PR #253, opened 2019-04-30
Add clip op

kahmed10:


src/targets/cpu/lowering.cpp
```diff
@@ -140,6 +140,21 @@ struct cpu_lrn
...
 };
 
+struct clip_op
+{
+    op::clip op;
+    std::string name() const { return "cpu::clip"; }
+    auto fcn() const
+    {
+        auto& max = op.max_val;
+        auto& min = op.min_val;
```
pfultz2: These variables can be values and not references.

src/targets/gpu/clip.cpp
```diff
@@ -0,0 +1,26 @@
...
+shape hip_clip::compute_shape(std::vector<shape> inputs) const
+{
+    inputs.pop_back();
+    return op.compute_shape(inputs);
+}
+
+argument
+hip_clip::compute(context& ctx, const shape& output_shape, const std::vector<argument>& args) const
+{
+    assert(output_shape == args.back().get_shape());
```
pfultz2: What is the reason for this assert?

kahmed10: This was copied from another compute function, so I don't know if it's necessary.

pfultz2: Well if the allocation is different we now have the `adjust_gpu_allocation` pass that will fix these allocations.

src/targets/gpu/include/migraphx/gpu/device/clip.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+void clip(hipStream_t stream,
+          const argument& result,
+          const argument& arg1,
+          const float& max,
```
pfultz2: These floats should be passed by value.

Issue #254, opened 2019-04-30
Commit 6f115a0 results in large increased in compile time


mvermeulen:
This week some of the models (densenet121, dpn92) started taking large amounts of time to compile.  I gave up after densenet had been running for >one hour.  The increase in compile time can be observed in a smaller test example - inceptioni32.onnx from torchvision.

Following are relative timings using:
```
prompt% time ./perf_onnx /home/mev/source/migraphx_onnx/torchvision/inceptioni32.onnx
```
With a preceding commit hash (a7461d7) the time to both compile and execute as well as runtime performance is
```
Rate: 10.0038/sec
Total time: 99.9621ms
Total instructions time: 103.339ms
Overhead time: 0.170403ms, -3.37656ms
Overhead: 0%, -3%

real	0m17.430s
user	0m12.705s
sys	0m1.974s
```
When the same is done with commit hash 6f115a0, the runtime is similar but total time (from compile) has gone up
```
Rate: 9.91012/sec
Total time: 100.907ms
Total instructions time: 104.257ms
Overhead time: 0.160416ms, -3.35007ms
Overhead: 0%, -3%

real	1m7.240s
user	1m2.554s
sys	0m1.888s
```

scxiao:
that commit has another issue that the following program will fail the MIOpen test by complaining no output after compilation.

    migraphx::program create_program() const
    {
        migraphx::program p;
        migraphx::shape s{migraphx::shape::float_type, {2, 3}};
        std::vector<float> data(2 * 3);
        std::iota(data.begin(), data.end(), 1.0f);
        auto l1 = p.add_literal(migraphx::literal(s, data));
        auto l2 = p.add_literal(migraphx::literal(s, data));
        p.add_instruction(migraphx::op::add{}, l1, l2);
        return p;
    };

pfultz2:
> that commit has another issue that the following program will fail the MIOpen test by complaining no output after compilation

That is expected since everything is const folded.

mvermeulen:
Fixed with PR 255.

PR #255, opened 2019-05-01
Fix slowness in eval

pfultz2:
This only runs eval if it knows all arguments can run eval. This avoids partially calculating some arguments even when the calculation is not needed. This should help address #254.

### DIFF
--- a/src/include/migraphx/instruction.hpp
+++ b/src/include/migraphx/instruction.hpp
@@ -74,7 +74,7 @@ struct instruction
 
     bool can_eval() const;
 
-    argument eval(bool check_eval=true) const;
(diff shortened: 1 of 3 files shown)


mvermeulen:
Latest changes address the compile-time slowdown.

On the models compiled in the performance benchmark suite, overall compile-time increases 3.8% which is within an expected margin of difference.  Runtime is still the same across these models as well, 0.0% change.

mvermeulen:
Note: Net effect on TF benchmark speeds:
resnet50_v1_50i64.pb - from 800.1 to 803.4 images/sec = +0.4%
resnet50_v2_50i64.pb - from 532.9 to 535.0 images/sec = +0.4%

PR #256, opened 2019-05-02
Depthwise convolution

kahmed10:
Reinterpret depthwise convolution as group convolution with group = number of input channels.


src/tf/tf.cpp
```diff
@@ -339,6 +340,62 @@ struct tf_parser
...
+            op.stride[0] = stride[2];
+            op.stride[1] = stride[3];
+        }
+        auto weights = args[1];
+        // check if weights are from a constant
+        if(weights->name() != "@param")
+        {
+            if(is_nhwc)
+            {
+                weights = prog.add_instruction(op::transpose{{1, 3, 0, 2}}, args[1]);
```
mei-ye: If weights is literal, you can transpose it into a new literal, then transpose instruction does not needed to be inserted into the program.

kahmed10: Deleted previous comment by accident, but the jist was that the weights have a different format: [out_channels, in_channels, h, w] compared to NCHW vs NHWC. I insert this additional transpose to get the literal into the correct format. The simplify_reshapes pass should replace the transposes with the correct reshape as well.

mei-ye: As long as the transpose is removed, I will have less concern.

pfultz2: It will be removed by the const propogation pass.

PR #257, opened 2019-05-02
Ensure reflect methods for all operators 

pfultz2:
This will throw a compile error on operators that are not empty and do not have a `reflect` method.

This will ensure that equality operators are checking if two operators are equivalent beyond just the name.


pfultz2:
The patch coverage is down because of deleting a lot of code in lowering.cpp, but the project coverage is increased so its safe to merge it now.

PR #258, opened 2019-05-09
Parse mean fix

kahmed10:
Still use GlobalAvgPool for now, but allow for a reduction in dims. The operator is used this way for nasnet.


scxiao:
also, more tests are need to cover code changes as suggested by Jinkens

PR #260, opened 2019-05-10
Int8 miopen call (gemm and conv)

scxiao:


src/targets/cpu/lowering.cpp
```diff
@@ -204,6 +204,60 @@ struct cpu_convolution
...
+            auto in_h = in[2];
+            auto in_w = in[3];
+
+            auto wei   = weights.get_shape().lens();
+            auto wei_n = wei[0];
+            auto wei_c = wei[1];
+            auto wei_h = wei[2];
+            auto wei_w = wei[3];
+
+            par_dfor(output_shape.lens()[0],
```
kahmed10: Can we replace this portion with a function shared with cpu_convolution? The only difference I see is the type of acc, other than that it looks the same.

src/targets/gpu/quant_convolution.cpp
```diff
@@ -0,0 +1,153 @@
...
+                                        &beta,
+                                        x_desc_vec4.get(),
+                                        arg_vec4_x.implicit());
+    if(status != miopenStatusSuccess)
+    {
+        MIGRAPHX_THROW("QUANT_CONVOLUTION: transform input tensor failed");
+    }
+
+    // pack input to vec4 format
+    status = miopenTransformTensor(ctx.get_stream().get_miopen(),
```
pfultz2: This should be done in a separate operator that is inserted during lowering.

src/targets/gpu/quant_gemm.cpp
```diff
@@ -0,0 +1,187 @@
...
+    auto dim_1      = n_dim - 1;
+    auto dim_0      = n_dim - 2;
+    auto arg_num    = args.size();
+    rocblas_int lda = args[0].get_shape().strides()[transa ? dim_1 : dim_0];
+    rocblas_int ldb = args[1].get_shape().strides()[transb ? dim_1 : dim_0];
+    rocblas_int ldc = args[arg_num - 1].get_shape().strides()[dim_0];
+
+    if(!transb)
+    {
+        device::pack_a(ctx.get_stream().get(), args[arg_num - 2], args[1]);
```
pfultz2: This should be done as a separate operator and inserted during lowering.

src/targets/gpu/quant_gemm.cpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    // need to pack A in this scenario, use the algorithm to pack B in the
+    // comment of the API
+    if(transa)
+    {
+        device::pack_b(ctx.get_stream().get(), args[arg_num - 3], args[0]);
```
pfultz2: This should be done as a separate operator and inserted during lowering.

src/targets/cpu/lowering.cpp
scxiao: there is another difference that input and output types are different in quant_convolution.

src/targets/gpu/quant_gemm.cpp
scxiao: done

src/targets/gpu/quant_convolution.cpp
scxiao: fixed

src/targets/gpu/quant_gemm.cpp
scxiao: fixed

PR #262, opened 2019-05-20
Enable more verbose TRACE_EVAL

pfultz2:
Setting `MIGRAPHX_TRACE_EVAL=2` will print out the outputs of each instruction.


mvermeulen:
I tried this with a perf_onnx example and got a segmentation violation
```
mev@chilecito:~/source/AMDMIGraphX/build/src/onnx$ env MIGRAPHX_TRACE_EVAL=2 ./perf_onnx /home/mev/source/migraphx_onnx/torchvision/resnet50i1.onnx 1 
[libprotobuf INFO google/protobuf/io/coded_stream.cc:610] Reading dangerously large protocol message.  If the message turns out to be larger than 2147483647 bytes, parsing will be halted for security reasons.  To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.
[libprotobuf WARNING google/protobuf/io/coded_stream.cc:81] The total number of bytes read was 102754719
Compiling ... 
Allocating params ... 
Running performance report ... 
Run instruction: @0 = check_context -> float_type, {}, {}
Ouput: 
Run instruction: scratch = @param:scratch -> float_type, {2408448}, {1}
Segmentation fault (core dumped)
```
Same program works when MIGRAPHX_TRACE_EVAL=1 is set.

Note: I have some smaller MNIST ONNX examples that wouldn't result in two million numbers printed that have similar symptoms for scratch space allocation, but reporting using perf_onnx since that can easily be reproduced.

PR #264, opened 2019-05-24
Add hip softmax

kahmed10:
Inefficient version of softmax to currently mitigate issues with MIOpen version not working. Either this version will be optimized in the future, or the MIOpen issue will be resolved.


src/tf/tf.cpp
```diff
@@ -339,6 +316,35 @@ struct tf_parser
             }
         }
 
+        if(contains(attributes, "padding"))
```
kahmed10: I have moved this section below strides as a side effect of another branch for testing. It adds another check for "valid" pad_mode, but otherwise is the same.

test/gpu/miopen.cpp
```diff
@@ -574,7 +574,7 @@ struct test_softmax : verify_program<test_softmax>
     migraphx::program create_program() const
     {
         migraphx::program p;
-        auto x = p.add_parameter("x", migraphx::shape{migraphx::shape::float_type, {5, 3, 4, 2}});
+        auto x = p.add_parameter("x", migraphx::shape{migraphx::shape::float_type, {5, 3, 1, 1}});
```
kahmed10: This test case was failing with the hip version of softmax. I think it only works for 2d cases at the moment.

kahmed10:
Changes were made to the implementation in order to support tf and pytorch's version of softmax.

src/include/migraphx/op/softmax.hpp
```diff
@@ -18,10 +18,23 @@ namespace op {
...
+    }
+
     std::string name() const { return "softmax"; }
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs}.has(1).only_dims(4);
+        check_shapes{inputs}.has(1).standard();
+        if(axis < 0 || axis >= inputs[0].lens().size())
+        {
+            MIGRAPHX_THROW("SoftMax: input axis value " + std::to_string(axis) +
```
scxiao: maybe we can add a shape test to trigger an exception for better code coverage.

kahmed10: I combined the shape tests for softmax and logsoftmax (with the edge cases on axis accounted for) with the latest commit.

scxiao: Thanks

src/targets/gpu/include/migraphx/gpu/softmax.hpp
```diff
@@ -1,16 +1,28 @@
...
+#include <migraphx/gpu/context.hpp>
+#include <utility>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
-struct context;
-
-struct miopen_softmax
```
pfultz2: Can we keep the miopen version of the operator?

kahmed10: fixed

PR #265, opened 2019-05-24
Transpose each layer of TF

pfultz2:
This will do a transpose at each layer to help ensure reshapes are setup correctly.

This is also fixes an issue in simplify_shapes where transpose were not reduced correctly.


kahmed10:
When merging #267 and #264, resnet50 v2 is failing.

pfultz2:
So this verified on: inception_resnet_v2, inception_v3, inception_v4, mobilenet_v1, resnet_v1_50, resnet_v2_50. So I think we can merge it in now.

src/driver/main.cpp
```diff
@@ -7,6 +7,14 @@
 #include <migraphx/onnx.hpp>
 #include <migraphx/stringutils.hpp>
 
+#include <migraphx/pass_manager.hpp>
+#include <migraphx/dead_code_elimination.hpp>
+#include <migraphx/eliminate_identity.hpp>
+#include <migraphx/eliminate_pad.hpp>
+#include <migraphx/propagate_constant.hpp>
+#include <migraphx/simplify_algebra.hpp>
+#include <migraphx/simplify_reshapes.hpp>
```
scxiao: Can we print out help info if no command line arguments?

mvermeulen:
Note: Gets same accuracy on TF/resnet50v2

PR #267, opened 2019-05-30
padding calculation for attribute

kahmed10:
Rewrite pad calculation to use for MIOpen calls to convolution and pooling. This link was used to derive the equation: http://deeplearning.net/software/theano/tutorial/conv_arithmetic.html


src/include/migraphx/op/pad_calc.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <utility>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+inline size_t calculate_padding(size_t weight_dim, size_t dilation)
+{
+    return (dilation * (weight_dim - 1)) / 2;
+}
```
pfultz2: Can we move this header out of the op directory since it is not an op?

kahmed10: fixed

PR #268, opened 2019-05-31
Logsoftmax modification

scxiao:
to be compatible with with pytorch output


src/include/migraphx/op/logsoftmax.hpp
```diff
@@ -30,7 +30,7 @@ struct logsoftmax
     shape compute_shape(std::vector<shape> inputs) const
     {
         check_shapes{inputs}.has(1).standard();
-        if(axis < 0 || axis > inputs[0].lens().size())
+        if(axis < 0 || axis >= inputs[0].lens().size())
```
kahmed10: This change isn't in #264, so the test case in op_shape_test.cpp I modified is accounting for the previous axis check. When either one gets merged in, we can change accordingly.

PR #269, opened 2019-06-05
Add unified driver

pfultz2:
This unifies the tf and onnx drivers into a single executable. This will do verifying and performance reports just like the original driver.


src/driver/main.cpp
```diff
@@ -0,0 +1,214 @@
...
+        auto m = c.params(p);
+        p.eval(m);
+        std::cout << p << std::endl;
+    }
+};
+
+struct perf : command<perf>
+{
+    compiler c;
+    unsigned n = 100;
```
kahmed10: Might be useful to print out what default value of n is set to.

src/driver/main.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+    program load()
+    {
+        program p;
+        if(type.empty())
+        {
+            if(ends_with(file, ".onnx"))
+                type = "onnx";
+            else
+                type = "tf";
```
kahmed10: tensorflow protobufs default to .pb file extensions, so I think we should do a check on the file for this

src/driver/argument_parser.hpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+};
+
+struct argument_parser
+{
+    struct argument
+    {
+        std::vector<std::string> flags;
+        std::function<bool(argument_parser&, const std::vector<std::string>&)> action{};
+        std::string type    = "";
```
kahmed10: can we rename type to something like file_type or graph_type? The function get_type_name<T>() in line 81 is used elsewhere in our code for a different purpose, and the names might get confusing.

pfultz2: This is the type of the variable so we can show it in the `--help` message it is unrelated to files or graphs. I can rename the `type` in the loader class to `file_type` to make it clearer.

src/driver/main.cpp
pfultz2: It prints out the default now with `--help` flag.

PR #273, opened 2019-06-14
Rnn optimization

scxiao:
1. combine matrix multiplication to improve performance
2. add to handle activation function names of all lower case and first letter capital


src/onnx/onnx.cpp
```diff
@@ -100,11 +100,17 @@ struct onnx_parser
 
     void init_actv_func()
     {
+        // Support name format of all lower case or the first letter capital
```
kahmed10: Is this necessary for the ops we have currently?

scxiao: Yes, the activate function is used in the rewrite_rnn function, we cannot just all the various parse_XXX function to process them...

PR #277, opened 2019-06-20
Padding fixes

kahmed10:
This PR fixes several issues with padding:
- Padding mode affected the calculation of the output shapes for convolution and pooling. Now it defaults to the padding_mode::default for calculating the shape. I left the actual pad modes in each operator for debugging purposes when calling print_program().
- The actual equation is modified to use the [tensorflow-onnx](https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/onnx_opset/nn.py#L154) version for calculation
- Padding for max pooling inserts std::lowest<> values instead of previously just 0

I also modified the CMakeLists.txt to fix the following protobuf issue:
```
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: (scc->visit_status.load(std::memory_order_relaxed)) == (SCCInfoBase::kRunning): 
```


src/targets/gpu/include/migraphx/gpu/softmax.hpp
```diff
@@ -34,7 +34,7 @@ struct miopen_softmax
         return migraphx::reflect(self.op, f);
     }
 
-    std::string name() const { return "gpu::softmax"; }
+    std::string name() const { return "gpu::miopen_softmax"; }
```
pfultz2: How about naming it `miopen::softmax`?

test/gpu/miopen.cpp
```diff
@@ -1573,6 +1573,26 @@ void manual_identity()
...
+    p.add_instruction(op, l0);
+    p.compile(migraphx::gpu::target{});
+    migraphx::program::parameter_map m;
+    for(auto&& x : p.get_parameter_shapes())
+    {
+        m[x.first] = migraphx::gpu::to_gpu(migraphx::generate_argument(x.second));
+    }
+    auto result = migraphx::gpu::from_gpu(p.eval(m));
+    std::cout << result << std::endl;
+}
```
pfultz2: This doesnt look like a test. The tests should use the `verify_program`.

src/tf/tf.cpp
```diff
@@ -567,21 +602,39 @@ struct tf_parser
...
+                }
+                else
+                {
+                    op.padding[0] = pads[0];
+                    op.padding[1] = pads[1];
+                }
             }
             else if(pad_mode.find("VALID") != std::string::npos)
             {
                 op.padding_mode = op::padding_mode_t::valid;
```
pfultz2: Dont we need to calculate padding for valid as well?

test/gpu/miopen.cpp
kahmed10: fixed with rename of test

src/tf/tf.cpp
kahmed10: According to the tensorflow-onnx link, they don't do anything for valid either. My guess is because tensorflow automatically inserts a pad (like in resnet v2) before performing convolution or pooling in order to make the input valid.

src/targets/gpu/include/migraphx/gpu/softmax.hpp
kahmed10: fixed

mvermeulen:
I ran this branch through imagenet.  Pass rates for mobilenet are 71.68% (top1) and 90.12% (top5)
```
50000 top1: 35841 top5: 45059
Overall - top1: 0.71682 top5: 0.90118
```

Will update this comment once my run of resnet50_v2 completes...

PR #278, opened 2019-06-24
Refactor device

pfultz2:
This should help simplify the device code so we can write more complex optimizations.


src/targets/gpu/device/include/migraphx/gpu/device/shape.hpp
```diff
@@ -0,0 +1,89 @@
...
+                const std::size_t slen   = s * len;
+                const std::size_t idx    = (i % slen) / s;
+                result += stride * idx;
+                s = slen;
+            }
+            return result;
+        }
+    }
+
+    MIGRAPHX_DEVICE_CONSTEXPR hip_index multi(std::size_t idx) const
```
scxiao: For the general shape, can we add a multi() function as
std::vector<std::size_t> multi(std::size_t idx) const?

pfultz2: We can add that to the host, but do we have a need for such method?

scxiao: cpu implementation of a few operator like the softmax need that. OK, I will add it in my code.

Issue #234, opened 2019-04-09
Change parse_axis to pass dims as parameter


kahmed10:
3d convolutions in tensorflow use NDHWC format, so handling 5 dims currently will not adjust to the correct format.

PR #284, opened 2019-06-25
Expand dims op

kahmed10:
This PR also addresses #234 because this may impact expand_dims.


src/tf/tf.cpp
```diff
@@ -117,17 +118,17 @@ struct tf_parser
...
-            auto new_idx         = parse_axis(i);
+            auto new_idx         = parse_axis(i, new_data.size());
             new_data.at(new_idx) = prev_data.at(i);
         }
         prev_data = new_data;
     }
 
     template <class T>
-    T parse_axis(const T& dim) const
+    T parse_axis(const T& dim, const size_t& num_dims) const
```
pfultz2: `num_dims` should be passed by value.

src/tf/tf.cpp
```diff
@@ -165,6 +166,7 @@ struct tf_parser
         add_mem_op("Const", &tf_parser::parse_constant);
         add_mem_op("Conv2D", &tf_parser::parse_conv);
         add_mem_op("DepthwiseConv2dNative", &tf_parser::parse_depthwiseconv);
+        add_mem_op("ExpandDims", &tf_parser::parse_expanddims);
```
pfultz2: The transpose flag should be set to false.

src/tf/tf.cpp
```diff
@@ -490,6 +492,25 @@ struct tf_parser
         return prog.add_instruction(op, {l0, new_weights});
     }
 
+    instruction_ref
+    parse_expanddims(const std::string&, const attribute_map&, std::vector<instruction_ref> args)
+    {
+        std::vector<size_t> input_dims = args[0]->get_shape().lens();
+        std::vector<int64_t> new_dims(input_dims.begin(), input_dims.end());
+        size_t num_dims = input_dims.size();
+        int32_t dim     = parse_axis(args[1]->eval().at<int32_t>(), num_dims);
```
pfultz2: The `parse_axis` function shouldn't be needed if we set the `transpose` flag to false.

src/tf/tf.cpp

src/tf/tf.cpp

src/tf/tf.cpp

PR #291, opened 2019-06-27
Add reduce sum operators

pfultz2:


src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp
```diff
@@ -78,9 +78,9 @@ __device__ auto block_reduce(index idx, Op op, T init, std::size_t n, F f)
     for(std::size_t s = 1; s < idx.nlocal(); s *= 2)
     {
         const std::size_t index = 2 * s * idx.local;
-        if(index < idx.nlocal())
+        if(index + s < idx.nlocal())
```
pfultz2: Why did you revert this change?

PR #304, opened 2019-07-04
Reduce mean operator

scxiao:
should be merged after PR#292 (argmax/argmin)


src/include/migraphx/op/reduce_sum.hpp
```diff
@@ -13,7 +14,7 @@ namespace op {
 
 struct reduce_sum
 {
-    std::vector<std::size_t> axes;
+    std::vector<int64_t> axes{};
```
pfultz2: The type should be `size_t` as the axes cannot be negative.

test/cpu_ops_test.cpp
```diff
@@ -1763,4 +1763,79 @@ TEST_CASE(reduce_sum_test12)
     EXPECT(results_vector == gold);
 }
 
+TEST_CASE(reduce_mean_test1)
```
kahmed10: can we rename these tests to reduce_mean_axis_{1,2,...}

kahmed10: I notice sum tests above have the same naming convention, but feel free to change them too

kahmed10: can we also get a test for no axis provided? The default should be to reduce all dimensions if no axis is provided, both for onnx and tf.

src/include/migraphx/op/reduce_sum.hpp
scxiao: There are negative input in the bert model, so keep it as int64_t

test/cpu_ops_test.cpp
scxiao: Added and add shape test to cover code change

src/include/migraphx/op/reduce_sum.hpp
kahmed10: Would it be better to reinterpret the negative input when parsing the protobuf? tf reduce_sum also takes in a negative value.

scxiao: I think this is just a design choice.

scxiao: Either way is fine, but we should be consistent across all our operators.

pfultz2: We usually use the negative indices in the operators.

test/op_shape_test.cpp
```diff
@@ -390,37 +390,111 @@ TEST_CASE(softmax) { test_softmax_variations<migraphx::op::softmax>(); }
...
+        throws_shape(migraphx::op::argmin{4}, input);
+    }
+}
+
+template <class T>
+void test_reduce_ops()
+{
+    {
+        migraphx::shape input{migraphx::shape::float_type, {2, 3, 4, 5}};
+        expect_shape(migraphx::shape{migraphx::shape::float_type, {1, 1, 1, 1}}, T{}, input);
```
kahmed10: This isn't the correct behavior in tensorflow. If no axis is provided, then the output should be a scalar. I believe it should be the same in onnx.

scxiao: According to Paul's suggestion, our operator definition does not have the keep_dim attribute, and the default behavior is to keep dim. Then in parse onnx, we keep_dim is false, we call squeeze to convert it to a scalar.

kahmed10: Ok, makes sense. If you want, you can leave a comment like: "operator keeps dims by default, unless user specifies in the protobuf"

scxiao: Will add a comment for that.

PR #305, opened 2019-07-05
Bert operators

scxiao:


src/onnx/onnx.cpp
```diff
@@ -899,6 +909,80 @@ struct onnx_parser
...
+            migraphx::shape s;
+            // empty input tensor, output is a scalar
+            if(args[0]->get_shape().elements() == 0)
+            {
+                s = migraphx::shape{type, {1}, {0}};
+            }
+            else
+            {
+                migraphx::argument in = args[0]->eval();
+                if(in.empty())
```
kahmed10: Maybe we should create a helper function that calls eval() on an input and throws an exception if eval() returns empty(). This seems to be quite repetitive otherwise throughout this file.

test/onnx/onnx_test.cpp
```diff
@@ -858,4 +858,104 @@ TEST_CASE(clip_test)
...
+    EXPECT(p == prog);
+}
+
+TEST_CASE(const_of_shape2)
+{
+    migraphx::program p;
+    migraphx::shape ss(migraphx::shape::int32_type, {3});
+    p.add_literal(migraphx::literal(ss, {2, 3, 4}));
+    migraphx::shape s(migraphx::shape::int64_type, {2, 3, 4});
+    std::vector<int64_t> vec(s.elements(), 10.0f);
```
kahmed10: the vector is type int64, but you are passing in a float value.

test/onnx/onnx_test.cpp
```diff
@@ -858,4 +858,104 @@ TEST_CASE(clip_test)
...
+    p.add_literal(migraphx::literal(s, vec));
+
+    auto prog = migraphx::parse_onnx("const_of_shape2.onnx");
+    EXPECT(p == prog);
+}
+
+TEST_CASE(const_of_shape3)
```
kahmed10: this test looks to be the same as const_of_shape1 except that the float value is 0 instead of 10. Is this test necessary?

test/onnx/onnx_test.cpp
```diff
@@ -858,4 +858,104 @@ TEST_CASE(clip_test)
...
+{
+    migraphx::program p;
+    auto l = p.add_parameter("x", migraphx::shape{migraphx::shape::half_type, {10}});
+    p.add_instruction(migraphx::op::convert{migraphx::shape::float_type}, l);
+
+    auto prog = migraphx::parse_onnx("cast_test.onnx");
+    EXPECT(p == prog);
+}
+
+TEST_CASE(const_of_shape1)
```
kahmed10: in general, can we rename these tests to be more clear for what they are testing, such as const_of_shape_float and const_of_shape_int64

src/onnx/onnx.cpp

test/onnx/onnx_test.cpp
scxiao: fixed

PR #310, opened 2019-07-09
add gpu implementation of the div operator

scxiao:


src/targets/gpu/device/div.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include <migraphx/gpu/device/nary.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+void div(hipStream_t stream, const argument& result, const argument& arg1, const argument& arg2)
+{
+    nary(stream, result, arg1, arg2)([](auto x, auto y) { return y / x; });
```
kahmed10: In tf and onnx, it's arg1 / arg2, so shouldn't it be x / y?

scxiao: If you look at how this function is called, you will know this is the correct way. There are tests to verify that.

kahmed10: Never mind, the argument order is flipped when calling compute(). Perhaps for clarity we should flip arg2 and arg1 in the definition (in sub as well), or at least be consistent with other operators like add and mul.

PR #311, opened 2019-07-09
Enable MIOpen Backend for HIP-Clang Compiler

aaronenyeshi:
We should also build the GPU libraries for HIP-Clang compiler.


pfultz2:
This should use `check_cxx_compiler_flag` instead as this wont work when using clang. And we need to use clang directly so that the correct flags will be passed to clang-tidy.

I think doing something like `check_cxx_compiler_flag(HAS_HIP "-x hip")` should enough.

aaronenyeshi:
@pfultz2 where is that function defined? How does it know HAS_HIP?

CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     endif()
 endif()
 
-if(CMAKE_CXX_COMPILER MATCHES ".*hcc")
+if(check_cxx_compiler_flag(HAS_HIP "-x hip"))
```
pfultz2: This needs to be separate lines:

```cmake
check_cxx_compiler_flag(HAS_HIP "-x hip")
if(HAS_HIP)
```

aaronenyeshi:
I tried using the check_cxx_compiler_flag option, and I am getting this with hip-clang compiler: 
-- Performing Test -x hip
-- Performing Test -x hip - Failed

pfultz2:
> I tried using the check_cxx_compiler_flag option, and I am getting this with hip-clang compiler

Hip clang doesnt recognize `-x hip` flag? I thought that was the flag needed to compile with hip-clang.

pfultz2:
I see it cant find the device files:

```
$ echo | /opt/rocm/hcc/bin/clang-9 -x hip -
clang-9: error: no such file or directory: 'hip.amdgcn.bc'
clang-9: error: no such file or directory: 'opencl.amdgcn.bc'
clang-9: error: no such file or directory: 'ocml.amdgcn.bc'
clang-9: error: no such file or directory: 'ockl.amdgcn.bc'
clang-9: error: no such file or directory: 'oclc_finite_only_off.amdgcn.bc'
clang-9: error: no such file or directory: 'oclc_daz_opt_off.amdgcn.bc'
clang-9: error: no such file or directory: 'oclc_correctly_rounded_sqrt_on.amdgcn.bc'
clang-9: error: no such file or directory: 'oclc_unsafe_math_off.amdgcn.bc'
clang-9: error: no such file or directory: 'oclc_isa_version_20.amdgcn.bc'
```

pfultz2:
We need to add `--cuda-host-only`. So we can do `check_cxx_compiler_flag(HAS_HIP "--cuda-host-only -x hip")` to detect hip-clang.

aaronenyeshi:
It seems both HCC's clang and HIP-Clang doesn't exit with positive.
HCC:
```
# echo | /opt/rocm/hcc/bin/clang --cuda-host-only -x hip -
/usr/lib/gcc/x86_64-linux-gnu/7.4.0/../../../x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
clang-9: error: linker command failed with exit code 1 (use -v to see invocation)
```
HIP-Clang
```
# echo | /opt/rocm/hip/bin/clang --cuda-host-only -x hip -
/usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

pfultz2:
> It seems both HCC's clang and HIP-Clang doesn't exit with positive.
HCC

You need to only use compile `-c`. This works `echo | /opt/rocm/hcc/bin/clang++ --cuda-host-only -x hip -c -`.

> So HIP-Clang CMake still having an issue:

I guess I should have read the documentation closer. The parameters are flipped. It should be `check_cxx_compiler_flag("--cuda-host-only -x hip" HAS_HIP)`. This works for me:

```
-- Performing Test HAS_HIP
-- Performing Test HAS_HIP - Success
```

aaronenyeshi:
Thank you so much, I've pushed an updated version, and it works with HIP-Clang as well.

PR #319, opened 2019-07-16
changes for bert to get the same result between cpu and gpu

scxiao:


src/onnx/onnx.cpp
```diff
@@ -257,14 +261,28 @@ struct onnx_parser
         return prog.add_instruction(op, std::move(args));
     }
 
-    instruction_ref
-    parse_softmax(const std::string&, const attribute_map&, std::vector<instruction_ref> args)
+    //    instruction_ref
```
kahmed10: can we have a separate PR that changes parse_softmax to use the new implementation?

scxiao: fixed. move this part to another PR

PR #322, opened 2019-07-17
Parse softmax using new implementation

scxiao:


kahmed10:
Can you change the tf parsing in this PR as well?

src/targets/gpu/gemm.cpp
```diff
@@ -171,6 +171,35 @@ shape miopen_gemm::compute_shape(const std::vector<shape>& inputs) const
 {
     std::vector<shape> input_shapes(inputs.begin(), inputs.begin() + inputs.size() - 1);
     check_shapes{input_shapes}.not_broadcasted();
+    auto a_strides = inputs[0].strides();
+    auto dim_0     = a_strides.size() - 2;
+    if(a_strides.size() > 2)
+    {
+        if(!std::all_of(a_strides.begin(), a_strides.begin() + dim_0, [&](auto batch_size) {
+               return std::all_of(a_strides.begin() + dim_0, a_strides.end(), [&](auto data_size) {
```
pfultz2: This doesnt seem to check for transposed batches. It would be better to get the max stride of the last two dimensions and then do `adjacent_find` to make sure each stride is greater than the max stride and greater than its previous stride.

src/tf/tf.cpp
```diff
@@ -715,14 +715,19 @@ struct tf_parser
...
+    instruction_ref parse_softmax(const std::string&,
+                                  const attribute_map& attributes,
+                                  std::vector<instruction_ref> args)
     {
-        auto dims = args.front()->get_shape().lens();
-        auto r =
-            prog.add_instruction(op::reshape{{long(dims[0]), long(dims[1]), 1, 1}}, args.front());
-        auto s = prog.add_instruction(op::softmax{}, r);
-        return prog.add_instruction(op::reshape{{long(dims[0]), long(dims[1])}}, s);
+        int axis = 1;
```
kahmed10: In tf, the default axis is actually -1, indicating to perform softmax on the last dimension. Also, if the input is a transpose, then softmax is currently failing. This is my current workaround in another branch, feel free to use it to make modifications:
```
add_mem_op("Softmax", &tf_parser::parse_softmax<op::softmax>, false);
```
```
    // template to facilitate the logsoftmax later
    template <class Op>
    instruction_ref parse_softmax(const std::string&,
                                  const attribute_map& attributes,
    {
        int axis = -1;
        auto num_dims = args[0]->get_shape().lens().size();
        if(contains(attributes, "axis"))
        {
            axis = static_cast<int>(attributes.at("axis").i());
        }
        if(axis < 0)
        {
            axis += num_dims;
        }

        return prog.add_instruction(Op{axis}, make_contiguous(args[0]));
    }
```

PR #325, opened 2019-08-01
add capture operator

scxiao:


src/quantization.cpp
```diff
@@ -103,5 +123,80 @@ void quantize(program& prog, const std::vector<std::string>& ins_names)
...
+// For the input of each input argument, we need to insert a
+// capture operator to compute the scale and shift
+void capture_arguments(program& prog,
+                       const std::vector<std::string>& ins_names,
+                       std::function<void(std::size_t, std::vector<argument>)> func)
+{
+
+    size_t num_quant_params = 0;
+    // the int8 quantization only support dot and convolution
+    std::vector<std::string> op_names = {"dot", "convolution"};
```
kahmed10: I think using a std::set would be better because then you could do something like if(not op_names == ins_names) MIGRAPHX_THROW(...) instead of doing find() and all_of()

src/include/migraphx/generate.hpp
```diff
@@ -82,6 +82,8 @@ std::vector<T> generate_tensor_data(const migraphx::shape& s, unsigned long seed
 {
     std::vector<T> result(s.elements());
     std::generate(result.begin(), result.end(), xorshf96_generator<T>{seed});
+    // divide a value to avoid integer overflow
+    std::transform(result.begin(), result.end(), result.begin(), [](auto i) { return i / 32; });
```
pfultz2: The max values on signed integers are 32, so this will make everything a 1 or 0. I dont think we want that.

pfultz2: And changes such as this, should be done in the `normalize` function.

src/include/migraphx/op/binary.hpp
```diff
@@ -30,23 +30,29 @@ struct binary : op_name<Derived>
...
         auto s2 = args[1].get_shape();
-        visit_all(result, args[0], args[1])([&](auto output, auto input1, auto input2) {
-            if(s1 == s2 and input1.get_shape().packed() and input2.get_shape().packed())
-            {
+        if(s1 == s2 and s1.packed())
+        {
+            shape std_shape{s1.type(), s1.lens()};
+            argument std_result{std_shape, result.data()};
+            argument std_arg0{std_shape, args[0].data()};
+            argument std_arg1{std_shape, args[1].data()};
```
pfultz2: Now thats a creative way to workaround the iterator problem.

src/include/migraphx/op/convert.hpp
```diff
@@ -20,11 +20,14 @@ namespace op {
 struct convert : unary<convert>
 {
     shape::type_t target_type = shape::half_type;
+    float scale               = 1.0f;
+    float shift               = 0.0f;
```
pfultz2: Why is scale and shift added here? And these should be arguments since they dont affect the shape.

src/include/migraphx/op/convert.hpp
```diff
@@ -35,10 +38,22 @@ struct convert : unary<convert>
 
     auto apply() const
     {
-        return [](auto x) { return x; };
+        return [&](auto x) {
+            float res = scale * x + shift;
```
pfultz2: This should be done with multiply and add operators.

src/include/migraphx/op/convert.hpp
```diff
@@ -35,10 +38,22 @@ struct convert : unary<convert>
...
+            if(target_type == shape::int8_type)
+            {
+                int factor = (res >= 0.0f) ? 1 : -1;
+                res        = res + factor * 0.5f;
+                res        = res > 127.0f ? 127.0f : res;
+                res        = res < -128.0f ? -128.0f : res;
```
pfultz2: Dont we have a clamp operator that does this?

kahmed10: We have the [clip](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/master/src/include/migraphx/op/clip.hpp) op that can be used for 47 and 48 above.

src/quantization.cpp
scxiao: The inputs ins_name could be a subset of op_name, they do not have to be the same.

src/include/migraphx/generate.hpp
scxiao: We already did that in the normalize function, so just remove it here.

src/include/migraphx/op/convert.hpp
scxiao: removed the scale and shift.

src/include/migraphx/op/convert.hpp
scxiao: This is going to be implemented in another PR for the int8_quantization.

src/include/migraphx/generate.hpp
```diff
@@ -82,6 +82,7 @@ std::vector<T> generate_tensor_data(const migraphx::shape& s, unsigned long seed
 {
     std::vector<T> result(s.elements());
     std::generate(result.begin(), result.end(), xorshf96_generator<T>{seed});
+    std::transform(result.begin(), result.end(), result.begin(), [](auto i) { return i; });
```
pfultz2: This seems to do nothing.

src/quantization.cpp
kahmed10: you could use std::includes(op_names.begin(), op_names.end(), ins_names.begin(), ins_names.end()) to check for a subset

pfultz2: Using `std::includes` is a good suggestion, but they must be sorted first, which can be done by putting it in a `std::set` or calling `std::sort` on the vector.

src/include/migraphx/program.hpp
```diff
@@ -126,6 +126,9 @@ struct program
     friend bool operator==(const program& x, const program& y);
     friend bool operator!=(const program& x, const program& y) { return !(x == y); }
 
+    std::shared_ptr<std::vector<std::pair<float, float>>> int8_quant_params =
```
pfultz2: These should not be stored in the program. This should be stored in an auxiliary data structure.

src/quantization.cpp
```diff
@@ -103,5 +116,80 @@ void quantize(program& prog, const std::vector<std::string>& ins_names)
...
+        // scale and shift is need for only int8 type, and we do not
+        // consider shift, so set shift to 0
+        std::vector<float> vec_val;
+        args.front().visit([&](auto output) { vec_val.assign(output.begin(), output.end()); });
+        auto max_val = *std::max_element(vec_val.begin(), vec_val.end());
+        auto min_val = *std::min_element(vec_val.begin(), vec_val.end());
+        auto max_abs = std::max(std::fabs(max_val), std::fabs(min_val));
+
+        param_pair.first                     = 127.0f / max_abs;
+        (*prog.int8_quant_params)[ins_index] = param_pair;
```
pfultz2: The quant params can be allocated outside of this lambda as a shared_ptr, and captured by value. Then this function should return the shared_ptr.

src/quantization.cpp
```diff
@@ -79,21 +100,13 @@ void quantize(program& prog, const std::vector<std::string>& ins_names)
...
-            {
-                // check the dead code case to avoid assert
-                bool output_empty = ins->outputs().empty();
-                auto ins_orig_type =
-                    prog.insert_instruction(std::next(ins), op::convert{orig_type}, ins);
-                if(!output_empty)
-                {
-                    prog.replace_instruction(ins, ins_orig_type);
-                }
+                prog.replace_instruction(ins, ins_orig_type);
```
pfultz2: The `quantize` function needs to call the `capture_arguments` function.

src/py/migraphx_py.cpp
```diff
@@ -186,6 +187,11 @@ PYBIND11_MODULE(migraphx, m)
         migraphx::quantize(p, ins_names);
     });
     m.def("quantize", [](migraphx::program& p) { migraphx::quantize(p, {"all"}); });
+    m.def("capture_arguments", [](migraphx::program& p, const std::vector<std::string>& ins_names) {
+        migraphx::capture_arguments(p, ins_names);
```
pfultz2: Why is this exposed in the python API? `capture_arguments` is only needed by the `quantize` function.

PR #330, opened 2019-08-12
Replace pooling with reduce_mean where possible

pfultz2:


kahmed10:
Can we also check in the parser for pooling ops to check if it meets the conditions to substitute for reduce_mean?

pfultz2:
> Can we also check in the parser for pooling ops to check if it meets the conditions to substitute for reduce_mean?

There is the `rewrite_pooling` pass that does that. Otherwise, we would need to write this twice in both of the frontends.

kahmed10:
> > Can we also check in the parser for pooling ops to check if it meets the conditions to substitute for reduce_mean?
> 
> There is the `rewrite_pooling` pass that does that. Otherwise, we would need to write this twice in both of the frontends.

I think we should revisit merging the frontend parser functions sometime in the future, but this will work for now.

PR #334, opened 2019-08-15
Bert ops

kahmed10:


src/tf/tf.cpp
```diff
@@ -804,30 +823,75 @@ struct tf_parser
...
             shrink_axis_mask = static_cast<uint32_t>(attributes.at("shrink_axis_mask").i());
 
+        for(size_t i = 0; i < num_axes; i++)
+        {
+            // the LSB corresponds to axis 0 when determining which axes to begin
+            if(((begin_mask >> i) & bitwise_compare) == 1)
+                begin_axes.push_back(1);
+            else
+                begin_axes.push_back(0);
+        }
```
pfultz2: Can you make this a function that takes the mask and returns the vector?

src/tf/tf.cpp
```diff
@@ -867,10 +929,16 @@ struct tf_parser
         if(instructions.count(name) == 0)
         {
             auto&& node = nodes.at(name);
+            // assert ops ignored
+            if(node.op() == "Assert" or contains(name, "Assert"))
```
pfultz2: Is it possible to add a test for this?

kahmed10: added

PR #336, opened 2019-08-17
Bugs fix related bert (batch running and quantization)

scxiao:


src/onnx/onnx.cpp
```diff
@@ -490,6 +490,16 @@ struct onnx_parser
         {
             axis = parse_value(attributes.at("axis")).at<int>();
         }
+
```
kahmed10: Can we make a function that adds a contiguous to the program if the arg isn't in standard shape? This can be done for parse_reshape() as well.

src/onnx/onnx.cpp
```diff
@@ -437,11 +447,7 @@ struct onnx_parser
             s.visit([&](auto v) { copy(v, std::back_inserter(op.dims)); });
         }
 
-        if(!args[0]->get_shape().standard())
-        {
-            args[0] = prog.add_instruction(op::contiguous{}, args[0]);
-        }
-
+        args[0] = make_contiguous(args[0]);
         return prog.add_instruction(op, args[0]);
```
pfultz2: Can we just do `return prog.add_instruction(op, make_contiguous(args[0]));`?

src/onnx/onnx.cpp
```diff
@@ -490,6 +496,9 @@ struct onnx_parser
         {
             axis = parse_value(attributes.at("axis")).at<int>();
         }
+        args[0] = make_contiguous(args[0]);
+        args[1] = make_contiguous(args[1]);
+
         op::gather op{axis};
         return prog.add_instruction(op, std::move(args));
```
pfultz2: Can we just do `prog.add_instruction(op, make_contiguous(args[0]), make_contiguous(args[1]))` here?

src/onnx/onnx.cpp

src/onnx/onnx.cpp
scxiao: changed accordingly

src/onnx/onnx.cpp
scxiao: changed accordingly

PR #348, opened 2019-08-28
Int8 quantize

scxiao:


src/include/migraphx/quantization.hpp
```diff
@@ -20,9 +22,30 @@ void quantize(program& prog);
...
+    return capture_arguments_impl(prog, t, ins_names);
+}
+
+void quantize_int8(program& prog,
+                   const target& t,
+                   std::vector<program::parameter_map>& calibration_args);
+void quantize_int8(program& prog,
+                   const target& t,
+                   std::vector<program::parameter_map>& calibration_args,
+                   const std::vector<std::string>& ins_names);
```
pfultz2: Can you use a default parameter for `ins_names`?

src/include/migraphx/quantization.hpp
```diff
@@ -20,9 +22,30 @@ void quantize(program& prog);
...
-std::shared_ptr<std::vector<std::pair<float, float>>> capture_arguments(program& prog);
+capture_arguments(program& prog, T&& t, const std::vector<std::string>& ins_names = {"dot"})
+{
+    static_assert(std::is_same<std::remove_cv_t<std::remove_reference_t<T>>, target>{} &&
+                      std::is_lvalue_reference<T>{},
+                  "Dangling reference to target!");
+    return capture_arguments_impl(prog, t, ins_names);
+}
+
+void quantize_int8(program& prog,
```
pfultz2: Shouldn't the `quantize` function be renamed to `quantize_fp16` as well?

src/include/migraphx/quantization.hpp
```diff
@@ -20,9 +22,30 @@ void quantize(program& prog);
...
+                   const target& t,
+                   std::vector<program::parameter_map>& calibration_args);
+void quantize_int8(program& prog,
+                   const target& t,
+                   std::vector<program::parameter_map>& calibration_args,
+                   const std::vector<std::string>& ins_names);
+void quantize_int8(program& prog,
+                   const std::vector<std::string>& ins_names,
+                   const std::vector<std::pair<float, float>>& quant_params);
```
pfultz2: This overload should just be in the cpp file.

src/include/migraphx/quantization.hpp
```diff
@@ -23,12 +23,14 @@ std::size_t capture_arguments(program& prog,
                               const std::vector<std::string>& ins_names,
                               const std::function<void(std::size_t, std::vector<argument>)>& func);
 
-std::shared_ptr<std::vector<std::pair<float, float>>> capture_arguments_impl(
-    program& prog, const target& t, const std::vector<std::string>& ins_names = {"dot"});
+std::shared_ptr<std::vector<std::pair<float, float>>>
+capture_arguments_impl(program& prog,
+                       const target& t,
+                       const std::vector<std::string>& ins_names = {"dot", "convolution"});
```
pfultz2: This is not needed for the impl function.

src/quantization.cpp
```diff
@@ -170,13 +308,13 @@ void quantize_int8(program& prog,
...
-                  << std::endl;
-    }
-    std::cout << std::endl;
+    // for(size_t i = 0; i < quant_params.size(); i++)
+    // {
+    //     auto param = quant_params.at(i);
+    //     std::cout << "index = " << i << ", scale = " << param.first << "\t" << param.second
+    //               << std::endl;
+    // }
+    // std::cout << std::endl;
```
pfultz2: Should this be guarded by an env variable?

src/include/migraphx/quantization.hpp
scxiao: fixed.

src/include/migraphx/quantization.hpp
scxiao: changed

src/include/migraphx/quantization.hpp
scxiao: Two reasons to add this API 1) with this API user can use their own scales for the quantization. So they can play with any algorithms to compute the scale. 2) I need this API to wrap up various unit test to cover all code changes.

src/include/migraphx/quantization.hpp
scxiao: we have a template function in this head file hat calls the impl function, so I think we need to have it here.

src/quantization.cpp
scxiao: fixed

src/include/migraphx/quantization.hpp
```diff
@@ -6,23 +6,46 @@
...
+#include <migraphx/env.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 struct program;
 
-void quantize(program& prog, const std::vector<std::string>& ins_names);
-void quantize(program& prog);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_INT8_QUANTIZATION_PARAMS)
```
pfultz2: Can this be put in the .cpp file?

src/quantization.cpp
```diff
@@ -115,7 +161,279 @@ void quantize(program& prog, const std::vector<std::string>& ins_names)
...
+// -128 ~ 127. To convert the float or double to int8, we need a scale and
+// a shift, then the convert can be done as v_int8 = fp * scale + shift.
+// To simplify the changes, we consider shift as 0.0f for now.
+void quantize_int8(program& prog,
+                   const std::vector<std::pair<float, float>>& quant_params,
+                   const std::vector<std::string>& ins_names)
+{
+    // For now, we only support the int8 quantization of gemm and convolution
+    std::vector<std::string> op_names = {"dot", "convolution"};
+    if(!std::all_of(ins_names.begin(), ins_names.end(), [&](auto name) {
```
kahmed10: We discussed in a previous PR that we can use a std::set for the op names, and then you can do std::includes for checking subsets of the ops rather than the logic you have currently. Are you still planning to do this?

src/quantization.cpp
```diff
@@ -115,7 +161,279 @@ void quantize(program& prog, const std::vector<std::string>& ins_names)
...
+        auto stride       = conv_op.stride;
+        auto dilation     = conv_op.dilation;
+        auto padding_mode = conv_op.padding_mode;
+        auto group        = conv_op.group;
+        auto adjust_factor =
+            std::round(1.0f / (ins_quant_params[0].first * ins_quant_params[1].first));
+
+        auto quant_conv = prog.insert_instruction(
+            ins,
+            op::quant_convolution{padding, stride, dilation, padding_mode, group},
```
kahmed10: Would it be worth making a constructor overload for quant_convolution that takes in a conv op and initializes the other fields, rather than getting all of the member variables (lines 260-264) in this function?

src/include/migraphx/quantization.hpp
scxiao: fixed

src/quantization.cpp
scxiao: Thanks. Changed accordingly.

src/quantization.cpp
scxiao: If we add such a constructor, I think we need another 6 overload constructors to handle the different scenarios when this operator is initialized. For now, looks like it is the only place that this scenario is used. If there are other locations we have these scenario, we can add it later.

PR #350, opened 2019-08-29
Refactor reduce ops

scxiao:


src/include/migraphx/op/reduce_op.hpp
```diff
@@ -0,0 +1,181 @@
...
+                shape& batch_shape,
+                std::vector<int64_t>& tuned_axes,
+                std::vector<std::size_t>& out_idx,
+                tensor_view<T>& output,
+                Op op) const
+    {
+        auto data_idx = out_idx;
+        T val         = op.init();
+        shape_for_each(batch_shape, [&](auto b_idx) {
+            for(auto axis : tuned_axes)
```
kahmed10: can we make a function that iterates over the tuned axes and returns the new axis data for both lines 141 and 157?

src/targets/gpu/include/migraphx/gpu/reduce_op.hpp
```diff
@@ -0,0 +1,77 @@
...
+    static auto reflect(Self& self, T f)
+    {
+        return migraphx::reflect(self.op, f);
+    }
+
+    std::string name() const
+    {
+        const std::string& name = get_type_name<Derived>();
+        // search the namespace gpu (::gpu::)
+        auto pos_ns = name.find("::gpu::");
```
kahmed10: can we use the utility function contains(name, "::gpu::") for this?

src/targets/gpu/include/migraphx/gpu/reduce_op.hpp
```diff
@@ -0,0 +1,77 @@
...
+            {
+                return std::string("gpu::") + name.substr(pos_name + 4);
+            }
+            else
+            {
+                return name.substr(pos_ns + 2);
+            }
+        }
+
+        return "unknown";
```
kahmed10: Can we have a different name for "unknown" to indicate it is related to the reduce op?

src/targets/gpu/include/migraphx/gpu/reduce_op.hpp
scxiao: We may not be able to since we need to pos info returned from the find function, while contain() does not return that. 

As for the purpose, the name returned from get_type_name() contains other info like namespace that we donot want. These code is to retrieve the actually operator name we want.

src/include/migraphx/op/reduce_op.hpp
scxiao: fixed.

src/targets/gpu/include/migraphx/gpu/reduce_op.hpp
kahmed10: can you clarify this behavior of removing namespace info in the code?

scxiao: for example, the name returned from get_type_name<Derived>() is "migraphx::version_1::gpu::hip_reduce_sum", this function is to extract the part "reduce_mean", then with the prefix "gpu::", we get the operator name "gpu::reduce_mean"

kahmed10: Sorry... I meant clarify to the reader by adding comments

src/targets/gpu/include/migraphx/gpu/reduce_op.hpp
scxiao: fixed

src/targets/gpu/include/migraphx/gpu/reduce_op.hpp
scxiao: OK, will add some comments for that.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -0,0 +1,186 @@
...
+    {
+        return x + y;
+    }
+
+    auto init() const { return zero(); }
+
+    auto output() const
+    {
+        return [=](auto val) { return val; };
+    }
```
pfultz2: These operators should be defined in the derived classes. We also need `input()` method as well so we can implement sum of squares.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -0,0 +1,186 @@
...
+    {
+        return T{0};
+    }
+};
+
+struct sum_op
+{
+    std::size_t elem_num = 1;
+    template <class T>
+    T operator()(T x, T y) const
```
pfultz2: Instead this should be a function named `op` and returns a lambda to make it consistent with the rest of the functions.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -0,0 +1,186 @@
...
+    template <class T>
+    operator T() const
+    {
+        return T{0};
+    }
+};
+
+struct sum_op
+{
+    std::size_t elem_num = 1;
```
pfultz2: This field is not used and can be removed.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -0,0 +1,186 @@
...
+    T operator()(T x, T y) const
+    {
+        return x + y;
+    }
+
+    auto init() const { return zero(); }
+
+    auto output() const
+    {
+        return [=](auto val) { return val / elem_num; };
```
pfultz2: Actually, this should probably be:

```cpp
    auto output(const argument& input) const
    {
        std::size_t elem_num = input.elements();
        return [=](auto val) { return val / elem_num; };
    }
```

pfultz2: Or maybe it should be just the shape.

src/include/migraphx/op/reduce_op.hpp
scxiao: fixed

src/include/migraphx/op/reduce_op.hpp
scxiao: fixed

src/include/migraphx/op/reduce_op.hpp
scxiao: changed accordingly

src/include/migraphx/op/reduce_op.hpp
scxiao: changed accordingly

src/include/migraphx/op/reduce_mean.hpp
```diff
@@ -1,110 +1,36 @@
...
 namespace op {
 
-struct reduce_mean
+struct mean_op
 {
-    std::vector<std::int64_t> axes{};
-
-    template <class Self, class F>
-    static auto reflect(Self& self, F f)
+    auto op() const
```
pfultz2: These methods should be in the `reduce_mean` class not in a seperate class.

src/include/migraphx/op/reduce_mean.hpp
```diff
@@ -1,110 +1,36 @@
...
-                if(axis < 0)
-                {
-                    axis += n_dim;
-                }
-            }
-        }
-
-        return tuned_axes;
-    }
+    auto init() const { return zero(); }
```
pfultz2: Can we provide this method in the base `reduce_op` class?

src/include/migraphx/op/reduce_mean.hpp
scxiao: fixed

PR #354, opened 2019-09-09
Add fusions for sigmoid and tanh

kahmed10:


src/targets/gpu/fuse_ops.cpp
```diff
@@ -517,11 +507,13 @@ void fuse_ops::apply(program& p) const
     // clang-format off
     match::find_matches(p, find_triadd{});
     match::find_matches(p, 
-        find_conv_bias_relu{ctx},
+        // find_conv_bias_relu{ctx},
```
pfultz2: Should this still be enabled?

PR #355, opened 2019-09-09
rename test_gpu_miopen

kahmed10:


kahmed10:
> I would suggest rename it to ops.cpp since cmake will add prefix test_gpu_. With the current name, the test name is test_gpu_gpu_ops_test

fixed

PR #356, opened 2019-09-10
Remove gemm copy and simplify rocblas call

scxiao:


src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -1,7 +1,10 @@
 #ifndef MIGRAPHX_GUARD_RTGLIB_GEMM_HPP
 #define MIGRAPHX_GUARD_RTGLIB_GEMM_HPP
 
+#include <rocblas-types.h>
```
pfultz2: These includes shouldn't go into the header file.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -10,21 +13,196 @@ namespace gpu {
...
+
+        return rocblas_datatype_f32_r;
+    }
+
     argument
-    compute(context& ctx, const shape& output_shape, const std::vector<argument>& args) const;
-    void batch_not_transposed(const std::vector<std::size_t>& strides) const;
+    compute(context& ctx, const shape& output_shape, const std::vector<argument>& args) const
+    {
+        bool transa     = args[0].get_shape().transposed();
```
pfultz2: I would prefer the compute method to go into a cpp file, or at least put this into a function that is implemented in the cpp file.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -10,21 +13,196 @@ namespace gpu {
...
+        std::vector<shape> in_shapes(inputs);
+        in_shapes.pop_back();
+        check_shapes{in_shapes}.not_broadcasted();
+        batch_not_transposed(inputs[0].strides());
+        batch_not_transposed(inputs[1].strides());
+
+        return op.compute_shape(in_shapes);
+    }
+
+    rocblas_datatype get_type(shape::type_t type) const
```
pfultz2: This doesnt need to be a member function.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -10,21 +13,63 @@ namespace gpu {
...
     static auto reflect(Self& self, F f)
     {
         return migraphx::reflect(self.op, f);
     }
 
-    std::string name() const { return "gpu::gemm"; }
-    shape compute_shape(const std::vector<shape>& inputs) const;
+    std::string name() const
+    {
+        if(op.name().find("quant_") != std::string::npos)
```
kahmed10: you can use contains() for the if condition. Also, you don't need an else statement here and can just return "gpu::gem".

scxiao: fixed

src/targets/gpu/include/migraphx/gpu/gemm.hpp
scxiao: moved to the cpp file

PR #362, opened 2019-09-13
Fix HIP-Clang GPU build issues

aaronenyeshi:
Add missing device attributes for GPU functions. GPU functions must be annotated with __device__ in HIP. Also, fix a few warning: nvcc does not allow '__device__' to appear after '()' in lambdas.

Here are the test results, I will file bugs for the failures:

99% tests passed, 2 tests failed out of 182

Total Test time (real) =  26.47 sec

### DIFF
--- a/src/targets/gpu/device/softmax.cpp
+++ b/src/targets/gpu/device/softmax.cpp
@@ -26,19 +26,19 @@ void softmax(hipStream_t stream, const argument& result, const argument& arg, in
         const std::size_t block_size     = compute_block_size(batch_item_num, max_block_size);
         gs_launch(stream,
                   batch_shape.elements() * block_size,
-                  block_size)([=](auto i, auto idx) __device__ {
+                  block_size)([=] __device__ (auto i, auto idx) {
             auto data_idx = batch.multi(i / block_size);
             using type    = device_type<std::remove_cv_t<typename decltype(input)::value_type>>;
             type init     = lowest();
 
             auto batch_max = block_reduce<max_block_size>(
-                idx, max{}, init, batch_item_num, [&](auto j) __device__ {
+                idx, max{}, init, batch_item_num, [&] __device__ (auto j) {
                     data_idx[axis] = j;
                     return input[data_idx];
                 });
 
             auto batch_sum =
-                block_reduce<max_block_size>(idx, sum{}, 0, batch_item_num, [&](auto j) __device__ {
+                block_reduce<max_block_size>(idx, sum{}, 0, batch_item_num, [&] __device__ (auto j) {
                     data_idx[axis] = j;
                     auto val       = input[data_idx] - batch_max;
                     return ::exp(to_hip_type(val));
--- a/src/targets/gpu/device/sub.cpp
+++ b/src/targets/gpu/device/sub.cpp
@@ -8,7 +8,7 @@ namespace device {
 
 void sub(hipStream_t stream, const argument& result, const argument& arg1, const argument& arg2)
 {
-    nary(stream, result, arg1, arg2)([](auto x, auto y) { return x - y; });
+    nary(stream, result, arg1, arg2)([] __device__ (auto x, auto y) { return x - y; });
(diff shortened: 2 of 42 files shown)


src/targets/gpu/device/softmax.cpp
```diff
@@ -26,19 +26,19 @@ void softmax(hipStream_t stream, const argument& result, const argument& arg, in
...
             auto batch_max = block_reduce<max_block_size>(
-                idx, max{}, init, batch_item_num, [&](auto j) __device__ {
+                idx, max{}, init, batch_item_num, [&] __device__ (auto j) {
                     data_idx[axis] = j;
                     return input[data_idx];
                 });
 
             auto batch_sum =
-                block_reduce<max_block_size>(idx, sum{}, 0, batch_item_num, [&](auto j) __device__ {
+                block_reduce<max_block_size>(idx, sum{}, 0, batch_item_num, [&] __device__ (auto j) {
```
pfultz2: Attributes for lambdas go after the parameters, specifiers, and `noexcept`.

src/targets/gpu/device/sub.cpp
```diff
@@ -8,7 +8,7 @@ namespace device {
 
 void sub(hipStream_t stream, const argument& result, const argument& arg1, const argument& arg2)
 {
-    nary(stream, result, arg1, arg2)([](auto x, auto y) { return x - y; });
+    nary(stream, result, arg1, arg2)([] __device__ (auto x, auto y) { return x - y; });
```
pfultz2: Why is `__device__` needed for this? It is only instantiated from a device function and is also `constexpr`.

yxsamliu: In CUDA/HIP syntax, kernel and device function can only call device function. By default this lambda is a host function, which cannot be called by a kernel or device function.

pfultz2: But the lambda is `constexpr` so it should run on either host or device.

src/targets/gpu/device/softmax.cpp
aaronenyeshi: It seems in CUDA/nvcc syntax, you will see warnings if you do that.
```
In file included from /root/AMDMIGraphX/src/targets/gpu/device/argmax.cpp:7:
In file included from /root/AMDMIGraphX/src/targets/gpu/include/migraphx/gpu/device/arg_op.hpp:9:
/root/AMDMIGraphX/src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp:227:85: warning: nvcc does not allow '__device__' to appear after '()' in lambdas [-Wcuda-compat]
        gs_launch(stream, nelements * block_size, block_size)([=](auto i, auto idx) __device__ {
```

mhbliao:
By default, lambda is not treated as constexpr until c++17.

src/targets/gpu/device/softmax.cpp
pfultz2: Cant we just add `-Wno-cuda-compat` to disable the warning? The issue is that it might break static analyzers that expect C++.

src/targets/gpu/device/sub.cpp
pfultz2: Isn't there a flag we can enable with clang to make `constexpr` functions be `__device__ __host__` as default?

pfultz2:
> By default, lambda is not treated as constexpr until c++17.

I see, then can we enable C++17 when using hip?

pfultz2:
This needs to run on our CI so we can merge it in. Currently, it seems to not work on PRs from forks. Can you create a branch in MIGraphX's repo and open a PR from there?

aaronenyeshi:
> This needs to run on our CI so we can merge it in. Currently, it seems to not work on PRs from forks. Can you create a branch in MIGraphX's repo and open a PR from there?

I've just moved it and opened PR #366

pfultz2:
Thanks

PR #363, opened 2019-09-17
Reduce_min/max operators

scxiao:


test/onnx/gen_onnx.py
```diff
@@ -1044,6 +1044,33 @@ def reducemean_test():
 
     return ([node], [x], [y])
 
+@onnx_test
```
kahmed10: Can you alphabetize the tests by moving them to the correct place in this file? Also, do the same for onnx_test.cpp. This way, when having both files side-by-side, it is easier to find the tests and debug the code if necessary.

scxiao: fixed

PR #366, opened 2019-09-18
Fix HIP-Clang GPU build issues

aaronenyeshi:
This is the same as #362 , just moved the fork branch to here.

Add missing device attributes for GPU functions. GPU functions must be annotated with device in HIP. Also, fix a few warning: nvcc does not allow 'device' to appear after '()' in lambdas.

Here are the test results, I will file bugs for the failures:

99% tests passed, 2 tests failed out of 182

Total Test time (real) = 26.47 sec


aaronenyeshi:
Rebased on develop

mvermeulen:
@aaronenyeshi do you have an understanding on the build failures? Is there a plan to address?

PR #368, opened 2019-09-24
Improve contiguous and concat performance

pfultz2:
This improves the performance by reading transposed input in order of the input rather then the order of the output.

The is also refactors concat to call contiguous so it can reuse the same code path.


src/targets/gpu/device/include/migraphx/gpu/device/nary.hpp
```diff
@@ -259,20 +290,25 @@ void nary_standard_impl(hipStream_t stream, F f, argument result, Arguments... a
...
+    MIGRAPHX_TRACE_NARY_FUNCTION
+    const auto shapes   = make_array(args.get_shape()...);
+    const bool standard = all_of(shapes, [](const shape& s) { return s.standard(); });
+    const bool packed   = all_of(shapes, [](const shape& s) { return s.packed(); });
+    const bool same_shapes =
+        all_of(shapes, [&](const shape& s) { return s == result.get_shape(); });
+    const bool same_input_shapes = all_of(shapes, [&](const shape& s) { return s == shapes[0]; });
+    if((result.get_shape().standard() and standard) or (packed and same_shapes))
         nary_standard_impl(stream, f, result, args...);
+    else if(packed and same_input_shapes)
```
kahmed10: do we have a test case for this?

pfultz2: Yes the `test_concat_transpose` will test this path.

kahmed10: I see that with concat, you're iterating through each input and calling contiguous() with a single argument. In this case, same_input_shapes will always be true since there's only one input shape. Do we have a test case where that's not true?

src/include/migraphx/permutation.hpp
```diff
@@ -0,0 +1,57 @@
+/*=============================================================================
```
kahmed10: is this from another source? If so, do we need to update the copyright date?

src/targets/gpu/device/include/migraphx/gpu/device/nary.hpp
pfultz2: For `test_concat_transpose` the `same_input_shapes` is true but `same_shapes` is not true. However, everything is packed so it will be called. If the inputs are non-packed and the output is then it will hit the final else.

src/include/migraphx/permutation.hpp
pfultz2: Oops this was auto-generated from my editor. I'll remove it.

src/targets/gpu/device/include/migraphx/gpu/device/nary.hpp
kahmed10: That makes sense, but I'm mainly asking if there are other operators to test that use nary() where the input shapes are not the same but the data is packed. If not, then I don't know if we need to compute another boolean just for this purpose.

src/include/migraphx/argument.hpp
```diff
@@ -43,6 +43,12 @@ struct argument : raw_data<argument>
 
     const shape& get_shape() const { return this->m_shape; }
 
+    argument reshape(const shape& s) const
+    {
+        argument self = *this;
```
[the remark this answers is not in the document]
pfultz2: The shape is not required to be standard.

PR #369, opened 2019-09-24
Yapf hook

kahmed10:
we can add a style file in the future if necessary, and the command line arg would be --style=file


Dockerfile
```diff
@@ -52,6 +52,9 @@ RUN pip install cget
 # Install rclone
 RUN pip install https://github.com/pfultz2/rclone/archive/master.tar.gz
 
+# Install yapf
+RUN pip install yapf
```
pfultz2: Shouldnt we pin the version for yapf?

kahmed10: fixed

PR #375, opened 2019-09-25
Ceil floor operators

scxiao:


test/onnx/onnx_test.cpp
```diff
@@ -8,6 +8,24 @@
 #include <migraphx/onnx.hpp>
 #include "test.hpp"
 
+// template <class Op>
```
kahmed10: unintended code?

scxiao: Yes, thanks for pointing out.

PR #378, opened 2019-09-27
bug_fix_for_gemm_copy

scxiao:


test/gpu/gemm.cpp
```diff
@@ -0,0 +1,75 @@
...
+    auto create_program = [] {
+        migraphx::program p;
+        migraphx::shape sa{migraphx::shape::float_type, {2, 16}};
+        migraphx::shape sb{migraphx::shape::float_type, {16, 8}};
+        migraphx::shape sc{migraphx::shape::float_type, {2, 8}};
+        auto pa = p.add_parameter("a", sa);
+        auto pb = p.add_parameter("b", sb);
+        auto pc = p.add_parameter("c", sc);
+        auto dr = p.add_instruction(migraphx::op::dot{}, pa, pb, pc);
+        p.add_instruction(migraphx::op::add{}, dr, dr);
```
pfultz2: Why isnt this added to the gpu_ops test instead?

scxiao: the issue appeared only when eval() is called for the second time or later....

src/targets/gpu/lowering.cpp
```diff
@@ -198,25 +198,17 @@ struct miopen_apply
...
-                    refs.back()   = copy_out;
-                    refs.push_back(copy_out);
-                }
+                beta = 0;
+                refs.push_back(output);
             }
             else
             {
-                refs.push_back(refs.back());
+                auto copy_out = prog->insert_instruction(ins, hip_copy{}, refs.back(), output);
```
pfultz2: This will always make a copy even if its only used once. Why dont you check for `refs.back()->outputs().size() > 1` anymore?

test/gpu/gemm.cpp
pfultz2: Lets run all of our gpu_ops_test twice.

PR #382, opened 2019-10-02
Accelerate calculating conflict table

scxiao:


src/schedule.cpp
```diff
@@ -303,38 +307,100 @@ struct stream_info
...
+            index_to_ins.push_back(merge.first);
+        }
+
+        par_for(concur_ins.size(), [&](auto ins_index) {
+            auto merge_first = index_to_ins[ins_index];
+            assert(concur_ins.count(ins1) > 0);
+            auto& merge_second = concur_ins.at(merge_first);
+
+            auto thr_id = std::this_thread::get_id();
+            if(thread_conflict_tables.count(thr_id) == 0)
```
pfultz2: The lock would need to include the conditional here.

pfultz2: Although, I think it would be better to extend `par_for` to add a sequential ids, and then we can just preallocate a vector.

src/schedule.cpp
```diff
@@ -304,6 +304,52 @@ struct stream_info
...
+        int size = static_cast<int>(merge_ins.size());
+        std::unordered_set<instruction_ref> checked_ins1_set;
+        for(int i = 0; i < size - 1; ++i)
+        {
+            std::unordered_set<instruction_ref> ins1_set;
+            for(auto ins1 : merge_ins[i])
+            {
+                if(checked_ins1_set.count(ins1) == 0)
+                    ins1_set.insert(ins1);
+            }
```
pfultz2: This can written with `copy_if`.

src/schedule.cpp
```diff
@@ -304,6 +304,52 @@ struct stream_info
...
+            checked_ins1_set.insert(ins1_set.begin(), ins1_set.end());
+
+            std::unordered_set<instruction_ref> ins2_set;
+            for(auto j = i + 1; j < size; ++j)
+            {
+                for(auto ins2 : merge_ins[j])
+                {
+                    if(checked_ins1_set.count(ins2) == 0)
+                        ins2_set.insert(ins2);
+                }
```
pfultz2: This can written with `copy_if`.

src/schedule.cpp
```diff
@@ -304,6 +304,52 @@ struct stream_info
...
+            }
+
+            for(auto ins1 : ins1_set)
+            {
+                auto p1 = std::distance(ins1, ins);
```
pfultz2: How do we know that `ins1` always precedes `ins`?

src/schedule.cpp
```diff
@@ -304,6 +304,52 @@ struct stream_info
...
+                for(auto ins2 : ins2_set)
+                {
+                    if(ins1 == ins2)
+                        continue;
+                    auto p2 = std::distance(ins2, ins);
```
pfultz2: How do we know that `ins2` always preceded `ins`?

src/schedule.cpp
```diff
@@ -304,6 +304,52 @@ struct stream_info
...
     }
 
+    void
+    compute_conflict_table(instruction_ref ins,
+                           const std::vector<std::vector<instruction_ref>>& merge_ins,
+                           std::unordered_map<instruction_ref, std::unordered_set<instruction_ref>>&
+                               thrd_conflict_table)
+    {
+        int size = static_cast<int>(merge_ins.size());
+        std::unordered_set<instruction_ref> checked_ins1_set;
```
pfultz2: Why is this called `checked_ins1_set`? There doesn't seem to be a `checked_ins2_set`?

src/schedule.cpp
scxiao: fixed

src/schedule.cpp
scxiao: changed accordly

src/schedule.cpp
scxiao: instructions in the concurrent table meets that requirement

src/schedule.cpp
scxiao: instructions in the concurrent table meets that requirement

src/include/migraphx/par_for.hpp
```diff
@@ -27,13 +27,26 @@ struct joinable_thread : std::thread
...
+
 template <class F>
 void par_for_impl(std::size_t n, std::size_t threadsize, F f)
 {
     if(threadsize <= 1)
     {
         for(std::size_t i = 0; i < n; i++)
-            f(i);
+            thread_invoke(i, 0, f);
+        ;
```
pfultz2: Is this extra semicolon necessary?

src/schedule.cpp
scxiao: fixed

PR #393, opened 2019-10-18
Pad calculation fix for nasnet

kahmed10:


pfultz2:
Can we add unit tests for the `calculate_padding` function?

kahmed10:
> Can we add unit tests for the `calculate_padding` function?

added

PR #396, opened 2019-10-24
Support initializer not in inputs

scxiao:


src/onnx/onnx.cpp
```diff
@@ -1426,17 +1426,14 @@ struct onnx_parser
         for(auto&& f : graph.initializer())
         {
             initializer_data[f.name()] = f;
```
pfultz2: Do we even need the `initializer_data` map now? We can just create the literals instead.

src/onnx/onnx.cpp
```diff
@@ -1426,17 +1426,14 @@ struct onnx_parser
...
             const std::string& name = input.name();
-            // Does the input have an initializer?
-            if(contains(initializer_data, name))
-            {
-                auto t             = initializer_data[name];
-                instructions[name] = prog.add_literal(parse_tensor(t));
-            }
-            else
+            // input not in initializer_data, so it is a real input
+            if(!contains(initializer_data, name))
```
pfultz2: This could be `!contains(instructions, name)` instead.

scxiao: changed accordingly

src/onnx/onnx.cpp
scxiao: removed it.

PR #401, opened 2019-11-05
Add Split op

kahmed10:
Should fix nasnet A large in tf. I didn't write an optimization pass for slices, but I have a line to check if the number of outputs for split is equal to 1, therefore returning a identity.


src/tf/tf.cpp
```diff
@@ -939,23 +1033,43 @@ struct tf_parser
...
+                    if(contains(input, ':'))
+                    {
+                        auto orig_iname = input.substr(0, input.find(':'));
+                        assert(name != orig_iname);
+                        this->parse_node(orig_iname);
+                        args.push_back(instructions.at(input));
+                    }
+                    else
+                    {
+                        iname = get_name(nodes.at(input));
```
pfultz2: Only the `iname` assignment should be in the `if`:

```cpp
                    if(contains(input, ':'))
                    {
                        iname = input.substr(0, input.find(':'));
                    }
                    else
                    {
                        iname = get_name(nodes.at(input));
                    }
                    assert(name != iname);
                    this->parse_node(iname);
                    args.push_back(instructions.at(iname));
```

src/tf/tf.cpp
```diff
@@ -939,23 +1033,43 @@ struct tf_parser
...
-                instructions[name] = prog.add_instruction(op::unknown{node.op()}, args);
+                result.push_back(prog.add_instruction(op::unknown{node.op()}, args));
             }
             else
             {
-                instructions[name] = ops[node.op()](get_attributes(node), args);
+                result = ops[node.op()](get_attributes(node), args);
+            }
+            // First output has no ":" delimiter
+            instructions[name] = result.front();
```
pfultz2: We should check that `result` is not empty, maybe with an `assert`.

src/tf/tf.cpp

src/tf/tf.cpp

src/tf/tf.cpp
```diff
@@ -1229,7 +1342,7 @@ struct tf_parser
             return literal{{shape_type}, data};
         return literal{{shape_type, dims}, data};
     }
-};
+}; // namespace MIGRAPHX_INLINE_NS
```
pfultz2: There shouldn't be a semicolon after a namespace closes.

src/tf/tf.cpp
```diff
@@ -939,23 +1033,42 @@ struct tf_parser
                     continue;
                 if(nodes.count(input) > 0)
                 {
-                    auto&& iname = get_name(nodes.at(input));
+                    std::string iname;
```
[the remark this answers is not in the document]
pfultz2: When there is multiple outputs, the node is actually the same even though node names are different(which is why we only process one of the nodes). Each different name represents the different output of the nodes.

src/tf/tf.cpp

PR #402, opened 2019-11-14
Improve concat gather

scxiao:


src/include/migraphx/op/concat.hpp
```diff
@@ -18,7 +18,7 @@ namespace op {
 
 struct concat
 {
-    std::size_t axis = 0;
+    int axis = 0;
```
pfultz2: Do we usually make these attributes `int64_t`?

src/tf/tf.cpp
```diff
@@ -357,7 +357,7 @@ struct tf_parser
     {
         // get index for axis within args
         size_t axis_idx = attributes.at("N").i();
-        size_t axis     = args[axis_idx]->eval().at<int64_t>();
+        int axis        = args[axis_idx]->eval().at<int64_t>();
```
pfultz2: Can you write this with `auto`?

src/tf/tf.cpp
```diff
@@ -664,8 +664,7 @@ struct tf_parser
             args.end(),
             std::back_inserter(unsqueezed_args),
             [&](instruction_ref arg) { return prog.add_instruction(op::unsqueeze{{axis}}, arg); });
-        return to_nhwc(
-            prog.add_instruction(op::concat{static_cast<size_t>(axis)}, unsqueezed_args));
+        return to_nhwc(prog.add_instruction(op::concat{static_cast<int>(axis)}, unsqueezed_args));
```
pfultz2: Is the `static_cast` necessary?

src/include/migraphx/op/concat.hpp
kahmed10: yes, normally they are int64 if the axis is negative.

src/tf/tf.cpp
scxiao: removed

src/include/migraphx/op/concat.hpp
scxiao: changed

src/tf/tf.cpp
scxiao: fixed

PR #404, opened 2019-11-26
Add support for builtin models in the driver

pfultz2:
An untrained model can be loaded from the driver with `--model` flag.

Also, there is a flag to output the onnx file as C++. It should be considered experimental until we have support for serialization.

Obviously test coverage will be lower for this PR.

### DIFF
--- a/src/driver/main.cpp
+++ b/src/driver/main.cpp
@@ -138,6 +138,7 @@ struct compiler
 struct read : command<read>
 {
     loader l;
+    bool cpp = false;
     bool graphviz = false;
     bool brief    = false;
     std::string output;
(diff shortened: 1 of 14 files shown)


src/driver/main.cpp
```diff
@@ -26,17 +27,21 @@ inline namespace MIGRAPHX_INLINE_NS {
...
-    bool optimize = false;
+    unsigned batch = 1;
+    bool is_nhwc   = true;
+    unsigned trim  = 0;
+    bool optimize  = false;
 
     void parse(argument_parser& ap)
     {
         ap(file, {}, ap.metavar("<input file>"));
+        ap(model, {"--model"}, ap.help("Load model"));
```
kahmed10: I would add the currently supported models in the help text: [resnet50|inceptionv3|alexnet]

kahmed10:
> An untrained model can be loaded from the driver with `--model` flag.
> 
> Also, there is a flag to output the onnx file as C++. It should be considered experimental until we have support for serialization.
> 
> Obviously test coverage will be lower for this PR.

Maybe we can create tests that read in the respective model's onnx file and compare that program length is the same. That way we could improve code coverage. In the future, we could also compare a program that may have different literals generated but the ops are the same.

pfultz2:
Right now there aren't any tests at all for the driver. We can try to add some in the future. For now, I just disable code coverage for the driver directory.

src/driver/main.cpp
pfultz2: Done.

PR #418, opened 2019-12-20
Handle optional outputs

scxiao:


src/onnx/onnx.cpp
```diff
@@ -1499,12 +1499,37 @@ struct onnx_parser
...
+                    {
+                        vec_ins_refs.push_back(x);
+                    }
+                }
+
+                for(auto ins : vec_ins_refs)
+                {
+                    if(ins->outputs().empty())
+                    {
+                        prog.remove_instruction(ins);
```
pfultz2: We should rely on DCE to remove the unused outputs.

src/onnx/onnx.cpp
```diff
@@ -1499,12 +1499,37 @@ struct onnx_parser
...
+                    }
+                }
```
pfultz2: Cant we just write this as:

```cpp
                assert(node.output().size() <= result.size());
                std::transform(node.output().begin(),
                               node.output().end(),
                               result.begin(),
                               std::inserter(instructions, instructions.end()),
                               [](auto&& x, auto&& y) { return std::make_pair(x, y); });
```
Since we put `node.output()` first it should stop when there isn't enough output parameters.

src/onnx/onnx.cpp
scxiao: If the ins to be removed is the last ins of the program, DCE will not work.

pfultz2: Ok we should probably add `identity` at the end of the program to return the final output(at least until we support multiple returns).

src/onnx/onnx.cpp
scxiao: Fixed accordingly

PR #426, opened 2020-01-07
Improve parsing of pooling operators related to onnxruntime

scxiao:


src/include/migraphx/op/pooling.hpp
```diff
@@ -48,6 +48,18 @@ struct pooling
         assert(lengths[0] <= (input.lens()[2] + 2 * padding[0]));
         assert(lengths[1] <= (input.lens()[3] + 2 * padding[1]));
 
+        // for padding mode same (either same_upper or same lower)
+        // output shape is computed as:
+        if(padding_mode == same)
```
pfultz2: `compute_shape` shouldn't use padding mode. Instead, the padding should be set so that the output shape will be correct. We do the same thing for convolution.

MIOpen doesn't take padding mode, so if we dont set the correct padding then it can't compute the pooling correctly.

src/onnx/onnx.cpp
```diff
@@ -406,27 +406,48 @@ struct onnx_parser
...
+                MIGRAPHX_THROW("PARSE_POOLING: padding should have 4 values");
             }
             if(padding[0] != padding[2] || padding[1] != padding[3])
             {
                 // insert zeros for pad op (args[0] has 4 dims)
                 padding = {0, 0, padding[0], padding[1], 0, 0, padding[2], padding[3]};
-                l0 = prog.add_instruction(op::pad{padding, std::numeric_limits<float>::lowest()},
-                                          l0);
+                // MaxPool
+                if(ends_with(name, "MaxPool"))
```
kahmed10: can you check for op.mode == "max" instead?

src/onnx/onnx.cpp
```diff
@@ -435,14 +456,54 @@ struct onnx_parser
...
+                        pads[7] = explicit_pads[1];
+                    }
+                    else if(s == "SAME_LOWER")
+                    {
+                        pads[2] = explicit_pads[0];
+                        pads[3] = explicit_pads[1];
+                    }
+
+                    // MaxPool
+                    if(ends_with(name, "MaxPool"))
```
kahmed10: same as previous comment

src/onnx/onnx.cpp
```diff
@@ -435,14 +456,54 @@ struct onnx_parser
         {
             copy(attributes["kernel_shape"].ints(), op.lengths.begin());
         }
+
         if(contains(attributes, "auto_pad"))
```
kahmed10: can we create a separate function for auto_pad? We can use it for conv and conv_transpose as well

src/include/migraphx/op/pooling.hpp
scxiao: fixed accordingly

src/onnx/onnx.cpp

PR #428, opened 2020-01-16
Instance norm op

kahmed10:


src/onnx/onnx.cpp
```diff
@@ -788,6 +789,41 @@ struct onnx_parser
...
+        auto l0              = prog.add_instruction(op::sqdiff{}, x, mean_bcast);
+        auto variance        = prog.add_instruction(op::reduce_mean{{2, 3}}, l0);
+        auto l1              = prog.add_instruction(op::sub{}, x, mean_bcast);
+        auto epsilon_literal = prog.add_literal(epsilon);
+        auto epsilon_bcast   = prog.add_instruction(op::multibroadcast{dims}, epsilon_literal);
+        auto variance_bcast  = prog.add_instruction(op::multibroadcast{dims}, variance);
+        auto l2              = prog.add_instruction(op::add{}, variance_bcast, epsilon_bcast);
+        auto l3              = prog.add_instruction(op::rsqrt{}, l2);
+        auto l4              = prog.add_instruction(op::mul{}, l1, l3);
+        auto scale_bcast     = prog.add_instruction(op::multibroadcast{dims}, scale);
```
scxiao: Can we multibroadcast shape [C] to [N, C, H, W]? 
maybe we need a unsqueeze of [C] to [C, 1, 1] first, then we can do a multibroadcast?

The unit test below has the C=1 happens to work here. If C is not 1, I believe the multibroadcast will not work.

src/onnx/onnx.cpp
```diff
@@ -788,6 +789,41 @@ struct onnx_parser
...
+        auto bias_bcast      = prog.add_instruction(op::multibroadcast{dims}, bias);
```
scxiao: Same question as the scale..

test/onnx/gen_onnx.py
```diff
@@ -902,6 +902,49 @@ def initializer_not_an_input():
     return ([node], [x], [y], [w])
 
 
+@onnx_test
+def instance_norm_test():
+    x = helper.make_tensor_value_info('0', TensorProto.FLOAT, [1, 1, 3, 3])
```
scxiao: I would suggest changing the C in the shape to a number other than 1 to cover the scenario I mentioned above.

src/onnx/onnx.cpp
kahmed10: I replaced with a regular broadcast since you are correct that multibroadcast won't work in this case. I modified the test to C != 1 to verify.

test/onnx/gen_onnx.py
kahmed10: fixed

PR #429, opened 2020-01-16
Conv transpose op

kahmed10:


src/onnx/onnx.cpp
```diff
@@ -277,6 +278,25 @@ struct onnx_parser
         });
     }
 
+    template <class T>
+    std::vector<int64_t> to_int64_vector(std::vector<T> input_vector)
```
pfultz2: This should be passed as const ref.

src/onnx/onnx.cpp
```diff
@@ -277,6 +278,25 @@ struct onnx_parser
...
+    {
+        std::vector<int64_t> output_vector(input_vector.begin(), input_vector.end());
+        return output_vector;
+    }
+
+    instruction_ref
+    add_bias(std::vector<instruction_ref> args, instruction_ref curr_ins, uint64_t axis)
```
pfultz2: The vector should be passed as const ref.

pfultz2: Also does this `add_bias` function work for convolution as well?

kahmed10: > Also does this `add_bias` function work for convolution as well?

yes it does. I replaced it in the parse_conv function

src/onnx/onnx.cpp

PR #430, opened 2020-01-24
Add C and C++ api

pfultz2:


tools/api.py
```diff
@@ -0,0 +1,837 @@
...
+    return with_handle
+
+
+def template_eval(template, **kwargs):
+    start = '<%'
+    end = '%>'
+    escaped = (re.escape(start), re.escape(end))
+    mark = re.compile('%s(.*?)%s' % escaped, re.DOTALL)
+    for key in kwargs:
+        exec ('%s = %s' % (key, kwargs[key]))
```
kahmed10: you can fix the format error by making sure "yapf" is pointing to the python3 version when doing formatting locally

PR #432, opened 2020-01-31
Add multibroadcast test to simplify_reshapes

kahmed10:


src/simplify_reshapes.cpp
```diff
@@ -21,7 +21,8 @@ const auto& reshaper_names()
         "reshape",
         "contiguous",
         "squeeze",
-        "unsqueeze"
+        "unsqueeze",
+        "mulibroadcast"
```
pfultz2: This is a list of operators that can be replaced by a `reshape` operator. The `multibroadcast` cannot, so instead this needs to be listed only under the `find_nop_reshapes` class.

kahmed10: You're right, apparently this change is already in our pass, so I am changing this PR to only be adding the test case.

PR #433, opened 2020-02-04
Change api to multiple prog outputs (only API change)

scxiao:


test/gpu/ops_test.cpp
```diff
@@ -154,17 +161,22 @@ void run_verify_program()
...
-        std::cout << std::endl;
+
+    std::size_t num = cpu_arg.size();
+    for(std::size_t i = 0; i < num; ++i)
+    {
+        bool passed = verify_args(migraphx::get_type_name<V>(), cpu_arg[i], gpu_arg[i]);
+        if(not passed)
+        {
+            V v;
+            auto p = v.create_program();
```
pfultz2: This should not be inside the loop. There should be a loop to set the `passed` variable and then this can be called.

scxiao: fixed

PR #436, opened 2020-02-07
Support multi program outputs

scxiao:


src/include/migraphx/builtin.hpp
```diff
@@ -63,6 +63,20 @@ struct param
     }
 };
 
+struct add_return
+{
+    std::vector<std::string> output_names;
```
pfultz2: Why do we store names?

scxiao: Two reasons: The main reason is this is required to support onnxruntime. We need to use input and output name to match our program inputs/outputs to that in the onnxruntime, so we can put correct buffer in our parameters. Another reason is by storing the output names, we can verify our program outputs match that in onnx files.  The output names can be empty here. If empty, we use output_1, 2,3... as the output names.

pfultz2: Why can't they match based on the index?

scxiao: Before calling prog.eval(), we need to set arguments retrieved by calling get_param_shapes(). The argument map includes both inputs and output and there is no guarantee about the order of the arguments (inputs are before outputs, and output parameter are in the same order as that in onnx file).  So how can find which parameter here is which program output?

scxiao: if we really want to use index to match, we have to allocate parameter buffer for each output parameter, then call prog.eval() to compute the result. Then call hipMemcpy to copy result from our allocated buffer to the tensor buffer in onnxruntime, but this introduce additional memory copy, which is not necessary. 

The approach I am using now is using the onnxruntime tensor buffer as the buffer in our argument, so after prog.eval(), the prog output is already in the onnxruntime tensor buffer.

src/include/migraphx/builtin.hpp
```diff
@@ -63,6 +63,13 @@ struct param
     }
 };
 
+struct add_return
```
pfultz2: Can we name the class `returns`?

src/include/migraphx/builtin.hpp
```diff
@@ -63,6 +63,13 @@ struct param
...
+{
+    std::string name() const { return "return"; }
```
pfultz2: The name should be `@return` as all the builtins start with `@`.

scxiao: instruction name cannot start with "@", see 

instruction_ref program::add_instruction(const operation& op, std::vector<instruction_ref> args)
{
    return insert_instruction(impl->instructions.end(), op, std::move(args));
}
instruction_ref program::insert_instruction(instruction_ref ins,
                                            const operation& op,
{
    assert(std::all_of(
               args.begin(), args.end(), [&](instruction_ref x) { return has_instruction(x); }) &&
           "Argument is not an exisiting instruction");
    assert(not starts_with(op.name(), "@"));

kahmed10: You can either change the symbol or change the assertion to make an exception for return.

src/include/migraphx/builtin.hpp
scxiao: removed the output name

src/include/migraphx/builtin.hpp
scxiao: Done

src/include/migraphx/builtin.hpp
scxiao: renamed the name to @return and add throw exception to compute function.

src/onnx/onnx.cpp
```diff
@@ -1759,18 +1759,28 @@ struct onnx_parser
...
-
-        if(instructions.count(oit->name()) > 0)
-        {
-            prog.add_instruction(op::identity{}, instructions[oit->name()]);
-        }
+        std::vector<std::string> all_output_names;
+        std::vector<std::string> prog_output_names;
+        std::transform(prog_output.begin(),
+                       prog_output.end(),
+                       std::back_inserter(all_output_names),
```
pfultz2: Why do we need to copy this to an extra vector? Why can't the `copy_if` use `prog_output` directly instead?

src/onnx/onnx.cpp
```diff
@@ -1816,9 +1826,9 @@ struct onnx_parser
             }
             else
             {
-                assert(node.output().size() <= result.size());
+                auto output_num = std::min<std::size_t>(node.output().size(), result.size());
```
pfultz2: Why was this changed?

scxiao: the rnn operator satisfies node.output().size() > result.size(), but the MaxPooling operator tis the case node.output().size() > result.size().

src/onnx/onnx.cpp
scxiao: According to our discussion, I will let it unchanged.

PR #446, opened 2020-02-19
Upgrade onnx proto file

scxiao:
the file onnx.proto is too old and causes crash when running onnxruntime tests, so update to a newer version and also update the protobuf version.


pfultz2:
Can you make this PR without the multiple return changes?

mvermeulen:
For curiosity, I built this branch and tried it against the list of test ONNX files.  No differences observed in which pass/fail or the MIGraphX dumps from read.

Good news it doesn't break anything, but unfortunately also doesn't fix any cases with MIGraphX standalone.

PR #447, opened 2020-02-20
Add horizontal fusion for pointwise operators

pfultz2:


test/simplify_algebra_test.cpp
```diff
@@ -381,4 +381,114 @@ TEST_CASE(simplify_add_conv_no_fusion_asymetrical_strides2)
...
+        auto concat1 = p2.add_instruction(migraphx::op::concat{0}, x, y);
+        auto concat2 = p2.add_instruction(migraphx::op::concat{0}, one, two);
+        auto sum     = p2.add_instruction(migraphx::op::add{}, concat1, concat2);
+        auto relu    = p2.add_instruction(migraphx::op::relu{}, sum);
+        p2.add_instruction(pass_op{}, relu);
+    }
+    EXPECT(p1 == p2);
+}
+
+TEST_CASE(simplify_concat_add_relu_broadcast1)
```
kahmed10: I would add a comment or rename the test name so that it explains that the sum of the literals gets broadcasted, and in the other test case the literals are first broadcasted then followed with a concat.

pfultz2: Done

PR #450, opened 2020-02-24
Imagescaler type mismatch

scxiao:
fix the type mismatch for the model fp16_tiny_yolov2


pfultz2:
There is a clang tidy issue for this PR:

```
/var/jenkins/workspace/GraphX_imagescaler_type_mismatch/src/onnx/onnx.cpp:1022:14: error: the variable 'input_lens' is copy-constructed from a const reference but is only used as const reference; consider making it a const reference [performance-unnecessary-copy-initialization,-warnings-as-errors]
        auto input_lens  = input_shape.lens();
             ^
        const  &

```

Issue #439, opened 2020-02-17
ONNX files mobilenet v1 and v2 fail at Opset 11


mvermeulen:
Mobilenet v1 and v2 fail for Opset 11 but pass for Opset 10.  The following error is given:
```
  what():  /home/mev/source/AMDMIGraphX/src/include/migraphx/check_shapes.hpp:58
: Wrong number of arguments: expected 1 but given 3
```
The gdb stack trace for this error and a dump of the ONNX file contents suggest perhaps the "clip" operator may be the issue.  Here are parts of the gdb stack trace
```
#5  0x00007ffff539e775 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x0000000000457e12 in migraphx::version_1::check_shapes::has<int> (this=this@entry=0x7fffffff21c0)
    at /home/mev/source/AMDMIGraphX-debug/src/include/migraphx/check_shapes.hpp:57
#7  0x000000000045c174 in migraphx::version_1::op::unary<migraphx::version_1::op::sigmoid>::compute_shape (this=<optimized out>, 
    inputs=std::vector of length 3, capacity 3 = {...}) at /home/mev/source/AMDMIGraphX-debug/src/include/migraphx/op/unary.hpp:15
#8  0x00007ffff6a16131 in migraphx::version_1::operation::private_detail_te_handle_type<migraphx::version_1::op::clip>::compute_shape (
    this=0x866370, input=...) at /home/mev/source/AMDMIGraphX-debug/src/include/migraphx/operation.hpp:554
#9  0x00007ffff58e22ce in migraphx::version_1::operation::compute_shape (input=std::vector of length 3, capacity 3 = {...}, 
    this=0x7fffffff2410) at /home/mev/source/AMDMIGraphX-debug/src/include/migraphx/operation.hpp:338
#10 migraphx::version_1::compute_shape (op=..., args=std::vector of length 3, capacity 3 = {...})
    at /home/mev/source/AMDMIGraphX-debug/src/instruction.cpp:234
#11 0x00007ffff58e8581 in migraphx::version_1::program::insert_instruction (this=this@entry=0x7fffffffd100, ins=..., op=..., 
    args=std::vector of length 3, capacity 3 = {...}) at /home/mev/source/AMDMIGraphX-debug/src/program.cpp:169
[...]

PR #455, opened 2020-02-26
Clip update for onnx

kahmed10:
#439


pfultz2:
Why dont we change the clip operator to take the min and max as input values?

kahmed10:
> Why dont we change the clip operator to take the min and max as input values?

Made changes to accept min and max as input args

src/onnx/onnx.cpp
```diff
@@ -325,16 +325,32 @@ struct onnx_parser
                                const attribute_map& attributes,
                                std::vector<instruction_ref> args)
     {
-        op::clip op;
+        auto input_lens = args[0]->get_shape().lens();
+        instruction_ref min_arg;
```
scxiao: If the operator has only one input and there is not "min" or "max", the min_arg and max_arg will be empty here. The logic would be something like 
        if(args.size() > 2)
        {
            max_arg = args[2];
        }
        else
        {
            float max_val = std::numeric_limits<float>::max();
            if(contains(attributes, "max"))
            {
                max_val = parse_value(attributes.at("max")).at<float>();
            }
            max_arg = prog.add_literal(max_val);
        }

        if(args.size() > 1)
        {
            min_arg = args[1];
        }
        else
        {
            float min_val = std::numeric_limits<float>::min();
            if(contains(attributes, "min"))
            {
                min_val = parse_value(attributes.at("min")).at<float>();
            }
            min_arg = prog.add_literal(min_val);
        }

If there is no min or max value provided, the default value is used. (Actually, for this case, the clip operator can be removed.)

src/onnx/onnx.cpp
```diff
@@ -325,16 +325,32 @@ struct onnx_parser
...
+            max_arg       = prog.add_literal(max_val);
         }
         if(contains(attributes, "min"))
         {
-            op.min_val = parse_value(attributes.at("min")).at<float>();
+            float min_val = parse_value(attributes.at("min")).at<float>();
+            min_arg       = prog.add_literal(min_val);
         }
-        return prog.add_instruction(op, std::move(args));
+        min_arg = prog.add_instruction(op::multibroadcast{input_lens}, min_arg);
```
scxiao: I am wondering why we need to multibroadcast? We can just use the index 0 to get the min/max value, right?

kahmed10: I added a multibroadcast to reuse nary() in the device function without adding more complexity.

src/onnx/onnx.cpp
kahmed10: I agree that I need to fix the logic. I am also debating on whether to assign default values to min/max or simply using a max operator when no max provided, and an identity if neither operators are provided.

scxiao: Yes, I am fine with not adding the operator if neither of min/max is used.

src/onnx/onnx.cpp
```diff
@@ -330,16 +330,48 @@ struct onnx_parser
...
+        if(min_used)
+            min_arg = prog.add_instruction(op::multibroadcast{input_lens}, min_arg);
+
+        if(max_used)
+            max_arg = prog.add_instruction(op::multibroadcast{input_lens}, max_arg);
+
+        if(min_used and max_used)
+            return prog.add_instruction(op::clip{}, args[0], min_arg, max_arg);
+        if(min_used)
+            return prog.add_instruction(op::max{}, args[0], min_arg);
```
pfultz2: Why is the max op being used with the min_arg?

kahmed10: If only the minimum value is clipped, you can simply call max(input, min_arg), which will prevent the input from going less than the minimum value.

pfultz2: I see its because we are clipping the min value.

PR #456, opened 2020-02-28
Add same padding mode for onnx

kahmed10:


src/onnx/onnx.cpp
```diff
@@ -470,7 +472,26 @@ struct onnx_parser
...
+
+                auto input_dims = l0->get_shape().lens();
+                std::vector<int64_t> pads(input_dims.size());
+                calculate_padding(0, pads, input_dims[2], op.stride[0], op.dilation[0], weight_h);
+                calculate_padding(1, pads, input_dims[3], op.stride[1], op.dilation[1], weight_w);
+
+                if(pads[0] != pads[2] || pads[1] != pads[3])
+                {
+                    std::vector<int64_t> padding = {0, 0, pads[0], pads[1], 0, 0, pads[2], pads[3]};
+                    l0 = prog.add_instruction(migraphx::op::pad{padding}, l0);
```
pfultz2: Could this be moved out of the conditionals so we can reuse the same logic for all cases of non-symmetrical padding?

src/onnx/onnx.cpp
```diff
@@ -470,7 +480,19 @@ struct onnx_parser
...
             {
-                op.padding_mode = op::padding_mode_t::same;
+                op.padding_mode                 = op::padding_mode_t::same;
+                std::vector<size_t> weight_dims = weights->get_shape().lens();
+                size_t weight_h                 = weight_dims[2];
+                size_t weight_w                 = weight_dims[3];
+
+                auto input_dims = l0->get_shape().lens();
+                std::vector<int64_t> padding(input_dims.size());
+                calculate_padding(
```
scxiao: we should also support same_lower by adding to flag to the function calculate_padding()

scxiao: Actually, you code change can be reused in both parse_conv and parse_pooling. I have  branch that calls your code change in this PR and used in both parse_conv and parse_pooling. The branch name is "correct_conv_shape_computation". This branch also support both the same_upper and same_lower mode. If you want, feel free to use the code changes in my branch.

src/onnx/onnx.cpp
```diff
@@ -470,7 +480,19 @@ struct onnx_parser
...
             if(s.find("SAME") != std::string::npos)
             {
-                op.padding_mode = op::padding_mode_t::same;
+                op.padding_mode                 = op::padding_mode_t::same;
+                std::vector<size_t> weight_dims = weights->get_shape().lens();
+                size_t weight_h                 = weight_dims[2];
+                size_t weight_w                 = weight_dims[3];
+
+                auto input_dims = l0->get_shape().lens();
+                std::vector<int64_t> padding(input_dims.size());
```
pfultz2: Can move this variable out of the if statement, so it can be reused for both padding modes?

src/onnx/onnx.cpp
```diff
@@ -470,7 +480,19 @@ struct onnx_parser
...
+                calculate_padding(
+                    0, padding, input_dims[2], op.stride[0], op.dilation[0], weight_h);
+                calculate_padding(
+                    1, padding, input_dims[3], op.stride[1], op.dilation[1], weight_w);
+
+                check_asym_padding(l0, padding, op);
```
pfultz2: Can this be moved out of the if statement so this function can be called once? (I dont think a separate function is necessary here either).

PR #459, opened 2020-03-05
CMakeLists.txt specifiy wrong debian package depends for hip_hcc

nelsonc-amd:
Change to hip-hcc.


pfultz2:
After merging master into develop, this is no longer needed.

PR #461, opened 2020-03-08
add improvements to pad calculation

kahmed10:
from @scxiao branch with additional improvements


src/onnx/onnx.cpp
```diff
@@ -1024,12 +1016,11 @@ struct onnx_parser
             auto&& bias_floats = attributes["bias"].floats();
             bias               = std::vector<float>(bias_floats.begin(), bias_floats.end());
         }
```
scxiao: The changes after this line is the old code, which should not be here. This might be due to you copied the file from my branch to here. I think you may need a merge instead of copying the file.

src/onnx/onnx.cpp
```diff
@@ -490,14 +458,21 @@ struct onnx_parser
...
                     1, padding, input_dims[3], op.stride[1], op.dilation[1], weight_w);
 
                 check_asym_padding(l0, padding, op);
             }
+
+            auto in_lens     = args[0]->get_shape().lens();
+            auto weight_lens = args[1]->get_shape().lens();
+            std::array<std::size_t, 2> k_lens;
+            k_lens[0] = weight_lens[2];
+            k_lens[1] = weight_lens[3];
```
pfultz2: Can you write this as `std::array<std::size_t, 2> k_lens = {weight_lens[2], weight_lens[2]}`?

kahmed10:
Retry DeepCode

PR #462, opened 2020-03-12
Add gelu optimization

kahmed10:


src/targets/gpu/fuse_ops.cpp
```diff
@@ -324,6 +333,45 @@ void move_standard_front(std::vector<instruction_ref>& args)
         std::swap(*it, args.front());
 }
 
+struct find_gelu
+{
+    auto matcher() const
+    {
+        return match::name("gpu::mul")(
+            match::arg(1)(match::name("gpu::add")(match::arg(0)(match::name("gpu::erf")(match::arg(
```
pfultz2: You need to check that the other operators are only used once. This excludes the entry node(ie `match::name('mul')`).

src/targets/gpu/fuse_ops.cpp
```diff
@@ -324,6 +333,45 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+                0)(match::name("gpu::div", "gpu::mul")(match::arg(0)(match::any().bind("x")))))))));
```
pfultz2: Why does it list both `div` and `mul`?

src/targets/gpu/fuse_ops.cpp
```diff
@@ -324,6 +333,45 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+
+        p.replace_instruction(ins, hip_gelu{}, x_ins, args.back());
+    }
+};
+
+struct find_add_gelu
+{
+    auto matcher() const
+    {
+        return match::name("gpu::gelu")(match::arg(0)(match::name("gpu::add").bind("add")));
```
pfultz2: Add needs to be checked for used_once.

src/targets/gpu/fuse_ops.cpp
kahmed10: fixed

src/targets/gpu/fuse_ops.cpp
kahmed10: fixed

src/targets/gpu/fuse_ops.cpp
kahmed10: The equation can be written either with a division or multiply by a reciprocal.

pfultz2: But if its a multiply shouldn't `x` bind to the input of the reciprocal? It seems like you should do:

```cpp
match::any_of(
    match::name("gpu::div")(match::arg(0)(match::any().bind("x"))),
    match::name("gpu::mul")(match::arg(0)(match::name("gpu::recip")(match::arg(0)(match::any().bind("x"))))))
```

kahmed10: We wouldn't have the input bound to the reciprocal in this case since it's already precomputed. It won't show up in the program after lowering.

pfultz2: But that is what you are using as an input to the operator. If its precomputed then why isn't the entire operation removed by constant propagation?

test/gpu/ops_test.cpp
```diff
@@ -722,6 +723,52 @@ struct test_triadd_broadcast : verify_program<test_triadd_broadcast>
...
+struct test_gelu : verify_program<test_gelu>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        std::vector<size_t> input_lens{1, 1, 5};
+        auto x            = p.add_parameter("x", {migraphx::shape::float_type, input_lens});
+        auto half         = p.add_literal(0.5f);
+        auto one          = p.add_literal(1.0f);
+        auto sqrt2        = p.add_literal(static_cast<float>(M_SQRT2));
```
pfultz2, line 735: Can you add a test changing the values of the literals to something else? Maybe change `half` to `2.0` and `sqrt2` to be the sqrt(3). It doesn't need to fuse necessarily, but we should make sure it is correct.

src/targets/gpu/fuse_ops.cpp
pfultz2: I see now. Its the second operator that is a recirpocal. Either way, you dont change the implementation if its a multiply or divide. That would make a difference. Since we transform the div to a mul, we can just match the mul instead of trying to match them both.

pfultz2: I guess you assume the second operand is either sqrt(2) or reciprocal of that, but you dont actually check any of the constants in the matcher.

test/gpu/ops_test.cpp
kahmed10: I will make my matcher more robust since the values for sqrt2 and half are not checked currently. Therefore, the test would fail.

PR #463, opened 2020-03-12
Python_scalar_argument_support

scxiao:
This PR is support scalar shape as an argument in python. I found this issue from the seq2seq translation model.


test/py/test_cpu.py
```diff
@@ -1,18 +1,49 @@
...
+
+    for key, value in p.get_parameter_shapes().items():
+        print("Parameter {} -> {}".format(key, value))
+        params[key] = migraphx.generate_argument(value)
+
+    # args = []
+    # args.append(np.random.randn(2, 3, 4, 5).astype(np.single))
+    # args.append(np.array(1).astype(np.single))
+    # params["0"] = migraphx.argument(args[0])
+    # params["1"] = migraphx.argument(args[1])
```
pfultz2: Why is this commented out?

Dockerfile
```diff
@@ -60,6 +60,9 @@ RUN pip install https://github.com/pfultz2/rclone/archive/master.tar.gz
 # Install yapf
 RUN pip3 install yapf==0.28.0
 
+# Install numpy
+RUN pip3 install numpy
```
pfultz2: I dont see numpy used anywhere.

scxiao: I tried to use numpy to create a scalar input for testing and added it here. But it still does not work. So changed to use others. Is it  the correct way to add numpy to be used in the above unit test?

pfultz2: So can this be removed? Or whats the plan?

scxiao: I removed this

test/py/test_cpu.py
scxiao: @pfultz2 When you have time, can you review this one? Thanks

test/py/test_cpu.py
```diff
@@ -1,18 +1,43 @@
...
+    print("Compiling ...")
+    p.compile(migraphx.get_target("cpu"))
+    print(p)
+    s2 = p.get_output_shapes()[-1]
+    assert s1 == s2
+    params = {}
+
+    for key, value in p.get_parameter_shapes().items():
+        print("Parameter {} -> {}".format(key, value))
+        params[key] = migraphx.generate_argument(value)
```
pfultz2, line 36: Does this actually create the scalar value?

pfultz2: Can we add assert to verify that?

scxiao: refined the test to cover the code change

Dockerfile
```diff
@@ -58,6 +58,9 @@ RUN pip3 install cget && pip3 install numpy
 # Install rclone
 RUN pip install https://github.com/pfultz2/rclone/archive/master.tar.gz
 
+# Install numpy
+pip install numpy
```
pfultz2, line 62: It seems this can be removed now?

scxiao: I am not sure for now. I am not sure wether memoryview() function in python2.7 can generate a scalar argument. The current code only generate the shape [1], which is not a scalar and the test failed in Debug build. So I need further investigation on it.

scxiao: If you look at the unit test, I still use the numpy to generate a scalar. We have the line pip3 install numpy for python version 3.x. But we do not have that for Python2.7, so I add this line here. Actually, I am not sure whether it works.

pfultz2: Can you use `create_buffer` to create the scalar instead? If the test is broken on python 2.7 then just skip it on python 2.

I would prefer not to require a dependency directly on numpy since it may not be installed(and wont get installed with rbuild).

PR #466, opened 2020-03-12
Improve parsing pad operator

scxiao:
This PR is based on the PR of support split operator, so should be committed after that PR.


src/onnx/onnx.cpp
```diff
@@ -1038,48 +1033,77 @@ struct onnx_parser
...
-    parse_pad(const std::string&, attribute_map attributes, std::vector<instruction_ref> args)
+    instruction_ref parse_pad(const std::string&, node_info info, std::vector<instruction_ref> args)
     {
         std::vector<int64_t> pads{};
-        float value = 0.0f;
-        if(contains(attributes, "pads"))
+        if(args.size() >= 2)
+        {
+            auto pad_ins = args.at(1);
+            if(!pad_ins->can_eval())
```
kahmed10: Similar to comment in #467

src/onnx/onnx.cpp
```diff
@@ -1038,48 +1033,77 @@ struct onnx_parser
...
         }
+
         // check if padding is actually being done (at least one value is nonzero)
         if(std::all_of(pads.begin(), pads.end(), [](const int& i) { return i == 0; }))
         {
             return prog.add_instruction(migraphx::op::identity{}, args.front());
         }
-        if(contains(attributes, "value"))
+
+        float value = 0.0f;
```
kahmed10: We should probably rewrite the pad operator to take an extra arg so that the value is passed in as an arg of type T. I am not sure if we should do it in this PR or in another one.

scxiao: The value can be used as an arg, but the pads cannot since it is related to the shape, and must be constant.

Considering change the value to an arg related to the PR for auto_pad processing, I would prefer do it later after this PR and the auto_pad PR are merged in. 

Create issue #471 to track that.

src/onnx/onnx.cpp

PR #467, opened 2020-03-14
Improve parsing slice

scxiao:


src/include/migraphx/op/slice.hpp
```diff
@@ -11,6 +11,8 @@
 #include <migraphx/config.hpp>
 #include <cmath>
 #include <utility>
+#include <iostream>
```
kahmed10: I don't think this include is needed anymore.

src/onnx/onnx.cpp
```diff
@@ -745,26 +745,72 @@ struct onnx_parser
...
         std::vector<size_t> dims = args[0]->get_shape().lens();
         size_t num_dims          = dims.size();
-        if(contains(attributes, "axes"))
+
+        // slice can have up to 5 inputs, we first check the 5th one
+        // to decide whether MIGRAPHX can handle this slice
+        if(args.size() == 5)
+        {
+            auto tind_ins = args.back();
+            if(!tind_ins->can_eval())
```
kahmed10: Can we make a function that checks can_eval and replace the conditionals on this line and 770 and 786?

src/include/migraphx/op/slice.hpp
scxiao: Fixed

src/onnx/onnx.cpp
scxiao: we have a function check_arg_empty() that can be used here.

PR #470, opened 2020-03-24
simplify div and sub by constants

kahmed10:


src/simplify_algebra.cpp
```diff
@@ -359,6 +402,8 @@ void simplify_algebra::apply(program& p) const
                             find_mul_conv{},
                             find_mul_add{},
                             find_concat_unary{},
+                            find_div_const{},
+                            find_sub_const{},
```
pfultz2: Dont put this inbetween the concat finders. Ideally, just put it before `find_concat_unary`.

test/onnx/gen_onnx.py
```diff
@@ -777,6 +777,25 @@ def deconv_stride_test():
...
+                                  vals=y.flatten().astype(np.float))
+
+    z = helper.make_tensor_value_info('2', TensorProto.FLOAT, [3])
+
+    node = onnx.helper.make_node('Div', inputs=[
+        '0',
+        '1',
+    ], outputs=['2'])
+
+    return ([node], [x], [z], [y_tensor])
```
pfultz2: Is this necessary? There is no onnx test for it. Also, I dont think we need an onnx test.

src/simplify_algebra.cpp
kahmed10: fixed

test/onnx/gen_onnx.py
kahmed10: fixed

src/include/migraphx/op/recip.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <array>
+#include <migraphx/op/unary.hpp>
+#include <migraphx/operation.hpp>
+#include <migraphx/check_shapes.hpp>
+#include <migraphx/stringutils.hpp>
+#include <migraphx/streamutils.hpp>
+#include <migraphx/literal.hpp>
+#include <migraphx/shape_for_each.hpp>
+#include <migraphx/config.hpp>
+#include <cmath>
```
scxiao, line 13: Some of the includes file may not be necessary here. I believe you can remove most of them.

src/simplify_algebra.cpp
```diff
@@ -7,8 +7,11 @@
 #include <migraphx/op/convolution.hpp>
 #include <migraphx/op/as_shape.hpp>
 #include <migraphx/op/broadcast.hpp>
+#include <migraphx/op/neg.hpp>
+#include <migraphx/op/recip.hpp>
 #include <migraphx/matcher.hpp>
 #include <migraphx/literal.hpp>
+#include <iostream>
```
scxiao, line 14: iostream may not be needed here.

kahmed10: fixed

src/include/migraphx/op/recip.hpp
kahmed10: fixed

PR #472, opened 2020-03-24
Horizontal fusions of gemms and convolutions

pfultz2:


pfultz2:
Here are the times for this PR vs develop:

* inceptionv4: 12.9815ms vs 13.9111ms
* bert: 24.9702ms vs 28.804ms

The bert module can be improved more by removing the extra contiguoug operators, which I believe @scxiao is working on.

The inception model can also be improved by removing the extra multiply operators. I plan to address that in a future PR.

src/simplify_algebra.cpp
```diff
@@ -346,20 +555,110 @@ struct find_add_convs
...
+        auto outputs = ins->outputs();
+        group_by(outputs.begin(), outputs.end(), each, pred);
+    }
+};
+
 void simplify_algebra::apply(program& p) const
 {
     // Run simplifications multiple times
-    for(int i = 0; i < 4; i++)
+    for(int i = 0; i < 8; i++)
```
kahmed10, line 648: should we set the number of iterations to an environment variable? I am assuming it was changed from 4 to 8 because with 4 iterations the program wasn't fully simplified.

src/simplify_algebra.cpp
```diff
@@ -243,6 +246,212 @@ struct find_concat_binary
...
+    }
+    return result;
+}
+
+struct find_splits
+{
+    auto matcher() const
+    {
+        return match::any(match::any_of[match::outputs()](match::name("slice")(
+            match::any_of[match::outputs()](match::name("add", "mul", "relu")))));
```
kahmed10, line 288: We should probably match all activations in the future, not just relu. Possibly the same thing for binary ops.

src/simplify_algebra.cpp
```diff
@@ -346,20 +555,110 @@ struct find_add_convs
...
+    auto dots  = std::count_if(ins->outputs().begin(), ins->outputs().end(), pred("dot"));
+    auto convs = std::count_if(ins->outputs().begin(), ins->outputs().end(), pred("convolution"));
+    return !(dots < 2 and convs < 2);
+}
+
+struct find_conv_dot_horiz_fusion
+{
+    auto matcher() const { return horiz_conv_dot(); }
+
+    static bool reaches(instruction_ref ins, const std::vector<instruction_ref>& args)
```
kahmed10, line 575: where is this being used?

src/simplify_algebra.cpp
pfultz2: I dont know if we need an env variable, I am not sure why we would need to change it. Ideally, I would like to change this to stop loops when there is no more matches.

src/simplify_algebra.cpp
pfultz2: Yes, I plan to have this include all unary elementwise operators in the future.

pfultz2:
@kahmed10 @scxiao can you review this?

kahmed10:
I am wondering if it'd be easier to keep track of the name of the input in the slice operator and the slice order whenever we parse split. In tensorflow, multiple outputs usually look something like "node:0," "node:1," "node:2..." This way, you can tell both the order of the output as well as the parent node.

src/shape.cpp
```diff
@@ -195,6 +195,14 @@ bool shape::scalar() const
 
 bool shape::standard() const { return impl->m_standard; }
 
+shape shape::normalize_standard() const
+{
+    if(this->standard())
```
scxiao, line 200: This function is going to implicitly convert a scalar type to the shape [1] with stride [1]

pfultz2: Yes, I think it should be expected. This is used to pass the shapes to miopen tensor descriptors but it doesn't support zero strides.

pfultz2:
@scxiao anymore feedback?

PR #474, opened 2020-03-27
Parse onnx using explicit input shape

scxiao:
migraphx changes to support dynamic input shape


src/include/migraphx/onnx.hpp
```diff
@@ -11,6 +11,7 @@ inline namespace MIGRAPHX_INLINE_NS {
 struct onnx_options
 {
     unsigned int batch_size = 1;
+    std::unordered_map<std::string, std::vector<std::size_t>> map_input_dims;
```
pfultz2: Why are we mapping to the inputs and not filling in the dim_params?

src/onnx/onnx.cpp
```diff
@@ -1914,8 +1915,14 @@ struct onnx_parser
             // input not in initializer_data, so it is a real input
             if(!contains(instructions, name))
             {
+                std::vector<std::size_t> dims;
+                if(map_input_dims.count(name) > 0)
+                {
+                    dims = map_input_dims.at(name);
+                }
+
                 // TODO: Get shape of input parameter
```
kahmed10, line 1924: do we still need this TODO?

src/include/migraphx/onnx.hpp
kahmed10: It seems that exported onnx files have a dim_param associated with the dim that is dynamic, but the name seems to be randomly generated (ex. "unk__271"). If there is a case where there is no dim_param in the onnx file but the shape is dynamic, we still need a way to handle that.

src/onnx/onnx.cpp
```diff
@@ -2105,7 +2112,9 @@ struct onnx_parser
         return literal{{shape_type, dims}, data.begin(), data.end()};
     }
 
-    static shape parse_type(const onnx::TypeProto& t, const unsigned int batch_size)
+    static shape parse_type(const onnx::TypeProto& t,
+                            const std::vector<std::size_t>& input_dims,
+                            std::size_t default_dim_value)
```
pfultz2, line 2117: Maybe `parse_type` should not be static and then we wont need to pass in these parameters.

src/driver/main.cpp
```diff
@@ -62,7 +62,7 @@ struct loader
             }
             std::cout << "Reading: " << file << std::endl;
             if(file_type == "onnx")
-                p = parse_onnx(file, onnx_options{batch});
+                p = parse_onnx(file, onnx_options{});
```
pfultz2, line 65: The batch should still be passed in from the driver.

src/api/migraphx.py
```diff
@@ -170,6 +170,21 @@ def program(h):
              const=True)
 
 
+@auto_handle
+def onnx_options(h):
+    h.constructor('create')
+    h.method(
+        'add_parameter_shape',
+        api.params(name='const char*', dim_num='size_t', dims='size_t*'),
```
pfultz2, line 178: This should just be `api.params(name='const char*', dims='std::vector<size_t>')`(although `migraphx::add_parameter_shape` will need to be updated to take a vector). The python will automatically create an additional parameter for the size of dims.

src/api/migraphx.py
```diff
@@ -170,6 +170,21 @@ def program(h):
              const=True)
 
 
+@auto_handle
+def onnx_options(h):
+    h.constructor('create')
+    h.method(
+        'add_parameter_shape',
```
pfultz2, line 177: I think this should be named as `set_input_parameter_shape` or `set_parameter_shape` instead to make it clearer.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -482,38 +482,72 @@ struct program : MIGRAPHX_HANDLE_BASE(program)
...
 
-inline program parse_onnx(const char* filename, migraphx_onnx_options options)
+struct onnx_options : MIGRAPHX_HANDLE_BASE(onnx_options)
 {
-    return program(make<migraphx_program>(&migraphx_parse_onnx, filename, &options), own{});
+    onnx_options() { this->make_handle(&migraphx_onnx_options_create); }
+
+    onnx_options(migraphx_onnx_options* p, own) { this->set_handle(p, own{}); }
+
+    onnx_options(migraphx_onnx_options* p, borrow) { this->set_handle(p, borrow{}); }
```
pfultz2, line 491: I dont think we need to borrow references for onnx options so I dont think this constructor is needed.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -482,38 +482,72 @@ struct program : MIGRAPHX_HANDLE_BASE(program)
     friend bool operator!=(const program& px, const program& py) { return !(px == py); }
 };
 
-inline program parse_onnx(const char* filename, migraphx_onnx_options options)
+struct onnx_options : MIGRAPHX_HANDLE_BASE(onnx_options)
 {
-    return program(make<migraphx_program>(&migraphx_parse_onnx, filename, &options), own{});
+    onnx_options() { this->make_handle(&migraphx_onnx_options_create); }
+
+    onnx_options(migraphx_onnx_options* p, own) { this->set_handle(p, own{}); }
```
pfultz2, line 489: The own parameter shouldn't be needed it should own by default.

pfultz2: Nevermind, we should keep for consistency.

src/onnx/onnx.cpp
scxiao: Fixed

src/onnx/onnx.cpp
scxiao: Fixed.

src/api/migraphx.py
scxiao: changed the name

src/api/include/migraphx/migraphx.hpp
scxiao: removed

src/driver/main.cpp
```diff
@@ -62,7 +62,7 @@ struct loader
             }
             std::cout << "Reading: " << file << std::endl;
             if(file_type == "onnx")
-                p = parse_onnx(file, onnx_options{batch});
+                p = parse_onnx(file, onnx_options{batch, {}});
```
pfultz2, line 65: The second parameter should not be necessary.

src/include/migraphx/onnx.hpp
pfultz2: This should be initialized with `std::unordered_map<std::string, std::vector<std::size_t>> map_input_dims{};` or `std::unordered_map<std::string, std::vector<std::size_t>> map_input_dims = {};`

PR #479, opened 2020-04-01
Parse graph node topologically

scxiao:
In this way, the parsed graph is more consistent with the nodes in the onnx file.


src/onnx/onnx.cpp
```diff
@@ -1861,9 +1860,40 @@ struct onnx_parser
...
+            std::vector<instruction_ref> args;
+            for(auto&& input : node.input())
+            {
+                if(input.empty())
+                {
+                    this->parse_undefined(input);
+                }
+                if(instructions.count(input) == 0)
+                {
+                    MIGRAPHX_THROW("PARSE_GRAPH: input " + input + " is unavailable");
```
pfultz2, line 1875: Is ONNX specified for the nodes to be in topological order? I could see how an exported may not make such a guarantee, which would mean we would throw an error on a valid onnx file.

pfultz2: I do see in the spec:

> Each computation dataflow graph is structured as a topologically sorted list of nodes that form a graph

So a topological order is guaranteed by the spec. So it is acceptable to throw an error here. We should probably say in the error message that it is invalid onnx file and that the nodes are not in order.

pfultz2: @scxiao Are you able to update the message?

scxiao: Thanks for the investigation. Yes, I updated the message to say onnx file is invalid with onordered nodes.

PR #480, opened 2020-04-02
Usability updates to onnx parsing

pfultz2:
This will now throw an error on missing operator and there is a flag to enable the older behaviour.

There is also a flag to print the program on parsing error.


kahmed10:
Do we plan on opening a separate PR for the tf parsing changes?

pfultz2:
> Do we plan on opening a separate PR for the tf parsing changes?

We can.

src/py/migraphx_py.cpp
```diff
@@ -181,12 +181,20 @@ PYBIND11_MODULE(migraphx, m)
...
+                                          {.batch_size             = batch_size,
+                                           .skip_unknown_operators = skip_unknown_operators,
+                                           .print_program_on_error = print_program_on_error});
           },
           "Parse onnx file",
           py::arg("filename"),
-          py::arg("batch_size") = 1);
+          py::arg("batch_size")             = 1,
+          py::arg("skip_unknown_operators") = false,
+          py::arg("print_program_on_error") = false);
```
kahmed10, line 197: Do we want the default behavior in the python API to also print_program_on_error like the driver?

pfultz2:
The failure is from code coverage, but I dont think we can recover. Its mainly missing the try/catch coverage.

src/py/migraphx_py.cpp
pfultz2: I dont think so, as this can be used by applications.

src/onnx/onnx.cpp
```diff
@@ -2112,21 +2116,25 @@ template <class... Ts>
...
-#ifndef NDEBUG
-    // Log the program when it can't be parsed
-    try
+    parser.batch_size             = options.batch_size;
+    parser.skip_unknown_operators = options.skip_unknown_operators;
+    if(options.print_program_on_error)
     {
-        parser.parse_from(std::forward<Ts>(xs)...);
+        // Log the program when it can't be parsed
+        try
```
kahmed10, line 2124: One way to get better coverage could be to call parse_onnx() in onnx_test.cpp with print_program_on_error=true.

test/onnx/onnx_test.cpp
```diff
@@ -1619,6 +1621,11 @@ TEST_CASE(unknown_test)
     EXPECT(p == prog);
 }
 
+TEST_CASE(unknown_test_throw)
+{
+    EXPECT(test::throws([&] { migraphx::parse_onnx("unknown_test.onnx"); }));
```
kahmed10, line 1626: you can repeat this call but add print_program_on_error=true to the options.

PR #488, opened 2020-04-07
Add recip op for gpu

kahmed10:


src/targets/gpu/device/recip.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include <migraphx/gpu/device/types.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+void recip(hipStream_t stream, const argument& result, const argument& arg)
+{
+    nary(stream, result, arg)([](auto x) __device__ { return 1 / (to_hip_type(x)); });
```
scxiao, line 12: Do we need to call the to_hip_type here?

kahmed10: fixed

PR #490, opened 2020-04-07
Segment fault c api

scxiao:


src/api/include/migraphx/migraphx.hpp
```diff
@@ -344,7 +344,10 @@ struct program_parameter_shapes : MIGRAPHX_HANDLE_BASE(program_parameter_shapes)
     std::vector<const char*> names() const
     {
         std::vector<const char*> result(this->size());
-        call(&migraphx_program_parameter_shapes_names, result.data(), this->get_handle_ptr());
+        if(result.size() > 0)
+        {
+            call(&migraphx_program_parameter_shapes_names, result.data(), this->get_handle_ptr());
```
pfultz2, line 349: Why is there a seg fault? Is there a way to fix this in the [`get_names`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/tools/api/api.cpp#L102) function instead?

test/api/test_cpu.cpp
```diff
@@ -22,4 +22,24 @@ TEST_CASE(load_and_run)
...
+{
+    auto p             = migraphx::parse_onnx("constant_fill_test.onnx");
+    auto shapes_before = p.get_output_shapes();
+    p.compile(migraphx::target("cpu"));
+    auto shapes_after = p.get_output_shapes();
+    CHECK(shapes_before.size() == 1);
+    CHECK(shapes_before.size() == shapes_after.size());
+    CHECK(bool{shapes_before.front() == shapes_after.front()});
+    migraphx::program_parameters pp;
+    auto param_shapes = p.get_parameter_shapes();
```
kahmed10, line 35: if there are no parameters to the program, why are we doing this part of the test?

src/api/include/migraphx/migraphx.hpp
scxiao: I think if the size of result is 0, result.data() is nullptr, and then it failed.

test/api/test_cpu.cpp
scxiao: For an onnx file, we may not know whether there are parameters. that is why we need that check.

kahmed10: we know this file has no parameters, so I would just add a check for program_parameters.empty() instead.

scxiao: I think if we add a check for param_shapes.empty(), and only if it is false we do the loop, then the following for loop becomes useless.

src/api/include/migraphx/migraphx.hpp
scxiao: The check if (out == nullptr) caused the problem. It has nothing to do with the get_name function. Is there a way to remove the check?

test/api/test_cpu.cpp
kahmed10: that's fine by me

src/api/include/migraphx/migraphx.hpp
pfultz2: Let me look into this.

pfultz2: I see, its not a seg fault, there is an exception thrown because the c api returns an error. I think changing here in the C++ api is acceptable.

pfultz2:
There is a clang tidy warning:

```
/var/jenkins/workspace/_AMDMIGraphX_segment_fault_c_api/src/api/include/migraphx/migraphx.hpp:347:12: error: the 'empty' method should be used to check for emptiness instead of 'size' [readability-container-size-empty,-warnings-as-errors]

        if(result.size() > 0)
           ^~~~~~~~~~~~~~~~~
           !result.empty()

```

PR #495, opened 2020-04-09
Fix bug in unsqueeze shape and improve test coverage

### LINKED ISSUE #494: Unsqueeze on scalar input causes incorrect output shape
This issue is related to BERT squad. There are three unsqueeze operators followed by a concat. However, two out of the three unsqueezes are scalars with shape (1) and stride (0). Our currrent implementation of unsqueeze sets the stride to (1) for scalars. The expected shape for the concat should be (1,1), which we are not seeing because of the current unsqueeze implementation.

pfultz2:
This should fix #494.


pfultz2:
Its trying to concat these args:

```
@86 = unsqueeze[axes={0}](@3) -> int64_type, {1, 1}, {0, 0}
@88 = unsqueeze[axes={0}](@87) -> int64_type, {1, 1}, {0, 0}
@0 = @literal{64} -> int64_type, {1}, {1}
```

Perhaps we are parsing the constant wrong.

src/include/migraphx/op/unsqueeze.hpp
```diff
@@ -58,6 +55,11 @@ struct unsqueeze
                 new_lens[i] = old_lens[p++];
             }
         }
+        if(input_shape.scalar())
```
kahmed10, line 58: As discussed earlier, we should revert this and modify the tests accordingly.

pfultz2:
Test have been updated.

PR #497, opened 2020-04-09
Fix bug with transpose and pad

pfultz2:


scxiao:
Looks like with this fix and the fix for eliminate_concat, the following models that have different results on CPU and GPU can have same results now:
opset7,8,9,10,11:
tf_inception_v1
tf_inception_v2
tf_mobilenet_v1_1.0_224
tf_mobilenet_v2_1.0_224
tf_mobilenet_v2_1.4_224
opset9/LSTM_Sequence_unpack

There are still 4 fp16 models that are incorrect

I will test this branch in onnxruntime, and let you guys know the output.

Issue #496, opened 2020-04-09
Slice: no axes set when arg is not present


kahmed10:
related to #485. In opset 10 , the arg for axes is optional, but if it is not present, we are currently leaving op.axes empty.

PR #498, opened 2020-04-10
Slice axes fix

kahmed10:
fix for #496


src/onnx/onnx.cpp
```diff
@@ -814,6 +814,13 @@ struct onnx_parser
             s.visit([&](auto v) { copy(v, std::back_inserter(op.starts)); });
         }
 
+        if(op.axes.empty())
+        {
+            std::vector<int64_t> axes(args[0]->get_shape().lens().size());
+            std::iota(axes.begin(), axes.end(), int64_t{0});
+            op.axes = axes;
+        }
+
```
scxiao, line 823: Do we need to make the change in the tf.cpp file to parse slice?
How about making this change in the function tune_attributes() in the file slice.hpp? Then we can change just one place.

pfultz2: We should do this in the onnx parsing because the optimization passes read this axes array so we should have it as "normal" as possible.

PR #502, opened 2020-04-11
Add OneHot op

kahmed10:
Basic implementation for bert squad. More work can be done to extend the capabilities according to the onnx spec.


src/onnx/onnx.cpp
```diff
@@ -1807,6 +1808,39 @@ struct onnx_parser
         return ret_ins;
     }
 
+    instruction_ref
+    parse_onehot(const std::string&, node_info info, std::vector<instruction_ref> args)
+    {
+        migraphx::argument depth_arg = args[1]->eval();
+        check_arg_empty(depth_arg, "ONEHOT: depth - dynamic shape not supported");
+        size_t depth = static_cast<size_t>(depth_arg.at<int32_t>());
```
pfultz2, line 1816: Why the `static_cast`? Why not read it as `size_t` from the argument?

src/onnx/onnx.cpp
```diff
@@ -1807,6 +1808,39 @@ struct onnx_parser
...
+
+        if(contains(info.attributes, "axis"))
+            axis = info.attributes.at("axis").i();
+        if(axis == -1)
+        {
+            shape s{shape::float_type, {depth, depth}};
+            auto l0 = prog.add_literal({s, depth_input});
+            return prog.add_instruction(op::gather{0}, {l0, args[0]});
+        }
+        MIGRAPHX_THROW("MIGraphX does not support axis != -1");
```
scxiao, line 1841: If you want, can you change the error message to "ONEHOT:  MIGraphX does not support axis != -1"?

PR #509, opened 2020-04-20
Fix build failure for HIP-Clang

yxsamliu:


CMakeLists.txt
```diff
@@ -128,6 +128,7 @@ rocm_enable_clang_tidy(
         -UNDEBUG
         -DMIGRAPHX_USE_CLANG_TIDY
         "-Dmain\\\\(...\\\\)=main\\\\(__VA_ARGS__\\\\) // NOLINT"
+        -Xclang -fallow-half-arguments-and-returns
```
pfultz2, line 131: These flags should be added to the migraphx_device target not to clang-tidy.

src/targets/gpu/device/include/migraphx/gpu/device/multi_index.hpp
```diff
@@ -31,7 +31,7 @@ struct multi_index
 };
 
 template <class ForStride>
-auto deduce_for_stride(ForStride fs) -> decltype(fs(id{}));
+__device__ __host__ auto deduce_for_stride(ForStride fs) -> decltype(fs(id{}));
```
pfultz2, line 34: It seems strange to add `__device__ __host__` to this function when it is unevaluated. Maybe it would be better to use `constexpr` for this.

CMakeLists.txt
yxsamliu: fixed

PR #510, opened 2020-04-20
Onehot operator

scxiao:


pfultz2:
Can't we implement this using mod+reshape operators instead of creating a new gpu operator? The onnx test using numpy here might be a good reference point:

https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/onehot.py#L13

scxiao:
> Can't we implement this using mod+reshape operators instead of creating a new gpu operator? The onnx test using numpy here might be a good reference point:
> 
> https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/onehot.py#L13

I looked into the above implementation, converting to migraphx operators, it is like the following scenario: we have indices, axis, value, and depth as the input, we need the _EQUAL_ operator
cast_indices = cast[to_int32 type](indices);
unsqueezed_indices = unsqueeze({axis}, cast_indices);
mod_indices = mod(unsqueezed_indices, depth);
std::vector<int32> vec(depth);
std::iota(vec.begin(), vec.end(), 0);
l = literal({int32_type, depth}, vec);
unsqueezed_l = unsqueeze({0, 1, ..., axis-1, rank+1}, l);
ones = equal(mod_indices, unsqueezed_l) (implicit multiple broadcast is required)
l_off_value = slice({0}, {0}, {1}, value)
l_on_value = slice({0}, {1}, {2}, value)
output = add(ones * (sub(l_on_value, l_off_value), l_off_value) (implicit multibroadcast)

PR #514, opened 2020-04-30
Fix handling of lowest values in pad operator 

pfultz2:


kahmed10:
To improve code coverage, I suggest adding cpu tests where padding is done with lowest() and max() values.

test/gpu/ops_test.cpp
```diff
@@ -2267,6 +2268,23 @@ struct test_pad_int8 : verify_program<test_pad_int8>
     }
 };
 
+struct test_pad_lowest : verify_program<test_pad_lowest>
```
kahmed10, line 2271: can you add a test for max as well?

PR #515, opened 2020-04-30
Add reflect pad mode

kahmed10:
Uses slices and concats to enable reflect pad mode support. Related to #449


test/onnx/onnx_test.cpp
```diff
@@ -1165,6 +1165,40 @@ TEST_CASE(pad_3arg_test)
     EXPECT(p == prog);
 }
 
+TEST_CASE(pad_reflect_test)
+{
+    migraphx::program p;
+    auto l0 = p.add_parameter("0", migraphx::shape{migraphx::shape::float_type, {2, 2}});
+    p.add_literal({migraphx::shape{migraphx::shape::int32_type, {4}}, {0, 2, 0, 1}});
```
pfultz2, line 1172: Why is this added? Its never used.

src/onnx/onnx.cpp
```diff
@@ -448,6 +448,56 @@ struct onnx_parser
...
+        std::vector<int> ldims(num_dims);
+        std::vector<int> rdims(num_dims);
+
+        std::vector<int64_t> axes(num_dims);
+        std::iota(axes.begin(), axes.end(), int64_t{0});
+
+        for(int i = 0; i < num_dims; i++)
+        {
+            ldims.at(i) = pads.at(i);
+            rdims.at(i) = pads.at(i + num_dims);
```
pfultz2, line 463: Why not initialize the vector with the values from pads?

```cpp
std::vector<int> ldims(pads.begin(), pads.begin()+num_dims);
std::vector<int> rdims(pads.begin()+num_dims, pads.end());
assert(ldims.size() == rdims.size());
```

src/onnx/onnx.cpp
```diff
@@ -448,6 +448,56 @@ struct onnx_parser
...
+            std::vector<size_t> dims = input->get_shape().lens();
+            std::vector<int64_t> starts(axes.size(), 0);
+            std::vector<int64_t> ends(dims.begin(), dims.end());
+            std::vector<instruction_ref> slices;
+
+            for(int j = 0; j < lcount; j++)
+            {
+                auto start_idx = (j + 1) % dims.at(i);
+                starts.at(i)   = start_idx;
+                ends.at(i)     = start_idx + 1;
```
pfultz2, line 485: `starts.at(i)`and `ends.at(i)` can be assigned to a pointer or iterator outside of the loop so it doesn't need to do a lookup each time.

test/onnx/onnx_test.cpp
kahmed10: This is because in the latest opset of onnx, the operator takes pads as an input. It is an artifact of parsing in a constant value (see other pad tests).

pfultz2: I see.

src/onnx/onnx.cpp
```diff
@@ -1151,6 +1200,8 @@ struct onnx_parser
         if(contains(info.attributes, "mode"))
```
scxiao, line 1200: The mode attribute should be moved to before if (args.size() == 3) since the input value is used only for the constant mode.

PR #516, opened 2020-04-30
Add simplify rsqrt pass

kahmed10:


src/simplify_algebra.cpp
```diff
@@ -391,6 +392,23 @@ struct find_sub_const
     }
 };
 
+struct find_rsqrt
+{
+    auto matcher() const
+    {
+        return match::name("recip")(
+            match::args(match::name("sqrt")(match::args(match::any().bind("x")))));
```
scxiao, line 400: Should we consider output of sqrt used only once?

kahmed10: fixed

src/simplify_algebra.cpp
```diff
@@ -396,8 +396,8 @@ struct find_rsqrt
 {
     auto matcher() const
     {
-        return match::name("recip")(
-            match::args(match::name("sqrt")(match::args(match::any().bind("x")))));
+        return match::name("recip")(match::args(
+            match::name("sqrt")(match::used_once(), match::args(match::any().bind("x")))));
```
scxiao, line 400: Can you add a unit test to verify this?

kahmed10: added

PR #517, opened 2020-05-02
Rnn variable seq lengths

scxiao:


pfultz2:
Is it possible to use [`GatherElements`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#GatherElements) to shift the sequences instead of implementing custom operators?

kahmed10:
Can you comment somewhere in the code a summary of how the shifting algorithm works?

scxiao:
> Can you comment somewhere in the code a summary of how the shifting algorithm works?

I will add some comments. Thanks

src/onnx/onnx.cpp
```diff
@@ -1697,10 +1713,16 @@ struct onnx_parser
             op::lstm{hidden_size, vec_actv_funcs, dirct, clip, input_forget}, std::move(args));
 
         // second output for last lstm output
-        auto last_output = prog.add_instruction(op::rnn_last_output{}, hidden_states);
+        std::vector<instruction_ref> vec_args;
+        vec_args.push_back(hidden_states);
+        if(seq_lens != prog.end())
+        {
+            vec_args.push_back(seq_lens);
```
pfultz2, line 1720: What if `seq_lens` is a constant?

src/onnx/onnx.cpp
```diff
@@ -1697,10 +1713,16 @@ struct onnx_parser
...
+        }
+        auto last_output = prog.add_instruction(op::rnn_last_hs_output{dirct}, vec_args);
 
         // third output for last cell output
-        auto last_cell_output = prog.add_instruction(op::lstm_last_cell_output{}, hidden_states);
+        auto last_cell_output = prog.add_instruction(op::lstm_last_cell_output{dirct}, vec_args);
```
pfultz2, line 1725: Why does `lstm_last_cell_output` need to take direction?

src/rewrite_rnn.cpp
```diff
@@ -983,10 +1003,10 @@ std::vector<instruction_ref> rewrite_rnn::lstm_cell(bool is_forward,
...
 
         // equation Ct = ft (.) Ct-1 + it (.) ct
-        auto ft_cell     = prog.insert_instruction(ins, op::mul{}, ft, sic);
-        auto it_ct       = prog.insert_instruction(ins, op::mul{}, it, ct);
-        auto cellt       = prog.insert_instruction(ins, op::add{}, ft_cell, it_ct);
-        last_cell_output = cellt;
+        auto ft_cell = prog.insert_instruction(ins, op::mul{}, ft, sic);
+        auto it_ct   = prog.insert_instruction(ins, op::mul{}, it, ct);
+        auto cellt   = prog.insert_instruction(ins, op::add{}, ft_cell, it_ct);
+        // last_cell_output = cellt;
```
kahmed10, line 1009: still needed?

src/onnx/onnx.cpp
scxiao: Only if seq_lens is a constant and all sequence lengths are the same, we do not need to call these kernels. But if the len is less than the seq_len in the input (of shape [seq_len, batch, input_size], we need to extend the output hidden state to the shape of [seq_len, num_direct, batch, hidden_size] from the shape [len, num_direct, batch, hidden_size] since len < seq_len.

src/onnx/onnx.cpp
scxiao: The cell output is of shape [num_direction, batch_size, hidden_size]. If the any sequence len is less than max_sequence_len, we need the direction information to know whether to retrieve the last element from all_cell_outputs (forward direction) or the first element from the all_cell_outputs (reverse direction)

src/rewrite_rnn.cpp
scxiao: I will remove it

src/onnx/onnx.cpp
pfultz2: But wont the shape be the same for the last element and first element?

scxiao: It is fine for the shape computation. For the case of different input sequence length, the algorithm used to compute the last output in forward and reverse direction is different. Please see the following:
       visit_all(result, args[0])([&](auto output, auto input) {
            args[1].visit([&](auto seq_lens) {
                par_for(output_shape.elements(), [&](auto i) {
                    auto idx = out_comp_s.multi(i);
                    auto b   = idx[2];
                    if(direction == rnn_direction::reverse or idx[1] == 1)
                    {
                        idx[0] = 0;
                    }
                    else
                    {
                        idx[0] = seq_lens[b] - 1;
                    }
                    output[i] = input(idx.begin(), idx.end());
                });
            });
        });
so we need the direction in computing the last output

src/include/migraphx/op/rnn_last_cell_output.hpp
```diff
@@ -1,33 +1,16 @@
...
-#include <cmath>
-#include <utility>
+#include <migraphx/op/rnn_last_output.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace op {
 
-struct lstm_last_cell_output
+struct lstm_last_cell_output : rnn_last_output<lstm_last_cell_output>
```
pfultz2, line 10: We should just add `rnn_last_hs_output` and `rnn_last_cell_output` with the understanding that `rnn_last_cell_output` is only used for lstms.

src/targets/gpu/include/migraphx/gpu/rnn_last_cell_output.hpp
```diff
@@ -0,0 +1,24 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+
+struct context;
+
+struct hip_lstm_last_cell_output : dev_rnn_last_output<op::lstm_last_cell_output>
+{
+    hip_lstm_last_cell_output() {}
+    hip_lstm_last_cell_output(op::lstm_last_cell_output o) : dev_rnn_last_output(o){};
+};
```
pfultz2, line 18: Why do we need different operators for this? Its the same operator.

src/targets/gpu/lowering.cpp
```diff
@@ -184,9 +187,13 @@ struct miopen_apply
         add_extend_op<hip_reduce_min, op::reduce_min>("reduce_min");
         add_extend_op<hip_reduce_prod, op::reduce_prod>("reduce_prod");
         add_extend_op<hip_reduce_sum, op::reduce_sum>("reduce_sum");
+        add_extend_op<hip_rnn_shift_output, op::rnn_shift_output>("rnn_shift_output");
+        add_extend_op<hip_rnn_shift_sequence, op::rnn_shift_sequence>("rnn_shift_sequence");
+        add_extend_op<hip_lstm_last_cell_output, op::lstm_last_cell_output>(
+            "lstm_last_cell_output");
+        add_extend_op<hip_rnn_last_hs_output, op::rnn_last_hs_output>("rnn_last_hs_output");
```
pfultz2, line 194: The rewrite_rnn should replace these operators with a single `rnn_last_output`.

src/include/migraphx/op/rnn_variable_seq_lens.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <migraphx/par_for.hpp>
+#include <migraphx/config.hpp>
+#include <cmath>
+#include <utility>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct rnn_shift_output
```
pfultz2, line 20: Can we prefix all these operators with rnn_variable_` to make it clear its only used for variable sequences.

src/include/migraphx/op/rnn_variable_seq_lens.hpp
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                    output(idx.begin(), idx.end()) = val;
+                });
+            });
+        });
+
+        return result;
+    }
+};
+
```
pfultz2, line 104: Also `rnn_last_output` should be in this header(and should probably be named `rnn_variable_last_output`.

src/rewrite_rnn.cpp
```diff
@@ -1099,6 +1140,146 @@ std::vector<operation> rewrite_rnn::lstm_actv_funcs(instruction_ref ins) const
...
+        auto last_hs_output_it = ins->outputs().begin();
+        while(last_hs_output_it != ins->outputs().end())
+        {
+            last_hs_output_it = std::find_if(last_hs_output_it, ins->outputs().end(), [](auto i) {
+                return i->name() == "rnn_last_hs_output";
+            });
+
+            if(last_hs_output_it != ins->outputs().end())
+            {
+                prog.replace_instruction(*last_hs_output_it, last_hs_output);
```
pfultz2, line 1214: Why are you replacing `last_hs_output_it` with `last_hs_output`?

src/rewrite_rnn.cpp
```diff
@@ -1099,6 +1140,146 @@ std::vector<operation> rewrite_rnn::lstm_actv_funcs(instruction_ref ins) const
...
+                                                    instruction_ref last_hs_output,
+                                                    op::rnn_direction dirct) const
+{
+    bool variable_seq_len = is_variable_seq_lens(prog, seq_lens);
+    instruction_ref result_ins{};
+    if(variable_seq_len)
+    {
+        result_ins = prog.insert_instruction(
+            std::next(ins), op::rnn_shift_output{"hidden_states", dirct}, ins, seq_lens);
+        prog.replace_instruction(ins, result_ins);
```
pfultz2, line 1201: This wont replace the last_hs_output placeholder.

src/include/migraphx/op/rnn_last_cell_output.hpp
scxiao: you mean rename the file name from rnn_last_cell_output.hpp to lstm_last_cell_output here?

src/targets/gpu/include/migraphx/gpu/rnn_last_cell_output.hpp
scxiao: Do you mean it is same with hip_rnn_last_hs_output? Yes, the algorithm used is the same, but we need the different name to indicate whether it is the last_hs_output or last_cell_output.

src/onnx/onnx.cpp
pfultz2: Yes, but you only need the direction for the actual operator. This shouldn't be there for the placeholder. The rewrite_rnn will change it over to use the operator with direction.

src/include/migraphx/op/rnn_last_cell_output.hpp
pfultz2: I actually meant rename the operator to `rnn_last_cell_output` so it matches the filename.

src/targets/gpu/include/migraphx/gpu/rnn_last_cell_output.hpp
pfultz2: You know if its last_hs_output or last_cell_output in rewrite_rnn.

src/targets/gpu/lowering.cpp
scxiao: fixed

src/onnx/onnx.cpp

src/include/migraphx/op/rnn_last_cell_output.hpp
scxiao: fixed

src/targets/gpu/include/migraphx/gpu/rnn_last_cell_output.hpp
scxiao: fixed

src/rewrite_rnn.cpp
scxiao: This is to replace the hs_outputs of the case of all sequences are the same(not the output we need) with the hs_outputs of the case of sequence are of variable length(final output)

src/rewrite_rnn.cpp
scxiao: This code path is what we implemented before for the case of all sequences are of the same length, and they are constants. 

As for the replae_instruciton, last_hs_output_it is a placeholder to the last hiddent state output, while last_hs_output is the instruction that can compute the last hs output that we want.

src/include/migraphx/op/rnn_variable_seq_lens.hpp
scxiao: how about rnn_var_sl_shift_sequence and rnn_var_sl_shift_output?

scxiao: renamed

src/include/migraphx/op/rnn_variable_seq_lens.hpp
scxiao: fixed

src/include/migraphx/op/rnn_last_output.hpp
```diff
@@ -1,33 +1,64 @@
...
     {
-        check_shapes{inputs, *this}.has(1);
         auto dims = inputs[0].lens();
 
         // remove the first dimension, remaing are output shape
         dims.erase(dims.begin());
         return {inputs[0].type(), dims};
     }
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
```
pfultz2, line 32: The `compute` method should be implemented in lowering.

src/include/migraphx/op/rnn_last_output.hpp
```diff
@@ -1,33 +1,64 @@
...
+#include <migraphx/op/common.hpp>
+#include <migraphx/op/name.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace op {
 
-struct rnn_last_output
+template <class Derived>
+struct rnn_last_output : op_name<Derived>
```
pfultz2, line 13: Shouldn't this be named `rnn_var_sl_last_output`?

src/include/migraphx/op/rnn_last_hs_output.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <migraphx/op/rnn_last_output.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct rnn_last_hs_output : rnn_last_output<rnn_last_hs_output>
+{
+    rnn_last_hs_output() {}
+    rnn_last_hs_output(rnn_direction dirct) : rnn_last_output(dirct) {}
```
pfultz2, line 13: This shouldn't take the direction. The direction is only needed for `rnn_var_sl_last_output` operator.

src/targets/gpu/lowering.cpp
```diff
@@ -184,9 +185,16 @@ struct miopen_apply
...
         add_extend_op<hip_reduce_prod, op::reduce_prod>("reduce_prod");
         add_extend_op<hip_reduce_sum, op::reduce_sum>("reduce_sum");
+        add_extend_op<hip_rnn_var_sl_shift_output, op::rnn_var_sl_shift_output>(
+            "rnn_var_sl_shift_output");
+        add_extend_op<hip_rnn_var_sl_shift_sequence, op::rnn_var_sl_shift_sequence>(
+            "rnn_var_sl_shift_sequence");
+        add_extend_op<hip_rnn_var_sl_last_output<op::rnn_last_cell_output>,
+                      op::rnn_last_cell_output>("rnn_last_cell_output");
+        add_extend_op<hip_rnn_var_sl_last_output<op::rnn_last_hs_output>, op::rnn_last_hs_output>(
+            "rnn_last_hs_output");
```
pfultz2, line 195: `rnn_last_hs_output` and `rnn_last_cell_output` should be replaced in rewrite_rnn.

src/include/migraphx/op/rnn_last_hs_output.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef MIGRAPHX_GUARD_OPERATORS_RNN_LAST_HS_OUTPUT_HPP
+#define MIGRAPHX_GUARD_OPERATORS_RNN_LAST_HS_OUTPUT_HPP
+
+#include <migraphx/op/rnn_var_sl_last_output.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct rnn_last_hs_output : rnn_var_sl_last_output
```
pfultz2, line 10: Why does this inherit `rnn_var_sl_last_output`?

src/rewrite_rnn.cpp
```diff
@@ -1099,6 +1141,172 @@ std::vector<operation> rewrite_rnn::lstm_actv_funcs(instruction_ref ins) const
...
+    instruction_ref result_ins{};
+    if(variable_seq_len)
+    {
+        result_ins = prog.insert_instruction(
+            std::next(ins), op::rnn_var_sl_shift_output{"hidden_states", dirct}, ins, seq_lens);
+        prog.replace_instruction(ins, result_ins);
+
+        // correct the direction used for the operator
+        auto last_hs_output_it = result_ins->outputs().begin();
+        while(last_hs_output_it != result_ins->outputs().end())
```
pfultz2, line 1206: It would be nice(but I dont know if there is time to update it) to implement this in terms of this `find_all` function(which could be added to ranges.hpp):

```cpp
template <class R>
using range_value = std::decay_t<decltype(*std::declval<R>().begin())>;

template<class Range, class Predicate>
std::vector<range_value<Range>> find_all(Range&& r, Predicate p)
{
    std::vector<range_value<Range>> result;
    std::copy_if(r.begin(), r.end(), std::back_inserter(result), p);
    return result;
}
```

src/include/migraphx/op/rnn_last_hs_output.hpp
scxiao: I want to reuse the compute_shape() function of rnn_var_sl_last_out, just fixed following your suggestion

PR #520, opened 2020-05-06
 Add more constructor to C api for shape 

pfultz2:


src/api/api.cpp
```diff
@@ -611,16 +629,16 @@ extern "C" migraphx_status migraphx_onnx_options_create(migraphx_onnx_options_t*
...
+    migraphx_onnx_options_t onnx_options, const char* name, size_t* dims, size_t dims_size)
 {
     return migraphx::try_([&] {
         if(onnx_options == nullptr)
             MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter onnx_options: Null pointer");
-        migraphx::set_input_parameter_shape((onnx_options->object), (name), (dims), (dim_num));
+        if(dims == nullptr)
+            MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter dims: Null pointer");
+        migraphx::set_input_parameter_shape(
+            (onnx_options->object), (name), (std::vector<size_t>(dims, dims + dims_size)));
```
scxiao, line 641: This change revert to the code in my previous PR parse_onnx_using_explicit_input_shape. The problem here is that if dim is nullptr, an exception is thrown.

But in the function onnx_options::set_input_parameter_shape(name, dims.data(), dim.size()) in migraphx.hpp, if dim is empty (used to represent a scalar shape in both numpy and ort), we will throw an exception.
That is the reason I made some changes in my PR python_scalar_input_shape.

src/api/include/migraphx/migraphx.h
```diff
@@ -77,6 +77,13 @@ migraphx_status migraphx_shape_create(migraphx_shape_t* shape,
                                       size_t* lengths,
                                       size_t lengths_size);
 
+migraphx_status migraphx_shape_create_with_strides(migraphx_shape_t* shape,
+                                                   migraphx_shape_datatype_t type,
+                                                   size_t* lengths,
+                                                   size_t lengths_size,
+                                                   size_t* strides,
+                                                   size_t strides_size);
+
```
scxiao, line 86: This function still cannot handle the case of a scalar shape for the function set_input_parameter_shape(name, dims), where dims is a empty vector

src/api/migraphx.py
```diff
@@ -176,9 +181,7 @@ def onnx_options(h):
     h.constructor('create')
     h.method(
         'set_input_parameter_shape',
-        api.params(name='const char*',
-                   dims='const size_t *',
```
scxiao, line 180: Same as above, the reason we change to this API is to process the case of dims.datat() is nullptr.

test/api/test_cpu.cpp
```diff
@@ -64,21 +64,22 @@ TEST_CASE(zero_parameter)
...
-    CHECK(bool{s1 == s1_orig});
+    auto s = migraphx::shape(migraphx_shape_float_type);
+    EXPECT(s.lengths().size() == 1);
+    EXPECT(s.strides().size() == 1);
+    EXPECT(s.lengths().front() == 1);
+    EXPECT(s.strides().front() == 0);
+}
 
-    migraphx::onnx_options option;
-    option.set_input_parameter_shape("1", {});
```
scxiao, line 76: This line will throw an exception and this is what we meet in ORT

src/api/api.cpp
pfultz2: Sorry, I am not sure why this was reverted, maybe a merge mistake. I will try to fix it.

test/api/test_cpu.cpp
pfultz2: I added this test back in and I fixed the null pointer error.

PR #521, opened 2020-05-07
Add gelu optimization

kahmed10:
#434


src/include/migraphx/matcher.hpp
```diff
@@ -535,6 +535,19 @@ auto same_shape(Ms... ms)
     return all_of(same_shape(ms)...);
 }
 
+template <class T>
+inline auto has_value(T x, float tolerance = 1e-6)
+{
+    return make_basic_pred_matcher([=](instruction_ref ins) {
+        auto l = ins->get_literal();
+        if(l.empty())
```
pfultz2, line 543: This should check that there is only one element(ie `ins->get_shape().elements() == 1`) since you are only checking the first element.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,115 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
 
+struct find_gelu
+{
+
+    static auto erf_fn()
+    {
+        return match::name("gpu::erf")(
+            match::arg(0)(match::used_once(),
+                          match::name("gpu::mul")(match::arg(0)(match::any().bind("x")),
+                                                  match::arg(1)(match::has_value(M_SQRT1_2)))));
```
pfultz2, line 342: Multiplication is commutative, so you should use `match::either_arg(0, 1)`. Instead of `match::any()`, you might need to do `match::none_of(match::has_value(M_SQRT1_2))`.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,115 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+    }
+
+    auto matcher() const
+    {
+        return match::name("gpu::mul")(match::arg(0)(match::name("gpu::mul")(
```
pfultz2, line 347: This should use `either_arg` as well.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,115 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+                                           match::arg(1)(match::args(match::has_value(0.5f))))),
```
pfultz2, line 348: This should be `either_arg`. There is no guarantee that this will be the second argument.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,115 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+                                       match::arg(1)(match::name("gpu::add")(
+                                           match::used_once(),
+                                           match::arg(0)(match::used_once(), erf_fn()),
+                                           match::arg(1)(match::args(match::has_value(1.0f))))));
```
pfultz2, line 352: This should be `either_arg(0, 1)`.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,115 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+
+    static auto pow_fn()
+    {
+        return match::name("gpu::pow")(match::arg(1)(match::args(match::has_value(3.0f))));
+    }
+
+    static auto tanh_fn()
+    {
+        return match::name("gpu::tanh")(match::arg(0)(match::name("gpu::mul")(
+            match::arg(0)(match::args(match::has_value(sqrt(M_2_PI)))),
```
pfultz2, line 396: Use `either_arg` here.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,115 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+            match::arg(1)(match::name("gpu::add")(match::arg(1)(
```
pfultz2, line 397: Use `either_arg` here.

kahmed10:
@pfultz2 changed the matcher functions to use either_arg() where appropriate.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,116 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+        return match::name("gpu::pow")(match::arg(1)(match::args(match::has_value(3.0f))));
+    }
+
+    static auto tanh_fn()
+    {
+        return match::name(
+            "gpu::tanh")(match::arg(0)(match::name("gpu::mul")(match::either_arg(0, 1)(
+            match::args(match::has_value(sqrt(M_2_PI))),
+            match::name("gpu::add")(match::any_arg(0, 1)(match::name("gpu::mul")(
+                match::either_arg(0, 1)(match::args(match::has_value(0.044715f)), pow_fn()))))))));
```
scxiao, line 400: I am not sure whether the pow_fn need a used_once here?

src/targets/gpu/fuse_ops.cpp
```diff
@@ -310,6 +331,116 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+    }
+
+    auto matcher() const
+    {
+        return match::name("gpu::mul")(match::either_arg(0, 1)(
+            match::any().bind("x"),
+            match::name("gpu::add")(match::any_arg(0, 1)(match::name("gpu::mul")(
+                match::either_arg(0, 1)(match::args(match::has_value(0.5f)), tanh_fn()))))));
```
scxiao, line 408: same for the tanh_fn(), do we need a used_once for it? Also for the gpu::mul here/

src/targets/gpu/device/gelu.cpp
```diff
@@ -0,0 +1,60 @@
...
+auto gelu_fn(T x) __device__
+{
+    return x * 0.5 * (1 + ::erf(x * M_SQRT1_2));
+}
+
+// 0.5 * x * (1 + tanh(sqrt(2 / pi) * (x + 0.044715 * pow(x, 3))))
+template <class T>
+auto gelu_fn_new(T x) __device__
+{
+    return 0.5 * x * (1 + tanh(sqrt(M_2_PI) * (x + 0.044715 * pow(x, 3))));
```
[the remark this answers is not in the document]
kahmed10: In this case, I don't think it matters since gelu_fn() will be called instead by default. There is a slight accuracy penalty, but the performance is much better.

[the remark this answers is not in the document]
kahmed10: I will change it anyways since the pow() function is much slower regardless.

src/targets/gpu/fuse_ops.cpp
kahmed10: changed

src/targets/gpu/fuse_ops.cpp
kahmed10: changed

Issue #511, opened 2020-04-26
bug in eliminate common subexpression


kahmed10:
Related to #475 performance regressions. I was able to reproduce the results by creating a test with multiple literals with the same value (like in relu6):
```
TEST_CASE(cse_test_literal)
{
    migraphx::program p1;
    {
        auto six1  = p1.add_literal(6);
        auto zero1  = p1.add_literal(0);
        auto six2  = p1.add_literal(6);
        auto zero2  = p1.add_literal(0);
        auto six3  = p1.add_literal(6);
        auto zero3  = p1.add_literal(0);

        auto sum1 = p1.add_instruction(migraphx::op::add{}, six1, zero1);
        auto sum2 = p1.add_instruction(migraphx::op::add{}, six2, zero2);
        auto sum3 = p1.add_instruction(migraphx::op::add{}, six3, zero3);
        auto sum4 = p1.add_instruction(migraphx::op::add{}, sum1, sum2);
        auto sum5 = p1.add_instruction(migraphx::op::add{}, sum3, sum4);
        p1.add_instruction(pass_op{}, sum5);
    }
    run_pass(p1);

    migraphx::program p2;
    {
        auto six  = p2.add_literal(6);
        auto zero = p2.add_literal(0);
        auto sum1 = p2.add_instruction(migraphx::op::add{}, six, zero);
        auto sum2 = p2.add_instruction(migraphx::op::add{}, sum1, sum1);
        auto sum3 = p2.add_instruction(migraphx::op::add{}, sum2, sum1);
        p2.add_instruction(pass_op{}, sum3);
    }
    EXPECT(p1 == p2);
}

...

FAILED: p1 == p2  [ @0 = @literal{0} -> int32_type, {1}, {0}
@1 = @literal{6} -> int32_type, {1}, {0}
@2 = @literal{0} -> int32_type, {1}, {0}
@3 = @literal{6} -> int32_type, {1}, {0}
@4 = add(@3,@2) -> int32_type, {1}, {0}
@5 = add(@1,@0) -> int32_type, {1}, {0}
@6 = add(@4,@5) -> int32_type, {1}, {0}
@7 = add(@5,@6) -> int32_type, {1}, {0}
@8 = pass(@7) -> int32_type, {1}, {0}
 == @0 = @literal{0} -> int32_type, {1}, {0}
@1 = @literal{6} -> int32_type, {1}, {0}
[...]

PR #524, opened 2020-05-10
cse literal fix

kahmed10:
Fixes issue where multiple literals of the same value were not completely getting removed. #511


src/eliminate_common_subexpression.cpp
```diff
@@ -14,19 +14,22 @@ template <class Range>
...
     {
         // Skip dead instructions
         if(ins->outputs().empty())
             continue;
-        // Find instruction with the same name
         auto found_instructions = range(instructions.equal_range(ins->name()));
         for(const auto& pp : found_instructions)
         {
             auto eq = pp.second;
+            if(processed_ins.find(eq) != processed_ins.end())
```
pfultz2, line 27: Maybe write this as `contains(processed_ins, eq)` instead. It seems clearer.

PR #525, opened 2020-05-12
Compute dominators

pfultz2:
This will utilize the dominators in our scheduling for computing the memory conflicts.


pfultz2:
Did you merge the #531 fix? Otherwise, it wont verify.

scxiao:
> Did you merge the #531 fix? Otherwise, it wont verify.

Yes, #531 is included in the test, so I am sure the issue is caused by the dominator PR

pfultz2:
@scxiao @kahmed10 I think this is ready to be reviewed.

src/opt/memory_coloring_impl.cpp
```diff
@@ -249,8 +249,8 @@ void memory_coloring_impl::verify()
 
             if(segment.begin == invalid_offset)
             {
-                if(!interval.is_live_on_entry)
-                    MIGRAPHX_THROW("interval is not live on entry");
+                // if(!interval.is_live_on_entry)
```
kahmed10, line 252: can these comments be removed?

test/memory_coloring_test.cpp
```diff
@@ -680,6 +714,3046 @@ TEST_CASE(test39)
     CHECK(no_allocate(*else_mod));
 }
 
+TEST_CASE(rnn_dom)
+{
```
kahmed10, line 718: should we move this to another file or generate the module in a way to avoid the tidy complexity warning?

src/opt/memory_coloring_impl.cpp
pfultz2: I am not sure, this seem to error on some cases, but I am not sure if these checks are needed or not.

test/memory_coloring_test.cpp
pfultz2: I dont think there is, I can just add a `NOLINT` to the function.

src/opt/memory_coloring_impl.cpp
kahmed10: right now it's commented out. Is it still throwing an error if deleted? Or are you saying we should keep these for reference in case we need to activate these checks in the future? If that's the case, maybe we should just add a comment to explain this.

kahmed10 (approved):

scxiao (approved):

pfultz2:
I think this can be merged.

scxiao:
I tried this PR for all MSFT models, no regression was observed.

Issue #481, opened 2020-04-02
add Range op


kahmed10:
found in opset 11 (used by bert squad example from [onnxruntime](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/python/tools/bert/notebooks/PyTorch_Bert-Squad_OnnxRuntime_GPU.ipynb))

PR #529, opened 2020-05-14
Add range op

kahmed10:
#481


src/onnx/onnx.cpp
```diff
@@ -1989,6 +1990,47 @@ struct onnx_parser
...
+            assert(delta_val != 0);
+
+            size_t num_elements = static_cast<size_t>(
+                ceil(static_cast<double>(limit_val - start_val) / static_cast<double>(delta_val)));
+
+            assert(num_elements > 0);
+
+            using type = decltype(start_val);
+
+            std::vector<type> range_vals;
```
pfultz2, line 2022: You should preallocate the vector.

src/onnx/onnx.cpp
```diff
@@ -1989,6 +1990,47 @@ struct onnx_parser
...
+
+            for(int64_t i = 0; i < num_elements; i++)
+            {
+                range_vals.push_back(start_val + static_cast<type>(i) * delta_val);
+            }
```
pfultz2, line 2027: You could write this using `std::generate`:

```cpp
std::generate(range_vals.begin(), range_vals.end(), [&]() { 
    auto result = start_val;
    start_val += delta_val;
    return result;
});
```

src/onnx/onnx.cpp

src/onnx/onnx.cpp
```diff
@@ -1989,6 +1990,47 @@ struct onnx_parser
...
+
+        auto start_arg = args[0]->eval();
+        check_arg_empty(start_arg, "PARSE_RANGE: start arg dynamic shape is not supported");
+        auto limit_arg = args[1]->eval();
+        check_arg_empty(limit_arg, "PARSE_RANGE: limit arg dynamic shape is not supported");
+        auto delta_arg = args[2]->eval();
+        check_arg_empty(delta_arg, "PARSE_RANGE: delta arg dynamic shape is not supported");
+
+        assert(args[0]->get_shape().elements() == args[1]->get_shape().elements() ==
+               args[2]->get_shape().elements() == 1);
```
pfultz2, line 2005: Actually, this does not do what is expected. `args[0]->get_shape().elements() == args[1]->get_shape().elements()` will return a boolean. It should be `args[0]->get_shape().elements() == 1 && args[1]->get_shape().elements() == 1 && args[2]->get_shape().elements() == 1`.

kahmed10: fixed

PR #531, opened 2020-05-18
Skip fusing group convolutions

pfultz2:


pfultz2:
It passes the CI, but not the coverage because there is no test for tracing added. I am not sure if its necessary right now.

scxiao:
> It passes the CI, but not the coverage because there is no test for tracing added. I am not sure if its necessary right now.

I am fine to merge it.

PR #537, opened 2020-05-21
Enable read support for n-dimensional ops

kahmed10:
Changes made to onnx to enable reading n-d batchnorm, conv, and pooling


src/include/migraphx/op/pooling.hpp
```diff
@@ -20,11 +20,11 @@ namespace op {
 
 struct pooling
 {
-    std::string mode                   = "average";
-    std::array<std::size_t, 2> padding = {{0, 0}};
-    std::array<std::size_t, 2> stride  = {{1, 1}};
-    std::array<std::size_t, 2> lengths = {{1, 1}};
-    padding_mode_t padding_mode        = default_;
+    std::string mode = "average";
+    std::vector<std::size_t> padding{0, 0};
```
pfultz2, line 24: Is it possible to write this as `std::vector<std::size_t> padding = {0, 0}`? The formatter will align these values.

kahmed10: I changed it for the other operators as well.

src/onnx/onnx.cpp
```diff
@@ -323,16 +323,33 @@ struct onnx_parser
                             Op& op,
                             float pad_val = 0)
     {
-        if(padding[0] != padding[2] || padding[1] != padding[3])
+        bool asym_padding = false;
+        size_t pad_ndims  = padding.size() / 2;
```
scxiao, line 327: we need a check here to ensure padding.size() is an even number.

src/include/migraphx/op/pad.hpp
```diff
@@ -51,6 +51,8 @@ struct pad
         return s;
     }
 
+    std::size_t pad_ndims() const { return pads.size() / 2; }
+
```
scxiao, line 55: We need a check to ensure pads.size() is an even number. 
Or we can separate the start and end pads of each dimension, then it is more readable when the graph is print out.

src/include/migraphx/op/pooling.hpp
```diff
@@ -40,29 +40,27 @@ struct pooling
...
-        check_shapes{inputs, *this}.has(1).only_dims(4);
+        check_shapes{inputs, *this}.has(1);
 
         const shape& input = inputs.at(0);
         auto t             = input.type();
 
-        assert(lengths[0] <= (input.lens()[2] + 2 * padding[0]));
-        assert(lengths[1] <= (input.lens()[3] + 2 * padding[1]));
+        auto input_lens = input.lens();
+        size_t kdims    = input_lens.size() - 2;
```
scxiao, line 50: We need a bunch of checks to ensure the dimensions of the inputs, padding, strides, lengths to match since it can be any numbers independently.

src/include/migraphx/op/quant_convolution.hpp
```diff
@@ -19,9 +19,9 @@ namespace op {
 
 struct quant_convolution
 {
-    std::array<std::size_t, 2> padding  = {{0, 0}};
-    std::array<std::size_t, 2> stride   = {{1, 1}};
-    std::array<std::size_t, 2> dilation = {{1, 1}};
+    std::vector<std::size_t> padding  = {0, 0};
+    std::vector<std::size_t> stride   = {1, 1};
+    std::vector<std::size_t> dilation = {1, 1};
```
scxiao, line 24: The compte_shape() function needs to be updated to reflect the multiple dimensions (same as the convolution). Also, same as above, we need check to ensure dimensions to match among inputs and attributes.

src/include/migraphx/op/batch_norm.hpp
```diff
@@ -42,7 +42,7 @@ struct batch_norm_inference
     shape compute_shape(std::vector<shape> inputs) const
     {
         check_shapes{inputs, *this}.has(5);
-        check_shapes{inputs.data(), inputs.data() + 1, *this}.only_dims(4);
+        check_shapes{inputs.data(), inputs.data() + 1, *this}.same_ndims();
```
pfultz2, line 45: You should update the gpu operator to check for 4 dims.

src/onnx/onnx.cpp
```diff
@@ -441,11 +458,18 @@ struct onnx_parser
...
-            std::vector<int64_t> padding(in_lens.size());
-            calculate_padding(
-                0, padding, in_lens[2], op.stride[0], dilation[0], k_lens[0], is_same_upper);
-            calculate_padding(
-                1, padding, in_lens[3], op.stride[1], dilation[1], k_lens[1], is_same_upper);
+            std::vector<int64_t> padding(2 * (in_lens.size() - 2));
+
+            for(size_t i = 0; i < padding.size() / 2; i++)
+            {
+                calculate_padding(i,
```
scxiao, line 465: The calculate_padding function needs to be changed as:
auto pad_ndim = pads.size() / 2;
pads[idx] = pad/2;
pads[idx + pad_ndim] = pad - pad / 2;

src/onnx/onnx.cpp
```diff
@@ -427,8 +444,8 @@ struct onnx_parser
     instruction_ref process_auto_pad_attribute(instruction_ref ins,
                                                node_info info,
                                                Op& op,
-                                               std::array<std::size_t, 2> k_lens,
-                                               std::array<std::size_t, 2> dilation,
+                                               std::vector<std::size_t> k_lens,
+                                               std::vector<std::size_t> dilation,
```
scxiao, line 448: Again, we need to check the dimensions of k_lens, dilation, in_lens match each other.

test/op_shape_test.cpp
```diff
@@ -74,6 +74,19 @@ TEST_CASE(convolution_shape)
...
+    migraphx::shape weights_1d{migraphx::shape::float_type, {4, 3, 3}};
+    expect_shape(output_1d, migraphx::op::convolution{{0}, {1}, {1}}, input_1d, weights_1d);
+
+    migraphx::shape output_3d{migraphx::shape::float_type, {4, 4, 1, 1, 1}};
+    migraphx::shape input_3d{migraphx::shape::float_type, {4, 3, 3, 3, 3}};
+    migraphx::shape weights_3d{migraphx::shape::float_type, {4, 3, 3, 3, 3}};
+    expect_shape(output_3d,
+                 migraphx::op::convolution{{0, 0, 0}, {1, 1, 1}, {1, 1, 1}},
+                 input_3d,
+                 weights_3d);
```
scxiao, line 89: After adding the checks, we can add some rainy tests to ensure our code can figured out the incorrect inputs.

src/include/migraphx/op/batch_norm.hpp
kahmed10: Should it be 4 dims on the GPU? I thought MIOpen can support up to 3D batchnorm?

src/include/migraphx/op/quant_convolution.hpp
kahmed10: I didn't make any changes to quant_convolution aside from making the array to vector in case there were any side effects. I will change accordingly.

src/include/migraphx/op/batch_norm.hpp
pfultz2: But for now, I dont think we call the API correctly for 3d convolutions.

src/onnx/onnx.cpp
```diff
@@ -548,51 +600,34 @@ struct onnx_parser
...
-            std::array<std::size_t, 2> k_lens = {weight_lens[2], weight_lens[3]};
+            std::vector<std::size_t> k_lens(weight_lens.begin() + 2, weight_lens.end());
             l0 = process_auto_pad_attribute(l0, info, op, k_lens, op.dilation, in_lens);
         }
         if(contains(info.attributes, "group"))
         {
             op.group = parse_value(info.attributes.at("group")).at<int>();
         }
 
+        recalc_conv_attributes(op, kdims);
```
scxiao, line 629: Should we fill the default value or just throw an exception? 
@pfultz2 what is your opinion?

kahmed10: We should always fill the default values for padding, strides, and dilations based on K-dims because onnx does not require these attributes when creating the node.

scxiao: For an attribute, I agree that if not value is provided, we should set the default value. But if an incorrect value (less than the required number of dimensions) is provided, we should throw an exception. Looks like the above logic will also pad the missing dimensions with the default value.

So, for each attribute, the logic would be like:
if (contains())
{
    
}
else
{
    std::fill_n();
}

if (op.padding.size() != kdims)
{
    MIGRAPHX_THROW()
}

src/onnx/onnx.cpp
```diff
@@ -529,13 +558,36 @@ struct onnx_parser
...
+
     template <class Op>
     instruction_ref
     parse_conv(const std::string&, node_info info, std::vector<instruction_ref> args)
     {
         Op op;
         auto l0      = args[0];
         auto weights = args[1];
+        auto in_lens = l0->get_shape().lens();
+        auto kdims   = in_lens.size() - 2;
```
pfultz2, line 589: Should we check that `in_lens.size()` is greater than `2`?

src/onnx/onnx.cpp
```diff
@@ -705,11 +740,13 @@ struct onnx_parser
     parse_pooling(const std::string& name, node_info info, std::vector<instruction_ref> args)
     {
         op::pooling op{ends_with(name, "MaxPool") ? "max" : "average"};
-        auto l0 = args[0];
+        auto l0      = args[0];
+        auto in_lens = l0->get_shape().lens();
+        auto kdims   = in_lens.size() - 2;
```
pfultz2, line 745: Should we check that `in_lens.size()` is greater than 2?

src/onnx/onnx.cpp
kahmed10: I think instead I will make a function that checks the individual attributes if present in the onnx file. This way it'll be easier to hit code coverage as well.

src/onnx/onnx.cpp

src/onnx/onnx.cpp

src/onnx/onnx.cpp
kahmed10: you can check to see the new changes I made to see if this is better. I added an extra test case for coverage as well.

src/onnx/onnx.cpp
```diff
@@ -529,13 +558,46 @@ struct onnx_parser
...
+
     template <class Op>
     instruction_ref
     parse_conv(const std::string&, node_info info, std::vector<instruction_ref> args)
     {
         Op op;
         auto l0      = args[0];
         auto weights = args[1];
+        auto in_lens = l0->get_shape().lens();
+        assert(in_lens > 2);
```
pfultz2, line 598: Shouldn't this be `in_lens.size() > 2`?

PR #541, opened 2020-05-26
Separate gpu unittests to multiple files

scxiao:


test/gpu/test_utils.hpp
```diff
@@ -0,0 +1,223 @@
...
+};
+std::array<std::function<void()>, 2> auto_print::handlers = {};
+
+template <class T>
+auto get_hash(const T& x)
+{
+    return std::hash<T>{}(x);
+}
+
+void compile_check(migraphx::program& p, const migraphx::target& t, bool show_trace = false)
```
pfultz2, line 92: This function needs to be declared as `inline`.

test/gpu/test_utils.hpp
```diff
@@ -0,0 +1,223 @@
...
+            catch(const std::exception& e)
+            {
+                std::cout << "    what(): " << e.what() << std::endl;
+            }
+            std::cout << std::endl;
+            for(auto&& handle : auto_print::handlers)
+                handle();
+        });
+    }
+    static std::array<std::function<void()>, 2> handlers;
```
pfultz2, line 71: To get rid of the clang-tidy issues, I think we need to make this a static local variable, and then only access it through a function:

```cpp
static std::array<std::function<void()>, 2>& handlers
{
    static std::array<std::function<void()>, 2> r = {};
    return r;
}
```

scxiao: fixed

PR #542, opened 2020-05-27
Bert fuse slice reshape trans contiguous

scxiao:


src/simplify_algebra.cpp
```diff
@@ -708,6 +710,109 @@ struct find_rsqrt
...
 
+auto slice_reshape_transpose_ops(const std::string& input, const std::string& reshape)
+{
+    return match::name("transpose")(
+        match::arg(0)(match::name("reshape")(match::arg(0)(match::name("contiguous")(
+                                                 match::arg(0)(match::name("slice").bind(input)))))
+                          .bind(reshape)));
+}
+
+struct find_slice_reshape_transpose
```
pfultz2, line 721: This should be split into two separate classes. There should be a `find_slice_reshape` that fuses the reshape across slices, and then a `find_slice_transpose` that fuses the transpose across slices.

pfultz2:
This also should use `get_splits` to ensure that the slices are from the same input.

src/simplify_algebra.cpp
```diff
@@ -708,6 +710,109 @@ struct find_rsqrt
...
+
+    void apply(program& p, match::matcher_result r) const
+    {
+        auto ins = r.result;
+        auto slc = r.instructions["slice"];
+        auto rsp = r.instructions["reshape"];
+
+        auto tran_op               = any_cast<op::transpose>(ins->get_operator());
+        auto perm                  = tran_op.dims;
+        std::vector<int64_t> perm0 = {0, 2, 1, 3};
```
pfultz2, line 785: Why are these hardcoded?

src/simplify_algebra.cpp
```diff
@@ -708,6 +710,109 @@ struct find_rsqrt
...
+        std::transform(ends.begin(), ends.end(), ends.begin(), [](auto i) { return i / 64; });
+
+        auto rsp_op = any_cast<op::reshape>(rsp->get_operator());
+        auto dims   = rsp_op.dims;
+        dims[2] *= 3;
+
+        auto input = slc->inputs().front();
+        auto r_ins = get_reshape_ins(p, input, dims);
+        auto t_ins = get_trans_ins(p, r_ins, perm);
+        p.replace_instruction(ins, op::slice{{1}, starts, ends}, t_ins);
```
pfultz2, line 812: Why is the axis hardcoded? It could be split across any axis.

src/simplify_algebra.cpp
```diff
@@ -708,6 +710,109 @@ struct find_rsqrt
...
+        if(slc_op.axes != axes)
+        {
+            return;
+        }
+
+        auto starts = slc_op.starts;
+        auto ends   = slc_op.ends;
+
+        std::transform(starts.begin(), starts.end(), starts.begin(), [](auto i) { return i / 64; });
+        std::transform(ends.begin(), ends.end(), ends.begin(), [](auto i) { return i / 64; });
```
pfultz2, line 803: Why do we divide this by `64`?

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,80 @@ struct find_rsqrt
...
+
+        // replace the original reshape with slice
+        int64_t i = 0;
+        for(auto in : vec_rsp)
+        {
+            p.replace_instruction(
+                in, op::slice{{axis}, {i * dim_size}, {(i + 1) * dim_size}}, rsp_ins);
+            ++i;
+        }
+        // remove the original slice instructions
```
pfultz2, line 782: We dont need to remove the instructions manually. They should be removed automatically by DCE.

src/simplify_algebra.cpp
scxiao: fixed accordingly

src/simplify_algebra.cpp

src/simplify_algebra.cpp

src/simplify_algebra.cpp
scxiao: removed

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,128 @@ struct find_rsqrt
...
+           }))
+        {
+            return;
+        }
+
+        auto axis    = any_cast<op::slice>(slc->get_operator()).axes[0];
+        auto in_lens = input->get_shape().lens();
+        if(!std::equal(in_lens.begin(), in_lens.begin() + axis, dims.begin(), dims.begin() + axis))
+        {
+            return;
```
pfultz2, line 762: Are you able to add a test case to cover this branch?

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,128 @@ struct find_rsqrt
...
+        // insert an transpose instruction
+        auto tr = p.insert_instruction(std::next(input), op::transpose{perm}, input);
+
+        // compute the axis in the slice
+        auto axis = any_cast<op::slice>(slc->get_operator()).axes.front();
+
+        auto it = std::find(perm.begin(), perm.end(), axis);
+        if(it == perm.end())
+        {
+            return;
```
pfultz2, line 826: Can you add a test case to cover this branch?

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,128 @@ struct find_rsqrt
...
+        auto input         = slc->inputs().front();
+        auto split_outputs = get_splits(input);
+        if(split_outputs.empty())
+        {
+            return;
+        }
+
+        // all transpose are the same
+        auto perm = any_cast<op::transpose>(trans->get_operator()).dims;
+        if(!std::all_of(split_outputs.begin(), split_outputs.end(), [&](auto i) {
```
kahmed10, line 807: I think it's possible to break this into two steps: first get the instructions for all of the split outputs. Then, you can create a function that calls std::all_of() on the instructions to check if the dims are the same. This function can also be reused for L750 above. Perhaps something like this:
```
template<class Op>
bool check_dims(std::vector<int64_t> dims, std::vector<instruction_ref> ins, const std::string& op_name)
{
    return std::all_of(ins.begin(), ins.end(), [&](auto i) {
        return(i->name() == op_name) and (any_cast<Op>(i->get_operator()).dims == dims);
    });
}
```

src/simplify_algebra.cpp
scxiao: Looks like this is not going to happen, I just removed this check and add test case to cover other checks.

src/simplify_algebra.cpp
scxiao: Added.

src/simplify_algebra.cpp
scxiao: Fixed

src/simplify_algebra.cpp
pfultz2: Well you could make it an assert to help improve debugging.

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,132 @@ struct find_rsqrt
     }
 };
 
+template <class Op>
+static bool check_dims(const std::string& name,
+                       std::vector<instruction_ref>& vec_ins,
+                       const std::vector<int64_t>& dims)
+{
+    return std::all_of(vec_ins.begin(), vec_ins.end(), [&](auto i) {
+        return (i->name() == name) and (any_cast<Op>(i->get_operator()).dims == dims);
```
pfultz2, line 725: There is no need to pass in the dims. This can be written as:

```cpp
static bool same_ops(const std::vector<instruction_ref>& vec_ins) {
    return std::all_of(vec_ins.begin(), vec_ins.end(), [&](auto i) {
        return i->get_operator() == vec_ins.front()->get_operator();
    };
}
```

The default equality operator will compare both the name and the dims.

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,132 @@ struct find_rsqrt
     }
 };
 
+template <class Op>
+static bool check_dims(const std::string& name,
+                       std::vector<instruction_ref>& vec_ins,
```
pfultz2, line 721: This should be const ref.

src/simplify_algebra.cpp
scxiao: added

src/simplify_algebra.cpp
scxiao: fixed

src/simplify_algebra.cpp
scxiao: fixed

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,131 @@ struct find_rsqrt
     }
 };
 
+template <class Op>
```
pfultz2, line 719: There is no need for this template parameter since its not used.

src/simplify_algebra.cpp
```diff
@@ -714,6 +716,131 @@ struct find_rsqrt
...
+static bool check_dims(const std::vector<instruction_ref>& vec_ins)
```
pfultz2, line 720: This should probably be called `same_ops` since it can check more than just dims.

pfultz2:
What perf improvements do you see with bert with this change?

PR #545, opened 2020-06-03
Embedding bag op (limited support)

kahmed10:
Tested with embedding bag found in the simple version of DLRM.


src/include/migraphx/op/common.hpp
```diff
@@ -31,6 +31,13 @@ enum class rnn_direction
     bidirectional,
 };
 
+enum reduce_mode_t
+{
+    sum,
+    mean,
+    max_
+};
```
pfultz2, line 39: This should go in the onnx.cpp file.

kahmed10: fixed

src/onnx/onnx.cpp
```diff
@@ -2105,6 +2106,47 @@ struct onnx_parser
         return l0;
     }
 
+    enum reduce_mode_t
```
pfultz2, line 2109: This should be `enum class` to avoid name collision.

src/onnx/onnx.cpp
```diff
@@ -2105,6 +2106,47 @@ struct onnx_parser
...
+    {
+        sum,
+        mean,
+        max
```
pfultz2, line 2113: This should probably set the value for each of this since this converted from an integer.

src/onnx/onnx.cpp

src/onnx/onnx.cpp

PR #550, opened 2020-06-11
ND convolution GPU support

kahmed10:
Currently no unit tests, but manually tested for correctness by using conv_1d_test.onnx and conv_3d_test.onnx.


src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -75,21 +75,31 @@ inline tensor_descriptor make_tensor(const migraphx::shape& os, bool pack = fals
     return t;
 }
 
+inline void insert_attribute_dim(std::vector<int>& padding,
```
scxiao, line 78: Maybe we can rename it to insert_attribute_1d_dims

src/targets/gpu/convolution.cpp
```diff
@@ -9,15 +9,38 @@ namespace gpu {
 shape miopen_convolution::compute_shape(const std::vector<shape>& inputs) const
 {
     check_shapes{inputs, *this}.has(4).standard();
-    return op.compute_shape({inputs.at(0), inputs.at(1)});
+    std::vector<shape> conv_inputs(inputs.begin(), inputs.begin() + 2);
+    check_shapes{conv_inputs, *this}.max_ndims(5);
+    return op.compute_shape(conv_inputs);
 }
+
+void recompute_shape(shape& input)
```
scxiao, line 17: How about the function name extend_1d_shape?

kahmed10:
> We need some 1d/3d unit tests to verify its correctness

We won't be able to add unit tests for correctness until the cpu implementation is finished

src/targets/gpu/include/migraphx/gpu/miopen.hpp
kahmed10: fixed

src/targets/gpu/convolution.cpp
kahmed10: I changed the name to "recompute_shape_to_2d"

src/targets/gpu/convolution.cpp
```diff
@@ -9,15 +9,38 @@ namespace gpu {
 shape miopen_convolution::compute_shape(const std::vector<shape>& inputs) const
 {
     check_shapes{inputs, *this}.has(4).standard();
-    return op.compute_shape({inputs.at(0), inputs.at(1)});
+    std::vector<shape> conv_inputs(inputs.begin(), inputs.begin() + 2);
+    check_shapes{conv_inputs, *this}.max_ndims(5);
+    return op.compute_shape(conv_inputs);
 }
+
+inline void recompute_shape_to_2d(shape& input)
```
pfultz2, line 17: This should return the new shape instead of modifying it.

src/targets/gpu/convolution.cpp
```diff
@@ -9,15 +9,38 @@ namespace gpu {
...
-    auto y_desc = make_tensor(output_shape);
+    shape x_shape = args[0].get_shape();
+    shape w_shape = args[1].get_shape();
+    shape y_shape = output_shape;
+
+    recompute_shape_to_2d(x_shape);
+    recompute_shape_to_2d(w_shape);
+    recompute_shape_to_2d(y_shape);
+
+    auto x_desc = make_tensor(x_shape);
```
pfultz2, line 41: These can be written as `auto x_desc = make_tensor(recompute_shape_to_2d(x_shape));`

src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -75,21 +75,31 @@ inline tensor_descriptor make_tensor(const migraphx::shape& os, bool pack = fals
...
         c_mode = miopenGroupConv;
-    miopenInitConvolutionDescriptor(c.get(),
-                                    c_mode,
-                                    op.padding[0],
-                                    op.padding[1],
-                                    op.stride[0],
-                                    op.stride[1],
-                                    op.dilation[0],
-                                    op.dilation[1]);
+    std::vector<int> padding(op.padding.begin(), op.padding.end());
```
pfultz2, line 94: Padding, stride, and dilation is represented as `std::size_t` not `int`.

src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -75,21 +75,31 @@ inline tensor_descriptor make_tensor(const migraphx::shape& os, bool pack = fals
     return t;
 }
 
+inline void insert_attribute_1d_dims(std::vector<int>& padding,
+                                     std::vector<int>& stride,
+                                     std::vector<int>& dilation)
+{
+    padding.insert(padding.begin(), 0);
+    stride.insert(stride.begin(), 1);
+    dilation.insert(dilation.begin(), 1);
```
pfultz2, line 84: Can we just default initialize these and then copy backwards into the vector?

src/targets/gpu/include/migraphx/gpu/miopen.hpp
kahmed10: The miopen call below requires the attribute data to be `int*`, so I just copied the values as ints here.

src/targets/gpu/include/migraphx/gpu/miopen.hpp
kahmed10: changed

src/targets/gpu/convolution.cpp
kahmed10: fixed

src/targets/gpu/convolution.cpp
kahmed10: fixed

src/targets/gpu/include/migraphx/gpu/miopen.hpp
pfultz2: Ok I see.

src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -91,12 +97,11 @@ inline convolution_descriptor make_conv(const T& op)
...
     if(op.group > 1)
         c_mode = miopenGroupConv;
-    std::vector<int> padding(op.padding.begin(), op.padding.end());
-    std::vector<int> stride(op.stride.begin(), op.stride.end());
-    std::vector<int> dilation(op.dilation.begin(), op.dilation.end());
 
-    if(op.kdims() == 1)
-        insert_attribute_1d_dims(padding, stride, dilation);
+    size_t kdims = op.kdims();
+    std::vector<int> padding = get_attributes(op.padding, kdims, 0);
```
pfultz2: I meant do something like this:

```cpp
std::vector<int> padding(std::min(2, kdims), 0);
std::copy_backwards(op.padding.begin(), op.padding.end(), padding.end());
```

kahmed10: this doesn't work if kdims = 3

kahmed10: I was able to fix the above by using `std::max` instead

pfultz2: Yea thats right, it should be `std::max`.

PR #553, opened 2020-06-17
Averagepooling asymetric pading

scxiao:


src/onnx/onnx.cpp
```diff
@@ -765,27 +828,23 @@ struct onnx_parser
             op.lengths = std::vector<size_t>(in_lens.begin() + 2, in_lens.end());
         }
 
-        if(contains(info.attributes, "pads"))
+        // does not support ceil_mode
+        int ceil_mode = 0;
+        if(contains(info.attributes, "ceil_mode"))
         {
-            if(contains(info.attributes, "auto_pad"))
+            ceil_mode = info.attributes.at("ceil_mode").i();
```
kahmed10, line 835: You can probably rewrite this so that you don't need to create an extra `ceil_mode` variable
```
if(info.attributes.at("ceil_mode").i() == 1) MIGRAPHX_THROW(...);
```

src/onnx/onnx.cpp
```diff
@@ -751,6 +754,66 @@ struct onnx_parser
...
+    {
+        // maxpooling or count_include_pad is 1, no change is required.
+        if(op.mode == "max" or count_include_pad == 1)
+        {
+            return;
+        }
+
+        // check assymetric padding
+        bool asym_padding = false;
+        assert(padding.size() % 2 == 0);
```
kahmed10, line 791: I made a function in PR #554 that you can use to check for asym_padding

scxiao: fixed

src/onnx/onnx.cpp

src/onnx/onnx.cpp
```diff
@@ -778,6 +779,51 @@ struct onnx_parser
...
     }
 
+    void
+    tune_padding_to_symmetric(int64_t& left, int64_t& right, const int stride, int64_t& s_start)
+    {
+        s_start = 0;
+        if(left > right)
+        {
+            right   = left;
+            s_start = 0;
```
kahmed10, line 789: is this needed since `s_start` is set to 0 in L785?

scxiao:
@pfultz2  When you get a chance, can you take a look this PR?

src/onnx/onnx.cpp
```diff
@@ -593,6 +553,56 @@ struct onnx_parser
         }
     }
 
+    template <class Op>
+    void cal_auto_padding_size(node_info info,
+                               Op& op,
+                               std::vector<std::size_t> k_lens,
+                               std::vector<std::size_t> dilation,
```
pfultz2, line 560: Can these be const ref?

src/onnx/onnx.cpp
```diff
@@ -593,6 +553,56 @@ struct onnx_parser
...
+                               const std::vector<std::size_t>& in_lens,
+                               std::vector<int64_t>& paddings)
```
pfultz2, line 562: It seems like this should be the return value of the function.

src/onnx/onnx.cpp
```diff
@@ -593,6 +553,56 @@ struct onnx_parser
         }
     }
 
+    template <class Op>
+    void cal_auto_padding_size(node_info info,
```
pfultz2, line 557: This method should be static.

src/onnx/onnx.cpp
```diff
@@ -593,6 +553,56 @@ struct onnx_parser
...
+                                  in_lens[i + 2],
+                                  op.stride[i],
+                                  dilation[i],
+                                  k_lens[i],
+                                  is_same_upper);
+            }
+        }
+    }
+
+    void check_padding_mode(node_info info, const std::string& op_name)
```
pfultz2, line 592: This method should be static as well.

src/onnx/onnx.cpp
```diff
@@ -778,6 +779,50 @@ struct onnx_parser
         return add_bias(args, l1, 1);
     }
 
+    void
+    tune_padding_to_symmetric(int64_t& left, int64_t& right, const int stride, int64_t& s_start)
```
pfultz2, line 783: This should be `static`.

src/onnx/onnx.cpp
```diff
@@ -778,6 +779,50 @@ struct onnx_parser
...
+        {
+            auto diff = right - left;
+            s_start   = (diff + stride - 1) / stride;
+            left      = left + s_start * stride;
+            right     = left;
+        }
+    }
+
+    template <class Op>
+    void tune_padding_size(Op& op,
```
pfultz2, line 800: This should be `static`. Also, should this be templated? This looks like it only works for pooling.

pfultz2:
Will these asymmetrical adjustments need to be applied for tf? Maybe we should seperate these functions out so they can be reused by tf.

src/onnx/onnx.cpp

src/onnx/onnx.cpp
scxiao: yes, paddings is the return value of the function.

src/onnx/onnx.cpp
scxiao: yes, fixed.

src/onnx/onnx.cpp
scxiao: fixed

src/onnx/onnx.cpp
scxiao: fixed

src/onnx/onnx.cpp
scxiao: fixed

PR #554, opened 2020-06-17
Nd deconv read support

kahmed10:


src/include/migraphx/op/deconvolution.hpp
```diff
@@ -39,25 +39,27 @@ struct deconvolution
     std::string name() const { return "deconvolution"; }
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(2).same_type().same_ndims().only_dims(4);
+        check_shapes{inputs, *this}.has(2).same_type().same_ndims().min_ndims(3);
```
scxiao, line 42: since GPU has a max_ndims(5), do we need add here?

kahmed10: the cpu implementation will be able to handle ndims > 5, but we restrict it for the gpu because of miopen

pfultz2: I agree we shouldn't limit max dims to 5.

src/onnx/onnx.cpp
```diff
@@ -705,44 +719,57 @@ struct onnx_parser
...
-        if(asymm_padding)
+        std::vector<int64_t> curr_shape(dims.begin() + 2, dims.end());
+        if(asym_padding)
         {
-            op::slice slice_op;
-            slice_op.axes   = {0, 1, 2, 3};
-            slice_op.starts = {0, 0, 0 + padding[0], 0 + padding[1]};
-            slice_op.ends   = {
-                dims[0], dims[1], curr_shape[0] - padding[2], curr_shape[1] - padding[3]};
+            std::vector<int64_t> axes(dims.size());
```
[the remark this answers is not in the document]

PR #557, opened 2020-06-22
Neg operator

scxiao:


src/onnx/onnx.cpp
```diff
@@ -834,6 +835,14 @@ struct onnx_parser
         return prog.add_instruction(op, l0);
     }
 
+    instruction_ref parse_neg(const std::string&, node_info, std::vector<instruction_ref> args)
+    {
+        auto s = args[0]->get_shape();
+        std::vector<float> zero(s.elements(), 0.0f);
+        auto l_zero = prog.add_literal(literal(s, zero));
+        return prog.add_instruction(op::sub{}, l_zero, args[0]);
```
kahmed10, line 843: we have a neg operator already implemented

scxiao: Looks like we do not have a GPU implementation for neg, so I use the sub operator in the gpu lowering.

PR #558, opened 2020-06-22
Nd deconv GPU support

kahmed10:


kahmed10:
> Can we add some gpu test for 1D and 3D?

I plan on adding tests once the cpu implementation is complete.

src/targets/gpu/deconvolution.cpp
```diff
@@ -9,15 +9,32 @@ namespace gpu {
 shape miopen_deconvolution::compute_shape(const std::vector<shape>& inputs) const
 {
     check_shapes{inputs, *this}.has(4).standard();
-    return op.compute_shape({inputs.at(0), inputs.at(1)});
+    std::vector<shape> conv_inputs(inputs.begin(), inputs.begin() + 2);
+    check_shapes{conv_inputs, *this}.max_ndims(5);
+    return op.compute_shape(conv_inputs);
 }
+
+inline shape recompute_shape_to_2d(const shape& input)
```
scxiao, line 17: How about the name reshape_to_2d() here?

src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -104,14 +104,17 @@ inline convolution_descriptor make_deconv(const T& op)
...
-                                    op.padding[1],
-                                    op.stride[0],
-                                    op.stride[1],
-                                    op.dilation[0],
-                                    op.dilation[1]);
+    int kdims                      = op.kdims();
+    std::vector<int> padding(std::max(2, kdims), 0);
+    std::vector<int> stride(std::max(2, kdims), 1);
+    std::vector<int> dilation(std::max(2, kdims), 1);
+
```
scxiao, line 111: Can we add a assert here to ensure padding.size() == op.padding.size(), same for stride and dilation.

scxiao:
> > Can we add some gpu test for 1D and 3D?
> 
> I plan on adding tests once the cpu implementation is complete.

So there is another PR for the CPU implementation?

src/targets/gpu/include/migraphx/gpu/miopen.hpp
kahmed10: it may not to be equal if the op is 1d, so this check is not necessary. However, when I call op.kdims(), there is an assert to make sure the attributes are consistent.

src/targets/gpu/deconvolution.cpp
kahmed10: I might rename it to reshape_if_1d to make it more clear that this is only done if the op is 1d

kahmed10:
> > > Can we add some gpu test for 1D and 3D?
> > 
> > 
> > I plan on adding tests once the cpu implementation is complete.
> 
> So there is another PR for the CPU implementation?

Yes, I plan to open another PR with those changes including cpu/gpu tests.

src/include/migraphx/op/deconvolution.hpp
```diff
@@ -37,27 +37,41 @@ struct deconvolution
...
+        {
+            MIGRAPHX_THROW("deconvolution: inconsistent attribute sizes");
+        }
+    }
+
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(2).same_type().same_ndims().only_dims(4);
+        check_shapes{inputs, *this}.has(2).same_type().same_ndims().min_ndims(3);
+        check_attribute_size();
```
scxiao, line 53: We need a check to ensure input dim num and attribute dim num match. This is also needed for the convolution. I spend two ours figure out a bug that input dim number is different from the attribute dim number since we did not report such a wrong input scenario.

kahmed10: fixed

src/include/migraphx/op/convolution.hpp
```diff
@@ -55,6 +55,10 @@ struct convolution
         const shape& weights = inputs.at(1);
         auto t               = input.type();
         size_t kdims         = input.lens().size() - 2;
+        if(kdims != this->kdims())
+        {
+            MIGRAPHX_THROW("quant_convolution: input k-dims does not match attribute size");
```
scxiao, line 60: The name here should be "convolution"

kahmed10: fixed

PR #561, opened 2020-06-24
Add conv ND for cpu

pfultz2:


src/targets/cpu/lowering.cpp
```diff
@@ -187,37 +187,61 @@ struct cpu_convolution
...
-                            const auto in_y  = start_y + y;
-                            const auto in_ch = group_id * wei_c + k;
-                            if(in_x >= 0 && in_x < in_h && in_y >= 0 && in_y < in_w)
-                                acc += input(o, in_ch, in_x, in_y) * weights(w, k, x, y);
-                        });
-                        output(o, w, i, j) = acc;
+                auto wei_lens                     = weights.get_shape().lens();
+                auto wei_n                        = wei_lens[0];
+                auto wei_c                        = wei_lens[1];
+                std::vector<std::size_t> win_size = {};
```
kahmed10, line 196: would it be better to initialize `win_size` with the iterators from `wei_lens` instead?
like: `std::vector<std::size_t>win_size(wei_lens.begin() + 1, wei_lens.end());`

pfultz2: Yea, let me update that.

src/targets/cpu/lowering.cpp
```diff
@@ -86,35 +86,39 @@ struct cpu_batch_norm_inference
...
-                    });
+
+                    par_dfor(num_batch, num_channels, image_height, image_width)(
+                        [&](std::size_t n, std::size_t c, std::size_t h, std::size_t w) {
+                            assert((variance(c, h, w) + epsilon) > 0);
+                            result(n, c, h, w) = gamma(c, h, w) *
+                                                     (buffer(n, c, h, w) - mean(c, h, w)) /
+                                                     std::sqrt(variance(c, h, w) + epsilon) +
+                                                 bias(c, h, w);
+                        });
```
scxiao, line 121: There is a merge error here, the deleted part should be kept...

test/cpu_ops_test.cpp
```diff
@@ -713,109 +714,6 @@ TEST_CASE(im2col_3x3_with_padding_test)
...
-                               -2.39032,
-                               1.17489,
-                               -0.183886,
-                               -0.453904,
-                               -0.239955,
-                               0.288275,
-                               -0.963948};
-    EXPECT(migraphx::verify_range(result_vector, gold));
-}
-
```
scxiao, line 818: This unit test should be kept

test/cpu_ops_test.cpp
```diff
@@ -862,41 +760,6 @@ TEST_CASE(batch_norm_inference_test)
...
-    std::vector<float> result_vector;
-    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
-    std::vector<float> gold = {
-        -0.220005, 0.569376, 0.372612, -0.126798,  -0.0452053, -0.165809, 0.211653,  -0.0783441,
-        0.739245,  0.404024, 0.370239, -0.0430317, -0.0277556, 0.368179,  0.126639,  0.272615,
-        0.0149929, 0.391911, 0.615216, -0.160635,  0.336706,   0.0836764, 0.0787094, -0.278108,
-        -0.103283, 0.585881, 0.458947, 0.156161,   -0.140408,  -0.386246, -0.243006, 0.661551};
-    EXPECT(migraphx::verify_range(result_vector, gold));
-}
-
```
scxiao, line 899: This unit test should be kept

kahmed10:
@mvermeulen the codecov is failing on this PR because we no longer use `par_dfor` in the conv cpu implementation.

mvermeulen:
Not sure why, but we had lcov build issues again
```
+ lcov --remove coverage.info /usr/* --output-file coverage.info
lcov: WARNING: negative counts found in tracefile coverage.info
lcov: ERROR: cannot write to coverage.info!
```

pfultz2:
I think I fixed the codecov error by using an absolute path. The code coverage difference is acceptable to merge I believe.

PR #562, opened 2020-06-25
Cpu batchnorm

scxiao:


scxiao:
> Can you add test cases for 1d and 3d batchnorm?

Added a few unit tests

PR #564, opened 2020-06-30
Gpu batchnorm

scxiao:


src/targets/gpu/batchnorm.cpp
```diff
@@ -8,16 +8,37 @@ namespace gpu {
 shape miopen_batch_norm_inference::compute_shape(const std::vector<shape>& inputs) const
 {
     check_shapes{inputs, *this}.has(6);
+    check_shapes{inputs.data(), inputs.data() + 1, *this}.same_ndims().max_ndims(5);
     return op.compute_shape({inputs.at(0), inputs.at(1), inputs.at(2), inputs.at(3), inputs.at(4)});
 }
 
+inline void reshape_1d_to_2d(shape& input)
+{
```
kahmed10, line 16: Can you rewrite this so that it returns a shape? Then in L39 you can write something like:
```
auto x_desc = make_tensor(reshape_1d_to_2d(x_shape));
```

PR #567, opened 2020-07-06
Bert squad eliminate contiguous

scxiao:
The previous optimization to reorder slice/reshape does not work for the BERT_SQUAD model, refined it to be able to optimize it.


src/simplify_algebra.cpp
```diff
@@ -760,29 +760,42 @@ struct find_split_reshape
...
 
         // insert the reshape instruction
-        auto rsp_ins = p.insert_instruction(std::next(input), op::reshape{rsp_lens}, input);
+        auto rsp_ins = p.insert_instruction(std::next(input), op::reshape{rsp_out_lens}, input);
 
         // replace the original reshape with slice
-        int64_t i = 0;
-        for(auto in : vec_rsp)
+        int64_t start = 0;
+        for(int i = 0; i < vec_rsp.size(); ++i)
```
pfultz2, line 794: Should `i` be `std::size_t`?

scxiao: thanks. fixed.

src/simplify_algebra.cpp
```diff
@@ -760,29 +760,42 @@ struct find_split_reshape
...
-        std::vector<int64_t> rsp_lens(tmp_lens.begin(), tmp_lens.end());
-        int64_t dim_size = rsp_lens[axis];
-        rsp_lens[axis] *= vec_rsp.size();
+        std::vector<int64_t> vec_dims(vec_rsp.size());
+        std::transform(vec_rsp.begin(), vec_rsp.end(), vec_dims.begin(), [&](auto is) {
+            return is->get_shape().lens()[rsp_axis];
+        });
+
+        std::vector<int64_t> rsp_out_lens(rsp_lens.begin(), rsp_lens.end());
+        rsp_out_lens[rsp_axis] = std::accumulate(vec_dims.begin(), vec_dims.end(), 0);
```
pfultz2, line 787: Also, this should be `std::accumulate(vec_dims.begin(), vec_dims.end(), int64_t{0})`.

src/simplify_algebra.cpp
```diff
@@ -760,29 +760,42 @@ struct find_split_reshape
...
+
+        // search the reshape output (standard shape) to decide which axis are
+        // in its output corresponding to the slc_dim_size
+        auto rsp_lens    = rsp->get_shape().lens();
+        auto rsp_strides = rsp->get_shape().strides();
+        rsp_strides.insert(rsp_strides.begin(), rsp_strides[0] * rsp_lens[0]);
+        auto ait = std::find(rsp_strides.begin(), rsp_strides.end(), slc_dim_size);
+        if(ait == rsp_strides.end())
         {
             return;
```
pfultz2, line 776: Is it possible to write a unit test for this branch?

src/simplify_algebra.cpp

src/simplify_algebra.cpp
scxiao: Yes, added one. Thanks.

PR #568, opened 2020-07-07
Optimize multiply across slices

pfultz2:


src/simplify_algebra.cpp
```diff
@@ -63,6 +63,92 @@ struct find_mul_conv
...
+            return;
+        if(slice_op.axes.front() != 1)
+            return;
+
+        auto slice_idx = std::distance(conv_ins, slice_ins);
+        if(std::any_of(conv_ins->outputs().begin(), conv_ins->outputs().end(), [&](auto i) {
+               if(i == slice_ins)
+                   return false;
+               // Slice should already be checked by the matcher
+               assert(i->name() == "slice");
```
kahmed10, line 105: is the assertion necessary? The matcher would return false if there was a conv with outputs that weren't slices

pfultz2: Yea I dont think it is necessary. The `any_cast` will check the operator as well.

scxiao:
I am running this PR against the models from MSFT now, if no regression, I will approve it

PR #569, opened 2020-07-08
Layernorm changes

kahmed10:


src/targets/gpu/device/layernorm.cpp
```diff
@@ -0,0 +1,66 @@
...
+                     relements;
+
+            idx.local_stride(relements, [&](auto j) __device__ { x(j) = x(j) - m; });
+
+            auto r = block_reduce<max_block_size>(
+                         idx,
+                         sum{},
+                         0,
+                         relements,
+                         [&](auto j) __device__ { return ::pow(to_hip_type(x(j)), 2); }) /
```
scxiao, line 51: Can we rewrite pow(x, 2) to x * x here?

src/targets/gpu/fuse_ops.cpp
```diff
@@ -332,6 +337,48 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+    void apply(program& p, match::matcher_result r) const
+    {
+        auto ins   = r.result;
+        auto x_ins = r.instructions["x"];
+        auto args  = ins->inputs();
+
+        p.replace_instruction(ins, hip_layernorm{}, x_ins, args.back());
+    }
+};
+
```
scxiao, line 381: Is this model used somewhere?

src/targets/gpu/device/layernorm.cpp
```diff
@@ -0,0 +1,66 @@
...
+        using value_type = typename decltype(input)::value_type;
+
+        const std::size_t max_block_size = 256;
+        const std::size_t block_size     = compute_block_size(relements, max_block_size);
+
+        gs_launch(stream, nelements * block_size, block_size)([=](auto i, auto idx) __device__ {
+            const auto out_idx  = i / block_size;
+            const auto base_idx = out_idx * relements;
+            value_type x_data[4];
+            auto x = [&](auto j) -> value_type& { return x_data[j - idx.local]; };
```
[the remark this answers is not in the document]
pfultz2: Yea, I think you are right. Although, we should probably keep track of the depth in the strided loop instead of using division(which can be slow), but I guess we can make that change later.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -332,6 +337,48 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+    static auto layernorm_tf()
+    {
+        return match::either_arg(0, 1)(
+            match::name("gpu::mul")(match::arg(0)(match::name("gpu::add").bind("x"))),
+            match::name("gpu::sub")(match::either_arg(0, 1)(
+                multibroadcast_op(),
+                match::name("gpu::mul")(
+                    match::arg(1)(match::name("gpu::mul")(match::arg(1)(multibroadcast_op()))
+
+                                      )))));
```
pfultz2, line 364: This never matches `reduce_mean`.  Does it need to be `multibroadcast_op(match::name("gpu::reduce_mean"))`?

src/targets/gpu/fuse_ops.cpp
```diff
@@ -332,6 +337,48 @@ void move_standard_front(std::vector<instruction_ref>& args)
...
+    }
+
+    auto matcher() const
+    {
+        return match::name("gpu::add", "gpu::div")(match::any_of(layernorm_onnx(), layernorm_tf()));
```
pfultz2, line 369: This shouldn't match `add` or `div` as these can mean very different things. Is the `div` even necessary? Dont we change it to multiply in simplify_algebra?

src/targets/gpu/fuse_ops.cpp
kahmed10: sorry I still have to add some more specific cases to the matcher that I missed, like you suggested

src/targets/gpu/fuse_ops.cpp
kahmed10: We only change it to multiply in simplify_algebra if the input is a constant value.

PR #574, opened 2020-07-15
integrate onnx backend test suit to migraphx

scxiao:
the backend unit test can be run as: (need to specify the PYTHONPATH to build/lib where the migraphx shared lib for python is located)

python onnx_backend_test_series.py
you can specify the target by -d CPU/GPU

There are a few unit test failed because of bug#573. There are a few other unit tests that are supposed to work also failed, need further investigation.


src/program.cpp
```diff
@@ -284,6 +285,10 @@ instruction_ref program::add_outline(const shape& s)
 instruction_ref program::add_parameter(std::string name, shape s)
 {
     assert(get_parameter_shape(name) == shape{});
+    if (name != "scratch")
```
pfultz2, line 288: Why are we checking for such a parameter name?

Also, we should be compiling with `offload_copy=true`, so this parameter won't get created.

CMakeLists.txt
```diff
@@ -191,8 +191,15 @@ rocm_create_package(
...
 add_subdirectory(doc)
 add_subdirectory(test)
 add_subdirectory(tools)
+
+set(DEST_DIR ${CMAKE_BINARY_DIR})
+file (COPY ${CMAKE_SOURCE_DIR}/src/py/backend/backend.py DESTINATION ${DEST_DIR}/lib/onnx_migraphx/.)
+file (COPY ${CMAKE_SOURCE_DIR}/src/py/backend/backend_rep.py DESTINATION ${DEST_DIR}/lib/onnx_migraphx/.)
+file (COPY ${CMAKE_SOURCE_DIR}/src/py/backend/__init__.py DESTINATION ${DEST_DIR}/lib/onnx_migraphx/.)
+file (COPY ${CMAKE_SOURCE_DIR}/test/py/onnx_backend_test_series.py DESTINATION ${DEST_DIR}/.)
+file (COPY ${CMAKE_SOURCE_DIR}/test/py/testdata DESTINATION ${DEST_DIR}/.)
```
pfultz2, line 204: `configure_file` should probably be used instead so that cmake will re copy this when the file changes.

Although, is it possible to achieve this without copying? What about naming the directory `py/onnx_migraphx`?

CMakeLists.txt
```diff
@@ -191,8 +191,15 @@ rocm_create_package(
...
+
```
pfultz2, line 205: How are these python scripts invoked for the tests?

src/program.cpp
scxiao: The memory_color pass add this parameter, and it is removed later by the dead_code_elimination. Maybe we need to check the flag offload_copy in this pass.

pfultz2: Actually, the `preallocate_param` pass removes this parameter regardless of what `offload_copy` is set to

pfultz2: Perhaps we need to update the `preallocate_param` pass so it is removing the parameter properly.

src/targets/cpu/lowering.cpp
```diff
@@ -275,41 +276,66 @@ struct cpu_deconvolution
...
-                            output(o, in_ch, out_x, out_y) +=
-                                input(o, w, i, j) * weights(w, k, x, y);
-                        }
-                    });
+            std::vector<std::size_t> win_size(wei.begin() + 1, wei.end());
+
+            auto out_lens = output_shape.lens();
+            auto kdims    = op.kdims();
+
+            par_for(output_shape.elements(), [&](auto i) {
```
pfultz2, line 289: This should have the ND deconv changes. This implementation is currently wrong.

CMakeLists.txt
scxiao: ok, will change to that.

scxiao: After trying the above, we get the following error as:
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/_pytest/config.py", line 13, in <module>
    import _pytest._code
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/_pytest/_code/__init__.py", line 3, in <module>
    from .code import Code  # noqa
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/_pytest/_code/code.py", line 11, in <module>
    reprlib = py.builtin._tryimport('repr', 'reprlib')
AttributeError: module 'py' has no attribute 'builtin'

Looks like the pytest module contains something conflict.

CMakeLists.txt
scxiao: I moved the file onnx_backend_test.py to the folder test/py, and added one line to the CMakeList.txt in this folder as:
add_py_test(backend onnx_backend_test WORKING_DIRECTORY ${TEST_ONNX_DIR})
now it can be invoked.

Dockerfile
```diff
@@ -59,7 +59,7 @@ ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
 # Install cget
-RUN pip3 install cget && pip3 install numpy
+RUN pip3 install cget && pip3 install numpy && pip3 install onnx
```
pfultz2: Can you pin the version of onnx?

scxiao: done

src/program.cpp
kahmed10: @scxiao was this resolved?

src/include/migraphx/program.hpp
```diff
@@ -86,10 +86,12 @@ struct program
 
     instruction_ref add_outline(const shape& s);
 
-    instruction_ref add_parameter(std::string name, shape s);
+    instruction_ref add_parameter(std::string name, shape s, bool is_input = true);
```
pfultz2, line 89: Can these changes be removed?

scxiao: We need this to indicate the scratch memory is not a input, so we can get correct input parameter names...

scxiao:
@pfultz2 I tried to use the Dockerfile here to build a docker container in my desktop, but looks like the container does not have the onnx, numpy modules. Only after I manually install onnx and numpy can we import these modules. Do you have any idea for that?

src/include/migraphx/program.hpp
pfultz2: We shouldn't rely on these parameter names as this is fragile. Can't we keep track of the parameter names before we compile?

src/program.cpp
```diff
@@ -319,6 +324,8 @@ shape program::get_parameter_shape(std::string name) const
         return {};
 }
 
+std::vector<std::string> program::get_parameter_names() const { return impl->input_names; }
```
pfultz2, line 327: This can return names to parameters that have been removed. This should iterate over the program and find all the parameters.

src/targets/gpu/lowering.cpp
```diff
@@ -229,6 +229,13 @@ struct miopen_apply
             for(auto& in : inputs)
             {
                 auto p_output = prog->insert_instruction(ret, hip_copy_from_gpu{}, in);
+                // the last copy back instruction needs a synchronization to ensure data copy
+                // completion
+                if(in == inputs.back())
+                {
+                    p_output = prog->insert_instruction(ret, hip_sync_device{}, p_output);
```
pfultz2, line 236: This can be moved out of the loop. And `p_output` doesn't need to be passed in.

pfultz2: Ah maybe `p_output` needs to be passed in so scheduling doesn't move it somewhere else. Of course, the scheduler could reorder these so the last input may not be the last copied. So it should take all copies as inputs.

src/targets/gpu/include/migraphx/gpu/hip.hpp
```diff
@@ -63,7 +63,7 @@ struct hip_sync
         return pack(f(self.tag, "tag"));
     }
 
-    std::string name() const { return "hip::sync"; }
+    std::string name() const { return "hip::sync_device"; }
     shape compute_shape(const std::vector<shape>& inputs) const
     {
         if(inputs.empty())
```
pfultz2, line 69: So this should always return `{}` so dead code elimination wont remove it regardless of inputs.

src/program.cpp
scxiao: We need to return the names of the same order as that in the onnx file. After compilation, we lose track of the name order in the program.

src/targets/gpu/lowering.cpp
scxiao: changed accordingly

src/program.cpp
pfultz2: Yea, there are several ways to do that. You can take the list of parameters found in the program and reverse them:

```cpp
std::vector<std::string> program::get_parameter_names() const 
{
    std::vector<std::string> result;
    for(auto&& ins : impl->instructions)
    {
        if(ins.name() == "@param")
        {
            auto&& name = any_cast<builtin::param>(ins.get_operator()).parameter;
            result.push_back(name);
        }
    }
    std::reverse(result.begin(), result.end();)
    return result;
}
```

Or check if `impl->input_names` is still a parameter:

```cpp
std::vector<std::string> program::get_parameter_names() const 
{
    std::vector<std::string> result = impl->input_names;
    std::unordered_set<std::string> params;
    for(auto&& ins : impl->instructions)
    {
        if(ins.name() == "@param")
        {
            auto&& name = any_cast<builtin::param>(ins.get_operator()).parameter;
            params.push_back(name);
        }
    }
    erase_if(result, [&](auto&& name) {
        return params.count(name) == 0;
    });
    return result;
}
```

Or add a field `order` to the parameter operator and sort them based on this:

```cpp
instruction_ref program::add_parameter(std::string name, shape s)
{
    assert(get_parameter_shape(name) == shape{});
    impl->instructions.push_front({builtin::param{std::move(name), nparam}, std::move(s), {}});
    nparam++;
    return impl->instructions.begin();
}
std::vector<std::string> program::get_parameter_names() const 
{
    std::vector<std::string> result;
    std::vector<builtin::param> params;
    for(auto&& ins : impl->instructions)
    {
        if(ins.name() == "@param")
        {
            auto&& param = any_cast<builtin::param>(ins.get_operator());
            params.push_back(param);
        }
    }
    std::sort(params.begin(), params.end(), by(std::less<>{}, [](auto&& p) { return p.order; }));
    std::transform(params.begin(), params.end(), std::back_inserter(params), [&](auto&& p) {
        return p.parameter;
    });
    return result;
}
```

scxiao: Thanks for the detailed info. I understand what you mean now. I think the second and third approaches work, so I just take the second one. For the first approach, the final name can be either the same or reversed of the order in the onnx file, so just reverse it may not work.

src/targets/gpu/include/migraphx/gpu/hip.hpp
scxiao: Fixed, and add a sync_device pass after the schedule pass to avoid a bunch of other issues.

src/targets/gpu/target.cpp
```diff
@@ -77,6 +78,7 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         dead_code_elimination{},
         write_literals{&ctx},
         schedule{gpu::schedule_model{ctx.get_current_device().nstreams()}, not enabled(MIGRAPHX_DISABLE_SCHEDULE_PASS{})},
+        sync_device{},
```
pfultz2, line 81: This should go after memory coloring.

PR #577, opened 2020-07-15
Add serialization framework

pfultz2:


src/include/migraphx/cloneable.hpp
```diff
@@ -0,0 +1,49 @@
+#ifndef MIGRAPHX_GUARD_RTGLIB_CLONABLE_HPP
+#define MIGRAPHX_GUARD_RTGLIB_CLONABLE_HPP
```
kahmed10, line 2: minor point, but we should probably keep the spelling of cloneable consistent

src/value.cpp
```diff
@@ -0,0 +1,417 @@
...
+const value* value::end() const { return begin() + size(); }
+
+value& value::front() { return *begin(); }
+const value& value::front() const { return *begin(); }
+value& value::back() { return *std::prev(end()); }
+const value& value::back() const { return *std::prev(end()); }
+value& value::at(std::size_t i)
+{
+    auto* a = if_array_impl(x);
+    if(a == nullptr)
```
kahmed10, line 266: for several of these functions, they're (mostly) doing the same error checking and throwing the same message. Is it worthwhile to create simple error functions that can be reused?

src/value.cpp
```diff
@@ -174,7 +175,14 @@ const std::vector<value>& value::get_object() const
     assert(r);
     return *r;
 }
-const std::vector<value>* value::if_object() const { return x ? x->if_array() : nullptr; }
+const std::vector<value>* value::if_object() const 
+{
+    auto* r =  x ? x->if_array() : nullptr;
+    assert(r == nullptr or std::none_of(r->begin(), r->end(), [](auto&& v) {
```
scxiao: This change fixed the missing key problem. But the following code has an assert failure in the get_object() function.

int main(int argc, char **argv) {
    migraphx::value v2 = {1, 2, 3};
    v2.push_back(2);
    std::cout << "v2_size = " << v2.size() << std::endl;
    v2.visit([](auto v) {
        migraphx::value vv = v;
    });

    migraphx::value v3{"abc", {v2, v2}};
    v3.visit([](auto v) {
        migraphx::value v1 = v;
    });
    std::cout << "v3_is_object = " << v3.is_object() << std::endl;
    std::cout << "v3_object = " << v3.get_object() << std::endl;
    std::cout << "v3 = " << v3 << std::endl;

    return 0;
}

pfultz2: The assertion seems correct as `v2` doesn't have a key, but maybe there is still an issue because `is_object` should be false.

pfultz2: It seems like `is_object` returns `false`. In fact this test passes:

```cpp
TEST_CASE(key_array)
{
    migraphx::value v2 = {1, 2, 3};
    migraphx::value v3{"abc", {v2, v2}};
    EXPECT(not v3.is_object());
    EXPECT(v3.is_array());
}
```

To make `v3` an object you need to use an extra set of brackets:

```cpp
migraphx::value v3{{"abc", {v2, v2}}};
```

scxiao: I think in this case v3{{"abc", {v2, v2}}}, v3 is an array of object containing 1 element. Then we do not support v3 = {"abc", {v2, v2}}? In release build (assert is gone), my program works as expected. In debug build, only the assert failed.

scxiao: That is the assert error I got yesterday. But I cannot reproduce it today. 

test_value1: /home/scxiao/Workplace/projects/MIGraph/src/value.cpp:185: const std::vector<value> *migraphx::version_1::value::if_object() const: Assertion `r == nullptr or std::none_of(r->begin(), r->end(), [](auto&& v) { return v.get_key().empty(); })' failed.

pfultz2: > I think in this case v3{{"abc", {v2, v2}}}, v3 is an array of object containing 1 element. Then we do not support v3 = {"abc", {v2, v2}}?

`{"abc", {v2, v2}}` means an array with the key of `abc`. To add the array without the key you would write just `{v2, v2}`. Thats because doing `{string, value}` becomes a value with a key of the string. When a list of values is past to the constructor if it is only 2 elements and the first element is a string then it is treated as a key-value pair instead of an array. If you want to treat it as an array then would need to explicitly initialize it `value::array{string, value}`. This will treat it as an array. You can also do `value::object{{"key1", v1}, {"key2", v2}}` to explicitly create objects.

A value with a `key` is not an object. And object is like an array except each value has corresponding `key`. So to create an object you need multiple key-value pairs. This is similar to how you initialize maps in C++. It is a list of pairs.

pfultz2:
> One note is the representation of the object is a little tricky. Its data is an array, so the input lamda function for the visit() should be different for an general array and an array from the data of an object.

We could pass an optional "tag" to the lambda so you can know which type it is.

scxiao:
> We could pass an optional "tag" to the lambda so you can know which type it is.

Once you make your change, I will change my part accordingly.

PR #583, opened 2020-07-16
Gelu matcher fix

kahmed10:


src/include/migraphx/matcher.hpp
```diff
@@ -567,6 +567,25 @@ inline auto has_value(T x, float tolerance = 1e-6)
     });
 }
 
+template <class T>
+inline auto has_value_multi(T x, float tolerance = 1e-6)
```
pfultz2, line 571: Is a new matcher necessary? Can't we just rename this to `has_value`?

src/targets/gpu/fuse_ops.cpp
```diff
@@ -335,14 +335,18 @@ void move_standard_front(std::vector<instruction_ref>& args)
 struct find_gelu
 {
 
+    static auto sqrt_1_2_val()
+    {
+        return match::any_of(match::has_value(M_SQRT1_2), match::has_value_multi(M_SQRT1_2));
```
pfultz2, line 340: If `has_value_multi` matches then `has_value` would match.

src/include/migraphx/matcher.hpp
kahmed10: I believe so. Initially I checked bert_mrpc, and the way we had written `has_value` was limited to literals of size 1. In the onnx file, it seems they had already broadcasted those values, so the matcher was incorrect.

src/targets/gpu/fuse_ops.cpp
kahmed10: fixed

Issue #575, opened 2020-07-15
bug in rewrite_pooling


scxiao:
Output shape is modified for the attached model
[avepool.zip](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/4922854/avepool.zip)
The shape change is in the pass rewrite_pooling.

PR #584, opened 2020-07-17
Rewrite avepooling bug

scxiao:
related to issue #575


kahmed10:
Maybe add 1 test for a non-2d case.

scxiao:
> Maybe add 1 test for a non-2d case.

Added a bunch of unit tests

PR #588, opened 2020-07-21
Workaround broken rocm packages on newer rocm

pfultz2:
Even though we install rocm packages, the symlink is missing, so this will add it in. This will allow the dockerfile to be built on newer rocm >= 3.1.


kahmed10:
Will approve assuming changes are made to get the CI images to build.

PR #589, opened 2020-07-22
Add solution_id to convolution

kahmed10:


src/targets/gpu/convolution.cpp
```diff
@@ -97,12 +97,44 @@ shape miopen_convolution::compile(context& ctx,
...
+                                                 x_desc.get(),
+                                                 cd.get(),
+                                                 y_desc.get(),
+                                                 solution_count,
+                                                 &solution_count,
+                                                 solutions.data());
+    if(status != miopenStatusSuccess)
+        MIGRAPHX_THROW("MIOpen Convolution: get solution failed");
+
+    solution_id = solutions.front().solution_id;
```
pfultz2, line 135: This should be done in the compile function.

pfultz2:
Also, the finalize method should only call `compile` if the solution_id is `0`.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -17,19 +17,25 @@ struct miopen_convolution
     shared<convolution_descriptor> cd;
     miopenConvFwdAlgorithm_t algo{};
     miopenHandle_t handle = nullptr;
+    uint64_t solution_id  = 0;
```
pfultz2, line 20: We can remove `algo` here since we have the solution_id.

kahmed10: we still use `algo` for checking fusions when it's using Winograd

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -17,19 +17,25 @@ struct miopen_convolution
...
+                    f(self.solution_id, "solution_id"));
     }
 
     std::string name() const { return "gpu::convolution"; }
     shape compute_shape(const std::vector<shape>& inputs) const;
     argument
     compute(context& ctx, const shape& output_shape, const std::vector<argument>& args) const;
-    shape compile(context& ctx, const shape& output_shape, std::vector<shape> inputs);
+    shape find(context& ctx, const shape& output_shape, std::vector<shape> inputs);
+    void get_solution(context& ctx, const shape& output_shape, std::vector<shape> inputs);
```
pfultz2, line 38: This can be removed, We dont define `get_solution` anywhere.

kahmed10: fixed

PR #592, opened 2020-07-29
Convert value to json

scxiao:


src/include/migraphx/serialize.hpp
```diff
@@ -171,6 +171,9 @@ void from_value(const value& v, T& x)
     detail::from_value_impl(rank<7>{}, v, x);
 }
 
+void value_to_json_string(const value& val, std::string& str);
+void value_from_json_string(const std::string& str, value& val);
```
pfultz2, line 175: This should be in a different header. And the function signatures should be:

```cpp
std::string to_json(const value& v);
value from_json(const std::string& str);
```

src/serialize.cpp
```diff
@@ -1,6 +1,7 @@
 #include <migraphx/serialize.hpp>
 #include <migraphx/argument.hpp>
 #include <migraphx/literal.hpp>
+#include <nlohmann/json.hpp>
```
pfultz2, line 4: This should be added to a json.cpp file, not in the same serialize header.

pfultz2:
So with the json library we can add `to_json` and `from_json` functions to handle conversions of custom types. It describes that [here](https://github.com/nlohmann/json#basic-usage). We can add something like these function to the .cpp file to do the conversion:

```cpp
void to_json(json& j, const value& v);
void from_json(const json& j, value& v);
```

And then we should be able to implictly convert back and forth from value and json.

src/serialize.cpp
```diff
@@ -31,5 +32,163 @@ void migraphx_from_value(const value& v, argument& a)
...
+        CASE_TYPE(number_float, double)
+        CASE_TYPE(number_integer, int64_t)
+        CASE_TYPE(number_unsigned, uint64_t)
+        CASE_TYPE(string, std::string)
+#undef CASE_TYPE
+
+    case json::value_t::array:
+    case json::value_t::object:
+    case json::value_t::binary:
+    case json::value_t::discarded: return false;
```
pfultz2, line 133: Shouldn't this throw an error instead?

src/serialize.cpp
```diff
@@ -31,5 +32,163 @@ void migraphx_from_value(const value& v, argument& a)
...
+    switch(type)
+    {
+    case json::value_t::null: val = migraphx::value(nullptr); break;
+
+    case json::value_t::object:
+        for(auto item : j.items())
+        {
+            auto key = item.key();
+            json v   = item.value();
+            if(!to_value(v, key, val))
```
pfultz2, line 151: Why does `key` need to be passed? You already use the `key` below.

src/serialize.cpp
```diff
@@ -31,5 +32,163 @@ void migraphx_from_value(const value& v, argument& a)
...
+        }
+        break;
+
+    case json::value_t::array:
+        for(auto& v : j)
+        {
+            if(!to_value(v, val))
+            {
+                migraphx::value mv;
+                value_from_json(v, mv);
```
pfultz2, line 166: Why is `value_from_json` called? `to_value` should already convert the json to a value?

src/serialize.cpp
```diff
@@ -31,5 +32,163 @@ void migraphx_from_value(const value& v, argument& a)
...
+
+bool to_value(const json& j, migraphx::value& val)
+{
+    auto type = j.type();
+    switch(type)
+    {
+    case json::value_t::null: val.push_back(migraphx::value(nullptr)); break;
+
+#define CASE_TYPE(vt, cpp_type) \
+    case json::value_t::vt: val.push_back(j.get<cpp_type>()); break;
```
pfultz2, line 121: Why is it calling `push_back`? It should be `val = j.get<cpp_type>()`.

src/serialize.cpp
```diff
@@ -31,5 +32,163 @@ void migraphx_from_value(const value& v, argument& a)
...
+
+template <class T, class U>
+void value_to_json(const std::pair<T, U>& x, json& j)
+{
+    json jj;
+    value_to_json(x.second, jj);
+    j[x.first] = jj;
+}
+
+void value_to_json(std::nullptr_t&, json& j) { j = {}; }
```
pfultz2, line 71: Wont the overload on line 40 take care of this case?

src/serialize.cpp
pfultz2: And why doesn't this do `val[key] = value_from_json(v)`? Why is this second `to_value` function even needed?

src/serialize.cpp
```diff
@@ -31,5 +32,163 @@ void migraphx_from_value(const value& v, argument& a)
...
+            {
+                migraphx::value mv;
+                value_from_json(v, mv);
+                val.push_back(mv);
+            }
+        }
+        break;
+
+#define CASE_TYPE(vt, cpp_type) \
+    case json::value_t::vt: val = j.get<cpp_type>(); break;
```
pfultz2, line 173: Can you just inline this instead of using a macro?

src/include/migraphx/serialize.hpp
scxiao: fixed

src/serialize.cpp
scxiao: fixed

scxiao:
> And then we should be able to implictly convert back and forth from value and json.

We have the function value_to_json(const value& val, std::string& str) and value_from_json(const std::string& str, value& val)

pfultz2:
> We have the function value_to_json(const value& val, std::string& str) and value_from_json(const std::string& str, value& val)

Yes but the json library wont call those functions. The point is to have the json library call these functions to do the conversion, so then our functions could be implemented like this:

```cpp
std::string to_json_string(const value& v)
{
    json j = v;
    return j.dump();
}
value from_json_string(const std::string& str)
{
    json j = json::parse(str);
    return j.get<value>();
}
```

scxiao:
> std::string to_json_string(const value& v)
> {
>     json j = v;
>     return j.dump();
> }
> value from_json_string(const std::string& str)
> {
>     json j = json::parse(str);

to Convert json to a value, I got the following build error:
constexpr variable 'value' must be initialized by a constant expression

Do you know the reason for that?

The code is:
```
void from_json(const json& j, value& val)
{
    val = value_from_json(j);
}

value from_json_string(const std::string& str)
{
    json j = json::parse(str);
    // value val = value_from_json(j);
    // return val;

   return j.get<migraphx::value>();
}
```

pfultz2:
> to Convert json to a value, I got the following build error:
> constexpr variable 'value' must be initialized by a constant expression

Do you have the full build error?

scxiao:
> > to Convert json to a value, I got the following build error:
> > constexpr variable 'value' must be initialized by a constant expression
> 
> Do you have the full build error?

Please check the attached file 
[build_error.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/5011568/build_error.txt)

I created another branch test_json that can reproduce the build error.

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ endif()
 
 project(migraphx)
 find_package(ROCM REQUIRED)
+find_package(nlohmann_json 3.8.0 REQUIRED)
```
pfultz2, line 22: I am not sure if its best to put the version here, I believe there is some backwards compatibility, but I am not sure.

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+void value_to_json(const std::vector<value>& x, json& j)
+{
+    for(auto& v : x)
+    {
+        if(v.get_key().empty())
+        {
+            json jj;
+            value_to_json(v, jj);
+            j.push_back(jj);
```
pfultz2, line 28: I believe this can be written as `j.push_back(v)`. You just need to forward declare the `to_json` and `from_json` functions at the top.

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+        }
+        // in the case of an object
+        else
+        {
+            value_to_json(v, j);
```
pfultz2, line 33: I think this can be written as `j[v.get_key()] = v.without_key()`.

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void value_to_json(std::nullptr_t&, json& j) { j = {}; }
+
+void value_to_json(const value& val, json& j)
+{
+    val.visit([&](auto v) { value_to_json(v, j); });
+}
+
+void to_json(json& j, const value& val) { value_to_json(val, j); }
```
pfultz2, line 53: Can this just be written as:

```cpp
void to_json(json& j, const value& val)
{
   val.visit([&](auto v) { value_to_json(v, j); });
}
```

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+std::string to_json_string(const value& val)
```
pfultz2, line 55: I think it would be best to move this to the bottom with `from_json_string`.

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    case json::value_t::number_integer: val = j.get<int64_t>(); break;
+
+    case json::value_t::number_unsigned: val = j.get<uint64_t>(); break;
+
+    case json::value_t::string: val = j.get<std::string>(); break;
+
+    case json::value_t::array:
+        std::transform(j.begin(), j.end(), std::back_inserter(val), [&](const auto& v) {
+            return value_from_json(v);
```
pfultz2, line 82: I think this can be written as `return v.get<value>()`. Also, I think it would be better to use the variable `x` or `jj`, to make it a littler clearer its `json` object and not `value`.

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+        });
+        break;
+
+    case json::value_t::object:
+        for(const auto& item : j.items())
+        {
+            const auto& key = item.key();
+            const json& v   = item.value();
+            val[key]        = value_from_json(v);
```
pfultz2, line 91: I think this can be written as `val[item.key()] = item.value().get<value>()`.

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+        }
+        break;
+
+    case json::value_t::binary:
+    case json::value_t::discarded: MIGRAPHX_THROW("Convert JSON to Value: type not supported!");
```
pfultz2, line 96: We should probably put the type in the error message:

```cpp
    case json::value_t::binary: MIGRAPHX_THROW("Convert JSON to Value: binary type not supported!");
    case json::value_t::discarded: MIGRAPHX_THROW("Convert JSON to Value: discarded type not supported!");
```

pfultz2:
Some additional tests: array of objects, an object of arrays, empty object, and empty array. The empty types may have some ambiguity to it, because when we serialize back from a `value` we dont know if its an object or an array. Ideally, we need to update the `visit` method to pass a tag.

src/json.cpp
```diff
@@ -0,0 +1,114 @@
...
+    return j.dump();
+}
+
+migraphx::value value_from_json(const json& j)
+{
+    migraphx::value val;
+    json::value_t type = j.type();
+    switch(type)
+    {
+    case json::value_t::null: val = migraphx::value(); break;
```
pfultz2, line 68: This is a nice workaround, but in the msgpack PR, I fix assignment from null so `val = nullptr` should work after merging the msgpack PR.

src/json.cpp
scxiao: fixed

src/json.cpp
scxiao: fixed

src/json.cpp
scxiao: fixed

src/json.cpp
scxiao: the v.get<value>() has a compilation error as I mentioned before.

src/json.cpp
scxiao: same build error as above

src/json.cpp
pfultz2: You need to put the `template` keyword there, or change the lambda from `const auto& v` to `const json& v`.

scxiao: Tried, looks like it does not work. I read more materials, from https://github.com/nlohmann/json#implicit-conversions, It is recommended to NOT USE implicit conversions FROM a JSON value, particularly for complex data types. Maybe the json implementation has some limitations for that?

src/json.cpp
scxiao: fixed

test/json_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+    migraphx::value ev   = {{"a", 1.2}, {"b", true}};
+    EXPECT(v == ev);
+}
+
+TEST_CASE(empty_object)
+{
+    migraphx::value v;
+    migraphx::value v1   = {{"a", v}};
+    std::string json_str = migraphx::to_json_string(v1);
+    EXPECT(json_str == "{\"a\":null}");
```
pfultz2, line 123: Thats not an empty object, I mean to have test like these:

```cpp
EXPECT(migraphx::from_json_string("{}") == migraphx::value::object{});
EXPECT(migraphx::from_json_string("[]") == migraphx::value::array{});
EXPECT(migraphx::to_json_string(migraphx::value::array{}) == "[]")
// Empty object becomes an array on the way back due to ambiguity of the visit functon
EXPECT(migraphx::to_json_string(migraphx::value::object{}) == "[]")
```

scxiao: added corresponding unit tests

src/json.cpp
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+}
+
+template <class T, class U>
+void value_to_json(const std::pair<T, U>& x, json& j)
+{
+    json jj;
+    value_to_json(x.second, jj);
+    j[x.first] = jj;
```
pfultz2, line 62: This code should never be called(and there is no testing coverage for this). The overload is still needed for the visit function, but you should just do `assert(false)` to check in debug builds that it is never called.

test/json_test.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+TEST_CASE(empty_value_rev)
+{
+    std::string json_str = "null";
+    migraphx::value v    = migraphx::from_json_string(json_str);
+    migraphx::value ev   = migraphx::value();
+    EXPECT(v == ev);
+}
+
+TEST_CASE(empty_array)
```
pfultz2, line 24: This test case should probably be called `null_array` since its an array of nulls and not an empty array.

test/json_test.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+TEST_CASE(object_value_rev)
+{
+    std::string json_str = R"({"a":1.2,"b":true})";
+    migraphx::value v    = migraphx::from_json_string(json_str);
+    migraphx::value ev   = {{"a", 1.2}, {"b", true}};
+    EXPECT(v == ev);
+}
+
+TEST_CASE(empty_object)
```
pfultz2, line 132: This test case should probably be called `null_object` and not empty, since it isn't empty.

src/json.cpp
scxiao: I just removed this function

PR #596, opened 2020-07-31
Add msgpack serialization

pfultz2:


pfultz2:
The code coverages fails because of the `debug_print`. I dont think a test case is really needed for that.

src/msgpack.cpp
```diff
@@ -0,0 +1,155 @@
...
+            {
+                v = o.as<double>();
+                break;
+            }
+            case msgpack::type::STR:
+            {
+                v = o.as<std::string>();
+                break;
+            }
+            case msgpack::type::BIN: { throw msgpack::type_error();
```
scxiao, line 48: Which error message is printed here? can we use the MIGRAPHX_THROW instead?

src/msgpack.cpp
```diff
@@ -0,0 +1,155 @@
...
+                migraphx::value r = migraphx::value::object{};
+                std::for_each(o.via.map.ptr,
+                              o.via.map.ptr + o.via.map.size,
+                              [&](const msgpack::object_kv& p) {
+                                  r[p.key.as<std::string>()] = p.val.as<migraphx::value>();
+                              });
+                v = r;
+                break;
+            }
+            case msgpack::type::EXT: { throw msgpack::type_error();
```
scxiao, line 71: Same as above

src/msgpack.cpp
```diff
@@ -0,0 +1,155 @@
...
+                migraphx::value r = migraphx::value::object{};
+                std::for_each(o.via.map.ptr,
+                              o.via.map.ptr + o.via.map.size,
+                              [&](const msgpack::object_kv& p) {
+                                  r[p.key.as<std::string>()] = p.val.as<migraphx::value>();
+                              });
+                v = r;
+                break;
+            }
+            case msgpack::type::EXT: { MIGRAPHX_THROW("msgpack BIN type not supported.");
```
scxiao, line 71: I think the msssage should be "msgpack EXT type not support", not BIN type

pfultz2: Ah good catch!

PR #599, opened 2020-08-07
Layernorm onnx support

kahmed10:
Works with pytorch exports of bert


src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp
```diff
@@ -242,6 +242,23 @@ constexpr index_int compute_block_size(index_int n, index_int max_block_size)
     return block_size;
 }
 
+inline std::vector<index_int> get_reduce_lens(const std::vector<size_t>& input_lens,
```
kahmed10, line 245: This file was refactored from previous changes, but it does not directly affect layernorm at the moment.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -737,6 +788,7 @@ struct find_conv_bias_relu
 
 void fuse_ops::apply(program& p) const
 {
+    match::find_matches(p, find_layernorm{});
```
pfultz2, line 791: Can this be put in the `find_matches` call below?

kahmed10: moved

src/targets/gpu/device/layernorm.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+// m = x - mean(x)
+// m / sqrt(mean(m ^ 2) + 1e-12)
+void layernorm(hipStream_t stream, const argument& result, const argument& arg1)
+{
+    auto relements    = arg1.get_shape().lens().back();
```
scxiao, line 15: we need to add an assert to ensure relements <= 1024, here

src/targets/gpu/device/layernorm.cpp
```diff
@@ -0,0 +1,64 @@
...
+        using value_type = typename decltype(input)::value_type;
+
+        const std::size_t max_block_size = 256;
+        const std::size_t block_size     = compute_block_size(relements, max_block_size);
+        const std::size_t block_size_div = encode_divisor(block_size);
+
+        gs_launch(stream, nelements * block_size, block_size)([=](auto i, auto idx) __device__ {
+            const auto out_idx  = i / block_size;
+            const auto base_idx = out_idx * relements;
+            value_type x_data[4];
```
scxiao, line 34: Should we initialize x_data to 0s?

pfultz2: There may be a performance cost initializing it.

src/targets/gpu/device/layernorm.cpp
kahmed10: fixed

src/targets/gpu/device/layernorm.cpp
kahmed10: I will leave it as is for now, unless there are some correctness issues in the future

PR #602, opened 2020-08-11
Move alloc copy to finalize

scxiao:


src/targets/gpu/include/migraphx/gpu/hip.hpp
```diff
@@ -151,7 +152,9 @@ struct hip_copy
     std::ptrdiff_t output_alias(const std::vector<shape>&) const { return 1; }
 };
 
-struct hip_load_memory
+void store_preallocated_param(context& ctx, const std::string& id, const argument& a);
```
pfultz2, line 155: Whats the reason to have this in a seperate function? Is the `context` an incomplete type?

scxiao: Yes, context in the header file is an incomplete type. A few other functions like the copy_to_gpu and copy_from_gpu are the same.

scxiao: @pfultz2 the write_literals pass has an if branch like the following:
            if(enabled(MIGRAPHX_COPY_LITERALS{}))
            {
                literal l  = ins->get_literal();
                auto pre   = p.add_literal(l);
                auto alloc = p.insert_instruction(std::next(pre), hip_allocate{l.get_shape()});
                p.replace_instruction(ins, hip_copy_to_gpu{}, pre, alloc);
            }
            else

Should we also change this part?

pfultz2: > Should we also change this part?

That shouldn't need to be changed.

PR #603, opened 2020-08-13
Bool type and equal operator

scxiao:


src/include/migraphx/generate.hpp
```diff
@@ -37,6 +37,12 @@ constexpr T normalize(unsigned long z)
     return z % max;
 }
 
+template <>
```
pfultz2, line 40: This should probably be `template<class T, MIGRAPHX_REQUIRES(std::is_same<T, bool>{})>`.

src/targets/gpu/device/equal.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+void equal(hipStream_t stream, const argument& result, const argument& arg1, const argument& arg2)
+{
+    result.visit([&](auto output) {
+        visit_all(arg1, arg2)([&](auto in1, auto in2) {
+            const auto* in1_ptr = device_cast(in1.data());
+            const auto* in2_ptr = device_cast(in2.data());
+            auto* out_ptr       = device_cast(output.data());
+            gs_launch(stream, result.get_shape().elements())(
+                [=](auto i) __device__ { out_ptr[i] = (in1_ptr[i] == in2_ptr[i]); });
```
pfultz2, line 17: This only works when the data is in standard shape or they are all packed and the same shape.

test/include/test.hpp
```diff
@@ -85,6 +85,18 @@ inline std::ostream& operator<<(std::ostream& s, const std::vector<T>& v)
     return s;
 }
 
+template <>
```
pfultz2, line 88: I think the `template <>` is not necessary.

src/include/migraphx/op/equal.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <migraphx/operation.hpp>
+#include <migraphx/check_shapes.hpp>
+#include <migraphx/float_equal.hpp>
+#include <migraphx/config.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct equal : binary<equal>
```
pfultz2, line 14: This needs to be added to the list of register ops.

src/targets/gpu/include/migraphx/gpu/equal.hpp
```diff
@@ -0,0 +1,19 @@
...
+#define MIGRAPHX_GUARD_RTGLIB_EQUAL_HPP
+
+#include <migraphx/gpu/oper.hpp>
+#include <migraphx/gpu/device/equal.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+
+struct hip_equal : binary_device<hip_equal, device::equal>
```
pfultz2, line 11: This needs to be added to the list of register ops.

src/onnx/onnx.cpp
```diff
@@ -2624,12 +2636,13 @@ struct onnx_parser
         case onnx::TensorProto::UINT32: shape_type = shape::uint32_type; break;
         case onnx::TensorProto::UINT64: shape_type = shape::uint64_type; break;
         case onnx::TensorProto::UINT8: shape_type = shape::uint8_type; break;
+        case onnx::TensorProto::BOOL: shape_type = shape::bool_type; break;
         case onnx::TensorProto::STRING:
-        case onnx::TensorProto::BOOL:
         case onnx::TensorProto::UNDEFINED:
         case onnx::TensorProto::COMPLEX64:
         case onnx::TensorProto::COMPLEX128:
-            break; // throw std::runtime_error("Unsupported type");
```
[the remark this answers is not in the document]
pfultz2: I dont remember.

PR #604, opened 2020-08-13
Register all operators in migraphx

pfultz2:


mvermeulen:
@pfultz2 what goes on with deepcode failure?  Is this something gating my merging it in?

pfultz2:
> what goes on with deepcode failure? Is this something gating my merging it in?

deeepcode is not a required status check, so you should be able to merge it in. The word `Required` is next to the checks that are required to pass, which is missing for deepcode and codecov.

PR #605, opened 2020-08-13
Add to_value/from_value to operation class

pfultz2:


kahmed10:
Was there a way to write an additional test to improve code coverage?

pfultz2:
I added a couple of tests.

PR #606, opened 2020-08-17
move init miopen fusion operator to finalize method

scxiao:


src/targets/gpu/fuse_ops.cpp
```diff
@@ -45,8 +45,9 @@ struct fusion
         return result;
     }
 
+    fusion() {}
```
pfultz2, line 48: Since we are adding a default constructor, we might need to add a `assert(fp)` at the beginning of each method.

scxiao: fixed.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -565,18 +565,21 @@ struct miopen_conv_bias
     template <class Self, class F>
     static auto reflect(Self& self, F f)
     {
+        assert(f.fp);
```
pfultz2, line 568: The asserts should be in the methods to the `fusion` class not in the operators class.

PR #607, opened 2020-08-19
Context serialization

scxiao:


tools/include/context.hpp
```diff
@@ -25,12 +26,32 @@ struct context
...
 <%
-interface('context',
-    virtual('finish', returns='void', const=True)
-)
+ interface('context',
+           virtual('finish', returns = 'void', const = True))
 %>
 
+// virtual('to_value', returns = 'value', const = True, default = 'to_value_context'),
+// virtual('from_value', v = 'const value&', default = 'from_value_context')
```
pfultz2, line 47: These lines should be added. Why are they commented out?

tools/include/context.hpp
```diff
@@ -25,12 +26,32 @@ struct context
 
 #else
 
+template <class T>
+value to_value_context(const T& x)
+{
+    return migraphx::to_value(x);
```
pfultz2, line 32: The fallback should be `return migraphx::value{}`.

tools/include/context.hpp
```diff
@@ -25,12 +26,32 @@ struct context
...
+}
+
+template <class T>
+void from_value_context(T& x, const value& v)
+{
+    return migraphx::from_value(v, x);
```
pfultz2, line 38: This should be empty.

tools/include/context.hpp
```diff
@@ -8,6 +8,7 @@
 #include <type_traits>
 #include <utility>
 #include <migraphx/config.hpp>
+#include <migraphx/serialize.hpp>
```
pfultz2, line 11: This should be changed to only include `value.hpp` and not `serialize.hpp`.

src/include/migraphx/context.hpp
```diff
@@ -117,9 +143,39 @@ struct context
...
 
+    template <class T>
+    static auto private_detail_te_default_to_value(char, T&& private_detail_te_self)
+        -> decltype(private_detail_te_self.to_value())
+    {
+        return private_detail_te_self.to_value();
+    }
+
+    template <class T>
+    static value private_detail_te_default_to_value(float, T&& private_detail_te_self)
```
scxiao, line 159: @pfultz2 How this function can be called?

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -168,11 +169,40 @@ struct context
...
+
+        return result;
+    }
+
+    void from_value(const value& v)
+    {
+        assert(v.contains("events"));
+        auto v_events = v.at("events");
+        assert(v_events.is_uint64());
+        std::size_t n_events = v_events.get_uint64();
```
pfultz2, line 186: Use `v_events.to<std::size_t>()` instead.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -168,11 +169,40 @@ struct context
...
+        this->create_events(n_events - 1);
+
+        assert(v.contains("streams"));
+        auto v_streams = v.at("streams");
+        assert(v_streams.is_uint64());
+        std::size_t n_streams = v_streams.get_uint64();
```
pfultz2, line 192: Same here.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -168,11 +169,40 @@ struct context
...
+
+    void from_value(const value& v)
+    {
+        assert(v.contains("events"));
+        auto v_events = v.at("events");
+        assert(v_events.is_uint64());
+        std::size_t n_events = v_events.get_uint64();
+        this->create_events(n_events - 1);
+
+        assert(v.contains("streams"));
```
pfultz2, line 189: There is no point in asserting this since `.at` will throw an exception if the key is missing.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -168,11 +169,40 @@ struct context
...
+        result["streams"] = current_device->nstreams();
+
+        return result;
+    }
+
+    void from_value(const value& v)
+    {
+        assert(v.contains("events"));
+        auto v_events = v.at("events");
+        assert(v_events.is_uint64());
```
pfultz2, line 185: Also this `assert` isn't needed when using `.to` as it will throw an exception if it can't be converted.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -180,16 +180,12 @@ struct context
 
     void from_value(const value& v)
     {
-        assert(v.contains("events"));
-        auto v_events = v.at("events");
-        assert(v_events.is_uint64());
-        std::size_t n_events = v_events.get_uint64();
+        auto v_events        = v.at("events");
+        std::size_t n_events = v_events.to<std::pair<std::string, std::size_t>>().second;
```
pfultz2, line 184: Actually, you can write `v_events.without_key().to<std::size_t>()`. In PR #616, I updated `value` class so the `without_key()` is not necessary.

PR #609, opened 2020-08-19
Dont print literal data in hip_copy_literal

pfultz2:


mvermeulen:
Kicking off a build and weekly perf run using this branch. Hopefully verifies the PR while also getting a weekly perf outputs more easily comparable with last week.

mvermeulen:
I can confirm that these changes no longer result in large output.  In process of doing complete performance run, but this is fixed and will merge when it passes.

PR #613, opened 2020-08-21
Improve layernorm performance

pfultz2:
This does a vectorized layernorm which seems to avoid the performance issue with using an array. This runs over 10x faster. It was running at 12ms on machine, and now it runs in under 1ms. The matcher is also updated so it won't fuse if its too large. 

It also adds `--fill1` to the verify driver so I can verify this change with bert(which it passes).

### DIFF
--- a/src/targets/gpu/fuse_ops.cpp
+++ b/src/targets/gpu/fuse_ops.cpp
@@ -234,8 +234,7 @@ MIGRAPHX_REGISTER_OP(hip_add_tanh)
 struct hip_layernorm : unary_device<hip_layernorm, &device::layernorm>
 {
     // Empty finalize to skip dimension reduction
-    void finalize(context&, const shape&, const std::vector<shape>&)
-    {}
+    void finalize(context&, const shape&, const std::vector<shape>&) {}
 };
 MIGRAPHX_REGISTER_OP(hip_layernorm)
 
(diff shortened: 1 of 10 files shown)


src/targets/gpu/fuse_ops.cpp
```diff
@@ -326,6 +328,23 @@ struct find_layernorm
         auto x_ins = r.instructions["x"];
         auto args  = ins->inputs();
 
+        // We dont fuse for non-standard layouts
+        if(not x_ins->get_shape().standard())
+            return;
+
+        auto relements = x_ins->get_shape().lens().back();
+
+        if(relements % 4 == 0)
```
kahmed10, line 337: this can be simplified to `if(relements > 1024 or (relements % 4 !=0 and relements > 256)) return;`

test/gpu/ops_test.cpp
```diff
@@ -1034,26 +1034,63 @@ struct test_layernorm : verify_program<test_layernorm>
     {
         migraphx::program p;
         std::vector<size_t> dims{1, 1, 5};
-        auto x        = p.add_parameter("x", migraphx::shape{migraphx::shape::float_type, dims});
-        auto scale    = p.add_parameter("scale", migraphx::shape{migraphx::shape::float_type, {5}});
-        auto bias     = p.add_parameter("bias", migraphx::shape{migraphx::shape::float_type, {5}});
+        auto x = p.add_parameter("x", migraphx::shape{migraphx::shape::float_type, dims});
```
kahmed10, line 1037: The only difference between this and the other test is `dims`. Can we write a function that adds the instructions for layernorm and takes `dims` as a parameter?

PR #615, opened 2020-08-21
Pooling ceil mode

scxiao:


src/include/migraphx/op/pooling.hpp
```diff
@@ -24,6 +24,7 @@ struct pooling
     std::vector<std::size_t> stride  = {1, 1};
     std::vector<std::size_t> lengths = {1, 1};
     padding_mode_t padding_mode      = default_;
+    int ceil_mode                    = 0;
```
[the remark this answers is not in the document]
pfultz2: > The padding mode is used in the tensorflow parsing..., so cannot be removed.

No, it is not used. Tensorflow sets it, but the operator never uses it.

src/targets/cpu/lowering.cpp
```diff
@@ -421,7 +421,11 @@ MIGRAPHX_REGISTER_OP(cpu_im2col)
 struct max_pool
 {
     static std::string name() { return "max"; }
-    static double start() { return std::numeric_limits<double>::lowest(); }
+    template <class T>
+    static double start()
```
pfultz2, line 425: This should return `T`.

src/targets/cpu/lowering.cpp
```diff
@@ -501,7 +510,7 @@ struct cpu_pooling : auto_register_op<cpu_pooling<Op>>
                     }
                 });
 
-                output[i] = type(Op::final(acc, pool_size));
+                output[i] = (pool_size == 0) ? type(acc) : type(Op::final(acc, pool_size));
```
pfultz2, line 513: I dont see why this is needed.

src/include/migraphx/op/pooling.hpp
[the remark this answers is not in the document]
scxiao: OK, I will change remove corresponding code in the tf.cpp to try.

src/targets/cpu/lowering.cpp
scxiao: In ceil_mode = true, case, some output elements corresponding to 0 input element. This is to avoid divided by 0 in averagePool

src/targets/cpu/lowering.cpp
```diff
@@ -421,25 +421,34 @@ MIGRAPHX_REGISTER_OP(cpu_im2col)
...
+    template <class T>
+    static double start()
+    {
+        return 0.0;
+    }
 
     static double apply(double x, double y) { return x + y; }
 
-    static double final(double x, double y) { return x / y; }
+    static double final(double x, std::size_t y) { return (y == 0) ? x : (x / y); }
```
pfultz2, line 451: Should this be `0` when `y == 0`? Or does this not matter?

src/onnx/onnx.cpp
```diff
@@ -701,14 +700,18 @@ struct onnx_parser
...
             cal_auto_padding_size(info,
                                   values,
                                   k_lens,
                                   values["dilation"].to_vector<std::size_t>(),
                                   in_lens,
                                   padding);
+            auto auto_pad = info.attributes["auto_pad"].s();
+            if(auto_pad.find("SAME") != std::string::npos)
+            {
+                values["padding_mode"] = to_value(op::padding_mode_t::same);
```
pfultz2, line 713: If you removed the `padding_mode` from pooling then this can be removed.

src/include/migraphx/op/pooling.hpp
```diff
@@ -64,12 +64,13 @@ struct pooling
         for(size_t i = 0; i < kdims; i++)
         {
             assert(lengths[i] <= input_lens[i + 2] + 2 * padding[i]);
+            std::size_t len = (ceil_mode)
+                                  ? ceil_divide<std::ptrdiff_t>(
+                                        input_lens[i + 2] + 2 * padding[i] - lengths[i], stride[i])
```
pfultz2, line 69: Can you assign this to a local variable and then pass it to the function? This will avoid repeating this argument twice.

src/onnx/onnx.cpp
scxiao: I removed the padding_mode from pooling. Here is for convolution.

src/targets/cpu/lowering.cpp
scxiao: Yes, when y == 0, x is also 0.

scxiao: so, either way is fine.

src/include/migraphx/op/pooling.hpp
scxiao: fixed

src/targets/cpu/lowering.cpp
scxiao: fixed

src/targets/cpu/lowering.cpp
scxiao: changed to return 0

PR #616, opened 2020-08-22
Add C++ ability to construct operators by name

pfultz2:
This also refactors the onnx parser and gpu lowering to use these functions so that we can improve our compile-time.


src/onnx/onnx.cpp
```diff
@@ -821,22 +888,27 @@ struct onnx_parser
...
         }
     }
 
     instruction_ref
     parse_pooling(const std::string& name, node_info info, std::vector<instruction_ref> args)
     {
-        op::pooling op{ends_with(name, "MaxPool") ? "max" : "average"};
+        std::string mode = ends_with(name, "MaxPool") ? "max" : "average";
+        operation op     = make_op("pooling", {{"mode", mode}});
+        // operation op = op::pooling{ends_with(name, "MaxPool") ? "max" : "average"};
```
kahmed10, line 901: Extra commented code here

src/targets/gpu/lowering.cpp
```diff
@@ -379,28 +336,26 @@ struct miopen_apply
         });
     }
 
-    template <class T>
-    void add_generic_op(std::string name)
+    void add_generic_op(const std::string& op_name, const std::string& gpu_name)
```
kahmed10, line 339: Rather than pass in the `gpu_name` to the function, can we just prepend `"gpu::"` to `op_name` in L346?

src/targets/gpu/lowering.cpp
```diff
@@ -379,28 +336,26 @@ struct miopen_apply
...
             refs.push_back(output);
 
-            return prog->replace_instruction(ins, T{}, refs);
+            return prog->replace_instruction(ins, make_op(gpu_name), refs);
         });
     }
 
-    template <class T, class Op>
-    void add_extend_op(std::string name)
+    void add_extend_op(const std::string& op_name, const std::string& gpu_name)
```
kahmed10, line 350: Same suggestion as above

src/targets/gpu/lowering.cpp
pfultz2: Well, I dont know if all operators in the future will follow that pattern. We could probably refactor this in the future if we think we wont break this convention.

kahmed10: I think it would be easier to change it now since I can't really think of any operator where it would differ. If in the future we need to have a different name, it might be easier to just have the `gpu_name` be an optional parameter.

Issue #581, opened 2020-07-16
python api cannot support int64 datatype


scxiao:
In the onnx backend tests, the following cases reported an error of int64 datatype not suppoted
test_operator_non_float_params_cpu
test_Embedding_sparse_cpu
test_gather_negative_indices_cpu

PR #618, opened 2020-08-24
Onnx datatype parsing

scxiao:
This is for issue #581 (Python API does not support int64 data type)


scxiao:
After checking the codecov report, looks like the only missing code are those for throwing exceptions.

PR #627, opened 2020-09-03
Some perf improvements to bert

pfultz2:
So this does two things:

* For commutative operators, its best to have the broadcasted type last
* Move gemm fusion into fuse ops so it can happen after other fusions are applied. So the add_gelu fusion can be used when possible.


kahmed10 (approved):

scxiao (approved):

scxiao:
I think this PR caused a few onnx unit tests failed, may need some check

src/targets/gpu/fuse_ops.cpp
```diff
@@ -743,6 +743,61 @@ struct find_conv_bias_relu
...
+    void apply(program& p, match::matcher_result r) const
+    {
+        auto ins      = r.result;
+        auto gemm_ins = r.instructions["gemm"];
+        auto c_ins    = r.instructions["c"];
+
+        auto gemm = any_cast<rocblas_gemm<op::dot>>(gemm_ins->get_operator());
+
+        // Already fused gemm
+        if(gemm.op.beta != 0)
```
kahmed10, line 763: this comparison is giving a compiler warning

PR #628, opened 2020-09-03
Dockerfile download onnx models to enable real model unit tests

scxiao:


download_models.sh
```diff
@@ -0,0 +1,13 @@
...
+tmp_dir=$ONNX_HOME/tmp/
+mkdir -p $model_dir
+mkdir -p $tmp_dir
+for name in bvlc_alexnet densenet121 inception_v1 inception_v2 \
+        resnet50 shufflenet squeezenet vgg19 zfnet512 
+do
+curl https://s3.amazonaws.com/download.onnx/models/opset_9/$name.tar.gz --output $tmp_dir/$name.tar.gz
+tar -xzvf $tmp_dir/$name.tar.gz --directory $model_dir && rm $tmp_dir/$name.tar.gz
+done
+
```
pfultz2, line 13: Can this script go under `tools/`?

scxiao: Yes, we can. I will do that.

test/py/onnx_backend_test.py
```diff
@@ -262,18 +262,20 @@ def create_backend_test(testname=None, target_device=None):
         backend_test.exclude(r'test_expand_shape_model3_cpu')
         backend_test.exclude(r'test_expand_shape_model4_cpu')
 
-        backend_test.exclude(r'test_bvlc_alexnet_cpu')
-        backend_test.exclude(r'test_densenet121_cpu')
+        #        backend_test.exclude(r'test_bvlc_alexnet_cpu')
+        #        backend_test.exclude(r'test_densenet121_cpu')
```
pfultz2, line 266: Can we remove these lines?

tools/download_models.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+model_dir=$ONNX_HOME/.onnx/models
```
pfultz2, line 3: How is `ONNX_HOME` set? I dont see it set in the Dockerfile. Also is it possible to set this to a default(ie `ONNX_HOME=$HOME`) when it wasn't set by the user?

scxiao: The env var ONNX_HOME is used in the onnx unit test. It indicates where the test models and data are stored. I think we can set to to whatever value we want. I will change to set it to home.

test/py/onnx_backend_test.py
scxiao: Removed.

tools/download_models.sh
pfultz2: But this script is called from the Dockerfile not onnx test suite. Can the script fallback to `$HOME` if `ONNX_HOME` is not defined?

Dockerfile
```diff
@@ -92,6 +92,11 @@ RUN cget -p $PREFIX install -f /dev-requirements.txt -DMIOPEN_CACHE_DIR=""
 
 RUN pip3 install onnx==1.7.0 numpy==1.18.5 typing==3.7.4 pytest==6.0.1
 
+# Download real models to run onnx unit tests
+ENV ONNX_HOME=$HOME
```
pfultz2, line 96: Is it possible to set this in the python script or cmake script(I think you can set the `ENVIRONMENT` test property)?

src/py/backend/backend.py
```diff
@@ -84,10 +90,19 @@ def prepare(cls, model, device=None, **kwargs):
                 raise RuntimeError(
                     "Incompatible device expected '{0}', got '{1}'".format(
                         device, get_device()))
+            old_stdout = sys.stdout
+            sys.stdout = mystdout = StringIO()
```
pfultz2, line 94: Cant we just write to the strings directly? Doesn't something like `"Program = {}".format(inf)` work?

scxiao: fixed.

tools/download_models.sh
scxiao: Done

tools/download_models.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ -z "$ONNX_HOME" ]
+then
+   ONNX_HOME=$HOME
+fi
+
+model_dir=$ONNX_HOME/.onnx/models
+tmp_dir=$ONNX_HOME/tmp/
+mkdir -p $model_dir
+mkdir -p $tmp_dir
+for name in bvlc_alexnet densenet121 inception_v1 inception_v2 \
```
kahmed10, line 12: can we rewrite so that it is easier to add/remove models in the future? Also, if we are disabling tests in onnx_backend_test.py, we should probably disable the the downloading of those files here.

scxiao: fixed.

kahmed10 (approved):

pfultz2 (approved):

pfultz2 (approved):

PR #629, opened 2020-09-04
Add parallel stream analysis

pfultz2:
This can help identify possible race issues in a migraphx program. This will run on the gpu tests.


src/analyze_streams.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <migraphx/instruction.hpp>
+#include <migraphx/errors.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+bool happens_before(const std::vector<std::size_t>& e1, const std::vector<std::size_t>& e2)
+{
+    return std::equal(e1.begin(), e1.end(), e2.begin(), e2.end(), std::less_equal<>{}) and
+           not std::equal(e1.begin(), e1.end(), e2.begin(), e2.end(), std::greater_equal<>{});
```
kahmed10, line 14: If I understand this correctly, if some or all values in e1 are less than e2, then e1 happens before e2. However, if all of the values are equal, e1 does not happen before e2. I am not sure why `greater_equal` is needed since the first predicate for `less_equal` would return false if any value is greater. I think the second predicate can be just `equal`.

pfultz2: The logic is that all of them should be less than equal and at least one of them is strictly less than. See [here](https://en.wikipedia.org/wiki/Vector_clock#Partial_ordering_property).

kahmed10: Both logic cases will work, but it won't be necessary to check `greater` in this case. You can change it to just `std::equal` for the second conditional or leave it otherwise.

kahmed10 (approved):

scxiao:
@pfultz2 The analyze_stream is not actually called right?

src/analyze_streams.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+std::vector<stream_race> analyze_streams(const program& p, const stream_model& m)
+{
+    using vector_clock = std::vector<std::size_t>;
+    std::vector<stream_race> races;
+    auto nstream = m.get_nstream();
+    std::vector<vector_clock> vclock(nstream, vector_clock(nstream));
```
scxiao, line 22: can non-diagonal-elements be non-zero?

pfultz2:
> The analyze_stream is not actually called right?

Its called by the unit test, and also by the gpu tests (here)[https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/629/files#diff-dd8cbdcad1cd67247f99f938ed5db789R134].

src/analyze_streams.cpp
pfultz2: Yea, otherwise it would be just a simple [Lamport timestamp](https://en.wikipedia.org/wiki/Lamport_timestamp).

scxiao (approved):

PR #630, opened 2020-09-08
Where op

scxiao:
Used on distillbert and gpt-2-ml model


src/onnx/onnx.cpp
```diff
@@ -2424,11 +2426,22 @@ struct onnx_parser
...
         return l;
     }
 
+    instruction_ref
+    parse_where(const std::string&, const node_info&, std::vector<instruction_ref> args)
+    {
+        auto type = args[1]->get_shape().type();
+        auto cond = prog.add_instruction(
+            make_op("convert", {{"target_type", shape::type_string(type)}}), args[0]);
+        auto diff = add_broadcastable_binary_op(args[1], args[2], "sub");
```
kahmed10, line 2440: can we make a comment in the code of the equation being used that is equivalent to the conditional statement?

scxiao: add one

src/include/migraphx/serialize.hpp
```diff
@@ -102,8 +103,15 @@ auto to_value_impl(rank<10>, const T& x) -> decltype(x.to_value())
     return x.to_value();
 }
 
+template <class T, MIGRAPHX_REQUIRES(std::is_same<T, shape::type_t>{})>
+value to_value_impl(rank<11>, const T& x)
```
pfultz2, line 107: Why is this overload added? The enum overload [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/serialize.hpp#L88) should handle this.

src/include/migraphx/serialize.hpp
```diff
@@ -6,6 +6,7 @@
 #include <migraphx/reflect.hpp>
 #include <migraphx/requires.hpp>
 #include <migraphx/rank.hpp>
+#include <migraphx/shape.hpp>
```
pfultz2, line 9: This header shouldn't depend on the shape class.

src/include/migraphx/serialize.hpp
scxiao: Without the overload, we need a constructor for the value with an enum as the input...

pfultz2 (approved):

kahmed10 (approved):

PR #635, opened 2020-09-15
Improve API for program parameters

pfultz2:
This improves the API usability. For python API, it allows passing a numpy array in directly in the python dict. For C++, the program parameters can be constructed using initializer lists.


kahmed10:
Is there a way to support both previous API and the new API change? Otherwise, we will have to update all of our external scripts/documentation as well.

kahmed10 (approved):

PR #636, opened 2020-09-16
Op constructor c/python api

scxiao:


src/py/migraphx_py.cpp
```diff
@@ -20,6 +22,94 @@
...
+        {
+            auto elem_v = kwargs_to_value(val.cast<py::kwargs>());
+            v.push_back(elem_v);
+        }
+        else if(py::isinstance<py::list>(val))
+        {
+            auto elem_v = list_to_value(val.cast<py::list>());
+            v.push_back(elem_v);
+        }
+        else if(py::isinstance<py::bool_>(val))
```
pfultz2, line 43: Can we write a visitor for this?

```cpp
template<class T, class F>
void visit_py(T x, F f) 
{
    if(py::isinstance<py::kwargs>(val))
    {
        f(val.cast<py::kwargs>());
    }
    else if(py::isinstance<py::list>(val))
    {
        f(val.cast<py::list>());
    }
    else if(py::isinstance<py::bool_>(val))
    {
        f(val.cast<bool>(););
    }
    else if(py::isinstance<py::int_>(val))
    {
        f(val.cast<int>());
    }
    else if(py::isinstance<py::float_>(val))
    {
        f(val.cast<float>());
    }
    else if(py::isinstance<py::str>(val))
    {
        f(val.cast<std::string>());
    }
    else
    {
        MIGRAPHX_THROW("Unsupported data type");
    }
}

migraphx::value migraphx_to_value(py::list lst)
{
    migraphx::value result = migraphx::value::array{};
    for(auto&& val:lst)
    {
        visit_py(val, [&](auto x) {
            result.push_back(migraphx::to_value(x));
        });
    }
    return result;
}
```

src/py/migraphx_py.cpp
```diff
@@ -20,6 +22,94 @@
 using half   = half_float::half;
 namespace py = pybind11;
 
+namespace migraphx {
+migraphx::value kwargs_to_value(py::kwargs kwargs);
+
+migraphx::value list_to_value(py::list lst)
```
pfultz2, line 28: Should be named `migraphx_to_value` so it will be called automatically by `migraphx::to_value`.

src/py/migraphx_py.cpp
```diff
@@ -20,6 +22,94 @@
...
+        else
+        {
+            MIGRAPHX_THROW("LIST_TO_VALUE: Unsupported data type ");
+        }
+    }
+
+    return v;
+}
+
+migraphx::value kwargs_to_value(py::kwargs kwargs)
```
pfultz2, line 68: Should be named `migraphx_to_value` so it will be called automatically by `migraphx::to_value`.

src/py/migraphx_py.cpp
```diff
@@ -186,6 +276,18 @@ PYBIND11_MODULE(migraphx, m)
         .def("__ne__", std::not_equal_to<migraphx::program>{})
         .def("__repr__", [](const migraphx::program& p) { return migraphx::to_string(p); });
 
+    py::class_<migraphx::operation>(m, "op")
+        .def(py::init([](const std::string& name, py::kwargs kwargs) {
+            migraphx::value v = migraphx::value::object{};
+            if(kwargs)
+            {
+                v = migraphx::kwargs_to_value(kwargs);
```
pfultz2, line 284: This should just be `v = migraphx::to_value(kwargs)`.

pfultz2:
Can some unit tests be added?

src/api/jsonize_attr_string.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    if(i == str.length())
+    {
+        return {};
+    }
+
+    return str.substr(start, i - start + 1);
+}
+
+bool is_value_a_string(const std::string& op_name, const std::string& key)
```
pfultz2, line 68: I dont think we need to put quotes around the values, just the keys.

src/api/jsonize_attr_string.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <migraphx/errors.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+// get all elements of an array or an object, including '[]' and '{}'
+std::string get_elements_string(const std::string& str, const std::size_t start, const char brkt)
```
pfultz2, line 16: We dont need to parse the json. We can just lex the json(and skip whitespace). When the next token is `:` then we can check if the previous token has quotes, if not then we add quotes while we serialize it to a string.

Each token can be an iterator(or pointer) pair so we dont have to make copies of the string as we lex.

pfultz2: Actually you could use something like this to tokenize the json string:

```cpp

template<class Predicate>
auto lex_while(P p)
{
    return [=](const char* start, const char* end) {
        return std::find_if(start, end, [&](char c) {
            return not p(c);
        });
    };
}

template<class Predicate>
auto lex_if(P p)
{
    return [=](const char* start, const char* end) {
        if (p(*start))
            return start+1;
        return start;
    };
}

using token = std::pair<const char*, const char*>;
using lexer = std::function<const char*(const char* start, const char* end)>;

std::vector<token> tokenize(const char* start, const char* end, std::vector<lexer> lexers)
{
    std::vector<token> result;
    while(start != end) 
    {
        bool error = true;
        for(auto l:lexers) 
        {
            auto next = l(start, end);
            if (next != start)
            {
                if (not std::all_of(start, next, &isspace))
                    result.emplace_back(start, next);
                start = next;
                error = false;
                break;
            }
        }
        if (error)
            MIGRAPHX_THROW("Lex error");
    }
    return result;
}

std::vector<token> json_tokenize(const std::string& s)
{
    std::vector<lexer> lexers;
    // Quote
    lexers.push_back([](const char* start, const char* end) {
        if (*start != '\"')
            return start;
        while(start != end and *start != '\"')
        {
            if (*start == '\\')
                start++;
            start++;
        }
    });
    // Whitespace
    lexers.push_back(lex_while(&isspace));
    // Punctation
    lexers.push_back(lex_if(&ispunct));
    // Identifier/number
    lexers.push_back(lex_while([](char c) { return isalnum(c) or '_' or '.' or '+'; }));
    return tokenize(s.data(), s.data()+s.length(), lexers);
}
```

As we lex identifiers, we can add quotes to them if it doesn't start with a number, which should add quotes for keys or values. You could also extend the lexer to tokenize floating point and integers separately.

src/api/json_tokenize.hpp
```diff
@@ -0,0 +1,18 @@
...
+#define MIGRAPHX_GUARD_API_RTGLIB_JSONIZE_ATTR_STRING_HPP
+
+#include <string>
+#include <migraphx/config.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+using siter = std::string::const_iterator;
+using token = std::pair<siter, siter>;
```
pfultz2, line 11: These should just be moved to the .cpp file.

src/api/json_tokenize.hpp
```diff
@@ -0,0 +1,18 @@
...
+
+std::string json_tokenize(const std::string& s);
```
pfultz2, line 13: This should probably be called something like `convert_json`, as tokenize implies we would be returning tokens.

pfultz2: Maybe we can unit tests for this function(without the constructor API):

```cpp
EXPECT(convert_json("{key: 1}") == "{\"key\": 1}");

File: src/api/json_tokenize.cpp [hunk 9]
```diff
@@ -0,0 +1,116 @@
...
+
+std::string json_tokenize(const std::string& s)
+{
+    siter start = s.begin();
+    siter end   = s.end();
+    std::vector<token> tokens;
+
+    while(start != end)
+    {
+        auto colon_iter = colon(start, end);
```
pfultz2, line 78: Wont this find a colon in a quote?

src/api/json_tokenize.cpp
```diff
@@ -0,0 +1,116 @@
...
+            MIGRAPHX_THROW("KEY: single quote cannot be a key!");
+        }
+    }
+
+    // find key start
+    --end;
+    while(start != end)
+    {
+        // match
+        if(*ke == '\"' and *end == '\"')
```
pfultz2, line 47: What if the quote is escaped? Then it shouldn't stop.

src/api/json_tokenize.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+    // find key end
+    --end;
+    while(start != end)
+    {
+        if(*end != ' ')
+        {
+            break;
+        }
+        --end;
```
pfultz2, line 26: This could be written with `find_if` and reverse iterators.

src/api/json_tokenize.cpp
```diff
@@ -0,0 +1,116 @@
...
+        {
+            break;
+        }
+        else if(*ke != '\"' and (std::ispunct(*end) or *end == ' '))
+        {
+            ++end;
+            break;
+        }
+        --end;
+    }
```
pfultz2, line 57: This loop could be written as this:

```cpp
if(*ke == '\"')
{
    end = std::find_if(std::make_reverse_iterator(end), std::make_reverse_iterator(start), [](char c) {
        return c == '\"';
    }).base();
}
else
{
    end = std::find_if(std::make_reverse_iterator(end), std::make_reverse_iterator(start), [](char c) {
        return std::ispunct(c) or c == ' ';
    }).base();
    end++;
}
```

The `*ke` condition is moved out of the loop, and we use `std::find_if`.

src/api/json_tokenize.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+siter colon(siter start, siter end)
+{
+    return std::find_if(start, end, [](auto c) { return c == ':'; });
+}
+
+std::pair<siter, siter> key(siter start, siter end)
+{
+    // find key end
+    --end;
```
pfultz2, line 19: Maybe add a `assert(*end == ':')`

src/api/json_tokenize.hpp
scxiao: Done

src/api/json_tokenize.cpp
scxiao: I just assume our usage does not have this scenario. Do we need to consider it? Also the case of single quote inside a quote?

src/api/json_tokenize.cpp
scxiao: So you mean we have a key with only quote at the end, and the start quote is missing?

src/api/json_tokenize.hpp
scxiao: Looks like we do not expose the api convert_to_json()

src/api/convert_to_json.cpp
```diff
@@ -0,0 +1,116 @@
...
+    std::vector<token> result;
+    while(start != end)
+    {
+        bool error = true;
+        for(auto l : lexers)
+        {
+            auto next = l(start, end);
+            if(next != start)
+            {
+                if(not std::all_of(start, next, &isspace))
```
pfultz2, line 45: This condition should be removed, so we dont strip whitespace in the output.

scxiao: removed.

test/convert_to_json.cpp
```diff
@@ -0,0 +1,60 @@
...
+    auto jstr       = migraphx::convert_to_json(str);
+    EXPECT(jstr == "{\"abc\" : {\"key\" : \"value\"}}");
+}
+
+TEST_CASE(key_space_value)
+{
+    std::string str = "{abc    : [key, value]}";
+    auto jstr       = migraphx::convert_to_json(str);
+    EXPECT(jstr == "{\"abc\"    : [\"key\", \"value\"]}");
+}
```
pfultz2, line 58: There is some more test cases that probably should be added:

* `{ key: 1 }`
* `{ key: -1 }`
* `{ key: 1.0 }`
* `{ key: -1.0 }`
* `{ key: 1e-10 }`
* `{ key: 1E+10 }`
* `{ key: nan }`
* `{ key: true }`
* `{ key: false }`
* `{ key: null }`

Quotes should be around the `key` but not the value.

scxiao: added a few more unit tests

pfultz2 (approved):

src/api/api.cpp
```diff
@@ -138,6 +141,24 @@ void quantize_int8_wrap(program& prog, const target& t, quantize_int8_options& o
...
+    }
+    auto op = make_op(name, v);
+
+    return op;
+}
+
+const char* op_name(const operation& op)
+{
+    const auto name = std::make_unique<std::string>(op.name());
+    return name->c_str();
```
pfultz2, line 159: Why cant this be `op.name().c_str()`?

pfultz2: Actually this wont work as the string will be deallocated causing a dangling reference. We need to update the operation interface to return a string_view. Let me look into a PR to update that.

pfultz2: Actually, `string_view` wont work either since its not guaranteed to be a null terminated string. I think the API needs to be change to fill a buffer. api.py can automatically do this, but a wrapper needs to be added. You can do something similar to the [`vector_c_wrap`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/tools/api.py#L700) for strings:

```python
@cwrap('std::string')
def string_c_wrap(p):
    t = 'const char*'
    if p.returns:
        if p.type.is_reference():
            p.add_param(t.add_pointer())
            p.bad_param('${name} == nullptr', 'Null pointer')
            p.cpp_write = '${type}(${name})'
            p.write = [
                '*${name} = ${result}.c_str()'
            ]
        else:
            p.add_param(t)
            p.bad_param('${name} == nullptr', 'Null pointer')
            p.cpp_write = '${type}(${name})'
            p.write = [
                'std::copy(${result}.begin(), ${result}.end(), ${name})'
            ]
    else:
        p.add_param(t)
        p.bad_param('${name} == nullptr', 'Null pointer')
        p.read = '${type}(${name})'
```

scxiao: Thanks for the help. I made changes accordingly.

src/api/migraphx.py
```diff
@@ -193,7 +193,7 @@ def operation(h):
     h.constructor('create',
                   api.params(name='const char*', attributes='const char*'),
                   fname='migraphx::create_op')
-    h.method('name', invoke='migraphx::op_name($@)', returns='const char*')
+    h.method('name', invoke='migraphx::op_name($@)', returns='std::string&')
```
pfultz2, line 196: This needs to be `std::string` without the `&` because it returns by value.

pfultz2: Also, `invoke='migraphx::op_name($@)'` is not needed because it will by default call `op.name()`.

tools/api.py
```diff
@@ -725,6 +725,28 @@ def vector_c_wrap(p):
...
+            p.bad_param('${name} == nullptr', 'Null pointer')
+            p.cpp_write = '${type}(${name})'
+            p.write = ['*${name} = ${result}.c_str()']
+        else:
+            p.add_param(t)
+            p.bad_param('${name} == nullptr', 'Null pointer')
+            p.cpp_write = '${type}(${name})'
+            p.write = [
+                'std::copy(${result}.begin(), ${result}.end(), ${name})'
+            ]
```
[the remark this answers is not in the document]
scxiao: Thanks, changed accordingly

src/api/migraphx.py
scxiao: fixed

src/api/api.cpp
```diff
@@ -676,6 +702,34 @@ migraphx_program_equal(bool* out, const_migraphx_program_t program, const_migrap
...
+extern "C" migraphx_status
+migraphx_operation_name(char* out, size_t out_size, migraphx_operation_t operation)
+{
+    return migraphx::try_([&] {
+        if(out == nullptr)
+            MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter out: Null pointer");
+        if(operation == nullptr)
+            MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter operation: Null pointer");
+        auto&& api_result = (operation->object).name();
+        auto* it = std::copy_n(api_result.begin(), std::min(api_result.size() + 1, out_size), out);
```
pfultz2, line 728: Why is this `api_result.size() + 1`? It seems like this would read out of bounds. This should probably be `std::min(api_result.size(), out_size - 1)`, and then do `*std::next(it) = 0`.

scxiao: I had thought the string contains a null terminator, so we just copy the null terminal to the output string. Looks like std::string does not contain that. changed according.

scxiao: After checking the function explanation and running the unit test, I changed it to:
std::min(api_result.size(), out_size - 1), and then do *it = 0

If you look a the algorithm in https://en.cppreference.com/w/cpp/algorithm/copy_n, it is like:
template< class InputIt, class Size, class OutputIt>
OutputIt copy_n(InputIt first, Size count, OutputIt result)
{
    if (count > 0) {
        *result++ = *first;
        for (Size i = 1; i < count; ++i) {
            *result++ = *++first;
        }
    }
    return result;
}
So the return value is the iterator after the last char that is copied. For our code, only when I do *it = 0, is the unit test correct.

Looks like the description has some problem.

scxiao:
@pfultz2 @mvermeulen If we look at the details, looks like all builds have finished, but here it still shows the onnx runtime is pending. So has it really finished?

pfultz2:
onnx stage passes, this can be merged in.

PR #639, opened 2020-09-17
Add build flag for fast math

kahmed10:


pfultz2:
The python docs need to be updated.

test/gpu/ops_test.cpp
```diff
@@ -2584,6 +2585,49 @@ void manual_identity()
...
+    options.fast_math = false;
+    p.compile(migraphx::gpu::target{}, options);
+    bool found_gelu_new = false;
+    for(auto ins : iterator_for(p))
+    {
+        if(ins->name() == "gpu::gelu_new")
+            found_gelu_new = true;
+    }
+    CHECK(found_gelu_new);
+}
```
pfultz2, line 2629: This should be put in a different cpp file.

pfultz2: Also, there should be two tests, one with it enabled and another with it disabled.

kahmed10: fixed

test/gpu/fast_math.cpp
```diff
@@ -0,0 +1,68 @@
...
+            found_gelu = true;
+    }
+    CHECK(found_gelu);
+}
+
+TEST_CASE(enable_fast_gelu)
+{
+    migraphx::program p = create_gelu();
+    p.compile(migraphx::gpu::target{});
+    check_gelu_version(p, "gpu::gelu");
```
pfultz2, line 56: This could be written with `any_of`:

```cpp
CHECK(any_of(p, [&](auto&& i) { return i.name() == "gpu::gelu"; });
```

scxiao (approved):

test/gpu/fast_math.cpp
kahmed10: fixed

pfultz2 (approved):

PR #646, opened 2020-09-24
Upsample op

scxiao:


src/onnx/onnx.cpp
```diff
@@ -171,6 +171,7 @@ struct onnx_parser
         add_mem_op("Split", &onnx_parser::parse_split);
         add_mem_op("Tile", &onnx_parser::parse_tile);
         add_mem_op("Transpose", &onnx_parser::parse_transpose);
+        add_mem_op("Upsample", &onnx_parser::parse_upsample);
```
kahmed10, line 174: Can you add a mem_op for `"Resize"` and refer it to `&onnx_parser::parse_upsample`?

pfultz2 (approved):

src/onnx/onnx.cpp
scxiao: We can reuse the algorithm, but we cannot directly do that since inputs, attributes are different. I will work on the Resize next week.

kahmed10 (approved):

PR #650, opened 2020-09-28
Added less and greater ops

turneram:


pfultz2:
Can you push this to branch instead of fork? For some reason our CI doesn't run on forks.

src/onnx/onnx.cpp
```diff
@@ -2486,6 +2445,28 @@ struct onnx_parser
         return l;
     }
 
+    instruction_ref
```
kahmed10, line 2448: After looking back at `parse_equal`, `parse_greater`, and `parse_less`, they are all basically the same except the operator name. You could probably refactor these functions into using just one `parse_compare` function, which takes in an additional parameter of `op_name` (see `parse_reduce_oper` function for an example).

You will also need to change L139, L145, and L150 to pass in the second argument of the `op_name` (see L164-168 as an example).

src/include/migraphx/op/greater.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct greater : binary<greater>
+{
+    value attributes() const
+    {
+        auto a           = base_attributes();
+        a["commutative"] = true;
```
pfultz2, line 19: I dont think greater is `commutative`.

PR #654, opened 2020-09-30
Less and greater

turneram:


src/onnx/onnx.cpp
```diff
@@ -2486,6 +2487,28 @@ struct onnx_parser
         return l;
     }
 
+    instruction_ref
```
kahmed10, line 2490: After looking back at parse_equal(), parse_greater(), and parse_less(), they are all basically the same except the operator name. You could probably refactor these functions into using just one parse_compare() function, which takes in an additional parameter of op_name (see parse_reduce_oper() function for an example).

You will also need to change L139, L145, and L150 to pass in the second argument of the op_name (see L164-168 for an example).

src/targets/gpu/include/migraphx/gpu/less.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <migraphx/argument.hpp>
+#include <migraphx/config.hpp>
+#include <hip/hip_runtime_api.h>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+void less(hipStream_t stream, const argument& result, const argument& arg1, const argument& arg2);
```
pfultz2, line 14: This header should go under  src/targets/gpu/include/migraphx/gpu/device/less.hpp. You also need to add a  src/targets/gpu/include/migraphx/gpu/less.hpp header to create the gpu operator:

```cpp
struct hip_less : binary_device<hip_equal, device::less>
{
};
```

Same thing applies to greater.

src/onnx/onnx.cpp
scxiao: @turneram, the combined function would be like:

    instruction_ref parse_compare_op(const std::string&,
                                 const std::string& op_name,
                                 const node_info&,
    {
        auto l = add_broadcastable_binary_op(args[0], args[1], op_name);
        if(l->get_shape().type() != shape::bool_type)
        {
            l = prog.add_instruction(make_op("convert", {{"target_type", shape::bool_type}}), l);
        }
        return l;
    }

Then the add_mem_op is called like:

add_mem_op("Equal", "equal", &onnx_parser::parse_compare_op);
add_mem_op("Greater", "greater", &onnx_parser::parse_compare_op);
add_mem_op("Less", "less", &onnx_parser::parse_compare_op);

PR #661, opened 2020-10-08
Switch to gcc 7 and drop gcc 5 

pfultz2:


pfultz2:
Closing as this is no longer valid.

PR #662, opened 2020-10-09
add hip-clang dockerfile

kahmed10:
Based on @scxiao 's onnx runtime dockerfile for hip-clang and rocm3.7


kahmed10:
@pfultz2 it looks like we already have a dockerfile for hip-clang used for our CI from PR #651. However, do we want another Dockerfile that builds from source? Otherwise, we should update the readme and point them to the `hip-clang.docker`.

pfultz2:
> However, do we want another Dockerfile that builds from source?

Yea we probably do, but I am not sure what to call it.

examples/Dockerfile.hip-clang
```diff
@@ -0,0 +1,30 @@
...
+  sh -c 'echo deb [arch=amd64] http://repo.radeon.com/rocm/apt/3.7/ xenial main > /etc/apt/sources.list.d/rocm.list'
+
+RUN apt-get update &&\
+    apt-get install -y sudo git bash build-essential cmake rocm-dkms libpython3.6-dev python3-pip miopen-hip rocblas half 
+
+# Install rbuild
+RUN pip3 install https://github.com/RadeonOpenCompute/rbuild/archive/master.tar.gz
+
+ENV PATH /opt/miniconda/bin:/code/cmake-3.14.3-Linux-x86_64/bin:${PATH}
+
```
scxiao, line 23: Looks like we did not  install the cmake-3.14.3 in this dockerfile, maybe you can add a line like:

RUN mkdir /code && cd /code && \
    wget https://github.com/Kitware/CMake/releases/download/v3.14.3/cmake-3.14.3-Linux-x86_64.sh && \
    chmod +x cmake-3.14.3-Linux-x86_64.sh && \
    ./cmake-3.14.3-Linux-x86_64.sh --prefix=/usr/local --exclude-subdir --skip-license

You may also need to include the wget in the apt install list

kahmed10: I think this was a remnant of taking your onnx runtime dockerfile. Is it worthwhile to use a newer version of cmake in this dockerfile? If so, then I'll make changes to install it.

scxiao (approved):

scxiao:
Build for this one is completed and can be merged.

PR #665, opened 2020-10-15
add get_main_module api

scxiao:


src/include/migraphx/program.hpp
```diff
@@ -139,6 +141,8 @@ struct program
     friend bool operator==(const program& x, const program& y);
     friend bool operator!=(const program& x, const program& y) { return !(x == y); }
 
+    module& get_main_module() { return *this; }
```
pfultz2, line 144: Can this be added to our python/C API?

scxiao: Thanks. Looks like I forgot. just added. will add corresponding unit tests tomorrow.

pfultz2: This should return `module*` not `module&`.

kahmed10 (approved):

src/api/api.cpp
```diff
@@ -616,11 +628,57 @@ migraphx_shapes_get(const_migraphx_shape_t* out, migraphx_shapes_t shapes, size_
...
+    return migraphx::try_([&] { destroy((module)); });
+}
+
+extern "C" migraphx_status migraphx_module_create(migraphx_module_t* module)
+{
+    return migraphx::try_(
+        [&] { *module = object_cast<migraphx_module_t>(allocate<migraphx::module>()); });
+}
+
+extern "C" migraphx_status migraphx_module_compile(migraphx_module_t module,
```
pfultz2, line 642: A module shouldn't have a compile function.

src/api/migraphx.py
```diff
@@ -163,8 +163,19 @@ def shapes(h):
              returns='const migraphx::shape&')
 
 
+@auto_handle
+def module(h):
+    h.constructor('create')
```
pfultz2, line 168: I dont think we should allow constructing a module. A module should only come from the program(ie `get_main_module`), and the program will own the module, so there is no need to expose a `destroy` or `create` function either.

src/include/migraphx/program.hpp

src/api/migraphx.py
pfultz2: So PR#673 adds the ability to not generate the `destroy` method. It just needs to be changed to `@auto_handle(ref=True)`.

scxiao: For the c++ api, looks like handle_base struct needs a destroy function, how can we handle that? definition of the handle base is 
template <class T, class D, D Deleter>
struct handle_base
{}

pfultz2 (approved):

PR #666, opened 2020-10-16
Fix bert fusions

pfultz2:


scxiao:
I will ran the msft model to check if there is a regression. If not, I will approve this PR

scxiao:
> I will ran the msft model to check if there is a regression. If not, I will approve this PR

No regression observed for this PR.

scxiao (approved):

src/instruction.cpp
```diff
@@ -219,6 +219,37 @@ void instruction::finalize(context& ctx)
...
+        else
+            os << "{" << ins.get_literal() << "}";
+    }
+    else
+    {
+        os << ins.get_operator();
+    }
+}
+
+void instruction::debug_print() const
```
kahmed10, line 238: where is this being used?

pfultz2: Its only used for debugging. We have `debug_print` in the program as well. This is here because sometime the program is unavailable.

kahmed10 (approved):

PR #667, opened 2020-10-19
Normalize ops

scxiao:


src/include/migraphx/normalize_op.hpp
```diff
@@ -0,0 +1,35 @@
...
+{
+    using type = T;
+};
+template <class T, class... Ts>
+using dependent_type = typename select_dependent_type<T, Ts...>::type;
+
+void normalize_op(operation& op, std::vector<shape> inputs);
+
+template <class T>
+shape normalize_compute_shape_op(T&& x, std::vector<shape> inputs)
```
pfultz2, line 25: This should be in the operation header, and under the `detail` namespace.

src/normalize_op.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <migraphx/normalize_op.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+void normalize_op(operation& op, std::vector<shape> inputs)
+{
+    int64_t n_dim = static_cast<int64_t>(inputs[0].lens().size());
+    value val     = op.to_value();
+    if(val.contains("axis"))
```
pfultz2, line 11: The name of the field should come from the `normalize_axes` attribute.

src/normalize_ops.cpp
```diff
@@ -0,0 +1,144 @@
...
+                                                "reduce_mean",
+                                                "reduce_min",
+                                                "reduce_prod",
+                                                "reduce_sum",
+                                                "softmax",
+                                                "squeeze",
+                                                "unsqueeze"};
+    for(auto ins : iterator_for(p))
+    {
+
```
pfultz2, line 109: This should just call your `normalize_op` function.

src/targets/cpu/target.cpp
```diff
@@ -16,7 +17,9 @@ std::string target::name() const { return "cpu"; }
 
 std::vector<pass> target::get_passes(migraphx::context&, const compile_options&) const
 {
-    return {rewrite_rnn{},
+    return {normalize_ops{},
+            dead_code_elimination{},
```
pfultz2, line 21: I dont think DCE is needed after this pass.

src/normalize_axes.cpp
```diff
@@ -0,0 +1,116 @@
...
+               !std::equal(ends.begin(), ends.end(), axis_lens.begin(), std::less_equal<>{}))
+            {
+                std::transform(ends.begin(),
+                               ends.end(),
+                               axis_lens.begin(),
+                               ends.begin(),
+                               [=](auto i, auto dim) {
+                                   i = (i < -dim) ? -dim : ((i > dim) ? dim : i);
+                                   return (i < 0) ? (i + dim) : i;
+                               });
```
pfultz2, line 102: Both the starts and ends do the same thing. It also looks like this logic could be applied to axes as well.

src/include/migraphx/op/slice.hpp
```diff
@@ -25,6 +26,8 @@ struct slice
         return pack(f(self.axes, "axes"), f(self.starts, "starts"), f(self.ends, "ends"));
     }
 
+    value attributes() const { return {{"axes", axes}, {"starts", starts}, {"ends", ends}}; }
```
pfultz2, line 29: This should be `{{"axes", 'axes', ''starts', 'ends'}`.

src/include/migraphx/op/argmin.hpp
```diff
@@ -21,14 +22,16 @@ struct argmin
         return pack(f(self.axis, "axis"));
     }
 
+    value attributes() const { return {{"axis", axis}}; }
```
pfultz2, line 25: I dont think we need a different attribute name for a single axis. We can tell whether its axes or axis by checking the type(ie `val.is_array()`).

src/normalize_axes.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <migraphx/stringutils.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+void normalize_axes(operation& op, const std::vector<std::size_t>& lens)
+{
+    int64_t n_dim = static_cast<int64_t>(lens.size());
+    auto val      = op.attributes();
+    if(val.contains("axis"))
```
pfultz2, line 12: The attribute should have the field names and not the actual field.

pfultz2: Also, the attribute name should probably be `normalize_axes`.

src/normalize_axes.cpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        auto axis = val["axis"].without_key().to<int64_t>();
+        if(op.name() == "flatten")
```
pfultz2, line 15: I would like to avoid checking operator names here, but I think we need a richer set of attributes. We can leave it for now.

src/normalize_axes.cpp
```diff
@@ -0,0 +1,116 @@
...
+            op.from_value(val);
+        }
+    }
+    else if(val.contains("axes"))
+    {
+        auto axes = val["axes"].without_key().to_vector<int64_t>();
+
+        if(op.name() == "unsqueeze")
+        {
+            n_dim = n_dim + axes.size();
```
pfultz2, line 45: What does this line do?

src/normalize_axes.cpp
scxiao: How about the following attributes for each field:
1) is_array
2) clip_max (slice)
3) clip_min (slice)
4) include_max (flatten)
5) include_min
6) use_output_rank (unsqueeze)

pfultz2: That sounds good, except we dont need an is_array attribute as we can infer that from the value.
Also, would this be in addition to the other attributes?

scxiao: For "Also, would this be in addition to the other attributes?", what does it mean for other attributes?

src/normalize_axes.cpp
```diff
@@ -5,114 +5,160 @@
...
 inline namespace MIGRAPHX_INLINE_NS {
 
-bool normalize_axes(operation& op, const std::vector<std::size_t>& lens)
+// different attributes
+// 1) use_input(default)/use_output
+// 2) use_rank(default)/use_len
+// 3) clip_min(default)/not_clip_min
+//   3.1) include_min(default)/exclude_min
+// 4) clip_max(default)/not_clip_max
+//   4.1) exclude_max(default)/include_max
```
pfultz2: I wonder if we should use a enum instead of raw strings. We can probably change that in a later PR.

src/include/migraphx/op/broadcast.hpp
```diff
@@ -35,10 +35,16 @@ struct broadcast
     std::string name() const { return "broadcast"; }
     shape compute_shape(std::vector<shape> inputs) const
     {
-        auto t     = inputs.at(0).type();
         auto input = inputs.at(0);
+        auto t     = input.type();
 
         std::vector<size_t> bcast_strides(broadcast_lens.size(), 0);
+        // the broacast op is deprecated now, so not handling the negative
```
pfultz2, line 42: What do you mean the broadcast op is deprecated? We should normalize the `axis`.

src/include/migraphx/op/gather.hpp
```diff
@@ -17,35 +18,33 @@ namespace op {
...
 
     template <class Self, class F>
     static auto reflect(Self& self, F f)
     {
         return pack(f(self.axis, "axis"));
     }
 
+    value attributes() const
+    {
+        value attr;
```
pfultz2, line 31: This is only for the normalize_axes is attribute. I would call it `normalize` or `normalize_axes` to make it a little bit clearer.

src/normalize_axes.cpp
scxiao: Changed

src/include/migraphx/op/broadcast.hpp
scxiao: This operator does not existed anymore, which mean no new model will use it and existing models donot have such a problem. Anyway, I can change it.

src/include/migraphx/op/gather.hpp
scxiao: OK, will change that.

scxiao: changed

src/include/migraphx/op/broadcast.hpp
scxiao: After checking our code, we do not parse the broadcast code at all, which means the only scenario we have the broadcast operator is we explicitly add it when necessary. In addition, there is no specification about the range of the axis attribute. And the way of converting negative value to positive is different other operators.

pfultz2: Its deprecated in onnx, but it doesn't mean its deprecated in migraphx. Did this operator support negative axis before?

src/include/migraphx/op/common.hpp
```diff
@@ -23,6 +23,23 @@ enum class rnn_direction
...
+//   4.1) exclude_max(default)/include_max
+enum class op_normalize_attributes
+{
+    use_len,
+    use_output,
+    clip_max,
+    clip_min,
+    include_max,
+    include_min
+};
```
pfultz2, line 41: Make a seperate header for this.

src/include/migraphx/op/common.hpp
```diff
@@ -23,6 +23,23 @@ enum class rnn_direction
...
 };
 
+// different attributes
+// 1) use_input(default)/use_output
+// 2) use_rank(default)/use_len
+// 3) clip_min(default)/not_clip_min
+//   3.1) include_min(default)/exclude_min
+// 4) clip_max(default)/not_clip_max
+//   4.1) exclude_max(default)/include_max
+enum class op_normalize_attributes
```
pfultz2, line 33: Could we name this `normalize_attribute`?

src/include/migraphx/op/gather.hpp
pfultz2: I meant rename the `attr` variable to `normalize` not the name of the attribute.

src/include/migraphx/op/broadcast.hpp
scxiao: No, we do not have negative axis. Actually, the axis is declared as uint64_t.

src/include/migraphx/op/common.hpp
scxiao: fixed

src/include/migraphx/op/gather.hpp
scxiao: fixed back

src/include/migraphx/op/common.hpp
scxiao: fixed

pfultz2 (approved):

pfultz2:
@kahmed10 Can you review this PR?

scxiao:
@kahmed10, @turneram Can you review this PR when you get a chance?

turneram (approved):

kahmed10 (approved):

PR #671, opened 2020-10-23
Split cpu and reference implementation

pfultz2:
It doesn't enable the cpu target yet(I plan to do that in a future PR).

* Converts `cpu::target` to `ref::target`
* Update driver so it can build and verify multiple targets
* Update gpu_ops test to run for multiple targets
* Split the verify tests into seperate cpp files


src/targets/ref/lowering.cpp
```diff
@@ -0,0 +1,974 @@
...
+                    });
+                });
+        }
+
+        return output;
+    }
+};
+MIGRAPHX_REGISTER_OP(cpu_batch_norm_inference)
+
+struct cpu_lrn
```
scxiao, line 130: should we rename the name to ref_lrn here?

test/quantization.cpp
```diff
@@ -1009,7 +1009,7 @@ TEST_CASE(int8_quantization_dot)
         m["a"] = migraphx::generate_argument(sa);
         m["c"] = migraphx::generate_argument(sc);
         std::vector<float> quant_result;
-        migraphx::target cpu_t = migraphx::cpu::target{};
+        migraphx::target cpu_t = migraphx::ref::target{};
```
scxiao, line 1012: Should we also cpu_t to ref_t here?

pfultz2:
I dont think its necessary to add tests fro env functions, so I think this can be merged in as well.

kahmed10 (approved):

test/gpu/quantization.cpp
```diff
@@ -4,21 +4,21 @@
...
 #include <migraphx/onnx.hpp>
-#include "test.hpp"
+#include <test.hpp>
 #include <migraphx/half.hpp>
 
 TEST_CASE(gpu_target_copy)
 {
     migraphx::target gpu_t = migraphx::gpu::target{};
-    migraphx::target cpu_t = migraphx::cpu::target{};
+    migraphx::target cpu_t = migraphx::ref::target{};
```
scxiao, line 21: maybe change the cpu_t to ref_t here?

test/gpu/quantization.cpp
```diff
@@ -4,21 +4,21 @@
...
     migraphx::shape s{migraphx::shape::int8_type, {2, 3, 4, 5}};
 
     auto cpu_arg_orig  = migraphx::generate_argument(s, 0x123456L);
```
scxiao, line 24: change to ref_arg_orig

test/gpu/quantization.cpp
```diff
@@ -83,7 +83,7 @@ TEST_CASE(int8_quantization)
         m["b"] = migraphx::generate_argument(sb);
         m["c"] = migraphx::generate_argument(sc);
         std::vector<float> cpu_result;
```
scxiao, line 85: change to ref_result

test/gpu/quantization.cpp
```diff
@@ -83,7 +83,7 @@ TEST_CASE(int8_quantization)
...
-        migraphx::target cpu_t = migraphx::cpu::target{};
+        migraphx::target cpu_t = migraphx::ref::target{};
```
scxiao, line 86: change to ref_t

test/gpu/quantization.cpp
scxiao: I just changed that and the following

scxiao (approved):

mvermeulen:
Is this one still pending?  Hard to tell if it passed all required steps.

pfultz2:
It looks like the job got cancelled, I am rebuilding it.

PR #678, opened 2020-11-05
Module impl

scxiao:


src/include/migraphx/program.hpp
```diff
@@ -139,9 +85,12 @@ struct program
     friend bool operator==(const program& x, const program& y);
     friend bool operator!=(const program& x, const program& y) { return !(x == y); }
 
+    module* get_main_module() { return &main_module; }
+    const module* get_main_module() const { return &main_module; }
+
     private:
     void assign(const program& p);
-    std::unique_ptr<program_impl> impl;
+    module main_module;
```
pfultz2, line 93: The module should be stored in the `program_impl` class.

src/program.cpp
```diff
@@ -20,81 +20,39 @@
...
+//         // TODO: Use all_of
+//         for(auto&& arg : ins->inputs())
+//         {
+//             assert(p.has_instruction(arg) && "Instruction not found");
+//             (void)arg;
+//         }
+
+//         print_func(ins, names);
+//     }
+// }
```
pfultz2, line 53: Is this commented out code necessary?

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
-               impl->instructions.begin(), impl->instructions.end(), [&](const instruction& x) {
-                   return std::addressof(*ins) == std::addressof(x);
-               }) != impl->instructions.end();
+    return main_module.has_instruction(ins);
 }
 
-std::size_t program::size() const { return impl->instructions.size(); }
-instruction_ref program::begin() const { return impl->instructions.begin(); }
-instruction_ref program::end() const { return impl->instructions.end(); }
+std::size_t program::size() const { return main_module.size(); }
```
pfultz2, line 97: Size should return the number of modules(which will always be `1` in this case).

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
+instruction_ref program::begin() const { return main_module.begin(); }
+instruction_ref program::end() const { return main_module.end(); }
```
pfultz2, line 99: This should iterate over the modules. Since we only have one module it can be written like this:

```cpp
module* program::begin() const { return &main_module; }
module* program::end() const { return begin()+1; }
```

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
-    // The else branch is to provide backward compatibility
-    else
-    {
-        return {last_ins.get_shape()};
-    }
-}
+std::vector<shape> program::get_output_shapes() const { return main_module.get_output_shapes(); }
 
-context& program::get_context() const { return impl->ctx; }
+context& program::get_context() const { return main_module.get_context(); }
```
pfultz2, line 103: The context should be stored in the program and not the module.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
-    }
-    else
-    {
-        return generic_eval(
-            *this, ctx, std::move(params), [&](auto&, auto f) { return check_context(f); });
-    }
+    return main_module.eval(std::move(params));
 }
 
 const int program_file_version = 1;
```
pfultz2, line 128: This should be incremented since we are changing the format.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
-    os << std::endl;
-
-    os << "Rate: " << rate << "/sec" << std::endl;
-    os << "Total time: " << total_time << "ms" << std::endl;
-    os << "Total instructions time: " << total_instruction_time << "ms" << std::endl;
-    os << "Overhead time: " << overhead_time << "ms"
-       << ", " << calculate_overhead_time << "ms" << std::endl;
-    os << "Overhead: " << std::round(overhead_percent) << "%"
-       << ", " << std::round(calculate_overhead_percent) << "%" << std::endl;
+    main_module.perf_report(os, n, std::move(params));
```
pfultz2, line 150: The perf_report should be in the program and not the module.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
-    run_passes(*this, t.get_passes(this->impl->ctx, options), options.trace);
-    auto invalid = this->validate();
-    if(invalid != impl->instructions.end())
-    {
-        auto index = std::distance(impl->instructions.begin(), invalid);
-        MIGRAPHX_THROW("Invalid program from compilation at instruction " + std::to_string(index));
-    }
-    this->finalize();
-}
+void program::compile(const target& t, compile_options options) { main_module.compile(t, options); }
```
pfultz2, line 109: The module class probably shouldn't have a compile method, but we can leave this for now.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
-    std::stringstream ss;
-    print_program(*this, [&](auto x, const auto& names) {
-        if(x == ins)
-        {
-            print_instruction(std::cout, x, names);
-            std::cout << std::endl;
-        }
-    });
-}
+void program::debug_print() const { main_module.debug_print(); }
```
pfultz2, line 153: debug_print should still do `std::cout << *this << std::endl`

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
+void program::debug_print(instruction_ref ins) const { main_module.debug_print(ins); }
```
pfultz2, line 154: This should probably removed from the program.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
 void program::debug_print(const std::vector<instruction_ref>& inss) const
 {
-    for(auto ins : inss)
-        debug_print(ins);
-    std::cout << std::endl;
+    main_module.debug_print(inss);
```
pfultz2, line 157: This should probably be removed from the program class, and only on the module class.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
 }
 
 void program::annotate(std::ostream& os, std::function<void(instruction_ref)> a) const
 {
-    print_program(*this, [&](auto ins, const auto& names) {
-        print_instruction(os, ins, names);
-        a(ins);
-        os << std::endl;
-    });
+    main_module.annotate(os, std::move(a));
```
pfultz2, line 193: This should print the module name. For now we can just hardcode this with `std::cout << "main:" << std::endl`.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
 
 program& program::sort()
 {
-    fix([&](auto self, auto ins) {
-        this->move_instruction(ins, this->begin());
-        for(auto child : ins->inputs())
-            self(child);
-    })(std::prev(this->end()));
-    assert(this->validate() == this->end());
+    main_module.sort();
```
pfultz2, line 198: `sort` should probably removed from the program, but we can leave it for now.

src/program.cpp
```diff
@@ -105,829 +63,148 @@ program::program(const program& p) { assign(p); }
...
 bool operator==(const program& x, const program& y) { return to_string(x) == to_string(y); }
 
 std::ostream& operator<<(std::ostream& os, const program& p)
 {
-    print_program(p, [&](auto ins, const auto& names) {
-        print_instruction(os, ins, names);
-        os << std::endl;
-    });
+    const auto* mm = p.get_main_module();
+    os << *mm;
```
pfultz2, line 207: This should print `main:` first, to show the instructions are coming from the main module.

src/include/migraphx/module.hpp
```diff
@@ -0,0 +1,152 @@
...
+    module& sort();
+
+    friend std::ostream& operator<<(std::ostream& os, const module& m);
+    friend bool operator==(const module& x, const module& y);
+    friend bool operator!=(const module& x, const module& y) { return !(x == y); }
+
+    private:
+    void assign(const module& m);
+    std::unique_ptr<module_impl> impl;
+};
```
pfultz2, line 147: Module class should have a `name()` method(and right now it can return `main`).

src/program.cpp
scxiao: Should we also add an API for the instruction begin() and end() of a program (it is actually the begin() and end instructions of the main module)?

src/program.cpp
scxiao: Looks like each module also need the ctx since it has a finalize() method, in which it calls the ins->finalize(ctx) function. Maybe we can change the module.finalize() to module.finalize(ctx). Is that OK?

src/program.cpp
scxiao: The module should have it, right?

kahmed10 (approved):

src/module.cpp
```diff
@@ -0,0 +1,677 @@
...
+
+void module::finalize(context& ctx)
+{
+    for(auto ins : iterator_for(*this))
+    {
+        ins->finalize(ctx);
+    }
+}
+
+const int module_file_version = 1;
```
pfultz2, line 422: The version should only be in the program, as this is used as a global-wide version to check for incompatibilities.

src/program.cpp
```diff
@@ -6,31 +6,30 @@
...
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 struct program_impl
 {
-    // A list is used to keep references to an instruction stable
-    std::list<instruction> instructions;
-    std::vector<std::string> input_names;
+    // A list is used to keep references to modules of the program
```
pfultz2, line 27: This is a map not a list.

src/program.cpp
```diff
@@ -69,7 +68,7 @@ static void print_program(const program& p, F print_func)
     std::unordered_map<instruction_ref, std::string> names;
     int count = 0;
 
-    for(auto ins : iterator_for(p))
+    for(auto ins : iterator_for(*p.get_main_module()))
```
pfultz2, line 71: The print function should reuse the print functions in the module.

src/program.cpp
```diff
@@ -111,342 +110,115 @@ program& program::operator=(program p)
 
 void program::assign(const program& p)
 {
-    // clean the current program
```
pfultz2, line 114: Cant this be implemented as `*this = p;`?

src/program.cpp
```diff
@@ -111,342 +110,115 @@ program& program::operator=(program p)
...
-    {
-        if(ins.name() == "@param")
-        {
-            auto&& name  = any_cast<builtin::param>(ins.get_operator()).parameter;
-            result[name] = ins.get_shape();
-        }
-    }
-    return result;
+    assert(contains(impl->modules, "main"));
+    return impl->modules["main"].begin();
```
pfultz2, line 165: Could you call `this->get_main_module()` instead of `impl->modules["main"]`? This can help get rid of the repeated asserts.

src/program.cpp
```diff
@@ -758,176 +443,68 @@ void program::debug_print(instruction_ref ins) const
...
 program& program::sort()
 {
-    fix([&](auto self, auto ins) {
-        this->move_instruction(ins, this->begin());
-        for(auto child : ins->inputs())
-            self(child);
-    })(std::prev(this->end()));
-    assert(this->validate() == this->end());
+    assert(contains(impl->modules, "main"));
+    impl->modules["main"].sort();
```
pfultz2, line 493: I think `program::sort()` can be removed.

scxiao: There are several locations that the program.sort() is called. Change all of them to prog.get_main_module()->sort()?

src/program.cpp
scxiao: fixed

src/program.cpp
scxiao: *this = p is going to be a shallow copy? the two program instances will share the instructions. I think we need a deep copy here.

scxiao: I just tested it, and *this = p does work for many our unit tests.

src/program.cpp
```diff
@@ -113,342 +86,115 @@ program& program::operator=(program p)
...
-            result[name] = ins.get_shape();
-        }
-    }
-    return result;
+    const auto* mm = this->get_main_module();
+    return mm->begin();
 }
 
-bool program::has_instruction(instruction_ref ins) const
+instruction_ref program::end() const
```
pfultz2, line 144: Program shouldn't have `begin` and `end` functions that return `instruction_ref`.

src/program.cpp
```diff
@@ -113,342 +86,115 @@ program& program::operator=(program p)
...
-            }
-        });
-    if(ins != this->end())
-        return ins->get_shape();
-    else
-        return {};
-}
-
-std::vector<std::string> program::get_parameter_names() const
+bool program::has_instruction(instruction_ref ins) const
```
pfultz2, line 130: This method should only be in the module.

src/program.cpp
```diff
@@ -113,342 +86,115 @@ program& program::operator=(program p)
...
 
 context& program::get_context() const { return impl->ctx; }
 
 instruction_ref program::validate() const
 {
-    return std::find_if(impl->instructions.begin(),
-                        impl->instructions.end(),
-                        [&](const instruction& i) { return !i.valid(impl->instructions.begin()); });
+    const auto* mm = this->get_main_module();
+    return mm->validate();
```
pfultz2, line 161: This should be ran over all modules.

src/program.cpp
```diff
@@ -113,342 +86,115 @@ program& program::operator=(program p)
...
 }
 
 void program::finalize()
 {
-    for(auto ins : iterator_for(*this))
-    {
-        ins->finalize(this->impl->ctx);
-    }
+    auto* mm = this->get_main_module();
+    mm->finalize(impl->ctx);
```
pfultz2, line 197: This should run `finalize` over all modules.

src/program.cpp
```diff
@@ -457,65 +203,8 @@ std::vector<argument> generic_eval(const program& p,
...
-            results.emplace(ins, trace(ins, [&] {
-                                return ins->get_operator().compute(ctx, ins->get_shape(), values);
-                            }));
-        }
-        assert(results.find(ins) != results.end());
-    }
-
-    return {results.at(std::prev(p.end()))};
+    const auto* mm = p.get_main_module();
+    return generic_eval(*mm, ctx, params, trace);
```
pfultz2, line 207: There shouldn't be a `generic_eval` in program. Instead `eval` from module should be called:

```cpp
std::vector<argument> program::eval(parameter_map params) const
{
    const auto* mm = p.get_main_module();
    mm->eval(params);
}
```

All the tracing handling should be done in the module.

src/program.cpp
```diff
@@ -597,37 +277,16 @@ void program::from_value(const value& v)
...
-                output = this->add_return(inputs);
-            else
-                output = this->add_instruction(op, inputs);
-        }
-        instructions[node.at("output").to<std::string>()] = output;
+        const auto& key = vv.get_key();
+        auto val        = vv.without_key();
+        module modl;
+        modl.from_value(val);
+        modl.finalize(this->impl->ctx);
```
pfultz2, line 287: We shouldn't finalize the module until they are all loaded.

src/program.cpp
```diff
@@ -597,37 +277,16 @@ void program::from_value(const value& v)
...
+        impl->modules[key] = modl;
     }
-    this->finalize();
```
pfultz2, line 630: This line should stay.

src/program.cpp
```diff
@@ -760,176 +419,66 @@ void program::debug_print(instruction_ref ins) const
...
-    return "m" + replace_string(name, "@", "x");
+    auto& ctx = this->impl->ctx;
+    generic_eval(*this, ctx, std::move(params), [](auto&&...) { return argument{}; });
 }
 
-static std::string cpp_op_var(const std::string& name, instruction_ref ins)
+void program::annotate(std::ostream& os, std::function<void(instruction_ref)> a) const
 {
-    return replace_string(name, "@", ins->name());
+    const auto* mm = this->get_main_module();
```
pfultz2, line 444: This should loop over all modules.

src/program.cpp
```diff
@@ -760,176 +419,66 @@ void program::debug_print(instruction_ref ins) const
...
-        }
-    });
-    os << "}" << std::endl;
+    os << "migraphx::program p;" << std::endl;
+    const auto* mm = this->get_main_module();
+    mm->print_cpp(os);
 }
 
-static std::string cpp_var_name(const std::string& name)
+void program::dry_run(std::unordered_map<std::string, argument> params) const
```
pfultz2, line 436: This should probably be added to module class.

src/program.cpp
```diff
@@ -760,176 +419,66 @@ void program::debug_print(instruction_ref ins) const
...
+void program::annotate(std::ostream& os, std::function<void(instruction_ref)> a) const
 {
-    return replace_string(name, "@", ins->name());
+    const auto* mm = this->get_main_module();
+    std::cout << mm->name() << ":" << std::endl;
+    mm->annotate(os, std::move(a));
 }
 
-static void print_op_attributes(std::ostream& os, const std::string& name, const operation& op)
+module* program::get_main_module()
```
pfultz2, line 449: This doesnt look like this function returns the module.

src/program.cpp
```diff
@@ -760,176 +419,66 @@ void program::debug_print(instruction_ref ins) const
...
 {
-    std::string x = to_string(op);
-    if(contains(x, "["))
+    if(!contains(impl->modules, "main"))
```
pfultz2, line 451: There is no need to search twice, just do `return &impl->modules["main"]`.

src/program.cpp
```diff
@@ -760,176 +419,66 @@ void program::debug_print(instruction_ref ins) const
...
-}
-
-void program::annotate(std::ostream& os, std::function<void(instruction_ref)> a) const
-{
-    print_program(*this, [&](auto ins, const auto& names) {
-        print_instruction(os, ins, names);
-        a(ins);
-        os << std::endl;
-    });
+    assert(contains(impl->modules, "main"));
```
pfultz2, line 461: I dont think this assert is necessary.

src/program.cpp
```diff
@@ -760,176 +419,66 @@ void program::debug_print(instruction_ref ins) const
...
 
 program& program::sort()
 {
-    fix([&](auto self, auto ins) {
-        this->move_instruction(ins, this->begin());
-        for(auto child : ins->inputs())
-            self(child);
-    })(std::prev(this->end()));
-    assert(this->validate() == this->end());
+    this->get_main_module()->sort();
```
pfultz2, line 467: This should sort all modules.

src/include/migraphx/generic_eval.hpp
```diff
@@ -0,0 +1,123 @@
...
+                           ins->inputs().end(),
+                           [&](auto arg) { return m.has_instruction(arg); }) &&
+               "PRINT_MODULE: Instruction not found");
+
+        print_func(ins, names);
+    }
+}
+
+template <class F>
+std::vector<argument> generic_eval(const module& p,
```
pfultz2, line 55: These functions should be moved to the module.cpp file. The module class should have sufficient methods that the program can call into.

src/program.cpp
scxiao: If a program does not have a has_instruction(), how can we know whether a program contains an instruction? Use the approach of going through all modules to do the check?

src/program.cpp
scxiao: I think a program actually contains a begin and end instruction, which is the begin() and end() of the main module. Do you mean we always need to use that of the main module?

src/program.cpp
scxiao: My previous change is like the above and I can change it back

scxiao: Should we keep the generic_eval() in the program.cpp? It is called in a couple of places.

src/program.cpp
scxiao: fixed

src/program.cpp
pfultz2: Ah sorry, yea just leave it as it is and move generic_eval into program.cpp.

src/program.cpp
pfultz2: Do we call `has_instruction` on a program? I think we just use this over modules. If we do need program-wide `has_instruction` then it should go through all of the modules.

src/program.cpp
pfultz2: The program is a collection of modules(so begin and end should iterate over modules, but I dont think we need that) and the module is a collection of instructions.

> Do you mean we always need to use that of the main module?

Yes we should use begin and end from the module.

src/program.cpp
pfultz2: Yea this can stay.

src/program.cpp
scxiao: fixed

src/program.cpp
pfultz2: The assignment operator should make a copy of the program.

src/program.cpp
```diff
@@ -68,35 +65,11 @@ static void print_instruction(std::ostream& os,
...
-        for(auto&& arg : ins->inputs())
-        {
-            assert(p.has_instruction(arg) && "Instruction not found");
-            (void)arg;
-        }
-
-        print_func(ins, names);
-    }
+    const auto* mm = p.get_main_module();
+    print_module(*mm, print_func);
```
pfultz2, line 69: This should print all modules.

src/program.cpp
scxiao: removed

src/program.cpp
scxiao: fixed

src/program.cpp
scxiao: fixed

src/program.cpp
scxiao: fixed

src/program.cpp
scxiao: fixed

src/program.cpp
```diff
@@ -564,31 +243,22 @@ value program::to_value() const
...
+    }
     return result;
 }
+
 void program::from_value(const value& v)
 {
     auto version = v.at("version").to<int>();
     if(version != program_file_version)
-        std::cout << "Warning: Version mismatch" << std::endl;
+        std::cout << "Warning: Program version mismatch" << std::endl;
```
pfultz2, line 260: This is the format version not program version. Ideally, we should probably throw an error because we usually increment this when the format changes in a non-backwards compatible way.

src/program.cpp
```diff
@@ -741,7 +390,7 @@ void program::perf_report(std::ostream& os, std::size_t n, parameter_map params)
 void program::debug_print() const { std::cout << *this << std::endl; }
 void program::debug_print(instruction_ref ins) const
```
pfultz2, line 391: I think the `debug_print` with an instruction should be done in the module class. Do we use this function from the program?

pfultz2: We can probably leave it for now.

src/program.cpp
```diff
@@ -741,7 +390,7 @@ void program::perf_report(std::ostream& os, std::size_t n, parameter_map params)
...
 {
-    if(ins == this->end())
+    if(ins == this->impl->modules["main"].end())
```
pfultz2, line 393: I would assume this function(since it in the program) would search all modules for the instruction.

src/program.cpp
scxiao: fixed

pfultz2: This condition can be removed.

src/program.cpp
scxiao: fixed

src/include/migraphx/program.hpp
```diff
@@ -126,22 +74,21 @@ struct program
...
     void debug_print() const;
     void debug_print(instruction_ref ins) const;
-    void debug_print(const std::vector<instruction_ref>& inss) const;
     void print_graph(std::ostream& os, bool brief = false) const;
     void print_cpp(std::ostream& os) const;
 
     void dry_run(parameter_map params) const;
 
-    void annotate(std::ostream& os, std::function<void(instruction_ref)> a) const;
+    void annotate(std::ostream& os, const std::function<void(instruction_ref)>& a) const;
```
pfultz2, line 82: This function in the program should probably pass the current module to the function(via `std::function<void(module* m, instruction_ref)>`.

src/module.cpp
```diff
@@ -0,0 +1,655 @@
...
+
+value module::to_value() const
+{
+    value result;
+    result["version"] = module_file_version;
```
pfultz2, line 400: The module file version should be removed.

src/include/migraphx/generic_eval.hpp
pfultz2: What I said is incorrect here. This function should be moved into program.cpp.

src/program.cpp
scxiao: fixed

src/include/migraphx/generic_eval.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include <algorithm>
+#include <set>
+#include <utility>
+#include <unordered_set>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+template <class F>
+void print_module(const module& m, F print_func)
```
pfultz2, line 26: This should be a method of the module. It should be changed to `void print_module(std::function<void(instruction_ref, std::unordered_map<instruction_ref, std::string>)> print_func)`, rather than using a template.

src/program.cpp
scxiao: removed

src/program.cpp
scxiao: fixed

src/include/migraphx/program.hpp
scxiao: This function calls the module method annotate(), which already have the module information.

src/program.cpp
scxiao: fixed

src/program.cpp
scxiao: fixed

src/module.cpp
scxiao: removed

src/include/migraphx/generic_eval.hpp
scxiao: fixed

src/include/migraphx/generic_eval.hpp
scxiao: moved

src/include/migraphx/program.hpp
```diff
@@ -126,22 +72,26 @@ struct program
 
     void debug_print() const;
     void debug_print(instruction_ref ins) const;
-    void debug_print(const std::vector<instruction_ref>& inss) const;
+    void
+    debug_print(const std::function<void(instruction_ref,
```
pfultz2, line 76: Why do we have this `debug_print` function?

src/module.cpp
```diff
@@ -0,0 +1,740 @@
...
+                os << ", " << cpp_var_name(names.at(input));
+            }
+            os << ");" << std::endl;
+        }
+    });
+}
+
+void module::annotate(std::ostream& os, std::function<void(instruction_ref)> a) const
+{
+    this->debug_print([&](auto ins, const auto& names) {
```
pfultz2, line 710: Debug function should only to be used for debugging. What about renaming it to just `print`?

src/program.cpp
```diff
@@ -741,159 +446,53 @@ void program::perf_report(std::ostream& os, std::size_t n, parameter_map params)
...
-                    os << "[label=" << enclose_name(to_string(ins->get_shape())) << "]";
-                os << ";" << std::endl;
-            }
-        }
-    });
-    os << "}" << std::endl;
-}
 
-static std::string cpp_var_name(const std::string& name)
+void program::debug_print(
```
pfultz2, line 474: This should be named just `print`.

src/module.cpp
```diff
@@ -0,0 +1,740 @@
...
+        os << ")";
+    }
+
+    // skip return instruction shape
+    if(ins->name() != "@return")
+        os << " -> " << ins->get_shape();
+}
+
+template <class F>
+std::vector<argument> generic_eval(const module& p,
```
pfultz2, line 65: This should be removed. Its not used by the module class and it duplicates the same code in program.cpp.

src/module.cpp
scxiao: renamed

src/program.cpp
scxiao: renamed

src/program.cpp
```diff
@@ -113,346 +79,103 @@ program& program::operator=(program p)
...
-
-    // TODO: Should it be an error if the output is empty?
-    if(ins->outputs().empty())
-    {
-        return rep;
-    }
-    // Make a copy of outputs which can be changed when calling replace_argument
-    auto outputs = ins->outputs();
-    for(auto out : outputs)
+    for(auto& modl_pair : p.impl->modules)
```
pfultz2, line 93: Is this loop necessary? Why cant you just write `impl->modules = p.impl->modules`?

test/simplify_algebra_test.cpp
```diff
@@ -222,8 +222,8 @@ TEST_CASE(simplify_mul_conv1)
     mm->add_instruction(pass_op{}, mul);
     EXPECT(conv->outputs().front()->name() == "mul");
     run_pass(p);
-    auto new_conv =
-        std::find_if(p.begin(), p.end(), [](auto&& ins) { return ins.name() == "convolution"; });
+    auto new_conv = std::find_if(
+        mm->begin(), mm->end(), [](auto&& ins) { return ins.name() == "convolution"; });
```
pfultz2, line 226: These tests should just run over a module. There is no need to create a program. Maybe we can fix this at a later time.

test/program_test.cpp
```diff
@@ -27,7 +27,11 @@ migraphx::program create_program()
 TEST_CASE(program_equality)
 {
     migraphx::program x = create_program();
-    migraphx::program y = create_program();
+    migraphx::program y;
+
+    y = x;
```
pfultz2, line 32: Why is this changed?

test/program_test.cpp
```diff
@@ -56,6 +60,38 @@ TEST_CASE(program_default_copy_construct)
...
 }
 
+TEST_CASE(program_print)
+{
+    migraphx::program p = create_program();
+    auto* mm            = p.get_main_module();
+    auto in1            = mm->end();
+
+    // print end instruction
+    p.debug_print(in1);
```
pfultz2, line 70: This should use the module.

test/rewrite_batchnorm_test.cpp
```diff
@@ -108,8 +108,8 @@ TEST_CASE(non_literal)
 
     migraphx::rewrite_batchnorm opt;
     opt.apply(*p2.get_main_module());
-    EXPECT(any_of(p1, &is_batch_norm));
-    EXPECT(none_of(p2, &is_batch_norm));
+    EXPECT(any_of(*p1.get_main_module(), &is_batch_norm));
+    EXPECT(none_of(*p2.get_main_module(), &is_batch_norm));
```
pfultz2, line 112: These tests should just create a module and not a program.

pfultz2: It would be better to fix this at a later time.

test/program_test.cpp
scxiao: I had thought it will cover the lines:
    else if(!impl->modules.empty())
    {
        impl->modules.clear();
    }
But looks like it does not go here.

test/program_test.cpp
scxiao: The purpose here is to cover the program::print() function

src/program.cpp
scxiao: fixed

pfultz2 (approved):

scxiao:
@mvermeulen Looks like all the builds have passed, we can merge it in

PR #680, opened 2020-11-06
Add flag to enable cpu backend

pfultz2:
Cpu backend is disabled by default.


scxiao (approved):

scxiao:
I think all builds pass for this PR, the code coverage issue is caused the CPU target not enabled by default?

pfultz2:
It seems like the coverage is missing from tmp_dir(which was added with hip compilation), but this is not part of this PR.

pfultz2:
I think this can be merged in already.

PR #681, opened 2020-11-06
Try to avoid ambiguous assign in value class

pfultz2:


scxiao (commented):
Maybe we can add a unit test for the code change

kahmed10 (approved):

pfultz2:
> Maybe we can add a unit test for the code change

There already is one. This is an attempt to fix this test failure that happens only on centos.

scxiao (approved):

Issue #686, opened 2020-11-11
Add default dim override in driver


kahmed10:
We currently have a `map_input_dims` map in our onnx parser that allows for overriding the default dims (regardless of parameterized value or not), but there is no way to access it in the driver.

scxiao:
code merge and close issue

PR #687, opened 2020-11-12
Driver dim override

scxiao:
For issue #686


src/driver/main.cpp
```diff
@@ -59,6 +60,10 @@ struct loader
            ap.set_value(true));
         ap(is_nhwc, {"--nchw"}, ap.help("Treat tensorflow format as nchw"), ap.set_value(false));
         ap(trim, {"--trim", "-t"}, ap.help("Trim instructions from the end"));
+        ap(param_dims,
+           {"--dim_val"},
+           ap.help("Dim of a parameter (format: \"name:{d1,d2,...,dn}\" (no space in the string))"),
```
kahmed10, line 65: can we change the braces to either brackets or remove them? Since there is an issue with parsing the args otherwise

src/driver/main.cpp
```diff
@@ -59,6 +60,10 @@ struct loader
            ap.set_value(true));
         ap(is_nhwc, {"--nchw"}, ap.help("Treat tensorflow format as nchw"), ap.set_value(false));
         ap(trim, {"--trim", "-t"}, ap.help("Trim instructions from the end"));
+        ap(param_dims,
+           {"--dim_val"},
```
pfultz2, line 64: This should be `--dim-val`

src/driver/main.cpp
pfultz2: The format should be something like `@name d1 d2 d3`. The `@` symbol is to distinguish between an input parameter and a dimension. And then you can convert it to the map with:

```cpp
std::unordered_map<std::string, std::vector<std::size_t>> map_input_dims;
std::string name = "";
for(auto&& x:param_dims)
{
    if (x[0] == '@')
        name = x.substr(1);
    else
        map_input_dims[name].push_back(value_parser<std::size_t>::apply(x));
}
```

src/driver/main.cpp
pfultz2: Maybe this should be named `--input-dim` to avoid confusion with dim parameters when we add them later.

src/driver/main.cpp
scxiao: The format @name d1 d2 d3 will be handled as multiple command line arguments, and we have no idea of how many arguments should be included for the --input-dim. Looks like even the format with quote "@name d1 d2 d3" does not work here either.

pfultz2: > The format @name d1 d2 d3 will be handled as multiple command line arguments

Yes.

> we have no idea of how many arguments should be included for the --input-dim.

We know it will stop when another argument is passed. 

> Looks like even the format with quote "@name d1 d2 d3" does not work here either.

I think we need to add `ap.append()` so it will append multiple arguments together.

scxiao: We have the ap.append(), see line 66. is there other usage of the ap.append()?

scxiao: I add the function call ap.nargs(2), and the above format works now. 
But one thing to note is, previously, the following command line works:
./driver perf --input-dim @name 1 2 3 file.onnx

but it does NOT works any more. Then reason is, in the above command, file.onnx is considered as part of the --input-dim options. so we have to use the following command instead:
./driver perf file.onnx --input-dim @name 1 2 3

pfultz2 (approved):

kahmed10 (approved):

PR #688, opened 2020-11-16
Use dnnl for cpu backend

pfultz2:


scxiao (approved):

src/include/migraphx/op/contiguous.hpp
```diff
@@ -44,6 +44,11 @@ struct contiguous
         });
         return result;
     }
+
```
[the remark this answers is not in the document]
pfultz2: Thats a good idea.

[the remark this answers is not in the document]
pfultz2: No we cant get rid of compute as this doesn't inherit from `unary`. There seems to be an issue with using `unary` here. I am not sure what the issue, we can look into this in the future.

pfultz2:
This is passing CI now. @kahmed10 can you review?

kahmed10 (commented):
There's an empty file for test/verify/main.cpp? Or is this a github bug?

pfultz2:
> There's an empty file for test/verify/main.cpp? Or is this a github bug?

Its not an empty file(as you can see [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/cpu-target/test/verify/main.cpp)) but it means there are no changes to the file. I am not sure why it shows if there are no changes to the file.

src/targets/cpu/include/migraphx/cpu/context.hpp
```diff
@@ -2,15 +2,65 @@
...
 
+#ifdef USE_DNNL
 struct context
 {
     void finish() const {}
+
+    template <class F>
+    void bulk_execute(std::size_t n, std::size_t min_grain, F f)
+    {
+#if 1
```
kahmed10, line 21: Why do we need a macro here?

pfultz2: Yea thats debug code that can be removed.

pfultz2: Fixed.

kahmed10 (approved):

PR #692, opened 2020-12-01
Conv weight fix

kahmed10:
Includes changes to broadcasting a scalar.


src/include/migraphx/op/broadcast.hpp
```diff
@@ -49,8 +49,6 @@ struct broadcast
         if(std::all_of(
                broadcast_lens.cbegin(), broadcast_lens.cend(), [&](auto x) { return x == 1; }))
         {
-            if(axis != 0)
-                MIGRAPHX_THROW("BROADCAST: when broadcasting tensor of size 1, axis should be 0");
```
pfultz2, line 53: Can you add a unit test for this?

kahmed10: We have unit tests for this if statement under `op_shape_test.cpp`. I also added a scalar test for broadcast in tf.

pfultz2: There isnt a test for non zero axis in op_shape_test.cpp, only for when axis is 0.

kahmed10: Added

pfultz2 (approved):

test/op_shape_test.cpp
```diff
@@ -346,6 +346,15 @@ TEST_CASE(broadcast)
...
+        std::vector<std::size_t> lens{1, 1};
+        migraphx::shape input{migraphx::shape::float_type, {1}, {0}};
+        expect_shape(migraphx::shape{migraphx::shape::float_type, {1, 1}, {0, 0}},
+                     migraphx::op::broadcast{1, lens},
+                     input);
+    }
+
     {
         std::vector<std::size_t> lens{1, 1};
         migraphx::shape input{migraphx::shape::float_type, {4, 1, 3}};
```
pfultz2, line 360: This test should be removed as it fails, since it no longer throws.

test/tf/gen_tf_pb.py
```diff
@@ -145,6 +145,14 @@ def biasadd_test(g1):
         tf.nn.bias_add(g1_input, g2_input, name='bias_add1')
 
 
+@tf_test
+def biasadd_scalar_test(g1):
```
pfultz2, line 149: I think you forgot to commit these new files.

test/op_shape_test.cpp
kahmed10: removed

test/tf/gen_tf_pb.py
kahmed10: added

src/include/migraphx/op/broadcast.hpp
scxiao: We may need to add a check that input either is a scalar or contains only one element? Then the shape unit test there can still pass.

pfultz2:
Can you merge this in first since it fixes a serious bug? And then we can open an issue to improve error checking in the broadcast operators?

mvermeulen:
@scxiao  is this ok to approve?

scxiao:
> @scxiao is this ok to approve?

If we use another PR for the checking. I am fine to approve this one.

scxiao (approved):

Issue #691, opened 2020-12-01
slim resnet50 fails to generate perf report: same_dims: add: Dimensions do not match


mvermeulen:
Running weekly performance on"
```
commit e09d54fe2311ff4c4d11a46678660adb7351a204
```
Results in the following error
```
root@paramaribo:/src/AMDMIGraphX/build/bin# ./driver perf /root/resnet50v2_i1.pb 
Compiling ... 
Reading: /root/resnet50v2_i1.pb
terminate called after throwing an instance of 'migraphx::version_1::exception'
  what():  /src/AMDMIGraphX/src/include/migraphx/check_shapes.hpp:122: same_dims: add: Dimensions do not match
Aborted (core dumped)
```
A "driver read" does work without error and I have attached results of both these commands
[resnet50.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/5620025/resnet50.txt)

I have also pushed a docker container including the TF protobuf saved in /root as the following:
```
mevermeulen/rocm-migraphx:resnet50-2020-11-30
```
To repeat
```
cd /src/AMDMIGraphX/build/bin
./driver perf /root/resnet50v2_i1.pb
```
This problem was not there on the build of MIGraphX on November 16th, but don't know the precise commit where things broke.

mvermeulen:
The following commit is OK:
```
commit 8443ecd17a562137c79e4e531455eee989d85a44
Author: Shucai Xiao <shucai.xiao@amd.com>
Date:   Mon Nov 16 15:30:05 2020 -0600
```
The following commit is broken:
```
commit e09d54fe2311ff4c4d11a46678660adb7351a204
Author: kahmed10 <15948690+kahmed10@users.noreply.github.com>
Date:   Wed Nov 25 20:28:18 2020 -0600
```
Still narrowing to specific commit that broke this.

kahmed10:
It looks to be an issue with `eliminate_pad.cpp` when it was refactored. When disabling this pass, the program works @pfultz2

mvermeulen:
Replacing eliminate_pad.cpp and eliminate_pad.hpp with version prior to  51acbdef  results in this passing.

scxiao:
It is related to the following change:
-    T op       = any_cast<T>(ins->get_operator());
-    op.padding = new_pads;
+    auto op = ins->get_operator();
+    op.from_value({"padding", new_pads});

[...]

PR #693, opened 2020-12-01
Eliminate pad discard attr

scxiao:
for issue #691


kahmed10:
We can probably add a test to `test/eliminate_pad.cpp` that creates a conv op with different strides. That would probably catch the issue found in #691

pfultz2:
#694 fixes this issue, and improves the testing.

PR #696, opened 2020-12-04
Refactor to use make_op almost everywhere

pfultz2:


src/onnx/onnx.cpp
```diff
@@ -1660,7 +1690,7 @@ struct onnx_parser
             auto&& perm_vals = info.attributes["perm"].ints();
             perm             = std::vector<int64_t>(perm_vals.begin(), perm_vals.end());
         }
-        return mm->add_instruction(migraphx::op::transpose{perm}, args.front());
```
kahmed10, line 1663: for this and L1724, I don't think we need to scope to migraphx when calling make_op (looks to be carried forward from existing code, which doesn't need it either)

pfultz2: Actually, I am refactoring the onnx parser, and removed the extra `migraphx::` there.

kahmed10 (approved):

scxiao:
@pfultz2 I ran this PR with the MSFT models and no regression is observed.

scxiao (approved):

PR #699, opened 2020-12-08
Refactor onnx parser

pfultz2:


src/onnx/include/migraphx/onnx/onnx_parser.hpp
```diff
@@ -0,0 +1,82 @@
...
+            return add_instruction(op, {xs...});
+        }
+    };
+    using node_map = std::unordered_map<std::string, onnx::NodeProto>;
+    using op_func  = std::function<std::vector<instruction_ref>(
+        const onnx_parser&, const node_info&, std::vector<instruction_ref>)>;
+    node_map nodes;
+    std::unordered_map<std::string, instruction_ref> instructions;
+    program prog                  = program();
+    bool is_pytorch               = false;
```
kahmed10, line 53: looking at the code before this change, I can't seem to find where this variable is used. If that's the case, can we remove?

kahmed10 (approved):

src/onnx/include/migraphx/onnx/onnx_parser.hpp
pfultz2: Yea I removed it.

src/onnx/parse_onehot.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+        std::vector<float> depth_input(depth * depth, 0.0f);
+        for(int i = 0; i < depth; i++)
+        {
+            depth_input[depth * i + i] = 1.0f;
+        }
+
+        auto type = args[2]->get_shape().type();
+        shape s{type, {depth, depth}};
+        auto l_val      = info.mm->add_literal({s, depth_input});
```
scxiao, line 38: I am wondering whether we should add a method add_literal to the node_info struct, then the add_literal() can be done like info.add_literal()?

src/onnx/parse_pad.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+        // compute slice indices in a periodic fashion
+        calc_reflect_indices(l_indices, *dims_it);
+        calc_reflect_indices(r_indices, *dims_it);
+
+        for(int idx : l_indices)
+        {
+            *starts_it = idx;
+            *ends_it   = *starts_it + 1;
+            slices.push_back(info.mm->add_instruction(
```
scxiao, line 73: Can we just call info.add_instruction() here?

src/onnx/parse_onehot.cpp
pfultz2: Yea, I added this.

src/onnx/parse_pad.cpp
pfultz2: Yea, done.

scxiao (approved):

PR #703, opened 2020-12-15
Add Quantize and Dequantize Linear op

kahmed10:


kahmed10:
There should be no change to src/onnx/onnx.cpp, it looks like a glitch on github's end.

pfultz2 (approved):

src/onnx/parse_dequantizelinear.cpp
```diff
@@ -0,0 +1,49 @@
...
+            axis = info.attributes.at("axis").i();
+
+        auto input_lens = args[0]->get_shape().lens();
+
+        auto sub_zero_point = args[0];
+
+        if(args.size() == 3)
+        {
+            auto zero_point = args[2];
+            if(axis != 1)
```
scxiao, line 30: Looks like the broadcasting here has some issues.
In my understanding, if it is a scalar shape, we can use multibroadcast; 
if it is a 1D tensor, it is a broadcast as make_op("broadcast", {{"axis", axis}, {"dims", input_lens}}) (not dependent on the axis value).

src/onnx/parse_dequantizelinear.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& /*parser*/,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        int axis = 1;
+        if(contains(info.attributes, "axis"))
+            axis = info.attributes.at("axis").i();
+
```
scxiao, line 22: We should be able to handle the case of axis to be minus value as:
if (axis < 0) axis += n_rank;

src/onnx/parse_quantizelinear.cpp
```diff
@@ -0,0 +1,52 @@
...
+struct parse_quantizelinear : op_parser<parse_quantizelinear>
+{
+    std::vector<op_desc> operators() const { return {{"QuantizeLinear"}}; }
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& /*parser*/,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        auto quant_type = shape::uint8_type;
```
scxiao, line 19: How do we know the default output type to be uint8_type?

test/onnx/gen_onnx.py
```diff
@@ -1016,6 +1016,38 @@ def deconv_stride_test():
...
+    arg1 = helper.make_tensor_value_info('1', TensorProto.FLOAT, [5])
+    arg2 = helper.make_tensor_value_info('2', TensorProto.INT8, [5])
+    arg_out = helper.make_tensor_value_info('out', TensorProto.FLOAT,
+                                            [1, 1, 5, 1])
+
+    node = onnx.helper.make_node('DequantizeLinear',
+                                 inputs=['0', '1', '2'],
+                                 outputs=['out'],
+                                 axis=2)
+
```
scxiao, line 1047: We can add a test case of the axis to be negative values.

scxiao (commented):
You can turned on the onnx unit tests in the onnx_backend_test.py by adding the two lines:
        backend_test.include(r'.*test_dequantizelinear*')
        backend_test.include(r'.*test_quantizelinear*')

Some of the unit tests failed for now. They are supposed to pass if our implementation is correct.

src/onnx/parse_quantizelinear.cpp
kahmed10: this is from the onnx spec which defaults the input to be uint_8

src/onnx/parse_dequantizelinear.cpp
scxiao: Another way for the broadcasting can be as:
For scalar shape, just do multibroadcasting, this is straightforward.
For the 1D case, we can first do a reshape as for the 1D tensor as {1, 1, (axis_dim use the dim of the 1D-tensor), 1), then do a multibroadcasting....

src/onnx/parse_dequantizelinear.cpp
kahmed10: fixed

test/onnx/gen_onnx.py
kahmed10: added

scxiao (approved):

src/include/migraphx/tune_axis.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <utility>
+#include <cstdint>
+#include <string>
+#include <migraphx/errors.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+inline int tune_axis(const int n_dim, const int axis, std::string op_name)
```
pfultz2: Can we make `op_name` have default value?

kahmed10: added

turneram (approved):

mvermeulen:
Is there a reason some of the required items still seem to be pending?  Will merge once they pass.

kahmed10:
Everything else was passing on the previous CI run, but for some reason github wasn't updating to reflect that. I just re-ran the job to see if it fixes the issue.

PR #704, opened 2020-12-16
Add missing has_op function

pfultz2:


kahmed10 (approved):

scxiao (approved):

scxiao:
With this PR, looks like the build and test both passed.

I did the following to pass the build.
1) Install the openmp as:
sudo apt install libomp-dev

2) added the following line to the requirements.txt file as:
oneapi-src/oneDNN@v2.0 -DDNNL_CPU_RUNTIME=OMP

Then the CPU target can be built.

PR #706, opened 2020-12-19
Revamp CI infrastucture

pfultz2:
* Add github actions to handle some of our CI jobs:
    - Build jobs for tidy, cppcheck, format, and CPU unit tests
    - Use docker layer caching to speed up building dockerfile
    - Use ccache to speed up compilation
    - Update rocm-cmake version to use clang-tidy cache to speed up analysis
* Update docker file to use hip-clang by default:
    - Use hip-clang for most jobs on jenkins
    - Update readme to point to hip clang
* Add a hcc.docker file for hcc builds. Only one jenkins job is run with hcc for running with UBSAN
* Update cppcheck version, as it can do analysis faster

Also, there is still is a hip-clang.docker file that is used by github actions, its almost the same version as Dockerfile, but skips installing the things needed for onnx runtime and onnx models. We can try to merge them together in the future and use build args to disable the steps for github runners.

This also doesn't show passing the required checks as this needs to be updated as the checks are different now. I'll update the required checks as it gets reviewed.


hcc.docker
```diff
@@ -0,0 +1,66 @@
...
+RUN pip3 install -r /doc-requirements.txt
+
+# Install dependencies
+ADD dev-requirements.txt /dev-requirements.txt
+ADD requirements.txt /requirements.txt
+# Manually ignore rocm dependencies
+RUN cget -p $PREFIX ignore \
+    RadeonOpenCompute/clang-ocl \
+    ROCm-Developer-Tools/HIP \
+    ROCmSoftwarePlatform/MIOpenGEMM \
```
scxiao, line 58: I am wondering why not install the miopen and rocblas using "apt install", this can avoid the long time used for building the source code...

Dockerfile
```diff
@@ -75,29 +57,31 @@ RUN pip3 install yapf==0.28.0
...
 RUN chmod +x /download_models.sh && /download_models.sh && rm /download_models.sh
 
+# Install dependencies
+ADD dev-requirements.txt /dev-requirements.txt
+ADD requirements.txt /requirements.txt
+# Manually ignore rocm dependencies
+RUN cget -p $PREFIX ignore \
+    RadeonOpenCompute/clang-ocl \
+    ROCm-Developer-Tools/HIP \
+    ROCmSoftwarePlatform/MIOpen \
```
scxiao, line 74: Same comments as the other places...

hcc.docker
pfultz2: I do that already, which is why I ignore them here.

pfultz2: Its installed by apt [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/706/files/ec060537bbc0a9fd75226193da551f35d0e47fe1#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R35).

scxiao: Sorry, I think I had considered the miopendriver as the miopen when I try this docker file.

scxiao:
I just tried the two docker files hcc.docker and hip-clang.docker, they work file.

scxiao (approved):

kahmed10 (approved):

pfultz2:
This can be merged now.

Issue #559, opened 2020-06-22
Pow operator enhancement


scxiao:
Pow operator needs to support base and exponents of different types, but we cannot support for now. need enhancement.

scxiao:
fixed and close

PR #707, opened 2021-01-06
Pow different data type

scxiao:
#559 This PR is to handle the case of different input data type for the pow operator


src/onnx/parse_pow.cpp
```diff
@@ -0,0 +1,75 @@
...
+        {
+            args[0] = info.add_instruction(make_op("convert", {{"target_type", tc}}), args[0]);
+        }
+
+        if(tc != te)
+        {
+            args[1] = info.add_instruction(make_op("convert", {{"target_type", tc}}), args[1]);
+        }
+
+        // return info.add_broadcastable_binary_op("pow", args[0], args[1]);
```
pfultz2, line 62: This comment can be removed.

pfultz2 (approved):

turneram (approved):

src/onnx/parse_pow.cpp
```diff
@@ -0,0 +1,74 @@
...
+struct parse_pow : op_parser<parse_pow>
+{
+    std::vector<op_desc> operators() const { return {{"Pow"}}; }
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& /*parser*/,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        auto tb = args[0]->get_shape().type();
```
kahmed10, line 48: I would prefer to be explicit with the variable names so that it's clear (or leave a comment in the code with the general formula). For example, maybe `tb = type_base, tc = type_exponent, tc = type_computed`

scxiao: fixed, changed to type_fullname

kahmed10 (approved):

src/onnx/parse_pow.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <migraphx/ranges.hpp>
+#include <migraphx/make_op.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace onnx {
+
+auto compute_type(shape::type_t t1, shape::type_t t2)
+{
+    const static std::unordered_map<shape::type_t, int> op_order = {{shape::int8_type, 1},
```
pfultz2, line 12: It looks `shape::type_t` cannot be used as a key since its not hashable. You should just use an int as the key and then convert the enum to an int with a `static_cast`.

scxiao: Thanks, just changed that.

PR #716, opened 2021-01-15
And_op

scxiao:
When analyzing the mlperf_ssd_resnet model, I added this operator so only loop operator is missed for the model.


kahmed10:
Can we call it `andl` (logical and) instead of `andd`?

pfultz2:
> Can we call it andl (logical and) instead of andd?

I think it would be better to call it `logical_and` and not use cryptic names.

scxiao:
> > Can we call it andl (logical and) instead of andd?
> 
> I think it would be better to call it `logical_and` and not use cryptic names.

Thanks. Actually, this is one thing that I need suggestions from you guys. I will use the logical_and as the name.

PR #717, opened 2021-01-16
Add binary type to serialization

pfultz2:
This will help serialization to json. I dont use json's binary type since it can't easily be roundtripped. Instead, I create an object with `bytes` and if we see this when parsing then it treats it as binary instead of an object.


src/include/migraphx/value.hpp
```diff
@@ -139,6 +140,19 @@ struct value
     using const_pointer   = const value_type*;
     using array           = std::vector<value>;
     using object          = std::unordered_map<std::string, value>;
+    struct binary : std::vector<std::uint8_t>
+    {
+        using base = std::vector<std::uint8_t>;
+        binary() {}
+        template <class Container>
+        explicit binary(const Container& c) : base(c.begin(), c.end())
```
scxiao, line 148: should be restrict the sizeof(c) == 1 here? otherwise, we will lose some data for the conversion.

PR #723, opened 2021-01-21
Normalize compute methods

pfultz2:
There is a normalized flags set to avoid running normalization everytime during `eval`.


scxiao (commented):
I think there are a few other operators also needing changes like gather, squeeze, and unsqueeze. Whatever operator having a normalize_compute_shape() needs corresponding changes.

pfultz2:
> I think there are a few other operators also needing changes like gather, squeeze, and unsqueeze

There is no change needed for these. Squeeze and unsqueeze just use the output_shape. In Gather, normalizations fixes the `axis` field, but it doesn't fix the indices that get passed in, so this can't be changed.

scxiao (approved):

kahmed10 (approved):

Issue #721, opened 2021-01-21
Unknown operator: Not


cagery:
Shows up at ***roberta-base-11.onnx*** from ONNX Model Zoo when _cpp_parse_load_save app_ is used to parse:
https://github.com/onnx/models/blob/master/text/machine_comprehension/roberta/model/roberta-base-11.onnx
```
 ./parse_load_save /cgMIGraphX/AMDMIGraphX_v1/examples/cpp_parse_load_save/models/text/machine_comprehension/roberta/model/roberta-base-11.onnx --parse onnx
Parsing ONNX File
MIGraphX Error: /cgMIGraphX/not_operator/AMDMIGraphX/src/onnx/onnx_parser.cpp:281: parse_graph: Unknown operator: Not
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to call function
Aborted (core dumped)
```

mvermeulen:
Merged

PR #735, opened 2021-01-29
changes for not operator

cagery:
This PR introduces unary NOT operator.

The problem is identified [in this issue.](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/721)

Compilation passed, and re-testing the issue is not showing the error any longer.


pfultz2:
This is a good start. Tests need to be added. We usually add a test in refs_ops_test.cpp. You can follow the example for the [`logical_and_test`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/ref_ops_test.cpp#L1825) test. We also need a test for onnx parsing, which we usually add this in onxx_test.cpp, but also we update gen_onnx.py so it can generate the onnx files. You can see the [`logical_or_test`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/onnx/onnx_test.cpp#L1618) as an example.

src/include/migraphx/op/unary_not.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <migraphx/operation.hpp>
+#include <migraphx/check_shapes.hpp>
+#include <migraphx/config.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct unary_not : unary<unary_not>
+{
```
pfultz2, line 14: Even though the class has to be `unary_not` so it won't conflict with C++ keywords, maybe we should still have the operator name as `not`, which could be done by adding a `name` method to the class(ie `std::string name() const { return "not"; }`.

cagery:
Thanks @pfultz2 ! I will look into them and get back soon.

test/onnx/onnx_test.cpp
```diff
@@ -1657,6 +1658,20 @@ TEST_CASE(logsoftmax_test)
     EXPECT(p == prog);
 }
 
+TEST_CASE(not_test)
```
kahmed10, line 1661: can you move this test to be in alphabetical order? Just to keep consistency for the gen_onnx and onnx_test files

kahmed10 (changes requested):

test/onnx/onnx_test.cpp
kahmed10: otherwise lgtm

cagery: Moved it, thanks!

test/onnx/gen_onnx.py
```diff
@@ -2013,6 +2012,14 @@ def neg_test():
 
     return ([node], [x], [y])
 
+@onnx_test
+def not_test():
+    x = helper.make_tensor_value_info('0', TensorProto.INT32, [4])
+    y = helper.make_tensor_value_info('1', TensorProto.INT32, [4])
```
scxiao, line 2018: The input here should be BOOL type

test/onnx/onnx_test.cpp
```diff
@@ -1914,6 +1915,19 @@ TEST_CASE(nonzero_int_test)
...
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    auto l0  = mm->add_parameter("0", migraphx::shape{migraphx::shape::int32_type, {4}});
```
scxiao, line 1922: again, the input here should be bool_type

scxiao (commented):
Looks like the GPU implementation is missed.

test/onnx/gen_onnx.py
cagery: Why does it have to be BOOL? Unary not should work on any data type, what am I missing?

cagery:
@scxiao , how do we implement an operator for GPU? Is it /src/targets/gpu ? Can you provide me a pointer from an existing op?

test/onnx/gen_onnx.py
scxiao: That is specified by the onnx operator. You can keep this test and one for the BOOL type.

src/include/migraphx/op/unary_not.hpp
scxiao: > @scxiao , how do we implement an operator for GPU? Is it /src/targets/gpu ? Can you provide me a pointer from an existing op?

Please refer to the logical_and operator.

scxiao:
> @scxiao , how do we implement an operator for GPU? Is it /src/targets/gpu ? Can you provide me a pointer from an existing op?

You can refer the logical_and operator

test/onnx/gen_onnx.py
cagery: Got it! Thanks Shucai.

cagery:
@scxiao gpu implementation and the bool tests are turned in, can you please take a look whenever you get a chance?

scxiao:
The gpu code is fine. I think we need two other changes: 
1. change the gpu/lowering.cpp file to lowering the not operator
2. add a gpu unit test in the folder test/verify for the not operator (since we do not have this unit test, changes mentioned in item 1 is not surfaced.)

cagery:
I have added the changes required for GPU. Compilation passes, and tests are passes as well. It should be fine now but please take another look whenever you get a chance. @scxiao @pfultz2 
```
[   RUN    ] test_not
[ COMPLETE ] test_not
```

```
[   RUN    ] not_test
[ COMPLETE ] not_test
[   RUN    ] not_bool_test
[ COMPLETE ] not_bool_test
```

**Two questions:**
1) Is there a way to generate the onnx test file automatically? 
I have to do the followings to generate onnx files, otherwise tests fail. It'd be better if I could automate this, maybe during compilation?:
```
python3 -c "import gen_onnx; gen_onnx.not_bool_test()"
python3 -c "import gen_onnx; gen_onnx.not_test()"
```

2) For `not_test` and `not_bool_test`, we set the golden expected results. How does that work for `test_not` under /verify/ ? I see all the tests under /verify/ are returning the migraphx::program only. Does it mean that we are using these cpp files to make sure such operators can be compiled on a target?

kahmed10:
You need to run the python formatter yapf (same version as docker container) on the gen_onnx.py file. Then the format test will pass.

scxiao:
> I have added the changes required for GPU. Compilation passes, and tests are passes as well. It should be fine now but please take another look whenever you get a chance. @scxiao @pfultz2
> 
> ```
> [   RUN    ] test_not
> [ COMPLETE ] test_not
> ```
> 
> ```
> [   RUN    ] not_test
> [ COMPLETE ] not_test
> [   RUN    ] not_bool_test
> [ COMPLETE ] not_bool_test
> ```
> 
> **Two questions:**
> 
>     1. ~Is there a way to generate the onnx test file automatically?
>        I have to do the followings to generate onnx files, otherwise tests fail. It'd be better if I could automate this, maybe during compilation?:~ _Added the onnx test files._
> 
> 
> ```
> python3 -c "import gen_onnx; gen_onnx.not_bool_test()"
> python3 -c "import gen_onnx; gen_onnx.not_test()"
> ```
> 
>     1. For `not_test` and `not_bool_test`, we set the golden expected results. How does that work for `test_not` under /verify/ ? I see all the tests under /verify/ are returning the migraphx::program only. Does it mean that we are using these cpp files to make sure such operators can be compiled on a target?

For the verify test, the return program will run on both the GPU and the REF target, then compare the two output to ensure they are the same.

scxiao (approved):

pfultz2 (approved):

PR #740, opened 2021-02-02
Log softmax nonstd input shape

scxiao:


src/include/migraphx/op/logsoftmax.hpp
```diff
@@ -30,8 +30,9 @@ struct logsoftmax
     std::string name() const { return "logsoftmax"; }
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(1).standard();
-        return inputs.at(0);
+        check_shapes{inputs, *this}.has(1);
+        auto lens = inputs.at(0).lens();
+        return {inputs.at(0).type(), lens};
```
pfultz2, line 35: If the shape is packed it should return `inputs.at(0)`.

scxiao: looks like transpose is not the case.

pfultz2: If the input is transposed then the output should maintain the same layout(there are other operators(such as convolution) that needs updated as well).

scxiao: We may need a pass to handle this case. Because of the transposed shape, the axis value in the onnx file need to change to the axis before the transpose.

pfultz2: I dont understand why the axis needs to be changed since its completely independent of the memory layout.

pfultz2: When we use other layouts such as NHWC, we will transpose it to NCHW. Our transpose operator doesn't change memory layout only the view to the memory so it still will be an NHWC memory layout.

The axis for softmax will be 1 for NCHW since it reduces over C. Even if the memory layout is NHWC the axis will still be `1` since the order of the axes are NCHW.

pfultz2 (approved):

kahmed10 (approved):

scxiao:
This PR can be merged now.

PR #741, opened 2021-02-03
Module operations

scxiao:
This PR is for a few operations related to modules:
1) Create and remove submodules
2) Print program containing submodules
3) Used submodules as instruction inputs

Compilation and execution related operations are NOT included.


src/module.cpp
```diff
@@ -241,6 +322,26 @@ instruction_ref module::move_instructions(instruction_ref src, instruction_ref d
...
+    this->impl->sub_modules.push_back(module(name));
+    module_ref sub_mdl = &this->impl->sub_modules.back();
+    sub_mdl->set_parent_module(this);
+
+    return sub_mdl;
+}
+
+std::vector<module_ref> module::get_sub_modules() const
+{
+    std::vector<module_ref> sub_modules(this->impl->sub_modules.size());
```
pfultz2, line 336: This function should create submodules from the instructions:

```cpp
    std::vector<module_ref> result;
    for(auto ins: *this)
        result.insert(result.end(), ins->module_inputs().begin(), ins->module_inputs().end());
    return result;
```

src/module.cpp
```diff
@@ -24,7 +24,9 @@ struct module_impl
 {
     // A list is used to keep references to an instruction stable
     std::list<instruction> instructions;
+    std::list<module> sub_modules;
```
pfultz2, line 27: The module should not own the submodules, as these can be referenced across multiple modules. The modules should be owned by the program class.

src/module.cpp
```diff
@@ -113,6 +125,28 @@ void module::assign(const module& m)
         }
         else
         {
+            // if there are sub_module inputs, need to make a copy of the submodule
+            auto arg_modules = ins->module_inputs();
```
pfultz2, line 129: The modules should only be replaced when copying the program not when copying just the module. When copying a module class it should still reference the same modules. We should add a `module_ref replace_module(module_ref mod,  module_ref rep)` method to the module class, so then the program will make a copy of all modules and then replace the module_refs with the copied modules:

```cpp
void program::assign(const program& p)
{
    if(!impl)
    {
        impl = std::make_unique<program_impl>();
    }
    else if(!impl->modules.empty())
    {
        impl->modules.clear();
    }
    impl->ctx         = p.impl->ctx;
    impl->target_name = p.impl->target_name;
    impl->modules     = p.impl->modules;
    for(auto&& mp:impl->modules) 
    {
        auto&& mod = mp.second;
        for(auto* old_mod:mod.get_sub_modules())
        {
            mod.replace_module(old_mod, &impl->modules.at(old_mod->name()));
        }
    }
}
```

src/include/migraphx/module.hpp
```diff
@@ -126,14 +147,17 @@ struct module
...
     friend std::ostream& operator<<(std::ostream& os, const module& m);
     friend bool operator==(const module& x, const module& y);
     friend bool operator!=(const module& x, const module& y) { return !(x == y); }
 
     private:
-    void assign(const module& m);
+    void assign(const module& m, std::unordered_map<instruction_ref, instruction_ref> ins_map);
     std::unique_ptr<module_impl> impl;
-    std::string module_name;
+    module_ref parent_mdl = nullptr;
```
pfultz2, line 160: Why do we need to store a pointer to the parent module?

src/program.cpp
```diff
@@ -89,7 +89,16 @@ void program::assign(const program& p)
     }
     impl->ctx         = p.impl->ctx;
     impl->target_name = p.impl->target_name;
-    impl->modules     = p.impl->modules;
```
pfultz2, line 92: Why is this line removed?

src/instruction.cpp
```diff
@@ -1,6 +1,7 @@
 #include <migraphx/instruction.hpp>
 #include <migraphx/builtin.hpp>
 #include <migraphx/erase.hpp>
+#include <migraphx/module.hpp>
```
pfultz2, line 4: Can the module class be forwarded declared?

src/instruction.cpp
```diff
@@ -82,7 +95,15 @@ bool instruction::valid() const
...
         try
         {
-            computed = compute_shape(op, arguments);
+            if(module_args.empty())
+            {
+                computed = compute_shape(op, arguments);
+            }
+            else
+            {
+                auto out_shapes = compute_shape(module_args[0]);
```
pfultz2, line 104: The output shape won't always be output shape of the module. I think the operator class will need a compute_shape that takes arguments and modules as inputs(`compute_shape(op, arguments, module_args)`).

src/module.cpp
```diff
@@ -120,6 +122,27 @@ void module::assign(const module& m)
         }
         else
         {
+            // if there are sub_module inputs, need to make a copy of the submodule
```
pfultz2, line 125: The module class should not copy the submodules. When copying a module the module_refs should be the same.

pfultz2:
Can you add a unit test like this?

```cpp
TEST_CASE(submodule_copy)
{
    module mm("main");
    auto x = mm.add_parameter("x", {migraphx::shape::int64_type});
    module sm("sub");
    sm.add_instruction(pass_op{}, x);
    mm.add_instruction(pass_op{}, {x}, {sm});

    auto mm2 = mm1;

    EXPECT(mm == mm2);
    EXPECT(mm.get_sub_modules() == mm2.get_sub_modules());
}
```

This will make sure the module copy works.

src/include/migraphx/op/iff.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <migraphx/functional.hpp>
+#include <migraphx/config.hpp>
+#include <cmath>
+#include <utility>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct iff
```
pfultz2, line 16: I think the class should be name `if_op`.

src/include/migraphx/op/iff.hpp
```diff
@@ -0,0 +1,37 @@
...
+    std::string else_sub_graph;
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.then_sub_graph, "then_sub_graph"),
+                    f(self.else_sub_graph, "else_sub_graph"));
+    }
+
+    std::string name() const { return "iff"; }
```
pfultz2, line 28: The name should be `if`.

src/include/migraphx/op/iff.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <utility>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct iff
+{
+    std::string then_sub_graph;
+    std::string else_sub_graph;
```
pfultz2, line 19: Why are the subgraphs strings? I dont think the operator would store the subgraph since that is already stored in the instruction.

pfultz2:
Maybe we should old off adding the `if` operator since its not complete yet. We can still test submodules by extending the `pass_op` to support module_args(and then the compute method would just pass the first input and ignore the modules).

src/program.cpp
scxiao: We cannot directly copy the modules one by one. The design here considers all modules are of the same level, which dies not consider the lexical scoping issue. Actually, these modules are parent-child relationship in the ONNX model. 

By assigning the modules one after another, it may cause the issue that one submodule needs an input from its parent module, but its parent module is not copies yet, so that input is unavailable.
 
To avoid this problem, we have to start from the main module, then go to its child submodule, and so on..... In this way, when a submodule needs an input from its parent, we can ensure it is available.

src/module.cpp
scxiao: When should we replace the old_mod_ref with the new_mod_ref?

src/program.cpp
scxiao: Looks like we are using a linear data structure to implement something like a tree, implementation of the functions like to_value/from_value/assign are a little weird.

src/module.cpp
pfultz2: The program would replace the refs since it can see all modules.

src/program.cpp
pfultz2: The module class should be copyable, but of course, the program class needs to update the references to the new modules.

I forgot about the instruction_refs that need to be updated. We can add a `replace_refs` method to the instruction class that takes two maps for module_ref and instruction_ref. This would update any reference(in inputs(), outputs(), and module_args) in the instruction class to use the new reference. We can then copy the program with something like this:

```cpp
void program::assign(const program &p)
{
  if (!impl)
    impl = std::make_unique<program_impl>();
  impl->ctx = p.impl->ctx;
  impl->target_name = p.impl->target_name;
  // Copy the modules using its copy contructor
  impl->modules = p.impl->modules;
  // Build a map from old module to new module
  std::unordered_map<module_ref, module_ref> mod_map;
  std::transform(
      impl->modules.begin(), impl->modules.end(), p.impl->modules.begin(),
      std::back_inserter(mod_map),
      [](auto &&x, auto &&y) { return std::make_pair(&x.second, &y.second); });
  // Build a map from old ins to new ins
  std::unordered_map<instruction_ref, instruction_ref> ins_map;
  for (auto &&pp : mod_map)
  {
    auto old_ins = iterator_for(pp.first);
    auto new_ins = iterator_for(pp.second);
    std::transform(old_ins.begin(), old_ins.end(), new_ins.begin(),
                   std::back_inserter(ins_map),
                   [](auto x, auto y) { return std::make_pair(x, y); });
  }
  // Update all references from all modules
  for (auto &&mp : impl->modules)
  {
    for (auto ins : iterator_for(mp.second))
      ins->replace_refs(mod_map, ins_map);
  }
}
```

I think from_value/to_value should be removed from the module class since it doesn't make sense for the module class to serialize its data when it references data from other classes. Instead,

- Modules are stored as an array instead of an object
- The modules are sorted into a breadth-first order so the parent module is always read or written first.
- The name of the inputs and outputs should be prefixed with the module name(ie `main:@1` instead of just `@1`).

> Looks like we are using a linear data structure to implement something like a tree

I am not sure if a tree would help. First, a tree implies all the nodes are connected which may not be the case with dead branches. Secondly, a tree could manage the relationships between modules but it wouldn't manage relationships between instructions.

This is what makes it more complicated. If we just needed relationships between modules we could use module names instead(as the name would be stable and wouldn't cause a dangling reference). However, if we want lexical scoping, then we need to update instruction_refs as well.

pfultz2:
For print, I think we can extend the module::print function to take in an initial set of names and then return the names created:

```cpp
std::unordered_map<instruction_ref, std::string> module::print(const std::function<
                   void(instruction_ref, const std::unordered_map<instruction_ref, std::string>&)>&
                       print_func,
                   std::unordered_map<instruction_ref, std::string> names = {}) const
{
    int count = 0;

    for(auto ins : iterator_for(*this))
    {
        std::string var_name;
        if(ins->name() == "@param")
        {
            var_name = any_cast<builtin::param>(ins->get_operator()).parameter;
        }
        else
        {
            var_name = this->name() + ":@" + std::to_string(count);
            count++;
        }
        names.emplace(ins, var_name);

        print_func(ins, names);
    }
    return names;
}
```

The `names` parameter should probably be a defaulted parameter taking an empty map. This is so the function can still be called without the parent names(which is needed so `==` can work). We need to update `print_instruction` so it prints a special name(perhaps a question mark`?`) when a name is not found instead of throwing an error.

Hopefully that makes sense.

src/module.cpp
scxiao: fixed.

src/include/migraphx/op/iff.hpp
scxiao: renamed.

src/include/migraphx/op/iff.hpp
scxiao: looks like the name() returned must be the same as the struct name, otherwise, there is a build error.

src/instruction.cpp
```diff
@@ -221,6 +295,50 @@ void instruction::finalize(context& ctx)
...
+        else
+            os << "{" << ins->get_literal() << "}";
+    }
+
+    if(!ins->inputs().empty())
+    {
+        char delim = '(';
+        for(auto&& arg : ins->inputs())
+        {
+            os << delim << names.at(arg);
```
pfultz2, line 319: I dont think you can do `names.at` here. Instead, you should check if `contains(names, arg)` otherwise print a `?`.

src/include/migraphx/module.hpp
```diff
@@ -109,21 +129,23 @@ struct module
...
+    void debug_print(instruction_ref ins,
+                     const std::unordered_map<instruction_ref, std::string>& names) const;
     void debug_print(const std::vector<instruction_ref>& inss) const;
-    void print(const std::function<void(instruction_ref,
-                                        const std::unordered_map<instruction_ref, std::string>&)>&
-                   print_func) const;
+
+    std::unordered_map<instruction_ref, std::string> 
+        print(const std::function<void(instruction_ref, const std::unordered_map<instruction_ref, std::string>&)>&
+              print_func, std::unordered_map<instruction_ref, std::string> names) const;
```
pfultz2, line 140: Names should be defaulted to an empty map.

src/module.cpp
```diff
@@ -478,28 +454,34 @@ void module::debug_print(instruction_ref ins) const
...
         this->debug_print(ins);
     std::cout << std::endl;
 }
 
-void module::print(const std::function<
+std::unordered_map<instruction_ref, std::string> module::print(const std::function<
                    void(instruction_ref, const std::unordered_map<instruction_ref, std::string>&)>&
-                       print_func) const
+                       print_func,
+                   std::unordered_map<instruction_ref, std::string> names = {}) const
```
pfultz2, line 482: The default parameter should go in the function definition in the header file, not here.

src/module.cpp
```diff
@@ -654,32 +632,66 @@ void module::print_cpp(std::ostream& os) const
...
+    	instruction::print(os, ins, ins_names);
         a(ins);
         os << std::endl;
     });
 }
 
+std::vector<module_ref> module::get_sub_modules() const
+{
+    std::vector<module_ref> vec_modules;
+    this->print([&](auto ins, auto) {
```
pfultz2, line 645: Why are we calling `print` to get submodules? We should just iterate over the instructions.

src/module.cpp
```diff
@@ -654,32 +632,66 @@ void module::print_cpp(std::ostream& os) const
...
     assert(this->validate() == this->end());
     return *this;
 }
 
+std::unordered_set<module_ref> module::get_parent_modules() const
+{
+    std::unordered_set<module_ref> parent_modules;
+    for(auto ins : iterator_for(*this))
+    {
+        auto& module_args = ins->module_inputs();
```
pfultz2, line 676: This looks like the submodules not the parent modules.

src/module.cpp
```diff
@@ -654,32 +632,66 @@ void module::print_cpp(std::ostream& os) const
...
+        for(auto& mod : module_args)
+        {
+            parent_modules.insert(mod);
```
pfultz2, line 679: This could use the range-based insertion(ie `parent_modules.insert(module_args.begin(), module_args.end())`).

src/include/migraphx/module.hpp
```diff
@@ -36,6 +39,8 @@ struct module
     // copy constructor
     module(const module&);
 
+    module(const module&, std::unordered_map<instruction_ref, instruction_ref>&);
```
pfultz2, line 42: Is this constructor used anywhere?

src/include/migraphx/op/iff.hpp
scxiao: removed

src/module.cpp
scxiao: it was written when we have the parent_mod in the module class. looks like it is get_sub_modules now, so I just removed this function since we already have the get_sub_modules().

src/include/migraphx/module.hpp
scxiao: It is not used. Actually, there is no implementation for this function at all. Just removed it.

src/module.cpp
scxiao: changed

src/include/migraphx/module.hpp
scxiao: changed

src/module.cpp
scxiao: changed

src/instruction.cpp
scxiao: changed

src/include/migraphx/module.hpp
```diff
@@ -2,11 +2,12 @@
 #define MIGRAPHX_GUARD_MIGRAPHLIB_MODULE_HPP
 
 #include <list>
+#include <unordered_set>
 #include <unordered_map>
 #include <migraphx/operation.hpp>
 #include <migraphx/literal.hpp>
 #include <migraphx/builtin.hpp>
-#include <migraphx/instruction_ref.hpp>
+#include <migraphx/instruction.hpp>
```
pfultz2, line 10: This shouldn't include the instruction header here.

src/include/migraphx/program.hpp
```diff
@@ -89,10 +88,21 @@ struct program
...
+    const module* get_module(const std::string& name) const;
+    void remove_module(const std::string& name);
+
     module* get_main_module();
     const module* get_main_module() const;
 
+    std::vector<const module*> get_modules() const;
+
     private:
+    void mod_from_val(module_ref mod, const value& v,
```
pfultz2, line 103: This could be declared static and moved to the cpp file.

src/include/migraphx/module.hpp
```diff
@@ -23,6 +24,8 @@ struct module_impl;
 
 using parameter_map = std::unordered_map<std::string, argument>;
 
+using module_ref = module*;
```
pfultz2, line 27: This is already declared in the module_ref header.

src/include/migraphx/module.hpp
scxiao: if just including "instruction_ref.hpp", there will be a build error for the 
std::unordered_map<instruction_ref, std::string> map_name = {}.

That is why i put the default argument value in the .cpp file last time.

pfultz2: I see, then we should add two overloads for this function. The one that doesnt have the names parameter can call the other overload and create the empty map.

scxiao: fixed.

src/include/migraphx/program.hpp

src/include/migraphx/module.hpp
scxiao: removed.

scxiao:
@pfultz2 There is a cppcheck like this:

```
Warning: method 'replace_instruction' can be made const [readability-make-member-function-const,-warnings-as-errors]
instruction_ref module::replace_instruction(instruction_ref ins,
                        ^
                                            const operation& op,
                                            std::vector<instruction_ref> args,
                                            std::vector<module_ref> module_args)

CMake Error at tidy-target-migraphx-module_cpp.cmake:78 (message):
  Clang tidy failed.
```

But if I put a const for this function as:

```
instruction_ref module::replace_instruction(instruction_ref ins,
                                            const operation& op,
                                            std::vector<instruction_ref> args,
                                            std::vector<module_ref> module_args) MIGRAPHX_TIDY_CONST
```

There is another cppcheck error reported:

```
Error while processing /workspace/host/projects/AMDMIGraphX/test/module_test.cpp.
/workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:76:16: error: no matching member function for call to 'replace_instruction' [clang-diagnostic-error]
        return replace_instruction(ins, op, {args...});
               ^
/workspace/host/projects/AMDMIGraphX/test/module_test.cpp:185:9: note: in instantiation of function template specialization 'migraphx::MIGRAPHX_INLINE_NS::module::replace_instruction<std::vector<std::_List_iterator<migraphx::MIGRAPHX_INLINE_NS::instruction>, std::allocator<std::_List_iterator<migraphx::MIGRAPHX_INLINE_NS::instruction>>>, std::vector<migraphx::MIGRAPHX_INLINE_NS::module *, std::allocator<migraphx::MIGRAPHX_INLINE_NS::module *>>>' requested here
    m1->replace_instruction(ins_pass, migraphx::make_op("if"), inputs, mod_inputs);
        ^
/workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:78:21: note: candidate function not viable: cannot convert initializer list argument to 'std::vector<instruction_ref>' (aka 'vector<_List_iterator<migraphx::MIGRAPHX_INLINE_NS::instruction>>')
    instruction_ref replace_instruction(instruction_ref ins,
                    ^
/workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:74:21: note: candidate template ignored: substitution failure: deduced incomplete pack <(no value)> for template parameter 'Ts'
    instruction_ref replace_instruction(instruction_ref ins, operation op, Ts... args)
                    ^
/workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:87:21: note: candidate function not viable: requires 2 arguments, but 3 were provided
    instruction_ref replace_instruction(instruction_ref ins, instruction_ref rep);
                    ^
/workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:82:21: note: candidate function not viable: requires 4 arguments, but 3 were provided
    instruction_ref replace_instruction(instruction_ref ins,
                    ^
```

scxiao:
> @pfultz2 There is a cppcheck like this:
> 
> ```
> Warning: method 'replace_instruction' can be made const [readability-make-member-function-const,-warnings-as-errors]
> instruction_ref module::replace_instruction(instruction_ref ins,
>                         ^
>                                             const operation& op,
>                                             std::vector<instruction_ref> args,
>                                             std::vector<module_ref> module_args)
> 
> CMake Error at tidy-target-migraphx-module_cpp.cmake:78 (message):
>   Clang tidy failed.
> ```
> 
> But if I put a const for this function as:
> 
> ```
> instruction_ref module::replace_instruction(instruction_ref ins,
>                                             const operation& op,
>                                             std::vector<instruction_ref> args,
>                                             std::vector<module_ref> module_args) MIGRAPHX_TIDY_CONST
> ```
> 
> There is another cppcheck error reported:
> 
> ```
> Error while processing /workspace/host/projects/AMDMIGraphX/test/module_test.cpp.
> /workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:76:16: error: no matching member function for call to 'replace_instruction' [clang-diagnostic-error]
>         return replace_instruction(ins, op, {args...});
>                ^
> /workspace/host/projects/AMDMIGraphX/test/module_test.cpp:185:9: note: in instantiation of function template specialization 'migraphx::MIGRAPHX_INLINE_NS::module::replace_instruction<std::vector<std::_List_iterator<migraphx::MIGRAPHX_INLINE_NS::instruction>, std::allocator<std::_List_iterator<migraphx::MIGRAPHX_INLINE_NS::instruction>>>, std::vector<migraphx::MIGRAPHX_INLINE_NS::module *, std::allocator<migraphx::MIGRAPHX_INLINE_NS::module *>>>' requested here
>     m1->replace_instruction(ins_pass, migraphx::make_op("if"), inputs, mod_inputs);
>         ^
> /workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:78:21: note: candidate function not viable: cannot convert initializer list argument to 'std::vector<instruction_ref>' (aka 'vector<_List_iterator<migraphx::MIGRAPHX_INLINE_NS::instruction>>')
>     instruction_ref replace_instruction(instruction_ref ins,
>                     ^
> /workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:74:21: note: candidate template ignored: substitution failure: deduced incomplete pack <(no value)> for template parameter 'Ts'
>     instruction_ref replace_instruction(instruction_ref ins, operation op, Ts... args)
>                     ^
> /workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:87:21: note: candidate function not viable: requires 2 arguments, but 3 were provided
>     instruction_ref replace_instruction(instruction_ref ins, instruction_ref rep);
>                     ^
> /workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:82:21: note: candidate function not viable: requires 4 arguments, but 3 were provided
>     instruction_ref replace_instruction(instruction_ref ins,
>                     ^
> ```

Do you have any suggestion on fixing it? use another name for the function?

```
instruction_ref module::replace_instruction(instruction_ref ins,
                                            const operation& op,
                                            std::vector<instruction_ref> args,
                                            std::vector<module_ref> module_args) MIGRAPHX_TIDY_CONST
```

pfultz2:
> Do you have any suggestion on fixing it? use another name for the function?

We should be able to use the same name for the function. It looks like from the compiler error it cannot convert the initializer list to a vector:

```
/workspace/host/projects/AMDMIGraphX/src/include/migraphx/module.hpp:78:21: note: candidate function not viable: cannot convert initializer list argument to 'std::vector<instruction_ref>' (aka 'vector<_List_iterator<migraphx::MIGRAPHX_INLINE_NS::instruction>>')
    instruction_ref replace_instruction(instruction_ref ins,
                    ^
```

I assume thats the overload you want to call.

pfultz2:
Ok, I think I see the issue. When calling the overload with module inputs(ie `m1->replace_instruction(ins_pass, migraphx::make_op("if"), inputs, mod_inputs);`) it calls the templated version, which then tries to pass instruction_ref vector and module_ref vector in the same initializer list(as `{inputs, mod_inputs}`), but since these are different types it fails.

Instead, we need to restrict the templated overload to only `instruction_ref` inputs. So we need to write something like:

```cpp
    template <class... Ts, MIGRAPHX_REQUIRES(std::is_same<Ts, instruction_ref>{}...)>
    instruction_ref replace_instruction(instruction_ref ins, operation op, Ts... args)
    {
        return replace_instruction(ins, op, {args...});
    }
```

We should probably do this for `insert_instruction` and `add_instruction` overloads as well.

scxiao:
> Ok, I think I see the issue. When calling the overload with module inputs(ie `m1->replace_instruction(ins_pass, migraphx::make_op("if"), inputs, mod_inputs);`) it calls the templated version, which then tries to pass instruction_ref vector and module_ref vector in the same initializer list(as `{inputs, mod_inputs}`), but since these are different types it fails.
> 
> Instead, we need to restrict the templated overload to only `instruction_ref` inputs. So we need to write something like:
> 
> ```c++
>     template <class... Ts, MIGRAPHX_REQUIRES(std::is_same<Ts, instruction_ref>{}...)>
>     instruction_ref replace_instruction(instruction_ref ins, operation op, Ts... args)
>     {
>         return replace_instruction(ins, op, {args...});
>     }
> ```
> 
> We should probably do this for `insert_instruction` and `add_instruction` overloads as well.

Many thanks. I made the change for the three functions.

src/include/migraphx/op/if_op.hpp
```diff
@@ -0,0 +1,32 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct if_op
+{
+    std::string name() const { return "if"; }
+
+    shape compute_shape(const std::vector<shape>&, std::vector<module_ref> mods) const
+    {
+        auto out_shapes = mods[0]->get_output_shapes();
```
pfultz2, line 23: We should still check the input shapes that there is 1 argument, and we should check there are two modules passed and they both have the same shape.

scxiao: fixed

pfultz2 (approved):

src/include/migraphx/operation.hpp
```diff
@@ -278,11 +347,10 @@ void from_value_op(T& x, const value& v)
  *      std::ptrdiff_t output_alias(const std::vector<shape>& input) const;
  *      void finalize(context& ctx,const shape& output,const std::vector<shape>& input) ;
  *      shape compute_shape(const std::vector<shape>& input) const;
- *      argument compute(context& ctx,const shape& output,const std::vector<argument>& input) const;
- *      argument compute(const shape& output,const std::vector<argument>& input) const;
- *      value to_value() const;
- *      void from_value(const value& v) ;
- *      value attributes() const;
+ *      shape compute_shape(const std::vector<shape>& inputs,const std::vector<module_ref>&
```
kahmed10, line 350: was this auto-generated? If not, it looks like we should probably fix the formatting so each new function is on a new line.

scxiao: Yes, it is auto-generated. I did not know why it is like this.

kahmed10 (approved):

scxiao:
This PR can be merged now. All the CI builds passed, but looks like it does not display correctly.

PR #742, opened 2021-02-04
Disable ort tests to pass jenkins

scxiao:
changes to run some tests one by one instead of simultaneously.


pfultz2 (approved):

scxiao:
@mvermeulen I think we can merge this PR now, then other PRs should pass the jenkins.
ssue #743 is created to track this issue.

PR #744, opened 2021-02-04
Add code object custom op

pfultz2:


sjw36 (approved):
Looks good to start. Please merge.

src/targets/gpu/include/migraphx/gpu/code_object_op.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+struct code_object_op
+{
+    value::binary code_object;
+    std::string symbol_name;
+    std::size_t global;
+    std::size_t local;
+    std::vector<shape> expected_inputs;
+    shape output;
+    kernel k;
```
sjw36, line 24: At some point we will want to be able to share the kernel for several code_object_op's since often in a model there are layers with identical dimensions and operations.

pfultz2: > since often in a model there are layers with identical dimensions and operations.

After applying horizontal fusions, I don't really see that much in models. We can try to see how to share the same binaries across different operators although it may be a little complicated because of serialization.

PR #747, opened 2021-02-11
Fix bug where instructions were not printed when doing TRACE_EVAL

pfultz2:
It just always printed that instruction was not part of the program. This also moves `print_instruction` to the instruction class and removed the duplicate function.


scxiao (approved):

mvermeulen:
Looks  like we might still be having some network issues...

PR #753, opened 2021-02-16
Bug simplify algegra

scxiao:


src/simplify_algebra.cpp
```diff
@@ -439,6 +444,39 @@ struct find_splits
         return groups;
     }
 
+    std::vector<int64_t> get_reduce_axes(instruction_ref ins) const
+    {
+        assert(contains(ins->name(), "reduce"));
+        std::vector<int64_t> axes;
```
pfultz2, line 450: This should just read the axes with `ins->get_operator().to_value()["axes"].to_vector<int64_t>()`.

src/simplify_algebra.cpp
```diff
@@ -453,6 +491,23 @@ struct find_splits
             if(op.name() == "slice")
                 continue;
 
+            if(contains(op.name(), "reduce"))
```
pfultz2, line 494: If we are going to support reduce then the matcher should be updated to match reduce.

src/simplify_algebra.cpp
```diff
@@ -453,6 +491,23 @@ struct find_splits
...
+                auto reduce_axes = get_reduce_axes(start);
+                // axes of slice and reduce op cannot have overlap
+                if(std::any_of(slc_axes.begin(), slc_axes.end(), [&](auto axis) {
+                       return (std::find(reduce_axes.begin(), reduce_axes.end(), axis) !=
+                               reduce_axes.end());
+                   }))
+                {
+                    continue;
+                }
+            }
```
pfultz2, line 509: We need and else here to check that the operator is pointwise. If its not reduce or pointwise then we should `continue`.

src/simplify_algebra.cpp
```diff
@@ -406,7 +406,7 @@ struct find_splits
     auto matcher() const
     {
         return match::any(match::any_of[match::outputs()](
-            match::name("slice")(match::any_of[match::outputs()](pointwise()))));
+            match::name("slice")(match::any_of[match::outputs()](pointwise(), match::name_contains("reduce")))));
```
pfultz2, line 409: Just like we have a `pointwise` matcher(defined [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/2ab73fd32145fa645da05c0e6e2fee77cd1889a0/src/simplify_algebra.cpp#L43)), could we add a `reduction` matcher to check for reduce operators?

pfultz2 (approved):

kahmed10 (approved):

scxiao:
Looks like all checks have passed, but it does not display here correctly. So this PR can be merged now.

PR #755, opened 2021-02-18
fix tf fusion regressions

kahmed10:
caused by transposes left in program that made matchers not recognize fusions


mvermeulen:
As noted separately in email, this seems to address the performance regression this week.  Tested with normal performance tests (minus onnx runtime tests).

scxiao (commented):
do we need some unit tests for the changes?

kahmed10:
Added unit tests

scxiao (approved):

kahmed10:
Retry DeepCode

Issue #722, opened 2021-01-21
Unknown operator: LessOrEqual


cagery:
Shows up at ***t5-decoder-with-lm-head-12.onnx*** from ONNX Model Zoo when _cpp_parse_load_save_ app is used to parse:
https://github.com/onnx/models/blob/master/text/machine_comprehension/t5/model/t5-decoder-with-lm-head-12.onnx

```
MIGraphX Error: /cgMIGraphX/AMDMIGraphX/src/onnx/onnx_parser.cpp:281: parse_graph: Unknown operator: LessOrEqual
```

mvermeulen:
Done

PR #758, opened 2021-02-19
Changes for LessOrEqual operator

cagery:
This PR implements LessOrEqual operator, and fixes issue #722.

- Ref and GPU are implemented.
- Compilation and tests pass.
```
100% tests passed, 0 tests failed out of 282
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ find_package(nlohmann_json 3.8.0 REQUIRED)
 
 include(ROCMSetupVersion)
 
-rocm_setup_version(VERSION 0.9)
+rocm_setup_version(VERSION 1.0)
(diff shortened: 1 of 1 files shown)


scxiao (commented):
I would prefer this operator to be implementation with the combination of less, or, and equal as
result = (a < b) or (a == b)
In this way, we do not need to implement the kernel, which can reduce our build time in some degree. In addition, it will help for the graph optimization since we have less operations to handle.

kahmed10:
It might be easier to implement it as y = not(greater(x))

pfultz2:
Actually, it could be implemented as `not(less(y, x))` by just flipping the inputs.

cagery:
Closed this branch as the changes here are unnecessary. Working on another branch.

PR #759, opened 2021-02-23
Changes to Jupyter Notebook of ResNet-50 Python Inference Example

cagery:
While trying `ResNet-50 Python Inference` example, I have encountered a few issues:

1. cv2 library could not be found. I am aware this is python binary/installation specific, and user has to make sure of it, but still added two lines at first cell to install the missing library. 
2. After executing the notebook, I found out latest `cv2` library version causes issues with `qt` plugin. Please see this link:
https://stackoverflow.com/questions/60042568/this-application-failed-to-start-because-no-qt-platform-plugin-could-be-initiali  
Thus, I made sure we install version `opencv-python==4.1.2.30`.
3. I utilized headless server, but I still wanted to make sure the functionality is there. Headless server could not open the video/image through `.imshow()`, so I have added a new cell just to get the labels. Labels seemed to be correct. (I have downloaded the sample video and watched it 😄 )

Let me know your comments/inputs.


examples/python_api_inference/resnet50_inference.ipynb
```diff
@@ -330,12 +286,28 @@
...
+   "source": [
+    "while (cap.isOpened()):\n",
+    "    start = time.perf_counter()\n",
+    "    ret, frame = cap.read()\n",
+    "    if not ret: break\n",
+    "    \n",
+    "    top_prediction = predict_class(frame)\n",
+    "    print(labels[top_prediction])\n",
+    "cap.release()"
+   ]
```
turneram, line 310: Looks good. Only suggestion is that we could maybe show a histogram of predictions once it is finished rather than having it print the prediction for every frame.

cagery: I like the idea. I will add it and circle back afterwards.

turneram (approved):

PR #760, opened 2021-02-25
Tf multi output support

kahmed10:


src/driver/main.cpp
```diff
@@ -65,6 +66,12 @@ struct loader
            ap.help("Dim of a parameter (format: \"@name d1 d2 dn\")"),
            ap.append(),
            ap.nargs(2));
+
+        ap(output_names,
+           {"--output-names"},
+           ap.help("Names of node output (format: \"name_1 name_2 name_n\")"),
+           ap.append(),
+           ap.nargs(2));
```
pfultz2, line 74: So what happens when the user doesn't pass in the output_names?

src/tf/tf_parser.cpp
```diff
@@ -286,9 +286,30 @@ void tf_parser::parse_graph(const tensorflow::GraphDef& graph)
...
     }
-
-    // Needs to add a ret instruction at the end of
-    // the program
+    auto last_ins = std::prev(mm->end());
+    if(last_ins != mm->end())
+    {
+        // Needs to add a ret instruction at the end of
+        // the program
+        if(output_node_names.empty())
```
pfultz2, line 294: There is no way to compute output_node_name, by just taking the nodes that are never used for input?

kahmed10: There could be nodes that remain from training that don't have the output being used. I think that's the main issue with checking if a node is never used for input.

pfultz2: I think you can compute the outputs with something like this:

```cpp
std::unordered_set<std::string> inputs;
for(auto&& p : nodes)
{
    auto&& node = p.second;
    std::copy(node.input().begin(), node.input().end(), std::inserter(inputs, inputs.end()));
}
std::unordered_set<std::string> outputs;
std::copy_if(nodes.begin(), nodes.end(), std::inserter(outputs, outputs.end()), [&](auto&& p) {
    const auto& name = p.first;
    const auto& node = p.second;
    if(contains(name, "^"))
        return false;
    for(const auto& x:{"NoOp", "Assert"})
    {
        if (node.op() == x)
            return false;
        if(contains(name, x))
            return false;
    }
    return inputs.count(name) == 0;
});
```

pfultz2: So when the user doesn't specify the outputs how to pick which output? Or is it random?

kahmed10: In TF, passing in the output name is enforced when running the graph or exporting it. Otherwise, TF will throw an error.

scxiao: Then where can a user get the output names?

kahmed10: They can use TF scripts that read pb files (similar to onnx), then pick the output node they want.

pfultz2 (approved):
This is good for initial support so it can be provided by release time. We can improve the inferring of the output names on a future PR.

src/driver/main.cpp
kahmed10: Currently it will just default to the last instruction in the program (same as before this PR)

PR #764, opened 2021-03-02
Infer outputs in tf

kahmed10:


kahmed10:
Some tf tests have been adjusted accordingly so that they hit the targeted code changes.

src/targets/gpu/device/unary_not.cpp
```diff
@@ -7,13 +7,9 @@ inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 namespace device {
 
-void unary_not(hipStream_t stream,
-                 const argument& result,
-                 const argument& arg
-                 )
+void unary_not(hipStream_t stream, const argument& result, const argument& arg)
```
kahmed10, line 10: these files were changed after a manual merge, but I believe the formatting is correct since it passes jenkins

scxiao (approved):

pfultz2 (approved):

PR #765, opened 2021-03-03
Module build exec

scxiao:
PR to build and execute a module...


src/auto_contiguous.cpp
```diff
@@ -12,6 +12,13 @@ void auto_contiguous::apply(module& p) const
 {
     for(auto ins : iterator_for(p))
     {
+        // apply to sub_modules
+        const auto& mod_inputs = ins->module_inputs();
+        for(const auto& mod : mod_inputs)
+        {
+            this->apply(*mod);
+        }
```
pfultz2, line 20: Passes should not call out to the submodules directly. Rather run_passes should call the passes for the child modules before calling it for the parent module.(I believe this is the reverse of what `get_sub_modules` return).

src/targets/gpu/if.cpp
```diff
@@ -0,0 +1,32 @@
...
+    return op.compute_shape(inputs, mods);
+}
+
+argument hip_if::compute(
+    const std::vector<argument>& args,
+    const std::vector<module_ref>& mods,
+    std::function<std::vector<argument>(
+        module_ref& mdl, const std::unordered_map<std::string, argument>& inputs)>& run) const
+{
+    auto arg_cond  = migraphx::gpu::from_gpu(args[0]);
```
pfultz2, line 20: I think it would be better to insert a copy_from_gpu instruction during lowering.

src/auto_contiguous.cpp
scxiao: Changed

src/targets/gpu/if.cpp
scxiao: Changed

src/onnx/onnx_parser.cpp
```diff
@@ -283,9 +290,11 @@ void onnx_parser::parse_graph(module* mod, const onnx::GraphProto& graph)
         }
         else
         {
-            result = ops[node.op_type()](
-                *this, {get_attributes(node), output_num, node.op_type(), mod}, args);
+            std::string node_name = node.op_type() + "_" + std::to_string(node_index);
+            result                = ops[node.op_type()](
+                *this, {get_attributes(node), output_num, node_name, mod}, args);
         }
+        ++node_index;
```
pfultz2, line 297: Can `instructions.size()` be used for this instead?

src/opt/memory_coloring_impl.hpp
```diff
@@ -145,7 +151,54 @@ struct memory_coloring_impl
...
+            {
+                for(const auto* ssmod : mod_args)
+                {
+                    calc_implicit_deps(*ssmod, pmod, ins, deps);
+                }
+            }
+        }
+    }
+
+    void calc_implicit_deps(module& p)
```
pfultz2, line 183: Doesnt the scheduler need this function? Also this should be made a member function to the module class and it should also return the map.

src/program.cpp
```diff
@@ -143,19 +143,28 @@ void program::compile(const target& t, compile_options options)
...
+    std::reverse(mods.begin(), mods.end());
+    auto pass_options = options;
+
+    for(const auto& mod : mods)
     {
-        auto index = std::distance(modl->begin(), invalid);
-        MIGRAPHX_THROW("Invalid module " + modl->name() + " from compilation at instruction " +
-                       std::to_string(index));
+        // submodules do not use offload_copy
+        pass_options.offload_copy = ((mod->name() == "main") ? options.offload_copy : false);
```
pfultz2, line 154: We shouldn't modify the compile options here.

pfultz2: Instead, lowering should check if the module name is `main` to do offload copying.

src/program.cpp
```diff
@@ -143,19 +143,28 @@ void program::compile(const target& t, compile_options options)
...
+        auto&& passes             = t.get_passes(this->impl->ctx, pass_options);
```
pfultz2, line 155: We should only get_passes once not inside the loop.

src/program.cpp
```diff
@@ -221,15 +233,42 @@ std::vector<argument> generic_eval(const module& p,
...
-                            }));
+
+            const auto& mod_args = ins->module_inputs();
+            if(not mod_args.empty())
+                results.emplace(
+                    ins, trace(ins, mod, [&] {
+                        return ins->get_operator().compute(
+                            values,
+                            mod_args,
+                            [&](module_ref smod,
```
pfultz2, line 244: This lambda can be assigned to a variable and then passed in. I think it will be cleaner.

src/program.cpp
```diff
@@ -221,15 +233,42 @@ std::vector<argument> generic_eval(const module& p,
...
-            results.emplace(ins, trace(ins, [&] {
-                                return ins->normalized_operator().compute(
-                                    ctx, ins->get_shape(), values);
-                            }));
+
+            const auto& mod_args = ins->module_inputs();
+            if(not mod_args.empty())
+                results.emplace(
+                    ins, trace(ins, mod, [&] {
+                        return ins->get_operator().compute(
```
pfultz2, line 241: This should be `ins->normalized_operator().compute(`.

src/program.cpp
```diff
@@ -221,15 +233,42 @@ std::vector<argument> generic_eval(const module& p,
...
+                                    else
+                                        MIGRAPHX_THROW("Input " + nm + " parameter of module: \"" +
+                                                       mod->name() + "\" not exist!");
+                                return generic_eval(smod, ctx, m, results, trace);
+                            });
+                    }));
+            else
+            {
+                results.emplace(ins, trace(ins, mod, [&] {
+                                    return ins->normalized_operator().compute(
```
pfultz2, line 263: Is this necessary? Won't `compute` call the non-module overload when the modules are empty? If not we can change this in a future PR, and no need to change here.

src/program.cpp
```diff
@@ -261,21 +300,24 @@ std::vector<argument> program::eval(parameter_map params) const
...
-            if(trace_level > 1 and ins->name().front() != '@' and ins->name() != "load")
-                std::cout << "Ouput: " << result << std::endl;
-            return result;
-        });
+        std::unordered_map<instruction_ref, std::string> names;
+        return generic_eval(
+            *this, ctx, std::move(params), [&](auto& ins, const module* mod, auto f) {
+                ctx.finish();
+                std::cout << "Run instruction: ";
+                mod->debug_print(ins, names);
```
pfultz2, line 308: How is names update here? Can't we just do `this->debug(ins)` here?

src/targets/ref/lowering.cpp
```diff
@@ -885,9 +886,41 @@ struct ref_rnn_var_sl_last_output
 };
 MIGRAPHX_REGISTER_OP(ref_rnn_var_sl_last_output)
 
+struct ref_if
```
pfultz2, line 889: Is this necessary? Why cant the ref use the original `if` op's `compute` method?

src/program.cpp
scxiao: Currently, the compute function with mod_args input does not have the context input. As I talked to you before, if context is added as an input, I met the problem of the compute_function not found in gpu computation.

src/program.cpp
scxiao: If we do not have the name input, instructions of main module cannot be printed out in submodules.

src/onnx/onnx_parser.cpp
scxiao: fixed

src/opt/memory_coloring_impl.hpp
scxiao: moved to module as a member function

src/program.cpp
scxiao: changed accordlingly

src/program.cpp
scxiao: changed accordingly

src/program.cpp
scxiao: fixed accordingly

src/program.cpp
scxiao: changed

src/targets/ref/lowering.cpp
scxiao: changed accordingly

src/include/migraphx/op/if_op.hpp
```diff
@@ -36,6 +36,31 @@ struct if_op
...
+            module_ref& mdl, const std::unordered_map<std::string, argument>& inputs)>& run) const
+    {
+        auto cond      = args.front().at<bool>();
+        module_ref mod = cond ? mods[0] : mods[1];
+        std::unordered_map<std::string, argument> params;
+        // const auto& out_shapes = mod->get_output_shapes();
+        assert(args.size() == 1 or args.size() == mod->get_output_shapes().size() + 1);
+        for(std::size_t i = 1; i < args.size(); ++i)
+        {
+            std::string name = mod->name() + ":#output_" + std::to_string(i - 1);
```
pfultz2, line 53: The arguments should go to all parameters in the submodule not just the ones named `#output`. It should be the union of all parameters in each module:

```cpp
std::set<std::string> pnames;
for(auto mod:mods)
{
    auto names = mod->get_parameter_names();
    pnames.insert(names.begin(), names.end());
}
assert(pname.size() < args.size());
std::transform(args.begin()+1, 
               args.end(),
               pnames.begin(),
               std::inserter(params, params.end()), 
               [](auto&& arg, auto&& name) {
                   return std::make_pair(name, arg);
               });
```

src/program.cpp
pfultz2: What do you mean? Are you saying `this->debug_print(ins)` is broken?

pfultz2:
Can you add unit tests to memory_coloring and schedule for implicit deps?

src/program.cpp
```diff
@@ -221,15 +225,44 @@ std::vector<argument> generic_eval(const module& p,
...
+                                   const std::unordered_map<std::string, argument>& inputs) {
+                // wrap up parameters for sub_modules
+                const auto& param_names = smod->get_parameter_names();
+                parameter_map m;
+                for(const auto& nm : param_names)
+                {
+                    if(contains(inputs, nm))
+                        m[nm] = inputs.at(nm);
+                    else
+                        MIGRAPHX_THROW("Input " + nm + " parameter of module: \"" + mod->name() +
```
pfultz2, line 240: Shouldn't this be `smod`?

src/program.cpp
```diff
@@ -221,15 +225,44 @@ std::vector<argument> generic_eval(const module& p,
...
+                                       "\" not exist!");
+                }
+
+                return generic_eval(smod, ctx, m, results, trace);
```
pfultz2, line 244: Why cant inputs be passed directly into `generic_eval`? We already check if a parameter is missing.

src/opt/memory_coloring_impl.cpp
```diff
@@ -185,8 +202,9 @@ void memory_coloring_impl::rewrite()
     std::vector<std::size_t> dims;
     dims.push_back((required_bytes + sizeof(float) - 1) / sizeof(float));
     shape s                       = {shape::float_type, dims};
-    instruction_ref scratch_param = p_program->add_parameter("scratch", s);
-    for(auto ins : iterator_for(*p_program))
+    std::string name              = p_mod->name() + "scratch";
```
pfultz2, line 205: Why can't this just be `scratch`?

src/program.cpp
scxiao: fixed

src/program.cpp
scxiao: fixed

src/opt/memory_coloring_impl.cpp
scxiao: the preallocation_param pass has an id "scratch", which is store per program. We need to use the mod name to differentiate the scratch for different modules.

pfultz2: The preallocation_param pass should be updated to store the parameter for each module it is used in.

src/targets/gpu/preallocate_param.cpp
```diff
@@ -16,8 +16,9 @@ void preallocate_param::apply(module& p) const
     {
         if(ins->name() != "@param")
             continue;
-        std::string id = any_cast<builtin::param>(ins->get_operator()).parameter;
-        if(id != param)
+        std::string id        = any_cast<builtin::param>(ins->get_operator()).parameter;
+        std::string mod_param = p.name() + param;
+        if(id != mod_param)
```
pfultz2, line 21: This should be:

```cpp
if (param != any_cast<builtin::param>(ins->get_operator()).parameter)
    continue;
std::string id = p.name() + ":" param;
```

This will match `scratch` parameters for all modules and give each one a unique id.

src/targets/gpu/write_literals.cpp
```diff
@@ -28,7 +28,7 @@ void write_literals::apply(module& p) const
             }
             else
             {
-                std::string id = "@literal:" + std::to_string(n);
+                std::string id = p.name() + "@literal:" + std::to_string(n);
```
pfultz2, line 31: It would be better with a colon after the module name: `std::string id = p.name() + ":@literal:" + std::to_string(n);`

src/targets/gpu/include/migraphx/gpu/hip.hpp
```diff
@@ -103,6 +103,14 @@ struct hip_copy_to_gpu
 
 struct hip_copy_from_gpu
 {
+    bool sync_copy = false;
```
pfultz2, line 106: This should be removed. A `hip_sync_device` should be inserted after the copy if it needs to be synced. We might need to insert an identity operator so scheduling won't change the order.

src/opt/memory_coloring_impl.cpp
scxiao: fixed

src/targets/gpu/preallocate_param.cpp
scxiao: fixed

src/targets/gpu/write_literals.cpp
scxiao: fixed

src/targets/gpu/include/migraphx/gpu/hip.hpp
scxiao: changed

pfultz2 (approved):

pfultz2:
@kahmed10 Can review this PR?

src/schedule.cpp
```diff
@@ -114,7 +123,9 @@ struct stream_info
             assert(ins != p.end());
             if(contains(partitions, ins))
                 return;
-            assert(p.has_instruction(ins));
+            if(not p.has_instruction(ins))
+                return;
+            // assert(p.has_instruction(ins));
```
kahmed10, line 128: do we need this assert comment here?

scxiao: removed.

kahmed10 (approved):

scxiao:
For this PR, the "deepcode-ci-bot" failed.

Issue #768, opened 2021-03-05
MLIR MIOpen Dialect integration (phase 1)


sjw36:
Integrate MLIR to enable MIOpen Dialect for iGEMM usage.

- Import MLIR libraries and C-API headers
- Add a pass to the GPU target before lowering that:
  - Find convolutions supported by MLIR dialect
  - Call MLIR C-API to generate kernel binaries
  - Substitute the gpu::convolution with a code_object_op

mvermeulen:
Closing as merged.

PR #769, opened 2021-03-06
MLIR MIOpen Dialect integration (phase 1) (#768)

sjw36:
* Added Findmlir.cmake (using environment variables to import)

* Added mlir_conv pass to GPU target

  * Apply to any gpu::convolution if supported by MLIR

  * Call MLIR C-API to generate iGEMM kernel with configuration from gpu::convolution

  * Capture binary in dictionary for matching convolutions

  * Build a code_object_op with the binary and execution dimensions

  * Substitute for the gpu::convolution


sjw36:
@pfultz2 this enables MLIR substitution but I am getting a crash in allocating params:

> Allocating params ...
> Memory access fault by GPU node-4 (Agent handle: 0x3338700) on address 0x1000. Reason: Page not present or supervisor privilege.
> Aborted (core dumped)

It will also require a local build of https://github.com/ROCmSoftwarePlatform/llvm-project-mlir once this PR is merged (https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/64).

cmake/Findmlir.cmake
```diff
@@ -0,0 +1,24 @@
+# This file allows users to call find_package(MLIR) and pick up our targets.
+
+set(HAS_MLIR_MIOPEN_SUPPORT False)
+
+if(DEFINED ENV{MLIR_SOURCE_PATH})
+  set(MLIR_SOURCE_PATH   "$ENV{MLIR_SOURCE_PATH}")
+  set(MLIR_BUILD_PATH    "$ENV{MLIR_BUILD_PATH}" )
```
pfultz2, line 7: We shouldn't rely on env variables to find dependencies. This should just use cmake's find_* command to find the different pieces needed. Ideally, it would be even nicer if mlir could provide a mlir-config.cmake so we dont need to this find module, but this is fine for a prototyping phase.

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,214 @@
...
+              auto flt = ins->inputs().at(1);
+              return prog->replace_instruction(ins, conv, {flt, inp});
+            }
+            return ins;
+        });
+    }
+
+};
+
+void target_mlir_conv::apply(module& p) const { mlir_apply{&p, this}.apply(); }
```
pfultz2, line 211: This follow the lowering passes, but I think it makes it more complicated in this case. You could use the matchers:

```cpp
struct find_mlir_convolution
{
    context * ctx = nullptr;
    auto matcher() const
    {
        return match::name("convolution");
    }

    void apply(module& m, match::matcher_result r) const
    {
        auto ins = r.result;
        // Replace ins with gpu::convolution
        ...
    }
};

void target_mlir_conv::apply(module& m) const 
{ 
    find_matches(m, find_mlir_convolution{&ctx});
}
```

Which could extend this to match more operators together(like conv+bias+activation).

However, if you just want to look at convolution, an even simpler way is to just use a simple for loop:

```cpp
void target_mlir_conv::apply(module& m) const 
{ 
    for(auto ins:iterator_for(m))
    {
        if (ins->name() != "convolution")
            continue;
        // Replace ins with gpu::convolution
        ...
    }
}
```

src/targets/gpu/CMakeLists.txt
```diff
@@ -252,6 +254,9 @@ endif()
 message(STATUS "clang-offload-bundler: ${MIGRAPHX_OFFLOADBUNDLER_BIN}")
 message(STATUS "extractkernel: ${MIGRAPHX_EXTRACT_KERNEL}")
 
+target_compile_definitions(migraphx_gpu PRIVATE "-DMIGRAPHX_MLIR_MIOPEN_SUPPORT")
```
pfultz2, line 257: We should have an explicit cmake flag to enable MLIR: `set(MIGRAPHX_ENABLE_MLIR Off CACHE BOOL "")` and then only add those flags if it is set.

src/targets/gpu/CMakeLists.txt
```diff
@@ -252,6 +254,9 @@ endif()
...
+target_link_libraries(migraphx_gpu PUBLIC MLIRMIOpenThin)
```
pfultz2, line 258: This should search for the library with `find_library`, similar to how MIOpen finds MLIR:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/CMakeLists.txt#L318

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return *pass->ctx;
+    }
+
+    void init()
+    {
+        assert(prog != nullptr);
+        assert(pass != nullptr);
+    }
+
+    execution_spec *make_mlir_binary(instruction_ref op_r)
```
pfultz2, line 65: I think this should just return by value. We know if it didn't create the binary if `binary.empty()` is true.

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,264 @@
...
+          " --out_channels " + std::to_string(out_t.lens()[1]) +
+          " --out_h "        + std::to_string(out_t.lens()[2]) +
+          " --out_w "        + std::to_string(out_t.lens()[3]);
+
+        auto bin_i = binary_map.find(mlir_options);
+        if (bin_i == binary_map.end()) {
+          size_t bin_size = 0;
+          char *bin_buffer = nullptr;
+
+          auto mlir_handle = miirCreateHandle(mlir_options.c_str());
```
pfultz2, line 140: The handle should be created with a managed pointer(which is just a `std::unique_ptr` with a custom deleter) so it will cleanup even if an exception occurs:

```cpp
using mlir_handle = MIGRAPHX_MANAGE_PTR(MiirHandle, miirDestroyHandle);

auto handle = mlir_handle{mlir_options.c_str()};
```

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+          if (miirLowerBin(mlir_handle) == MIIR_SUCCESS &&
+              miirBufferGet(mlir_handle, nullptr, &bin_size) == MIIR_SUCCESS) {
+            bin_buffer = new char[bin_size];
```
pfultz2, line 144: This is never deleted, use [`make_shared_array`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/make_shared_array.hpp#L11) here.

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,264 @@
...
+      const size_t offset = 0;
+      auto inst_t = inst->get_shape();
+      refs.push_back(inst);
+      refs.push_back(inst);
+      refs.push_back(get_literal(offset)); // offset
+
+      auto &lens = inst_t.lens();
+      for (int i = 0; i < lens.size(); ++i) {
+        refs.push_back(get_literal(lens[i]));
+      }
```
pfultz2, line 210: This can be written with std::transform instead.

src/targets/gpu/mlir_conv.cpp
sjw36: That would induce a copy, would be very expensive with the binary vector copy.

pfultz2: NRVO should elide the copy, but even if that is not used, the compiler should implicitly move the value in C++14(according to [1579](http://open-std.org/JTC1/SC22/WG21/docs/cwg_defects.html#1579)), which will avoid a copy, and just do some pointer swaps.

sjw36: But the execution_spec is potentially shared from the unordered_map (which owns the object). That way multiple conv2d's with the same config will not call the compiler every time.

pfultz2: I see, so if the map owns it, it should be `std::unordered_map<std::string, execution_spec>` and then this function can return a pointer to the element in the map(it doesn't seem like shared ownership is necessary), but you will need to add elements to the map with `binary_map.emplace()` so it can move(rather than copy) the object into the map.

sjw36: True. will do.

sjw36: Will use shared_ptr instead since we still want nullptr's in the map.

src/targets/gpu/mlir_conv.cpp
sjw36: How does that call miirCreateHandle?

pfultz2: Oops, that should be `auto handle = mlir_handle{miirCreateHandle(mlir_options.c_str())};`

src/targets/gpu/mlir_conv.cpp
pfultz2: Why do we want nullpointers in the map? Whats the difference between the key not being in the map and the key being in the map but the value is null?

Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     apt-utils \
     build-essential \
     clang-format-5.0 \
+    clang \
```
pfultz2, line 16: Why are we installing clang?

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,291 @@
...
+    void add_memref_descriptor(std::vector<instruction_ref>& refs, instruction_ref inst)
+    {
+        const size_t offset = 0;
+        auto inst_t         = inst->get_shape();
+        refs.push_back(inst);
+        refs.push_back(inst);
+        refs.push_back(get_literal(offset)); // offset
+
+        // dim sizes
+        std::for_each(inst_t.lens().begin(), inst_t.lens().end(), [this, &refs](const auto lval) {
```
pfultz2, line 229: It would be better to use implicit `[&]` capture for the lambda.

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,291 @@
...
+            refs.push_back(get_literal(lval));
+        });
```
pfultz2, line 231: This can be done with `std::transform` using `std::back_inserter`:

```cpp
std::transform(inst_t.lens().begin(), inst_t.lens().end(), std::back_inserter(refs), [&](const auto& lval) {
    return get_literal(lval);
});
```

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,291 @@
...
+            refs.push_back(out);
+
+            prog->replace_instruction(ins, conv, refs);
+        }
+    }
+
+    void apply()
+    {
+        init();
+        for(auto it = prog->begin(); it != prog->end(); it++)
```
pfultz2, line 271: This can be written as `for(auto it:iterator_for(*prog))`

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+#include <stdio.h>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+
+struct mlir_apply
+{
+    module* prog          = nullptr;
```
pfultz2, line 38: The variable should be named `mod` and not `prog` as its a module and not a program.

src/targets/gpu/mlir_conv.cpp
```diff
@@ -0,0 +1,291 @@
...
+        {
+            if(it->name() == "convolution")
+            {
+                replace_conv_op(it);
+            }
+        }
+    }
+};
+
+void mlir_conv::apply(module& p) const
```
pfultz2, line 281: It would be better for the parameter to be `m`(for module) instead of `p`(which would stand for `program`).

src/targets/gpu/mlir_conv.cpp
pfultz2: Wait, I see, null means it tried but failed to compile a kernel, whereas a missing key means it hasn't even tried.

src/include/migraphx/value.hpp
```diff
@@ -140,9 +140,9 @@ struct value
     using const_pointer   = const value_type*;
     using array           = std::vector<value>;
     using object          = std::unordered_map<std::string, value>;
-    struct binary : std::vector<std::uint8_t>
+    struct binary : std::vector<std::int8_t>
```
pfultz2, line 143: Is there a reason you can't cast it to `char*` like what we do for [msgpack](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/msgpack.cpp#L88)?

This also breaks the unit tests.

Dockerfile
sjw36: No reason. Removing..

src/targets/gpu/mlir_conv.cpp
sjw36: Cool, I will try that.

src/include/migraphx/value.hpp
sjw36: No problem, I tried this first but didn't work so I added the cast and forgot to remove this.

src/targets/gpu/mlir_conv.cpp
sjw36: We prefer explicit capture in MLIR but I can match your preference here.

src/targets/gpu/CMakeLists.txt
```diff
@@ -259,6 +261,20 @@ endif()
 message(STATUS "clang-offload-bundler: ${MIGRAPHX_OFFLOADBUNDLER_BIN}")
 message(STATUS "extractkernel: ${MIGRAPHX_EXTRACT_KERNEL}")
 
+set(MIGRAPHX_ENABLE_MLIR ON CACHE BOOL "")
```
pfultz2, line 264: This should be set to `Off` by default.

src/include/migraphx/compile_options.hpp
```diff
@@ -11,6 +11,7 @@ struct compile_options
 {
     bool offload_copy = false;
     bool fast_math    = true;
+    bool enable_mlir  = false;
```
pfultz2, line 14: I think for now we shouldn't need a flag. By default, migraphx doesn't use MLIR so this flag seems unnecessary. We can use the `MIGRAPHX_ENABLE_MLIR` build flag to decide if its enabled or not.

The problem with a flag like this is that none of the verification tests enable it. With a build time flag we can easily add a Jenkins stage with `-DMIGRAPHX_ENABLE_MLIR=On` and it will run all the verification tests with mlir enable. Once we get further along to use MLIR by default, then we can think about how we might want to enable or disable it.

pfultz2:
We can add a stage to jenkins after `clang_release` to run and build mlir:

```groovy
rocmtest clang_debug: rocmnode('vega') { cmake_build ->
    stage('Hip Clang Debug') {
        def sanitizers = "undefined"
        def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers}"
        cmake_build("/opt/rocm/llvm/bin/clang++", "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}'")
    }
}, clang_release: rocmnode('vega') { cmake_build ->
    stage('Hip Clang Release') {
        cmake_build("/opt/rocm/llvm/bin/clang++", "-DCMAKE_BUILD_TYPE=release")
        stash includes: 'build/*.deb', name: 'migraphx-package'
    }
}, mlir_debug: rocmnode('vega') { cmake_build ->
    stage('MLIR Debug') {
        def sanitizers = "undefined"
        def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers}"
        cmake_build("/opt/rocm/llvm/bin/clang++", "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}'")
    }
}
```

src/include/migraphx/compile_options.hpp
sjw36: gone.

sjw36:
I believe all issues are addressed. Anything else @pfultz2 ?

pfultz2:
There is still an issue on jenkins:

```
/var/jenkins/workspace/MLLibs_AMDMIGraphX_PR-769/src/targets/gpu/mlir_conv.cpp:71:70: error: unused parameter 'op_r' [-Werror,-Wunused-parameter]
```

Its really only unused when MLIR is not enabled, so you should probably add a `(void)op_r` line on the `#else` part of the ifdef.

Also, can you add a stage to jenkins with MLIR enabled? I mentioned how [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/769#issuecomment-825887677).

pfultz2:
Also, it looks formatting is needed again:

```
sh -c clang-format-5.0 -style=file ./src/targets/gpu/mlir_conv.cpp | diff - ./src/targets/gpu/mlir_conv.cpp 
285c285,288
< void mlir_conv::apply(module& m) const { mlir_apply{&m, this}.apply(); }
---
> void mlir_conv::apply(module& m) const
> {
>     mlir_apply{&m, this}.apply();
> }
```

sjw36:
what is up with the false error on deepcode?

pfultz2 (approved):

sjw36:
still blocked..

pfultz2:
There's a new compiler warning from merging in the tuple type PR:

```
[2021-04-28T01:15:33.019Z] /var/jenkins/workspace/MLLibs_AMDMIGraphX_PR-769/src/targets/gpu/mlir_conv.cpp:82:20: error: enumeration value 'tuple_type' not handled in switch [-Werror,-Wswitch]

[2021-04-28T01:15:33.019Z]             switch(s.type())

[2021-04-28T01:15:33.019Z]                    ^
```

sjw36:
@mvermeulen looks like it passed all required. I don't have merge permissions, please merge. thx

PR #777, opened 2021-03-18
Propagate data layout in the operators

pfultz2:


kahmed10 (approved):

src/permutation.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return {s.type(), reorder_dims(s.lens(), permutation), reorder_dims(s.strides(), permutation)};
+}
+
+std::vector<int64_t> invert_permutation(const std::vector<int64_t>& permutation)
+{
+    return sort_permutation(permutation, std::less<>{});
+}
+
+std::vector<int64_t> find_permutation(const shape& s)
+{
```
scxiao, line 22: If the shape is broadcasted (containing more than one 0 in the strides) and transposed, can this function find the correct permutation?

src/permutation.cpp
```diff
@@ -0,0 +1,55 @@
...
+    std::vector<std::int64_t> result(s.lens().size());
+    std::iota(result.begin(), result.end(), 0);
+    std::stable_sort(result.begin(), result.end(), by(std::greater<>{}, [&](auto x) {
+                         return std::make_tuple(s.strides()[x], s.lens()[x]);
+                     }));
+    return result;
+}
+
+std::vector<int64_t> find_permutation(const std::vector<shape>& shapes)
+{
```
scxiao, line 32: In my understanding, this function is to return the permutation with the least number of input shapes, right?

src/permutation.cpp
pfultz2: It can return a correct permutation, but there could be other interpretations. This is why I usually skip broadcasted inputs.

src/permutation.cpp
pfultz2: The least number of input shapes? I dont understand that.

This function will return the permutation that is used the most in the inputs.

scxiao: Now, I understand what it does. I missed the std::less<>{} part and had thought it returns the permutation used least.

scxiao (approved):

PR #778, opened 2021-03-22
add a flag to disable cpu target in some verification unit tests

scxiao:


pfultz2:
I think I would rather have `disable_test_for` method in the `run_verify` similar to `disable_parallel_for`:

```cpp
void disable_test_for(const std::string& name, const std::vector<std::string>& tests);
````

Then you will need to add a field to `target_info` to store the disabled tests(ie `std::set<std::string> disabled_tests`). The `target_info` is available in the `verify` method by calling `get_target_info(target_name)`. Then in the main.cpp add the tests you want to disable:

```cpp
int main(int argc, const char* argv[])
{
    run_verify rv;
    rv.disable_test_for("cpu", {"test_if"});
    rv.add_validation_for("gpu", &validate_gpu);
    rv.run(argc, argv);
}
```

test/verify/run_verify.cpp
```diff
@@ -146,7 +147,10 @@ void run_verify::verify(const std::string& name, const migraphx::program& p) con
         for(const auto& tname : target_names)
         {
             target_info ti = get_target_info(tname);
-            auto t         = migraphx::make_target(tname);
+            // if tests disabled, skip running it
+            if(migraphx::contains(ti.disabled_tests, name))
```
pfultz2, line 151: This should be done in the loop above where we add the target names, so that we dont run the ref if there is no target to run.

test/verify/main.cpp
```diff
@@ -47,5 +47,6 @@ int main(int argc, const char* argv[])
 {
     run_verify rv;
     rv.add_validation_for("gpu", &validate_gpu);
+    rv.disable_test_for("cpu", {""});
```
pfultz2, line 50: It would be better to do `{}` instead. Of course, we still wont get full code coverage, but I think its fine.

pfultz2 (approved):

scxiao:
Looks like all my code changes are covered, not sure why there is a coverage percentage decrease.

PR #779, opened 2021-03-25
Bert squad example

cagery:
Adds BERT-SQuAD example using MIGraphX python interface. Sample dataset and three questions are provided.
There is Jupyter notebook and also simple python script with steps, given in Readme file.


examples/python_bert_squad_example/bert-squad-migraphx.py
```diff
@@ -0,0 +1,99 @@
...
+import sys
+
+import tokenization
+from run_onnx_squad import *
+
+print(sys.path)
+
+migx_lib_path = "/opt/rocm/lib"
+if migx_lib_path not in sys.path:
+    sys.path.append(migx_lib_path)
```
pfultz2, line 15: The MIGraphX package should setup the python path when installed, so this seems unnecessary. Perhaps we have a bug in our packaging.

pfultz2: We use these site packages for python [here](https://github.com/RadeonOpenCompute/rocm-cmake/blob/master/share/rocm/cmake/ROCMCreatePackage.cmake#L138). Perhaps, that is not the correct path on newer pythons.

pfultz2: I see in the python notebook it shows the path `/usr/local/lib/python3.6/dist-packages` which we don't set up. Let me update this in the rocm-cmake.

I think these lines should be removed, and we fix our packages.

cagery: Makes sense, thank you Paul. I will remove the lines.

turneram (commented):
These may just have to do with how I had certain things configured, but I did have to go through a few extra steps to get it to run.

1.) I had to downgrade tensorflow to 1.14.0. $ pip3 install tensorflow==1.14.0 didn't work, but it did work with
$ python3 -m pip install --upgrade https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-1.14.0-py3-none-any.whl

2.) $ apt install rccl

3.) tokenization.py was giving the error "Error: Module 'tensorflow' has no attribute 'gfile'"
I tried changing the call on line 126 to tf.io.gfile..., and it still wasn't working. But changing the import to "import tensorflow.compat.v1 as tf" resolved that error.

examples/python_bert_squad_example/bert-squad-migraphx.py
cagery: Tested .deb you proposed, it works, removed the workaround. Thanks!

cagery:
@turneram glad it worked after those changes. For RCCL, I did not need that, but I wonder where it is coming from. Do you have the error it gave for rccl?

turneram:
> @turneram glad it worked after those changes. For RCCL, I did not need that, but I wonder where it is coming from. Do you have the error it gave for rccl?

I can't recall which cell it came from, , but the error message was "ImportError: librccl.so.1: cannot open shared object file: No such file or directory"

examples/python_bert_squad_example/BERT-Squad.ipynb
```diff
@@ -0,0 +1,284 @@
...
+    "## Requirements "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip3 install -r requirements_migraphx.txt"
```
scxiao, line 30: can we rename the file name to "requrements_bertsquad.txt"? I think the requirements is for the bert model.

scxiao (commented):
There are two other files bert-squad-bench.py and bert-squad-ort.py that are not mentioned in the readme file. What are they used for?

examples/python_bert_squad_example/BERT-Squad.ipynb
cagery: Done

cagery:
> There are two other files bert-squad-bench.py and bert-squad-ort.py that are not mentioned in the readme file. What are they used for?

I should not have turned them in, reverted them back. Bench was for benchmarking the stages of inference, and ORT one is inference with onnxruntime.

cagery:
> Is there a way we can import and utilize `tokenizers` or `tokenization` libraries rather than having a large `tokenization.py` added to the repo?

As far as I see this is a bert-squad specific tokenization file, `run_onnx_squad.py` also needs this file. I have looked at the run_onnx_squad.py, and applying one of the available tokenizers (maybe `BertWordPieceTokenizer`) would require refactoring the code.

scxiao (approved):

turneram (approved):

kahmed10 (approved):

pfultz2 (approved):

examples/python_bert_squad_example/BERT-Squad.ipynb
```diff
@@ -0,0 +1,284 @@
...
+   "source": [
+    "output_dir = 'predictions'\n",
+    "os.makedirs(output_dir, exist_ok=True)\n",
+    "output_prediction_file = os.path.join(output_dir, \"predictions.json\")\n",
+    "output_nbest_file = os.path.join(output_dir, \"nbest_predictions.json\")\n",
+    "write_predictions(eval_examples, extra_data, all_results, n_best_size,\n",
+    "                  max_answer_length, True, output_prediction_file,\n",
+    "                  output_nbest_file)\n",
+    "\n",
+    "import json\n",
```
kahmed10, line 256: This is giving a deepcode CI warning. Can we remove this duplicate import?

kahmed10 (changes requested):

examples/python_bert_squad_example/bert-squad-migraphx.py
```diff
@@ -0,0 +1,90 @@
...
+
+output_dir = 'predictions'
+os.makedirs(output_dir, exist_ok=True)
+output_prediction_file = os.path.join(output_dir, "predictions.json")
+output_nbest_file = os.path.join(output_dir, "nbest_predictions.json")
+write_predictions(eval_examples, extra_data, all_results, n_best_size,
+                  max_answer_length, True, output_prediction_file,
+                  output_nbest_file)
+
+import json
```
kahmed10, line 87: here as well

kahmed10 (changes requested):

examples/python_bert_squad_example/bert-squad-migraphx.py
cagery: Just removed them both. Thanks!

kahmed10 (approved):

PR #781, opened 2021-03-29
Cpu fusions using post_ops

pfultz2:


src/targets/cpu/fuse_ops.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+MIGRAPHX_PRED_MATCHER(without_post_ops, instruction_ref ins)
+{
+    auto v = ins->get_operator().to_value();
+    return v.contains("post_ops") and v["post_ops"].empty();
+}
+
+bool workaround_dnnl_broken_post_ops(const operation& op, const operation& post_op)
+{
+    if(contains({"dnnl::dot", "dnnl::convolution"}, op.name()))
```
kahmed10, line 31: so all convolutions and gemm fusions are broken currently?

pfultz2: They most likely need to use the 3 argument form. When using an 'add' post_op for convolution or dot it does work correctly but it usually falls back on a ref implementation which is incredibly slow. Other post_ops don't seem to work consistently.

This restricts it to only use post_ops when it is binary+eltwise. It does work in some other scenarios but its hard to characterize when it works and when it doesn't.

kahmed10: I am not sure that we should call the function `workaround_dnnl_broken_post_ops` if it's just op+eltwise. Maybe something like `check_op_eltwise`? Within the function, we could also include a comment about why these fusions aren't supported currently.

pfultz2: It is a workaround. dnnl shouldn't return an algorithm when the set of ops are not supported. I would prefer to keep the name whereas `check_op_eltwise` name would not be accurate if we refine this workaround in the future.

kahmed10 (approved):

scxiao (approved):

src/include/migraphx/op/broadcast.hpp
```diff
@@ -66,6 +66,7 @@ struct broadcast
     {
         return {std::move(output_shape), std::move(args.at(0).data)};
     }
+    bool is_borrowed() const { return true; }
```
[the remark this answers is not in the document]
pfultz2: It means it uses a weak reference to the memory in the inputs(ie constructing an argument from `.data()` of another argument).

PR #784, opened 2021-03-31
Optimize resize and where operators

scxiao:


src/module.cpp
```diff
@@ -479,7 +479,8 @@ std::unordered_map<instruction_ref, std::string> module::print(
         }
         else
         {
-            var_name = this->name() + ":@" + std::to_string(count);
+            var_name = "@" + std::to_string(count);
+            var_name = this->name().empty() ? var_name : (this->name() + ":" + var_name);
```
pfultz2, line 483: I dont think this changed should be in this PR.

src/simplify_algebra.cpp
```diff
@@ -78,6 +77,162 @@ struct find_mul_conv
...
 };
 
+struct find_resize
+{
+    auto matcher() const
+    {
+        return match::name("gather")(
+            match::args(match::name("reshape")(match::arg(0).bind("data")),
+                        match::is_constant().bind("ind")),
+            match::all_of[match::outputs()](match::used_once(), pointwise()));
```
pfultz2, line 87: I dont think its necessary to check if the outputs are pointwise or that they are used once. This reshape+broadcast transformation should work even if the outputs are convolution.

src/simplify_algebra.cpp
```diff
@@ -78,6 +77,162 @@ struct find_mul_conv
...
+        }
+
+        // if ind is not constant, cannot optimize
+        std::vector<int> vec_ind;
+        auto arg_ind = ins_ind->eval();
+        if(arg_ind.empty())
+        {
+            return;
+        }
+        arg_ind.visit([&](auto v) { vec_ind.assign(v.begin(), v.end()); });
```
pfultz2, line 118: You need to check the values of `arg_ind`.  It might be easier to recompute the indices for a resize and then check that they match.

src/simplify_algebra.cpp
```diff
@@ -78,6 +77,162 @@ struct find_mul_conv
...
+            out_dims.insert(out_dims.begin() + ii + 1, scales[ii]);
+        }
+
+        auto in_rsp   = ins_rsp->inputs().front();
+        auto rsp_data = p.insert_instruction(
+            ins_rsp, migraphx::make_op("reshape", {{"dims", in_dims}}), in_rsp);
+        auto mb_rsp = p.insert_instruction(
+            ins_rsp, migraphx::make_op("multibroadcast", {{"output_lens", out_dims}}), rsp_data);
+
+        // reshape other inputs of the pointwise operator
```
pfultz2, line 163: This should be in a seperate matcher.

src/simplify_algebra.cpp
```diff
@@ -78,6 +77,162 @@ struct find_mul_conv
...
+        auto ins_ind  = r.instructions["ind"];
+
+        // if ind is not constant, cannot optimize
+        std::vector<bool> vec_ind;
+        auto arg_ind = ins_ind->eval();
+        if(arg_ind.empty())
+        {
+            return;
+        }
+        arg_ind.visit([&](auto v) { vec_ind.assign(v.begin(), v.end()); });
```
pfultz2, line 207: I think you need to check the values of the indices here as well.

test/simplify_algebra_test.cpp
```diff
@@ -2132,4 +2132,53 @@ TEST_CASE(reorder_slice_trans_diff_perm)
...
+        auto diff = m.add_instruction(migraphx::make_op("sub"), rspy, mbx);
+        std::vector<int64_t> orig_dims = {1, 1, 4, 6};
+        auto r = m.add_instruction(migraphx::make_op("reshape", {{"dims", orig_dims}}), diff);
+        m.add_return({r});
+
+        return m;
+    };
+
+    auto m1 = create_optimized_module();
+    EXPECT(m == m1);
```
pfultz2, line 2181: Can this just be `EXPECT(m == create_optimized_module())`?

pfultz2:
We should add a test for `find_where` as well.

src/simplify_algebra.cpp
```diff
@@ -78,6 +77,162 @@ struct find_mul_conv
...
+        }
+
+        auto pw_out = p.insert_instruction(ins_pw, ins_pw->get_operator(), pw_inputs);
+        std::vector<int64_t> out_dims_rsp(out_lens.begin(), out_lens.end());
+        p.replace_instruction(
+            ins_pw, migraphx::make_op("reshape", {{"dims", out_dims_rsp}}), pw_out);
+    }
+};
+
+struct find_where
```
pfultz2, line 186: This should go in simplify_reshapes as well.

src/simplify_algebra.cpp
pfultz2: `find_resize` should just transform `gather` to the reshape+broadcast+reshape. Another matcher should find a pointwise binary operator with one input of `reshape` with non-contiguous inputs, and then transform it to reshape the second argument and apply a reshape to the output.

src/module.cpp
scxiao: OK, I will move to the other PR I have.

src/simplify_algebra.cpp
scxiao: I have a constant requirement for the ind in the matcher, so I did not put the checker here.

src/simplify_algebra.cpp
```diff
@@ -78,6 +77,162 @@ struct find_mul_conv
     }
 };
 
+struct find_resize
```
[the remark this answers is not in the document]
scxiao: OK, will move to the simplify_reshape

src/simplify_algebra.cpp

src/simplify_algebra.cpp
scxiao: anyway, I added the check

src/simplify_algebra.cpp
scxiao: separate it into two optimizations.

src/simplify_algebra.cpp
scxiao: added the check

test/simplify_algebra_test.cpp
scxiao: fixed

src/simplify_algebra.cpp
scxiao: moved

src/simplify_reshapes.cpp
```diff
@@ -318,11 +318,209 @@ struct find_nested_concat
...
+    }
+};
+
+struct find_reshape_cont
+{
+    auto matcher() const
+    {
+        return match::name("reshape")(match::args(match::name("contiguous").bind("cont")),
+                                      match::used_once(),
+                                      match::any_of[match::outputs()](match::pointwise()));
```
pfultz2, line 472: Instead of matching the outputs, it should match the input to pointwise is a reshape, something like this:

```cpp
    auto matcher() const
    {
        return match::pointwise(match::either_arg(0, 1)(match::name("reshape")(
            match::args(match::name("contiguous").bind("cont")),
            match::used_once())));
    }
```

src/simplify_reshapes.cpp
```diff
@@ -318,11 +318,209 @@ struct find_nested_concat
...
+            p.replace_instruction(ins, inputs.at(1));
+        }
+    }
+};
+
+struct find_reshape_cont
+{
+    auto matcher() const
+    {
+        return match::name("reshape")(match::args(match::name("contiguous").bind("cont")),
```
pfultz2, line 470: This should check that the input is non-contiguous after skipping contiguous operators: `skip(name("contiguous"))(not_standard_shape().bind("input"))`

src/simplify_reshapes.cpp
```diff
@@ -318,11 +318,209 @@ struct find_nested_concat
...
+        }
+
+        // if ind is not constant, cannot optimize
+        std::vector<int> vec_ind;
+        auto arg_ind = ins_ind->eval();
+        if(arg_ind.empty())
+        {
+            return;
+        }
+        arg_ind.visit([&](auto v) { vec_ind.assign(v.begin(), v.end()); });
```
pfultz2, line 359: Where do you check the indices?

src/simplify_reshapes.cpp
scxiao: I just used the above matcher with a minor change.

src/simplify_reshapes.cpp
```diff
@@ -318,11 +319,227 @@ struct find_nested_concat
...
+        std::vector<bool> equal(out_shape.elements());
+        par_for(out_shape.elements(), [&](auto i) {
+            auto out_idx = out_shape.multi(i);
+            auto in_idx  = out_idx;
+            for(std::size_t ii = 0; ii < out_idx.size(); ++ii)
+            {
+                in_idx[ii] = out_idx[ii] - (out_idx[ii] % scales[ii]);
+            }
+            equal[i] = (vec_ind[out_shape.index(out_idx)] == vec_ind[out_shape.index(in_idx)]);
```
pfultz2, line 387: I dont think this is thread-safe, because `std::vector` stores each element as one bit instead of one byte.

src/simplify_reshapes.cpp
```diff
@@ -318,11 +319,227 @@ struct find_nested_concat
...
+        }
+        arg_ind.visit([&](auto v) { vec_ind.assign(v.begin(), v.end()); });
+        std::vector<bool> equal(out_shape.elements());
+        par_for(out_shape.elements(), [&](auto i) {
+            auto out_idx = out_shape.multi(i);
+            auto in_idx  = out_idx;
+            for(std::size_t ii = 0; ii < out_idx.size(); ++ii)
+            {
+                in_idx[ii] = out_idx[ii] - (out_idx[ii] % scales[ii]);
+            }
```
pfultz2, line 386: This can be written with `std::transform`.

src/simplify_reshapes.cpp
```diff
@@ -318,11 +319,227 @@ struct find_nested_concat
...
+        // if ind is not constant, cannot optimize
+        std::vector<int> vec_ind;
+        auto arg_ind = ins_ind->eval();
+        if(arg_ind.empty())
+        {
+            return;
+        }
+        arg_ind.visit([&](auto v) { vec_ind.assign(v.begin(), v.end()); });
+        std::vector<bool> equal(out_shape.elements());
+        par_for(out_shape.elements(), [&](auto i) {
```
pfultz2, line 380: So this should use the thread-id parameter and so the equal vector only needs to be the number of cores:

```cpp
std::vector<char> equal(std::thread::hardware_concurrency(), 1);
par_for(out_shape.elements(), [&](auto i, auto tid) {
    // Short-circuit
    if (equal[tid] == 0)
        return;
    // Calculate indices
    // ...
    // Assign to equal vector
    equal[tid] = ...;
});
```

pfultz2:
So I tested this with unet but the contiguous operators are still there.

src/simplify_reshapes.cpp
scxiao: Changed to use transform...

pfultz2: I meant the `for` loop can be written with `std::transform`(and it still can):

``cpp
std::transform(out_idx.begin(), out_idx.end(), scales.begin(), in_idx.begin(), [](auto out, auto scale) {
    return out - (out % scale);
});
```

File: src/simplify_reshapes.cpp [hunk 15]
```diff
@@ -318,11 +318,230 @@ struct find_nested_concat
...
+        auto arg_ind = ins_ind->eval();
+        if(arg_ind.empty())
+        {
+            return;
+        }
+        arg_ind.visit([&](auto v) { vec_ind.assign(v.begin(), v.end()); });
+        std::vector<bool> equal(out_shape.elements());
+        std::vector<int> index(out_shape.elements());
+        std::iota(index.begin(), index.end(), 0);
+        std::transform(index.begin(), index.end(), equal.begin(), [&](auto i) {
```
pfultz2, line 381: This should use `all_of` so it can short-circuit.

src/simplify_reshapes.cpp
scxiao: fixed.

scxiao:
> unet

@pfultz2 So where is the unet model?

pfultz2:
> So where is the unet model?

https://github.com/mvermeulen/rocm-migraphx/blob/master/saved-models/onnx-misc/taau_low_res_downsample_d2s_for_infer_time_fp16_opset11.onnx

You need to run with `--input-dim @inputs 1 1920 1080 8`.

scxiao:
> > So where is the unet model?
> 
> https://github.com/mvermeulen/rocm-migraphx/blob/master/saved-models/onnx-misc/taau_low_res_downsample_d2s_for_infer_time_fp16_opset11.onnx
> 
> You need to run with `--input-dim @inputs 1 1920 1080 8`.

Looks like the matcher in the comments:

```
    auto matcher() const
    {
        return match::pointwise(
            match::nargs(2),
            match::either_arg(0, 1)(
                match::name("reshape")(match::args(match::name("contiguous").bind("cont")))
                    .bind("rsp"),
                match::used_once()));
    }
```

cannot match the pointwise(reshape(contigusou))), but the following works:

```
        return match::name("reshape")(match::args(match::name("contiguous").bind("cont")),
                                      match::used_once(),
                                      match::any_of[match::outputs()](match::pointwise()));        
```

To me, both should work, not sure what is the reason for that.

scxiao:
> > > So where is the unet model?
> > 
> > 
> > https://github.com/mvermeulen/rocm-migraphx/blob/master/saved-models/onnx-misc/taau_low_res_downsample_d2s_for_infer_time_fp16_opset11.onnx
> > You need to run with `--input-dim @inputs 1 1920 1080 8`.
> 
> Looks like the matcher in the comments:
> 
> ```
>     auto matcher() const
>     {
>         return match::pointwise(
>             match::nargs(2),
>             match::either_arg(0, 1)(
>                 match::name("reshape")(match::args(match::name("contiguous").bind("cont")))
>                     .bind("rsp"),
>                 match::used_once()));
>     }
> ```
> 
> cannot match the pointwise(reshape(contigusou))), but the following works:
> 
> ```
>         return match::name("reshape")(match::args(match::name("contiguous").bind("cont")),
>                                       match::used_once(),
>                                       match::any_of[match::outputs()](match::pointwise()));        
> ```
> 
> To me, both should work (actually, both work for the unit test ```reshape_cont```), not sure what is the reason for that.

pfultz2:
> To me, both should work(actually, both work for the unit test reshape_cont), not sure what is the reason for that.

The `used_once` is applied to the argument that is not a `reshape` input. The `either_arg` matcher takes two inputs for either argument. It should be `match::any`:

```cpp
    auto matcher() const
    {
        return match::pointwise(
            match::nargs(2),
            match::either_arg(0, 1)(
                match::name("reshape")(match::args(match::name("contiguous").bind("cont")))
                    .bind("rsp"),
                match::any()));
    }
```

I also dont think `used_once` is necessary here either. This fixes the issue in unet.

scxiao:
> > To me, both should work(actually, both work for the unit test reshape_cont), not sure what is the reason for that.
> 
> The `used_once` is applied to the argument that is not a `reshape` input. The `either_arg` matcher takes two inputs for either argument. It should be `match::any`:
> 
> ```c++
>     auto matcher() const
>     {
>         return match::pointwise(
>             match::nargs(2),
>             match::either_arg(0, 1)(
>                 match::name("reshape")(match::args(match::name("contiguous").bind("cont")))
>                     .bind("rsp"),
>                 match::any()));
>     }
> ```
> 
> I also dont think `used_once` is necessary here either. This fixes the issue in unet.

Thanks. fixed it

src/simplify_reshapes.cpp
```diff
@@ -318,11 +318,229 @@ struct find_nested_concat
...
+            match::args(match::name("reshape")(match::arg(0)(match::name("concat").bind("data"))),
+                        match::is_constant().bind("ind")));
+    }
+
+    void apply(module& p, match::matcher_result r) const
+    {
+        auto ins     = r.result;
+        auto concat  = r.instructions["data"];
+        auto ins_ind = r.instructions["ind"];
+        // if ind is not constant, cannot optimize
```
kahmed10, line 434: why are we unable to check this within the matcher?

scxiao: We already have the constant check in the matcher. Just remove this one.

kahmed10: ok, we can probably remove the comment above this as well

scxiao: Thanks. removed

kahmed10 (approved):

src/simplify_reshapes.cpp
```diff
@@ -318,11 +318,224 @@ struct find_nested_concat
...
+        {
+            if(scales[ii] == 1 or in_dims[ii] == 1)
+            {
+                continue;
+            }
+
+            out_dims[ii] = in_dims[ii];
+            in_dims.insert(in_dims.begin() + ii + 1, 1);
+            out_dims.insert(out_dims.begin() + ii + 1, scales[ii]);
+        }
```
pfultz2, line 407: It seems like it would be better to zip the scales and dims together:

```cpp
std::vector<std::pair<std::size_t, std::size_t>> dim_scales;
std::transform(in_lens.begin(), in_lens.end(), out_lens.begin(), dim_scales.begin(), [](auto x, auto y) {
    return std::make_pair(x, y / x);
});
```

Then you could create the vector of dims by a simple iteration:

```cpp
for(auto&& p:dim_scales)
{
    in_dims.push_back(p.first);
    out_dims.push_back(p.first);
    if(p.first == 1 or p.second == 1)
        continue;
    in_dims.push_back(1);
    out_dims.push_back(p.second);
}
```

This would also avoid inserting in the middle of the vector since that is slow.

scxiao:
@pfultz2 Can you review this PR when you get a chance?

pfultz2:
> Can you review this PR when you get a chance?

Did you see my comments above?

src/simplify_reshapes.cpp
scxiao: fixed

src/simplify_reshapes.cpp
```diff
@@ -318,11 +319,233 @@ struct find_nested_concat
...
+                              in_idx.begin(),
+                              [&](auto io, auto scale) { return io - (io % scale); });
+               return vec_ind[i] == vec_ind[out_shape.index(in_idx)];
+           }))
+        {
+            return;
+        }
+
+        // wrap up shapes for multibroadcast
+        std::vector<std::pair<std::size_t, std::size_t>> dim_scales;
```
pfultz2, line 396: Looking at this further, it seems it would be better to zip the in_lens and out_lens into one vector, and this should get rid of some of these intermediate vectors such as `is_multi`, `scales`, and `dim_scales`. Maybe we can fix that in a later PR and get this merged in.

pfultz2 (approved):

PR #789, opened 2021-04-05
Super resolution example

cagery:
This PR introduces super resolution example, which scales low-res image to high-res, using [sub-pixel conv layer](https://arxiv.org/abs/1609.05158).

Provides requirements file and notebook to run the example.


kahmed10:
My only concern with this example is that the original image is already smaller than the input, so we are calling `resize_cover()` to first upscale the input. It would be better to start with an image that is the expected output size, downscale, then use the network to upscale. You could also possibly display the PSNR as well, but that's up to you.

kahmed10 (approved):

cagery:
@kahmed10 Thanks for the suggestion!
I have added PSNR. I am making PSNR measurements on output image vs upscaled input. 
The new input image is not the expected output size but is larger than the expected input dims. 400x400. 
Let me know what you think.

turneram (approved):

scxiao (commented):
Can you add some information like the "Steps" as that in the BERT model PR?

cagery:
> Can you add some information like the "Steps" as that in the BERT model PR?

Just added the steps, thanks Shucai!

pfultz2 (approved):

scxiao (approved):

PR #796, opened 2021-04-09
Create .o files in the build directory

pfultz2:


scxiao (commented):
I think we need to remove those .o files from the repo, and commit that, right?

pfultz2:
> I think we need to remove those .o files from the repo, and commit that, right?

They have already been removed.

scxiao (approved):

scxiao:
@mvermeulen I think this PR can be merged now.

PR #797, opened 2021-04-09
Prefix scan operator

turneram:
This PR adds the prefix_scan operator class. Prefix_scan_sum is made to the same specifications as ONNX CumSum operator. The prefix_scan class can be extended for other operators than sum.


pfultz2:
There seems to be .o and .pyc files added to this PR and should be removed.

src/onnx/parse_prefix_scan.cpp
```diff
@@ -0,0 +1,70 @@
...
+    bool reverse = false;
+
+    if(contains(info.attributes, "exclusive"))
+    {
+        exclusive = parser.parse_value(info.attributes.at("exclusive")).at<bool>();
+    }
+
+    if(contains(info.attributes, "reverse"))
+    {
+        reverse = parser.parse_value(info.attributes.at("reverse")).at<bool>();
```
pfultz2, line 46: Can we use a `transpose` to do the reverse?

src/onnx/parse_prefix_scan.cpp
```diff
@@ -0,0 +1,70 @@
...
+instruction_ref parse_prefix_scan_oper(const std::string& op_name,
+                                  const onnx_parser& parser,
+                                  onnx_parser::node_info info,
+                                  std::vector<instruction_ref> args)
+{
+    int64_t axis = 0; 
+    migraphx::argument in = args[1]->eval();
+    if (in.empty())
+    {
+        std::cout << "No axis given. Using axis=0" << std::endl;
```
pfultz2, line 19: Shouldn't this be an error if the axis is missing?

src/onnx/parse_prefix_scan.cpp
```diff
@@ -0,0 +1,70 @@
...
+        in.visit([&](auto input) { axis_in.assign(input.begin(), input.end()); });
+        axis = axis_in[0];
+        auto n_dims = args[0]->get_shape().lens().size();
+        if (axis >= n_dims or axis < -int64_t(n_dims)) 
+        {
+            MIGRAPHX_THROW("Axis " + std::to_string(axis) + " is out of bounds for shape with " + std::to_string(n_dims) + " dimensions");
+        }
+        if (axis < 0 and axis >= -int64_t(n_dims)) 
+        {
+            axis += n_dims;
```
pfultz2, line 33: The negative axis should be handled by normalization.

src/onnx/parse_prefix_scan.cpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        std::cout << "No axis given. Using axis=0" << std::endl;
+    }
+    else 
+    {
+        std::vector<std::size_t> axis_in;
+        in.visit([&](auto input) { axis_in.assign(input.begin(), input.end()); });
+        axis = axis_in[0];
+        auto n_dims = args[0]->get_shape().lens().size();
+        if (axis >= n_dims or axis < -int64_t(n_dims))
```
pfultz2, line 27: This should be checked in normalize_compute_shape.

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                });
+            });
+        }
+    }
+    
+    argument compute(const shape& output_shape, std::vector<argument> args) const
+    {
+        argument result{output_shape};
+        this->prefix_scan(args[0], result);
```
pfultz2, line 95: This could be done in-place:

```cpp
argument result = args[0];
this->prefix_scan(result);
```

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,111 @@
...
+        normalize["axis"] = value::array{normalize_attribute::include_min};
+        return {{"normalize_axes", normalize}};
+    }
+
+    shape normalize_compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(1);
+        auto s          = inputs.at(0);
+
+        return {s.type(), s.lens()};
```
pfultz2, line 44: This should do `return s`.

src/include/migraphx/op/prefix_scan_op.hpp
pfultz2: Actually, this could be done in parallel by slicing the shape and then doing `partial_sum` in the inner loop since we are only doing this across one axis. A rough idea of how this can be done:

```cpp
argument result = args[0];
auto s = result.get_shape();
auto lens = s.lens();
auto size = lens[axis];
lens[axis] = 1;
auto slice = shape{s.type(), lens, s.strides()};
    par_for(slice.elements(), [&](auto i) {
        // Get pointer to first element for prefix_sum
        auto* start = output.data()+slice.index(i);
        std::partial_sum(start, start+size, start, self.op());
    });
});
```

This will need to be adjusted for an exclusive scan.

pfultz2: also the reverse can be done with `std::partial_sum(std::make_reverse_iterator(start), std::make_reverse_iterator(start+size), start, self.op())`

src/onnx/parse_prefix_scan.cpp
pfultz2: Nevermind, I dont think thats possible.

src/onnx/parse_prefix_scan.cpp
turneram: The only way I've gotten the onnx CumSum node in gen_onnx.py to actually work is to pass in 'axis' as a tensor_value_info rather than an actual tensor or constant. So when I run onnx_test.cpp, having it throw an error causes the test to fail. If I can figure out how to get an actual value for 'axis' into the onnx CumSum node, then I'll fix this.

src/include/migraphx/op/prefix_scan_op.hpp
turneram: > This could be done in-place:
> 
> ```c++
> argument result = args[0];
> this->prefix_scan(result);
> ```

Exclusive mode either requires two copies or either a preliminary or follow-up pass to shift the elements.

turneram: > Actually, this could be done in parallel by slicing the shape and then doing `std::partial_sum` in the inner loop since we are only doing this across one axis. A rough idea of how this can be done:
> 
> ```c++
> argument result = args[0];
> auto s = result.get_shape();
> auto lens = s.lens();
> auto size = lens[axis];
> lens[axis] = 1;
> auto slice = shape{s.type(), lens, s.strides()};
> result.visit([&](auto output) {
>     par_for(slice.elements(), [&](auto i) {
>         // Get pointer to first element for prefix_sum
>         auto* start = output.data()+slice.index(i);
>         std::partial_sum(start, start+size, start, self.op());
>     });
> });
> ```
> 
> This will need to be adjusted for an exclusive scan.

The issue that seems to be happening with using partial_sum is that the elements are only contiguous when `axis = dimensions - 1`. 

Also, I think we can parallelize along all axes other than the one being summed along. So for 3-D and axis=0, both inner loops can be parallelized; 3-D and axis=1, the outer and inner loops can be parallelized; or 3-D and axis=2, the two outer loops can be parallelized. I have an idea for how to do this, but I think it would require having a maximum number of dimensions and padding tensors smaller than that with size 1 dimensions. Can we assume 3 is the max?

pfultz2: > The issue that seems to be happening with using partial_sum is that the elements are only contiguous when axis = dimensions - 1.

That's right, it only works if the strides are 1 for the dimension. We need to slice it into another tensor_view:

```cpp
argument result = args[0];
auto s = result.get_shape();
auto slice = shape{s.type(), {s.lens()[axis]}, {s.strides()[axis]}};
auto lens = s.lens();
lens[axis] = 1;
auto batch = shape{s.type(), lens, s.strides()};
    using type = decltype(output);
    par_for(batch.elements(), [&](auto i) {
        auto* start = output.data()+batch.index(i);
        type x{slice, start};
        std::partial_sum(x.begin(), x.end(), x.begin(), self.op());
    });
});
```

> Can we assume 3 is the max?

No we can't.

src/onnx/parse_prefix_scan.cpp
pfultz2: @kahmed10 @scxiao Are you able to help Alan generate the correct onnx?

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,101 @@
...
+            using type = decltype(output);
+            par_for(batch.elements(), [&](auto i) {
+                auto* start = output.data()+batch.index(i);
+                type x{slice, start};
+                if (reverse) {
+                    if (exclusive) {
+                        type y{slice_ex, start+1};
+                        type z{slice_ex, start};
+                        std::copy(y.begin(), y.end(), z.begin());
+                        *z.end() = 0;
```
pfultz2, line 72: You can't deference an end iterator. This should probably be:

```cpp
std::copy(x.begin()+1, x.end(), x.begin());
x.back() = 0;
```

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                    std::partial_sum(std::make_reverse_iterator(x.end()), std::make_reverse_iterator(x.begin()), std::make_reverse_iterator(x.end()), self.op());
+                }
+                else {
+                    if (exclusive) {
+                        type y{slice_ex, start};
+                        std::copy_backward(y.begin(), y.end(), x.end());
+                        *x.begin() = 0;
```
pfultz2, line 80: I am not sure the copy is needed. For exclusive case, couldn't you do something like `std::partial_sum(x.begin(), x.end()-1, x.begin()+1, self.op())` and `x.front() = 0`?

src/include/migraphx/shape_for_each.hpp
```diff
@@ -29,6 +29,27 @@ void shape_for_each(const migraphx::shape& s, F f)
...
+                       ss.strides().end(),
+                       ss.lens().begin(),
+                       indices.begin(),
+                       [&](std::size_t stride, std::size_t len) {
+                           assert(len > 0 and stride > 0);
+                           return ((i - 1) / stride) % len;
+                       });
+        call(indices);
+    }
+}
```
pfultz2, line 51: I think this can be removed now.

src/include/migraphx/op/prefix_scan_op.hpp
turneram: I was getting errors when using pointer arithmetic on tensor_view iterators, but just tried it with pre-increment/decrement and that works.

src/include/migraphx/op/prefix_scan_op.hpp
turneram: I tried that and I don't believe it works. 
For a quick example,
```
#include <iostream>
#include <numeric>
#include <vector>

int main()
{
  std::vector<int> x = {1, 2, 3, 4, 5};
  std::partial_sum(x.begin(), x.end()-1, x.begin()+1);
  x.front() = 0;
  for (auto& xi: x) {
      std::cout << xi << ", ";
  }
  return 0;
}
```

The output should be {0, 1, 3, 6, 10}, but it gives {0, 1, 2, 4, 8}

src/include/migraphx/shape_for_each.hpp
turneram: Removed

src/include/migraphx/op/prefix_scan_op.hpp
pfultz2: I see the operators for `+` are not implemented correctly.

src/include/migraphx/op/prefix_scan_op.hpp
pfultz2: You're right, we overwrite the values we need for partial_sum.

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,97 @@
...
+        normalize["axis"] = value::array{normalize_attribute::include_min};
+        return {{"normalize_axes", normalize}};
+    }
+
+    shape normalize_compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(1);
+        auto s          = inputs.at(0);
+        auto n_dims = s.lens().size();
+        if (axis >= n_dims or axis < -int64_t(n_dims))
```
pfultz2, line 44: I dont think the `axis < -int64_t(n_dims)` condition is necessary since normalization will make it always positive.

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    shape normalize_compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(1);
+        auto s          = inputs.at(0);
+        auto n_dims = s.lens().size();
+        if (axis >= n_dims or axis < -int64_t(n_dims)) 
+        {
+            MIGRAPHX_THROW("Axis " + std::to_string(axis) + " is out of bounds for shape with " + std::to_string(n_dims) + " dimensions");
```
pfultz2, line 46: It would be good to add a test case for this error in op_shape_test.cpp.

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include <migraphx/config.hpp>
+#include <migraphx/value.hpp>
+#include <migraphx/op/normalize_attribute.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+template <class Derived>
+struct prefix_scan_op : op_name<Derived>
```
kahmed10, line 18: is there a reason why we need to call it `prefix_scan_op` instead of `prefix_scan`?

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,104 @@
...
+{
+    int64_t axis;
```
scxiao, line 20: I think the axis had better be an input, not an axis. Then axis input for this operator does not have to be constant. 
The reason we put axis as an attribute for other operators (like gather, concat) is that different axis values generates different output shapes, and migraphx cannot support that. But this operator is different. Its output shape is always the same, so the axis can be dynamic.

src/include/migraphx/op/prefix_scan_op.hpp
pfultz2: Yea, if we want to do prefix_scan with other operators besides sum, then this class can be reused.

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,104 @@
...
+    value attributes() const
+    {
+        value normalize;
+        normalize["axis"] = value::array{normalize_attribute::include_min};
+        return {{"normalize_axes", normalize}};
+    }
+
+    shape normalize_compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(1);
```
scxiao, line 40: we may need a standard() check here since it looks like algorithm below assume this.

src/include/migraphx/operators.hpp
```diff
@@ -62,6 +62,7 @@
 #include <migraphx/op/quant_convolution.hpp>
 #include <migraphx/op/quant_dot.hpp>
 #include <migraphx/op/pow.hpp>
+#include <migraphx/op/prefix_scan_sum.hpp>
```
scxiao, line 65: we can move the file name above a little, so they are in alphabetical order. The "pow.hpp" should also be moved

src/include/migraphx/op/prefix_scan_sum.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace op {
+
+struct prefix_scan_sum : prefix_scan_op<prefix_scan_sum>
+{
+    prefix_scan_sum() {}
+    prefix_scan_sum(int64_t ax) : prefix_scan_op(ax) {}
+    prefix_scan_sum(int64_t ax, bool excl) : prefix_scan_op(ax, excl) {}
+    prefix_scan_sum(int64_t ax, bool excl, bool rev) : prefix_scan_op(ax, excl, rev) {}
+
+    auto op() const
```
scxiao, line 22: we can add a name function function to indicate it is the "cumsum" operator.

scxiao (commented):
Looks like we do not have a GPU implementation for this op yet

src/onnx/parse_prefix_scan.cpp
scxiao: I refined the script and committed the changes. Not sure whether that is what you want.

scxiao (commented):
Since we need the axis to be a constant, which is NOT that specified in the ONNX specification, none of the onnx backend unit tests can run as:

```
ERROR: test_cumsum_1d_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 248, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 278, in run
    prepared_model = self.backend.prepare(model, device)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 105, in prepare
    return cls.prepare(bin, device, **kwargs)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 92, in prepare
    inf = migraphx.parse_onnx_buffer(model)
RuntimeError: /home/scxiao/Workplace/projects/AMDMIGraphX/src/onnx/checks.cpp:12: check_arg_empty: PARSE_PREFIX_SCAN: axis - dynamic shape not supported

======================================================================
ERROR: test_cumsum_1d_exclusive_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 248, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 278, in run
    prepared_model = self.backend.prepare(model, device)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 105, in prepare
    return cls.prepare(bin, device, **kwargs)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 92, in prepare
    inf = migraphx.parse_onnx_buffer(model)
RuntimeError: /home/scxiao/Workplace/projects/AMDMIGraphX/src/onnx/checks.cpp:12: check_arg_empty: PARSE_PREFIX_SCAN: axis - dynamic shape not supported

======================================================================
ERROR: test_cumsum_1d_reverse_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 248, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 278, in run
    prepared_model = self.backend.prepare(model, device)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 105, in prepare
    return cls.prepare(bin, device, **kwargs)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 92, in prepare
    inf = migraphx.parse_onnx_buffer(model)
RuntimeError: /home/scxiao/Workplace/projects/AMDMIGraphX/src/onnx/checks.cpp:12: check_arg_empty: PARSE_PREFIX_SCAN: axis - dynamic shape not supported

======================================================================
ERROR: test_cumsum_1d_reverse_exclusive_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 248, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 278, in run
    prepared_model = self.backend.prepare(model, device)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 105, in prepare
    return cls.prepare(bin, device, **kwargs)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 92, in prepare
    inf = migraphx.parse_onnx_buffer(model)
RuntimeError: /home/scxiao/Workplace/projects/AMDMIGraphX/src/onnx/checks.cpp:12: check_arg_empty: PARSE_PREFIX_SCAN: axis - dynamic shape not supported

======================================================================
ERROR: test_cumsum_2d_axis_0_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 248, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 278, in run
    prepared_model = self.backend.prepare(model, device)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 105, in prepare
    return cls.prepare(bin, device, **kwargs)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 92, in prepare
    inf = migraphx.parse_onnx_buffer(model)
RuntimeError: /home/scxiao/Workplace/projects/AMDMIGraphX/src/onnx/checks.cpp:12: check_arg_empty: PARSE_PREFIX_SCAN: axis - dynamic shape not supported

======================================================================
ERROR: test_cumsum_2d_axis_1_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 248, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 278, in run
    prepared_model = self.backend.prepare(model, device)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 105, in prepare
    return cls.prepare(bin, device, **kwargs)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 92, in prepare
    inf = migraphx.parse_onnx_buffer(model)
RuntimeError: /home/scxiao/Workplace/projects/AMDMIGraphX/src/onnx/checks.cpp:12: check_arg_empty: PARSE_PREFIX_SCAN: axis - dynamic shape not supported

======================================================================
ERROR: test_cumsum_2d_negative_axis_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 248, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/home/scxiao/anaconda3/lib/python3.6/site-packages/onnx/backend/test/runner/__init__.py", line 278, in run
    prepared_model = self.backend.prepare(model, device)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 105, in prepare
    return cls.prepare(bin, device, **kwargs)
  File "/home/scxiao/Workplace/projects/AMDMIGraphX/build/lib/onnx_migraphx/backend.py", line 92, in prepare
    inf = migraphx.parse_onnx_buffer(model)
RuntimeError: /home/scxiao/Workplace/projects/AMDMIGraphX/src/onnx/checks.cpp:12: check_arg_empty: PARSE_PREFIX_SCAN: axis - dynamic shape not supported
```

We cannot verify the correctness of our implementation.

pfultz2:
> We cannot verify the correctness of our implementation.

We dont have a GPU backend for this so even if we were to use a dynamic axis, we couldn't run it. We should tackle the GPU implementation in a later PR as it is fairly involved.

Also, there are several tests to verify the implementation is correct, so its not like this is untested.

src/include/migraphx/op/prefix_scan_sum.hpp
pfultz2: The is named `prefix_scan_sum` so it seems obvious that this will do a sum.

src/include/migraphx/op/prefix_scan_op.hpp
scxiao: If the input shape is broadcasted and result is assigned as ```argument result = args[0];``` as below, the result will not have enough buffer to store the results. 

For the transposed shape, I think we might need to tune the axis (I am not very sure, but it might be necessary)

pfultz2: > If the input shape is broadcasted and result is assigned

We can probably add a `not_broadcasted` as I dont think it makes much sense to do a prefix sum on a broadcasted tensor.

> For the transposed shape, I think we might need to tune the axis (I am not very sure, but it might be necessary)

No we don't for the same reason we don't need to tune the axis for softmax or reduce.

kahmed10 (approved):

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -0,0 +1,104 @@
...
+        auto s      = inputs.at(0);
+        auto n_dims = s.lens().size();
+        if(axis >= n_dims)
+        {
+            MIGRAPHX_THROW("Axis " + std::to_string(axis) + " is out of bounds for shape with " +
```
pfultz2, line 45: Can you add a test for this line in op_shape_test.cpp?

turneram: Can you take a look at the test I added to op_shape _test in this PR? I'm not sure why that isn't covering this.

pfultz2 (approved):

test/ref_ops_test.cpp
```diff
@@ -3593,4 +3593,369 @@ TEST_CASE(unsqueeze_test)
...
+        migraphx::make_op("prefix_scan_sum", {{"axis", 0}, {"exclusive", true}, {"reverse", true}}),
+        l0);
+    p.compile(migraphx::ref::target{});
+    auto result = p.eval({}).back();
+    std::vector<float> results_vector;
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> gold{20.0, 18.0, 15.0, 11.0, 6.0, 0.0};
+    EXPECT(results_vector == gold);
+}
+
```
scxiao, line 3960: Now the unit test are sorted in the alphabetical order, can you move the tests so they are in order in the file?

turneram: Fixed

scxiao (approved):

mvermeulen:
@turneram seems to have been failing checks, even when restarted.

turneram:
> @turneram seems to have been failing checks, even when restarted.

@mvermeulen Not entirely sure why that's happening. It appears to be a tidy warning from `/src/include/migraphx/verify.hpp`. Within "files changed", it shows up at the bottom under "Unchanged files with check annotations" with the error: 
`no member named 'begin' in 'migraphx::MIGRAPHX_INLINE_NS::shape' [clang-diagnostic-error]`

I think the issue started after the most recent merge from develop.

pfultz2:
The error is from passing shape to `verify_range`:

```
test/ref_ops_test.cpp:3862:22: note: in instantiation of function template specialization 'migraphx::MIGRAPHX_INLINE_NS::verify_range<std::vector<float, std::allocator<float>>, migraphx::MIGRAPHX_INLINE_NS::shape>' requested here
    EXPECT(migraphx::verify_range(results_vector, s));
                     ^
```

Which is a line you changed in your PR [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/797/files#diff-cc9d7124ad9c037970afe103cb8e14283bbaf86217b080652850a9b8912e064dR3862).

turneram:
> The error is from passing shape to `verify_range`:
> 
> ```
> test/ref_ops_test.cpp:3862:22: note: in instantiation of function template specialization 'migraphx::MIGRAPHX_INLINE_NS::verify_range<std::vector<float, std::allocator<float>>, migraphx::MIGRAPHX_INLINE_NS::shape>' requested here
>     EXPECT(migraphx::verify_range(results_vector, s));
>                      ^
> ```
> 
> Which is a line you changed in your PR [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/797/files#diff-cc9d7124ad9c037970afe103cb8e14283bbaf86217b080652850a9b8912e064dR3862).
> 
> This is causing compilation errors on the other stages.

Looks like the last few lines of softmax_test got pasted over the last few lines of sub_test at some point. Not sure how that happened, but should be fixed now.

PR #798, opened 2021-04-09
Onnx 1.8 support

scxiao:


Dockerfile
```diff
@@ -80,6 +78,9 @@ RUN cget -p $PREFIX install ccache@v4.1
 # Install newer cmake for onnx runtime
 RUN cget -p /opt/cmake install kitware/cmake@v3.13.0
 
+# Install onnx 1.8.1
+RUN pip3 install onnx numpy typing pytest
```
pfultz2, line 82: Isn't this already installed in the `install_prereqs.sh` script?

src/include/migraphx/op/logsoftmax.hpp
```diff
@@ -12,7 +12,7 @@ namespace op {
 
 struct logsoftmax
 {
-    int64_t axis = 1;
+    int64_t axis = -1;
```
pfultz2, line 15: Why are we changing the default axis?

src/onnx/parse_squeeze.cpp
```diff
@@ -0,0 +1,62 @@
...
+    {
+        return {{"Squeeze", "squeeze"}, {"Unsqueeze", "unsqueeze"}};
+    }
+
+    bool needs_contiguous(const std::string& op_name) const
+    {
+        return contains({"squeeze", "unsqueeze"}, op_name);
+    }
+
+    void assign_axes(const operation& op, const std::vector<int64_t>& axes) const
```
pfultz2, line 23: This function looks like a nop. What is it supposed to be doing?

src/onnx/parse_squeeze.cpp
```diff
@@ -0,0 +1,62 @@
...
+namespace onnx {
+
+struct parse_squeeze : op_parser<parse_squeeze>
+{
+    std::vector<op_desc> operators() const
+    {
+        return {{"Squeeze", "squeeze"}, {"Unsqueeze", "unsqueeze"}};
+    }
+
+    bool needs_contiguous(const std::string& op_name) const
```
pfultz2, line 18: All the operators listed need contiguous, so this function doesn't seem necessary.

Dockerfile
pfultz2: Also, the pinned versions should not be dropped.

scxiao: Yes, you are right. I removed this line and the line for 1.7.0 in the Dockerfile. 

Actually, I am wondering whether we can install two python versions (like 3.6 and 3.7), and install different onnx versions on them?

src/include/migraphx/op/logsoftmax.hpp
scxiao: The onnx 1.8.1 changed the default value from 1 to -1.

src/onnx/parse_squeeze.cpp
scxiao: I forgot a return value. Fixed.

src/onnx/parse_squeeze.cpp
scxiao: fixed.

src/include/migraphx/op/logsoftmax.hpp
pfultz2: Then this should be changed in the onnx parser.

Dockerfile
pfultz2: Is the reason to install two different python version so we can have two different onnx versions installed? As we can create two different venvs to install onnx. I dont think we should do that in this PR, but we could look into cmake automatically setting up the venvs so the the tests will know where to find them.

src/onnx/parse_quantizelinear.cpp
```diff
@@ -72,6 +72,21 @@ struct parse_quantizelinear : op_parser<parse_quantizelinear>
...
+        if(min_arg->get_shape() != s)
+        {
+            min_arg = info.add_instruction(make_op("multibroadcast", {{"output_lens", out_lens}}),
+                                           min_arg);
+        }
+        if(max_arg->get_shape() != s)
+        {
+            max_arg = info.add_instruction(make_op("multibroadcast", {{"output_lens", out_lens}}),
+                                           max_arg);
+        }
```
pfultz2, line 88: This can use `add_broadcastable_binary_op` instead.

src/onnx/parse_reduce_op.cpp
```diff
@@ -15,15 +16,42 @@ instruction_ref parse_reduce_oper(const std::string& op_name,
...
+        arg_axes.visit([&](auto s) { axes.assign(s.begin(), s.end()); });
+    }
+    else if(contains(info.attributes, "axes"))
     {
         axes.clear();
         auto&& attr_axes = info.attributes["axes"].ints();
         axes             = std::vector<int64_t>(attr_axes.begin(), attr_axes.end());
     }
 
+    int noop_with_empty_axes = 0;
```
pfultz2, line 34: Shouldn't this be a `bool`?

src/onnx/parse_squeeze.cpp
```diff
@@ -0,0 +1,57 @@
...
+    {
+        auto v = op.to_value();
+        for(auto&& x : v)
+        {
+            if(x.get_key() == "axes")
+            {
+                x = axes;
+                break;
+            }
+        }
```
pfultz2, line 28: Why not write this as `v["axes"] = axes`?

Dockerfile
scxiao: yes, we want two different onnx versions installed, so we can test both onnx versions.
OK, we can push later to add that.

src/include/migraphx/op/logsoftmax.hpp
scxiao: So, we will set different default values for different onnx versions, right?

src/onnx/parse_quantizelinear.cpp
scxiao: The "clip" here is not a binary, we add another function for the broadcasting?

src/onnx/parse_reduce_op.cpp
scxiao: The ONNX operator uses an int type, but I can change it to bool here.

src/onnx/parse_squeeze.cpp
scxiao: changed

src/include/migraphx/op/logsoftmax.hpp
pfultz2: > So, we will set different default values for different onnx versions, right?

If we want backwards compatibility to 1.7 then we will probably need to do this. More importantly, onnx may change default values but that should not necessarily change our default values. Rather, the onnx parser can decide a default value that is different than ours.

Is the onnx version saved in the onnx file?

tools/install_prereqs.sh
```diff
@@ -6,7 +6,7 @@
 apt update && apt install -y python3-pip rocm-cmake rocblas miopen-hip openmp-extras
 
 # install onnx package for unit tests
-pip3 install onnx==1.7.0 numpy==1.18.5 typing==3.7.4 pytest==6.0.1
+pip3 install onnx==1.8.1 numpy==1.18.5 typing==3.7.4 pytest==6.0.1 packaging
```
pfultz2, line 9: `packaging` version should be pinned.

src/onnx/onnx_parser.cpp
```diff
@@ -29,6 +29,12 @@ static onnx_parser::attribute_map get_attributes(const onnx::NodeProto& node)
 static literal
 create_literal(shape::type_t shape_type, const std::vector<size_t>& dims, const char* data)
 {
+    // empty input
+    auto elem_num =
+        std::accumulate(dims.begin(), dims.end(), std::size_t(1), std::multiplies<std::size_t>());
+    if(elem_num == 0)
+        return {};
```
pfultz2, line 36: There should be a test for this.

scxiao:
@pfultz2 @kahmed10 @cagery Can you review this PR?

Dockerfile
```diff
@@ -73,6 +71,8 @@ ADD requirements.txt /requirements.txt
 COPY ./tools/install_prereqs.sh /
 RUN /install_prereqs.sh /usr/local / && rm /install_prereqs.sh
 
+RUN pip3 install onnx==1.8.1 numpy==1.18.5 typing==3.7.4 pytest==6.0.1 packaging
```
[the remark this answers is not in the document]
kahmed10: we should probably refactor our Dockerfile as a whole to make sure there aren't duplicates in install_prereqs.sh

src/onnx/onnx_parser.cpp
```diff
@@ -238,6 +257,26 @@ void onnx_parser::parse_from(const void* data, std::size_t size)
...
 
+void onnx_parser::parse_opset_version(const onnx::ModelProto& model)
+{
+    const auto& opset_import = model.opset_import();
+    int64_t version          = -1;
+    for(const auto& opset : opset_import)
+    {
+        if(opset.has_version())
+        {
+            version = (version < opset.version()) ? opset.version() : version;
```
kahmed10, line 268: can we rewrite this as `version = std::max(version, opset.version())`?

src/onnx/onnx_parser.cpp
```diff
@@ -238,6 +257,26 @@ void onnx_parser::parse_from(const void* data, std::size_t size)
     }
 }
 
+void onnx_parser::parse_opset_version(const onnx::ModelProto& model)
```
pfultz2, line 260: Can we make this a static function that returns the opset version: `std::size_t get_opset_version(const onnx::ModelProto& model)`? And then we can set the version with `opset_version = get_opset_version(model)`.

tools/install_prereqs.sh
scxiao: pinned

Dockerfile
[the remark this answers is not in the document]
scxiao: just checked the docker file, there should be no duplicates anymore.

src/onnx/onnx_parser.cpp
scxiao: added one

src/onnx/onnx_parser.cpp
scxiao: changed accordly

src/onnx/onnx_parser.cpp
scxiao: changed

pfultz2 (approved):

kahmed10 (approved):

scxiao:
@cagery Can you review this PR when you get a chance? We want to get it in today, thanks!

cagery (approved):

PR #802, opened 2021-04-15
Inline subgraph

scxiao:


src/include/migraphx/op/if_op.hpp
```diff
@@ -39,10 +39,12 @@ struct if_op
     }
 
     argument compute(
+        context& ctx,
```
pfultz2, line 42: Context should not be required here.

kahmed10 (approved):

src/include/migraphx/inline_subgraph.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include <string>
+#include <migraphx/instruction_ref.hpp>
+#include <migraphx/config.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+struct module;
+
+struct inline_subgraph
```
pfultz2, line 13: I think the pass should be called `inline_module`, to be consistent with terminology.

src/inline_subgraph.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <migraphx/program.hpp>
+#include <migraphx/instruction.hpp>
+#include <migraphx/make_op.hpp>
+#include <migraphx/ranges.hpp>
+#include <migraphx/iterator_for.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+void inline_subgraph::apply(module& p) const
```
pfultz2, line 11: Should be `module& m`.

src/inline_subgraph.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+        // cond is constant, inline the corresponding subgraph and discard the other one
+        else
+        {
+            inline_submodule(p, ins);
+        }
+    }
+}
+
+void inline_subgraph::inline_submodule(module& p, instruction_ref ins) const
```
pfultz2, line 89: This can be a free function defined only in the cpp file.

pfultz2: Also should be `module& m`.

src/include/migraphx/inline_subgraph.hpp
scxiao: fixed

src/inline_subgraph.cpp
scxiao: fixed

src/inline_subgraph.cpp
scxiao: fixed

src/inline_module.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+        else
+        {
+            auto mod_args = sins->module_inputs();
+            auto inputs   = sins->inputs();
+            std::vector<instruction_ref> copy_inputs(inputs.size());
+            std::transform(inputs.begin(), inputs.end(), copy_inputs.begin(), [&](auto i) {
+
+                assert(contains(map_ins, i) or m.has_instruction(i));
+                return m.has_instruction(i) ? i : map_ins[i];
```
pfultz2, line 48: I dont think this will work as the instruction could reference an instruction in the parent's module. You could do `smod.has_instruction(i) ? map_ins[i] : i` instead, but `assert` needs to be updated.

scxiao: fixed

src/inline_module.cpp
```diff
@@ -0,0 +1,158 @@
...
+                return contains(map_ins, i) ? map_ins[i] : i;
+            });
+
+            if(sins->name() == "@return")
+            {
+                mod_outputs = copy_inputs;
+                break;
+            }
+
+            if(mod_args.empty())
```
pfultz2, line 55: Is this check necessary? Wont `insert_instruction` work the same when the module args are empty?

src/inline_module.cpp
```diff
@@ -0,0 +1,158 @@
...
+                auto loffset = m.add_literal(literal(sidx, vec_offset));
+                auto moffset = m.insert_instruction(ins, make_op("mul"), mcond, loffset);
+                auto f_ind   = m.insert_instruction(ins, make_op("sub"), lidx, moffset);
+                auto ins_ind = m.insert_instruction(
+                    ins, make_op("convert", {{"target_type", shape::int32_type}}), f_ind);
+                auto l01 = m.insert_instruction(ins, make_op("concat", {{"axis", 0}}), l0, l1);
+                auto rl  = m.insert_instruction(
+                    ins, make_op("reshape", {{"dims", {l0->get_shape().elements() * 2}}}), l01);
+                auto r = m.insert_instruction(ins, make_op("gather", {{"axis", 0}}), rl, ins_ind);
+                m.replace_instruction(ins_outputs.at(i), r);
```
pfultz2, line 146: Why is all this needed? Shouldn't it just insert a `make_tuple` of the multiple arguments?

src/inline_module.cpp
```diff
@@ -0,0 +1,158 @@
...
+                else
+                {
+                    // no return instruction, last instruction is output
+                    mod_outputs.push_back(last);
+                }
+
+                // only one output is considered for now
+                for(const auto& out : mod_outputs)
+                {
+                    auto mod_out = out->eval();
```
pfultz2, line 111: Actually, I think it would be simpler to leave const propagation to the const propagate pass and then just take of inlining when the condition is known.

src/onnx/parse_if.cpp
```diff
@@ -1,3 +1,4 @@
+#include "migraphx/instruction_ref.hpp"
```
pfultz2, line 1: This should be `<migraphx/instruction_ref.hpp>`

test/inline_module_test.cpp
```diff
@@ -0,0 +1,569 @@
...
+#include <basic_ops.hpp>
+#include <migraphx/operators.hpp>
+#include <migraphx/make_op.hpp>
+
+#include <test.hpp>
+
+void run_pass(migraphx::program& p)
+{
+    auto* mm = p.get_main_module();
+    migraphx::run_passes(*mm, {migraphx::inline_module{}, migraphx::dead_code_elimination{}});
```
pfultz2, line 14: I think these two lines can be removed.

test/inline_module_test.cpp
```diff
@@ -0,0 +1,569 @@
...
+    migraphx::run_passes(p, {migraphx::dead_code_elimination{}});
```
pfultz2, line 15: You need to add `migraphx::inline_module` here.

src/inline_module.cpp
scxiao: fixed

src/onnx/parse_if.cpp
scxiao: fixed

test/inline_module_test.cpp
scxiao: fixed

test/inline_module_test.cpp
scxiao: removed

src/inline_module.cpp
scxiao: The reason is that the gather does not work for the tuple type.

src/inline_module.cpp
scxiao: Actually, I am thinking about extending the if operator as the following:
1) With module input, we have the current implementation. 
2) No module inputs but with 3 input arguments (cond, then_arg, else_arg). Then if cond is true, return "then_arg"; if cond is false, return else_arg. 
With this extension, this optimization is much simpler, and the optimization for "find_where_op" can be further optimized.

Any opinion?

scxiao: I just created another branch (extend the if, further optimize for here and the find_where_op) to implemented what I just mentioned. So for this PR, I would prefer keep this.

scxiao: just removed this optimization

src/inline_module.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void inline_module::apply(module& m) const
+{
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != "if")
+            continue;
+
+        auto arg_cond = ins->inputs().front()->eval();
```
pfultz2, line 79: So the `arg_cond` should be passed to `inline_module` so it doesn't have to be called `eval` twice.

scxiao:
@pfultz2 When you get a chance, you can take a look to see whether you have additional comments? Thanks

pfultz2 (approved):

PR #806, opened 2021-04-21
Increase the default local threads for grid strided loops

pfultz2:
This does improve performance in some cases.


src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
```diff
@@ -73,7 +73,7 @@ MIGRAPHX_DEVICE_CONSTEXPR auto gs_invoke(F&& f, index_int i, index) -> decltype(
     return f(i);
 }
 
-inline auto gs_launch(hipStream_t stream, index_int n, index_int local = 1024)
+inline auto gs_launch(hipStream_t stream, index_int n, index_int local = 4096)
```
scxiao, line 76: is the 4096 applicable to our older hardware like vega10?

mvermeulen:
Not sure why I see this, but building on 4.1 and running unit tests, I see a few failures:
	 59 - test_gpu_quantization (Child aborted)
	275 - test_api_gpu (Child aborted)
	276 - test_verify_general (Failed)
	277 - test_verify_rnn (Failed)
	289 - test_py_3.6_backend (Failed)

After running these tests, I ran tests I normally run for weekly performance, I see considerable improvements as shown below in the summary.  Also a fair variation between different models.
perf-threads

Assuming the testing failures I saw are a transient phasing issue, this otherwise looks good.

mvermeulen:
I also see correctness issues when running imagenet tests, e.g. resnet50 onnx only has 46 correct tests out of 50,000 vs 37654 expected.

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
scxiao: I just ran rocminfo on my desktop, here is what I got:

```
  Wavefront Size:          64(0x40)                           
  Workgroup Max Size:      1024(0x400)  
```
So the max number of threads in a workgroup is 1024.

scxiao:
> I also see correctness issues when running imagenet tests, e.g. resnet50 onnx only has 46 correct tests out of 50,000 vs 37654 expected. I have pushed the docker image I tested with to: mevermeulen/rocm-migraphx:20210422-threads

In my understanding, for the tests whose work item number is larger than 1024 will be wrong since HIP cannot support a workgroup containing more than 1024 threads.

mvermeulen:
This sounds like something we might have addressed in TVM by calling:

hipDeviceGetAttribute(&value, hipDeviceAttributeMaxThreadsPerBlock, device.device_id)

mvermeulen:
Closing.

PR #811, opened 2021-04-26
Int8 gemm support

scxiao:


src/targets/gpu/lowering.cpp
```diff
@@ -311,7 +312,19 @@ struct miopen_apply
...
 
-            return mod->replace_instruction(ins, rocblas_gemm<Op>{Op{op.alpha, beta}}, refs);
+            bool int8_x4_format = true;
+#if ROCBLAS_VERSION_MAJOR >= 2 && ROCBLAS_VERSION_MINOR >= 38
+            if(contains(op.name(), "quant_"))
+            {
+                auto& ctx = get_context();
+                rocblas_gemm_flags flag;
+                rocblas_query_int8_layout_flag(ctx.get_stream().get_rocblas(), &flag);
+                int8_x4_format = (flag == rocblas_gemm_flags_pack_int8x4);
```
pfultz2, line 322: This should be set in the init method.

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -1,10 +1,16 @@
+#include "migraphx/instruction_ref.hpp"
```
pfultz2, line 1: This should be `#include <migraphx/instruction_ref.hpp>`.

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -16,24 +22,42 @@ void pack_int8_args::apply(module& p) const
     {
         if(ins->name() == "gpu::quant_gemm")
         {
+            auto&& op = any_cast<rocblas_gemm<op::quant_dot>>(ins->get_operator());
```
pfultz2, line 25: Is there a reason this doesn't query the flag here?

src/targets/gpu/lowering.cpp
scxiao: moved

src/targets/gpu/pack_int8_args.cpp
scxiao: changed

src/targets/gpu/pack_int8_args.cpp
scxiao: What does it mean?

kahmed10: I think he is asking to use the rocBLAS query API here instead of lowering.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -19,6 +19,7 @@ template <class Op>
 struct rocblas_gemm
 {
     Op op;
+    bool int8_x4_format = true;
```
pfultz2, line 22: This needs to be added to `reflect` so it will be saved. Most likely it will need to be `pack_join(migraphx::reflect(self.op, f), pack(f(self.int8_x4_format, "int8_x4_format")))`

src/targets/gpu/lowering.cpp
```diff
@@ -43,6 +43,7 @@
 #include <utility>
 #include <functional>
 #include <algorithm>
+#include <rocblas.h>
```
pfultz2, line 46: You should include `migraphx/gpu/rocblas.hpp` header instead.

src/targets/gpu/pack_int8_args.cpp
pfultz2: Dont use `any_cast`, instead read the operator with `to_value`.

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -69,6 +93,76 @@ shape pack_int8_args::pack_int8_shape(const shape& s) const
...
+    auto pad_lens                  = lens;
+    pad_lens[lens.size() + offset] = pad_k;
+    std::vector<int64_t> pad_dims(lens.size() * 2, 0);
+    auto ret_ins = ins;
+    if(pad_k != k)
+    {
+        pad_dims[lens.size() + offset] = pad_k - k;
+        shape ps{s.type(), pad_lens};
+        auto ins_out = p.insert_instruction(ins, hip_allocate{ps});
+        op::pad pad{pad_dims};
```
pfultz2, line 111: Use `make_op` to construct the pad operator.

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -1,10 +1,16 @@
+#include <iterator>
 #include <migraphx/gpu/pack_int8_args.hpp>
 #include <migraphx/gpu/int8_gemm_pack.hpp>
 #include <migraphx/gpu/int8_conv_pack.hpp>
+#include <migraphx/gpu/gemm.hpp>
 #include <migraphx/gpu/hip.hpp>
+#include <migraphx/gpu/pad.hpp>
```
pfultz2, line 7: Remove the gemm and pad header files.

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -69,6 +93,76 @@ shape pack_int8_args::pack_int8_shape(const shape& s) const
     return {s.type(), lens, strides};
 }
 
+instruction_ref pack_int8_args::pad_ins(module& p, instruction_ref ins, int offset) const
```
pfultz2, line 96: `module& m`

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -69,6 +93,76 @@ shape pack_int8_args::pack_int8_shape(const shape& s) const
...
+        shape ps{s.type(), pad_lens};
+        auto ins_out = p.insert_instruction(ins, hip_allocate{ps});
+        op::pad pad{pad_dims};
+        ret_ins = p.insert_instruction(std::next(ins), hip_pad{pad}, ins, ins_out);
+    }
+
+    return ret_ins;
+}
+
+std::vector<instruction_ref> pack_int8_args::pad_inputs(module& p, instruction_ref ins) const
```
pfultz2, line 118: `module& m`

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -69,6 +93,76 @@ shape pack_int8_args::pack_int8_shape(const shape& s) const
...
+    auto sa     = in0->get_shape();
+    bool transa = sa.transposed();
+    if(transa)
+    {
+        auto perm  = find_permutation(sa);
+        auto t_in  = in0->inputs().front();
+        int offset = static_cast<int>(perm.back()) - static_cast<int>(perm.size());
+        auto p_in  = pad_ins(p, t_in, offset);
+        auto val   = in0->get_operator().to_value();
+        assert(val.contains("dims"));
```
pfultz2, line 132: This seems to assume that when the input is transposed there is a transpose operator, which is not a safe assumption.

src/targets/gpu/pack_int8_args.cpp
pfultz2, line 96: This can be a free function in the cpp file.

src/targets/gpu/pack_int8_args.cpp
pfultz2: This can be a free function in the cpp file.

pfultz2:
Can you add unit tests for `pack_int8_args`?

src/targets/gpu/pack_int8_args.cpp
scxiao: fixed

src/targets/gpu/lowering.cpp
scxiao: changed.

src/targets/gpu/pack_int8_args.cpp
scxiao: changed

src/targets/gpu/pack_int8_args.cpp
scxiao: removed

src/targets/gpu/pack_int8_args.cpp
scxiao: changed

src/targets/gpu/pack_int8_args.cpp
scxiao: changed

src/targets/gpu/pack_int8_args.cpp
scxiao: changed

test/verify/quant_dot_3args_5.cpp
```diff
@@ -0,0 +1,26 @@
...
+        migraphx::shape m2_shape{migraphx::shape::int8_type, {7, 6}};
+        migraphx::shape m3_shape{migraphx::shape::int32_type, {2, 7}};
+
+        auto l1  = mm->add_parameter("a", m1_shape);
+        auto tl1 = mm->add_instruction(migraphx::make_op("transpose", {{"dims", {1, 0}}}), l1);
+        auto l2  = mm->add_parameter("b", m2_shape);
+        auto tl2 = mm->add_instruction(migraphx::make_op("transpose", {{"dims", {1, 0}}}), l2);
+        auto l3  = mm->add_parameter("c", m3_shape);
+        mm->add_instruction(
+            migraphx::make_op("quant_dot", {{"alpha", 3}, {"beta", 2}}), tl1, tl2, l3);
```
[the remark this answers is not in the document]
scxiao: removed the third argument

scxiao:
After asking the rocblas guy by sending them a microbenchmark, looks like our call to rocblas is correct. The error for int8 gemm in mi100 is a limitation in the current rocblas implementation. Specifically, for int8 input of matrix A (mxk) and matrix B(k x n), the m and n have to be at least 4. They say this limitation will be removed in rocm4.3

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -1,54 +1,174 @@
...
+    std::copy(inputs.begin() + 2, inputs.end(), std::back_inserter(ret_inputs));
+
+    return ret_inputs;
+}
+
+void pack_int8_args::apply(module& m) const
+{
+    for(auto ins : iterator_for(m))
     {
         if(ins->name() == "gpu::quant_gemm")
```
kahmed10, line 113: do we want to collapse this if statement as well by early returning if it's not quant_gemm?

kahmed10: and then similarly, maybe we can reuse the variables by merging the else if statement following

kahmed10: also, we can probably make functions for handling the different operators

scxiao:
I tried the rocm4.3 container, looks like it still requires the matrix size to be equal to or bigger than 4, i.e. if I put either m or n less than 4, the output is incorrect.

src/targets/gpu/pack_int8_args.cpp
kahmed10: I suppose it's fine as is for now.

kahmed10 (approved):

src/targets/gpu/pack_int8_args.cpp
scxiao: fixed

scxiao:
> Can you add unit tests for `pack_int8_args`?

added some unit tests

test/verify/quant_dot_3args_3.cpp
```diff
@@ -10,9 +10,9 @@ struct quant_dot_3args_3 : verify_program<quant_dot_3args_3>
     {
         migraphx::program p;
         auto* mm = p.get_main_module();
-        migraphx::shape m1_shape{migraphx::shape::int8_type, {2, 8}};
+        migraphx::shape m1_shape{migraphx::shape::int8_type, {4, 8}};
         migraphx::shape m2_shape{migraphx::shape::int8_type, {7, 8}};
-        migraphx::shape m3_shape{migraphx::shape::int32_type, {2, 7}};
+        migraphx::shape m3_shape{migraphx::shape::int32_type, {4, 7}};
```
pfultz2, line 15: Why do we change this to `4`? We should have tests for <4 matrices.

scxiao: If less than 4, and the QA test our code on an MI100, this unit test will fail.

scxiao: The plan is: when rocm4.4 is available and hopefully less than 4 cases are supposed, we will add these tests back. 
I created an issue #835 for tack this problem.

pfultz2: Alright, well can we just disable the test for the gpu instead of changing it? You can disable tests for the gpu in test/verify/main.cpp.

scxiao: fixed

pfultz2 (approved):

PR #813, opened 2021-04-28
Use hipStreamSynchronize instead of device sync

### LINKED ISSUE #807: Why use hipDeviceSynchronize instead of hipStreamSynchronize?
If I load more than one onnx models in multi-thread program(one model in one thread), there will be several streams in the program. When I use migraphx::gpu::from_gpu to get result, hipDeviceSynchronize will have a bad effect. Why not use hipStreamSynchronize?

pfultz2:
Fixes #807.


scxiao (commented):
I am wondering whether we need to add a hip_sync_stream like the hip_sync_device then we can use hip_sync_stream in some places to replace the hip_sync_device?

pfultz2:
> I am wondering whether we need to add a hip_sync_stream like the hip_sync_device then we can use hip_sync_stream in some places to replace the hip_sync_device?

Yea, we probably can.

scxiao:
> > I am wondering whether we need to add a hip_sync_stream like the hip_sync_device then we can use hip_sync_stream in some places to replace the hip_sync_device?
> 
> Yea, we probably can.

Do you want it to be in a separate PR?

qianqing13579 (commented):
@pfultz2 ,when use hipStreamSynchronize instead of device sync,I think that the program::eval() function should use ctx.finish() to do stream_sync  if I use program::eval() to do inference. At the same time, the from_gpu() function needs to be changed when I use it to get data.Is that right?

pfultz2:
> I think that the program::eval() function should use ctx.finish() to do stream_sync if I use program::eval() to do inference.

We dont call ctx.finish() in eval() so the result can be passed on to other GPU workloads asynchronously.

> At the same time, the from_gpu() function needs to be changed when I use it to get data.Is that right?

There is no stream for `from_gpu` to use. You can use [`hipMemcpyAsync`](https://rocmdocs.amd.com/en/latest/ROCm_API_References/HIP_API/Memory-Management.html#hipmemcpyasync) to copy with a stream, or you can compile with `offload_copy = true` and migraphx will insert implicit async copies to and from the gpu.

src/targets/gpu/hip.cpp
```diff
@@ -146,6 +147,8 @@ void gpu_sync()
         MIGRAPHX_THROW("hip device synchronization failed: " + hip_error(status));
 }
 
+void gpu_sync(const context& ctx) { ctx.finish(); }
```
scxiao, line 150: Should we use the name stream_sync here?

scxiao (approved):

kahmed10 (approved):

PR #817, opened 2021-05-04
Bug split optimization

scxiao:
This issue surfaced for the resize linear mode


src/simplify_algebra.cpp
```diff
@@ -421,9 +439,18 @@ struct find_splits
                 if(it == split->outputs().end())
                     break;
                 assert((*it)->name() != "slice");
-                // If there is a duplicate bail
```
pfultz2, line 424: Why is this comment removed?

src/simplify_algebra.cpp
```diff
@@ -421,9 +439,18 @@ struct find_splits
...
+
                 if(contains(group, *it))
                     return {};
+
+                // there are should be no dependent between instructions in the group
+                if(std::any_of(group.begin(), group.end(), [&](auto i) {
+                       return is_dependent(*it, i) or is_dependent(i, *it);
```
pfultz2, line 448: Can't the `contains(group, *it)` be removed?

src/simplify_algebra.cpp
```diff
@@ -421,9 +439,18 @@ struct find_splits
                 if(it == split->outputs().end())
                     break;
                 assert((*it)->name() != "slice");
-                // If there is a duplicate bail
+
                 if(contains(group, *it))
                     return {};
+
+                // there are should be no dependent between instructions in the group
```
pfultz2, line 446: dependent -> dependency

src/simplify_algebra.cpp
```diff
@@ -403,6 +404,23 @@ struct find_splits
...
+        return fix<bool>([&](auto self, auto ins) -> bool {
+            if(ins == ins2)
+                return true;
+
+            if(contains(traversed, ins))
+                return false;
+
+            traversed.insert(ins);
+            const auto& inputs = ins->inputs();
+            return std::any_of(inputs.begin(), inputs.end(), [&](auto in) { return self(in); });
```
pfultz2, line 420: Wont this traverse into the parent module?

src/simplify_algebra.cpp
scxiao: put it back

src/simplify_algebra.cpp
scxiao: removed

src/simplify_algebra.cpp
scxiao: changed

src/simplify_algebra.cpp
scxiao: add the check whether input is in the same module

pfultz2 (approved):

PR #819, opened 2021-05-06
Resize linear mode support

scxiao:


src/onnx/parse_resize.cpp
```diff
@@ -71,6 +71,96 @@ const auto& get_original_idx_op(const std::string& mode)
     return idx_ops.at(mode);
 }
 
+using vvv = std::vector<std::vector<std::vector<std::size_t>>>;
```
kahmed10, line 74: onnx runtime uses a naming scheme like v, v1, v2. Do we want to do the same to make it easier to read?

turneram (approved):

src/onnx/parse_resize.cpp
pfultz2: This is only used in one place so can we just remove the typedef?

scxiao: removed the typedef

pfultz2 (approved):

kahmed10 (approved):

PR #822, opened 2021-05-07
upgrade protobuf to 3.16.0 to be compatible with onnxruntime

scxiao:
As we discussed in the standup meeting, there is an error happened when running models in onnxruntime. After investigation, the reason is the protobuf (version 3.11.0) used in the migraphx dependency is out of date. After upgrading to 3.16.0, it works fine, and no regression is observed for migraphx unit tests. So we can upgrade protobuf to 3.16.0 to fix the issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-google/protobuf@v3.11.0 -DCMAKE_POSITION_INDEPENDENT_CODE=On -X subdir -Dprotobuf_BUILD_TESTS=Off
+google/protobuf@v3.16.0 -DCMAKE_POSITION_INDEPENDENT_CODE=On -X subdir -Dprotobuf_BUILD_TESTS=Off
 RadeonOpenCompute/rocm-cmake@ececd2eccae4d01e7ec154efe90ac43ebf4df317 --build
 ROCmSoftwarePlatform/rocBLAS@abd98a2b48b29326ebaef471630786a548622c06
 ROCmSoftwarePlatform/MIOpen@2.4.0


mvermeulen:
I can verify that this addresses the issues for ONNX runtime.

kahmed10 (approved):

pfultz2:
What is the issue exactly? Will this newer version work with an older version of onnx?

scxiao:
> What is the issue exactly? Will this newer version work with an older version of onnx?

Yes, it works well for all the models to test onnxruntime. But there is a bug surfaced, and I have a fix for that in PR #823.

scxiao:
> As we discussed in the standup meeting, there is an error happened when running models in onnxruntime. After investigation, the reason is the protobuf (version 3.11.0) used in the migraphx dependency is out of date. After upgrading to 3.16.0, it works fine, and no regression is observed for migraphx unit tests. So we can upgrade protobuf to 3.16.0 to fix the issue.



> As we discussed in the standup meeting, there is an error happened when running models in onnxruntime. After investigation, the reason is the protobuf (version 3.11.0) used in the migraphx dependency is out of date. After upgrading to 3.16.0, it works fine, and no regression is observed for migraphx unit tests. So we can upgrade protobuf to 3.16.0 to fix the issue.



> But why do we need to use a newer version of protobuf because onnx runtime is using a newer version?

yes, ort is using a newer version 3.15.6 which is incompatiable with the version 3.11.0 that we are using, so we need to update protobuf to a newer version.

mvermeulen:
Symptoms I observed were a runtime error running models in ORT docker, e.g.
```
[libprotobuf FATAL /code/onnxruntime/cmake/external/protobuf/src/google/protobuf/stubs/common.cc:87] This program was compiled against version 3.11.0
 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.15.8).  Contact the program author for an update.  If 
you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version ver
ification failed in "/migraphx/build/src/tf/tensor_shape.pb.cc".)
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  This program was compiled against version 3.11.0 of the Protocol Buffer runtime library, which is not compatible with the installed versio
n (3.15.8).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version o
f Protocol Buffers as your link-time library.  (Version verification failed in "/migraphx/build/src/tf/tensor_shape.pb.cc".)
```

scxiao:
> ```cmake
> target_link_libraries(migraphx_tf PRIVATE tf-proto "-Wl,--exclude-libs,ALL")
> ```

Just tried, looks like it does not work, same error message reported.

```
[libprotobuf FATAL /code/onnxruntime/cmake/external/protobuf/src/google/protobuf/stubs/common.cc:87] This program was compiled against version 3.11.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.16.0).  Contact the program author for an update. 
```

scxiao:
> > ```cmake
> > target_link_libraries(migraphx_tf PRIVATE tf-proto "-Wl,--exclude-libs,ALL")
> > ```
> 
> Just tried, looks like it does not work, same error message reported.
> 
> ```
> [libprotobuf FATAL /code/onnxruntime/cmake/external/protobuf/src/google/protobuf/stubs/common.cc:87] This program was compiled against version 3.11.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.16.0).  Contact the program author for an update. 
> ```

Some more update. 
For the previous test, I set the flag only for the onnx, but did not do it for the tf (that might be why it failed for tf, not for onnx).  I just set the flags for both onnx and tf, and looks like it works this time.

scxiao:
> > > ```cmake
> > > target_link_libraries(migraphx_tf PRIVATE tf-proto "-Wl,--exclude-libs,ALL")
> > > ```
> > 
> > 
> > Just tried, looks like it does not work, same error message reported.
> > ```
> > [libprotobuf FATAL /code/onnxruntime/cmake/external/protobuf/src/google/protobuf/stubs/common.cc:87] This program was compiled against version 3.11.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.16.0).  Contact the program author for an update. 
> > ```
> 
> Some more update.
> For the previous test, I set the flag only for the onnx, but did not do it for the tf (that might be why it failed for tf, not for onnx). I just set the flags for both onnx and tf, and looks like it works this time. Sorry for forgetting the tf part.

So, we change the flag as the fix for this issue?

mvermeulen:
Seems reasonable. Using the flag in both cases helps with a hypothesis that at link-time calls from ORT were being resolved to MIGraphX version of protobuf and hence clashing with headers those files were compiled with.

In general, this flag also keeps similar things from happening, having external software linking to MIGraphX included archive libraries and either resolving incorrectly (runtime) or causing symbol clashes (compiletime) - so think is is a good thing to have in any case.

PR #823, opened 2021-05-10
Bug as shape

scxiao:
in the optimization find_add_conv, change the input shape to be standard if not. This bug surfaced when running the pnasnet_large model from msft.

### DIFF
--- a/test/simplify_algebra_test.cpp
+++ b/test/simplify_algebra_test.cpp
@@ -426,11 +426,11 @@ TEST_CASE(simplify_add_conv_1x1_diff_strides1)
 TEST_CASE(simplify_add_conv_1x1_diff_strides1_transpose)
 {
     migraphx::module m;
-    auto x = m.add_parameter("x", {migraphx::shape::float_type, {1, 14, 128, 14}});
+    auto x  = m.add_parameter("x", {migraphx::shape::float_type, {1, 14, 128, 14}});
     auto tx = m.add_instruction(migraphx::make_op("transpose", {{"dims", {0, 2, 1, 3}}}), x);
     auto w =
         m.add_literal(migraphx::generate_literal({migraphx::shape::float_type, {256, 128, 1, 1}}));
-    auto y = m.add_parameter("y", {migraphx::shape::float_type, {1, 128, 28, 28}});
+    auto y  = m.add_parameter("y", {migraphx::shape::float_type, {1, 128, 28, 28}});
     auto ty = m.add_instruction(migraphx::make_op("transpose", {{"dims", {0, 1, 3, 2}}}), y);
     auto v =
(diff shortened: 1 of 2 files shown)


pfultz2:
Why does `compute_stride_shape` not work with transposed input? As we should fix it so it the shape is correct with transposed inputs otherwise this optimization is worthless with non NCHW layouts.

test/simplify_algebra_test.cpp
```diff
@@ -423,6 +423,29 @@ TEST_CASE(simplify_add_conv_1x1_diff_strides1)
...
+    migraphx::module m;
+    auto x  = m.add_parameter("x", {migraphx::shape::float_type, {1, 14, 128, 14}});
+    auto tx = m.add_instruction(migraphx::make_op("transpose", {{"dims", {0, 2, 1, 3}}}), x);
+    auto w =
+        m.add_literal(migraphx::generate_literal({migraphx::shape::float_type, {256, 128, 1, 1}}));
+    auto y  = m.add_parameter("y", {migraphx::shape::float_type, {1, 128, 28, 28}});
+    auto ty = m.add_instruction(migraphx::make_op("transpose", {{"dims", {0, 1, 3, 2}}}), y);
+    auto v =
+        m.add_literal(migraphx::generate_literal({migraphx::shape::float_type, {256, 128, 1, 1}}));
+    auto conv1 = m.add_instruction(migraphx::make_op("convolution"), tx, w);
```
pfultz2, line 437: I dont think this convolution is needed for the test.

scxiao:
> Why does `compute_stride_shape` not work with transposed input? As we should fix it so the shape is correct with transposed inputs otherwise this optimization is worthless with non NCHW layouts.

The compute_stride_shape works with the transposed input, but the as_shape does not. The as_shape needs a standard shape input. So if the input shape is transposed, as_shape throws an exception.

pfultz2:
So Cagri doesn't need a step operator. He just needs a reverse, so if you want you could implement a `forward_step` in this PR, but don't worry about the onnx frontend for slice yet.

scxiao:
> So Cagri doesn't need a step operator. He just needs a reverse, so if you want you could implement a `forward_step` in this PR, but don't worry about the onnx frontend for slice yet.

Yes, I can implement it. Just to clarify:
1) This operator is to compute the strided output for an input, and is able to process transpose input shape. Should we also process other case like broadcast? 
2) operator name: should we use "step" or "forward_step"? 
3) The "axes" and "step" indicate the axes and the corresponding stride of the axis we want to change, right?
4) For the algorithm, we will try to avoid the data copy, right?

pfultz2:
> This operator is to compute the strided output for an input, and is able to process transpose input shape. Should we also process other case like broadcast?

Yes, and we should handle non-packed tensors as well. The lens just need to be divisible by the step.

scxiao:
We will add a step operator to resolve the problem here, so close this PR

PR #825, opened 2021-05-12
Update rbuild to handle dependencies better and remove unused dependencies

pfultz2:


requirements.txt
```diff
@@ -1,7 +1,4 @@
 google/protobuf@v3.11.0 -DCMAKE_POSITION_INDEPENDENT_CODE=On -X subdir -Dprotobuf_BUILD_TESTS=Off
```
scxiao, line 1: Should we upgrade the protobuf to a newer version?

scxiao (approved):

PR #827, opened 2021-05-13
Model optimizer

turneram:
Command line tool using Python API to allow for easy compilation and serialization for end-to-end examples/applications.


scxiao (commented):
Two general comments:
1) We can add a link to the example/readme.md file to point this optimizer
2) Can we mention something about how to use the serialized model?

examples/python_model_optimizer/optimize_model.py
```diff
@@ -0,0 +1,102 @@
...
+    if not os.path.dirname(output_file):
+        output_file = "./saved_models/" + output_file
+        os.makedirs("./saved_models", exist_ok=True)
+    if not os.path.basename(output_file):
+        output_file += os.path.basename(model_name)
+else:
+    output_file = "./saved_models/" + os.path.basename(model_name)
+    os.makedirs("./saved_models", exist_ok=True)
+
+if ".onnx" in output_file:
```
kahmed10, line 86: There might be an easier way to remove the extension from the output filename. Maybe like this?
```
output_file = os.path.splitext(ouput_file)[0]
```

examples/python_model_optimizer/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+positional arguments:
+  model                 Path to ONNX or TF Protobuf file
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --nchw                Treat Tensorflow format as nchw. (Default is nhwc)
+  -t {gpu,cpu,ref}, --target {gpu,cpu,ref}
+                        Compilation target
+  -d, --disable_fast_math
```
pfultz2, line 25: The flags should use dashes instead of underscores(ie `--disable-fast-math`)

examples/python_model_optimizer/README.md
```diff
@@ -0,0 +1,55 @@
...
+                        Disable optimized math functions
+  -o, --offload_copy_off
```
pfultz2, line 27: Probably, better to do `--disable-offload-copy` as its consistent with other flags.

PR #839, opened 2021-05-22
Step op

scxiao:
To resolve the problem as shape requires a standard input shape...


kahmed10 (approved):

kahmed10:
Do we want to add parsing for steps in Slice, even if it's currently only positive values?

src/include/migraphx/op/step.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+        if(std::any_of(axes.begin(), axes.end(), [&](auto axis) { return axis >= in_lens.size(); }))
+        {
+            MIGRAPHX_THROW("STEP: axis value is out of range!");
+        }
+
+        auto lens    = in_lens;
+        auto strides = input.strides();
+        auto it_step = steps.begin();
+        for(const auto& axis : axes)
```
pfultz2, line 51: I think it would be better to use the indices since everything is random access:

``cpp
for(auto i : range(axes.size()))
{
    auto axis = axes[i];
    auto step = steps[i];
    lens[axis] = (in_lens[axis] + step - 1) / step;
    strides[axis] *= step;
}
```

File: src/simplify_algebra.cpp [hunk 2]
```diff
@@ -682,29 +662,21 @@ struct find_add_convs
             {
                 if(a_op.stride < b_op.stride)
                 {
-                    auto n = compute_stride_factor(b_op, a_op);
+                    int64_t n = static_cast<int64_t>(compute_stride_factor(b_op, a_op));
```
pfultz2, line 665: Why is `int64_t` needed?

scxiao: I had thought n has to be int64_t  type to initialize the value. Looks like it does not have to be. changed it.

src/include/migraphx/op/step.hpp
scxiao: changed

test/verify/test_step_brcst_trans.cpp
```diff
@@ -0,0 +1,24 @@
+
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+
+struct test_step_brcst_trans : verify_program<test_step_brcst_trans>
```
pfultz2, line 7: Name this `test_step_broadcast_transpose`.

test/verify/test_step_trans.cpp
```diff
@@ -0,0 +1,22 @@
+
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+
+struct test_step_trans : verify_program<test_step_trans>
```
pfultz2, line 7: This should probably be `test_step_transpose`

test/verify/test_step_brcst_trans.cpp
scxiao: fixed

test/verify/test_step_trans.cpp
scxiao: fixed

pfultz2 (approved):

PR #843, opened 2021-05-28
Quantize linear ops

turneram:
Adds MIGraphX operators quantizelinear and dequantizelinear, refactors parsers for de/quantizelinear, and adds passes to rewrite de/quantizelinear ops as combination of equivalent operators.


src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -0,0 +1,42 @@
...
+    int axis = 1;
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.axis, "axis"));
+    }
+
+    std::string name() const { return "dequantizelinear"; }
+    shape compute_shape(std::vector<shape> inputs) const
```
scxiao, line 32: this method should use the name normalize_compute_shape since the axis needs normalize if it is negative.

src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return {inputs[0].type(), inputs[0].lens()};
```
scxiao, line 34: The output data type is float according to the onnx specification.
Also, I would suggest keep the input shape, so output shape is {shape::float_type, inputs[0].lens(), inputs[0].strides()}

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,42 @@
...
+    int axis = 1;
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.axis, "axis"));
+    }
+
+    std::string name() const { return "quantizelinear"; }
+    shape compute_shape(std::vector<shape> inputs) const
```
scxiao, line 32: same as above, the method name should be normalize_compute_shape()

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return {inputs[0].type(), inputs[0].lens()};
```
scxiao, line 34: The output shape is as follows:
1) If it has three inputs, then output type is the same as the third input, i.e., zero_offset data type.
2) if it has two inputs, the output shape is int8, also, we should keep the lens and stride so it could be more optimized.

src/include/migraphx/rewrite_dequantizelinear.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+struct module;
+
+/**
+ * Rewrite dequantizelinear to equivalent operators
+ */
+struct rewrite_dequantizelinear
+{
+    std::string name() const { return "rewrite_dequantizelinear"; }
+    void apply(module& p) const;
```
scxiao, line 18: we had better use module& m, the name "p" here is we use a program here before, but we did not change it to "m" when we change the program to module.

src/onnx/parse_dequantizelinear.cpp
```diff
@@ -21,40 +20,19 @@ struct parse_dequantizelinear : op_parser<parse_dequantizelinear>
...
             axis = info.attributes.at("axis").i();
 
-        auto input_lens = args[0]->get_shape().lens();
-        int n_dim       = static_cast<int>(input_lens.size());
-
-        auto sub_zero_point = args[0];
+        auto x       = args[0];
+        auto x_scale = args[1];
 
         if(args.size() == 3)
```
scxiao, line 26: We do not need this check here. The operator can have two or three inputs. So we can just do:

```
return info.add_instruction(make_op("dequantizelinear", {{"axis", axis}}), args)
```

scxiao: The same for the quantizelinear operator

scxiao (approved):

src/include/migraphx/op/dequantizelinear.hpp
turneram: Fixed

src/include/migraphx/op/dequantizelinear.hpp
turneram: Fixed

src/include/migraphx/op/quantizelinear.hpp
turneram: Fixed

src/include/migraphx/op/quantizelinear.hpp
turneram: Fixed

src/include/migraphx/rewrite_dequantizelinear.hpp
turneram: Fixed

src/onnx/parse_dequantizelinear.cpp
turneram: I had it this way so that rewrite_(de)quantizelinear.cpp could assume that there's always a third argument. Better to keep it the way it is, or make your suggested change and add the check to the rewrite...cpp files?

pfultz2:
We should not be rewriting the quantize operators back to multiplies. There should be a pass that first replaces DQ -> Op -> Q with the quantized op version for dot and convolution. Then later remove all the Q->DQ and DQ->Q pairs. After that we should not have any quantized linear ops to rewrite.

pfultz2:
We still want the Q/DQ pairs so we can set the scales when calling the libraries. It harder to deduce the scales from the multiply/convert ops.

src/rewrite_dequantizelinear.cpp
```diff
@@ -0,0 +1,73 @@
+#include <migraphx/rewrite_dequantizelinear.hpp>
+#include <migraphx/instruction.hpp>
+#include <migraphx/iterator_for.hpp>
+#include <migraphx/op/dequantizelinear.hpp>
+#include <migraphx/op/reshape.hpp>
```
scxiao, line 5: we can remove all the include "#include <migraphx/op/****.hpp>" and use the make_op function.

src/rewrite_dequantizelinear.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void rewrite_dequantizelinear::apply(module& m) const
+{
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != "dequantizelinear")
+            continue;
+
+        auto&& op         = any_cast<op::dequantizelinear>(ins->get_operator());
+        auto axis         = op.axis;
```
scxiao, line 22: The axis can be computed as:

```
    auto val = ins->get_operator().to_value();
    assert(val.contains("axis"));
    int axis = val.at("axis").to<int>();
```

and the type can be obtained as:

```
    auto type    = ins->get_shape().type();
```

scxiao: same for the quantizelinear rewrite.

scxiao (commented):
By combine qdq rewrite into one pass, the apply function can be rewritten as:

```
void rewrite_qdq::apply(module& m) const
{
    {
        if(ins->name() == "quantizelinear")
        {
            apply_quantizelinear(m, ins);
        }
        else if(ins->name() == "dequantizelinear")
        {
            apply_dequantizelinear(m, ins);
        }
    }
}
```

src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -0,0 +1,69 @@
...
+        if(args.size() == 3)
+            x_zero_point = args[2];
+
+        auto x       = args[0];
+        auto x_scale = args[1];
+        argument result{output_shape};
+        visit_all(x, x_zero_point)([&](auto input, auto zero_pts) {
+            visit_all(result, x_scale)([&](auto output, auto scales) {
+                auto num_scales = scales.size();
+                auto num_zeros  = zero_pts.size();
```
pfultz2, line 50: The scales and zero_pts should be broadcasted to the same tensor size. This will allow scaling across different axis in the future.

src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -0,0 +1,69 @@
...
+    int axis = 1;
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.axis, "axis"));
+    }
+
+    std::string name() const { return "dequantizelinear"; }
+    shape normalize_compute_shape(std::vector<shape> inputs) const
```
pfultz2, line 33: This should be `compute_shape` as there is no normalization done.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,87 @@
...
+            y_zero_point = args[2];
+        }
+
+        int max_quant = 255;
+        int min_quant = 0;
+        if(quant_type == shape::int8_type)
+        {
+            max_quant = 127;
+            min_quant = -128;
+        }
```
pfultz2, line 58: The min and max should come from `std::numeric_limits`, which means we need to set these inside the visitor:

```cpp
visit_all(result, y_zero_point)([&](auto output, auto zero_pts) {
    using quant_type = typename decltype(output)::value_type;
    auto min_value = std::numeric_limits<quant_type>::min();
    auto max_value = std::numeric_limits<quant_type>::max();
});
```

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,87 @@
...
+        auto x       = args[0];
+        auto y_scale = args[1];
+        argument result{output_shape};
+        visit_all(x, y_scale)([&](auto input, auto scales) {
+            visit_all(result, y_zero_point)([&](auto output, auto zero_pts) {
+                auto num_scales = scales.size();
+                auto num_zeros  = zero_pts.size();
+                par_for(output_shape.elements(), [&](auto i) {
+                    auto idx = output_shape.multi(i);
+                    float data =
```
pfultz2, line 69: Use `double` and `int64_t` for calculation in the reference implementation.

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <migraphx/instruction.hpp>
+#include <migraphx/iterator_for.hpp>
+#include <migraphx/make_op.hpp>
+#include <migraphx/tune_axis.hpp>
+#include <migraphx/program.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+void apply_quantizelinear(module& m, instruction_ref& ins)
```
pfultz2, line 11: Dont pass the `instruction_ref` by reference.

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,164 @@
...
+    {
+        max_quant = 127;
+        min_quant = -128;
+    }
+
+    auto max_arg = m.add_literal(max_quant);
+    auto min_arg = m.add_literal(min_quant);
+
+    auto input_lens = x->get_shape().lens();
+    int n_dim       = static_cast<int>(input_lens.size());
```
pfultz2, line 39: Use `auto` instead of `static_cast`.

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,164 @@
...
+        y_zero_point = m.add_literal(0);
+
+    int max_quant = 255;
+    int min_quant = 0;
+
+    if(quant_type == shape::int8_type)
+    {
+        max_quant = 127;
+        min_quant = -128;
+    }
```
pfultz2, line 33: You can use the visit method in the shape to get the min and max:

```cpp
int64_t qmin = 0;
int64_t qmax = 0;

ins->get_shape().visit([&](auto as) {
    qmin = as.min();
    qmax = as.max();
});
```

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    instruction_ref y_zero_point;
+    auto quant_type = ins->get_shape().type();
+    if(ins->inputs().size() == 3)
+        y_zero_point = ins->inputs()[2];
+    else
+        y_zero_point = m.add_literal(0);
+
+    int max_quant = 255;
+    int min_quant = 0;
```
pfultz2, line 27: Use `int64_t` to avoid any overflows.

test/onnx/onnx_test.cpp
```diff
@@ -16,14 +17,15 @@
...
 {
     migraphx::onnx_options options;
     options.skip_unknown_operators = true;
     auto prog                      = migraphx::parse_onnx(name, options);
     auto* mm                       = prog.get_main_module();
-    if(eliminate_deadcode)
-        migraphx::run_passes(*mm, {migraphx::dead_code_elimination{}});
+    if(run_passes)
+        migraphx::run_passes(*mm,
+                             {migraphx::dead_code_elimination{}, migraphx::rewrite_quantization{}});
```
pfultz2, line 28: `rewrite_quantization` should go before dead_code_elimination,

test/dequantizelinear_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    migraphx::program p1 = create_program();
+    p1.compile(migraphx::ref::target{});
+    auto result = p1.eval({}).back();
+    std::vector<float> results_vector(9);
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> gold{-256, -200, -100, -2, 0, 2, 100, 200, 254};
+    EXPECT(results_vector == gold);
+}
```
pfultz2, line 70: Arent these tests already in ref_ops_test?

pfultz2:
There should be a verify test for these ops as well.

src/include/migraphx/op/dequantizelinear.hpp
turneram: Using `scales[idx[axis] % num_scales]` produces the correct results for per-tensor and per-axis scaling.

test/dequantizelinear_test.cpp
turneram: The tests are the same, but (de)quantizelinear_test checks the correctness of the rewrite method and ref_ops_test checks the correctness of the compute method.

src/include/migraphx/op/dequantizelinear.hpp
pfultz2: But you dont normalize the axis though.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,90 @@
...
+        auto x       = args[0];
+        auto y_scale = args[1];
+
+        auto output_lens = output_shape.lens();
+        auto tuned_axis  = tune_axis(output_lens.size(), axis, this->name());
+        std::vector<size_t> bcast_strides(output_lens.size(), 0);
+
+        if(y_scale.get_shape().elements() != 1)
+            bcast_strides[tuned_axis] = 1;
+        migraphx::shape bcast_scales{y_scale.get_shape().type(), output_lens, bcast_strides};
```
pfultz2, line 58: The broadcast should happen in the onnx frontend by inserting a broadcast or multibroadcast op.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,90 @@
...
+        y_scale                   = y_scale.reshape(bcast_scales);
+        bcast_strides[tuned_axis] = 0;
+
+        if(y_zero_point.get_shape().elements() != 1)
+            bcast_strides[tuned_axis] = 1;
+        migraphx::shape bcast_zeros{y_zero_point.get_shape().type(), output_lens, bcast_strides};
+        y_zero_point = y_zero_point.reshape(bcast_zeros);
+
+        argument result{output_shape};
+        visit_all(x, y_scale)([&](auto input, auto scales) {
```
scxiao, line 68: x and x_zero_point can be different types, we cannot call visit_all here. One implementation can be like:
(broadcast can be done in parsing onnx)

```
    {
        std::vector<int8_t> zeros(output_shape.elements(), 0);
        argument zero_point{output_shape, zeros.data()};
        if (args.size() == 3)
        {
            zero_point = args.at(2);
        }

        visit_all(zero_point, result)([&](auto zp, auto output) {
            using quant_type = typename decltype(output)::value_type;
            auto min_value   = std::numeric_limits<quant_type>::min();
            auto max_value   = std::numeric_limits<quant_type>::max();
            args.at(0).visit([&](auto x) {
                args.at(1).visit([&](auto scale) {
                    par_for(output_shape.elements(), [&](auto i) {
                        int64_t quantized = static_cast<int>(std::round(x[i] / scale[i])) +
                                            static_cast<int>(zp[i]);
                        output[i] = std::max(static_cast<int64_t>(min_value),
                                            std::min(static_cast<int64_t>(max_value), quantized));
                    });
                });
            });
        });

        return result;
    }
```

src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -0,0 +1,82 @@
...
+        return pack(f(self.axis, "axis"));
+    }
+
+    std::string name() const { return "dequantizelinear"; }
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        return {shape::float_type, inputs[0].lens(), inputs[0].strides()};
+    }
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
```
scxiao, line 39: The compute function can be something like:

```
    {
        std::vector<int8_t> zeros(output_shape.bytes(), 0);
        argument zero_point{output_shape, zeros.data()};
        if (args.size() == 3)
        {
            zero_point = args.at(2);
        }

        visit_all(result, args.at(1))([&](auto output, auto scale) {
            visit_all(args.at(0), zero_point)([&](auto x, auto zp) {
                par_for(output_shape.elements(), [&](auto i) {
                    output[i] = static_cast<float>((x[i] - zp[i])) * scale[i];
                });                
            });
        });

        return result;
    }
```
if all inputs are broadcasted.

src/include/migraphx/op/quantizelinear.hpp
scxiao: BTW, if all broadcast in parsing onnx, we do not need the "attribute" in this operator, and the name "compute_shape" do not need to change to "normalize_compute_shape" any more. Same for dequantizelinear.

src/include/migraphx/op/dequantizelinear.hpp
pfultz2: That is not correct, `output` and `zero_point` may not have the same type.  The `zero_point` should match the type of the input. Also, the calculation should use `int64_t` and `double`.

src/include/migraphx/op/quantizelinear.hpp
pfultz2: Here `zero_point` and `output` have the same type.

scxiao: Yes, that is what I wrote in the comments

src/include/migraphx/op/dequantizelinear.hpp
scxiao: The implementation does NOT require output and zero_point have the same type. It requires:
x and zero_point have the same type. 
according to the onnx specification, output and scale have the same type. That is why I wrote like this.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -52,16 +52,18 @@ struct quantizelinear
...
-                                         std::min(static_cast<int64_t>(max_value), quantized));
+        visit_all(result, y_zero_point)([&](auto output, auto zero_pts) {
+            x.visit([&](auto input) {
+                y_scale.visit([&](auto scales) {
+                    using quant_type = typename decltype(output)::value_type;
+                    auto min_value   = std::numeric_limits<quant_type>::min();
+                    auto max_value   = std::numeric_limits<quant_type>::max();
+                    par_for(output_shape.elements(), [&](auto i) {
+                        int64_t quantized = static_cast<int>(std::round(input[i] / scales[i])) +
+                                            static_cast<int>(zero_pts[i]);
```
pfultz2, line 63: Use `int64_t`.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -52,16 +52,18 @@ struct quantizelinear
...
+                        output[i] = std::max(static_cast<int64_t>(min_value),
```
pfultz2, line 64: Declare the variables as `int64_t` instead of casting.

src/onnx/parse_dequantizelinear.cpp
```diff
@@ -15,46 +15,49 @@ struct parse_dequantizelinear : op_parser<parse_dequantizelinear>
...
-
-        auto scale = args[1];
-        if(not(scale->get_shape().elements() == 1))
-        {
-            axis  = tune_axis(n_dim, axis, opd.op_name);
-            scale = info.add_instruction(
-                make_op("broadcast", {{"axis", axis}, {"dims", input_lens}}), scale);
-        }
-        return info.add_broadcastable_binary_op("mul", dequant_input, scale);
+        return info.add_instruction(make_op("dequantizelinear"), args[0], x_scale);
```
pfultz2, line 60: There doesn't seem to be a test for this line.

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,99 @@
...
+        auto sub_zp_int32 =
+            m.insert_instruction(ins, make_op("convert", {{"target_type", shape::int32_type}}), x);
+        auto sub      = m.insert_instruction(ins, make_op("sub"), sub_zp_int32, zero_point_int32);
+        dequant_input = m.insert_instruction(
+            ins, make_op("convert", {{"target_type", shape::float_type}}), sub);
+    }
+    else
+    {
+        dequant_input =
+            m.insert_instruction(ins, make_op("convert", {{"target_type", shape::float_type}}), x);
```
pfultz2, line 77: Probably should add a unit test for this line.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <cmath>
+#include <utility>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct quantizelinear
+{
+    int axis = 1;
```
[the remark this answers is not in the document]
scxiao: The axis attribute is not used, so can be removed here.

src/onnx/parse_dequantizelinear.cpp
```diff
@@ -15,46 +15,49 @@ struct parse_dequantizelinear : op_parser<parse_dequantizelinear>
...
         {
-            auto zero_point = args[2];
-            if(not(zero_point->get_shape().elements() == 1))
+            auto x_zero_point = args[2];
+            if(x_zero_point->get_shape().elements() != 1)
             {
-                axis       = tune_axis(n_dim, axis, opd.op_name);
-                zero_point = info.add_instruction(
-                    make_op("broadcast", {{"axis", axis}, {"dims", input_lens}}), zero_point);
+                auto tuned_axis = tune_axis(n_dim, axis, opd.op_name);
```
scxiao, line 45: This line has a duplicate later, maybe we can move out.

src/onnx/parse_quantizelinear.cpp
```diff
@@ -12,83 +12,51 @@ struct parse_quantizelinear : op_parser<parse_quantizelinear>
...
         {
-            auto zero_point = args[2];
-            if(not(zero_point->get_shape().elements() == 1))
+            auto y_zero_point = args[2];
+            if(y_zero_point->get_shape().elements() != 1)
             {
-                axis       = tune_axis(n_dim, axis, opd.op_name);
-                zero_point = info.add_instruction(
-                    make_op("broadcast", {{"axis", axis}, {"dims", input_lens}}), zero_point);
+                auto tuned_axis = tune_axis(n_dim, axis, opd.op_name);
```
scxiao, line 45: same as above, this line can be moved out, so we can avoid a duplicate line later.

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,95 @@
...
+    int64_t max_quant = 0;
+    int64_t min_quant = 0;
+    ins->get_shape().visit_type([&](auto as) {
+        max_quant = as.max();
+        min_quant = as.min();
+    });
+    auto max_arg = m.add_literal(static_cast<int>(max_quant));
+    auto min_arg = m.add_literal(static_cast<int>(min_quant));
+
+    auto div            = m.insert_instruction(ins, make_op("div"), x, y_scale);
```
scxiao, line 27: There is no guarantee that x and y_scale are of the same type, we need a check and type conversion here. We also need to add a unit test for that.

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,95 @@
...
+        min_arg = m.insert_instruction(
+            ins, make_op("multibroadcast", {{"output_lens", output_lens}}), min_arg);
+    }
+    if(max_arg->get_shape() != s)
+    {
+        max_arg = m.insert_instruction(
+            ins, make_op("multibroadcast", {{"output_lens", output_lens}}), max_arg);
+    }
+
+    auto saturate = m.insert_instruction(ins, make_op("clip"), add_zero_point, min_arg, max_arg);
```
scxiao, line 53: This one may also have a different type. add_zero point has the type int32, but min/max_arg have the type either int8 or uint8

src/rewrite_quantization.cpp
```diff
@@ -0,0 +1,95 @@
...
+    m.replace_instruction(ins, make_op("convert", {{"target_type", quant_type}}), saturate);
+}
+
+void apply_dequantizelinear(module& m, instruction_ref ins)
+{
```
scxiao, line 58: I would suggest converting everything just to the output type (float_type), which is going to be simpler. For example, we can do it like:

```
void apply_dequantizelinear(module& m, instruction_ref ins) const
{
    assert(ins->name() == "dequantizelinear");
    auto type    = ins->get_shape().type();

    auto sub_zero_point = inputs[0];
    if(sub_zero_point->get_shape().type() != type)
    {
        sub_zero_point =
            m.insert_instruction(ins, make_op("convert", {{"target_type", type}}), sub_zero_point);
    }
    if(inputs.size() == 3)
    {
        auto zero_point = inputs[2];
        if(zero_point->get_shape().type() != type)
        {
            zero_point =
                m.insert_instruction(ins, make_op("convert", {{"target_type", type}}), zero_point);
        }
        sub_zero_point = m.insert_instruction(ins, make_op("sub"), sub_zero_point, zero_point);
    }

    auto scale = inputs[1];
    m.replace_instruction(ins, make_op("mul"), sub_zero_point, scale);
}
```

src/onnx/parse_quantizelinear.cpp
turneram: I just noticed a potential issue with merging them into a single call before the if-statements. In the case where x is 1-D and no axis is specified, the axis will get the default value of 1, which is out-of-bounds for the 1-D tensor. If we leave them as they are within the if-statements, then they get skipped in that case because y_scale and y_zero_point will have to also be 1-D, so it will execute the multibroadcast branches instead of the broadcast branches.

scxiao: Oh, yes, you are right, then no need to change this part.

pfultz2:
So tf_test.cpp  includes <migraphx/operators.hpp> which will bring in a lot of extra symbols, which could be the reason for this(as you added operators to this header). We can avoid including this header since we use `make_op` in most places. We will just need to include the operator headers directly for the tests that are not using `make_op`.

src/rewrite_quantization.cpp
scxiao: The implementation can be something like:

```
    auto inputs  = ins->inputs();
    auto scale = inputs[1];
    auto scale_type = scale->get_shape().type();
    if(inputs[0]->get_shape().type() != scale_type)
    {
        inputs[0] =
            m.insert_instruction(ins, make_op("convert", {{"target_type", scale_type}}), inputs[0]);
    }
    auto div            = m.insert_instruction(ins, make_op("div"), inputs[0], scale);
    auto add_zero_point      = m.insert_instruction(ins, make_op("round"), div);

    if(inputs.size() == 3)
    {
        auto zero_point = inputs[2];
        if(scale_type != zero_point->get_shape().type())
        {
            zero_point = m.insert_instruction(
                ins, make_op("convert", {{"target_type", scale_type}}), zero_point);
        }
        add_zero_point = m.insert_instruction(ins, make_op("add"), add_zero_point, zero_point);
    }

    auto s           = add_zero_point->get_shape();
    const auto& lens = s.lens();
    std::vector<int64_t> out_lens(lens.begin(), lens.end());
    std::vector<int> min_data(s.elements(), min_val);
    std::vector<int> max_data(s.elements(), max_val);
    auto min_arg = m.add_literal(literal(s, min_data));
    auto max_arg = m.add_literal(literal(s, max_data));

    auto saturated = m.insert_instruction(ins, make_op("clip"), add_zero_point, min_arg, max_arg);
    m.replace_instruction(ins, make_op("convert", {{"target_type", type}}), saturated);
```

test/verify/test_quantizelinear.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <migraphx/make_op.hpp>
+
+struct test_quantizelinear : verify_program<test_quantizelinear>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+
+        migraphx::shape sx{migraphx::shape::float_type, {2, 2, 2}};
```
scxiao, line 14: Maybe we can add a test with the sx of type int32 or double, in this way, we can test the apply_quantizelinear() with x and y_scale of different data type.

scxiao (approved):

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -0,0 +1,71 @@
...
+            y_zero_point = args.at(2);
+        }
+
+        argument result{output_shape};
+        visit_all(result, y_zero_point)([&](auto output, auto zero_pts) {
+            x.visit([&](auto input) {
+                y_scale.visit([&](auto scales) {
+                    using quant_type = typename decltype(output)::value_type;
+                    auto min_value   = std::numeric_limits<quant_type>::min();
+                    auto max_value   = std::numeric_limits<quant_type>::max();
```
pfultz2, line 52: Declare the variables as `int64_t` instead of casting.

turneram: Doing that works fine locally, but when I pushed it that way to github, one of the CI tests would fail with `Warning: overflow in implicit constant conversion`.

kahmed10 (approved):

scxiao:
@turneram There is a merge conflict, can you fix it?

scxiao:
@pfultz2 When you get a chance, can you take a look at this PR?

pfultz2 (approved):

PR #846, opened 2021-06-02
Reverse Op

cagery:
Adds reverse op


src/targets/gpu/device/reverse.cpp
```diff
@@ -0,0 +1,37 @@
...
+argument reverse(hipStream_t stream, argument result, argument arg1, int64_t axis)
+{
+    auto s                = arg1.get_shape();
+    auto axis_dim_size    = s.lens()[axis];
+    std::size_t nelements = s.elements();
+    visit_all(result, arg1)([&](auto output1, auto input1) {
+        hip_visit_views(output1, input1, s)([&](auto output, auto input, auto hs) {
+            // const auto* in_ptr = device_cast(input.data());
+            // auto* out_ptr = device_cast(output.data());
+            gs_launch(stream, nelements, 256)([=](auto i) __device__ {
```
pfultz2, line 22: Is there a reason you need to set the local size to 256?

src/targets/gpu/device/reverse.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+argument reverse(hipStream_t stream, argument result, argument arg1, int64_t axis)
+{
+    auto s                = arg1.get_shape();
+    auto axis_dim_size    = s.lens()[axis];
+    std::size_t nelements = s.elements();
+    visit_all(result, arg1)([&](auto output1, auto input1) {
+        hip_visit_views(output1, input1, s)([&](auto output, auto input, auto hs) {
+            // const auto* in_ptr = device_cast(input.data());
+            // auto* out_ptr = device_cast(output.data());
```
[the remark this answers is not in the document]
cagery: Yes, removed.

test/ref_ops_test.cpp
```diff
@@ -3560,6 +3560,46 @@ TEST_CASE(reshape_test)
...
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    auto l   = mm->add_literal(migraphx::literal{in_shape, data});
+    int axis = 1;
+    mm->add_instruction(migraphx::make_op("reverse", {{"axis", axis}}), l);
+    p.compile(migraphx::ref::target{});
+    auto result = p.eval({}).back();
+    std::vector<float> results_vector;
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> target_data = {16, 15, 14, 13, 12, 11, 10, 9,  8,  7,  6,
```
kahmed10, line 3577: It might be easier to make a copy of the original data, then call std::reverse on it. This way if the test changes in the future, it will be easier to manage.

cagery: What kind of test changes do you refer to?

kahmed10: right now, you have the target_data values hardcoded. I'm suggesting to generate these numbers and call std::reverse on them.

cagery: What is the benefit? Hardcoded is less computation vs generating copy and reversing it is additional computation. I am not seeing clearly the benefit of it.

kahmed10: It's preference, but if the test changes or becomes more complex, you would have to redo the values. The other option is to just reverse an existing vector in the expected behavior. I'm okay either way for this PR.

pfultz2: Using `std::reverse` would help ensure the data is correct, whereas hardcoded data may not always be obvious.

src/targets/gpu/device/reverse.cpp
pfultz2: Yea, defaulting to 1024 should better utilize the GPU. If you remove the parameter it will default to 1024.

test/ref_ops_test.cpp
cagery: I understand but then with the same idea, we should not have had any hardcoded gold vectors in these tests, e.g. round_test or pow_test, there are many of them, for which we have standard functions to use.

src/targets/gpu/device/reverse.cpp
cagery: Okay makes sense, will replace it accordingly. Related, why does gather op use 256, any specific reason ?https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/0a9c18a582721085b211abb842c5b77c0fcce12b/src/targets/gpu/device/gather.cpp#L27) ?

test/ref_ops_test.cpp
cagery: Replaced gold vectors with `std::reverse` and `std::swap_ranges` depending on axis. Retested, looks ok.

src/targets/gpu/device/reverse.cpp
cagery: Changed it to 1024

test/ref_ops_test.cpp
kahmed10: > I understand but then with the same idea, we should not have had any hardcoded gold vectors in these tests, e.g. round_test or pow_test, there are many of them, for which we have standard functions to use.

Right, I think moving forward once we refactor other tests, we should replace the hardcoded values there as well.

kahmed10 (approved):

src/targets/gpu/device/reverse.cpp
scxiao: @cagery For whatever reason, 1024 for gather does not work correctly.

test/verify/test_reverse.cpp
```diff
@@ -0,0 +1,18 @@
...
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        migraphx::shape s{migraphx::shape::float_type, {4, 16}};
+        auto a0  = mm->add_parameter("data", s);
+        int axis = 0;
+        mm->add_instruction(migraphx::make_op("reverse", {{"axis", axis}}), a0);
+        return p;
+    }
+};
```
scxiao, line 18: Maybe you can add a test for axis to be a negative number.

pfultz2 (approved):

scxiao (approved):

test/verify/test_reverse.cpp
cagery: Added `test/verify/test_reverse_negaxis.cpp`

PR #847, opened 2021-06-04
Update parse slice

cagery:
following `reverseOp` branch, this introduces changes to `parse_slice.cpp` for negative stepping of slice using reverse op.


src/onnx/parse_slice.cpp
```diff
@@ -77,7 +86,43 @@ struct parse_slice : op_parser<parse_slice>
...
+                else if((start_v < 0) & (end_v > INT_MIN) & (end_v < 0))
+                {
+                    op.ends[axis]   = lens[axis] + start_v + 1;
+                    op.starts[axis] = end_v - INT_MIN;
+                }
+            }
+
+            auto ins = info.add_instruction(op, args[0]);
+
+            for(auto axis : axes_v)
```
scxiao, line 116: Considering the cases here, I would suggest extend the reverse op to support multiple axes. We need the following changes:
1) Change the axis attribute in the reverse from int64_t axis to "std::vector<int64_t> axes"
2) For the ref compute function, change it to:

```
    argument compute(const shape& s, std::vector<argument> args) const
    {
        argument result{s};
        auto lens = s.lens();
        visit_all(result, args.front())([&](auto output, auto input) {
            shape_for_each(s, [&](const auto& out_idx) {
                auto in_idx              = out_idx;
                for (const auto& axis : axes)
                {
                    in_idx[axis] = lens[axis] - 1 - out_idx[axis];
                }
                output[s.index(out_idx)] = input[s.index(in_idx)];
            });
        });

        return result;
    }
```

For the device function, the implementation is:

```
argument reverse(hipStream_t stream, argument result, argument arg1, const std::vector<int64_t> axes)
{
    auto s       = arg1.get_shape();
    auto lens    = s.lens();
    std::size_t nelements = s.elements();
    visit_all(result, arg1)([&](auto output1, auto input1) {
        hip_visit_views(output1, input1, s)([&](auto output, auto input, auto hs) {
            gs_launch(stream, nelements)([=](auto i) __device__ {
                auto idx     = hs.multi(i);
                auto in_idx  = idx;
                for (auto axis : axes)
                {
                    in_idx[axis] = lens[axis] - 1 - idx[axis];
                }
                output[idx]  = input[in_idx];
            });
        });
    });

    return result;
}
```

scxiao: With supporting axes, we donot need the for loop here anymore.

src/onnx/parse_slice.cpp
```diff
@@ -30,7 +32,14 @@ struct parse_slice : op_parser<parse_slice>
             step_arg.visit([&](auto s) { steps.assign(s.begin(), s.end()); });
             if(!std::all_of(steps.begin(), steps.end(), [](auto s) { return s == 1; }))
             {
-                MIGRAPHX_THROW("PARSE_SLICE: cannot handle step other than 1");
+                if(std::all_of(steps.begin(), steps.end(), [](auto s) { return s == -1; }))
```
pfultz2, line 35: This should check `abs(s) == 1` then throw the error.

src/onnx/parse_slice.cpp
```diff
@@ -20,6 +20,8 @@ struct parse_slice : op_parser<parse_slice>
     {
         op::slice op;
 
+        bool reverse_direction = false;
```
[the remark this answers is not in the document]
scxiao: I think we do not even need the steps here since we did not change the slice implementation at all. We only change the way parsing the slice operator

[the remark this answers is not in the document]
scxiao: BTW, since we have the step operator now, we might be able to handle steps of not 1

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +86,39 @@ struct parse_slice : op_parser<parse_slice>
             op.axes = axes;
         }
 
-        return info.add_instruction(op, args[0]);
+        if(reverse_direction)
+        {
+
+            migraphx::argument axes_arg = args.at(3)->eval();
+            std::vector<int> axes_v;
+            axes_arg.visit([&](auto s) { axes_v.assign(s.begin(), s.end()); });
```
pfultz2, line 94: Only the axes that have a negative step should be copied to the reverse operator:

```cpp
std::vector<int64_t> raxes;
for(auto i:range(op.axes))
{
    if (steps[i] < 0)
        raxes.push_back(op.axes[i]);
}
```

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +86,39 @@ struct parse_slice : op_parser<parse_slice>
...
+
+            for(auto axis : axes_v)
+            {
+                auto start_v = op.starts[axis];
+                auto end_v   = op.ends[axis];
+                if((start_v < 0) && (end_v < INT_MIN))
+                {
+                    op.ends[axis]   = lens[axis] + start_v + 1;
+                    op.starts[axis] = 0;
+                }
```
pfultz2, line 106: Why are you normalizing the start and ends? This seems unnecessary.

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +86,39 @@ struct parse_slice : op_parser<parse_slice>
...
+                else if((start_v < 0) && (end_v > INT_MIN) && (end_v < 0))
+                {
+                    op.ends[axis]   = lens[axis] + start_v + 1;
+                    op.starts[axis] = end_v - INT_MIN;
+                }
+            }
+
+            auto ins = info.add_instruction(op, args[0]);
```
pfultz2, line 114: This should be moved outside of the if statement and then only construct reverse when raxes is not empty:

```cpp
auto ins = info.add_instruction(op, args[0]);
if (not raxes.empty())
    return info.add_instruction(make_op("reverse", {{"axes", raxes}}), ins);
else
    return ins;
```

src/onnx/parse_slice.cpp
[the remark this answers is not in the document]
pfultz2: But we need the steps variable so we can insert step/reverse operators. It should be a variable here since we can't store it in the operators(whereas axes/start/ends are store in the slice op). We also won't store the steps directly in the step operator either since it we need to process it for reverse.

> BTW, since we have the step operator now, we might be able to handle steps of not 1

Yea, I just assume for now he was just adding reversing, but it should be done in a way we can easily insert step later on.

src/onnx/parse_slice.cpp
pfultz2: You can easily add the step operator here:

```cpp
auto ins = info.add_instruction(op, args[0]);
if (std::any_of(steps.begin(), steps.end(), [](auto s) { return std::abs(s) != 1 }))
{
    std::vector<int64_t> nsteps;
    std::transform(steps.begin(), steps.end(), std::back_inserter(nsteps), [](auto s) {return std::abs(s); });
    ins = info.add_instruction(make_op("step", {{"axes", op.axes}, {"steps", nsteps}}));
}
if (not raxes.empty())
    return info.add_instruction(make_op("reverse", {{"axes", raxes}}), ins);
else
    return ins;
```

But this would probably entail more unit testing, so I guess we can add it later on since you need it to run a model already.

src/onnx/parse_slice.cpp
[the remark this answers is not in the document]
scxiao: ****

> But we need the steps variable so we can insert step/reverse operators. It should be a variable here since we can't store it in the operators(whereas axes/start/ends are stored in the slice op). We also won't store the steps directly in the step operator either since we need to process it for reverse.
> 
> > BTW, since we have the step operator now, we might be able to handle steps of not 1
> 
> Yea, I just assume for now he was just adding reversing, but it should be done in a way we can easily insert step later on.

I had thought this is the slice operator implementation file, so I think we should not have the steps. Yes, I agree we should have the "step" in the parse_slice file.

src/onnx/parse_slice.cpp
cagery: one of the cases was start = -1 and end = INT_MIN, this caused issues (miscalculation the boundaries from slice op implementation) so I have normalized it, as if it is slice op with step=1

src/onnx/parse_slice.cpp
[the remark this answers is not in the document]
cagery: Makes sense, will do

src/onnx/parse_slice.cpp
pfultz2: I think we can just swap those values:

```cpp
for(auto axes:raxes)
    std::swap(op.starts[axes], op.ends[axes]);
```

It should use the same normalization that slice already uses, if not, then we should fix our normalization.

scxiao: The slice start and ends works as [starts, end), i.e., including starts, execluding ends. so, for positive steps, starts (smaller indices) are included, ends(larger values) are  excluded. For negative steps, starts(larger indices) are included and ends (smaller indices) are excluded, so we cannot just do a swap.  If all starts and ends are positive, we need to add one after swapping. 

For negative starts and ends, the scenarios is more complex. For example, -1 indicates the last dimension, i.e. dim_size -1. If we add 1 directly, it becomes 0, which is the first dimension.

src/onnx/parse_slice.cpp
cagery: Done

src/onnx/parse_slice.cpp
cagery: Done

src/onnx/parse_slice.cpp
cagery: Done

src/onnx/parse_slice.cpp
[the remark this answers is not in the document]
cagery: Done, added steps as well.

cagery:
Addressed the latest inputs. I have not added the tests yet, will do in a bit. 
Currently seeing a test problem with `logsoftmax_nonstd_input_test` and trying to debug it. 
```
RUN    ] logsoftmax_nonstd_input_test
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 3) >= this->size() (which is 1)
<end of output>
Test time =   0.17 sec
```

cagery:
> Addressed the latest inputs. I have not added the tests yet, will do in a bit.
> Currently seeing a test problem with `logsoftmax_nonstd_input_test` and trying to debug it.
> 
> ```
> RUN    ] logsoftmax_nonstd_input_test
> terminate called after throwing an instance of 'std::out_of_range'
>   what():  vector::_M_range_check: __n (which is 3) >= this->size() (which is 1)
> <end of output>
> Test time =   0.17 sec
> ```

the test was missing axis=-1, added it.

src/onnx/parse_slice.cpp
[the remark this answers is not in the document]
cagery: I had to remove adding step operator itself. Variable is still there.

src/onnx/parse_slice.cpp
cagery: I have created the issue #855 to do the addition of step operator later.

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +78,37 @@ struct parse_slice : op_parser<parse_slice>
...
+                if(op.starts[axis] == 0)
+                {
+                    op.starts[axis] = INT_MAX;
+                }
+                op.ends[axis] += 1;
+                std::swap(op.starts[axis], op.ends[axis]);
+            }
+        }
+
+
```
scxiao, line 103: please remove the unnecessary empty lines

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +78,37 @@ struct parse_slice : op_parser<parse_slice>
             op.axes = axes;
         }
 
-        return info.add_instruction(op, args[0]);
+        std::vector<int64_t> raxes;
+        if(std::any_of(steps.begin(), steps.end(), [](auto s) { return s < 0; }))
```
pfultz2, line 82: You dont need this branch, as the for loops will skip when `raxes` is empty.

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +78,37 @@ struct parse_slice : op_parser<parse_slice>
...
+            }
+        }
+
+        
+
+        auto ins = info.add_instruction(op, args[0]);
+        if(not raxes.empty())
+        {
+            return info.add_instruction(make_op("reverse", {{"axes", raxes}}), ins);
+        }
```
pfultz2, line 109: Really small, but maybe remove the braces so its consistent with the else branch or add braces to the else.

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +78,37 @@ struct parse_slice : op_parser<parse_slice>
...
+            
+            auto lens = args[0]->get_shape().lens();
+            for(auto axis : raxes)
+            {
+                op.starts[axis] += 1;
+                if(op.starts[axis] == 0)
+                {
+                    op.starts[axis] = INT_MAX;
+                }
+                op.ends[axis] += 1;
```
pfultz2, line 98: This seems like a simple way to adjust the starts and ends. Looks good!

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +78,37 @@ struct parse_slice : op_parser<parse_slice>
...
+        std::vector<int64_t> raxes;
+        if(std::any_of(steps.begin(), steps.end(), [](auto s) { return s < 0; }))
+        {
+            for(auto i : op.axes)
+            {
+                if(steps[i] < 0)
+                    raxes.push_back(op.axes[i]); // populate negative only axis indices
+            }
+            
+            auto lens = args[0]->get_shape().lens();
```
pfultz2, line 90: It looks like this variable is never used.

src/onnx/parse_slice.cpp
```diff
@@ -77,7 +78,37 @@ struct parse_slice : op_parser<parse_slice>
             op.axes = axes;
         }
 
-        return info.add_instruction(op, args[0]);
+        std::vector<int64_t> raxes;
+        if(std::any_of(steps.begin(), steps.end(), [](auto s) { return s < 0; }))
+        {
+            for(auto i : op.axes)
```
pfultz2: This should be `for(auto i : range(op.axes.size()))`. This should fix the UBSAN error.

src/onnx/parse_slice.cpp
cagery: Artifact, removed.

src/onnx/parse_slice.cpp
cagery: Removed the braces, looks cleaner!

src/onnx/parse_slice.cpp
cagery: Done

src/onnx/parse_slice.cpp
cagery: I had to add this, otherwise `logsoftmax_nonstd_input_test` kept segfaulting:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 3) >= this->size() (which is 1)
```

src/onnx/parse_slice.cpp
cagery: Undefined behavior is gone but now I am seeing ASAN problem:

```
SUMMARY: AddressSanitizer: heap-buffer-overflow (/opt/rocm/lib/libmigraphx_onnx.so.1+0x8a7c0a) 
Shadow bytes around the buggy address:
[...]
```

gdb:
```
addition of unsigned offset to 0x602000000dd0 overflowed to 0x602000000dc8
```

backtrace:
```
#6  0x00007ffff71b5c24 in migraphx::version_1::onnx::onnx_parser::parse_graph (this=<optimized out>, mod=<optimized out>, graph=...)
    at /code/AMDMIGraphX/src/onnx/onnx_parser.cpp:332
#7  0x00007ffff71b3c53 in migraphx::version_1::onnx::onnx_parser::parse_from (this=<optimized out>, is=..., Python Exception <class 'OverflowError'> int too big to convert: 
name=)
    at /code/AMDMIGraphX/src/onnx/onnx_parser.cpp:235
```

pfultz2: Can you push your changes? I'll take a look tomorrow to see what the issue is.

cagery: Done, thanks!

src/onnx/parse_slice.cpp
pfultz2: I think thats masking a bug.

src/onnx/parse_slice.cpp
```diff
@@ -4,6 +4,7 @@
 #include <migraphx/ranges.hpp>
 #include <migraphx/instruction.hpp>
 #include <migraphx/make_op.hpp>
+#include <cstdlib>
```
pfultz2, line 7: This header can be removed.

cagery: done

PR #853, opened 2021-06-09
Loop operator

scxiao:
This PR is for the implementation of the Loop operator for opset version 13.
Notes: 1) Default max iteration number is 10 if no max iteration number is provided
           2) To change the max iter number, a user can set the max_loop_iterations in the onnx_option struct when parsing a model.
           3) The returned shape of the scan output is from the max_loop_iterations even the actual loop num is less than that. This issue also applies to other operators like NonZero and NonMaxSuppression. A issue #948 is created to track this and to be resolved later.


src/onnx/parse_loop.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    std::vector<instruction_ref> parse(const op_desc& /*opd*/,
+                                       onnx_parser& parser,
+                                       const onnx_parser::node_info& info,
+                                       std::vector<instruction_ref> args) const
+    {
+        // default value of the max_iter_num
+        int64_t max_iter_num = parser.max_iter_num;
+        // iteration input is empty
+        if(args.at(0)->name() == "undefined")
```
kahmed10, line 27: should we make a test case for this?

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,164 @@
...
+    argument compute(const shape& out_shape,
+                     const std::vector<argument>& args,
+                     const std::vector<module_ref>& mods,
+                     const std::function<std::vector<argument>(
+                         module_ref&, const std::unordered_map<std::string, argument>&)>& run) const
+    {
+        auto iter_num = args.at(0).at<int64_t>();
+        if(iter_num > max_iter_num)
+        {
+            MIGRAPHX_THROW("LOOP compute(): iter_num " + std::to_string(iter_num) +
```
kahmed10, line 90: I think the only way this would happen is if the user manually creates a loop operator and sets the input arg for `iter_num` to be larger than `max_iter_num`. In onnx parsing, we are currently always setting `max_iter_num` to whatever is specified by either the input arg or the onnx options. Should we just make this an assert?

src/eliminate_data_type.cpp
```diff
@@ -3,17 +3,19 @@
...
 #include <migraphx/make_op.hpp>
 #include <migraphx/instruction.hpp>
+#include <migraphx/ranges.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 void eliminate_data_type::apply(module& m) const
 {
+    static std::vector<std::string> skip_op_names = {"convert", "get_tuple_elem", "if", "loop"};
```
pfultz2, line 13: This should be declared `const`.

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,164 @@
...
+                // name is for input
+                if(io_index.second)
+                {
+                    params[name] = in_args.at(io_index.first);
+                }
+            }
+
+            auto mod_args = run(mod, params);
+            // copy loop carried dependency from mod outputs to inputs
+            std::copy(mod_args.begin(), mod_args.begin() + dep_num + 1, in_args.begin() + 1);
```
pfultz2, line 134: What is `dep_num`?

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,164 @@
...
+            cond = mod_args.at(0).at<bool>();
+
+            // concat scan outputs
```
pfultz2, line 137: If we are concating the output, I wonder if we can reuse the concat operator by maybe moving the implementation to a common function.

src/module.cpp
```diff
@@ -221,8 +221,6 @@ instruction_ref module::replace_instruction(instruction_ref ins,
 
 instruction_ref module::replace_instruction(instruction_ref ins, instruction_ref rep)
 {
-    assert(has_instruction(ins));
```
pfultz2, line 224: Why is this `assert` removed?

src/onnx/include/migraphx/onnx/onnx_parser.hpp
```diff
@@ -58,25 +59,30 @@ struct onnx_parser
...
+                         std::unordered_map<std::string, instruction_ref>& instructions);
 
     static int64_t get_opset_version(const onnx::ModelProto& model);
 
     void parse_from(std::istream& is, std::string name = "");
     void parse_from(const void* data, std::size_t size);
-    void parse_graph(module* mod, const onnx::GraphProto& graph);
+    void parse_graph(module* mod,
+                     const onnx::GraphProto& graph,
+                     std::unordered_map<std::string, instruction_ref> instructions = {},
```
pfultz2, line 84: Why is this passed as a parameter instead of using the member variable?

src/onnx/onnx_parser.cpp
```diff
@@ -222,45 +224,72 @@ int64_t onnx_parser::get_opset_version(const onnx::ModelProto& model)
...
+                map_param_names[name] = param_name;
+            }
+            mod_instructions[param_name] = mod->add_parameter(param_name, s);
         }
     }
 
+    for(const auto& n_ins : mod_instructions)
+    {
+        instructions[n_ins.first] = n_ins.second;
+    }
```
pfultz2, line 270: This can be `std::copy`.

src/pass_manager.cpp
```diff
@@ -1,3 +1,4 @@
+#include "migraphx/module_ref.hpp"
```
pfultz2, line 1: Use angle brackets.

src/targets/gpu/loop.cpp
```diff
@@ -0,0 +1,134 @@
...
+    std::vector<argument> out_args = {cpy_args.at(input_num)};
+    auto mod_outputs               = cpy_args.back().get_sub_objects();
+    out_args.insert(out_args.end(), mod_outputs.begin(), mod_outputs.end());
+
+    int64_t iter = 0;
+    for(iter = 0; (iter < iter_num) and cond; ++iter)
+    {
+        // copy iter num and cond to device memory
+        (void)hipMemcpy(in_args.at(0).data(), &iter, sizeof(int64_t), hipMemcpyHostToDevice);
+        (void)hipMemcpy(in_args.at(1).data(), &cond, sizeof(bool), hipMemcpyHostToDevice);
```
pfultz2, line 77: Why cant you use `to_gpu` to copy this to the GPU?

src/targets/ref/lowering.cpp
```diff
@@ -1,4 +1,5 @@
 
+#include "migraphx/module_ref.hpp"
```
pfultz2, line 2: Angle brackets for includes.

test/verify/run_verify.cpp
```diff
@@ -99,7 +99,22 @@ std::pair<migraphx::program, std::vector<migraphx::argument>> run_verify::run_ta
...
+                {
+                    arg_outs.push_back(t.allocate(ss));
+                }
+                m[x.first] = migraphx::argument(arg_outs);
+            }
+            else
+            {
+                m[x.first] = t.allocate(x.second);
+            }
+        }
```
pfultz2, line 117: This should be done in the allocate function. And `argument` constructor should be extended to support allocating tuple types.

test/verify/test_loop.cpp
```diff
@@ -0,0 +1,45 @@
+
+#include "migraphx/literal.hpp"
```
pfultz2, line 2: Angle brackets.

src/targets/gpu/loop.cpp
```diff
@@ -0,0 +1,134 @@
...
+                  const std::vector<module_ref>& mods,
+                  const std::function<std::vector<argument>(
+                      module_ref&, const std::unordered_map<std::string, argument>&)>& run) const
+{
+    auto cpy_args = args;
+    std::vector<argument> cpu_args;
+    cpu_args.push_back(cpy_args.at(1));
+    cpu_args.push_back(cpy_args.at(3));
+    cpy_args.erase(cpy_args.begin() + 3);
+    cpy_args.erase(cpy_args.begin() + 1);
```
pfultz2, line 50: Why erase these? Why not just copy the arguments you need?

src/targets/gpu/loop.cpp
```diff
@@ -0,0 +1,134 @@
...
+                }
+                else
+                {
+                    params[name] = out_args.at(io_index.first);
+                }
+            }
+        }
+
+        auto mod_args = run(mod, params);
+        gpu_sync();
```
pfultz2, line 106: This should be `ctx.finish()`

src/targets/gpu/loop.cpp
pfultz2, line 77: Use the type safe `copy_to_gpu` and `copy_from_gpu`.

src/targets/gpu/loop.cpp
```diff
@@ -0,0 +1,134 @@
...
+hip_loop::compute(const shape&,
+                  const std::vector<argument>& args,
+                  const std::vector<module_ref>& mods,
+                  const std::function<std::vector<argument>(
+                      module_ref&, const std::unordered_map<std::string, argument>&)>& run) const
+{
+    auto cpy_args = args;
+    std::vector<argument> cpu_args;
+    cpu_args.push_back(cpy_args.at(1));
+    cpu_args.push_back(cpy_args.at(3));
```
pfultz2, line 48: I dont see the point of this vector. I think it would be better to have variable names for each cpu_arg:

```cpp
const auto& cond_arg = args.at(1));
```

src/targets/gpu/loop.cpp
pfultz2: And instead of doing `in_args.at(0)` can we assign a name to it such as `gpu_cond_arg = args.at(0)`.

src/eliminate_data_type.cpp
scxiao: fixed

src/include/migraphx/op/loop.hpp
scxiao: It is the number of loop carried dependencies.

src/onnx/include/migraphx/onnx/onnx_parser.hpp
scxiao: In this way, subgraph can access variables in parent graph but not in the other way.

src/onnx/onnx_parser.cpp
scxiao: fixed

src/pass_manager.cpp
scxiao: fixed

src/targets/ref/lowering.cpp
scxiao: fixed

src/onnx/include/migraphx/onnx/onnx_parser.hpp
pfultz2: That would be an invalid onnx file. I would prefer to keep it as a member variable. If we want to make sure we dont reference it by accident then we could remove the instructions in the map when `parse_graph` finishes.

test/verify/run_verify.cpp
scxiao: fixed

src/targets/gpu/loop.cpp
scxiao: fixed

src/include/migraphx/onnx.hpp
```diff
@@ -15,6 +15,7 @@ struct onnx_options
     /// Explicitly specify the dims of an input
     std::unordered_map<std::string, std::vector<std::size_t>> map_input_dims = {};
     /// Continue parsing onnx file if an unknown operator is found
+    int64_t max_iter_num        = 10;
```
pfultz2, line 18: The comment above this applies to `skip_unknown_operators`. Also, please add a comment for this attribute.

pfultz2:
Is is possible to unify the loop implementation and have a class with customization points to deal with the difference between cpu and gpu? Something like:

```cpp
template<class LoopModel>
argument run_loop(LoopModel loop_model, ...)
{
    // Do loop op
}
```

A loop model class can be created for cpu and another one for gpu. I am not sure exactly how the loop model class would look, but maybe you have a better idea since you have implemented both loop for both cpu and gpu.

src/include/migraphx/onnx.hpp
scxiao: fixed

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+        return shape(ins_out_shapes);
+    }
+
+    struct ref_loop
+    {
+        int64_t max_iter_num = 0;
+
+        template <class T>
+        void copy_arg(context&, const argument& arg, T& var, bool from_var) const
```
pfultz2, line 67: You should have two overloads instead of the boolean parameter:

```cpp
template <class T>
void copy_arg(context&, const argument& src, T& dst) const;
template <class T>
void copy_arg(context&, const T& src, const argument& dst) const;
```

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,139 @@
...
+        {
+            char* ptr = reinterpret_cast<char*>(&var);
+            if(from_var)
+            {
+                std::copy(ptr, ptr + arg.get_shape().bytes(), arg.data());
```
pfultz2, line 72: This could be a bufferoverun if arg is much larger than the `sizeof(T)`. I think it would be better to write this as `*arg.cast<T> = var`.

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,139 @@
...
+            }
+            else
+            {
+                std::copy(arg.data(), arg.data() + arg.get_shape().bytes(), ptr);
```
pfultz2, line 76: Yea, I think this could be written as `var = *arg.cast<T>()`

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,139 @@
...
+            {
+                const auto& mod_out  = mod_scan_outs.at(i);
+                const auto& scan_out = scan_outputs.at(i);
+
+                auto* in_data        = mod_out.data();
+                auto* out_data       = scan_out.data();
+                std::size_t out_size = mod_out.get_shape().bytes();
+                std::copy(in_data, in_data + out_size, out_data + iter * out_size);
+            }
+        }
```
pfultz2, line 95: This looks more like it appends the `mod_scan_outs`, Maybe we should rename this function `append`?

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,139 @@
...
+                std::copy(ptr, ptr + arg.get_shape().bytes(), arg.data());
+            }
+            else
+            {
+                std::copy(arg.data(), arg.data() + arg.get_shape().bytes(), ptr);
+            }
+        }
+
+        void concat_scan_outputs(const std::vector<argument>& mod_scan_outs,
+                                 const std::vector<argument>& scan_outputs,
```
pfultz2, line 81: The parameter names makes it harder to see what is input and what is output.

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,139 @@
...
+            assert(mod_scan_outs.size() == scan_outputs.size());
+            for(auto i : range(mod_scan_outs.size()))
+            {
+                const auto& mod_out  = mod_scan_outs.at(i);
+                const auto& scan_out = scan_outputs.at(i);
+
+                auto* in_data        = mod_out.data();
+                auto* out_data       = scan_out.data();
+                std::size_t out_size = mod_out.get_shape().bytes();
+                std::copy(in_data, in_data + out_size, out_data + iter * out_size);
```
pfultz2, line 93: We should assert that `out_data` has enough buffer to copy.

src/include/migraphx/run_loop.hpp
```diff
@@ -0,0 +1,119 @@
...
+    std::string param_prefix = "#" + mod->name() + "_in_";
+
+    std::vector<argument> in_args(args.begin(), args.begin() + input_num);
+    std::vector<argument> out_args = {args.at(input_num)};
+    auto mod_outputs               = args.back().get_sub_objects();
+    out_args.insert(out_args.end(), mod_outputs.begin(), mod_outputs.end());
+
+    std::vector<argument> scan_outputs(out_args.begin() + dep_num + 1, out_args.end());
+    int64_t iter = 0;
+    for(iter = 0; (iter < iter_num) and cond; ++iter)
```
pfultz2, line 66: I think its cleaner to write `(iter < iter_num and cond` without the extra parens.

src/targets/gpu/loop.cpp
```diff
@@ -0,0 +1,70 @@
...
+        else
+        {
+            copy_from_gpu(ctx, arg, arg_var);
+        }
+    }
+
+    void
+    concat_scan_outputs(const std::vector<argument>&, const std::vector<argument>&, const int) const
+    {
+    }
```
pfultz2, line 40: Why is this empty?

src/generate.cpp
```diff
@@ -6,32 +6,81 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+        std::transform(sub_ss.begin(), sub_ss.end(), std::back_inserter(sub_args), [&](auto ss) {
+            argument temp_arg;
+            ss.visit_type([&](auto as) {
+                using type = typename decltype(as)::type;
+                auto v     = fill_tensor_data<type>(ss, value);
+                temp_arg   = {ss, v};
+            });
+
+            return temp_arg;
+        });
```
pfultz2, line 22: This should call generate recursively because if this is a tuple within a tuple it will throw an error.

src/targets/gpu/loop.cpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        if(iter >= max_iter_num)
+            return;
+
+        auto elem_num = max_iter_num - iter;
+        for(const auto& out : scan_outputs)
+        {
+            auto s    = out.get_shape();
+            auto size = s.bytes() / max_iter_num;
+            (void)hipMemset(out.data() + iter * size, 0, size * elem_num);
```
pfultz2, line 52: This should be a kernel so it can run async.

pfultz2: You can implement it in migraphx::gpu::device using `nary`.

src/include/migraphx/op/loop.hpp
scxiao: fixed

src/include/migraphx/op/loop.hpp
scxiao: fixed

src/include/migraphx/op/loop.hpp
scxiao: fixed

src/include/migraphx/op/loop.hpp
scxiao: fixed

src/include/migraphx/op/loop.hpp
scxiao: renamed them.

src/include/migraphx/run_loop.hpp
scxiao: fixed

src/targets/gpu/loop.cpp
scxiao: This have been done implicitly by use the argument as the mod output parameters.

src/generate.cpp
scxiao: fixed

src/targets/gpu/loop.cpp
scxiao: fixed

src/targets/gpu/loop.cpp
pfultz2: Couldn't you do something similiar where you allocate for the mod output parameters on ref implementation? And then we can remove this method.

src/include/migraphx/run_loop.hpp
```diff
@@ -0,0 +1,120 @@
...
+                  T& ctx,
+                  std::vector<argument> args,
+                  const std::vector<module_ref>& mods,
+                  const std::function<std::vector<argument>(
+                      module_ref&, const std::unordered_map<std::string, argument>&)>& run)
+{
+    std::vector<std::vector<argument>> results;
+    // process argu lists
+    auto iter_num = args.at(0).at<int64_t>();
+    auto cond     = args.at(2).at<bool>();
```
pfultz2, line 47: Why dont we put the iter_num and cond as the first two parameters? This will make it easier to slice the args.

src/include/migraphx/run_loop.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    args.erase(args.begin() + 2);
+    args.erase(args.begin());
+
+    auto input_num = args.size() - 2;
+    auto dep_num   = input_num - 2;
+
+    module_ref mod           = mods.at(0);
+    auto param_name_shapes   = mod->get_parameter_shapes();
+    std::string param_prefix = "#" + mod->name() + "_in_";
```
pfultz2, line 57: I understand why we add a prefix when doing the onnx parser so each parameter can be orderded correctly. However, for `run_loop` it should just sort the parameter names and the index into this sorted vector is parameter that should be used.

If we(or a user) build a loop operator manually then we might not name this operators exactly the same. The important part is that each arg is associated with each parameter as its sorted.

src/onnx/include/migraphx/onnx/onnx_parser.hpp
pfultz2: I still think this would be good to keep instructions as a member variable.

src/include/migraphx/onnx.hpp
```diff
@@ -18,6 +18,8 @@ struct onnx_options
     bool skip_unknown_operators = false;
     /// Print program if an error occurs
     bool print_program_on_error = false;
+    /// Max iter num for the loop operator
+    int64_t max_iter_num = 10;
```
pfultz2, line 22: This should be named `max_iterations` or `max_loop_iterations`.

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,139 @@
...
+#include <utility>
+#include <set>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct loop
+{
+    int64_t max_iter_num = 0;
```
pfultz2, line 26: This should also be named `max_iterations`.

src/api/migraphx.py
```diff
@@ -243,6 +243,11 @@ def onnx_options(h):
         api.params(value='size_t'),
         invoke='migraphx::set_default_dim_value($@)',
     )
+    h.method(
+        'set_default_loop_num',
```
pfultz2, line 247: This should be named `set_max_loop_iterations`.

src/targets/gpu/loop.cpp
scxiao: In the reference implementation, returned arguments for output parameters are allocated in operators, I am not sure if there is a way to use an external buffers store the output parameters.

src/include/migraphx/run_loop.hpp
scxiao: The problem is, some input parameters can be deleted when we compile a module. If some parameters are removed, even we sort the parameters, we do not know the relationship between input arguments and corresponding parameters since some parameters could have been removed. Maybe we should not remove input parameters in dead_code_elimination, then we can do what you proposed above.

src/onnx/include/migraphx/onnx/onnx_parser.hpp
scxiao: fixed

src/targets/gpu/loop.cpp
scxiao: fixed

kahmed10 (approved):

src/include/migraphx/op/loop.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+        template <class T>
+        void copy(context&, T src, const argument& dst) const
+        {
+            *dst.cast<T>() = src;
+        }
+
+        void append(const std::vector<argument>& iter_state,
+                    const std::vector<argument>& concatenated_outputs,
+                    const int iter) const
```
pfultz2, line 76: Dont make a value parameter `const`.

src/include/migraphx/run_loop.hpp
```diff
@@ -0,0 +1,124 @@
...
+template <class LoopModel, class T>
+argument run_loop(const LoopModel& model,
+                  T& ctx,
+                  std::vector<argument> args,
+                  const std::vector<module_ref>& mods,
+                  const std::function<std::vector<argument>(
+                      module_ref&, const std::unordered_map<std::string, argument>&)>& run)
+{
+    auto get_output_index = [](const std::string& name) {
+        std::string out_prefix = "#output_";
```
pfultz2, line 24: The naming of output indices as `#output_` is specific to the GPU backend. Other backends could name them differently. Why do you need to check for the output params? Maybe this should added to the loop_model, maybe a function like `std::map<std::string, int> get_output_params(const module& m)`?

src/include/migraphx/run_loop.hpp
```diff
@@ -0,0 +1,124 @@
...
+        auto loc               = name.find(out_prefix);
+        if(loc != std::string::npos)
+        {
+            int index = std::stoi(name.substr(loc + out_prefix.size()));
+            return index;
```
pfultz2, line 29: It says there is no coverage here. There are other places where there is no coverage in this file we should probably add unit tests for. I assume there is no coverage because output params are only used by the GPU backend. Perhaps, we can create mock loop_model and some fake output params to check with the unit tests.

src/onnx/onnx_parser.cpp
```diff
@@ -224,28 +224,42 @@ int64_t onnx_parser::get_opset_version(const onnx::ModelProto& model)
...
+    // the parent module
+    std::unordered_map<std::string, instruction_ref> pmod_insts_bkup;
+    for(auto& ins_p : mod_insts)
+    {
+        if(contains(instructions, ins_p.first))
+        {
+            pmod_insts_bkup[ins_p.first] = instructions[ins_p.first];
+        }
+        instructions[ins_p.first] = ins_p.second;
+    }
```
pfultz2, line 261: What is this doing? Is this to handle parameter shadowing? I thought shadowing is not permitted by ONNX spec from [Nodes](https://github.com/onnx/onnx/blob/master/docs/IR.md#nodes):

> The graph MUST use single static assignment for all node outputs, which means that all node output names MUST be unique within a graph. In the case of a nested subgraph, a node output name MUST be distinct from the names from the outer scopes that are visible in the nested subgraph.

Is that an incorrect interpretation? Do you see onnx models that are shadowing params and constants?

src/targets/gpu/device/fill.cpp
```diff
@@ -0,0 +1,17 @@
+#include <migraphx/gpu/device/fill.hpp>
+#include <migraphx/gpu/device/nary.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+void fill(hipStream_t stream, const argument& result, const unsigned long& val)
```
pfultz2, line 9: Pass `val` by value.

src/targets/gpu/lowering.cpp
```diff
@@ -469,9 +470,48 @@ struct miopen_apply
...
+    {
+        apply_map.emplace("loop", [=](instruction_ref ins) {
+            std::vector<instruction_ref> inputs = ins->inputs();
+            // copy max_iter from gpu to cpu
+            auto cpu_max_iter =
+                mod->insert_instruction(ins, make_op("hip::copy_from_gpu"), inputs.at(0));
+            auto cpu_cond =
+                mod->insert_instruction(ins, make_op("hip::copy_from_gpu"), inputs.at(1));
+            auto synced_max_iter =
+                mod->insert_instruction(ins, make_op("hip::sync_stream"), cpu_max_iter);
```
pfultz2, line 485: Shouldn't this take both `cpu_max_iter` and `cpu_cond` as inputs? So it will sync the stream after copying both variables?

test/ref_ops_test.cpp
```diff
@@ -2439,6 +2439,100 @@ TEST_CASE(logsoftmax_test_axis_3)
     EXPECT(migraphx::verify_range(results_vector, s));
 }
 
+TEST_CASE(loop_test)
```
pfultz2, line 2442: Can we put the ref loop unit tests in a seperate cpp file?

test/ref_ops_test.cpp
```diff
@@ -2439,6 +2439,100 @@ TEST_CASE(logsoftmax_test_axis_3)
...
+        std::vector<std::vector<int64_t>> res;
+        for(auto& arg : rets)
+        {
+            std::vector<int64_t> vec;
+            arg.visit([&](auto v) { vec.assign(v.begin(), v.end()); });
+            res.push_back(vec);
+        }
+
+        return res;
+    };
```
pfultz2, line 2497: I think it would be better to make this a regular function and make all the test cases seperate `TEST_CASE` cases in the test suite. This will make it easier to debug in the future as we can just run one test case(instead of all of them).

src/onnx/onnx_parser.cpp
scxiao: We have another scenario that subgraph and parent graph have parameters with the same name. In this case, when we parse the subgraph, the instruction (stored in the mem var ```instructions```  corresponding to the parameter in the subgraph will overwrite that of the parent graph. After parsing subgraph returns, if we do not restore the parameter of the parent graph and and the parent graph uses this parameter as input for some other operators, inputs of the operator are the parameter of the subgraph.

In this scenario, the name is not output of any node, so the above specification does not apply, and there is no specification saying parent graph and subgraph cannot use the same name as parameter names. The only thing I read from the notes is:

>In nested subgraphs used as attribute values, users MUST NOT use the same name as both a subgraph initializer and subgraph input unless the corresponding op's specification explicitly allows it.

but this does not apply, either. 

I got the example with the above scenario some time ago online, but I cannot find it anymore. Also, when I created the onnx file from the script, no error is reported.

src/include/migraphx/op/loop.hpp
scxiao: fixed

src/include/migraphx/run_loop.hpp
scxiao: moved this to the gpu implementation

src/include/migraphx/run_loop.hpp
scxiao: Yes, this part of code only applied to the GPU implementation.

src/targets/gpu/lowering.cpp
scxiao: fixed

test/ref_ops_test.cpp
scxiao: changed

test/ref_ops_test.cpp
scxiao: changed

src/onnx/onnx_parser.cpp
scxiao: @pfultz2 Changed accordingly and added a unit test for the run loop. Could you please take a look again? Thanks

src/onnx/onnx_parser.cpp
```diff
@@ -224,28 +224,42 @@ int64_t onnx_parser::get_opset_version(const onnx::ModelProto& model)
...
+        if(!contains(mod_insts, name))
         {
+            // ONNX specification does not specify hwo to deal with the
+            // scenario that a nested subgraph contains a parameter with the
+            // name existed in its parent graph.
+            // In the current implementation, MIGraphX throws an exception for that.
+            if(contains(instructions, name))
+            {
+                MIGRAPHX_THROW("module \"" + mod->name() + "\" has parameter name \"" + name +
+                               "\" existing in paraent graph!");
```
pfultz2, line 247: `parent`.

test/run_loop_test.cpp
```diff
@@ -0,0 +1,290 @@
...
+        {
+            auto lens = out_s.lens();
+            lens.insert(lens.begin(), max_iterations);
+            ins_out_shapes.push_back({out_s.type(), lens});
+        }
+
+        return migraphx::shape(ins_out_shapes);
+    }
+
+    struct test_loop
```
pfultz2, line 78: Couldn't this inherit from `ref_loop` then just override `get_output_params`?

scxiao: changed

src/onnx/onnx_parser.cpp
scxiao: Sorry, fixed. Thanks!

pfultz2 (approved):

PR #854, opened 2021-06-09
Test runner

scxiao:


src/api/api.cpp
```diff
@@ -171,6 +171,18 @@ operation create_op(const char* name, const char* attributes)
     return op;
 }
 
+std::size_t get_parameter_number(program& p) { return p.get_parameter_names().size(); }
```
pfultz2, line 174: I dont see a reason to add this API as this is the same as `p.get_parameter_names().size()`, which the user can call directly.

pfultz2:
Is there a reason we need to keep this as seperate app and call only through our API for C++? Also, is it necessary to have a python and C++ version? It seems to duplicate a lot of code.

If we decide to only do C++, I think it would be better to add a C++ driver for this(under src/driver), this way we can reuse a lot of the classes we already have. You can use `argument_parser` class to parse arguments. We can also add a `literal parse_tensor_from(std::istream& is)` to our onnx parser to parse the tensors.

If we just do this in python, then we can put this python file under `test/onnx` or `tools/`.

test/py/onnx_backend_test.py
```diff
@@ -1,9 +1,12 @@
-import sys
+import sys, os
 if sys.version_info < (3, 0):
     sys.exit()
 
+# add the lib folder (containing migraphx) to sys.path
+sys.path.append(os.getcwd() + "/lib")
```
pfultz2, line 6: CMake should set the correct python path when running the backend tests, so this shouldn't be added.

test_runner/python/test_runner.py
```diff
@@ -0,0 +1,195 @@
+import sys, os
+import numpy as np
+import argparse
+import onnx
+from onnx import numpy_helper
+
+# add the lib folder (containing migraphx) to sys.path
+sys.path.append(os.getcwd() + "/lib")
```
pfultz2, line 8: I prefer not to make assumptions like that.

scxiao:
Removed the c++ version of the test_runner

test/py/test_runner.py
```diff
@@ -0,0 +1,192 @@
...
+        file_name = dir_name + '/' + item
+        if os.path.isfile(file_name) and file_name.endswith('.onnx'):
+            return item
+
+    return ''
+
+
+def read_pb_file(filename):
+    try:
+        pfile = open(filename, 'rb')
```
kahmed10, line 52: I think we should fix this deepcode issue by using `with(open(filename, 'rb')) as pfile:`

test/py/test_runner.py
```diff
@@ -0,0 +1,192 @@
...
+
+    return ret
+
+
+def tune_input_shape(model, input_data):
+    input_data_keys = input_data.keys()
+    param_shapes = model.get_parameter_shapes()
+    input_shapes = {}
+    for name, s in param_shapes.items():
+        assert name in input_data_keys
```
kahmed10, line 130: I think we can fix this deepcode warning and just say `assert name in input_data`

scxiao: fixed

test/py/test_runner.py
scxiao: fixed

kahmed10 (approved):

test/py/test_runner.py
```diff
@@ -0,0 +1,186 @@
...
+        ret = check_correctness(gold_output_data, output_data)
+        if ret:
+            correct_num += 1
+
+        output_str = "PASSED" if ret else "FAILED"
+        print("\tCase {}: {}".format(case_name, output_str))
+
+    print("\nTest \"{}\" has {} cases:".format(test_name, case_num))
+    print("\t Passed: {}".format(correct_num))
+    print("\t Failed: {}".format(case_num - correct_num))
```
pfultz2, line 182: If the test failed then we should return an error.

pfultz2:
It seems like the test_runner.py would go under tools/. Does the test_runner need our onnx backend?

CMakeLists.txt
```diff
@@ -241,3 +241,4 @@ foreach(py_file ${backend_files})
     configure_file(${py_file} ${DEST_DIR}/lib/onnx_migraphx/. COPYONLY)
 endforeach(py_file)
 configure_file(${CMAKE_SOURCE_DIR}/test/py/onnx_backend_test.py ${DEST_DIR}/onnx_backend_test.py COPYONLY)
+configure_file(${CMAKE_SOURCE_DIR}/test/py/test_runner.py ${DEST_DIR}/migraphx_test_runner.py COPYONLY)
```
pfultz2, line 244: Whats the reason for this?

pfultz2 (approved):

Issue #855, opened 2021-06-09
slice+step+reverse parsing


cagery:
during the testing slice+step+reverse for slice onnx operator experienced:

```
[   RUN    ] slice_5arg_negstep_test
terminate called after throwing an instance of 'migraphx::version_1::exception'
  what():  /code/AMDMIGraphX/src/include/migraphx/op/step.hpp:45: compute_shape: STEP: axis value is out of range!
```
Initial debug showed:
axis values:
-1,-2

lens: 
2, 4

--> Maybe normalization is needed?
--> Maybe better handling make_op(step) in parse_slice.cpp?

I am creating this issue to track, implement step into `parse_slice.cpp` and also understand how to fix it at next milestone.

cagery:
Fixed with #858

PR #858, opened 2021-06-11
Step op for parse_slice.cpp

### LINKED ISSUE #856: Codecov: line not covered by tests, parse_slice
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/847/checks?check_run_id=2797091629

TODO: Add tests for coverage.

cagery:
Adds step operator to `parse_slice.cpp` for abs(steps) != 1 for #855. 

This should also fix #856.


scxiao (commented):
Please add an unit test in the test/onnx/verify_onnx.cpp to ensure the implementation is correct.
Also, please add a gpu test to cover the scenario of negative steps in slice.

cagery:
Normalization in slice operator does not work for some cases, such as start = -1, end = -5 for an array of length 5, causing test failures. I am not sure if normalization by default supports such start > end & both negative cases. I am leaning towards enhancing parse_slice.cpp for such cases. Any thoughts? @scxiao @pfultz2

scxiao:
> Normalization in slice operator does not work for some cases, such as start = -1, end = -5 for an array of length 5, causing test failures. I am not sure if normalization by default supports such start > end & both negative cases. I am leaning towards enhancing parse_slice.cpp for such cases. Any thoughts? @scxiao @pfultz2

Can you put your model or the program at some place?

cagery:
> > Normalization in slice operator does not work for some cases, such as start = -1, end = -5 for an array of length 5, causing test failures. I am not sure if normalization by default supports such start > end & both negative cases. I am leaning towards enhancing parse_slice.cpp for such cases. Any thoughts? @scxiao @pfultz2
> 
> Can you put your model or the program at some place?

It is happening with the current test `slice_5arg_reverse_test`

scxiao:
> > > Normalization in slice operator does not work for some cases, such as start = -1, end = -5 for an array of length 5, causing test failures. I am not sure if normalization by default supports such start > end & both negative cases. I am leaning towards enhancing parse_slice.cpp for such cases. Any thoughts? @scxiao @pfultz2
> > 
> > 
> > Can you put your model or the program at some place?
> 
> It is happening with the current test `slice_5arg_reverse_test`

I assume this is the test in the onnx test, right? I cannot reproduce the program. Could you provide the onnx file or the program? Thanks

cagery:
> > > > Normalization in slice operator does not work for some cases, such as start = -1, end = -5 for an array of length 5, causing test failures. I am not sure if normalization by default supports such start > end & both negative cases. I am leaning towards enhancing parse_slice.cpp for such cases. Any thoughts? @scxiao @pfultz2
> > > 
> > > 
> > > Can you put your model or the program at some place?
> > 
> > 
> > It is happening with the current test `slice_5arg_reverse_test`
> 
> I assume this is the test in the onnx test, right? I cannot reproduce the program. Could you provide the onnx file or the program? Thanks

Here it is:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/33d6a736226550d522d179e4de623900b5d49e65/test/onnx/slice_5arg_reverse_test.onnx

scxiao:
> > > > > Normalization in slice operator does not work for some cases, such as start = -1, end = -5 for an array of length 5, causing test failures. I am not sure if normalization by default supports such start > end & both negative cases. I am leaning towards enhancing parse_slice.cpp for such cases. Any thoughts? @scxiao @pfultz2
> > > > 
> > > > 
> > > > Can you put your model or the program at some place?
> > > 
> > > 
> > > It is happening with the current test `slice_5arg_reverse_test`
> > 
> > 
> > I assume this is the test in the onnx test, right? I cannot reproduce the program. Could you provide the onnx file or the program? Thanks
> 
> Here it is:
> https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/33d6a736226550d522d179e4de623900b5d49e65/test/onnx/slice_5arg_reverse_test.onnx

I tried the onnx file, that is the program i got:

```
main:@0 = @literal{-1, -3} -> int32_type, {2}, {1}
main:@1 = @literal{-5, -1} -> int32_type, {2}, {1}
main:@2 = @literal{-1, -2} -> int32_type, {2}, {1}
main:@3 = @literal{-1, 1} -> int32_type, {2}, {1}
0 = @param:0 -> float_type, {5, 5}, {5, 1}
main:@4 = slice[axes={-1, -2},starts={-4, -3},ends={2147483647, -1}](0) -> float_type, {2, 4}, {5, 1}
main:@5 = reverse[axes={-1}](main:@4) -> float_type, {2, 4}, {4, 1}
main:@6 = @return(main:@5)
```

any problem with it?

cagery:
Slice should produce start = 1 and end = 5, but now I see that would happen internally in slice implementation. I all along thought the print would be coming with the normalized axes but that's not the case, -4 and INT_MAX will produce [1,5) which is the expected case. Thus it never matched to my test case. Thanks @scxiao ! I am going add slice+reverse+step test case, do cleanup and ask for the final review.

cagery:
- added slice+reverse test
- added slice+step+reverse test
- returning shape for reverse operator with `with_lens()` function

cagery:
This should be ready for another review, can you please take a look? @scxiao @pfultz2

src/include/migraphx/op/reverse.hpp
```diff
@@ -41,7 +41,8 @@ struct reverse
     {
         auto lens = inputs[0].lens();
         auto type = inputs[0].type();
-        return shape{type, lens};
+        shape reverse_shape{type, lens};
+        return reverse_shape.with_lens(type, lens);
```
pfultz2, line 45: This can just be `return inputs[0].with_lens(inputs[0].lens())`. Maybe we could add an method like `make_packed` to `shape` class to do this, but its not necessary for this PR.

src/onnx/parse_slice.cpp
```diff
@@ -97,6 +93,22 @@ struct parse_slice : op_parser<parse_slice>
...
+            std::transform(steps.begin(), steps.end(), std::back_inserter(nsteps), [](auto s) {
+                return std::abs(s);
+            });
+            auto lens      = args[0]->get_shape().lens().size();
+            auto axes_norm = op.axes;
+            for(auto& axis : axes_norm)
+            {
+                while(axis < 0)
+                    axis = axis + lens; // need to be positive for step
+            }
```
pfultz2, line 108: We should normalize the axes in the reverse operator.

pfultz2: via the normalize attribute.

cagery: Do you mean step operator? [Reverse](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/e00479af6dc513d8f83e6da773014dee51ce1c2c/src/include/migraphx/op/reverse.hpp#L36) already normalizes axes. I had to add the line because step needed axis normalized, otherwise complained.

src/include/migraphx/op/reverse.hpp
cagery: Done, thanks!

src/onnx/parse_slice.cpp
pfultz2: Sorry, I meant the step.

cagery: Done

pfultz2 (approved):

src/onnx/parse_slice.cpp
```diff
@@ -97,6 +93,15 @@ struct parse_slice : op_parser<parse_slice>
...
 
         auto ins = info.add_instruction(op, args[0]);
+        if(std::any_of(steps.begin(), steps.end(), [](auto s) { return std::abs(s) != 1; }))
+        {
+            std::vector<int64_t> nsteps;
+            std::transform(steps.begin(), steps.end(), std::back_inserter(nsteps), [](auto s) {
+                return std::abs(s);
+            });
+            ins =
+                info.add_instruction(make_op("step", {{"axes", op.axes}, {"steps", nsteps}}), ins);
```
scxiao, line 103: After checking a unit test in this PR, I think there is a problem here. Specifically,
 
If we need a reverse after the step operator, we need to do some shifts about the ends in the slice operator. 
One simple example: we have input [1, 2, 3,4], and the start is 3 and end is -INF, and we have a step 2. In this case, the output should be [4, 2]. But with our current implementation, the output is [3, 1], which is wrong.

The reason is, for our current implementation, the slice operator has the start to be 0, and end to be 4, then if we do a step of 2, we get 1, and 3, then after the reverse, we get [3, 1]. So when we have both step and reverse, we need to do some shift about the ends so the last element of the slice output is selected by the step operator. 

This issue causes the verification unit test to be incorrect.

test/onnx/verify_onnx.cpp
```diff
@@ -408,6 +408,85 @@ TEST_CASE(selu_test)
...
+                               13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24};
+
+    migraphx::parameter_map pp;
+    pp["0"] = migraphx::argument(sh_data, data.data());
+
+    auto result = p.eval(pp).back();
+    std::vector<float> result_vector;
+    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
+
+    std::vector<float> gold = {13, 11};
```
scxiao, line 486: I think the output of this unit test should be [14, 12].

test/verify/test_slice_reverse.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+struct test_slice_reverse : verify_program<test_slice_reverse>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        migraphx::shape s{migraphx::shape::int32_type, {3, 5}};
+        auto x = mm->add_parameter("x", s);
+        mm->add_literal({{migraphx::shape::int32_type, {2}}, {-1, 1}});
```
scxiao, line 15: This literal is not used, so can be removed.

test/verify/test_slice_step_reverse.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+struct test_slice_step_reverse : verify_program<test_slice_step_reverse>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        migraphx::shape s{migraphx::shape::int32_type, {7, 5}};
+        auto x = mm->add_parameter("x", s);
+        mm->add_literal({{migraphx::shape::int32_type, {2}}, {-2, 2}});
```
scxiao, line 15: This literal can be removed.

src/onnx/parse_slice.cpp
cagery: Makes sense. Changed the operation order in parse_slice.cpp and also related test files. Also removed the unused literals. Thanks a lot @scxiao

test/verify/test_slice_step_reverse.cpp
```diff
@@ -11,8 +11,7 @@ struct test_slice_step_reverse : verify_program<test_slice_step_reverse>
         migraphx::program p;
         auto* mm = p.get_main_module();
         migraphx::shape s{migraphx::shape::int32_type, {7, 5}};
-        auto x = mm->add_parameter("x", s);
-        mm->add_literal({{migraphx::shape::int32_type, {2}}, {-2, 2}});
+        auto x         = mm->add_parameter("x", s);
         auto slice_out = mm->add_instruction(
             migraphx::make_op("slice", {{"axes", {0, 1}}, {"starts", {0, 2}}, {"ends", {2, -1}}}),
             x);
```
scxiao, line 17: Maybe we can add another unit test in which "reverse" is before "step".

cagery: 👍🏼  Done!

scxiao (approved):

PR #870, opened 2021-06-30
Nonzero op extension

scxiao:
support dynamic output shape...


src/targets/gpu/include/migraphx/gpu/hip.hpp
```diff
@@ -144,17 +144,16 @@ struct hip_copy_from_gpu
...
         if(args.size() == 1)
         {
-            argument result = allocate_gpu(output_shape, true);
+            argument result = allocate_gpu(args.front().get_shape(), true);
             gpu_copy(ctx, args[0], result);
             return result;
         }
         copy_from_gpu(ctx, args[0], args[1]);
-        return args[1];
+        return {args[0].get_shape(), args[1].data()};
```
pfultz2, line 156: This should be `args[1]`. This can cause a dangling reference, since the result will not own the memory.

src/targets/gpu/include/migraphx/gpu/hip.hpp
```diff
@@ -144,17 +144,16 @@ struct hip_copy_from_gpu
...
         return inputs.at(0);
     }
-    argument
-    compute(context& ctx, const shape& output_shape, const std::vector<argument>& args) const
+    argument compute(context& ctx, const shape&, const std::vector<argument>& args) const
     {
         if(args.size() == 1)
         {
-            argument result = allocate_gpu(output_shape, true);
+            argument result = allocate_gpu(args.front().get_shape(), true);
```
pfultz2, line 151: `output_shape` should be the same otherwise we have a bug in computing the shape.

src/targets/gpu/include/migraphx/gpu/hip.hpp
```diff
@@ -126,7 +126,7 @@ struct hip_copy_to_gpu
         argument result = args[1].share();
         gpu_copy(ctx, input, result);
         // Associate the input since it was registered with hip
-        return {result.get_shape(), [input, result]() mutable { return result.data(); }};
+        return {args[0].get_shape(), [input, result]() mutable { return result.data(); }};
```
pfultz2, line 129: This should not be necessary.

pfultz2:
This should probably be implemented with prefix_sum. Since we don't have a gpu implementation yet, we can at least insert synchronization in lowering and just use the ref implementation,

pfultz2:
> This should probably be implemented with prefix_sum

Maybe it would be better to have a `filter` operator that takes in a tensor of booleans to filter, but we would still need scan for the GPU.

src/targets/gpu/include/migraphx/gpu/hip.hpp
scxiao: This change is to support the dynamic output shape, which is required by the nonzero operator.
The scenario is: when we call the compute_shape method, we set the output shape to a max possible value, then in the compute method, if the actual output has less elements, i.e., a different shape, we will change the output shape to that from the compute method. 

Then in the gpu_copy, we need to do the same thing. In other words, if number of elems in the input is less than that of the output, we can still do the copy and change the output shape to be the same as the input. 

This change will not hurt any scenario that we can supported for now, and it will support more scenarios like the nonzero and the loop operators. But there are still scenarios, in whcih it will fail  in compute method of some operators.

pfultz2: The compute method should return the same shape as the `compute_shape`.

pfultz2: I would like to assert that this is true, but there are bugs in the `int8_pack_args` that need to be fixed first.

src/onnx/parse_nonzero.cpp
```diff
@@ -31,30 +31,35 @@ struct parse_nonzero : op_parser<parse_nonzero>
...
-            vec_data.assign(val.begin(), val.end());
-            indices = nonzero_indices(vec_data);
-        });
+        if(data_arg.empty())
+        {
+            return info.add_instruction(make_op("nonzero"), args);
+        }
+        else
+        {
+            std::vector<std::size_t> indices;
```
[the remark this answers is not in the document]
scxiao: If we do not do it here, the output shape will have to be the max value, even if it is not the max size. In the pass, we cannot change the max size to its actual value anymore.

src/targets/gpu/device/nonzero.cpp
```diff
@@ -0,0 +1,66 @@
...
+    // only one prefix sum
+    const index_int block_size = 256;
+    arg_idx.visit([&](auto in_idx) {
+        auto* ptr = device_cast(in_idx.data());
+        gs_launch(stream, block_size, block_size)([=](auto, auto idx) __device__ {
+            block_scan<block_size>(idx,
+                                   sum{},
+                                   0,
+                                   elem_num,
+                                   [&](auto j) { return ptr[j]; },
```
pfultz2, line 40: You could do `return float_equal(in_ptr[i], 0) ? 0 : 1` and you can skip the first kernel.

src/targets/gpu/device/nonzero.cpp
```diff
@@ -0,0 +1,66 @@
...
+        auto* idx     = arg_idx.cast<int64_t>();
+        auto* out_ptr = device_cast(output.data());
+        hip_visit_all(arg_data.get_shape())([&](auto si) {
+            gs_launch(stream, elem_num)([=](auto i) __device__ {
+                auto out_idx = idx[i] - 1;
+                auto index   = si.multi(out_idx);
+                for(std::size_t j = 0; j < index.size(); ++j)
+                {
+                    out_ptr[j * elem_num + out_idx] = index[j];
+                }
```
pfultz2, line 55: Can we zero out the unused parts of output here instead of calling `fill`?

src/targets/gpu/device/nonzero.cpp
```diff
@@ -0,0 +1,66 @@
...
+    const index_int block_size = 256;
+    arg_idx.visit([&](auto in_idx) {
+        auto* ptr = device_cast(in_idx.data());
+        gs_launch(stream, block_size, block_size)([=](auto, auto idx) __device__ {
+            block_scan<block_size>(idx,
+                                   sum{},
+                                   0,
+                                   elem_num,
+                                   [&](auto j) { return ptr[j]; },
+                                   [&](auto j, auto x) { ptr[j] = x; });
```
pfultz2, line 41: I wonder if you could write the output here:

```cpp
[&](auto j, auto x) { 
    auto out_idx = x - 1;
    auto index   = si.multi(out_idx);
    for(std::size_t k = 0; k < index.size(); ++k)
    {
        out_ptr[k * elem_num + out_idx] = index[k];
    }
}
```

src/targets/gpu/device/include/migraphx/gpu/device/float_equal.hpp
```diff
@@ -0,0 +1,47 @@
+#ifndef MIGRAPHX_GUARD_RTGLIB_GPU_DEVICE_FLOAT_EQUAL_HPP
+#define MIGRAPHX_GUARD_RTGLIB_GPU_DEVICE_FLOAT_EQUAL_HPP
+
+#include <hip/hip_runtime.h>
+#include <migraphx/half.hpp>
+#include <migraphx/config.hpp>
+#include <migraphx/tensor_view.hpp>
```
pfultz2, line 7: I dont think this include is needed.

src/targets/gpu/device/include/migraphx/gpu/device/float_equal.hpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    template <class T, class U>
+    __device__ __host__ bool operator()(T x, U y) const
+    {
+        return float_equal_fn::apply<common_type<T, U>>(x, y);
+    }
+};
+
+__device__ static constexpr float_equal_fn float_equal{};
```
pfultz2, line 40: How about we just make this a function instead?

src/targets/gpu/device/include/migraphx/gpu/device/float_equal.hpp
```diff
@@ -0,0 +1,47 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+template <class... Ts>
+using common_type = typename std::common_type<Ts...>::type;
+
+struct float_equal_fn
+{
+    template <class T, MIGRAPHX_REQUIRES(is_floating_point<T>{})>
```
pfultz2, line 19: I am not sure if `is_floating_point` will return true for `__fp16`. I think we should just define a `is_floating_point` trait in the types.hpp header. This will avoid the dependency on <migraphx/half.hpp>.

src/include/migraphx/op/nonzero.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
+    {
+        auto elem_num = args.front().get_shape().elements();
+        std::vector<int64_t> vec_idx(elem_num);
+        args.front().visit([&](auto data) {
+            par_for(elem_num, [&](auto i) { vec_idx[i] = (float_equal(data[i], 0)) ? 0 : 1; });
+        });
+
+        std::partial_sum(vec_idx.begin(), vec_idx.end(), vec_idx.begin());
```
pfultz2, line 38: We dont need to use scan/partial_sum for the ref implementation. In fact, it would be better to use the straightforward implementation for the ref implementation so we can make sure our GPU implementation using scan is correct.

src/targets/gpu/device/nonzero.cpp
scxiao: fixed

src/targets/gpu/device/nonzero.cpp
scxiao: fixed

src/targets/gpu/device/nonzero.cpp
scxiao: fixed

scxiao: After additional test, the out_idx in the line auto index   = si.multi(out_idx); should be i, and not every thread can write to the output. Only thread i with idx[i] != idx[i - 1] can write to the output.

src/targets/gpu/device/include/migraphx/gpu/device/float_equal.hpp
scxiao: @pfultz2 I changed it to a function and added a trait in the types.hpp for the __fp16, but looks like it does not work. More specifically, after I changed it to a function, it begin not working. Can you help check what is wrong with that?

src/include/migraphx/op/nonzero.hpp
scxiao: fixed

src/targets/gpu/device/include/migraphx/gpu/device/float_equal.hpp
scxiao: Never mind. I was able to fix it and it works now.

src/targets/gpu/device/include/migraphx/gpu/device/float_equal.hpp
scxiao: changed to a function.

src/targets/gpu/device/include/migraphx/gpu/device/float_equal.hpp
scxiao: removed

src/targets/gpu/device/nonzero.cpp
pfultz2: I dont follow, can you explain some more?

pfultz2: Why cant you use `auto index  = si.multi(j)`?

scxiao: A simple example:
we have input 2, 0, 3, 0, 1 (5 elements), then the output should be
out[0] = 0, out[1] = 2, out[2] = 4.

In our implementation, after the prefix_sum, we got:
1, 1, 2, 2, 3

This means different j's have the same "x", and the different j's are written to the same output location since we use "x" to indicate where to write the "j". 

The approach I used is for each "j" after the prefix_sum, we compare its "x" to that of "j - 1". Only if the x's are different, we write "j" to the output.

pfultz2: We shouldn't need to check the adjacent element, we can just check `not float_equal(ptr[j], 0)`.

scxiao: Good idea, changed. Thanks!

pfultz2 (approved):

PR #872, opened 2021-07-02
Add inclusive scan on the GPU

pfultz2:


scxiao (approved):

scxiao:
Maybe we can remove the reverse operator here. In the case of it require reverse, we can add a reverse operator before the scan, after computing the scan output, do a reverse for the scan output.

turneram (approved):

PR #873, opened 2021-07-06
Build error rocbals flag

scxiao:


kahmed10:
are the backend test changes required?

scxiao:
> are the backend test changes required?

The onnx1.8.1 added these two unit test, but we have no support for them yet. To avoid we meet en error in the future, I just disabled them.

kahmed10 (approved):

PR #874, opened 2021-07-07
Remove Alpha Beta from onnx gemm parsing

umangyadav:
This PR only changes the frontend onnx parser.  Parses does the mutiplication with alpha and beta after doing necessary broadcasting. It then calls to `dot` operator. 

PR doesn't include changes related to following: 

Removal of alpha and beta fields from `dot` operator requires larger changes including in `quant_dot`. 

Removal of `decompose` pass also requires further changes.

### DIFF
--- a/src/onnx/parse_gemm.cpp
+++ b/src/onnx/parse_gemm.cpp
@@ -42,10 +42,7 @@ struct parse_gemm : op_parser<parse_gemm>
         // swap the last two elements
         std::swap(*perm.rbegin(), *(perm.rbegin() + 1));
 
-        auto l1 = (transa) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), args[0])
-                           : args[0];
-        auto l2 = (transb) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), args[1])
-                           : args[1];
+        auto l1 = args[0];
 
         if(alpha != 1.0f)
         {
@@ -54,6 +51,10 @@ struct parse_gemm : op_parser<parse_gemm>
             l1 = info.add_instruction(make_op("convert", {{"target_type", l1->get_shape().type()}}),
                                       alpha_l1);
         }
+
+        l1      = (transa) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), l1) : l1;
+        auto l2 = (transb) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), args[1])
+                           : args[1];
 
         if(args.size() == 3 && beta != 0.0f)
         {
--- a/src/onnx/parse_gemm.cpp
+++ b/src/onnx/parse_gemm.cpp
@@ -56,9 +56,9 @@ struct parse_gemm : op_parser<parse_gemm>
         auto l2 = (transb) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), args[1])
                            : args[1];
 
-        if(args.size() == 3 && beta != 0.0f)
+        if(args.size() == 3)
         {
-            if(beta != 1.0f && args[2]->get_shape().elements() > 0)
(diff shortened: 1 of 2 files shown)


pfultz2 (approved):

src/onnx/parse_gemm.cpp
```diff
@@ -46,9 +46,19 @@ struct parse_gemm : op_parser<parse_gemm>
...
                            : args[1];
+
+        if(alpha != 1.0f)
+        {
+            auto alpha_literal   = info.add_literal(alpha);
+            auto alpha_broadcast = info.add_instruction(
+                make_op("multibroadcast", {{"output_lens", l1->get_shape().lens()}}),
+                alpha_literal);
+            l1 = info.add_instruction(make_op("mul"), l1, alpha_broadcast);
+        }
```
scxiao, line 57: The alpha_broadcast can be of different data type from l1, so we need a check to see if l1 and alpha_broadcast have the same data type. If not, need to add a convert operator. But there is another problem if we do the data type conversion.

According to the ONNX spec, alpha a float, and l1 can be int32, for this case, we need to convert alpha(float) to a int32, which will lose data precision. For example, 0.1 will be converted to 0. 

The same is for the beta.

scxiao: @pfultz2 any opinion on this?

pfultz2: We can convert it to float and do the multiply and then convert back. I think we should just move this common type logic to `add_broadcastable_binary_op`, which should simplify this to:

```cpp
auto multiply = info.add_broadcastable_binary_op("mul", l1, alpha_literal);
l1 = info.add_instruction(make_op("convert", {{"target_type", l1->get_shape().type()}}), multiply);
```

pfultz2: So PR #875 adds such support. So once its merged then `add_broadcastable_binary_op` can be used here.

scxiao: This type conversion is fine. The main problem is that, for example, if l1 is int32, then alpha will be converted from float to int32. This convert will lose information. 
One example is: alpha is 0.5(float), after the type conversion, alpha is converted to 0(int32), which is totally wrong.

pfultz2: > if l1 is int32, then alpha will be converted from float to int32.

With PR #875, it wont convert `alpha` to int32. Instead, `l1` is converted to float, which is why we need this conversion back to int32.

umangyadav: Made changes to make use of `broadcastable_binary_op`

src/onnx/parse_gemm.cpp
```diff
@@ -46,9 +46,19 @@ struct parse_gemm : op_parser<parse_gemm>
...
+
         if(args.size() == 3)
```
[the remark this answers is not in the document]
umangyadav: Done

src/onnx/parse_gemm.cpp
```diff
@@ -46,9 +46,18 @@ struct parse_gemm : op_parser<parse_gemm>
                            : args[0];
         auto l2 = (transb) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), args[1])
                            : args[1];
-        if(args.size() == 3)
+
+        if(alpha != 1.0f)
         {
-            if(beta != 0.f && args[2]->get_shape().elements() > 0)
+            auto alpha_literal = info.add_literal(alpha);
+            auto alpha_l1      = info.add_broadcastable_binary_op("mul", alpha_literal, l1);
```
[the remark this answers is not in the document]
scxiao: We already have a unit test gemm_ex_test to cover the scenario we talked, and I just created an issue to track the issue, so no additional changes are needed.

scxiao (approved):

PR #877, opened 2021-07-08
Topk op

scxiao:


src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,234 @@
...
+        bool larger;
+        template <class T, class U>
+        auto operator()(T x, U y) const
+        {
+            return (larger ? (x < y) : (x > y));
+        }
+    };
+
+    template <class T>
+    void swap(T& v1, T& v2) const
```
pfultz2, line 66: Why not use `std::swap`?

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,234 @@
...
+    {
+        T v = v1;
+        v1  = v2;
+        v2  = v;
+    }
+
+    template <class T, class Op>
+    void heapify(const T& data,
```
pfultz2, line 74: Can the heap be created using the `std::make_heap`, `std::push_heap` and `std::sort_heap` functions instead?

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+    struct compare_op
+    {
+        bool larger;
+        template <class T, class U>
+        auto operator()(T x, U y) const
+        {
+            return (larger ? (x < y) : (x > y));
+        }
+    };
```
pfultz2, line 63: The compare_op should be passed in as a template parameter so we can pass `std::less<>{}` or `std::greater<>{}`. The same would apply to the GPU implementation, as we are doing an extra comparison everytime we compare elements.

src/targets/gpu/device/topk.cpp
```diff
@@ -0,0 +1,230 @@
...
+    auto comp_lens  = in_lens;
+    comp_lens[axis] = 1;
+    shape comp_s{in_s.type(), comp_lens};
+    std::size_t elem_num = comp_s.elements();
+
+    hip_visit_all(val_res, arg, out_s, in_s, comp_s)(
+        [&](auto out_val, auto input, auto oss, auto iss, auto css) {
+            auto* data = device_cast(input.data());
+            auto* out  = device_cast(out_val.data());
+            auto* ind  = reinterpret_cast<int64_t*>(ind_res.data());
```
pfultz2, line 194: Do `ind_res.cast<int64_t>()` instead.

src/include/migraphx/op/topk.hpp
scxiao: fixed

src/include/migraphx/op/topk.hpp
scxiao: fixed in both ref and gpu implementation.

src/targets/gpu/device/topk.cpp
scxiao: fixed

src/include/migraphx/op/topk.hpp
scxiao: since we do not have a strided iterator yet, so I just keep the current implementation.

src/generate.cpp
```diff
@@ -6,32 +6,81 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+            }
+            else
+            {
+                using type = typename decltype(as)::type;
+                auto v     = generate_tensor_data<type>(s, seed);
+                result     = {s, v};
+            }
+        });
+    }
+
```
pfultz2, line 83: Are these changes necessary?

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <migraphx/ranges.hpp>
+#include <migraphx/par_for.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct topk
+{
+    int64_t k;
```
pfultz2, line 18: This should be initialized as well.

src/targets/gpu/device/topk.cpp
```diff
@@ -0,0 +1,225 @@
...
+namespace gpu {
+namespace device {
+
+template <class T>
+__device__ inline void swap(T& v1, T& v2)
+{
+    T v = v1;
+    v1  = v2;
+    v2  = v;
+}
```
pfultz2, line 21: Doesn't `std::swap` work?

src/targets/gpu/device/topk.cpp
```diff
@@ -0,0 +1,225 @@
...
+    build_heap(arr, ind, i, oss, iss, css, n, axis, op);
+    auto idx  = css.multi(i);
+    auto idxj = idx;
+    for(int j = n - 1; j > 0; --j)
+    {
+        idxj[axis] = j;
+        swap(ind[oss.index(idx)], ind[oss.index(idxj)]);
+        heap_heapify(arr, ind, i, oss, iss, css, j, 0, axis, op);
+    }
+}
```
pfultz2, line 149: Is it possible to make a heap class to reduce the number of parameters that need to be passed to these functions?

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,132 @@
...
+        {
+            indices.back() = val;
+        }
+        std::push_heap(indices.begin(), indices.end(), comp);
+    }
+
+    template <class T>
+    void heap_sort(std::vector<int>& indices, T comp) const
+    {
+        std::make_heap(indices.begin(), indices.end(), comp);
```
pfultz2, line 65: Is this necessary? Haven't we already called `make_heap` on `indices`?

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,132 @@
...
+        std::sort_heap(indices.begin(), indices.end(), comp);
+    }
+
+    template <class T>
+    void topk_value(std::vector<int>& indices, std::size_t n, T comp) const
+    {
+        std::make_heap(indices.begin(), indices.end(), comp);
+        for(int i = indices.size(); i < n; ++i)
```
pfultz2, line 73: This could be `for(auto i:range(indices.size(), n))`

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,132 @@
...
+        shape s_ind{shape::int64_type, lens};
+
+        return shape({s_val, s_ind});
+    }
+
+    template <class T>
+    void heap_update(std::vector<int>& indices, const int& val, T comp) const
+    {
+        std::pop_heap(indices.begin(), indices.end(), comp);
+        if(comp(val, indices.back()))
```
pfultz2, line 55: Shouldn't this always be true? Maybe this should be an `assert`?

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+    template <class T>
+    void topk_value(std::vector<int>& indices, std::size_t n, T comp) const
+    {
+        std::make_heap(indices.begin(), indices.end(), comp);
+        for(int i = indices.size(); i < n; ++i)
+        {
+            heap_update(indices, i, comp);
+        }
+    }
```
pfultz2, line 77: I think we could make a heap class.

```cpp
template<class Compare>
struct heap 
{
    std::vector<std::size_t> indices;
    Compare compare;

    heap(std::size_t k, Compare comp)
    : indices(k), compare(std::move(comp))
    {
        std::iota(indices.begin(), indices.end(), 0);
        std::make_heap(indices.begin(), indices.end(), compare);
    }

    void push(std::size_t val)
    {
        std::pop_heap(indices.begin(), indices.end(), compare);
        if(compare(val, indices.back()))
            indices.back() = val;
        std::push_heap(indices.begin(), indices.end(), compare);
    }

    void sort()
    {
        std::sort_heap(indices.begin(), indices.end(), compare);
    }
};

template<class Compare>
heap<Compare> make_heap(std::size_t k, Compare compare)
{
    return {k, std::move(compare)};
}
```

Maybe a similar class could be created for the GPU(but using a different implementation).

src/include/migraphx/op/topk.hpp
scxiao: fixed

src/include/migraphx/op/topk.hpp
scxiao: added

src/include/migraphx/op/topk.hpp
scxiao: @pfultz2 added a gpu heap class, please review.

src/generate.cpp
kahmed10: I think he needs this for his tests to work.

src/targets/gpu/device/include/migraphx/gpu/device/heap.hpp
```diff
@@ -0,0 +1,142 @@
...
+    make_heap(T* ind, index_int n, Index ind_idx, Compare comp)
+    {
+        for(int j = 1; j < n; ++j)
+        {
+            heapify_up(ind, j, ind_idx, comp);
+        }
+    }
+
+    MIGRAPHX_DEVICE_CONSTEXPR inline void
+    push_heap(T* ind, index_int loc, Index ind_idx, Compare comp)
```
pfultz2, line 103: The parameters `ind` and `ind_idx` are laready member variables so they dont need to be passed in.

src/targets/gpu/device/include/migraphx/gpu/device/heap.hpp
```diff
@@ -0,0 +1,142 @@
...
+        data[data_index(size - 1)] = val;
+        push_heap(data, size - 1, data_index, compare);
+    }
+
+    MIGRAPHX_DEVICE_CONSTEXPR void sort() { sort_heap(data, size, data_index, compare); }
+
+    MIGRAPHX_DEVICE_CONSTEXPR T* get_sorted() { return data; }
+
+    private:
+    MIGRAPHX_DEVICE_CONSTEXPR inline void swap(T& v1, T& v2)
```
pfultz2, line 37: This should be static.

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+        std::pop_heap(data.begin(), data.end(), compare);
+        if(compare(val, data.back()))
+            data.back() = val;
+        std::push_heap(data.begin(), data.end(), compare);
+    }
+
+    void sort() { std::sort_heap(data.begin(), data.end(), compare); }
+
+    std::vector<T>& get_sorted() { return data; }
```
pfultz2, line 37: This should probably just be named `get`.

src/targets/gpu/device/topk.cpp
```diff
@@ -0,0 +1,84 @@
...
+namespace gpu {
+namespace device {
+
+argument topk(hipStream_t stream,
+              argument val_res,
+              argument ind_res,
+              argument arg,
+              int64_t k,
+              int64_t axis,
+              bool largest)
```
pfultz2, line 23: This function should be templated with a compare function(only defined in the cpp file), so it doesnt check a boolean on every compare inside the gpu code:

```cpp
template<class Compare>
argument topk(hipStream_t stream,
              argument val_res,
              argument ind_res,
              argument arg,
              int64_t k,
              int64_t axis,
              Compare compare);
```

And then add two device functions that pass in the different compare:

```cpp
argument topk_largest(hipStream_t stream,
                      argument val_res,
                      argument ind_res,
                      argument arg,
                      int64_t k,
                      int64_t axis);
{
    return topk(stream, val_res, ind_res, arg, k, axis, std::less<>{});
}

argument topk_smalles(hipStream_t stream,
                      argument val_res,
                      argument ind_res,
                      argument arg,
                      int64_t k,
                      int64_t axis)
{
    return topk(stream, val_res, ind_res, arg, k, axis, std::greater<>{});
}
```

Then the compute method can check the `largest` flag and call the correct device function.

src/targets/gpu/device/include/migraphx/gpu/device/heap.hpp
scxiao: fixed

src/targets/gpu/device/include/migraphx/gpu/device/heap.hpp
scxiao: fixed

src/targets/gpu/device/topk.cpp
scxiao: fixed

src/include/migraphx/heap.hpp
scxiao: fixed

src/onnx/parse_topk.cpp
```diff
@@ -0,0 +1,58 @@
...
+    std::vector<instruction_ref> parse(const op_desc& /*opd*/,
+                                       const onnx_parser& parser,
+                                       onnx_parser::node_info info,
+                                       std::vector<instruction_ref> args) const
+    {
+        int64_t k = 0;
+        if(args.size() == 2)
+        {
+            auto arg_k = args.at(1)->eval();
+            check_arg_empty(arg_k, "PARSE_TopK: k input must be constant");
```
kahmed10, line 24: while we are still supporting only constant k values, we should open a separate issue to work on optimizations passes for k. For example, if k=1, we can treat as a max operator, and if k=nelements, then return identity.

src/onnx/parse_topk.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+        int64_t axis = -1;
+        if(contains(info.attributes, "axis"))
+        {
+            axis = parser.parse_value(info.attributes.at("axis")).at<int>();
+        }
+
+        auto topk_ret = info.add_instruction(
+            make_op("topk", {{"k", k}, {"axis", axis}, {"largest", largest}}), args.at(0));
+        auto out_s = topk_ret->get_shape();
```
kahmed10, line 46: when will topk not return a tuple type?

src/eliminate_data_type.cpp
```diff
@@ -3,17 +3,19 @@
...
 
 void eliminate_data_type::apply(module& m) const
 {
+    static std::vector<std::string> skip_op_names = {"convert", "get_tuple_elem"};
     for(auto ins : iterator_for(m))
     {
         if(ins->name()[0] == '@')
             continue;
-        if(ins->name() == "convert")
+        if(contains(skip_op_names, ins->name()))
```
pfultz2, line 18: Just do `if(contains({"convert", "get_tuple_elem"}, ins->name()))`.

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+    void sort() { std::sort_heap(data.begin(), data.end(), compare); }
+
+    std::vector<T>& get() { return data; }
+};
+
+template <class T, class Compare>
+heap<T, Compare> make_heap(std::vector<T>& val, Compare compare)
+{
+    return {std::move(val), std::move(compare)};
+}
```
pfultz2, line 44: Can we just put this in the topk header as this is specific to opk? Even better if this can be nested under topk.

src/onnx/parse_topk.cpp
pfultz2: Those are great ideas. Can you open an issue about these optimizations?

src/targets/gpu/hip.cpp
```diff
@@ -101,8 +101,21 @@ hip_ptr write_to_gpu(const T& x)
...
+            auto p = share(allocate_gpu(ss.bytes() + 1, host));
+            return argument{ss, [p]() mutable { return reinterpret_cast<char*>(p.get()); }};
+        });
+        return argument(sub_args);
+    }
+    else
+    {
+        auto p = share(allocate_gpu(s.bytes() + 1, host));
+        return {s, [p]() mutable { return reinterpret_cast<char*>(p.get()); }};
+    }
```
pfultz2, line 118: With #919 this should not be necessary.

src/targets/gpu/device/topk.cpp
```diff
@@ -0,0 +1,104 @@
+#include "migraphx/gpu/device/shape.hpp"
```
pfultz2, line 1: angle brackets for include.

src/onnx/parse_topk.cpp
kahmed10: Actually I'll need to test the expected behavior since it looks like there are multiple outputs for the op.

src/include/migraphx/op/topk.hpp
```diff
@@ -0,0 +1,110 @@
...
+struct topk
+{
+    int64_t k    = 1;
+    int64_t axis = 0;
+    bool largest = true;
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.k, "k"), f(self.axis, "axis"), f(self.largest, "largest"));
```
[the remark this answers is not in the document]
scxiao: The reason is: we call the heap function in the stl lib for the ref implementation, but we cannot do that for the GPU implementation, so we have to implement the heap function ourselves. Because we do not know the specific algorithm used in the stl lib, so if we do not sort the output, there is no guarantee that the ref and gpu implementation have the same outputs, and our GPU tests will fail. To avoid that, we always sort the outputs.

src/onnx/parse_topk.cpp
scxiao: Yes, it also output the indices of the corresponding elements.

src/onnx/parse_topk.cpp
scxiao: That is specified by the onnx spec, it has two outputs, so we have to use the tuple type.

src/include/migraphx/heap.hpp
scxiao: These APIs are the same as that provided by the STL library. I think it is a general heap implementation.

src/eliminate_data_type.cpp
scxiao: fixed

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+        std::push_heap(data.begin(), data.end(), compare);
+    }
+
+    void sort() { std::sort_heap(data.begin(), data.end(), compare); }
+
+    std::vector<T>& get() { return data; }
+};
+
+template <class T, class Compare>
+heap<T, Compare> make_heap(std::vector<T>& val, Compare compare)
```
pfultz2, line 41: Pass by value since you are moving.

src/include/migraphx/heap.hpp
pfultz2: Yea but the `update` method is very specific to topk(as we only push the value onto the heap if it will be a top value).

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+    heap(const std::vector<T>& val, Compare comp) : data(val), compare(std::move(comp))
+    {
+        for(int i = 2; i <= data.size(); ++i)
+        {
+            std::push_heap(data.begin(), data.begin() + i, compare);
+        }
+    }
+
+    void update(const T val)
```
pfultz2, line 27: Dont make the parameter `const` since its passed by value. Also, this should probably be named `try_push`, since it will only push the value if it fits.

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+        std::pop_heap(data.begin(), data.end(), compare);
+        if(compare(val, data.back()))
```
pfultz2, line 30: Shouldnt we do `if(compare(val. data.front())` before calling `pop_heap`?

src/include/migraphx/heap.hpp
pfultz2: Also, `priority_queue` provides a standard API for a heap, but it doesn't provide `sort` or `get` methods.

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <cmath>
+#include <algorithm>
+#include <vector>
+#include <cassert>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+template <class T, class Compare>
+struct heap
```
pfultz2, line 14: This should probably be called `heap_vector` since this only works with `vector`.

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+{
+    std::vector<T> data;
+    Compare compare;
+
+    heap(const std::vector<T>& val, Compare comp) : data(val), compare(std::move(comp))
+    {
+        for(int i = 2; i <= data.size(); ++i)
+        {
+            std::push_heap(data.begin(), data.begin() + i, compare);
```
pfultz2, line 23: Why don't we call `make_heap` here?

src/include/migraphx/heap.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+    void update(const T val)
+    {
+        std::pop_heap(data.begin(), data.end(), compare);
+        if(compare(val, data.back()))
+            data.back() = val;
+        std::push_heap(data.begin(), data.end(), compare);
+    }
+
+    void sort() { std::sort_heap(data.begin(), data.end(), compare); }
```
pfultz2, line 35: If `sort` is called then this is no longer a heap. This should return a copy of the vector sorted, otherwise this breaks the class invariant.

src/onnx/parse_topk.cpp
pfultz2: We can return multiple outputs when parsing onnx, so we dont always need to use a tuple.

kahmed10: If we don't always need to use a tuple, then I'm not sure why we need the assert on L47.

src/include/migraphx/heap.hpp
scxiao: No need to do that, the if(compare(val, data.back())) is the same as what you mention before pop_back().

src/include/migraphx/heap.hpp
scxiao: fixed

src/include/migraphx/heap.hpp
scxiao: fixed

src/include/migraphx/heap.hpp
scxiao: Previous, I had tried to make the unsorted output to be the same between the ref and gpu, by using the push_heap, we know specific content value in the heap, so we can get the same heap between ref and gpu. That is why I use the push_heap to push element one by one. If we always sort the out, yes, we can use make_heap.

src/include/migraphx/heap.hpp
scxiao: fixed.

scxiao: @pfultz2 We need to make the same above change for the gpu implementation, right?

pfultz2: I dont think we can make a copy on the GPU. We can move it to the topk.cpp file so it will only be used with topk(it could be confusing for other operators).

src/include/migraphx/heap.hpp
pfultz2: Alright, then lets use `make_heap` as I think that will be faster.

src/include/migraphx/heap.hpp
pfultz2: Yes, but it requires swapping the values, and then pushing it back on the heap. We can avoid popping and pushing to the heap if we know the value wont be added.

src/onnx/parse_topk.cpp
scxiao: The problem is, when we compute the output shape of this operator, we do not know whether one or two outputs are needed, so we have to provide two outputs, and the output shape is a tuple. This is from the compute_shape() method, then in the compute() method, based on the output_shape computed from the compute_shape() method, we return an argument with the tuple_type, this is required by our design since we cannot change the output shape computed from the compute_shape() method.

I know for the RNN Operators, we have multiple outputs, but we did not use the tuple type, the reason for that is, we compute a first output, which contains information for the second output, then we insert another operator (rnn_last_output) to retrieve the second output from the first output. This is a special case so we can avoid the tuple type. 

If we understand in another way, for topk, we get the tuple output argument, then we insert the get_tuple_elem operator, this is very similar as what we did for the RNN operator. With our current design and architecture, I do not know how to avoid the tuple type. Of course, we can come up with anther design to not use the tuple.

src/include/migraphx/heap.hpp
scxiao: move the comparison to the start of the function.

src/include/migraphx/heap.hpp
scxiao: move the gpu heap to the topk.cpp file.

src/targets/gpu/hip.cpp
scxiao: removed this change

pfultz2 (approved):

src/onnx/parse_topk.cpp
kahmed10: I was mainly bringing up that the assert would always pass? Is it necessary?

src/targets/gpu/include/migraphx/gpu/device/topk.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <migraphx/argument.hpp>
+#include <migraphx/config.hpp>
+#include <hip/hip_runtime_api.h>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+argument topk_smallest(hipStream_t stream,
```
kahmed10, line 13: Wouldn't it be better to pass in `largest` (from `op.largest`) here and consolidate into one `topk()` function? Then in topk.cpp, you can just use the ternary operator on `largest` and call topk with either `std::less` or `std::greater`.

scxiao: This current approach is what Paul proposed, with this approach, we can avoid checking the op.largest every time the compare operator is used.

kahmed10 (approved):

turneram (approved):

src/onnx/parse_topk.cpp
scxiao: I just removed it.

PR #878, opened 2021-07-11
Move AMDGPU_TARGETS to before including dependencies

yjwong:
This is similar to https://github.com/ROCmSoftwarePlatform/rocSPARSE/pull/213. Since AMDGPU_TARGETS is marked as CACHED, and it's also included in rocBLAS, we can never override the value of AMDGPU_TARGETS. This means MIGraphX cannot be compiled for older GPU like gfx803.

Resolves https://github.com/RadeonOpenCompute/ROCm/issues/1269.

### DIFF
--- a/src/targets/gpu/CMakeLists.txt
+++ b/src/targets/gpu/CMakeLists.txt
@@ -1,3 +1,4 @@
+set(AMDGPU_TARGETS "gfx803;gfx900;gfx906" CACHE STRING "")
 
 list(APPEND CMAKE_PREFIX_PATH /opt/rocm /opt/rocm/hip /opt/rocm/hcc)
 find_package(miopen)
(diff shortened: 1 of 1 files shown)


pfultz2:
This variable is only meant for hcc(and should probably be removed). When using hip-clang, we use the default targets defined hip-config.cmake.

Also, the variable to set the targets should be `GPU_TARGETS`(as documented [here](https://rocmdocs.amd.com/en/latest/Installation_Guide/Using-CMake-with-AMD-ROCm.html#using-hip-in-cmake)). `AMDGPU_TARGETS` does work for backwards-compatible reasons, but is undocumented and may not work in the future.

yjwong:
Ah, that makes sense, thanks for the pointer to `hip-config.cmake`. I encountered this as I was trying to build onnxruntime with the MIGraphX execution provider.

It seems there are a lot of existing projects in the ecosystem not using the right way documented in the link above, to the extent of hardcoding `--amdgpu-target` with `HIP_CLANG_FLAGS` (which I understand seems to be a deprecated parameter since it's used by HCC).

I will try to submit PRs to the upstream projects and close this PR then. Thanks!

pfultz2:
> I encountered this as I was trying to build onnxruntime with the MIGraphX execution provider.

Until changes are made upstream, you can override this variable when you invoke cmake the first time when building MIGraphX from source(ie `cmake -DGPU_TARGETS=gfx803;gfx900;gfx906 ..`) and then it will only build for the targets you specify.

PR #882, opened 2021-07-14
[Example] U-Net Image Segmentation

cagery:
Provides simple image segmentation example with U-Net.


examples/python_unet/unet_inference.ipynb
```diff
@@ -0,0 +1,294 @@
...
+      "\n",
+      "unet_13_256.onnx    100%[===================>]  65.85M  25.8MB/s    in 2.5s    \n",
+      "\n",
+      "2021-07-21 20:44:45 (25.8 MB/s) - ‘unet_13_256.onnx’ saved [69052820/69052820]\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "!wget -nc https://github.com/cagery/unet-onnx/raw/main/unet_13_256.onnx"
```
turneram, line 63: The model downloaded here causes the kernel to restart when I try to parse it. Trying to read the model from the driver fails with the only output being `Killed`. Cagri was unable to reproduce the error; curious to see if you encounter the same issue or not @kahmed10 .

cagery: The only difference I can think of is that GPUs are different. 
On ROCm versions 4.1 and 4.2, I still do not see the error. 
I, although, see `MIOpen(HIP): Error [BaseGetSolution] result.out_pix_tile1 == 0 || result.out_pix_tile0 == 0` but this does not cause any failures. Maybe this error causes different behavior on your GPU and causing it to kill?

kahmed10: I have ROCm 4.2, not seeing the error either.

turneram (approved):

kahmed10:
Can we get a requirements.txt file for this example?

examples/python_unet/unet_inference.ipynb
cagery: I have provided another unet onnx file that uses transposed convolutions instead of bilinear to @turneram, he did not see the issue with transposed conv. Might be worth looking into it as a separate issue for why bilinear upsampling is causing issues in his system.

cagery:
> Can we get a requirements.txt file for this example?

Done

examples/python_unet/README.md
```diff
@@ -0,0 +1,9 @@
+# U-Net Image Segmentation Inference with AMD MIGraphX
+
+This examples provides a simple example for utilizing U-Net ONNX model for image segmentation, using AMD MIGraphX graph optimization engine for fast inference.
+
+## How-to
+Please utilize the notebook given `unet_inference.ipynb`.
+
+## Model Details
+ONNX model utilized in this example can be found [here](https://github.com/cagery/unet-onnx).
```
kahmed10, line 9: this should be updated with new model link.

examples/python_unet/unet_inference.ipynb
```diff
@@ -0,0 +1,213 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9de6f2a7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "mask = model.run({'inputs':input_im}) # Your first inference would take longer than the following ones.\n",
+    "output_mask = np.array(mask[0])\n",
+    "output_mask.shape"
```
kahmed10, line 158: do we want to wrap this with a print() to be consistent with other cells?

examples/python_unet/README.md
cagery: done thanks!

kahmed10 (approved):

PR #889, opened 2021-07-20
Optimize Q/DQ Format Pass

turneram:
Adds a pass to handle ONNXRuntime fake quantization format. The optimization pass consists of three sub-passes that 1) add Q->DQ pairs before add ops that are inserted by the onnx parser, 2) replaces `DQ->Quantizable_Op->Q` with `Quantized_Op` (quant_convolution, quant_dot), and 3) removes all remaining Q->DQ instances.


src/targets/gpu/target.cpp
```diff
@@ -60,6 +61,7 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         normalize_ops{},
         decompose{},
         dead_code_elimination{},
+        optimize_qdq_format{},
```
pfultz2, line 64: Can we name this `simplify_qdq`? That will be consistent in naming with the other simplify passes we have.

test/optimize_qdq_format_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include <migraphx/dead_code_elimination.hpp>
+#include <migraphx/pass_manager.hpp>
+#include <migraphx/matcher.hpp>
+
+#include <migraphx/serialize.hpp>
+
+#include <migraphx/verify.hpp>
+
+bool is_convolution(migraphx::instruction& ins) { return ins.name() == "convolution"; }
+bool is_dot(migraphx::instruction& ins) { return ins.name() == "dot"; }
```
pfultz2, line 20: This can be passed as constant.

test/optimize_qdq_format_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+struct match_find_qdq
+{
+    bool match_found = false;
+
+    auto matcher() const
+    {
+        return migraphx::match::name("dequantizelinear")(
+            migraphx::match::arg(0)(migraphx::match::name("quantizelinear")));
+    }
+    void apply(migraphx::module&, const migraphx::match::matcher_result&) { match_found = true; }
```
pfultz2, line 31: The `apply` method should be `const`.

test/optimize_qdq_format_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+    migraphx::program p1 = create_program();
+    migraphx::program p2 = create_program();
+
+    migraphx::optimize_qdq_format opt;
+    opt.apply(*p2.get_main_module());
+
+    match_find_qdq m1;
+    match_find_qdq m2;
+    migraphx::match::find_matches(*p1.get_main_module(), m1);
+    migraphx::match::find_matches(*p2.get_main_module(), m2);
```
pfultz2, line 189: In the matcher.cpp test file there is a `find_match` function which should work better than trying to using the match finder.  This will let you pass the matcher directly:

```cpp
auto match_qdq = migraphx::match::name("dequantizelinear")(
                 migraphx::match::arg(0)(migraphx::match::name("quantizelinear")));
auto ins = find_match(mod, match_qdq);
```

You can copy this function here or just add it to matcher.hpp file so it can be reused across multiple files.

test/optimize_qdq_format_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+                                           {0.00260497},
+                                           {0.0235534},
+                                           {0.0235534},
+                                           {0.023622},
+                                           {0.0162265},
+                                           {0.0138147}};
+    std::vector<std::vector<int>> ints{{-35}, {-127}, {-18}, {0}};
+
+    auto create_program = [&]() {
+        migraphx::program p;
```
pfultz2, line 56: Dont create a program. Just create the module.

src/optimize_qdq_format.cpp
```diff
@@ -0,0 +1,169 @@
...
+            if(arg->name() == "dequantizelinear")
+            {
+                auto q = arg->inputs().front();
+                if(q->name() == "quantizelinear")
+                {
+                    arg        = q->inputs().front();
+                    replace_op = true;
+                }
+            }
+        }
```
pfultz2, line 150: Use `std::find_if` here instead of the for loop.

src/optimize_qdq_format.cpp
```diff
@@ -0,0 +1,169 @@
...
+                    mismatch = true;
+                }
+            }
+        }
+
+        if(replace and not mismatch)
+        {
+            if(ins->name() == "convolution")
+            {
+                auto conv_op = any_cast<op::convolution>(ins->get_operator());
```
pfultz2, line 104: Can you use `make_op` to construct the convolution operator?

src/optimize_qdq_format.cpp
```diff
@@ -0,0 +1,169 @@
...
+                {
+                    arg     = q1;
+                    replace = true;
+                }
+                else
+                {
+                    mismatch = true;
+                }
+            }
+        }
```
pfultz2, line 98: The for loop can be rewritten with `std::all_of`.

pfultz2:
It would be good to add some smaller unit tests. Like just a module with DQ/Conv/Q and then check the module matches the expected transformation. I would like to see unit tests with something like:

- DQ/Conv/Q
- DQ/Dot/Q
- DQ/BiasAdd/Q
- DQ/Conv/Q -> DQ/BiasAdd/Q
- DQ/Dot/Q -> DQ/Add/Q
- DQ/Conv/Q -> DQ/Dot/Q

src/optimize_qdq_format.cpp
turneram: I think we need to be able to iterate over all inputs in case multiple input tensors have QDQs. Using find_if would stop after the first input with QDQ.

scxiao:
When I try to compile the model "mobilenetv2-7.quant.onnx" and compile it on GPU, I met a error as:

```
projects/AMDMIGraphX/src/include/migraphx/check_shapes.hpp:115: same_type: add: Types do not match
```

src/include/migraphx/matcher.hpp
```diff
@@ -263,6 +263,20 @@ matcher_result match_instruction(module& mod, instruction_ref ins, M&& m)
     return result;
 }
 
+/// Find first instance of a matching instruction in a module
+template <class M>
+migraphx::match::matcher_result find_match(migraphx::module& modl, M&& m)
```
pfultz2, line 268: No need for qualified names such as `migraphx::match` to `migraphx` here.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,484 @@
...
+    migraphx::module m1;
+    {
+        auto pa  = m1.add_parameter("pa", s7);
+        auto p3  = m1.add_parameter("p3", s4);
+        auto l4  = m1.add_literal(s1, ints[1]);
+        auto l5  = m1.add_literal(s1, ints[2]);
+        auto l7  = m1.add_literal(s1, ints[3]);
+        auto l14 = m1.add_literal(s8, floats[5]);
+        auto l15 = m1.add_literal(s8, floats[6]);
+        auto l16 = m1.add_literal(s8, floats[7]);
```
pfultz2, line 158: Can these be named more clearly? Are these for scales or shifts? Secondly, it would be good to put the numbers directly instead of from the vectors. Finally, you should use values that are simpler to calculate such as 0.5 or 2.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+        auto mb1 = m1.add_instruction(
+            migraphx::make_op("multibroadcast", {{"output_lens", {1280, 320, 1, 1}}}), l16);
+        auto mb2 = m1.add_instruction(
+            migraphx::make_op("multibroadcast", {{"output_lens", {1280, 320, 1, 1}}}), l7);
+        auto d1  = m1.add_instruction(migraphx::make_op("dequantizelinear"), p3, mb1, mb2);
```
pfultz2, line 164: I think it would be good to have a function to add these operators to make the test a little more concise:

```cpp
instruction_ref add_quantize_op(module& m, const std::string& name, instruction_ref x, instruction_ref scale, instruction_ref shift)
{
    auto lens = x->get_shape().lens();
    auto scale_mb = m.add_instruction(make_op("multibroadcast", {{"output_lens", lens}}), scale);
    auto shift_mb = m.add_instruction(make_op("multibroadcast", {{"output_lens", lens}}), shift);
    return m.add_instruction(make_op(name), x, scale_mb, shift_mb);
}
```

Then you can add the op with `auto d1 = add_quantize_op("dequantizelinear"), p3, scale1, shift1);` and dont have to worry about inserting the multibroadcasts.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,484 @@
...
+                                              {0.0162265},
+                                              {0.0138147}};
+static std::vector<std::vector<int>> ints{{-35}, {-127}, {-18}, {0}};
+
+TEST_CASE(conv)
+{
+    migraphx::module m1;
+    {
+        auto pa  = m1.add_parameter("pa", s7);
+        auto p3  = m1.add_parameter("p3", s4);
```
pfultz2, line 152: Can we use better names for the parameters? There is a `p3` and `pa` but no `p1` or `p2`. It would be better name it `input` and `weights`.

test/simplify_qdq_test.cpp
pfultz2: Also the variable should be named something like `dq_p3`(or `dq_input`), so it easier to see how it was transformed.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,484 @@
...
+        auto mb15 = m2.add_instruction(
+            migraphx::make_op("multibroadcast", {{"output_lens", {1, 1280, 7, 7}}}), l14);
+        auto mb16 = m2.add_instruction(
+            migraphx::make_op("multibroadcast", {{"output_lens", {1, 1280, 7, 7}}}), l4);
+        auto d6 = m2.add_instruction(migraphx::make_op("dequantizelinear"), c1, mb15, mb16);
+        m2.add_return({d6});
+    }
+
+    migraphx::simplify_qdq sqdq;
+    sqdq.apply(m1);
```
pfultz2, line 227: Most of the time we add a `run_pass` function in the test file to do this:

```cpp
void run_pass(migraphx::module& m)
{
    migraphx::simplify_qdq sqdq;
    sqdq.apply(m);
}
```

So we have a central place to update how the pass is applied.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+    static std::unordered_set<std::string> s = {"convolution", "dot"};
+    return s;
+}
+
+struct match_find_add_bias
+{
+    auto matcher() const
+    {
+        return match::name("dequantizelinear")(match::arg(0)(
```
scxiao, line 28: This assumes there is always an "add" after "convolution", and after the "add" there is a QDQ pair. This is not true. (Maybe the mobilenetv2 is the case), for example, in the Alexnet (I shared you the model), one scenario is like: 

```
main:@128 = multibroadcast[output_lens={1, 64, 27, 27}](main:@63) -> float_type, {1, 64, 27, 27}, {0, 0, 0, 0}
main:@129 = multibroadcast[output_lens={1, 64, 27, 27}](main:@64) -> int8_type, {1, 64, 27, 27}, {0, 0, 0, 0}
main:@130 = quantizelinear(main:@127,main:@128,main:@129) -> int8_type, {1, 64, 27, 27}, {46656, 729, 27, 1}
main:@131 = multibroadcast[output_lens={1, 64, 27, 27}](main:@63) -> float_type, {1, 64, 27, 27}, {0, 0, 0, 0}
main:@132 = multibroadcast[output_lens={1, 64, 27, 27}](main:@64) -> int8_type, {1, 64, 27, 27}, {0, 0, 0, 0}
main:@133 = dequantizelinear(main:@130,main:@131,main:@132) -> float_type, {1, 64, 27, 27}, {46656, 729, 27, 1}
main:@134 = convolution[padding={2, 2, 2, 2},stride={1, 1},dilation={1, 1},group=1,padding_mode=0](main:@133,main:@102) -> float_type, {1, 192, 27, 27}, {139968, 729, 27, 1}
main:@135 = broadcast[axis=1,dims={1, 192, 27, 27}](main:@105) -> float_type, {1, 192, 27, 27}, {0, 1, 0, 0}
main:@136 = add(main:@134,main:@135) -> float_type, {1, 192, 27, 27}, {139968, 729, 27, 1}
main:@137 = pooling[mode=max,padding={0, 0, 0, 0},stride={2, 2},lengths={3, 3},ceil_mode=0](main:@136) -> float_type, {1, 192, 13, 13}, {32448, 169, 13, 1}
```

After the convolution, there is no QDQ.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,156 @@
...
+        m.replace_instruction(add, migraphx::make_op("add"), add_args);
+    }
+};
+
+struct match_find_quantizable_ops
+{
+    auto matcher() const
+    {
+        return match::name("dequantizelinear")(match::arg(0)(match::name("quantizelinear")(
+            match::arg(0)(match::name(get_quantizable_op_names()).bind("qop")))));
```
scxiao, line 70: This is also assumes there are QDQ after the operator, which is not true. According to our previous discussion, and what ORT mentioned before, we can only assume there is QDQ for each argument of the operators to be quantized. 

Because of this reason, for the alexnet model I shared with you, there are two convolution ops that are not converted to quant_convolution. You can run the model for a try.

scxiao (approved):

scxiao:
> When I try to compile the model "mobilenetv2-7.quant.onnx" and compile it on GPU, I met a error as:
> 
> ```
> projects/AMDMIGraphX/src/include/migraphx/check_shapes.hpp:115: same_type: add: Types do not match
> ```
> 
> Note that: in the program, before compiling on the GPU, I called the quantize_fp16 for the program.

Maybe you can ignore this problem and try to resolve the issues I mentioned in other comments first, then we can come back to check this.

scxiao (changes requested):
I clicked the approve by mistake.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+            auto alpha     = 1;
+            auto beta      = (qop->inputs().size() == 3) ? dot_op.beta : 0.0f;
+            qop            = m.insert_instruction(
+                qop, migraphx::make_op("quant_dot", {{"alpha", alpha}, {"beta", beta}}), qop_args);
+            dq_scale   = m.add_literal(static_cast<float>(scale_val));
+            zero_point = m.add_literal(static_cast<int>((-1.0f * beta) / scale_val));
+        }
+        else
+        {
+            MIGRAPHX_THROW(qop->name() + " does not currently have a quantized implementation.");
```
pfultz2, line 153: We shouldn't throw an exception. We should only apply the transformation for dot and convolution.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+    auto shift_mb =
+        m.insert_instruction(ins, make_op("multibroadcast", {{"output_lens", lens}}), shift);
+    return m.insert_instruction(ins, make_op(name), x, scale_mb, shift_mb);
+}
+
+struct match_find_quantizable_ops
+{
+    auto matcher() const
+    {
+        return match::name(get_all_op_names())(match::arg(0)(
```
pfultz2, line 87: You can use `match::any()` instead of `match::name(get_all_op_names())`.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,537 @@
...
+        auto t2 = m1.add_parameter("t2", sh2);
+        auto sc = m1.add_literal(ss, scale);
+        auto z  = m1.add_literal(zs, zero);
+
+        auto q1 = add_quantize_op(m1, "quantizelinear", t1, sc, z);
+        auto d1 = add_quantize_op(m1, "dequantizelinear", q1, sc, z);
+        auto q2 = add_quantize_op(m1, "quantizelinear", t2, sc, z);
+        auto d2 = add_quantize_op(m1, "dequantizelinear", q2, sc, z);
+        auto dot =
+            m1.add_instruction(migraphx::make_op("dot", {{"alpha", 1}, {"beta", 0}}), d1, d2);
```
pfultz2, line 56: Shouldn't there be a `quantizelinear` after the dot operator?

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+                qop, migraphx::make_op("quant_dot", {{"alpha", alpha}, {"beta", beta}}), qop_args);
+            dq_scale   = m.add_literal(static_cast<float>(scale_val));
+            zero_point = m.add_literal(static_cast<int>((-1.0f * beta) / scale_val));
+        }
+        else
+        {
+            MIGRAPHX_THROW(qop->name() + " does not currently have a quantized implementation.");
+        }
+
+        dq              = insert_quantize_op(m, op, "dequantizelinear", qop, dq_scale, zero_point);
```
pfultz2, line 156: We shouldn't be inserting a `dequantizelinear` as that should already have been inserted by the model.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+    std::vector<int64_t> zero_points;
+    zp_lit.visit([&](const auto zp) {
+        std::transform(
+            zp.begin(), zp.end(), std::back_inserter(zero_points), [](auto&& z) { return z; });
+    });
+    return std::all_of(
+        zero_points.begin(), zero_points.end(), [](const auto& z) { return z == 0; });
+}
+
+double get_scale(const instruction_ref& ins)
```
pfultz2, line 48: This should be done with the matcher. You can use `skip_broadcasts` to skip the broadcasts and then bind the instruction.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+    std::vector<float> scales;
+    scale_lit.visit([&](auto sl) {
+        std::transform(
+            sl.begin(), sl.end(), std::back_inserter(scales), [](auto&& s) { return s; });
+    });
+    double epsilon = 1e-6;
+    if(not std::all_of(scales.begin(), scales.end(), [&](const auto& s) {
+           return std::abs(s - scales.front()) < epsilon;
+       }))
+        MIGRAPHX_THROW("Multiple scales not currently supported");
```
pfultz2, line 64: We should check in the matcher that it uses only one scale. You can create custom matcher with `MIGRAPHX_PRED_MATCHER`. 

Also since this is something we dont support we should have multiple tests for this.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+        if(q0->get_shape().type() != migraphx::shape::int8_type or
+           q1->get_shape().type() != migraphx::shape::int8_type)
+            return;
+
+        // Only zero_point==0 currently supported
+        auto dq0_args  = dq0->inputs();
+        auto dq1_args  = dq1->inputs();
+        bool all_zeros = true;
+        all_zeros &= inputs_are_zeros(dq0_args.at(2));
+        all_zeros &= inputs_are_zeros(dq1_args.at(2));
```
pfultz2, line 115: You can use the `has_value` matcher to check if the values are zeros.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    auto matcher() const
+    {
+        return match::name(get_all_op_names())(match::arg(0)(
+            match::name(get_quantizable_op_names())(
+                match::arg(0)(match::name("dequantizelinear")(
+                    match::arg(0)(match::name(get_all_op_names()).bind("q0")).bind("dq0"))),
+                match::arg(1)(match::name("dequantizelinear")(
+                    match::arg(0)(match::name(get_all_op_names()).bind("q1")).bind("dq1"))))
+                .bind("qop")));
```
pfultz2, line 93: This matcher doesnt check for `quantizelinear`. I would write this matcher like this:

```cpp
auto dequantizelinear_op(const std::string& name, const std::string& scale)
{
    return match::name("quantizelinear")(
        match::arg(0)(match::any().bind(name)),
        match::arg(1)(match::skip_broadcasts(match::any().bind(scale))),
        match::arg(2)(match::skip_broadcasts(match::none_of(match::has_value(0))))
    );
}

auto matcher() const
{
    match::name("quantizelinear")(match::args(
        match::name(get_quantizable_op_names())(
            match::arg(0)(dequantizelinear_op("x1", "scale1")),
            match::arg(1)(dequantizelinear_op("x2", "scale2"))
        )
    ));
}
```

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,537 @@
...
+                                                       {{"padding", {0, 0, 0, 0}},
+                                                        {"stride", {1, 1}},
+                                                        {"dilation", {1, 1}},
+                                                        {"group", 1},
+                                                        {"padding_mode", 0}}),
+                                     d5,
+                                     d1);
+        auto bc1 = mm.add_instruction(
+            migraphx::make_op("broadcast", {{"axis", 1}, {"dims", {1, 1280, 7, 7}}}), d2);
+        auto a1 = mm.add_instruction(migraphx::make_op("add"), c1, bc1);
```
pfultz2, line 493: This is missing a q/dq pair for the convolution output before it goes into `add`.

pfultz2:
Also it would be good to have a unit test with just and an add with qdq pairs: q/dq -> add -> dq/q

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+        // Only INT8 type currently supported
+        if(q0->get_shape().type() != migraphx::shape::int8_type or
+           q1->get_shape().type() != migraphx::shape::int8_type)
+            return;
+
+        // Only zero_point==0 currently supported
+        auto dq0_args  = dq0->inputs();
+        auto dq1_args  = dq1->inputs();
+        bool all_zeros = true;
+        all_zeros &= inputs_are_zeros(dq0_args.at(2));
```
scxiao, line 114: Should this be a logical and instead of a bitwise one?

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+        auto scale     = get_scale(dq0_args.at(1)) * get_scale(dq1_args.at(1));
+        auto qop_args  = qop->inputs();
+        qop_args.at(0) = q0;
+        qop_args.at(1) = q1;
+        instruction_ref dq;
+        instruction_ref dq_scale;
+        instruction_ref zero_point;
+        if(qop->name() == "convolution")
+        {
+            auto conv_op = any_cast<op::convolution>(qop->get_operator());
```
scxiao, line 128: This can be done in a simpler way as:

```
auto conv_val = qop->get_operator().to_value();
qop          = m.insert_instruction(qop, migraphx::make_op("quant_convolution", conv_val), qop_args);
```

src/simplify_qdq.cpp
scxiao: There is no dequantizelinear added after the operator. We confirmed this from MSFT guys before and the quantized model from ORT does not have it, either.

test/simplify_qdq_test.cpp
scxiao: No, there is no quantizelinear after the dot operator

src/simplify_qdq.cpp
pfultz2: I see, sorry for the confusion then.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+    else
+        scale_lit = ins->get_literal();
+    std::vector<float> scales;
+    scale_lit.visit([&](auto sl) {
+        std::transform(
+            sl.begin(), sl.end(), std::back_inserter(scales), [](auto&& s) { return s; });
+    });
+    double epsilon = 1e-6;
+    if(not std::all_of(scales.begin(), scales.end(), [&](const auto& s) {
+           return std::abs(s - scales.front()) < epsilon;
```
pfultz2, line 62: We have `float_equal` that can be used.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+        scale_lit = ins->inputs().at(0)->get_literal();
+    else
+        scale_lit = ins->get_literal();
+    std::vector<float> scales;
+    scale_lit.visit([&](auto sl) {
+        std::transform(
+            sl.begin(), sl.end(), std::back_inserter(scales), [](auto&& s) { return s; });
+    });
+    double epsilon = 1e-6;
+    if(not std::all_of(scales.begin(), scales.end(), [&](const auto& s) {
```
pfultz2, line 61: This should be skip the first element: `std::all_of(scales.begin()+1, scales.end(), ...)`.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    literal scale_lit;
+    if(ins->name() == "multibroadcast" or ins->name() == "broadcast")
+        scale_lit = ins->inputs().at(0)->get_literal();
+    else
+        scale_lit = ins->get_literal();
+    std::vector<float> scales;
+    scale_lit.visit([&](auto sl) {
+        std::transform(
+            sl.begin(), sl.end(), std::back_inserter(scales), [](auto&& s) { return s; });
```
pfultz2, line 58: Actually the copy to a vector is not needed, You can just do `all_of` inside the visit:

```cpp
    scale_lit.visit([&](auto s) {
        all_same = std::all_of(s.begin()+1, s.end(), [&](const auto& scale) {
            return float_equal(scale, s.front());
        });
    });
```

src/simplify_qdq.cpp
pfultz2: I guess the above is wrong since I was assuming a quantizelinear on the output, but we will need to skip quantizelinear ops:

```cpp
MIGRAPHX_PRED_MATCHER(has_same_value, instruction_ref ins) 
{
    if (!ins->name() != "@literal")
        return false;
    bool all_same = false;
    ins->get_literal().visit([&](auto s) {
        all_same = std::all_of(s.begin()+1, s.end(), [&](const auto& scale) {
            return float_equal(scale, s.front());
        });
    });
    return all_same;
}

auto dequantizelinear_op(const std::string& name, const std::string& scale)
{
    return match::name("dequantizelinear")(
        match::arg(0)(match::skip(match::name("quantizelinear"))(mathc::any().bind(name))),
        match::arg(1)(match::skip_broadcasts(has_same_value().bind(scale))),
        match::arg(2)(match::none_of(match::has_value(0)))
    );
}

auto matcher() const
{
    match::name(get_quantizable_op_names())(
        match::arg(0)(dequantizelinear_op("x1", "scale1")),
        match::arg(1)(dequantizelinear_op("x2", "scale2"))
    );
}
```

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,143 @@
...
+            auto conv_val = qop->get_operator().to_value();
+            dq            = m.insert_instruction(
+                qop, migraphx::make_op("quant_convolution", conv_val), qop_args);
+            dq_scale   = m.add_literal(static_cast<float>(scale));
+            zero_point = m.add_literal(0);
+        }
+        else if(qop->name() == "dot")
+        {
+            auto dot_op    = any_cast<op::dot>(qop->get_operator());
+            auto scale_val = dot_op.alpha / scale;
```
scxiao, line 90: This should be scale_val = dot_op.alpha * scale.

scxiao: The scale in my implementation is the reciprocal of the scale here, so we should use mulitply.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,143 @@
...
+            auto alpha     = 1;
+            auto beta      = (qop->inputs().size() == 3) ? dot_op.beta : 0.0f;
```
scxiao, line 92: the beta here is implicitly converted to an integer in the quant_dot operator, so if beta is not an integer, it will caused error. 
We should always set beta to zero, and use beta * C as the zeropoint argument in the dequantizelinear operator.

scxiao: 1. qop_args should do pop_back to remove the C from the input. Otherwise, there is will an error since C is float type.
2. Zero point should be (-1.0f * beta / scale_val) * C.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,143 @@
...
+            zero_point = m.add_literal(static_cast<int>((-1.0f * beta) / scale_val));
+        }
+
+        auto lens = dq->get_shape().lens();
+        auto scale_mb =
+            m.insert_instruction(qop, make_op("multibroadcast", {{"output_lens", lens}}), dq_scale);
+        auto shift_mb = m.insert_instruction(
+            qop, make_op("multibroadcast", {{"output_lens", lens}}), zero_point);
+        dq = m.insert_instruction(qop, make_op("dequantizelinear"), dq, scale_mb, shift_mb);
+        m.replace_instruction(qop, dq);
```
scxiao, line 105: I would suggest create a unit test as (pseudo code):

```

a = m.add_parameter("a", sa);
b = m.add_parameter("b", sb);
c = m.add_parameter("c", sc);
sa = 0.33f;
sb = 0.72f;
qa = m.add_instruction(make_op("quantizelinear"), a, sa);
dqa = m.add_instruction(make_op("quantizelinear"), qa, sa); 
qb = m.add_instruction(make_op("quantizelinear"), b, sb);
dqb = m.add_instruction(make_op("quantizelinear"), qb, sb); 
auto d = m.add_instruction(make_op("dot", {{"alpha", 0.3}, {"beta", 0.5}}, dqa, dqb, c)
m.add_return({d});
```

For the above program:
case 1) replace dot with quant_dot, create an input, run the program, get result1
case 2), remove the qdq for both a and b, create the same input as in case 1), run the program, get result 2

result1 should be equal to result2

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,143 @@
...
+    for(auto ins : iterator_for(m))
+    {
+        bool replace_op = false;
+        auto args       = ins->inputs();
+        for(auto&& arg : args)
+        {
+            if(arg->name() == "dequantizelinear")
+            {
+                auto q = arg->inputs().front();
+                if(q->name() == "quantizelinear")
```
scxiao, line 120: We need to check to ensure q and dq have the same scale and zero point to remove them.
Also, please add a unit test for this scenario.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,636 @@
...
+                                     weights);
+        auto d6 = add_quantize_op(m2, "dequantizelinear", c1, sc1, z1);
+        m2.add_return({d6});
+    }
+
+    run_pass(m1);
+    EXPECT(m1 == m2);
+}
+
+TEST_CASE(conv_multi_scale)
```
scxiao, line 318: What is the reason that the convolution is not replaced by quant_conv?

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,636 @@
...
+    auto match_qdq = migraphx::match::name("dequantizelinear")(
+        migraphx::match::arg(0)(migraphx::match::name("quantizelinear")));
+    auto ins1 = migraphx::match::find_match(mod1, match_qdq);
+    auto ins2 = migraphx::match::find_match(mod2, match_qdq);
+
+    EXPECT((ins1.result != mod1.end()) and (ins2.result == mod2.end()));
+    EXPECT(any_of(mod1, &is_convolution));
+    EXPECT(none_of(mod2, &is_convolution));
+    EXPECT(any_of(mod1, &is_dot));
+    EXPECT(none_of(mod2, &is_dot));
```
scxiao, line 633: Please add the unit test as I mentioned above. Also create a similar unit test for the convolution.

scxiao:
@turneram I tried the alexnet model, looks like none of the convolution or dot is replaced by quantized version, so please test the real model to ensure it works correctly.

src/simplify_qdq.cpp
pfultz2: We should porbably ignore beta and alpha values, since we are planning to remove them. If alpha is not 1 and beta not 0 then it should just bail.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,143 @@
...
+        {
+            auto conv_val = qop->get_operator().to_value();
+            dq            = m.insert_instruction(
+                qop, migraphx::make_op("quant_convolution", conv_val), qop_args);
+            dq_scale   = m.add_literal(static_cast<float>(scale));
+            zero_point = m.add_literal(0);
+        }
+        else if(qop->name() == "dot")
+        {
+            auto dot_op    = any_cast<op::dot>(qop->get_operator());
```
pfultz2, line 89: Can we create `dot` with `make_op`?

src/targets/cpu/target.cpp
```diff
@@ -47,6 +48,7 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
     std::set<shape::type_t> unsupported_types(shape::types().begin(), shape::types().end());
     unsupported_types.erase(shape::type_t::float_type);
     return {normalize_ops{},
+            simplify_qdq{},
```
pfultz2, line 51: We dont need this for CPU since we dont support the quantize ops yet.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,636 @@
...
+    migraphx::shape s7{migraphx::shape::float_type, {1, 320, 7, 7}};
+    migraphx::shape s8{migraphx::shape::float_type, {320}};
+
+    std::vector<int> zero{0};
+
+    migraphx::module m1;
+    {
+        auto input   = m1.add_parameter("input", s7);
+        auto weights = m1.add_parameter("weights", s4);
+        auto l1      = m1.add_literal(s1, zero);
```
pfultz2, line 331: Do `auto zero = m1.add_literal(std::int8_t{0})` instead.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,636 @@
...
+        auto l2      = m1.add_literal(migraphx::generate_literal(s8, 0));
```
pfultz2, line 332: Name these literals as `zero` and `scale`.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,636 @@
...
+    std::vector<float> scale{0.5};
+    std::vector<int> zero{0};
+
+    migraphx::module m1;
+    {
+        auto input   = m1.add_parameter("input", s7);
+        auto weights = m1.add_parameter("weights", s4);
+        auto bias    = m1.add_parameter("bias", s6);
+        auto l1      = m1.add_literal(s1, zero);
+        auto l2      = m1.add_literal(s8, scale);
```
pfultz2, line 388: You should be able to create the literal as `auto scale = m1.add_literal(0.5f)`.

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,636 @@
...
+TEST_CASE(conv_bias_add)
+{
+    migraphx::shape s1{migraphx::shape::int8_type, {1}};
+    migraphx::shape s4{migraphx::shape::int8_type, {1280, 320, 1, 1}};
+    migraphx::shape s6{migraphx::shape::int32_type, {1280}};
+    migraphx::shape s7{migraphx::shape::float_type, {1, 320, 7, 7}};
+    migraphx::shape s8{migraphx::shape::float_type, {1}};
+
+    std::vector<float> scale{0.5};
+    std::vector<int> zero{0};
```
pfultz2, line 380: There should be no need for these extra vector classes as you can construct literal scalars with `add_literal(value)`.

src/simplify_qdq.cpp
pfultz2: > I would suggest create a unit test as (pseudo code):

I would prefer not to add a test case like that with `dot` having non-default alpha and beta since we are in the process of removing these.

src/simplify_qdq.cpp
pfultz2: We should bail if alpha is not 1.

turneram:
> @turneram I tried the alexnet model, looks like none of the convolution or dot is replaced by quantized version, so please test the real model to ensure it works correctly.

It looks like the problem is that the convolution weights have non-zero zero points and the B inputs for the dots do as well. Mobilenet seems to also suffer from this. Do we plan on adding support for non-zero shifts?

src/simplify_qdq.cpp
turneram: Should I still add a similar test case, but with default alpha and beta?

pfultz2:
> It looks like the problem is that the convolution weights have non-zero zero points and the B inputs for the dots do as well. Mobilenet seems to also suffer from this. Do we plan on adding support for non-zero shifts?

I dont think we need to worry about that for now. The math for non-zero shifts is a little more complicated, so lets just focus on zero shifts for now.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,166 @@
...
+        return false;
+    auto literal2 = ins2->get_literal();
+
+    bool are_equal = false;
+    visit_all(literal1, literal2)([&](const auto l1, const auto l2) {
+        are_equal = std::equal(l1.begin(), l1.end(), l2.begin(), float_equal);
+    });
+
+    return are_equal;
+}
```
pfultz2, line 128: You can do this:

```cpp
bool compare_literals(instruction_ref ins1, instruction_ref ins2)
{
    auto x = ins1.eval();
    if (x.empty())
        return false;
    auto y = ins2.eval();
    if (y.empty())
        return false;
    return x == y;
}
```

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,748 @@
...
+        auto scale2 = m2.add_literal(0.4f);
+        auto zero   = m2.add_literal(std::int8_t{0});
+
+        auto q1  = add_quantize_op(m2, "quantizelinear", t1, scale1, zero);
+        auto d1  = add_quantize_op(m2, "dequantizelinear", q1, scale2, zero);
+        auto q2  = add_quantize_op(m2, "quantizelinear", t2, scale1, zero);
+        auto d2  = add_quantize_op(m2, "dequantizelinear", q2, scale2, zero);
+        auto add = m2.add_instruction(migraphx::make_op("add"), d1, d2);
+        m2.add_return({add});
+    }
```
pfultz2, line 110: If `m2` is the same as `m1` then just do `migraphx::module m2 = m1;`

src/simplify_qdq.cpp
scxiao: Then we need to change the parse_gemm to ensure alpha and beta are always 1.0f and 0.0f, respectively.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+void remove_qdq_pairs(module& m)
```
scxiao, line 130: This function can be simplified as:

```
void remove_qdq_pairs(module& m)
{
    {
        auto args       = ins->inputs();
        for(auto&& arg : args)
        {
            if(arg->name() == "dequantizelinear")
            {
                auto q = arg->inputs().front();
                if((q->name() == "quantizelinear") and
                   compare_literals(arg->inputs().at(1), q->inputs().at(1)) and
                   compare_literals(arg->inputs().at(2), q->inputs().at(2)))
                {
                    instruction::replace_argument(ins, arg, q->inputs().front());
                }
            }
        }
    }
}
```
so it does not require all inputs have qdq's to remove the qdq pairs.

src/simplify_qdq.cpp
```diff
@@ -0,0 +1,166 @@
...
+        return false;
+    auto literal1 = ins1->get_literal();
+
+    if(ins2->name() == "broadcast" or ins2->name() == "multibroadcast")
+        ins2 = ins2->inputs().front();
+    if(!ins2->can_eval())
+        return false;
+    auto literal2 = ins2->get_literal();
+
+    bool are_equal = false;
```
scxiao, line 122: There is another scenario that: if literal1 and literal2 are of different shapes, but all their values are the same, this function will return false, while it should return true. So the comparison can be written as:

```
    bool are_equal = false;
    visit_all(literal1, literal2)([&](const auto l1, const auto l2) {
        bool equal1 = std::equal(l1.begin(), l1.end(), l2.begin(), float_equal);
        auto val = *l1.begin();
        bool equal2 = std::all_of(l1.begin() + 1, l1.end(), [&](auto v) { return float_equal(v, val); }) and 
             std::all_of(l2.begin(), l2.end(), [&](auto v) { return float_equal(v, val); });
        are_equal = equal1 or equal2;
    });
```

test/simplify_qdq_test.cpp
```diff
@@ -0,0 +1,748 @@
...
+
+        run_pass(*m1);
+    }
+
+    migraphx::program p2;
+    {
+        auto* m2 = p2.get_main_module();
+        auto a   = m2->add_parameter("a", sh1);
+        auto b   = m2->add_parameter("b", sh2);
+        auto c   = m2->add_parameter("c", sh3);
```
scxiao, line 723: since beta is 0, so we do not need 'c' anymore.

scxiao: In the file parse_gemm.cpp, there is one change needed to set beta always to be 0. The last few lines should be changed to:

```
        l1      = (transa) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), l1) : l1;
        auto l2 = (transb) ? info.add_instruction(make_op("transpose", {{"dims", perm}}), args[1])
                           : args[1];

        auto ret = info.add_instruction(make_op("dot", {{"alpha", 1.0f}, {"beta", 0.0f}}), l1, l2);

        if(args.size() == 3)
        {
            if(not float_equal(beta, 0.0f) && args[2]->get_shape().elements() > 0)
            {
                auto out_lens   = l1->get_shape().lens();
                out_lens.back() = l2->get_shape().lens().back();
                auto l3         = args[2];
                auto l3_lens    = l3->get_shape().lens();
                if(!std::equal(out_lens.begin(), out_lens.end(), l3_lens.begin(), l3_lens.end()))
                {
                    l3 = info.add_instruction(
                        make_op("multibroadcast", {{"output_lens", out_lens}}), args[2]);
                }
                auto beta_literal = info.add_literal(beta);
                auto beta_l3      = info.add_broadcastable_binary_op("mul", l3, beta_literal);
                if(beta_l3->get_shape().type() != dot_type)
                {
                    beta_l3 = info.add_instruction(make_op("convert", {{"target_type", dot_type}}),
                                                   beta_l3);
                }

                ret = info.add_instruction(make_op("add"), ret, beta_l3);
            }
        }
```

Of course, the related to gemm onnx_test should be changed accordingly.

scxiao (approved):

pfultz2 (approved):

PR #891, opened 2021-07-21
qdq for quantization and include subgraph

scxiao:


src/quantization.cpp
```diff
@@ -1,3 +1,6 @@
+#include "migraphx/float_equal.hpp"
+#include "migraphx/instruction_ref.hpp"
+#include "migraphx/literal.hpp"
```
pfultz2, line 3: Use angle brackets for the headers.

src/quantization.cpp
```diff
@@ -16,6 +19,7 @@
 #include <migraphx/stringutils.hpp>
 #include <migraphx/ranges.hpp>
 #include <migraphx/target.hpp>
+#include <new>
```
pfultz2, line 22: Is this header needed?

src/quantization.cpp
```diff
@@ -48,58 +60,20 @@ instruction_ref insert_quant_ins(module& modl,
...
-            insert_loc, make_op("multibroadcast", {{"output_lens", rounded_lens}}), min_clip);
-        auto clipped_ins =
-            modl.insert_instruction(insert_loc, make_op("clip"), rounded_ins, min_clip, max_clip);
+        auto zero_point = modl.add_literal(static_cast<int8_t>(shift));
+        auto ins_scale  = modl.add_literal(1.0f / scale);
+        auto lens       = ins->get_shape().lens();
+        ins_scale       = modl.insert_instruction(
+            insert_loc, make_op("multibroadcast", {{"output_lens", lens}}), ins_scale);
+        zero_point = modl.insert_instruction(
+            insert_loc, make_op("multibroadcast", {{"output_lens", lens}}), zero_point);
```
pfultz2, line 74: The attributes need to be renamed `out_lens` now.

src/quantization.cpp
```diff
@@ -112,80 +86,153 @@ instruction_ref insert_quant_ins(module& modl,
...
-    std::unordered_map<instruction_ref, instruction_ref> map_fp16;
-    for(auto ins : iterator_for(*mm))
+    auto inputs = ins->inputs();
+    for(auto in : inputs)
+    {
+        if(in->get_shape().type() == shape::half_type)
+        {
+            continue;
+        }
+        else if(in->get_shape().type() == shape::float_type or
```
pfultz2, line 98: No need for `else` after `continue`.

src/quantization.cpp
```diff
@@ -112,80 +86,153 @@ instruction_ref insert_quant_ins(module& modl,
...
+            assert(in->name() == "convert" and
+                   in->inputs().front()->get_shape().type() == shape::half_type);
+            instruction::replace_argument(ins, in, in->inputs().front());
+        }
+    }
+}
+
+void quantize_fp16(module& m,
+                   const std::vector<std::string>& ins_names,
+                   std::unordered_map<instruction_ref, instruction_ref>& map_fp16,
```
pfultz2, line 110: Why are we passing in this map?

src/quantization.cpp
```diff
@@ -112,80 +86,153 @@ instruction_ref insert_quant_ins(module& modl,
...
+// This function is to convert any instructions specified in the input
+// from double or float to float16 by inserting a convert operator.
+// For the conversion, there could be cases of overflowing, but it
+// is very rare in the area of deeping learning, so we just do a
+// truncate of the input to get the fp16.
+void quantize_fp16(program& prog, const std::vector<std::string>& ins_names)
+{
+    auto* mm = prog.get_main_module();
+    std::unordered_map<instruction_ref, instruction_ref> map_fp16;
+    quantize_fp16(*mm, ins_names, map_fp16, false);
```
pfultz2, line 233: This seems backwards. Shouldn't we start at the leaf modules first and then do the main module last?

pfultz2: Actually, `run_passes` already does this. Can we convert this to a pass and then pass it to `run_passes`?

```cpp
void quantize_fp16(program& prog, const std::vector<std::string>& ins_names)
{
    run_passes(prog, {quantize_fp16_pass{ins_names}});
}
```

src/quantization.cpp
```diff
@@ -489,12 +492,11 @@ void quantize_int8(program& prog,
 
 // For the input of each input argument, we need to insert a
 // capture operator to compute the scale and shift
-std::size_t capture_arguments(program& prog,
-                              const std::vector<std::string>& ins_names,
-                              const std::function<void(std::size_t, std::vector<argument>)>& func)
+void capture_arguments(module& m,
+                       const std::vector<std::string>& ins_names,
+                       const std::function<void(std::size_t, std::vector<argument>)>& func,
+                       std::size_t& num_quant_params)
```
pfultz2, line 498: Whats the purpose of this? Wont we know this based on the value of the last capture op?

pfultz2, line 498: This should be returned by the function.

src/include/migraphx/op/capture.hpp
```diff
@@ -29,7 +30,7 @@ struct capture
 
     shape compute_shape(std::vector<shape> inputs) const { return inputs.front(); }
 
-    argument compute(const shape&, std::vector<argument> args) const
+    argument compute(context&, const shape&, const std::vector<argument>& args) const
```
pfultz2, line 33: Why is this changed?

pfultz2:
For int8 we could use the pass manager as well. We would have a pass to insert capture ops, run the model, and then run passes to quantize based on quantized parameters.

src/include/migraphx/quantize_fp16.hpp
```diff
@@ -0,0 +1,26 @@
...
+struct program;
+
+/**
+ * quantize a program to fp16
+ */
+struct quantize_fp16_pass
+{
+    std::vector<std::string> ins_names = {"all"};
+    std::string name() const { return "quantize_fp16"; }
+    void apply(program& prog) const;
```
pfultz2, line 20: Why does this take the `program` and not the `module`?

src/quantize_fp16.cpp
```diff
@@ -0,0 +1,165 @@
...
+                auto it      = std::find_if(in_outs.begin(), in_outs.end(), [](auto o) {
+                    return (o->name() == "convert" and o->get_shape().type() == shape::half_type);
+                });
+                assert(it != in_outs.end());
+                input_fp16 = *it;
+                converted_inputs.push_back(input_fp16);
+            }
+            // if the input is a convert operator, uses its input
+            // as its current input
+            else if(input->name() == "convert" and
```
pfultz2, line 104: There is no need to check if the input is a `convert`, it should always insert a convert and then simplify_reshapes can remove the redundant ones.

src/quantize_fp16.cpp
```diff
@@ -0,0 +1,165 @@
...
+        if(ins->name() == "@return")
+        {
+            if(quantize_inout)
+            {
+                convert_outputs_fp16(ins);
+            }
+            break;
+        }
+
+        if(ins->name() == "@param" or ins->name() == "@literal")
```
pfultz2, line 52: We should just skip on this. We should not be inserting a `convert`.

src/quantize_fp16.cpp
```diff
@@ -0,0 +1,165 @@
...
+            {
+                // check the dead code case to avoid assert
+                auto ins_orig_shape = m.insert_instruction(
+                    std::next(ins), make_op("convert", {{"target_type", orig_shape.type()}}), ins);
+                m.replace_instruction(ins, ins_orig_shape);
+            }
+        }
+        m.replace_instruction(ins, op, converted_inputs, mod_inputs);
+    }
+}
```
pfultz2, line 150: This just seems more complicated then it needs to be. We already have passes that can remove redundant convert ops(and if its not sufficient with modules then we can look at expanding the optimization passes). So the quantize can just focus on inserting the ops needed to quantize a single op. Basically, insert a convert back to the original type, and then insert converts to fp16 for each of the inputs that are floats:

```cpp
{
    if (contains({"@return", "convert"}, ins->name()))
        continue;
    if (ins->intputs().empty())
        continue;

    auto r = m.insert_instruction(std::next(ins), make_op("convert", {{"target_type", s.type()}}));
    m.replace_instruction(ins, r);

    auto inputs = ins->inputs();
    std::transform(inputs.begin(), inputs.end(), inputs.begin(), [&](auto input) {
        auto s = input->get_shape();
        if(s.type() != shape::float_type and s.type() != shape::double_type)
            return input;
        return m.insert_instruction(ins, make_op("convert", {{"target_type", shape::half_type}}), input);
    });
    m.replace_instruction(ins, ins->get_operator(), inputs);
}
```

We always skip `@return`, `convert`(regardless of the target type), and operators with no inputs.

pfultz2: This can be done using just a module pass. And there is no need to traverse modules or keep track of instruction in the parent modules.

src/quantize_int8.cpp
```diff
@@ -0,0 +1,300 @@
...
+        // it to an integer as the new_alpha in the quant_dot
+        instruction_ref input_c{};
+        if(converted_inputs.size() == 3)
+        {
+            input_c = converted_inputs.back();
+            converted_inputs.pop_back();
+        }
+
+        auto quant_dot = modl.insert_instruction(
+            ins, make_op("quant_dot", {{"alpha", 1}, {"beta", 0}}), converted_inputs);
```
pfultz2, line 49: Why is this inserting a `quant_dot`? This should just insert the qdq pairs and then we already have a later pass to rewrite it to `quant_dot`.

src/quantize_fp16.cpp
scxiao: This is true for the main module, but for submodules, I think we need to convert "return" and "params" to fp16.

src/quantize_int8.cpp
scxiao: If we do what you suggested, then the ref target will never be able to run the quant_dot since ref target does not rewrite it to the quant_dot. Is that what we want?

pfultz2: Then we should probably add the qdq transform to the ref. This is because verifying with int8 could be off in the driver since we dont run calibration data.

src/quantize_fp16.cpp
pfultz2: We should skip quantizing ops that have submodules for now(we can still quantize the submodule) as I think there is more work to make it work correctly. 

If we change an input arg to be fp16 and the corresponding parameter is still fp32 then we have created an invalid program. We ultimately need a way to change the arguments and then update the corresponding parameters in the submodules together, but let's not tackle this for now.

src/include/migraphx/quantize_fp16.hpp
scxiao: changed to module

src/quantize_fp16.cpp
scxiao: fixed

src/quantize_int8.cpp
scxiao: I just added this pass in the quantization_int8() api implementation, so we will get the int8 version of the ops to be quantized.

src/simplify_reshapes.cpp
```diff
@@ -176,6 +176,23 @@ struct find_nested_convert
...
+        auto ins     = mr.result;
+        auto x       = ins->inputs().front();
+        auto outputs = x->outputs();
+        if(x->get_shape() == ins->get_shape())
+        {
+            m.replace_instruction(ins, x);
+            m.move_instruction(ins, m.end());
+        }
+    }
+};
```
pfultz2: The `find_nop_reshapes` should already handle this case.

src/include/migraphx/quantize_int8.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * capture inputs of operators to be quantized to int8
+ */
+struct capture_arguments_pass
+{
+    std::vector<std::string> ins_names = {"dot", "convolution"};
+    std::function<void(std::size_t, std::vector<argument>)> f{};
+    std::string name() const { return "capture_arguments"; }
+    void apply(program& prog) const;
```
pfultz2, line 23: This could be a module pass as well, and we can just add a pointer `std::size_t* num_params` to this class to track the number of captures globally.

src/quantize_int8.cpp
```diff
@@ -0,0 +1,196 @@
...
+        std::vector<instruction_ref> dq_inputs;
+        for(auto input : inputs)
+        {
+            // calculate the index of each instruction to be quantized
+            std::size_t ins_index =
+                (map_ins_index.count(input) > 0) ? map_ins_index[input] : quant_param_index++;
+            map_ins_index[input] = ins_index;
+
+            auto param = quant_params[map_ins_index[input]];
+            ins_quant_params.push_back(param);
```
pfultz2, line 72: It would be simpler to keep the capture args so you can track which quant_param to use.

src/include/migraphx/quantize_int8.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * quantize a program to int8
+ */
+struct quantize_int8_pass
+{
+    std::vector<std::string> ins_names = {"dot", "convolution"};
+    std::vector<std::pair<float, float>> quant_params;
+    std::string name() const { return "quantize_int8"; }
+    void apply(program& prog) const;
```
pfultz2, line 34: This should be a module pass as well.

src/simplify_reshapes.cpp
scxiao: removed

src/quantization.cpp
```diff
@@ -568,12 +105,50 @@ capture_arguments_impl(program& prog, const target& t, const std::vector<std::st
...
+    // insert capture operator
+    const auto& int8_quant_params = calc_quantize_params(prog, t, calibration, ins_names);
+    run_passes(prog,
+               {quantize_int8_pass{ins_names, int8_quant_params},
+                eliminate_common_subexpression{},
+                dead_code_elimination{},
+                simplify_reshapes{},
+                dead_code_elimination{},
+                simplify_qdq{},
+                dead_code_elimination{}});
```
pfultz2, line 151: Why do we pass all these passes? We should run through these passes when we compile the program.

scxiao: This is to keep the quantize_int8 doing the same thing as the previous implementation. In other words, after we call this API, we get a int8 version of the program, and it can run on either gpu, cpu, or ref. With the additional passes, we are still running the program in fp32 if we not have the later passes on some target.

pfultz2: But now a target doesn't have control if it wants to skip running a dot or conv as a quantized dot or conv. For example, the cpu backend doesn't run anything as int8 currently.

pfultz2: We can open an issue to address this in a future PR. This doesn't really change the current behavior.

pfultz2 (approved):

src/quantization.cpp
scxiao: Thanks. #943 is created to track this.

turneram (approved):

PR #910, opened 2021-08-16
Exclude param from deadcode elimiation

scxiao:


test/gpu/quantization.cpp
```diff
@@ -57,6 +57,7 @@ TEST_CASE(int8_quantization)
         }
 
         auto result = t.copy_from(p.eval(m).back());
+        std::cout << "result = " << result << std::endl;
```
kahmed10, line 60: do we need this change?

scxiao: fixed

src/dead_code_elimination.cpp
```diff
@@ -60,7 +60,10 @@ void dead_code_elimination::apply(module& m) const
                 leaf->clear_arguments();
                 assert(bidistance(m, last, leaf) < 0);
                 assert(leaf != ins);
-                m.move_instruction(leaf, m.end());
+                if(leaf->name() != "@param" or
+                   (leaf->name() == "@param" and
+                    any_cast<builtin::param>(leaf->get_operator()).parameter == "scratch"))
+                    m.move_instruction(leaf, m.end());
```
pfultz2, line 66: Can you add a test to dead_code_elimintation_test.cpp?

src/dead_code_elimination.cpp
```diff
@@ -60,7 +60,10 @@ void dead_code_elimination::apply(module& m) const
                 leaf->clear_arguments();
                 assert(bidistance(m, last, leaf) < 0);
                 assert(leaf != ins);
-                m.move_instruction(leaf, m.end());
+                if(leaf->name() != "@param" or
+                   (leaf->name() == "@param" and
+                    any_cast<builtin::param>(leaf->get_operator()).parameter == "scratch"))
```
pfultz2, line 65: Why are we checking the name?

scxiao: After some investigation, the problem is, in the "proallocate_param" pass, the scratch memory is replaced by the preallocate_param instruction as:

```
        std::string id = m.name() + ":" + param;
        auto r         = m.insert_instruction(ins, model.preallocate(ins->get_shape(), id));
        m.replace_instruction(ins, r);
```
So, we do not need the scratch parameter anymore. But with the change in the dead code elimination, we cannot remove the scratch parameter, then when we run a program on the GPU, there is a scratch parameter that is not provided, so an error is reported.

I can change this pass to remove the scratch param right after the replace_instruction as:

```
void preallocate_param::apply(module& m) const
{
    auto last = std::prev(m.end());
    {
        if(ins->name() != "@param")
            continue;
        if(param != any_cast<builtin::param>(ins->get_operator()).parameter)
            continue;
        std::string id = m.name() + ":" + param;
        auto r         = m.insert_instruction(ins, model.preallocate(ins->get_shape(), id));
        m.replace_instruction(ins, r);
        m.move_instruction(ins, m.end());
    }
    m.remove_instructions(std::next(last), m.end());
}
```

Then we do not need to check the parameter name in the dead code elimination anymore.

src/dead_code_elimination.cpp
scxiao: It is covered in other tests, but anyway, I add one test.

src/dead_code_elimination.cpp
pfultz2: > I can change this pass to remove the scratch param right after the replace_instruction

I think thats a better change.

src/schedule.cpp
```diff
@@ -510,6 +522,9 @@ void schedule::apply(module& p) const
     if(enabled(MIGRAPHX_TRACE_COMPILE{}) or enabled(MIGRAPHX_TRACE_SCHEDULE{}))
     {
         p.annotate(std::cout, [&](auto ins) {
+            if(ins->name() == "@param" and not contains(si.weights, ins))
```
pfultz2, line 525: Condecov says there is no coverage for this line. Is this a false positive?

test/simplify_reshapes_test.cpp
```diff
@@ -707,20 +707,21 @@ TEST_CASE(optimize_where_true)
         return m;
     };
 
-    auto create_opt_module = [&](std::string name) {
+    auto create_opt_module = [&](bool cond) {
```
pfultz2, line 710: Maybe name this `return_x` or something?

src/schedule.cpp
scxiao: To cover this line, we need one of the env vars (MIGRAPHX_TRACE_COMPILE and MIGRAPHX_TRACE_SCHEDULE) to be set. Do we have a way to set them?

test/simplify_reshapes_test.cpp
scxiao: renamed

pfultz2 (approved):

scxiao:
@kahmed10 When you get a chance, could you please take a look at this PR? Thanks!

src/schedule.cpp
```diff
@@ -239,6 +239,18 @@ struct stream_info
...
+        // move dangling parameter to the front so as not be removed
+        auto ins = std::next(last);
+        while(ins != p.end())
+        {
+            auto next = std::next(ins);
+            if(ins->name() == "@param")
+            {
+                p.move_instruction(ins, p.begin());
+            }
+            ins = next;
```
kahmed10, line 252: why can't we remove `next` and just do `inst = std::next(ins)`?

scxiao: After we do the move_instruction, ins is the first element in the list, which is invalid for the while loop.

kahmed10 (approved):

PR #912, opened 2021-08-16
unary scalar input processing

scxiao:
Unary returns a non-scalar shape for input scalar input shape, which does not make sense.


src/shape.cpp
```diff
@@ -258,6 +258,10 @@ bool shape::transposed() const
 bool shape::broadcasted() const
 {
     assert(this->lens().size() == this->strides().size());
+    // a scalar is not considered as broadcasted.
+    if(this->lens().size() == 1 and this->lens()[0] == 1 and this->strides()[0] == 0)
```
pfultz2, line 262: Scalar should be considered broadcast.

pfultz2:
What is the reason for making this change?

scxiao:
> What is the reason for making this change?

The unary ops changes a scalar input to a shape as {1}, {1}, this causes a shape mismatch of the cond parameter between the input/output of the submodule for the loop operator.

src/shape.cpp
scxiao: OK, I removed this change.

pfultz2 (approved):

PR #916, opened 2021-08-20
Change attributes names to be more consistent and reflect better meaning

### LINKED ISSUE #786: refine attribute name in a few operators
We have a few operators "broadcast", "reshape", "multibroadcast" that have an output dims attribute, but each of them use a different name. We can change to use the same name "out_dims" or "out_lens" for all of them. 

The "transpose" operator has an attribute permutation, but we use a name "dims", we can change it to the name "perm" to better reflect the meaning...

umangyadav:
Fixes #786 

Changes 

broadcast operator's attribute name "dims" to "out_lens"
multibroadcast operators' attribute name "output_lens" to "out_lens"
~~reshape op's attribute name "dims" to "out_lens"~~
transpose op's attribute name "dims" to ~~"perm"~~ to "permutation"

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -96,7 +96,7 @@ instruction_ref insert_common_op(module& m,
         if(input->get_shape().lens() != common.lens())
         {
             input = m.insert_instruction(
-                ins, make_op("multibroadcast", {{"output_lens", common.lens()}}), input);
+                ins, make_op("multibroadcast", {{"out_lens", common.lens()}}), input);
         }
         if(input->get_shape().type() != common.type())
         {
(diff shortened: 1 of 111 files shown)


pfultz2:
> transpose op's attribute name "dims" to "perm"

I think it would be better to use `permutation` instead of just `perm`.

umangyadav:
> > transpose op's attribute name "dims" to "perm"
> 
> I think it would be better to use `permutation` instead of just `perm`.

Done. Changed it to "permutation"

scxiao (approved):

pfultz2:
> reshape op's attribute name "dims" to "out_lens"

I dont think this should be renamed. Dims is different than the out_lens, since you can have `-1` or `0`.

umangyadav:
> > reshape op's attribute name "dims" to "out_lens"
> 
> I dont think this should be renamed. Dims is different than the out_lens, since you can have `-1` or `0`.

Agree.  I can drop that commit.

pfultz2 (approved):

PR #919, opened 2021-08-23
Enable constructing argument with tuple and buffer

pfultz2:


scxiao (commented):
I am wondering whether we need to consider the alignment? For example, the tuple has sub shapes with size 255, 2, 3, should we allocate more buffer than the sum, then starting address of every sub argument can be aligned?

scxiao (approved):

kahmed10 (approved):

scxiao:
@pfultz2 Looks like it needs a unit test with both shape and data to construct an argument.

PR #918, opened 2021-08-20
Refactor where op

turneram:

PR #921, opened 2021-08-24
Add ThresholdedRelu to ONNX parser

turneram:
Depends on PR #918


src/onnx/parse_thresholdedrelu.cpp
```diff
@@ -0,0 +1,40 @@
...
+struct parse_thresholdedrelu : op_parser<parse_thresholdedrelu>
+{
+    std::vector<op_desc> operators() const { return {{"ThresholdedRelu"}}; }
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& parser,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        double alpha;
```
scxiao, line 20: I think alpha is float, you can do this as:

```
        float alpha = 1.0f;
        if(contains(info.attributes, "alpha"))
            alpha = parser.parse_value(info.attributes.at("alpha")).at<float>();
```

test/onnx/onnx_test.cpp
```diff
@@ -3782,6 +3782,40 @@ TEST_CASE(tanh_test)
     EXPECT(p == prog);
 }
 
+TEST_CASE(thresholdedrelu_default_test)
+{
```
scxiao, line 3786: Also, please enable the unit tests in the file src/test/py/onnx_backend_test.py

pfultz2 (approved):

src/onnx/parse_thresholdedrelu.cpp
```diff
@@ -0,0 +1,40 @@
...
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        float alpha = 1.0;
+        if(contains(info.attributes, "alpha"))
+            alpha = parser.parse_value(info.attributes.at("alpha")).at<float>();
+
+        auto x_shape = args[0]->get_shape();
+        std::vector<double> zeros(x_shape.elements(), 0);
+        std::vector<double> alphas(x_shape.elements(), alpha);
```
pfultz2, line 26: Actually the scalar values should be broadcasted.

pfultz2: `add_common` might help to implicitly add broadcasts.

scxiao (approved):

src/onnx/parse_thresholdedrelu.cpp
pfultz2: It will still need to compute the multi_index.

pfultz2: Broadcasting will make it faster even with index calculation since it will access a lot less memory.

PR #929, opened 2021-08-25
Changes to support both OneDNN and ZenDNN builds

kahmed10:
- updated CMakeLists.txt (might need some refactoring for cleanup)
- created macro to switch between OneDNN and ZenDNN API calls
- disabled some operators for lowering on ZenDNN since some unit tests currently fail
- moved deconvolution compute method to operator header file so it can fallback to ref implementation


pfultz2 (approved):

src/targets/cpu/dnnl.cpp
```diff
@@ -2,6 +2,9 @@
 
 #if defined(__GNUC__) && __GNUC__ <= 5
 namespace std {
+#ifdef MIGRAPHX_ENABLE_ZENDNN
+namespace dnnl = zendnn;
```
umangyadav, line 6: Wouldn't `dnnl=zendnn` assignment come from `dnnl.hpp` ? If so, do we need another assignment here ?

src/targets/cpu/include/migraphx/cpu/dnnl.hpp
```diff
@@ -7,14 +7,26 @@
 #include <migraphx/register_op.hpp>
 #include <migraphx/check_shapes.hpp>
 #include <unordered_map>
-#include <dnnl.hpp>
 #include <migraphx/errors.hpp>
 #include <migraphx/assert.hpp>
+#ifdef MIGRAPHX_ENABLE_ZENDNN
+#include <zendnn.hpp>
+#else
+#include <dnnl.hpp>
```
umangyadav, line 15: `dnnl.hpp` including itself ? I guess not, can you add a comment here where it is coming from ?

kahmed10: This file is `<migraphx/cpu/dnnl.hpp>`, and it's referencing the actual OneDNN and ZenDNN headers here.

src/targets/cpu/dnnl.cpp
kahmed10: I am getting compiler errors with this removed.

umangyadav: If errors are not easily fixable then, it is okay, we can keep this.

pfultz2: dnnl.hpp does not do `dnnl=zendnn` as it is completely unaware of zendnn. Also zendnn does not provide such an alias either, so this is necessary.

umangyadav (approved):

PR #931, opened 2021-08-28
Allow creating modules in a module pass

pfultz2:
This adds a `module_pass_manager` class that can be used for module passes. It provides a `create_module` method so module passes can create modules.

This is needed for pointwise ops, so each pointwise op can be fused in a submodule.


tools/include/pass.hpp
```diff
@@ -24,17 +26,42 @@ struct pass
     /// A unique name used to identify the pass
     std::string name() const;
     /// Run the pass on the module
+    void apply(module_pass_manager& mpm) const;
     void apply(module& m) const;
```
umangyadav, line 30: Same here ? 

Also curious why do we have the same logic both in `src/.../pass.hpp` and `tools/.../pass.hpp` ?

src/include/migraphx/pass.hpp
```diff
@@ -24,20 +26,45 @@ struct pass
     /// A unique name used to identify the pass
     std::string name() const;
     /// Run the pass on the module
+    void apply(module_pass_manager& mpm) const;
     void apply(module& m) const;
```
umangyadav, line 30: Do we need this method that applies on Module directly ?

tools/include/pass.hpp
pfultz2: > Also curious why do we have the same logic both in src/.../pass.hpp and tools/.../pass.hpp ?

The header under tools/include are ran through te.py to generate the type erasure code. This generated file is then placed under src/include. See [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/wiki/Developer's-Guide#type-erasure).

src/include/migraphx/pass.hpp
pfultz2: Yes as all of our passes are written using this method.

umangyadav: Then, Should all of them now use module_pass_manager instead 
?

pfultz2: I dont see a reason to change it.

scxiao (approved):

umangyadav (approved):

PR #937, opened 2021-08-31
Add ThresholdedRelu to onnx parser

### LINKED ISSUE #888: Add ThresholdedRelu
https://github.com/onnx/onnx/blob/master/docs/Operators.md#ThresholdedRelu

Used in SSD models (Mask RCNN/Faster RCNN).

turneram:
Add ability to parse ThresholdedRelu ONNX operator. 

Resolves #888


src/onnx/parse_thresholdedrelu.cpp
```diff
@@ -0,0 +1,41 @@
...
+    {
+        float alpha = 1.0;
+        if(contains(info.attributes, "alpha"))
+            alpha = parser.parse_value(info.attributes.at("alpha")).at<float>();
+
+        auto x_shape = args[0]->get_shape();
+
+        auto lit_zero = info.add_literal(migraphx::literal{migraphx::shape{x_shape.type()}, {0}});
+        auto lit_alpha =
+            info.add_literal(migraphx::literal{migraphx::shape{x_shape.type()}, {alpha}});
```
umangyadav, line 28: Curious if alpha is always literal or could there be case where alpha is also tensor. 

Secondly, Does alpha has be to always greater than zero ? Because, output of normal ReLU is always non-negative.

src/onnx/parse_thresholdedrelu.cpp
```diff
@@ -0,0 +1,41 @@
...
+struct parse_thresholdedrelu : op_parser<parse_thresholdedrelu>
+{
+    std::vector<op_desc> operators() const { return {{"ThresholdedRelu"}}; }
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& parser,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        float alpha = 1.0;
```
umangyadav, line 20: Any reason why default value is 1 ? If it is zero then, it can behave like Normal ReLU.

test/onnx/onnx_test.cpp
```diff
@@ -3782,6 +3782,63 @@ TEST_CASE(tanh_test)
     EXPECT(p == prog);
 }
 
+TEST_CASE(thresholdedrelu_default_test)
```
umangyadav, line 3785: Just wondering, 
All three tests different in alpha value or data type. I was wondering if there is thing like test parametrization. For example as in `pytest`.

umangyadav (approved):

src/onnx/parse_thresholdedrelu.cpp
turneram: Alpha is an attribute that gets passed in, so it will always be a single value that gets braodcasted to the shape of the input tensor.

src/onnx/parse_thresholdedrelu.cpp
turneram: That is per the onnx specification

scxiao (approved):

kahmed10 (approved):

pfultz2 (approved):

PR #949, opened 2021-09-10
Add flag to bypass passes on modules 

pfultz2:
This is needed to bypass passes when fusing pointwise operators into a module.


src/module.cpp
```diff
@@ -28,6 +28,7 @@ struct module_impl
     std::unordered_set<instruction*> instruction_set;
     std::string name;
     uint32_t nparams = 0;
+    bool bypass      = false;
```
scxiao, line 31: Do we need to copy the bypass flag in the module::assign() method?

scxiao (approved):

umangyadav (approved):

scxiao (approved):

PR #951, opened 2021-09-13
Dpp opts for wavefront 32

kahmed10:
- For now, checks wavefront size, then changes implementation and number of threads for DPP reduce
- Added a few more tests with larger shapes


src/targets/gpu/device/layernorm.cpp
```diff
@@ -86,10 +86,13 @@ __device__ void layernorm(index_int i,
...
 
     auto mean = [&](auto z) {
-        return auto_block_reduce<MaxBlockSize>(
-                   idx, sum{}, value_type(0), relements_v, [=](auto) { return z; }) /
-               value_type(relements);
-
+        auto m = auto_block_reduce<MaxBlockSize>(
+                     idx, sum{}, value_type(0), relements_v, [=](auto) { return z; }) /
+                 value_type(relements);
+#if __AMDGCN_WAVEFRONT_SIZE == 32
```
pfultz2, line 92: This is a workaround, so we should have a `ifdef` to define the workaround:

```cpp
#ifndef MIGRAPHX_WORKAROUND_NAVI_DPP_SYNC
#if __AMDGCN_WAVEFRONT_SIZE == 32
#define MIGRAPHX_WORKAROUND_NAVI_DPP_SYNC 1
#else
#define MIGRAPHX_WORKAROUND_NAVI_DPP_SYNC 0
#endif
#endif
```

So we can disable the workaround to submit a bug report to the compiler.

kahmed10: fixed

pfultz2 (approved):

src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp
```diff
@@ -118,9 +116,11 @@ __device__ inline void dpp_reduce(float& x, sum)
                      "s_nop 1\n"
                      "v_add_f32 %0 %0 %0 row_shr:8 bank_mask:0xc\n"
                      "s_nop 1\n"
```
hgaspar, line 118: What happens if we increment the n_op to 2?  Is the __syncthreads still needed?

However, overall, a much better way to implement waits for gds/lds memory transactions to become visible is:
s_waitcnt     lgkmcnt(0)

Can this allow getting rid of __syncthreads().  
Besides the __syncthreads you added in this workaround, there are also other __syncthreads in this file, which, on a superficial look, appear unnecessary. syncthreads arguably does too much, and if we can avoid it, we should...

kahmed10: I have tried with increasing n_op to 2, and I still wasn't able to get the correct results.

As for the other __syncthreads(), I can try opening a separate PR if we can remove them, but for now I didn't want to modify that part of the code.

src/targets/gpu/device/layernorm.cpp
```diff
@@ -86,10 +94,13 @@ __device__ void layernorm(index_int i,
...
     auto mean = [&](auto z) {
-        return auto_block_reduce<MaxBlockSize>(
-                   idx, sum{}, value_type(0), relements_v, [=](auto) { return z; }) /
-               value_type(relements);
-
+        auto m = auto_block_reduce<MaxBlockSize>(
+                     idx, sum{}, value_type(0), relements_v, [=](auto) { return z; }) /
+                 value_type(relements);
+#if MIGRAPHX_WORKAROUND_NAVI_DPP_SYNC
+        __syncthreads();
```
pfultz2, line 101: Some other things to try for a workaround is `__builtin_amdgcn_s_barrier`, `__threadfence`, or `__threadfence_block`.

kahmed10: looks like `__builtin_amdgcn_s_barrier` passes correctness while improving workaround performance.

hgaspar (approved):

PR #952, opened 2021-09-14
Roialign

scxiao:
Implementation of the roialign operator. For now, we have only the ref implementation. When we run a model on the GPU, we fall back the execution to use the ref implementation.


src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,323 @@
...
+                                      bin_size_h,
+                                      bin_size_w,
+                                      roi_bin_grid_h,
+                                      roi_bin_grid_w,
+                                      pre_calc);
+
+                std::vector<int64_t> lens1 = {channels, pooled_height, pooled_width};
+                std::vector<std::size_t> comp_lens1(lens1.begin(), lens1.end());
+                shape comp_s1{migraphx::shape::float_type, comp_lens1};
+                std::vector<int64_t> vec_index(channels, 0);
```
turneram, line 282: What is the reason for having this as a vector of 0s? And if it's always passing in the value 0, what is the purpose of the `index` parameter for `calc_pooling`?

scxiao: As you can see, the calc_pooling is called in the shape_for_each(), so index is 0 for the first time calc_pooling called, and for the next time, the initial value of index is not 0 anymore. It is increased in the calc_pooling func and index is a reference in the function calc_pooling.

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+                                      roi_start_h,
+                                      roi_start_w,
+                                      bin_size_h,
+                                      bin_size_w,
+                                      roi_bin_grid_h,
+                                      roi_bin_grid_w,
+                                      pre_calc);
+
+                std::vector<int64_t> lens1 = {channels, pooled_height, pooled_width};
+                std::vector<std::size_t> comp_lens1(lens1.begin(), lens1.end());
```
umangyadav, line 278: Wouldn't it implicitly cast if you say 
`vector<size_t> comp_lens1 = {channels, pooled_height, pooled_width}` ?

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+                         float bin_size_w,
+                         int64_t roi_bin_grid_h,
+                         int64_t roi_bin_grid_w,
+                         std::vector<pos_weight>& pos_weights) const
+    {
+        shape_for_each(comp_s, [&](auto idx) {
+            auto ph    = idx[0];
+            auto pw    = idx[1];
+            auto iy    = idx[2];
+            auto ix    = idx[3];
```
pfultz2, line 100: These should be arrays as well.

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+                pc.w1    = 0;
+                pc.w2    = 0;
+                pc.w3    = 0;
+                pc.w4    = 0;
+                return;
+            }
+
+            y          = (y <= 0) ? 0 : y;
+            x          = (x <= 0) ? 0 : x;
+            auto y_low = static_cast<int64_t>(y);
```
pfultz2, line 127: Just do `int64_t y_low = y;` instead.

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+            }
+
+            float ly = y - y_low;
+            float lx = x - x_low;
+            float hy = 1.0f - ly;
+            float hx = 1.0f - lx;
+            float w1 = hy * hx;
+            float w2 = hy * lx;
+            float w3 = ly * hx;
+            float w4 = ly * lx;
```
pfultz2, line 151: Can you just assign the values directly to `pos_weight`?

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+                float roi_end_w   = static_cast<float>(roi[roi_s.index({n, 2})] * spatial_scale);
+                float roi_end_h   = static_cast<float>(roi[roi_s.index({n, 3})] * spatial_scale);
+
+                // Force malformed ROIs to be 1x1
+                float roi_width =
+                    (roi_end_w - roi_start_w) > 1.0f ? (roi_end_w - roi_start_w) : 1.0f;
+                float roi_height =
+                    (roi_end_h - roi_start_h) > 1.0f ? (roi_end_h - roi_start_h) : 1.0f;
+                float bin_size_h = static_cast<float>(roi_height / pooled_height);
+                float bin_size_w = static_cast<float>(roi_width / pooled_width);
```
pfultz2, line 246: These casts can be removed as well.

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+                        int64_t roi_bin_grid_w,
+                        const std::vector<pos_weight>& pos_weights,
+                        int64_t& index,
+                        Op op) const
+    {
+        double output_val   = op.init();
+        const int64_t count = roi_bin_grid_h * roi_bin_grid_w;
+        for(int64_t iy = 0; iy < roi_bin_grid_h; iy++)
+        {
+            for(int64_t ix = 0; ix < roi_bin_grid_w; ix++)
```
pfultz2, line 201: Use `dfor` for these two loops.

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+
+        double final(double x, std::size_t y) { return (y == 0) ? 0.0 : (x / y); }
+    };
+
+    template <class T, class Op>
+    double calc_pooling(const T* data,
+                        int64_t roi_bin_grid_h,
+                        int64_t roi_bin_grid_w,
+                        const std::vector<pos_weight>& pos_weights,
+                        int64_t& index,
```
pfultz2, line 194: Return the index and output as a tuple instead of using an input parameter. You can use `std::tie` to assign each value to the vector.

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+        int64_t pooled_width  = out_lens[3];
+        const auto& x_lens    = args.at(0).get_shape().lens();
+        auto height           = x_lens[2];
+        auto width            = x_lens[3];
+        auto roi_s            = args.at(1).get_shape();
+
+        visit_all(result, args.at(0), args.at(1))([&](auto output, auto x, auto roi) {
+            const auto* batch_indices = args.at(2).cast<int64_t>();
+            par_for(n_rois, [&](auto n) {
+                const auto* bottom_data  = x.data();
```
pfultz2, line 232: Use iterators instead: `auto bottom_data = x.begin();`. You will need to update `calc_pooling` to remove the pointer.

src/include/migraphx/op/roialign.hpp
scxiao: I remember there is an build error for this case when I did the rewrite_rnn, see:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/8829d6ab2b75dd5c4e223029b83457f59de71e5b/src/rewrite_rnn.cpp#L258

Looks like it works fine now, just fixed it.

src/include/migraphx/op/roialign.hpp
scxiao: what does it mean? put "ph, pw, iy, ix" in an array?

src/include/migraphx/op/roialign.hpp
scxiao: fixed

src/include/migraphx/op/roialign.hpp
scxiao: fixed

src/include/migraphx/op/roialign.hpp
scxiao: fixed

src/include/migraphx/op/roialign.hpp
pfultz2: Well you would make two arrays:

```cpp
std::array<int64_t, 2> p = {idx[0}, idx[1]};
std::array<int64_t, 2> i = {idx[2], idx[3]};
```

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+#include <vector>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct roialign
+{
+    std::string coord_trans_mode = "half_pixel";
+    std::string mode             = "avg";
```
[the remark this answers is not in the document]
pfultz2: Can we open an issue to track this?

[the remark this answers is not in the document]
scxiao: Issue #983 is created.

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,321 @@
...
+        float w4;
+    };
+
+    void calc_pos_weight(const int64_t height,
+                         const int64_t width,
+                         const shape& comp_s,
+                         float roi_start_h,
+                         float roi_start_w,
+                         float bin_size_h,
+                         float bin_size_w,
```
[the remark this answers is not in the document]
scxiao: fixed

src/include/migraphx/op/roialign.hpp
```diff
@@ -0,0 +1,263 @@
...
+            // save weights and indeces
+            results[index].w = {hy * hx, hy * lx, ly * hx, ly * lx};
+        });
+
+        return results;
+    }
+
+    struct max_pool
+    {
+        double init() { return std::numeric_limits<double>::lowest(); }
```
hgaspar, line 137: All these doubles and floats in this file look like they could be templatized?

test/onnx/gen_onnx.py
```diff
@@ -3835,6 +3835,7 @@ def resize_upsample_pf_test():
     return ([node], [X], [Y], [scale_tensor])
 
 
+@onnx_test
```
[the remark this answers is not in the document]
scxiao: yes, it is in the onnx_test lists.

test/onnx/gen_onnx.py
```diff
@@ -3857,6 +3858,41 @@ def resize_upsample_pc_test():
...
+def roialign_default_test():
```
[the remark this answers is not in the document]
scxiao: I think the unit test "roialign_test" cover all non-default cases.

test/op_shape_test.cpp
```diff
@@ -1544,4 +1544,26 @@ TEST_CASE(where_broadcast_input)
     expect_shape(s2, migraphx::make_op("where"), s3, s1, s2);
 }
 
+TEST_CASE(roialign_test)
+{
+    migraphx::shape sx{migraphx::shape::float_type, {3, 4, 5, 6}};
+    migraphx::shape srois{migraphx::shape::float_type, {2, 4}};
+    migraphx::shape sbi{migraphx::shape::int64_type, {2}};
```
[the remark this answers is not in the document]
scxiao: The int64 datatype is required by the onnx specification.

cagery (approved):
LGTM.

PR #953, opened 2021-09-15
Make `file_options` an opaque object for ABI compatibility

umangyadav:


kahmed10 (approved):

src/api/api.cpp
```diff
@@ -1,3 +1,4 @@
+#include <migraphx/operation.hpp>
```
scxiao, line 1: Do we need this include here?

src/api/api.cpp
```diff
@@ -320,6 +316,16 @@ struct migraphx_onnx_options
     migraphx::onnx_options object;
 };
 
+extern "C" struct migraphx_file_options;
```
scxiao, line 319: How about the rename this to serialization_option? @pfultz2

src/api/api.cpp
umangyadav: It is auto-generated from `tools/.../api.cpp`  It is needed.

src/api/api.cpp
pfultz2: Lets just keep the name the same.

pfultz2 (approved):
I dont think we need to worry about the code coverage here.

scxiao (approved):

PR #954, opened 2021-09-15
Add multinomial op

### LINKED ISSUE #821: Add Multinomial op

turneram:
Add multinomial op to onnx parser with ref and GPU implementations.

The onnx parser inserts a literal of shape {batch_size, sample_size} with random values in the range [0, 1) and inserts existing ops to compute the cumulative density function. The multinomial operator multiplies the random values by the sum of the CDF and returns the index of the first element of the CDF that is greater than the result, representing samples randomly drawn from [0, class_size) that follow the log-probability distribution.

Resolves #821


src/onnx/parse_multinomial.cpp
```diff
@@ -0,0 +1,63 @@
...
+    {
+        int dtype = 6; // 6=in32, 7=int64
+        if(contains(info.attributes, "dtype"))
+            dtype = parser.parse_value(info.attributes.at("dtype")).at<int>();
+
+        size_t sample_size = 1;
+        if(contains(info.attributes, "sample_size"))
+            sample_size = parser.parse_value(info.attributes.at("sample_size")).at<int>();
+
+        float seed = 0.0;
```
kahmed10, line 29: According to the onnx spec, we need to generate a random seed if it's not provided. That means that for your onnx test, we should always have the seed field specified in order to have deterministic results on our CI.

turneram: Added

src/include/migraphx/op/multinomial.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <utility>
+#include <random>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct multinomial
+{
+    int dtype = 6;
```
scxiao, line 26: I would suggest use the shape::type_t as the attribute here. Then in the parse_multinominial, we can do the map from dtype to shape::type_t.

src/onnx/parse_multinomial.cpp
```diff
@@ -0,0 +1,64 @@
...
+        if(contains(info.attributes, "sample_size"))
+            sample_size = parser.parse_value(info.attributes.at("sample_size")).at<int>();
+
+        float seed = static_cast<float>(
+            std::chrono::high_resolution_clock::now().time_since_epoch().count());
+        if(contains(info.attributes, "seed"))
+            seed = parser.parse_value(info.attributes.at("seed")).at<float>();
+
+        // Compute cumulative density function
+        auto maxes =
```
scxiao, line 35: Where do these algos come from?

src/targets/gpu/multinomial.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <migraphx/gpu/context.hpp>
+#include <migraphx/tune_axis.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+
+shape hip_multinomial::compute_shape(const std::vector<shape>& inputs) const
+{
+    check_shapes{inputs, *this}.has(3).only_dims(2);
```
scxiao, line 12: Can we call the op.compute_shape() for the output shape?

src/include/migraphx/op/multinomial.hpp
turneram: Switched

src/targets/gpu/multinomial.cpp
turneram: Fixed

src/onnx/parse_multinomial.cpp
turneram: These accomplish the same computation found in MultinomialCompute up until the the Generate each sample comment from [here](https://github.com/microsoft/onnxruntime/blob/8d737f977056444a307f1b7f0bcd402fba62d790/onnxruntime/core/providers/cpu/generator/random.cc).

src/targets/gpu/device/multinomial.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+template <class T1, class T2>
+void __device__
+set_output(T1& cdf, T1& dist, T2& output, const size_t& i, size_t& j, const size_t& n)
+{
+    for(; j < n; ++j)
+    {
+        if(cdf[j] > (dist[i] * cdf[n - 1]))
+        {
+            output[i] = j;
```
scxiao, line 22: Compared to the ref implementation, I think this line should be

```
output[i] = j - start_of_j;
```
But why the test passed?

test/ref_ops_test.cpp
```diff
@@ -2687,6 +2688,58 @@ TEST_CASE(mul_test)
     EXPECT(migraphx::verify_range(results_vector, gold));
 }
 
+TEST_CASE(multinomial_test)
```
scxiao, line 2691: Can you turn on the multinomimal test in the file test/py/onnx_backend_test.py to check whether they can pass? Particularly for those with a see input.

src/targets/gpu/device/multinomial.cpp
turneram: You're right. I think the test was passing because it was for a single batch, which means `start_of_j` was just 0.

src/include/migraphx/op/multinomial.hpp
```diff
@@ -37,13 +37,7 @@ struct multinomial
         check_shapes{inputs, *this}.has(2).only_dims(2);
         size_t sample_size = inputs.back().lens().back();
 
-        if(dtype == 6)
```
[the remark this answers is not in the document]
turneram: Added

test/op_shape_test.cpp
```diff
@@ -1020,14 +1020,6 @@ TEST_CASE(multibroadcast)
     }
 }
 
-TEST_CASE(multinomial)
```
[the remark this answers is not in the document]
turneram: Re-added

src/onnx/parse_multinomial.cpp
```diff
@@ -0,0 +1,70 @@
...
+            dtype = parser.parse_value(info.attributes.at("dtype")).at<int>();
+        if(dtype == 7)
+            output_type = shape::type_t::int64_type;
+        else if(dtype != 6)
+            MIGRAPHX_THROW("Invalid output type: " + std::to_string(dtype) +
+                           ". Valid types are 6 (INT32) and 7 (INT64).");
+
+        size_t sample_size = 1;
+        if(contains(info.attributes, "sample_size"))
+            sample_size = parser.parse_value(info.attributes.at("sample_size")).at<int>();
```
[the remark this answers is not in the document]
turneram: Changed here and for the other attributes

src/targets/gpu/device/multinomial.cpp
turneram: Fixed and updated the test

src/include/migraphx/op/multinomial.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <migraphx/operation.hpp>
+#include <migraphx/check_shapes.hpp>
+#include <migraphx/stringutils.hpp>
+#include <migraphx/streamutils.hpp>
+#include <migraphx/literal.hpp>
+#include <migraphx/config.hpp>
+#include <migraphx/par_for.hpp>
+#include <migraphx/value.hpp>
+#include <migraphx/op/normalize_attribute.hpp>
+#include <migraphx/tune_axis.hpp>
```
pfultz2, line 15: There is a lot of includes that seem like they are not needes. Can you trim this down to only the ones needed?

src/onnx/parse_multinomial.cpp
```diff
@@ -0,0 +1,65 @@
...
+            migraphx::make_op("prefix_scan_sum", {{"axis", 1}, {"exclusive", false}}), cdf);
+
+        // Pre-compute random distribution
+        std::mt19937 gen(seed);
+        std::uniform_real_distribution<> dis(0.0, 1.0);
+        size_t batch_size = args[0]->get_shape().lens().front();
+        migraphx::shape dist_shape{migraphx::shape::float_type, {batch_size, sample_size}};
+
+        std::vector<float> random_dist(batch_size * sample_size);
+        std::transform(random_dist.begin(), random_dist.end(), random_dist.begin(), [&](auto) {
```
pfultz2, line 53: Use `std::generate` instead of `std::transform`.

src/targets/gpu/device/multinomial.cpp
```diff
@@ -0,0 +1,54 @@
...
+    size_t sample_size = result.get_shape().lens().back();
+
+    hip_visit_all(arg0, arg1)([&](auto cdf, auto dist) {
+        result.visit([&](auto out) {
+            hip_visit_views(out)([&](auto output) {
+                gs_launch(stream, batch_size * sample_size)([=](auto i) __device__ {
+                    auto idx     = output.get_shape().multi(i);
+                    size_t begin = idx.front() * class_size;
+                    size_t end   = (idx.front() + 1) * class_size;
+                    set_output(cdf, dist, output, i, begin, end);
```
pfultz2, line 44: Can you add a `constexpr` upper_bound so this could be ran on the device(this could just be added to the .cpp file for now)?

```cpp
template<class Iterator, class T>
constexpr Iterator upper_bound(Iterator first, Iterator last, const T& value)
{
    Iterator it;
    typename std::iterator_traits<Iterator>::difference_type count, step;
    count = std::distance(first, last);
 
    while (count > 0) {
        it = first; 
        step = count / 2; 
        std::advance(it, step);
        if (!(value < *it)) {
            first = ++it;
            count -= step + 1;
        } 
        else
            count = step;
    }
    return first;
}
```

src/include/migraphx/op/multinomial.hpp
turneram: Done

src/onnx/parse_multinomial.cpp
turneram: Changed here and throughout tests

src/targets/gpu/device/multinomial.cpp
turneram: Added

scxiao (approved):

src/include/migraphx/op/multinomial.hpp
```diff
@@ -0,0 +1,64 @@
...
+        argument result{output_shape};
+        size_t batch_size  = output_shape.lens().front();
+        size_t class_size  = args[0].get_shape().lens().back();
+        size_t sample_size = output_shape.lens().back();
+
+        visit_all(args[0], args[1])([&](auto cdf, auto dist) {
+            result.visit([&](auto output) {
+                par_for(batch_size * sample_size, [&](auto i) {
+                    auto idx        = args[1].get_shape().multi(i);
+                    auto* cdf_begin = cdf.data() + (idx[0] * class_size);
```
pfultz2, line 47: This should probably use iterators instead of pointers so it can work with non-standard shape. I think replacing `cdf.data()` with `cdf.begin()` should work.

src/targets/gpu/multinomial.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <migraphx/tune_axis.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+
+shape hip_multinomial::compute_shape(std::vector<shape> inputs) const
+{
+    inputs.pop_back();
+    return op.compute_shape(inputs);
```
pfultz2, line 13: This should require standard shape for inputs.

pfultz2: This has been fixed yet.

turneram: Ah, my bad. Should the check only occur here, or should the `compute_shape` method for the ref version also check for standard shape?

pfultz2: The ref version shouldn't need to check, since it should work with non-standard inputs shapes.

src/include/migraphx/op/multinomial.hpp
```diff
@@ -0,0 +1,64 @@
...
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.dtype, "dtype"));
+    }
+
+    std::string name() const { return "multinomial"; }
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(2).only_dims(2).standard();
```
pfultz2, line 26: Yea, `standard()` can be removed here. Since you use the iterators then it will work for any shape. On the GPU implementation, pointers are used(even when doing `.begin()` unfortunately) so thats why the GPU needs standard shape.

src/targets/gpu/multinomial.cpp
turneram: Okay, got it.

src/include/migraphx/op/multinomial.hpp
turneram: Removed

pfultz2 (approved):

src/onnx/parse_multinomial.cpp
kahmed10: Just one last minor suggestion: can you add a comment on the source for this algorithm? For exmaple in onnx runtime source code, they mention tensorflow, and we can mention we followed a similar variation.

turneram:
@kahmed10 I've added the comments you requested

kahmed10 (approved):

scxiao:
All the checks passed, it just does not display correctly, so this PR can be merged.

PR #945, opened 2021-09-08
Remove alpha and beta attributes from dot operator

umangyadav:
~~WIP:~~

This PR aims to remove alpha and beta attributes from dot operator completely. 

Previously dot operator was defined as ` C = alpha * A . B + beta * C` where `*` is scalar multiplication and `.` is dot product or matrix multiplication depending on dimension of the inputs. 

Aim is to have the definition of dot operator as ` C = A . B` without having `alpha` or `beta`.

In order to achieve the same effect as alpha and beta (1) it multiplies the one of the inputs to the dot operator with `alpha` value. (2) if beta is present then, multiplies the `C` with  beta and then adds into the output from step 1. 

PR adds a method to handle above two steps with necessary type conversions and broadcasting, and applies those changes to the tests and src files. 

~~PR is still WIP because~~
1. ~~One of the quantization test fails. Investigation going on~~. Tolerance is increased.
2.  ~~Changes in the lowering to not use alpha or beta fields~~
3. ~~Final removal alpha and beta fields from dot op~~ 
4. ~~Changes in decompose pass and test~~

PR #961, opened 2021-09-20
Remove alpha and beta from `dot` and `quant_dot`

umangyadav:
#945


src/targets/gpu/fuse_ops.cpp
```diff
@@ -697,52 +697,6 @@ struct find_conv_bias_relu
         apply_conv_bias<miopen_conv_bias_relu>(*ctx, p, std::move(r));
     }
 };
-
-struct find_gemm_add
```
umangyadav, line 701: @pfultz2  This one relies on setting `beta` value of 1 for the `dot` operator. That's why i'd removed. 

I can add beta field to `rocblas_gemm` and use it.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -717,7 +717,7 @@ struct find_gemm_add
         auto gemm = any_cast<rocblas_gemm<op::dot>>(gemm_ins->get_operator());
 
         // Already fused gemm
-        if(not float_equal(gemm.op.beta, 0))
+        if(not float_equal(gemm.beta, 0))
```
umangyadav, line 720: @pfultz2  This is conflicting with the design. 

In order to allow this fusion, I have to add beta flag to `rocblas_gemm`, which contains the `dot` op as one of the members.  If i do this fusion, i'll have to allow 3 input arguments to the `dot` operator. If `dot` doesn't have any `beta` field itself then, allowing 3 input args has undefined meaning.

test/quantization.cpp
```diff
@@ -753,8 +737,7 @@ TEST_CASE(dot_half_1arg)
         auto* mm = p.get_main_module();
         migraphx::shape s{migraphx::shape::half_type, {9, 9}};
         auto x = mm->add_parameter("x", s);
-        auto r =
-            mm->add_instruction(migraphx::make_op("dot", {{"alpha", 1.0f}, {"beta", 0.0f}}), x, x);
+        auto r = migraphx::add_apply_alpha_beta<float>(*mm, {x, x}, "dot");
```
scxiao, line 740: You can just use 

```
auto r  = mm->add_instruction(migraphx::make_op("dot"), x, x);
```

test/quantization.cpp
```diff
@@ -782,8 +765,7 @@ TEST_CASE(dot_half_1arg)
                                    zp_b);
         auto qb  = mm->add_instruction(migraphx::make_op("quantizelinear"), x, scale_b, zp_b);
         auto dqb = mm->add_instruction(migraphx::make_op("dequantizelinear"), qb, scale_b, zp_b);
-        auto r   = mm->add_instruction(
-            migraphx::make_op("dot", {{"alpha", 1.0f}, {"beta", 0.0f}}), dqa, dqb);
+        auto r   = migraphx::add_apply_alpha_beta<float>(*mm, {dqa, dqb}, "dot");
```
scxiao, line 768: Same as above, no need to call this function.

test/simplify_qdq_test.cpp
```diff
@@ -127,12 +128,11 @@ TEST_CASE(dot)
...
-        auto d1 = add_quantize_op(m1, "dequantizelinear", q1, scale, zero);
-        auto q2 = add_quantize_op(m1, "quantizelinear", t2, scale, zero);
-        auto d2 = add_quantize_op(m1, "dequantizelinear", q2, scale, zero);
-        auto dot =
-            m1.add_instruction(migraphx::make_op("dot", {{"alpha", 1}, {"beta", 0}}), d1, d2);
+        auto q1  = add_quantize_op(m1, "quantizelinear", t1, scale, zero);
+        auto d1  = add_quantize_op(m1, "dequantizelinear", q1, scale, zero);
+        auto q2  = add_quantize_op(m1, "quantizelinear", t2, scale, zero);
+        auto d2  = add_quantize_op(m1, "dequantizelinear", q2, scale, zero);
+        auto dot = migraphx::add_apply_alpha_beta<float>(m1, {d1, d2}, "dot");
```
scxiao, line 135: Just call add_instruction here.

test/simplify_qdq_test.cpp
```diff
@@ -168,22 +167,19 @@ TEST_CASE(dot_non_zero_point)
...
```
scxiao, line 174: Same

test/quantization.cpp
umangyadav: Done

test/quantization.cpp
umangyadav: done

test/simplify_qdq_test.cpp
umangyadav: done

test/simplify_qdq_test.cpp
umangyadav: done

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,72 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+template <typename T = float>
+instruction_ref insert_apply_alpha_beta(module& m,
+                                        instruction_ref pos,
+                                        const std::vector<instruction_ref>& args,
+                                        const std::string& op_name,
+                                        T alpha = 1.0f,
+                                        T beta  = 0.0f)
+{
+    auto l1       = args[0];
```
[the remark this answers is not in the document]
umangyadav: added.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -19,6 +21,7 @@ template <class Op>
 struct rocblas_gemm
 {
     Op op;
+    float beta          = 0;
```
pfultz2, line 24: There should be an `alpha` field here as well.

umangyadav: I can add it. but it won't be used. Only reason we have beta is that it allows fusion Gemm+Add fusion.

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,77 @@
...
+                                        const std::string& op_name,
+                                        T alpha = 1.0f,
+                                        T beta  = 0.0f)
+{
+    if(!(op_name == "dot" or op_name == "quant_dot"))
+        MIGRAPHX_THROW("Alpha and Beta calculation can be only be applied to either dot or "
+                       "quant_dot operator");
+
+    auto l1       = args[0];
+    auto l2       = args[1];
```
pfultz2, line 26: What is `l1` and what is `l2`? Can you use better names? Maybe `arg0` and `arg1`?

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,77 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+template <typename T = float>
+instruction_ref insert_apply_alpha_beta(module& m,
+                                        instruction_ref pos,
+                                        const std::vector<instruction_ref>& args,
+                                        const std::string& op_name,
+                                        T alpha = 1.0f,
+                                        T beta  = 0.0f)
+{
+    if(!(op_name == "dot" or op_name == "quant_dot"))
```
pfultz2, line 21: I dont think we should check the op_name is "dot" or "quant_dot". This function can be used for convolutions as well.

src/include/migraphx/apply_alpha_beta.hpp
scxiao: Maybe name "a" and "b" since we usually say a * b in matrix multiplication.

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,77 @@
...
+        MIGRAPHX_THROW("Alpha and Beta calculation can be only be applied to either dot or "
+                       "quant_dot operator");
+
+    auto l1       = args[0];
+    auto l2       = args[1];
+    auto dot_type = l1->get_shape().type();
+    if(!float_equal(static_cast<float>(alpha), 1.0f))
```
pfultz2, line 28: Cast to float shouldn't be needed, as `float_equal` will use the common type.

pfultz2: This should be written as `float_equal(alpha, 1)` so it avoids a float comparison when using integers.

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,77 @@
...
+        l1 = insert_common_op(m, pos, migraphx::make_op("mul"), {alpha_literal, l1});
+        if(l1->get_shape().type() != dot_type)
+        {
+            l1 = m.insert_instruction(pos, make_op("convert", {{"target_type", dot_type}}), l1);
+        }
+    }
+    auto dot_res = m.insert_instruction(pos, migraphx::make_op(op_name), l1, l2);
+    if(args.size() == 3)
+    {
+        if(not float_equal(static_cast<float>(beta), 0.0f) && args[2]->get_shape().elements() > 0)
```
pfultz2, line 40: Same here.

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+template <typename T = float>
+instruction_ref insert_apply_alpha_beta(module& m,
+                                        instruction_ref pos,
+                                        const std::vector<instruction_ref>& args,
+                                        const std::string& op_name,
+                                        T alpha = 1.0f,
+                                        T beta  = 0.0f)
```
pfultz2, line 19: This function should accept that values using the `literal` class so this implementation can be moved to the cpp file. We can then provide a templated method to implicitly convert them to `literal`:

```cpp
instruction_ref insert_apply_alpha_beta_impl(module& m,
                                        instruction_ref pos,
                                        const std::vector<instruction_ref>& args,
                                        const std::string& op_name,
                                        literal alpha,
                                        literal beta);

template <class T = float>
instruction_ref insert_apply_alpha_beta(module& m,
                                        instruction_ref pos,
                                        const std::vector<instruction_ref>& args,
                                        const std::string& op_name,
                                        T alpha = 1.0f,
                                        T beta  = 0.0f)
{
    return insert_apply_alpha_beta_impl(m, pos, args, op_name, literal{alpha}, literal{beta});
}
```

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    auto arg0     = args[0];
+    auto arg1     = args[1];
+    auto dot_type = arg0->get_shape().type();
+    if(!float_equal(alpha, 1.0f))
```
pfultz2, line 24: Drop the `f` suffix and make it just `1.0`

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,74 @@
...
+        arg0 = insert_common_op(m, pos, migraphx::make_op("mul"), {alpha_literal, arg0});
+        if(arg0->get_shape().type() != dot_type)
+        {
+            arg0 = m.insert_instruction(pos, make_op("convert", {{"target_type", dot_type}}), arg0);
+        }
+    }
+    auto dot_res = m.insert_instruction(pos, migraphx::make_op(op_name), arg0, arg1);
+    if(args.size() == 3)
+    {
+        if(not float_equal(beta, 0.0f) && args[2]->get_shape().elements() > 0)
```
pfultz2, line 36: Drop the `f` suffix here as well.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
umangyadav: alpha is added.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: renamed.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: removed the Check.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: done.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: done

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,74 @@
...
+instruction_ref insert_apply_alpha_beta(module& m,
+                                        instruction_ref pos,
+                                        const std::vector<instruction_ref>& args,
+                                        const std::string& op_name,
+                                        T alpha = 1.0f,
+                                        T beta  = 0.0f)
+{
+    auto arg0     = args[0];
+    auto arg1     = args[1];
+    auto dot_type = arg0->get_shape().type();
```
pfultz2: Maybe name this variable `input_type`.

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <migraphx/common.hpp>
+#include <migraphx/module.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+template <typename T = float>
+instruction_ref insert_apply_alpha_beta(module& m,
+                                        instruction_ref pos,
+                                        const std::vector<instruction_ref>& args,
+                                        const std::string& op_name,
```
pfultz2, line 17: If we want to use this for other operators like convolution or if we later add attributes to be set on `dot`, then we accept an `operation` rather than an op name.

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,74 @@
...
+    if(args.size() == 3)
+    {
+        if(not float_equal(beta, 0.0f) && args[2]->get_shape().elements() > 0)
+        {
+            auto out_lens   = arg0->get_shape().lens();
+            out_lens.back() = arg1->get_shape().lens().back();
+            auto arg2       = args[2];
+            auto arg2_lens  = arg2->get_shape().lens();
+            dot_type        = arg2->get_shape().type();
+            if(!std::equal(out_lens.begin(), out_lens.end(), arg2_lens.begin(), arg2_lens.end()))
```
pfultz2: This can be written as `out_lens != arg2_lens`. No need to use `std::equal` here.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: Done.  Added a source file.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: Done

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: Done

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: Done

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: Done

src/include/migraphx/apply_alpha_beta.hpp
pfultz2: Can the parameter be changed to accept `operation` instead of just the op_name?

test/program_test.cpp
```diff
@@ -160,9 +161,7 @@ TEST_CASE(program_copy)
         auto para1 = mm1->add_parameter("m1", s1);
         auto para2 = mm1->add_parameter("m2", s2);
         auto para3 = mm1->add_parameter("m3", s3);
-        mm1->add_instruction(
-            migraphx::make_op("dot", {{"alpha", 0.31f}, {"beta", 0.28f}}), para1, para2, para3);
-
+        migraphx::add_apply_alpha_beta<float>(*mm1, {para1, para2, para3}, "dot", 0.31f, 0.28f);
```
pfultz2, line 164: Can the `<float>` be removed? It should deduce the template type.

test/quantization.cpp
```diff
@@ -431,7 +432,7 @@ TEST_CASE(op_capture)
         auto pb = mm->add_parameter("b", s2);
         auto pc = mm->add_parameter("c", s2);
         auto pa = mm->add_instruction(migraphx::make_op("add"), p1, p2);
-        auto ps = mm->add_instruction(migraphx::make_op("dot"), pa, pb, pc);
+        auto ps = migraphx::add_apply_alpha_beta<float>(*mm, {pa, pb, pc}, "dot", 1, 1);
```
pfultz2, line 435: `<float>` can be removed? It should deduce the type from the template.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: Done

test/program_test.cpp
umangyadav: Done

test/quantization.cpp
umangyadav: Done

src/include/migraphx/apply_alpha_beta.hpp
```diff
@@ -0,0 +1,43 @@
...
+                                        T alpha = 1.0f,
+                                        T beta  = 0.0f)
+{
+    return insert_apply_alpha_beta(m, pos, args, op, literal{T{alpha}}, literal{T{beta}});
+}
+
+template <typename T = float>
+instruction_ref add_apply_alpha_beta(module& m,
+                                     const std::vector<instruction_ref>& args,
+                                     const std::string& op_name,
```
pfultz2, line 34: This should take the operation here too so its consistent.

src/onnx/parse_matmul.cpp
```diff
@@ -66,9 +66,8 @@ struct parse_matmul : op_parser<parse_matmul>
                     make_op("multibroadcast", {{"out_lens", l1_broadcasted_lens}}), l1);
             }
         }
-
-        auto dot_res =
-            info.add_instruction(make_op(opd.op_name, {{"alpha", 1}, {"beta", 0}}), bl0, bl1);
+        instruction_ref dot_res;
+        dot_res          = info.add_instruction(make_op(opd.op_name), bl0, bl1);
```
pfultz2, line 70: Why is this two lines? It should just be `auto dot_res = info.add_instruction(make_op(opd.op_name), bl0, bl1)`.

umangyadav: I think it was automatic formatting. Let me check though.

src/include/migraphx/apply_alpha_beta.hpp
umangyadav: It would be consistent, but this method is invoked in many places in tests and each time, i'll have to pass `migraphx::make_op(" ")`. I thought passing `op_name` was just shorter.

src/apply_alpha_beta.cpp
```diff
@@ -0,0 +1,57 @@
...
+            a = m.insert_instruction(pos, make_op("convert", {{"target_type", input_type}}), a);
+        }
+    }
+    auto dot_res = m.insert_instruction(pos, op, a, b);
+    if(args.size() == 3)
+    {
+        if(not float_equal(beta.at<float>(0), 0.0) && args[2]->get_shape().elements() > 0)
+        {
+            auto out_lens   = a->get_shape().lens();
+            out_lens.back() = b->get_shape().lens().back();
```
scxiao, line 34: The out_lens can be computed as:

```
auto out_lens = dot_res.get_shape().lens();
```

The above approach applies only to dot.

src/apply_alpha_beta.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if(!float_equal(alpha.at<float>(0), 1.0))
+    {
+        auto alpha_literal = m.add_literal(alpha);
+        a                  = insert_common_op(m, pos, migraphx::make_op("mul"), {alpha_literal, a});
+        if(a->get_shape().type() != input_type)
+        {
+            a = m.insert_instruction(pos, make_op("convert", {{"target_type", input_type}}), a);
+        }
+    }
+    auto dot_res = m.insert_instruction(pos, op, a, b);
```
scxiao, line 28: maybe we can rename dot_res to op_res

src/include/migraphx/apply_alpha_beta.hpp
scxiao: should we use:

```
T alpha = T{1},
```
here?

src/apply_alpha_beta.cpp
umangyadav: Done

umangyadav:
> The `compute_shape` method of `dot` and `quant_dot` needs to be updated to require two inputs, something like: `check_shapes{inputs, *this}.same_type().has(2)`. Also the checks for `inputs.size() == 3` can be removed as well.

Done.

pfultz2 (approved):

scxiao (approved):

PR #963, opened 2021-09-22
Add remaining random ops for Barracuda models

### LINKED ISSUE #907: Support the Random Operators for Unity Barracuda
Support for the following operators...
* RandomNormal
* RandomNormalLike

This is a breakout issue from  #824 .

### LINKED ISSUE #959: Support the RandomUniform Operators for Unity Barracuda
Support for the following operators...
* RandomUniform
* RandomUniformLike

This is a breakout issue from  #824 .

turneram:
Add RandomNormal, RandomNormalLike, RandomUniform, and RandomUniformLike to onnx parser and onnx tests 

Each pair of Random\*/Random\*Like is implemented using a single op_parser because the ops share the same essential attributes and algorithm with the difference that Random\*Like get the output type and/or shape from an input argument and Random* take both from attributes. 

Resolves #907 
Resolves #959


umangyadav (approved):

src/onnx/parse_randomnormal_ops.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+        shape out_shape;
+        if(contains(info.attributes, "shape"))
+        {
+            std::vector<int> out_lens;
+            literal ls = parser.parse_value(info.attributes.at("shape"));
+            ls.visit([&](auto s) { out_lens.assign(s.begin(), s.end()); });
+            out_shape = shape{out_type, out_lens};
+        }
+        else if(args.size() == 1)
```
umangyadav, line 58: for the `Random*Like` it would make use of this, it takes both input and output as args, so shouldn't this be size of 2 ?

test/onnx/gen_onnx.py
```diff
@@ -2947,6 +2947,186 @@ def quantizelinear_neg_axis_test():
...
+    scale = 1.5
+    seed = 0.0
+    input = helper.make_tensor_value_info('input', TensorProto.FLOAT16,
+                                          [2, 3, 4])
+    output = helper.make_tensor_value_info('output', TensorProto.FLOAT16,
+                                           [2, 3, 4])
+
+    node = onnx.helper.make_node('RandomNormalLike',
+                                 inputs=['input'],
+                                 outputs=['output'],
```
umangyadav, line 3015: Here it passes both inputs and outputs. Does that reflect as `args` of size 2 during the parsing ?

src/onnx/parse_randomnormal_ops.cpp
```diff
@@ -0,0 +1,82 @@
...
+                          const onnx_parser& parser,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        int dtype      = 1;
+        bool use_dtype = false;
+        if(contains(info.attributes, "dtype"))
+        {
+            dtype     = info.attributes.at("dtype").i();
+            use_dtype = true;
```
kahmed10, line 30: I don't think we need `use_dtype`, since the default is float and would be overriden if the attribute value is changed. For L63, we can just construct the shape using `out_type`.

turneram: The reason I added `use_dtype` is because the default behavior for Random*Like is to use the shape and type of the input tensor, but there are two cases where you want to use only the shape but not the type: 1) the `dtype` attribute is set, and 2) the type of the input tensor is not half, single, or double precision float.

src/onnx/parse_randomnormal_ops.cpp
```diff
@@ -0,0 +1,82 @@
...
+        if(contains(info.attributes, "shape"))
+        {
+            std::vector<int> out_lens;
+            literal ls = parser.parse_value(info.attributes.at("shape"));
+            ls.visit([&](auto s) { out_lens.assign(s.begin(), s.end()); });
+            out_shape = shape{out_type, out_lens};
+        }
+        else if(args.size() == 1)
+        {
+            if(valid_types.find(args[0]->get_shape().type()) == valid_types.end())
```
scxiao, line 60: For this case, if the data_type in the input is invalid, I would prefer throwing out an exception as:

```
            if(not use_dtype)
            {
                if(valid_types.find(args[0]->get_shape().type()) == valid_types.end())
                {
                    MIGRAPHX_THROW();
                }
                out_type = args[0]->get_shape().type();
           }
```

pfultz2, line 60: Does `contains(valid_types, args[0]->get_shape().type())` work here?

src/onnx/parse_randomnormal_ops.cpp
kahmed10: I think I see the issue. According to the onnx spec, RandomNormal and RandomUniform have `dtype` as a mandatory attribute whereas for the others `dtype` is optional. Since we are parsing both in the same cpp file, maybe add a comment for why the boolean is necessary?

kahmed10 (approved):

test/onnx/gen_onnx.py
turneram: The size of `args` will just be the the number of inputs.

src/onnx/parse_randomnormal_ops.cpp
turneram: Added and updated tests

turneram: Fixed

scxiao (approved):

pfultz2 (approved):

PR #966, opened 2021-09-28
Correcting hip package dependencies

ashutom:
Hip package names have been changed because of the new rocm packaging standards. Hence the hip-hcc dependency needs to be changed to hip-rocclr hip-base

Checked the changes to be working with ROCm 4.3 4.4 & 4.5


causten (approved):
This change passes in our CI

Issue #948, opened 2021-09-09
Operator dynamic output shape


scxiao:
Some operators output dynamic shape for different inputs (even input shapes are the same), but MIGraphX can only generate fixed output shape. Because of this, MIGraphX always use the max possible output shape even the actual output shape is smaller.

This problem applies to: Loop, NonZero, NonMaxSuppressoin. We need to come up with a general approach to resolve this problem.

causten:
Is this a dup of #689  ?

CharlieL7:
For reference we currently create a buffer of maximum size for NonMaxSuppression using the `max` `dynamic_dimension` values. Will likely do the same for other reference operators with dynamic shape functions.

CharlieL7:
Dynamic shapes are not fully done (need more operators, GPU support), but at a stage where we can close this issue for now.

PR #968, opened 2021-09-30
Nms op ref implementation

scxiao:
This PR is the ref implementation of the nonmaxsuppression operator. It always returns the max possible output shape, which is the problem tracked in issue #948.


src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -0,0 +1,259 @@
...
+        float x2_max{};
+        float y2_max{};
+        float intersection_x_min{};
+        float intersection_x_max{};
+        float intersection_y_min{};
+        float intersection_y_max{};
+
+        const float* box1 = boxes_data + 4 * box_index1;
+        const float* box2 = boxes_data + 4 * box_index2;
+        // center_point_box_ only support 0 or 1
```
kahmed10, line 84: should we be checking if the user passes in a value != 0 or 1? also comment has extra underscore

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -0,0 +1,259 @@
...
+        if(0 == center_point_box)
```
kahmed10, line 85: might be more consistent to have the variable first: `if(center_point_box == 0)`

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -0,0 +1,259 @@
...
+        selected_boxes_inside_class.reserve(output_shape.elements());
+
+        auto scores = make_view<float>(args.at(1).get_shape(), args.at(1).cast<float>());
+        shape comp_s{shape::float_type, {batch_num, class_num}};
+        shape_for_each(comp_s, [&](auto idx) {
+            auto bidx = idx[0];
+            auto cidx = idx[1];
+
+            std::size_t score_offset = (bidx * class_num + cidx) * box_num;
+            const float* batch_boxes = args.at(0).cast<float>() + bidx * box_num * 4;
```
pfultz2, line 200: Use an iterator instead of a raw pointer.

src/include/migraphx/op/nonmaxsuppression.hpp
pfultz2: The box should be converted to a canonical format and then we use one way to calculate.

pfultz2: I think the canonical format should be `[y1, x1, y2, x2]`.

src/include/migraphx/op/nonmaxsuppression.hpp
scxiao: Changed to a bool type, so no need  to do this check anymore.

src/include/migraphx/op/nonmaxsuppression.hpp
scxiao: add a function batch_box()

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -0,0 +1,259 @@
...
+#include <migraphx/tensor_view.hpp>
+#include <migraphx/shape_for_each.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct nonmaxsuppression
+{
+    int center_point_box = 0;
```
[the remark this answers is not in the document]
scxiao: changed

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -0,0 +1,259 @@
...
+    struct box_info_ptr
+    {
+        float score{};
+        int64_t index{};
+
+        inline bool operator<(const box_info_ptr& rhs) const
+        {
+            return score < rhs.score || (float_equal(score, rhs.score) && index > rhs.index);
+        }
+    };
```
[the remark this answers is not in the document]
scxiao: the last comparison is ">", so we cannot use the pair unless we overload the < operator for this pair.

PR #969, opened 2021-10-04
Fusion of pointwise operators

pfultz2:
This adds a pass to fuse pointwise operators into one "pointwsie" op that has a submodule which does the calculation.


src/fuse_pointwise.cpp
```diff
@@ -0,0 +1,155 @@
...
+{
+    auto last = std::prev(m->end());
+    if(last->name() == "@return")
+        return last->inputs().front();
+    return last;
+}
+
+std::vector<instruction_ref> append_pointwise_module(instruction_ref ins, instruction_ref output)
+{
+    module_ref pm = ins->module_inputs().at(0);
```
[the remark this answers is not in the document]
pfultz2: The `.at` methods will throw an exception if its empty, so an `assert` shouldn't be needed.

src/fuse_pointwise.cpp
```diff
@@ -0,0 +1,148 @@
...
+        m.replace_instruction(ins, *it);
+        m.move_instruction(*it, ins);
+    }
+}
+
+void fuse_pointwise::apply(module_pass_manager& mpm) const
+{
+    create_pointwise_modules(mpm);
+    mpm.run_pass(dead_code_elimination{});
+    for(int i = 0; i < 8; i++)
```
umangyadav, line 140: Instead of for loop, is there a way to check if there are no changes then break out of the loop ?

pfultz2: Thats a good idea. We probably still want the for loop to have upper limit.

test/fuse_pointwise.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+TEST_CASE(scalar_input)
+{
+    migraphx::shape s{migraphx::shape::float_type, {2, 3}};
+    migraphx::program p1;
+    {
+        auto* mm = p1.get_main_module();
+        auto x   = mm->add_parameter("x", s);
+        auto one = mm->add_literal(1.0f);
+        auto y =
```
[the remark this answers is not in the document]
pfultz2: Yes, because it is left unused.

src/include/migraphx/op/pointwise.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+        if(pm->get_output_shapes().size() != 1)
+            MIGRAPHX_THROW("submodule should have only one output.");
+
+        auto type = pm->get_output_shapes().front().type();
+
+        return shape::from_permutation(type, inputs.front().lens(), find_permutation(inputs));
+    }
+
+    argument compute(const shape& output_shape,
```
kahmed10, line 45: shouldn't we add a ref test to make sure that the compute works?

Issue #971, opened 2021-10-04
SpaceToDepth Operator


causten:
Support the ONNX operator https://github.com/onnx/onnx/blob/master/docs/Operators.md#SpaceToDepth

This is part of the Operator support needed for full Unity Barracuda operators.
https://docs.unity3d.com/Packages/com.unity.barracuda@2.1/manual/SupportedOperators.html#SpaceToDepth

This issue is a child of #824

causten:
Merged

PR #979, opened 2021-10-12
SpaceToDepth operator

umangyadav:
Inverse of DepthToSpace op... 

One difference is that, [onnx spec](https://github.com/onnx/onnx/blob/master/docs/Operators.md#SpaceToDepth) has `mode` attribute for the DepthToSpace but not for the SpaceToDepth. We can add if required. 

Solves #971

### DIFF
--- a/cppcheck.rules
+++ b/cppcheck.rules
@@ -154,6 +154,24 @@
 </rule>
 <rule>
     <tokenlist>normal</tokenlist>
+    <pattern><![CDATA[((?:(?:\w+|<|>|::) )*(?:\w+|>)(?: &|\*)*) (\w) ; \2 = static_cast < \1 > (\([^()]*(?-1)*[^()]*\)) ;]]></pattern>
+    <message>
+        <id>RedundantCast</id>
+        <severity>style</severity>
+        <summary>Static cast is redundant.</summary>
+    </message>
+</rule>
+<rule>
+    <tokenlist>normal</tokenlist>
+    <pattern><![CDATA[auto (\w) ; \1 = static_cast < (?:(?:\w+|<|>|::) )*(?:\w+|>)(?: &|\*)* > (\([^()]*(?-1)*[^()]*\)) ;]]></pattern>
+    <message>
+        <id>RedundantCast</id>
+        <severity>style</severity>
+        <summary>Static cast is redundant.</summary>
+    </message>
+</rule>
+<rule>
+    <tokenlist>normal</tokenlist>
     <pattern><![CDATA[\? (true|false) : (true|false)]]></pattern>
     <message>
         <id>RedundantConditionalOperator</id>


cagery (approved):

src/onnx/parse_spacetodepth.cpp
```diff
@@ -0,0 +1,60 @@
...
+        {
+            blocksize = info.attributes.at("blocksize").i();
+        }
+        if(blocksize < 1)
+        {
+            // blockSize less than 1 would rather result in DepthToSpace instead of SpaceToDepth
+            MIGRAPHX_THROW("SpaceToDepth: blocksize is less than 1");
+        }
+        // calculate dimensions
+        auto res_shape = s.lens(); // {N, C, H, W}
```
scxiao, line 32: I would suggest using the name "res_lens" and "trans_lens" to be more natural.

test/onnx/verify_onnx.cpp
```diff
@@ -63,6 +63,49 @@ TEST_CASE(depthtospace_simple_test)
...
+    pp1["x"]     = migraphx::argument(s_x_1, data_in.data());
+    auto result1 = p1.eval(pp1).back();
+    std::vector<float> result_vector1;
+    result1.visit([&](auto output) { result_vector1.assign(output.begin(), output.end()); });
+    // depth to space
+    auto p2 = migraphx::parse_onnx("depthtospace_simple_test.onnx");
+    p2.compile(migraphx::ref::target{});
+    migraphx::shape s_x_2{migraphx::shape::float_type, {1, 8, 2, 3}};
+    migraphx::parameter_map pp2;
+    pp2["x"]     = migraphx::argument(s_x_2, result_vector1.data());
```
scxiao, line 102: This line can be:

```
pp2["x"]     = result1;
```
So the test looks a little simpler

umangyadav: Done

umangyadav:
> Looks good to me, only concern is openvino has mode implemented[ it looks like](https://docs.openvinotoolkit.org/latest/openvino_docs_ops_movement_SpaceToDepth_1.html) but barracuda documentation shows onnx specs only, so I think we should be fine but might worth double checking with barracuda.

I checked barracuda, it follows onnx spec only.

test/onnx/verify_onnx.cpp
```diff
@@ -99,7 +99,7 @@ TEST_CASE(spacetodepth_depthtospace_test)
     p2.compile(migraphx::ref::target{});
     migraphx::shape s_x_2{migraphx::shape::float_type, {1, 8, 2, 3}};
     migraphx::parameter_map pp2;
-    pp2["x"]     = migraphx::argument(s_x_2, result_vector1.data());
+    pp2["x"]     = result1;
```
scxiao, line 102: Then lines 95 and 96 can be removed.

scxiao (approved):

Issue #905, opened 2021-08-04
DepthToSpace support: DepthToSpace followed by Elementwise op (e.g. add)


hgaspar:
Encoder-decoder pipelines, and ML-based upscaling pipelines employ processing steps where the output of a layer has greater spatial dimensions than the input. Initially, such layers were implemented in terms of a simple resize operation. The more modern approach is to use learnable resizing operations, such as deconvolutions, and convolutions+depth2space operations.

After the resizing operation (be in a simple nearest neighbor resizing), or a fancier deconvolution, or convolution+depth to space operation, a common operation that follows is an elementwise add, which combines the output of the previous layer (from depth to space), with a skip connection.

The present issue pertains only to pipelines that contain the following types of subgraph:

(depth to space -> buff 1)+ elementwise add (buff 0, buff 1), where buff 0 is from a skip connection.

In such cases, if depth to space can be viewed as a reshaping of the input tensor of add, and if buff 1 is only used by that add (a common case), then a beneficial fusion would be to never perform the depth to space explicitly, rather just reshape the corresponding input of the add operation.

Note:
convolution + depth2Space + add can be fused in two different ways:

1. (convolution + depth2Space) + add 
2. convolution + (depth2Space + add)

Those should be numerically equivalent, and timed (either manually, or in an autotuning scheme), to deduce what is the optimal approach, for a given pipeline.

umangyadav:
#987 Solved this one for now. Closing this.

PR #987, opened 2021-10-22
Add tests for the DepthToSpace+Binary pointwise operations fusion

umangyadav:
In migraphx, DepthToSpace (d2s) is implemented as reshape --> transpose --> contiguous --> reshape.

If there is trailing binary pointwise operator after depthToSpace then, migraphx can move binary operator before contiguous and reshape of the depthtospce. 

So, it becomes `reshape-->transpose-->binary_op-->contiguous-->reshape`. 

Explicit contiguous wouldn't be required since binary_op outputs standard shape. So, it becomes `reshape-->transpose-->binary-->reshape`. 

`simplify_reshapes` already has matcher that can do this transformation. This PR adds test for cases like `depthtospace +binary op`. 

solves #905

### DIFF
--- a/test/simplify_reshapes_test.cpp
+++ b/test/simplify_reshapes_test.cpp
@@ -1031,4 +1031,92 @@ TEST_CASE(transpose_contiguous_unsqueeze_unary)
     EXPECT(m1 == m2);
 }
 
+TEST_CASE(find_transpose_contiguous_reshape_binary_packed)
+{
+    migraphx::module m1;
+    {
+        auto x  = m1.add_parameter("x", {migraphx::shape::float_type, {2, 128, 28, 28}});
+        auto w1 = m1.add_literal(
+            migraphx::generate_literal({migraphx::shape::float_type, {256, 128, 1, 1}}));
+        auto conv1 = m1.add_instruction(
+            migraphx::make_op("convolution",
+                              {{"padding", {0, 0}}, {"stride", {1, 1}}, {"dilation", {1, 1}}}),
+            x,
+            w1); // (2, 256, 28, 28)
+        auto w2 = m1.add_literal(
+            migraphx::generate_literal({migraphx::shape::float_type, {512, 256, 1, 1}}));
+        auto conv2 = m1.add_instruction(
+            migraphx::make_op("convolution",
+                              {{"padding", {0, 0}}, {"stride", {2, 2}}, {"dilation", {1, 1}}}),
+            conv1,
+            w2); // (2, 512, 14, 14)
+
+        auto conv2_rsp1 = m1.add_instruction(
+            migraphx::make_op("reshape", {{"dims", {2, 2, 2, 128, 14, 14}}}), conv2);
+        auto conv2_trans = m1.add_instruction(
+            migraphx::make_op("transpose", {{"permutation", {0, 3, 4, 1, 5, 2}}}), conv2_rsp1);
+        auto conv2_cont = m1.add_instruction(migraphx::make_op("contiguous"), conv2_trans);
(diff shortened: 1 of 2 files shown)


umangyadav:
removed the matcher for the depthToSpace+Binary. 

it turns out, that case can already be handled by `find_reshape_cont` which does the required matching and changes on the program. 

there are also two more matchers that  are matched for depthToSpace like test case that are 1. `find_transpose` 2. `find_reshaper`.  I can verify that even though they are matched, but application of it wouldn't result in different program for depthToSpace+binary like cases.

scxiao (approved):

scxiao:
@umangyadav I think you need to change the description of the PR since we do not do the actual optimization anymore.

umangyadav:
> @umangyadav I think you need to change the description of the PR since we do not do the actual optimization anymore.

Did changes.

test/simplify_reshapes_test.cpp
```diff
@@ -1031,4 +1031,91 @@ TEST_CASE(transpose_contiguous_unsqueeze_unary)
...
+        m2.add_instruction(pass_op{}, add_rsp);
+    }
+    EXPECT(m1 == m2);
+}
+
+TEST_CASE(transpose_contiguous_reshape_binary_broadcast)
+{
+    auto create_module = []() {
+        migraphx::module m;
+        migraphx::shape sx{migraphx::shape::float_type, {1, 4, 1}};
```
pfultz2, line 1101: This should just be `migraphx::shape sx{migraphx::shape::float_type, {4}}`

test/simplify_reshapes_test.cpp
```diff
@@ -1031,4 +1031,91 @@ TEST_CASE(transpose_contiguous_unsqueeze_unary)
...
+            m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", {2, 4, 6}}}), x);
+        auto y_trans =
+            m.add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 2, 3, 1}}}), y);
+        auto y_cont = m.add_instruction(migraphx::make_op("contiguous"), y_trans);
+        auto y_rsp = m.add_instruction(migraphx::make_op("reshape", {{"dims", {2, 4, 6}}}), y_cont);
+        auto r     = m.add_instruction(migraphx::make_op("add"), y_rsp, x_brcst);
+        m.add_return({r});
+        return m;
+    };
+    auto m1 = create_module();
```
pfultz2, line 1116: Dont use the lambda function. Instead, create the `module m1`(like in the previous test) and then assign `module m2 = m1` before running the pass.

pfultz2: We used to use lambdas in the tests because the program/module wasn't copyable. This has been fix a while ago, so its much simpler to just copy the module.

test/simplify_reshapes_test.cpp
umangyadav: broadcast can not happen just with {4}.


```
  what():  /home/umayadav/repo/AMDMIGraphX/src/include/migraphx/op/multibroadcast.hpp:54: compute_shape: MULTIBROADCAST: input shape {4} cannot be broadcasted to {2, 4, 6}!
Aborted (core dumped)
```

test/simplify_reshapes_test.cpp
umangyadav: changed.

test/simplify_reshapes_test.cpp
pfultz2: You need to use `broadcast` instead of `multibroadcast`.

umangyadav: Done.

PR #988, opened 2021-10-22
Add fp16 verify to driver

kahmed10:
Allows `--fp16` to be used in the driver to compare the target fp16 result and the ref fp32 result.


src/driver/argument_parser.hpp
```diff
@@ -106,10 +114,22 @@ struct argument_parser
...
+    std::string as_string_value(rank<0>, const T&)
+    {
+        throw std::runtime_error("Can't convert to string");
+    }
+
     template <class T, MIGRAPHX_REQUIRES(not is_multi_value<T>{})>
     std::string as_string_value(const T& x)
     {
-        return to_string(x);
+        return as_string_value(rank<2>{}, x);
```
scxiao, line 132: The max rank above is 1, should we use rank<1> here?

kahmed10: fixed

src/driver/verify.cpp
```diff
@@ -43,24 +51,24 @@ void verify_program(const std::string& name,
                     const program& p,
                     const target& t,
                     compile_options options,
+                    precision quantize,
                     const parameter_map& inputs,
                     double tolerance)
 {
     auto x = run_ref(p, inputs);
```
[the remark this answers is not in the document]
kahmed10: The user should be able to pass in a different tolerance value when comparing to fp32 ref. It might vary depending on the model, so having the user specify the tolerance might be the best option.

scxiao (approved):

src/driver/argument_parser.hpp
```diff
@@ -28,6 +29,13 @@ inline namespace MIGRAPHX_INLINE_NS {
 #define MIGRAPHX_DRIVER_STATIC static
 #endif
 
+enum class precision
+{
+    fp32,
+    fp16,
+    int8
+};
```
pfultz2, line 37: This should go in a different header not in the `argument_parser.hpp`.

pfultz2: Perhaps add a `precision.hpp` header to put this enum in.

pfultz2 (changes requested):

pfultz2 (approved):

cagery (approved):
LGTM

PR #992, opened 2021-10-28
Conditionally enable pointwise fusion

pfultz2:
This enables the pointwise fusions using the `MIGRAPHX_ENABLE_POINTWISE_FUSION` env variable. Its disabled by default since MIOpen fusions need to be refactored.

This also adds a `compile_ops` pass to compile the pointwise modules. All tests except `test_gpu_fast_math` passes with `MIGRAPHX_ENABLE_POINTWISE_FUSION=1` set.


src/include/migraphx/op/pointwise.hpp
```diff
@@ -26,19 +26,17 @@ struct pointwise
...
-                MIGRAPHX_THROW("Mismatch type");
-        }
 
         if(pm->get_output_shapes().size() != 1)
             MIGRAPHX_THROW("submodule should have only one output.");
 
         auto type = pm->get_output_shapes().front().type();
 
+        // Scalar output if all inputs are scalar
+        if(inputs.front().elements() == 1 and
```
kahmed10, line 36: Do we want to add at test for this?

src/targets/gpu/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ add_library(migraphx_gpu
     batch_norm_inference.cpp
     clip.cpp
     code_object_op.cpp
+    compile_ops.cpp
```
umangyadav, line 125: Should we call it compile_pointwise_ops ?

pfultz2: No, this can be used to compile more ops in the future.

src/fuse_pointwise.cpp
```diff
@@ -13,6 +13,8 @@ inline namespace MIGRAPHX_INLINE_NS {
 
 static literal get_scalar(instruction_ref ins)
 {
+    if(ins->name() == "contiguous")
+        return get_scalar(ins->inputs().front());
```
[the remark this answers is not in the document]
pfultz2: For cases like this(such as a scalar or broadcasted shape) we skip const-folding of contiguous as it makes the tensor use more memory(and is inefficient) after lowering. Whereas if we keep the contiguous there, it can be removed if the lowered operator allows broadcasted or scalar shapes as inputs, and the original tensor shape can be preserved.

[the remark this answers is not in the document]
pfultz2: Also if folded the contiguous then we wouldn't know thats its scalar, which we try to embed scalars into the code generation.

src/include/migraphx/op/pointwise.hpp
pfultz2: Done

src/cpp_generator.cpp
```diff
@@ -84,8 +86,11 @@ void cpp_generator::fmap(const std::function<std::string(std::string)>& f) { imp
 std::string cpp_generator::generate_point_op(const operation& op,
                                              const std::vector<std::string>& args)
 {
-    auto v = op.to_value();
-    return interpolate_string(op.attributes()["point_op"].to<std::string>(),
+    auto v          = op.to_value();
+    auto attributes = op.attributes();
+    if(not attributes.contains("point_op"))
+        MIGRAPHX_THROW("op is missing point_op attribute: " + op.name());
+    return interpolate_string(attributes["point_op"].to<std::string>(),
```
scxiao, line 93: The interpolate_string called here is:

```
template <class F>
inline std::string
interpolate_string(const std::string& input, F f, std::string start = "${", std::string end = "}")
{
    std::string result = "";
    result.reserve(input.size());
    auto it = input.begin();
    while(it != input.end())
    {
        auto next_start = std::search(it, input.end(), start.begin(), start.end());
        auto next_end   = std::search(next_start, input.end(), end.begin(), end.end());
        result.append(it, next_start);
        if(next_start == input.end())
            break;
        auto r = f(next_start + start.size(), next_end - end.size() + 1);
        result.append(r.begin(), r.end());
        it = next_end + 1;
    }
    return result;
}

```
I think the next_end-end.size + 1 in the line ```auto r = f(next_start + start.size(), next_end - end.size() + 1);``` does not make sense. The reason is next_end is the start of "end". Move the iterator forward does not make sense.

pfultz2: > Move the iterator forward does not make sense.

Its not moving the iterator forward, its returning the distance.

scxiao: Then what does it mean for "next_end - end.size() + 1"? My understanding is it is the location end.length() characters before the "end".

pfultz2: You right it is moving the iterator by `end.size() + 1`. Let me investigate that more.

pfultz2: So I fixed this and added more tests which found a few more bugs, which I fixed.

src/cpp_generator.cpp
```diff
@@ -120,7 +125,12 @@ std::string cpp_generator::str() const { return impl->fs.str(); }
 cpp_generator::function cpp_generator::generate_module(const module& m)
 {
     function f;
-    f.set_name(m.name()).set_types(m).set_body(
+    auto name = transform_string(m.name(), [](char c) {
+        if(with_char(::isalnum)(c) or c == '_')
+            return c;
+        return '_';
+    });
+    f.set_name(name).set_types(m).set_body(
```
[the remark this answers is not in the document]
pfultz2: Fixed.

pfultz2:
@scxiao @kahmed10 @umangyadav Are you able to review again?

kahmed10 (approved):

umangyadav (approved):

scxiao (approved):

PR #994, opened 2021-10-28
Update Docker to ROCm 4.5 and support Navi on Jenkins

kahmed10:


Jenkinsfile
```diff
@@ -110,6 +112,11 @@ rocmtest clang_debug: rocmnode('vega') { cmake_build ->
         def debug_flags = "-g -O2 -fno-omit-frame-pointer -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers}"
         cmake_build("/opt/rocm/llvm/bin/clang++", "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DMIGRAPHX_ENABLE_GPU=Off -DMIGRAPHX_ENABLE_CPU=On -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}'")
     }
+}, clang_release_navi: rocmnode('navi21') { cmake_build ->
+    stage('HIP Clang Release Navi') {
+        cmake_build("/opt/rocm/llvm/bin/clang++", "-DCMAKE_BUILD_TYPE=release")
+        stash includes: 'build/*.deb', name: 'migraphx-package'
```
pfultz2, line 118: Dont stash the debs. Onnx runtime only runs on vega, right now.

test/msgpack.cpp
```diff
@@ -11,7 +11,7 @@ std::vector<char> msgpack_buffer(const T& src)
     msgpack::pack(buffer, src);
     buffer.seekg(0);
     std::string str = buffer.str();
-    return std::vector<char>(str.data(), str.data() + str.size());
+    return std::vector<char>(str.data(), str.data() + str.size()); // NOLINT
```
pfultz2, line 14: Why does this use `NOLINT`?

kahmed10: I tried rewriting using a brace initializer list, but then I get a tidy error that the local variable `str` should not be used within the initializer list when returning from the function. I had a workaround commit where I just replaced `str` with `buffer.str()`, but then the msgpack test failed.

Jenkinsfile
kahmed10: fixed

umangyadav (approved):

pfultz2 (approved):

PR #995, opened 2021-11-03
Failing fusion plan workaround

turneram:
Add workaround for devices that do not support miopen conv fusions

Fixes the error: `what():  /code/AMDMIGraphX/src/targets/gpu/fuse_ops.cpp:131: compile: Compiling fusion plan failed ` that was occuring on MI200.

### DIFF
--- a/src/targets/gpu/fuse_ops.cpp
+++ b/src/targets/gpu/fuse_ops.cpp
@@ -27,6 +27,7 @@
 #include <migraphx/array.hpp>
 #include <migraphx/op/clip.hpp>
 #include <cmath>
+#include <set>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
@@ -153,9 +154,9 @@ struct fusion
     }
 };
 
-std::vector<std::string> get_supported_archs()
+std::unordered_set<std::string> get_supported_archs()
 {
-    static std::vector<std::string> supported_archs{"gfx900", "gfx906", "gfx908", "gfx1030"};
+    static std::unordered_set<std::string> supported_archs{"gfx900", "gfx906", "gfx908", "gfx1030"};
     return supported_archs;
 }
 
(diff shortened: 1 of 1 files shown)


src/targets/gpu/fuse_ops.cpp
```diff
@@ -152,6 +153,12 @@ struct fusion
     }
 };
 
+std::vector<std::string> get_supported_archs()
+{
+    static std::vector<std::string> supported_archs{"gfx900", "gfx906", "gfx908"};
```
kahmed10, line 158: We also want to add gfx1030.

turneram: Added

kahmed10 (approved):

src/targets/gpu/fuse_ops.cpp
```diff
@@ -152,6 +153,12 @@ struct fusion
     }
 };
 
+std::vector<std::string> get_supported_archs()
```
kahmed10, line 156: Actually instead of a std::vector, can you use a std::set<std::string> instead? And then I believe you can change the logic to use `if contains(supported_archs, device_name)` for L173

src/targets/gpu/fuse_ops.cpp
```diff
@@ -152,6 +154,12 @@ struct fusion
     }
 };
 
+std::unordered_set<std::string> get_supported_archs()
```
pfultz2, line 157: This should return a const ref: `const std::unordered_set<std::string>&`.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -161,6 +169,10 @@ MIGRAPHX_PRED_MATCHER(bias_shape, instruction_ref ins)
 
 MIGRAPHX_PRED_MATCHER(fusable_conv, instruction_ref ins)
 {
+    const auto device_name     = split_string(get_device_name(), ':').front();
+    const auto supported_archs = get_supported_archs();
```
pfultz2, line 173: Remove this variable, and just use `get_supported_archs()` directly in the contains: `contains(get_supported_archs(), device_name)`.

src/targets/gpu/fuse_ops.cpp
turneram: Fixed

src/targets/gpu/fuse_ops.cpp
turneram: Removed

pfultz2 (approved):

kahmed10 (approved):

PR #996, opened 2021-11-04
Test runner match input output using tensor names

scxiao:
Two changes were made:

1. Previous implementation assumes inputs and outputs .pb files are ordered, but it is not the case. So, we should use the name of the tensors in the input/output .pb files to match the input and output in the onnx model. (This change applies to the BERT_Squad model)
2. When parsing a model with dynamic input shape, current implementation uses the default batch_size for the unknown dims, which can cause parsing error for some cases (e.g. mask_rcnn model). The solution is we first read an input to get the shape, then use these shapes to parse the onnx model.


tools/test_runner.py
```diff
@@ -54,36 +54,113 @@ def read_pb_file(filename):
...
     return outputs
 
 
+def model_parameter_names(model_file_name):
+    with open(model_file_name, 'rb') as pfile:
+        data_str = pfile.read()
+        model_proto = onnx.ModelProto()
+        model_proto.ParseFromString(data_str)
+        init_names = set([(i.name) for i in model_proto.graph.initializer])
+        param_names = []
```
kahmed10, line 114: This can  be rewritten with list comprehension:
```suggestion
        param_names = [ input for input in model_proto.graph.input if input.name not in init_names]
```

tools/test_runner.py
```diff
@@ -54,36 +54,113 @@ def read_pb_file(filename):
...
+
+    if len(name_array) < len(shape_array):
+        param_shape_map = {}
+        for i in range(len(param_names)):
+            param_shape_map[param_names[i]] = shape_array[i]
+
+        return param_shape_map
+
+    for name in param_names:
+        if not name in param_shape_map.keys():
```
kahmed10, line 152: Deepcode used to be able to check for this, but you do not need to use keys() anymore when searching for the key:
```suggestion
        if not name in param_shape_map:
```

tools/test_runner.py
```diff
@@ -54,36 +54,113 @@ def read_pb_file(filename):
...
+
+    return param_shape_map
+
+
 def run_one_case(model, param_map):
     # convert np array to model argument
     pp = {}
     for key, val in param_map.items():
-        print("input = {}".format(val))
+        #print("input: {} = {}".format(key, val))
```
kahmed10, line 163: needed?

tools/test_runner.py
```diff
@@ -168,9 +257,12 @@ def main():
 
         # run the model and return outputs
         output_data = run_one_case(model, input_data)
+        # gold_output_data = []
```
kahmed10, line 260: needed?

tools/test_runner.py
```diff
@@ -142,21 +217,34 @@ def main():
...
+    param_shapes = get_input_shapes(sample_case, param_names)
+    for name, dims in param_shapes.items():
+        print("Input: {}, shape: {}".format(name, dims))
+    print()
+
     # read and compile model
-    model = migraphx.parse_onnx(model_path_name)
-    param_names = model.get_parameter_names()
+    model = migraphx.parse_onnx(model_path_name, map_input_dims=param_shapes)
     output_shapes = model.get_output_shapes()
```
kahmed10, line 237: I think this line can be removed since it's not being used anywhere.

tools/test_runner.py
scxiao: fixed

tools/test_runner.py
scxiao: fixed

tools/test_runner.py
scxiao: removed

tools/test_runner.py
scxiao: removed

kahmed10 (approved):

scxiao:
@pfultz2 When you get a chance, could you please take a look at this PR?

pfultz2 (approved):

PR #997, opened 2021-11-05
Turn on gemm unit tests

scxiao:
This PR is to turn on a few gemm unit test with int8 input datatype. Before rocm4.4, int8 input data type requires matrix size to be no less than 4 in rocblas implementation. Because of this limitation, we turned off a few gemm unit tests with int8 input data type. 

This limitation is removed in rocm4.4, so after we upgrade to rocm4.5, we can turn on these unit tests. Also we change to unit test conv_bn_add to adding instructions to module instead of program.


test/verify/test_conv_bn_add.cpp
```diff
@@ -2,44 +2,44 @@
...
+//        auto relu1 = mm->add_instruction(migraphx::make_op("relu"), x);
+//        auto conv1 = mm->add_instruction(migraphx::make_op("convolution"), relu1, w);
+//        auto bn1   = add_bn(*mm, conv1, ochannels, 1);
+//        auto relu2 = mm->add_instruction(migraphx::make_op("relu"), y);
+//        auto conv2 = mm->add_instruction(migraphx::make_op("convolution"), relu2, v);
+//        auto bn2   = add_bn(*mm, conv2, ochannels, 1);
+//        auto sum   = mm->add_instruction(migraphx::make_op("add"), bn1, bn2);
+//        mm->add_instruction(migraphx::make_op("relu"), sum);
+//        return p;
+//    }
```
pfultz2, line 44: Can we uncomment out this test and then just disable it for the "gpu"?

kahmed10 (approved):

test/verify/test_conv_bn_add.cpp
scxiao: fixed

pfultz2 (approved):

scxiao:
@causten I think this PR can be merged, the pending check is passed, it is not displayed correctly.

PR #1001, opened 2021-11-09
Non std shape auto contiguous

scxiao:
This PR is to resolve a problem in parsing the ssd-10 model. 

The problem is, after inserting contiguous in the auto_contiguous pass, standard output shape of some operators becomes non-standard. Then, if the next operator requires standard input shape, an exception is throw. 

For example, if we pass the following model:
  Input (standard shape) -> transpose (transposed) -> softmax (transposed) -> transpose (standard) -> gather. 
It works fine, and no contiguous is required. 

In the auto_contiguous pass, a contiguous is inserted after the first transpose. Then we need to replace the first transpose with the contiguous and recompute all shapes. When it comes to the gather operator, its input is a transposed shape, and an exception is thrown.

The solution is in the recompute_shape() function. If it is called by the auto_contiguous pass and shape of an instruction is changed, and the shape is non_standard, we do not recompute shape of its output. The reason is: since its output shape is non_standard, a contiguous op will be added after the instruction, which will recompute shape for later operators.

### DIFF
--- a/test/auto_contiguous_test.cpp
+++ b/test/auto_contiguous_test.cpp
@@ -101,4 +101,42 @@ TEST_CASE(after_param_broadcast)
     EXPECT(not m.get_output_shapes().back().broadcasted());
 }
 
+TEST_CASE(two_transpose_gather)
+{
+    auto create_module = [] 
+    {
+        migraphx::module m;
+        auto data = m.add_parameter("2x2", {migraphx::shape::float_type, {2, 3, 4, 5}});
+        auto ind = m.add_parameter("ind", {migraphx::shape::float_type, {2, 3}});
+        auto td = m.add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 2, 3, 1}}}), data);
+        auto sd = m.add_instruction(migraphx::make_op("softmax", {{"axis", 2}}), td);
+        auto bd = m.add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 3, 1, 2}}}), sd);
+        auto r = m.add_instruction(migraphx::make_op("gather", {{"axis", 2}}), bd, ind);
+        m.add_return({r});
+
+        return m;
+    };
+
+    auto m = create_module();
+    run_pass(m);
+
+    auto create_cont_module = [] 
+    {
+        migraphx::module m;
+        auto data = m.add_parameter("2x2", {migraphx::shape::float_type, {2, 3, 4, 5}});
+        auto ind = m.add_parameter("ind", {migraphx::shape::float_type, {2, 3}});
+        auto td = m.add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 2, 3, 1}}}), data);
+        auto ctd = m.add_instruction(migraphx::make_op("contiguous"), td);
+        auto sd = m.add_instruction(migraphx::make_op("softmax", {{"axis", 2}}), ctd);
+        auto bd = m.add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 3, 1, 2}}}), sd);
+        auto cbd = m.add_instruction(migraphx::make_op("contiguous"), bd);
+        auto r = m.add_instruction(migraphx::make_op("gather", {{"axis", 2}}), cbd, ind);
+        m.add_return({r});
+
+        return m;
+    };
+
+    EXPECT(m == create_cont_module());
+}
+
 int main(int argc, const char* argv[]) { test::run(argc, argv); }
--- a/test/auto_contiguous_test.cpp
+++ b/test/auto_contiguous_test.cpp
@@ -103,14 +103,15 @@ TEST_CASE(after_param_broadcast)
 
 TEST_CASE(two_transpose_gather)
 {
-    auto create_module = [] 
-    {
+    auto create_module = [] {
         migraphx::module m;
         auto data = m.add_parameter("2x2", {migraphx::shape::float_type, {2, 3, 4, 5}});
-        auto ind = m.add_parameter("ind", {migraphx::shape::float_type, {2, 3}});
-        auto td = m.add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 2, 3, 1}}}), data);
+        auto ind  = m.add_parameter("ind", {migraphx::shape::float_type, {2, 3}});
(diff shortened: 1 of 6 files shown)


pfultz2:
Could we fix this issue by moving backwards in auto_contiguous? This way later instructions that do need contiguous will already have it inserted.

test/auto_contiguous_test.cpp
```diff
@@ -101,4 +101,44 @@ TEST_CASE(after_param_broadcast)
     EXPECT(not m.get_output_shapes().back().broadcasted());
 }
 
+TEST_CASE(two_transpose_gather)
+{
+    auto create_module = [] {
```
pfultz2, line 106: Dont use a lambda function. Instead just use an anonymous scope:

```cpp
migraphx::module m1;
{
    // Add operators
    auto x = m1.add_instruction(...);
}
run_pass(m1);

migraphx::module m2;
{
    // Add operators
    auto x = m2.add_instruction(...);
}
EXPECT(m1 == m2);
```

scxiao:
> Could we fix this issue by moving backwards in auto_contiguous? This way later instructions that do need contiguous will already have it inserted.

Looks like it does not work. For example, for the unit test I added, if go backward, when we check the second transpose, its output shape is standard, so no contiguous is added. Then when it comes to the first transpose, a contiguous is added, and recompute_shape is called, which will cause the problem that input shape of gather is not standard.

pfultz2:
Why does gather require standard shape? It uses the `tensor_view` to index the elements so it should work for any shape.

umangyadav (approved):

test/verify/test_nonstd_gather.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        migraphx::shape s{migraphx::shape::float_type, {3, 3}};
+        migraphx::shape s_indices{migraphx::shape::int32_type, {2, 2}};
+        std::vector<int> indices{1, 1, 0, 2};
+        auto d  = mm->add_parameter("data", s);
+        auto td = mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {1, 0}}}), d);
```
umangyadav, line 17: Have you checked if simplify_reshapes is not removing two transposes ? since they generate a `no_op`.

umangyadav: Even if they are not removed, `auto_contiguous` will add contiguous for both transposes that would make input to gather a standard shape. 

Because, after insertion of `contigous` in `auto_contiguous` shapes are recomputed for all its ouputs. that would make output of second transpose non-standard and `auto_contigous` will add contigous for it. So gather would receive standard shape.

umangyadav, line 17: Wouldn't auto_contiguous make it standard shape for the gather again?

test/verify/test_nonstd_gather.cpp
```diff
@@ -0,0 +1,26 @@
...
+        auto ind = mm->add_literal(migraphx::literal{s_indices, indices});
+        auto tind =
```
umangyadav, line 19: Why does it need to transpose indices as well ?

src/include/migraphx/op/gather.hpp
```diff
@@ -38,7 +38,6 @@ struct gather
 
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(2).standard();
```
pfultz2, line 41: Shouldn't we still have `check_shapes{inputs, *this}.has(2)`?

test/verify/test_nonstd_gather.cpp
pfultz2: This is true the reference implementation runs auto_contiguous so non-standard shape is not really tested. We should probably have a seperate unit test running `eval` without compilng and compare the results using standard shape input and non-standard shape inputs(they should be the same result).

test/verify/test_nonstd_gather.cpp
scxiao: The purpose it to test transposed index input will works for gather...

src/include/migraphx/op/gather.hpp
scxiao: Yes, you are right. we should keep the check for the number of inputs..

test/verify/test_nonstd_gather.cpp
umangyadav: Yes, but since we are just comparing with ref implementation output, it should also work if index is not transposed. I have similar test for the `scatter` which fails if index is not transposed.

test/verify/test_nonstd_gather.cpp
umangyadav: I checked for the CPU as well. It is also using standard shape. 

```
 983 Pass: dead_code_elimination
 984 module: "main"
 985 main:@0 = cpu::preallocate[shape=float_type, {21}, {1},id=main:scratch] -> float_type, {21}, {1}
 986 main:@1 = cpu::literal -> float_type, {2, 2}, {2, 1}
 987 data = @param:data -> float_type, {3, 3}, {3, 1}
 988 main:@2 = transpose[permutation={1, 0}](data) -> float_type, {3, 3}, {1, 3}
 989 main:@3 = load[offset=0,end=36](main:@0) -> float_type, {3, 3}, {3, 1}
 990 main:@4 = dnnl::reorder[post_ops={}](main:@2,main:@3) -> float_type, {3, 3}, {3, 1}
 991 main:@5 = load[offset=36,end=84](main:@0) -> float_type, {3, 2, 2}, {4, 2, 1}
 992 main:@6 = cpu::gather[axis=1](main:@4,main:@1,main:@5) -> float_type, {3, 2, 2}, {4, 2, 1}
 993 main:@7 = @return(main:@6)

```

umangyadav: Because `cpu::gather` requires standard shape.

scxiao: @pfultz2 I removed the compile() function for the unit test I added for the gather operator, so input to the gather operator is non-standard. 

Actually, I am wondering whether we should add a list of unit tests for all operators that contain the compute method without compiling the program to see whether they work correctly.

pfultz2 (approved):

PR #1005, opened 2021-11-11
Handle removing contiguous on operators that use modules

pfultz2:
Currently, eliminate_contiguous will never remove contiguous for operators that use module inputs due to the fact that it doesn't pass the module inputs to `compute_shape`.

* Update to pass the module inputs correctly to `compute_shape`
* Fix the overloads of `compute_shape` so that when passed an empty vector of module inputs it will call the overload without module inputs
* Add tests with contiguous and pointwise module function.
* Move `add_pointwise` function to a seperate header to reuse across different tests


src/pass_manager.cpp
```diff
@@ -82,6 +84,8 @@ module& get_module(module_pass_manager& mpm) { return mpm.get_module(); }
 
 void run_passes(module& mod, const std::vector<pass>& passes, tracer trace)
 {
+    if(enabled(MIGRAPHX_TRACE_PASSES{}))
```
umangyadav, line 87: Don't we already have `MIGRAPHX_TRACE_COMPILE` flag that would instantiate `tracer` with `std::cout` ?

umangyadav: I think you want to dump pass output with `run_passes` without having to call `compile`.  Is that why is it ?

umangyadav (approved):

tools/include/operation.hpp
```diff
@@ -112,77 +112,53 @@ auto normalize_compute_shape_op(rank<2>, const T& x, const std::vector<shape>& i
...
-shape normalize_compute_shape_op(rank<0>,
-                                 const T& x,
-                                 const std::vector<shape>&,
-                                 const std::vector<module_ref>&)
+shape mod_compute_shape_op(rank<0>,
+                           const T& x,
+                           const std::vector<shape>& inputs,
+                           const std::vector<module_ref>& mod_args)
 {
+    if(mod_args.empty())
```
scxiao, line 150: Why not put the if branch here in a different function like other overloaded function?

src/pass_manager.cpp
pfultz2: The unit tests dont run compile.

tools/include/operation.hpp
pfultz2: What do you mean? You can't make a separate overload for this check as this is a runtime check not a compile-time check.

kahmed10 (approved):

scxiao (approved):
@pfultz2 There is one unit test failed, which needs a minor change.

PR #1006, opened 2021-11-16
Remove standard shape requirement for Ops

umangyadav:
Currently in MIGraphX, a number of ops have put unnecessary restriction that input parameters should have standard shape requirement. 

Following are the list of the ops that require **_standard shape_** for the inputs  and my assessment if they really need or not : 

- ArgMax  --> Works without standard shape requirement
- ArgMin --> Works without standard shape requirement
- As_shape --> It has standard shape requirement and uses reshape, not sure how it works and it is not used in any tests.
- Flatten --> Requires standard, because it uses reshape. Flattens into 2D matrix up-to some input axis.
- Gather --> Works without standard shapes
- If_Op --> Input is boolean value, okay to be standard
- Loop --> Inputs are related to trip count, termination condition etc. okay to have standard shapes
- NMS --> Inputs are Box-Shapes, threadshold, scores etc., Okay to have them standard shapes.
- NonZero --> Should work without having standard shape, but fails on GPU on removing the standard shape requirement
- Reshape --> Need to have standard shape
- RoIAlign --> has three inputs 1. Data in NCHW 2. ROI coordinates 3. Batch indices  
-- ROI coordinates and batch indices should be standard shapes, no reason for them to be non-standard. BatchIndices is just a vector. 
-- Input NCHW can potentially be non-standard
- Scatter --> takes three inputs, data, indices and updates, shouldn't require standard shapes, it is just updating the values in the input tensor. Indices and Updates should have same shapes, not sure if there could be cases where they could be non-standard.
- Squeeze --> Shouldn't need standard shapes, it removes axis with shape `1`. 
- TopK --> Should work without having the standard requirement, extract top K elements. Inputs are just value of `k` and `axis`. 
- Unsqueeze --> Shouldn't require standard shapes, it adds an axis


src/include/migraphx/check_shapes.hpp
```diff
@@ -165,6 +165,13 @@ struct check_shapes
         return *this;
     }
 
+    const check_shapes& not_tuple_type() const
```
pfultz2, line 168: I would prefer to have a `allow_tuple_type` instead since the majority of operators dont allow tuples. This might require a `bool` flag that is checked in the destructor. Maybe we can put this in a seperate PR.

scxiao:
@umangyadav Overall, I think the changes should be like:
1) Remove the standard() shape requirement in ops that can be removed. For the corresponding GPU implementation, if necessary, added the standard shape requirement.
2) For unit tests, for the ops whose standard shape check is removed, added a ref unit test with input non-standard shape, DO NOT compile the program, call the eval() function to get an output. Make a copy of the program, compile the program, call eval(), check the two outputs to ensure they are the same.

src/targets/gpu/device/scatter.cpp
```diff
@@ -15,20 +15,22 @@ argument scatter(
...
             const auto* upd_ptr     = device_cast(update.data());
             const auto* indices_ptr = device_cast(indices.data());
             gs_launch(stream, inds.elements())([=](auto i) __device__ {
-                auto out_idx    = s1.multi(i);
-                auto index      = indices_ptr[i];
-                index           = index < 0 ? index + axis_dim_size : index;
-                out_idx[axis]   = index;
-                output[out_idx] = upd_ptr[i];
+                auto out_idx    = inds1.multi(i);
+                auto upd_idx    = upds1.index(out_idx);
```
pfultz2, line 29: just do `auto upd_idx = out_idx`.

src/targets/gpu/device/scatter.cpp
```diff
@@ -15,20 +15,22 @@ argument scatter(
...
+                auto idx        = indices_ptr[i];
+                idx             = idx < 0 ? idx + axis_dim_size : idx;
+                out_idx[axis]   = idx;
+                output[out_idx] = upd_ptr[upd_idx];
```
pfultz2, line 33: Just do `output[out_idx] = update[upd_idx]`.

src/targets/gpu/device/scatter.cpp
```diff
@@ -15,20 +15,22 @@ argument scatter(
 {
     auto ds            = arg0.get_shape();
     auto inds          = arg1.get_shape();
+    auto upds          = arg2.get_shape();
     auto axis_dim_size = ds.lens()[axis];
-    hip_visit_all(result, arg0, inds)([&](auto output, auto data, auto s1) {
+    hip_visit_all(result, arg0, inds, upds)([&](auto output, auto data, auto inds1, auto upds1) {
```
pfultz2, line 20: I am not sure why we pass `inds` and `upds` shapes. We can get the shapes using `indices.get_shape()` and `update.get_shape()`.

src/targets/gpu/device/scatter.cpp
```diff
@@ -15,20 +15,22 @@ argument scatter(
...
         hip_visit_all(arg1, arg2)([&](auto indices, auto update) {
             const auto* upd_ptr     = device_cast(update.data());
             const auto* indices_ptr = device_cast(indices.data());
             gs_launch(stream, inds.elements())([=](auto i) __device__ {
-                auto out_idx    = s1.multi(i);
-                auto index      = indices_ptr[i];
-                index           = index < 0 ? index + axis_dim_size : index;
-                out_idx[axis]   = index;
-                output[out_idx] = upd_ptr[i];
+                auto out_idx    = inds1.multi(i);
```
pfultz2, line 28: Just do `auto out_idx = indices.get_shape().multi(i)`.

src/include/migraphx/op/scatter.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef MIGRAPHX_GUARD_OPERATORS_SCATTER_HPP
 #define MIGRAPHX_GUARD_OPERATORS_SCATTER_HPP
 
+#include "migraphx/argument.hpp"
```
pfultz2, line 4: Use angle brackets. And this be moved below with the other migraphx includes.

src/targets/gpu/device/scatter.cpp
umangyadav: These changes are triggering a series of Compile-time errors.

src/targets/gpu/device/scatter.cpp
umangyadav: These changes are triggering a series of Compile-time errors.

src/targets/gpu/device/scatter.cpp
umangyadav: These changes are triggering a series of Compile-time errors.

umangyadav:
> @umangyadav Overall, I think the changes should be like:
> 
> 1. Remove the standard() shape requirement in ops that can be removed. For the corresponding GPU implementation, if necessary, added the standard shape requirement.
> 2. For unit tests, for the ops whose standard shape check is removed, added a ref unit test with input non-standard shape, DO NOT compile the program, call the eval() function to get an output. Make a copy of the program, compile the program, call eval(), check the two outputs to ensure they are the same.

@scxiao  
1. Yes, I have made changes in that manner. 
2. I've added ref tests comparing result with the program that is not compiled. 

Also, i've added verify tests wherever possible.

test/ref_ops_test.cpp
```diff
@@ -4662,6 +4825,23 @@ TEST_CASE(unsqueeze_test)
...
+    auto l0 = mm->add_literal(migraphx::literal{s1, data});
+    auto l0_trans =
+        mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {2, 0, 1}}}), l0);
+    mm->add_instruction(migraphx::make_op("unsqueeze", {{"axes", {2}}}), l0_trans);
+    auto p_uncompiled = p;
+    p.compile(migraphx::ref::target{});
+    auto result          = p.eval({}).back();
+    auto expected_result = p_uncompiled.eval({}).back();
+    EXPECT(result.get_shape() == expected_result.get_shape());
+}
```
pfultz2, line 4843: Can these nonstd tests be moved to a seperate .cpp file?

src/targets/gpu/device/scatter.cpp
pfultz2: Whats the compile error?

umangyadav: Following is the snippet : 

```
774 /home/umayadav/repo/AMDMIGraphX/src/targets/gpu/device/include/migraphx/gpu/device/shape.hpp:36:41: note: candidate function not viable: no known conversion from 'hip_array<[...], 5>' t    o 'hip_array<[...], 1U aka 1>' for 1st argument
775     MIGRAPHX_DEVICE_CONSTEXPR index_int index(hip_index x) const { return x.dot(strides); }
776                                         ^
777 /home/umayadav/repo/AMDMIGraphX/src/targets/gpu/device/include/migraphx/gpu/device/shape.hpp:38:41: note: candidate function not viable: no known conversion from 'migraphx::gpu::device:    :hip_array<unsigned int, 5>' to 'std::initializer_list<index_int>' (aka 'initializer_list<unsigned int>') for 1st argument
778     MIGRAPHX_DEVICE_CONSTEXPR index_int index(std::initializer_list<index_int> x) const
779                                         ^
780 /home/umayadav/repo/AMDMIGraphX/src/targets/gpu/device/include/migraphx/gpu/device/shape.hpp:46:41: note: candidate function not viable: no known conversion from 'migraphx::gpu::device:    :hip_array<unsigned int, 5>' to 'migraphx::gpu::device::index_int' (aka 'unsigned int') for 1st argument
781     MIGRAPHX_DEVICE_CONSTEXPR index_int index(index_int i) const

```
It happens when i do `output[out_idx] = update[upd_idx]`.   during the tensor-view of `update[upd_idx]` it calls into `index()` and then errors about no matching function.

test/ref_ops_test.cpp
umangyadav: Moved into a new file.

src/include/migraphx/check_shapes.hpp
umangyadav: Removed it for now.

src/targets/gpu/device/scatter.cpp
scxiao: @umangyadav You can refer the gpu/device/where.cpp file to use the if constexpr to see whether it can be fixed.

pfultz2: I see, there are two visits which causes a rank mismatch. You can use `if constexpr` to check if the rank of the shapes match, something like: `if constexpr(update.get_shape().size() == output.get_shape().size())`. This can go before `gs_launch`.

umangyadav: Thanks. I've made changes.

src/include/migraphx/op/roialign.hpp
```diff
@@ -42,7 +42,8 @@ struct roialign
 
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(3).standard();
+        check_shapes{inputs, *this}.has(3);
+        check_shapes{{inputs.at(1), inputs.at(2)}, *this}.standard();
```
scxiao, line 46: It should be inputs[0] and inputs[2] to be standard.

scxiao: Line 201 to 203, can be changed to
                std::array<float, 2> roi_starts = {
                    static_cast<float>(roi(n, 1) * spatial_scale),
                    static_cast<float>(roi(n, 0) * spatial_scale)};

src/include/migraphx/op/scatter.hpp
```diff
@@ -37,25 +38,28 @@ struct scatter
...
     {
         argument result{output_shape};
+        const shape s_output_shape{output_shape.type(), output_shape.lens()};
         // max dimension in axis
         auto axis_dim_size = output_shape.lens()[axis];
         visit_all(result, args[0], args[2])([&](auto output, auto data, auto update) {
             std::copy(data.begin(), data.end(), output.begin());
             args[1].visit([&](auto indices) {
                 auto ind_s = indices.get_shape();
+                const shape std_ind_s{ind_s.type(), ind_s.lens()};
```
scxiao, line 56: ind_s in the shape_for_each should use the std_ind_s.
All the xxxx[s.indx(idx)] can be changed to xxxx(idx.begin(), idx.end())

src/include/migraphx/op/scatter.hpp
```diff
@@ -37,25 +38,28 @@ struct scatter
...
-        check_shapes{inputs, *this}.has(3).standard();
+        check_shapes{inputs, *this}.has(3);
+        check_shapes{{inputs.at(1), inputs.at(2)}, *this}.same_dims();
         return inputs.front();
     }
 
     argument compute(const shape& output_shape, std::vector<argument> args) const
     {
         argument result{output_shape};
+        const shape s_output_shape{output_shape.type(), output_shape.lens()};
```
scxiao, line 49: This can be removed.

src/targets/gpu/nonzero.cpp
```diff
@@ -8,6 +8,7 @@ namespace gpu {
 
 shape hip_nonzero::compute_shape(std::vector<shape> inputs) const
 {
+    check_shapes({inputs.front()}, *this).has(1).standard();
```
scxiao, line 11: The device function will have problem if input is a broadcasted shape for the nonzero gpu implementation.

test/verify/test_arg_ops.cpp
```diff
@@ -2,34 +2,54 @@
...
 #include <migraphx/program.hpp>
 #include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
 #include <migraphx/op/argmax.hpp>
 #include <migraphx/op/argmin.hpp>
 
-template <class T, int Axis>
-struct test_arg_ops : verify_program<test_arg_ops<T, Axis>>
+template <class T, int Axis, bool Transpose>
+struct test_arg_ops : verify_program<test_arg_ops<T, Axis, Transpose>>
```
scxiao, line 10: For the operators that are changed, I would suggested added a case with broadcasted input shape test case.
Maybe another case of sliced input shape.

src/include/migraphx/op/roialign.hpp
umangyadav: > 
From what i understand,
input[0] is input tensor.  input[1] is ROI co-ordinates and input[2] is batch indices.  So, why input[0] need to be standard ?

scxiao: If you look at the line 

```
               const auto bottom_data   = x.begin();
```
It means args[0] is read according to the offset, so the input shape should be standard.
But for args[1], it is access is read from the tensor_view [] operator, which can handle non-standard shape.

scxiao:
@umangyadav Is it ready for review?

scxiao:
@umangyadav I think we have several unit tests failed with this change, could you please fix it?

umangyadav:
This PR involves multiple different ops that are causing failures. I've opened another PR with limited Ops changes in #1042. Closing this PR for now.

PR #1007, opened 2021-11-16
Parallel compilation

pfultz2:


kahmed10:
Can we add a test case that would ensure that the instructions after compilation are in the correct order?

umangyadav (approved):

kahmed10 (approved):

PR #1008, opened 2021-11-17
Fix fusable_conv whitespace bug

turneram:
Fix whitespace bug in fusable_conv matcher and add unit test


causten:
will this be needed in rocm 4.5 too?

kahmed10:
> will this be needed in rocm 4.5 too?

Currently it's faster to not fuse. However, once MIOpen tuned kernels are in, we will definitely want to reenable fusions. That should be after 4.5 timeframe.

pfultz2 (approved):

kahmed10 (approved):

PR #1009, opened 2021-11-20
ROCm package file/folder reorganization(SWDEV-291455) 

arvindcheru:
Summary of the Changes:
- HIP Path updated (to rocmpath as per folder reorg structure)
- Updated header inclusions (for rocblas - <compnm>/*.h)


examples/migraphx/cpp_parse_load_save/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ project (PLS)
 set (CMAKE_CXX_STANDARD 14)
 set (EXAMPLE parse_load_save)
 
-list (APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)
+list (APPEND CMAKE_PREFIX_PATH /opt/rocm)
```
pfultz2, line 7: These paths still need to be there so this can build with older rocm versions.

src/targets/gpu/CMakeLists.txt
```diff
@@ -277,7 +277,6 @@ else()
         PATH_SUFFIXES bin
         HINTS ${CMAKE_CXX_COMPILER_PATH}
         PATHS
-            /opt/rocm/hip
```
pfultz2, line 280: This is for hcc, I wouldn't worry about removing it.

src/targets/gpu/include/migraphx/gpu/rocblas.hpp
```diff
@@ -3,7 +3,7 @@
 
 #include <migraphx/manage_ptr.hpp>
 #include <migraphx/config.hpp>
-#include <rocblas.h>
+#include <rocblas/rocblas.h>
```
pfultz2, line 6: You need to check that there is a rocblas/rocblas.h header before including it and then fallback to using rocblas.h when the header is missing.

causten:
Why did this get closed?  The code never merged so did you mean to close it?

PR #1010, opened 2021-11-21
Fix pointwise compile error with half sqrt

pfultz2:


kahmed10 (approved):

pfultz2:
@causten I think this can be merged now.

PR #1014, opened 2021-11-22
Softmax perf optimization

scxiao:
This PR is to optimize the softmax on GPU, three changes are made:
1) Change the number of threads in a block from 256 to 128
2) Increase the max number of blocks in the kernel from 256 to 1M. 
3) For the case that the axis is the last dimension, we removed the computation of index since it is not required.

With these change, we can get about 2x speedup compared to the develop branch for the softmax op used in the BertSquad model.


src/targets/gpu/device/softmax.cpp
```diff
@@ -20,34 +20,58 @@ void softmax(hipStream_t stream, const argument& result, const argument& arg, in
     migraphx::shape batch_shape{result.get_shape().type(), batch_lens};
 
     hip_visit_all(result, arg, batch_shape)([&](auto output, auto input, auto batch) {
-        const index_int max_block_size = 256;
+        const index_int max_block_size = 128;
```
pfultz2, line 23: This may not be a good choice when the reduction axis is very large. If this works better in other models that use softmax(such as resnet50 and inception) then we can keep it for now and then improve these choices when we convert this to the jit model.

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
```diff
@@ -76,7 +76,7 @@ MIGRAPHX_DEVICE_CONSTEXPR auto gs_invoke(F&& f, index_int i, index) -> decltype(
 inline auto gs_launch(hipStream_t stream, index_int n, index_int local = 1024)
 {
     index_int groups  = (n + local - 1) / local;
-    index_int nglobal = std::min<index_int>(256, groups) * local;
+    index_int nglobal = std::min<index_int>(1048576, groups) * local;
```
pfultz2, line 79: Was this tested on vega20? This also affects all kernels not just softmax. So how does this affect the overall performance of models(not just bert)?

scxiao:
@pfultz2 When you get a chance, could you please take a look at it again?

pfultz2 (approved):

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
kahmed10: Can we change this value to something more generic like 1,000,000 instead of a random number?

kahmed10:
Works well on gfx1030.

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
scxiao: changed to 1 billion.

scxiao: @kahmed10 Do you have other comments for this PR?

kahmed10 (approved):

PR #1019, opened 2021-12-02
Handle miopen fusions when using pointwise fusions

pfultz2:


src/targets/gpu/fuse_ops.cpp
```diff
@@ -596,7 +708,8 @@ struct miopen_conv_bias
         f    = fusion(inputs[0]);
         conv = f.create_conv(op, inputs[1]);
         bias = f.create_bias(inputs[3]);
-        f.compile(ctx);
+        if(not f.compile(ctx))
+            MIGRAPHX_THROW("Failed to compile fusion plan");
```
kahmed10, line 712: Should we specify which fusion we are attempting here so that it is more helpful for debugging? Such as convolution attributes, input and filter size, etc

pfultz2: This is just to keep the legacy behavior. This will be removed when we switch to pointwise fusions as this will have a more graceful fallback rather than throwing an error.

kahmed10 (approved):

scxiao (commented):
Do we have unit test for this optimization?

umangyadav (approved):

scxiao (approved):

PR #1020, opened 2021-12-02
Fuse convert ops

pfultz2:


scxiao (commented):
Do we need a unit test for this change?

kahmed10 (approved):

pfultz2:
> Do we need a unit test for this change?

It is already covered by our verification tests when using `MIGRAPHX_ENABLE_POINTWISE_FUSION=1`.

scxiao (approved):

PR #1041, opened 2022-01-11
Add env variable to dump tests to a file

pfultz2:
This will make it easier to run the test in the driver.


test/verify/run_verify.cpp
```diff
@@ -125,6 +127,8 @@ void run_verify::verify(const std::string& name, const migraphx::program& p) con
     using result_future =
         std::future<std::pair<migraphx::program, std::vector<migraphx::argument>>>;
     auto_print::set_terminate_handler(name);
+    if(migraphx::enabled(MIGRAPHX_DUMP_TEST{}))
+        migraphx::save(p, name + ".mx");
```
kahmed10, line 131: Any particular reason to use `.mx` as the extension? In test/serialize_program.cpp, we are using `.dat`.

pfultz2: It seems more descriptive than using `.dat`. Perhaps, it would be better to use `.mxr`(for migraphx IR) which is the name MLIR uses for migraphx's dialect.

kahmed10: Then can we change it to .mxr for all serialized migraphx programs?

scxiao: I think this is the binary format (msgpack), right? I checked other unit test, we just use the extension msgpack to indicate it is a msgpack file. Also, in the driver, we use the extension "binary", maybe we can unify all these. And it helps use the driver to load the dumped program.

scxiao: Can we also dump the compiled programs? since it could be a bug from compiling the program.

scxiao: @pfultz2 @kahmed10 So we agree to create another PR to unify the extension for the msgpack format, right? maybe we can use the extension "mpk"?

scxiao (approved):

kahmed10 (approved):

test/verify/run_verify.cpp
pfultz2: > maybe we can use the extension "mpk"?

I would prefer the extension "mxr".

PR #1042, opened 2022-01-11
Remove Standard Shape requirement for ArgOps

umangyadav:
This is the same as #1006  but only includes changes for the arg ops i.e. argmax, argmin. 

All previous review comments have been incorporated in this PR.

### DIFF
--- a/test/verify/test_arg_ops.cpp
+++ b/test/verify/test_arg_ops.cpp
@@ -2,34 +2,92 @@
 #include "verify_program.hpp"
 #include <migraphx/program.hpp>
 #include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
 #include <migraphx/op/argmax.hpp>
 #include <migraphx/op/argmin.hpp>
 
-template <class T, int Axis>
-struct test_arg_ops : verify_program<test_arg_ops<T, Axis>>
+template <class T, int Axis, int NonStdShape>
+struct test_arg_ops : verify_program<test_arg_ops<T, Axis, NonStdShape>>
 {
     migraphx::program create_program() const
     {
         migraphx::program p;
         auto* mm = p.get_main_module();
-        migraphx::shape s{migraphx::shape::float_type, {2, 3, 4, 1025}};
+        migraphx::shape s{migraphx::shape::float_type, {2, 1, 4, 1025}};
         auto param = mm->add_parameter("data", s);
+        switch(NonStdShape)
+        {
+        case 0:
+            param = mm->add_instruction(
+                migraphx::make_op("transpose", {{"permutation", {0, 2, 3, 1}}}), param);
+            break;
+        case 1:
+            param = mm->add_instruction(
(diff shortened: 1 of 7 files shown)


src/include/migraphx/op/scatter.hpp
```diff
@@ -37,25 +38,28 @@ struct scatter
...
 
     argument compute(const shape& output_shape, std::vector<argument> args) const
     {
         argument result{output_shape};
+        const shape s_output_shape{output_shape.type(), output_shape.lens()};
         // max dimension in axis
         auto axis_dim_size = output_shape.lens()[axis];
         visit_all(result, args[0], args[2])([&](auto output, auto data, auto update) {
             std::copy(data.begin(), data.end(), output.begin());
             args[1].visit([&](auto indices) {
```
scxiao, line 54: the std::copy() will not work, either since data can be nonstd shape, while output will always be standard.

scxiao: Do something like:

```
            shape_for_each(output_shape, [&](const auto& idx) {
                output(idx.begin(), idx.end()) = data(idx.begin(), idx.end());
            });
```

umangyadav:
After discussing with @scxiao ,  We've come to an agreement that it would be better to make changes per Operation and assess the benefits of allowing nonstd shape per Op.   We are considering "transpose", "broadcast" and "slice" as nonstd shapes. 

Reasons are : 
1) For example, scatter op can support transpose and slice as non-std, but the broadcast requires additional changes in how we compute output shape. 
2) Therefore, the template of code changes is not similar across all the ops, which would make reviewing difficult
3) Difficult to debug tests failure. 
4) Assess the benefit of non-std shape per Op. 

 For that matter, I've futher reduced the scope of this PR for just ArgOps i.e. argmax and argmin.

test/verify/test_arg_ops.cpp
```diff
@@ -2,34 +2,92 @@
...
-template struct test_arg_ops<migraphx::op::argmax, -1>;
-template struct test_arg_ops<migraphx::op::argmax, -2>;
-
-template struct test_arg_ops<migraphx::op::argmin, 0>;
-template struct test_arg_ops<migraphx::op::argmin, 1>;
-template struct test_arg_ops<migraphx::op::argmin, 2>;
-template struct test_arg_ops<migraphx::op::argmin, 3>;
-template struct test_arg_ops<migraphx::op::argmin, -3>;
-template struct test_arg_ops<migraphx::op::argmin, -4>;
+template struct test_arg_ops<migraphx::op::argmax, 0, 1>;
```
scxiao, line 39: The nonstdshape value should be 0, 1, 2, 3 according to the switch() above, I think.

umangyadav: Yes, thanks, made the change.

scxiao (approved):

turneram (approved):

PR #1050, opened 2022-01-17
Fix hang with CSE pass when using submodules

kahmed10:
Added an example test case that causes a hang in CSE. The workaround is to currently skip the replacement if any of the output instructions don't belong to the submodule.


src/eliminate_common_subexpression.cpp
```diff
@@ -30,9 +30,15 @@ void cse_range(module& p, Range&& r)
                 continue;
             if(*eq != *ins)
                 continue;
+            auto outputs = eq->outputs();
+            for(auto output : outputs)
+            {
+                if(not p.has_instruction(output))
+                    continue;
```
pfultz2, line 37: This loop looks like it doesn't do anything.

kahmed10: This skips over the replacement of instructions within submodules since it checks if the current module contains the output instruction or not.

pfultz2: > This skips over the replacement of instructions

How? After the loop finishes it will call `replace_instruction` regardless of what `p.has_instruction(output)` returns.

kahmed10: I see what you mean, I originally intended to skip the outer for loop, but this is not happening in the above change. For some reason the reordering of getting the outputs before calling replace is fixing the issue.

scxiao (approved):

test/eliminate_common_subexpression_test.cpp
```diff
@@ -142,4 +148,63 @@ TEST_CASE(cse_test_literal)
...
+        auto in_v2 = body2->add_parameter("#loop_module_in_2", s);
+        auto l2    = body2->add_literal(migraphx::literal(si, vd));
+        auto ad2   = body2->add_instruction(migraphx::make_op("add"), l2, l2);
+        auto val2  = body2->add_instruction(migraphx::make_op("add"), in_v2, ad2);
+        auto cond3 = body2->add_instruction(
+            migraphx::make_op("convert", {{"target_type", migraphx::shape::bool_type}}), b1);
+        body2->add_return({cond3, val2, val2});
+
+        auto rl1 = mm->add_instruction(
+            migraphx::make_op("loop", {{"max_iterations", 1}}), {in_cond, in_val}, {body1});
```
pfultz2, line 193: Can the test be written to use `if` instead of `loop`? It seems loop is lot more complicated, but maybe not since there is only one submodule.

test/eliminate_common_subexpression_test.cpp
```diff
@@ -142,4 +148,63 @@ TEST_CASE(cse_test_literal)
...
+        auto rl2 = mm->add_instruction(
+            migraphx::make_op("loop", {{"max_iterations", 1}}), {in_cond, in_val}, {body2});
+        auto r0 = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), rl1);
+        auto r1 = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 1}}), rl1);
+        auto r2 = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), rl2);
+        auto r3 = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 1}}), rl2);
+
+        mm->add_return({r0, r1, r2, r3});
```
pfultz2, line 201: Why can't you just do `mm->add_return({rl1, rl2})` instead of `get_tuple_elem`?

test/eliminate_common_subexpression_test.cpp
```diff
@@ -142,4 +148,63 @@ TEST_CASE(cse_test_literal)
...
+        body2->add_parameter("#loop_module_in_1", sc);
+        auto in_v2 = body2->add_parameter("#loop_module_in_2", s);
+        auto l2    = body2->add_literal(migraphx::literal(si, vd));
+        auto ad2   = body2->add_instruction(migraphx::make_op("add"), l2, l2);
+        auto val2  = body2->add_instruction(migraphx::make_op("add"), in_v2, ad2);
+        auto cond3 = body2->add_instruction(
+            migraphx::make_op("convert", {{"target_type", migraphx::shape::bool_type}}), b1);
+        body2->add_return({cond3, val2, val2});
+
+        auto rl1 = mm->add_instruction(
```
pfultz2, line 192: Could these variables be named better like `loop1`, `loop2`?

test/eliminate_common_subexpression_test.cpp
kahmed10: I initially saw the cse hang behavior from a model with loop in it, so I just decided with making a test from one of our other loop submodule tests. There's only one submodule per loop, so it should be fine.

test/eliminate_common_subexpression_test.cpp
kahmed10: fixed

test/eliminate_common_subexpression_test.cpp
kahmed10: fixed

pfultz2 (approved):

PR #1056, opened 2022-01-20
Tests to discover where we are using too much space in github

causten:
a bunch of prints to dump file system size at various points in the github runner


causten:
this was for debugging only.

PR #1059, opened 2022-01-24
Keep std shape

scxiao:
This PR is the resolve two problems in the issue#999, i.e., non_standard_shape input to reshape and reduce_mean.
Three fixes:
1) Any operator that has a standard shape requirement will add a contiguous input for its input.
2) Eliminate_contiguous, when computing whether a contiguous can be removed, we should use all the updated args, not just the one that is being checked.
3) In two optimization in the simplify_reshape, we remove the contiguous in the reshaper name list, since eliminate_contiguous will remove the contiguous if it can be removed. 

the solution is add an attribute to the operator that requires standard input shape, then in the auto_contiguous pass, add a contiguous to every input of such operators.


pfultz2:
Could we instead add an attribute to operators to say it needs input to be standard, and then in auto_contiguous insert contiguous for those operators before inserting contiguous everywhere else?

src/onnx/onnx_parser.cpp
```diff
@@ -70,7 +70,7 @@ static literal from_repeated(shape::type_t t, const T& r)
 
 instruction_ref onnx_parser::node_info::make_contiguous(instruction_ref ins) const
 {
-    if(ins->get_shape().standard())
+    if(ins->name() == "contiguous")
```
pfultz2, line 73: This doesn't seem like this would really fix the issue. Users could still manually build a custom program without a contiguous before reshape, but auto_contiguous could still insert a contiguous that will cause this error. This seems like this should be fixed in auto_contiguous.

src/simplify_reshapes.cpp
```diff
@@ -26,7 +26,6 @@ const auto& reshaper_names()
     static const std::unordered_set<std::string> names = {
         "flatten",
         "reshape",
-        "contiguous",
```
pfultz2, line 29: We should still remove redundant contiguous in simplify_reshapes, but only when this doesn't precede an operator that requires standard.

kahmed10:
I don't see any new tests being written that would otherwise need to use the proposed changes.

scxiao:
@pfultz2 I added an attribute for standard input shape, can you take a look again?

src/auto_contiguous.cpp
```diff
@@ -10,8 +10,35 @@ inline namespace MIGRAPHX_INLINE_NS {
 
 void auto_contiguous::apply(module& p) const
 {
+    std::string key = "standard_input_shape";
```
kahmed10, line 13: Should we rename this variable to `std_shape`?

kahmed10 (approved):

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -66,7 +66,7 @@ struct reduce_op : op_name<Derived>
     {
         value normalize;
         normalize["axes"] = value::array{normalize_attribute::include_min};
-        return {{"normalize_axes", normalize}};
+        return {{"normalize_axes", normalize}, {"std_shape", true}};
```
pfultz2, line 69: reduce doesn't require standard shape.

src/include/migraphx/op/reshape.hpp
```diff
@@ -26,6 +27,8 @@ struct reshape
         return pack(f(self.dims, "dims"));
     }
 
+    value attributes() const { return {{"std_shape", true}}; }
```
pfultz2, line 30: Maybe the attribute should be called `require_std_shape`.

src/auto_contiguous.cpp
```diff
@@ -10,8 +10,35 @@ inline namespace MIGRAPHX_INLINE_NS {
 
 void auto_contiguous::apply(module& p) const
 {
+    std::string key = "std_shape";
+    for(auto ins : reverse_iterator_for(p))
+    {
+        auto&& attr = ins->get_operator().attributes();
+        if((attr.contains(key) and attr.at(key).to<bool>()))
```
pfultz2, line 17: This can be written as `if (attr.get(key, false))`

src/include/migraphx/op/reduce_op.hpp
scxiao: fixed

src/include/migraphx/op/reshape.hpp
scxiao: fixed

scxiao:
@pfultz2 Can you take a look at this PR again?

src/targets/gpu/target.cpp
```diff
@@ -93,10 +93,10 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         dead_code_elimination{},
         eliminate_common_subexpression{},
         dead_code_elimination{},
+        auto_contiguous{},
         simplify_algebra{},
         simplify_reshapes{},
         simplify_algebra{},
-        auto_contiguous{},
```
pfultz2, line 99: Why is this moved? I think contiguous could hinder some optimizations.

scxiao: I moved the adding contiguous to parse onnx. Reverted this change.

pfultz2 (approved):

PR #1060, opened 2022-01-24
Parse upsample

scxiao:
This is for another part in the issue#999, not linear mode for the Upsample operator, the solution is just use the parse_resize function to parse the upsample operator. Actually, upsample is just part of the resize, it is just a subset of the resize operator.


kahmed10 (approved):

turneram (approved):

scxiao:
> Can you add a onnx test? And is there a test to enable for our onnx backend tests as well?

Added.

pfultz2 (approved):

PR #1061, opened 2022-01-24
Add python type annotations to api.py

pfultz2:
This will also check the types using mypy on the CI.


tools/generate.sh
```diff
@@ -1,9 +1,16 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 SRC_DIR=$DIR/../src
-ls -1 $DIR/include/ | xargs -n 1 -P $(nproc) -I{} -t bash -c "python3.6 $DIR/te.py $DIR/include/{} | clang-format-5.0 -style=file > $SRC_DIR/include/migraphx/{}"
+PYTHON=python3
+if type -p python3.6 > /dev/null ; then
```
kahmed10, line 4: Can't we just do something like $PYTHON --version, then parse to make sure it is at least 3.6?

pfultz2: Python 3 might not be mapped to 3.6 or 3.8. Also, version comparisons are not simple to do in bash.

kahmed10: But currently if python3 is not mapped to 3.6 or 3.8, then we proceed with executing anyways. Are we wanting to version check?

pfultz2: This is just to make it work on ubuntu 20.04 which doesn't have python3.6.

kahmed10: Yes, previously this script was pinned to python3.6 and ubuntu 20.04 comes with 3.8 by default, so it would fail. But it seems that we want 3.6 minimum and anything greater than that should work? Otherwise, if a user has a version < 3.6, then currently the bash script would still try to execute the python commands.

There is a way to use `sort` to do this that might be better:
```
[ "$( (echo "${MIN_VERSION}" ; echo "${CURR_VERSION}") | sort -V | head -1)" == "${MIN_VERSION}" ]
```

tools/api.py
```diff
@@ -23,65 +24,65 @@ class Template(string.Template):
...
 
-    def is_reference(self):
+    def is_reference(self) -> bool:
         return self.name.endswith('&')
 
-    def is_const(self):
+    def is_const(self) -> bool:
         return self.name.startswith('const ')
 
     def is_variadic(self):
```
umangyadav, line 39: Should this have return type of `bool` as well ?

tools/api.py
```diff
@@ -23,65 +24,65 @@ class Template(string.Template):
...
         return self.name.startswith('...')
 
-    def add_pointer(self):
+    def add_pointer(self) -> 'Type':
         return Type(self.name + '*')
 
     def add_reference(self):
```
umangyadav, line 45: return type of `Type` ?

tools/api.py
```diff
@@ -113,20 +114,20 @@ def str(self):
 
 
 class CFunction:
-    def __init__(self, name):
+    def __init__(self, name: str) -> None:
         self.name = name
```
umangyadav, line 118: `self.name : str`

tools/api.py
```diff
@@ -144,41 +145,45 @@ def substitute(self, form):
...
+    def __init__(self,
+                 name: str,
+                 type: str,
+                 optional: bool = False,
+                 returns: bool = False) -> None:
         self.name = name
         self.type = Type(type)
         self.optional = optional
-        self.cparams = []
+        self.cparams: List[Tuple[str, str]] = []
```
umangyadav, line 170: Do we need to add types for all other members e.g. `self.name`, `self.optional` or just the required important ones ?

tools/api.py
```diff
@@ -323,47 +332,56 @@ def share_params(self):
...
             self.returns.update()
         self.create_cfunction()
 
-    def inputs(self):
+    def inputs(self) -> str:
         return ', '.join([p.input() for p in self.params])
 
-    def input_map(self):
-        m = {}
+    # TODO: Shoule we remove Optional?
```
umangyadav, line 346: I think it should be okay to have it `Optional` since they are function's input and return args.

tools/generate.sh
pfultz2: Well that doesn't really work, it always outputs the `${MIN_VERSION}`. I think we can leave it for now, and you can add a check for the python version in a future PR if you want to.

kahmed10 (approved):

umangyadav (approved):

PR #1065, opened 2022-01-25
Add Mean op ONNX parser

### LINKED ISSUE #1034: Implement ONNX Operator Mean
Needed for some recent model support requests. The original models are from TensorFlow. The tf2onnx tool indicates that the TF Operator "CombinedNonMaxSuppression" has paths that call Mean

turneram:
Add ONNX parser for operator Mean

Resolves #1034


kahmed10:
What do you think of the following?
Concat all inputs (after multibroadcasting), then call ReduceMean on the first axis with keepdims = False. I believe this would result in the same shape as the current output shape that you calculate.

scxiao (approved):

src/onnx/parse_mean.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        std::for_each(args.begin() + 1, args.end(), [&](auto& data_i) {
+            if(data_i->get_shape().type() != output_type)
+                MIGRAPHX_THROW("Parse Mean: All inputs must have the same data type.");
+
+            auto data_i_lens = data_i->get_shape().lens();
+            if(not std::equal(
+                   data_i_lens.begin(), data_i_lens.end(), output_lens.begin(), output_lens.end()))
+                data_i = info.add_instruction(
+                    migraphx::make_op("multibroadcast", {{"out_lens", output_lens}}), data_i);
```
pfultz2, line 94: Can we use `add_common_op` to insert the broadcasts?

src/onnx/parse_mean.cpp
```diff
@@ -0,0 +1,107 @@
...
+                          output_lens.end()))
+            mean = info.add_instruction(
+                migraphx::make_op("multibroadcast", {{"out_lens", output_lens}}), mean);
+
+        // Pre-divide each tensor element-wise by n to reduce risk of overflow during summation
+        mean = info.add_instruction(migraphx::make_op("div"), mean, divisor);
+
+        std::for_each(args.begin() + 1, args.end(), [&](auto& data_i) {
+            if(data_i->get_shape().type() != output_type)
+                MIGRAPHX_THROW("Parse Mean: All inputs must have the same data type.");
```
pfultz2, line 88: If we use `add_common_op` then we shouldn't need this restriction.

src/onnx/parse_mean.cpp
```diff
@@ -0,0 +1,107 @@
...
+                          mean->get_shape().lens().end(),
+                          output_lens.begin(),
+                          output_lens.end()))
+            mean = info.add_instruction(
+                migraphx::make_op("multibroadcast", {{"out_lens", output_lens}}), mean);
+
+        // Pre-divide each tensor element-wise by n to reduce risk of overflow during summation
+        mean = info.add_instruction(migraphx::make_op("div"), mean, divisor);
+
+        std::for_each(args.begin() + 1, args.end(), [&](auto& data_i) {
```
pfultz2, line 86: This should be `std::accumulate(args.begin() + 1, args.end(), args[0], [&](auto mean, auto data_i) { ... })` instead.

pfultz2:
> Concat all inputs (after multibroadcasting), then call ReduceMean on the first axis with keepdims = False. I believe this would result in the same shape as the current output shape that you calculate.

We could try that as an optimization later, but I assume the number mean arguments is small.

src/onnx/parse_mean.cpp
pfultz2: Actually, `info.add_broadcastable_binary_op` would work as well.

turneram: Changed

src/onnx/parse_mean.cpp
turneram: Changed

kahmed10 (approved):

PR #1068, opened 2022-01-27
Support nonstandard shapes for the Squeeze Op

umangyadav:
Support slice, broadcast and transpose shapes for the squeeze op.

### DIFF
--- a/src/include/migraphx/op/squeeze.hpp
+++ b/src/include/migraphx/op/squeeze.hpp
@@ -37,22 +37,30 @@ struct squeeze
     std::string name() const { return "squeeze"; }
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(1).standard();
+        check_shapes{inputs, *this}.has(1);
         auto input_shape = inputs[0];
         auto type        = input_shape.type();
         auto old_lens    = input_shape.lens();
-
+        auto old_strides = input_shape.strides();
         if(std::any_of(axes.begin(), axes.end(), [&](auto axis) { return old_lens[axis] != 1; }))
         {
             MIGRAPHX_THROW("squeeze axis dimension should be equal to 1");
         }
         std::vector<std::size_t> new_lens;
+        std::vector<std::size_t> new_strides;
         if(axes.empty())
         {
             std::copy_if(old_lens.begin(),
                          old_lens.end(),
                          std::back_inserter(new_lens),
                          [](auto len) { return len != 1; });
+            for(std::size_t i = 0; i < old_lens.size(); i++)
+            {
+                if(old_lens[i] != 1)
+                {
+                    new_strides.push_back(old_strides[i]);
+                }
+            }
         }
         else
         {
@@ -61,19 +69,20 @@ struct squeeze
                 if(std::find(axes.begin(), axes.end(), i) == axes.end())
                 {
                     new_lens.push_back(old_lens[i]);
+                    new_strides.push_back(old_strides[i]);
                 }
             }
         }
-
         if(new_lens.empty())
         {
             return shape{type};
         }
         else
         {
-            return shape{type, new_lens};
+            return shape{type, new_lens, new_strides};
         }
     }
+
     argument compute(shape output_shape, std::vector<argument> args) const
     {
         return args[0].reshape(output_shape);
--- a/src/include/migraphx/op/squeeze.hpp
+++ b/src/include/migraphx/op/squeeze.hpp
@@ -50,21 +50,18 @@ struct squeeze
         std::vector<std::size_t> new_strides;
         if(axes.empty())
         {
-            std::copy_if(old_lens.begin(),
-                         old_lens.end(),
-                         std::back_inserter(new_lens),
-                         [](auto len) { return len != 1; });
-            for(std::size_t i = 0; i < old_lens.size(); i++)
+            for(auto i : range(old_lens.size()))
             {
                 if(old_lens[i] != 1)
                 {
+                    new_lens.push_back(old_lens[i]);
                     new_strides.push_back(old_strides[i]);
                 }
             }
         }
         else
         {
-            for(std::size_t i = 0; i < old_lens.size(); i++)
+            for(auto i : range(old_lens.size()))
             {
                 if(std::find(axes.begin(), axes.end(), i) == axes.end())
                 {
--- a/test/ref_ops_nonstd_shape_test.cpp
+++ b/test/ref_ops_nonstd_shape_test.cpp
@@ -73,9 +73,10 @@ TEST_CASE(squeeze_transpose_test)
     p.compile(migraphx::ref::target{});
     auto result          = p.eval({}).back();
     auto expected_result = p_uncompiled.eval({}).back();
+    // contiguous is required to read the values in standard shaped order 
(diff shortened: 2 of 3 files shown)


src/include/migraphx/op/squeeze.hpp
```diff
@@ -37,22 +37,30 @@ struct squeeze
...
         }
         std::vector<std::size_t> new_lens;
+        std::vector<std::size_t> new_strides;
         if(axes.empty())
         {
             std::copy_if(old_lens.begin(),
                          old_lens.end(),
                          std::back_inserter(new_lens),
                          [](auto len) { return len != 1; });
+            for(std::size_t i = 0; i < old_lens.size(); i++)
```
pfultz2, line 57: Do `for(auto i:range(old_lens.size()))` instead.

src/include/migraphx/op/squeeze.hpp
```diff
@@ -37,22 +37,30 @@ struct squeeze
...
+            {
+                if(old_lens[i] != 1)
+                {
+                    new_strides.push_back(old_strides[i]);
+                }
```
pfultz2, line 62: The `copy_if` from above could be combined with this loop so you would just do:

```cpp
                if(old_lens[i] != 1)
                {
                    new_strides.push_back(old_strides[i]);
                    new_lens.push_back(old_lens[i]);
                }
```

src/include/migraphx/op/squeeze.hpp
umangyadav: Done. Thanks.

src/include/migraphx/op/squeeze.hpp
umangyadav: Done thanks.

scxiao (approved):

turneram (approved):

test/ref_ops_nonstd_shape_test.cpp
```diff
@@ -55,4 +55,62 @@ TEST_CASE(argmin_test_nonstd_shape)
...
+    auto l0 = mm->add_literal(migraphx::literal{s1, data});
+    auto l0_trans =
+        mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {1, 2, 3, 0, 4}}}), l0);
+    mm->add_instruction(migraphx::make_op("squeeze"), l0_trans);
+    auto p_uncompiled = p;
+    p.compile(migraphx::ref::target{});
+    auto result         = p.eval({}).back();
+    auto* mm_uncompiled = p_uncompiled.get_main_module();
+    // ref implementation always does the auto_contiguous pass
+    migraphx::run_passes(*mm_uncompiled, {migraphx::auto_contiguous{}});
```
scxiao, line 73: I think adding a contiguous here does not test what you want. In this case, the output of the uncompiled program is a transposed shape, you need to manually do a transpose for the output according to the remaining dimension in the above transpose op (permutation is {1, 2, 0}).  and compared the output to that of the compiled program.

scxiao (commented):
I think the transposed input maybe incorrect after removing the standard input shape requirement.

scxiao:
@causten Please hold merging this PR since we can improve the unit test. I will let  you know after it is fixed.

pfultz2 (approved):

test/ref_ops_nonstd_shape_test.cpp
pfultz2: `contiguous` will preserve order of elements but does not preserve memory layout. However, this test only checks the shape and not the values in the tensor, which is a problem.

test/ref_ops_nonstd_shape_test.cpp
```diff
@@ -55,4 +55,62 @@ TEST_CASE(argmin_test_nonstd_shape)
...
+    auto expected_result = p_uncompiled.eval({}).back();
+    EXPECT(result.get_shape() == expected_result.get_shape());
```
pfultz2, line 75: You need to check the values are the same: `EXPECT(result == expected_result)`

pfultz2 (changes requested):

test/ref_ops_nonstd_shape_test.cpp
umangyadav: Made the change.

test/ref_ops_nonstd_shape_test.cpp
pfultz2: If you add `auto_contiguous` then you need to add `eliminate_contiguous` to make sure it wont remove a `contiguous` thats needed. Although, we shouldn't be running `auto_contiguous` on the uncompiled version.

umangyadav: @scxiao @pfultz2 I've added a contiguous op instead of using `auto_contiguous` pass.

scxiao (approved):

test/ref_ops_nonstd_shape_test.cpp
```diff
@@ -55,4 +55,71 @@ TEST_CASE(argmin_test_nonstd_shape)
...
 
+TEST_CASE(squeeze_transpose_test)
+{
+    migraphx::program p;
+    auto* mm                = p.get_main_module();
+    std::vector<float> data = {
+        1.2255,  1.6834,  -2.0305, -0.3221, 0.4701,  0.2583, 0.7545,  2.5758,  -1.6849,
+        0.0928,  0.9022,  -0.8765, -0.4090, 0.9301,  2.0724, -1.5706, 0.4867,  -0.1493,
+        0.6957,  -0.2179, 0.7142,  0.7177,  0.0183,  1.3497, 1.2255,  1.6834,  -2.0305,
+        -0.3221, 1.2255,  1.6834,  -2.0305, -0.3221, 1.2255, 1.6834,  -2.0305, -0.3221};
```
pfultz2, line 66: Just use `generate_literal` instead.

test/ref_ops_nonstd_shape_test.cpp
```diff
@@ -55,4 +55,71 @@ TEST_CASE(argmin_test_nonstd_shape)
...
+        mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {1, 2, 3, 0, 4}}}), l0);
+    mm->add_instruction(migraphx::make_op("squeeze"), l0_trans);
+    auto p_uncompiled = p;
+    p.compile(migraphx::ref::target{});
+    auto result          = p.eval({}).back();
+    auto expected_result = p_uncompiled.eval({}).back();
+    // contiguous is required to read the values in standard shaped order
+    auto tr_op               = migraphx::make_op("contiguous");
+    auto std_expected_result = tr_op.compute(result.get_shape(), {expected_result});
+    EXPECT(result == std_expected_result);
```
pfultz2, line 79: Can you add check for the shape so we can see what the final shape should be in the test? Something like:

```cpp
EXPECT(result.get_shape() == migraphx::shape{migraphx::shape::float_type, {4, 3, 3}})`
```

test/ref_ops_nonstd_shape_test.cpp
umangyadav: Done.

test/ref_ops_nonstd_shape_test.cpp
umangyadav: Done.

pfultz2 (approved):

PR #1069, opened 2022-01-27
Update missing copyright information for run_onnx_squad.py 

causten:
Add an explicit license for two pictures as they were taken by
an AMD employee at the time.  Update copyright for a python file
since it was updated at the time of creation in this repo to adhere
to 80 column width guidelines


examples/vision/python_super_resolution/License
```diff
@@ -0,0 +1 @@
+Copyright (C) 2021 Advanced Micro Devices, Inc. All rights reserved.
```
pfultz2, line 1: What license is this? Why is this different than the license [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/LICENSE)?

pfultz2:
> Add an explicit license for two pictures

Where is the license for the image? The images dont seem to be modified.

kahmed10 (approved):

mvermeulen (approved):
Overall changes look good to me.
Only question I have is noted in Paul's comment of whether a LICENSE file is OK expressing just copyright or whether it needs to explicitly reference/include a license.

examples/vision/python_super_resolution/License
causten: I had a meeting with Legal.  After showing what it looks like and the top level MIT License they walked back their request to have a license specific to the 2 jpg files.  I have added a commit remove these licenses.

pfultz2 (approved):

PR #1074, opened 2022-02-02
Add ScatterND operator

turneram:
Add onnx parser and ref and gpu implementations of ONNX op ScatterND


src/include/migraphx/op/scatternd.hpp
```diff
@@ -0,0 +1,74 @@
...
+        visit_all(result, args[0], args[2])([&](auto output, auto data, auto updates) {
+            std::copy(data.begin(), data.end(), output.begin());
+            args[1].visit([&](auto indices) {
+                auto updates_shape = updates.get_shape();
+                // k = index length, r = rank(data)
+                // k<r => update slices, k=r => update elements
+                auto k = indices.get_shape().lens().back();
+                par_for(updates_shape.elements(), [&](const auto i) {
+                    // updates and indices share the first dimension
+                    auto offset       = updates_shape.multi(i).front();
```
scxiao, line 51: Offset is not the first dimension of update, it is the first q - 1 dimensions of update, then append a 0 at the end to form a q-dim index, then compute the offset as: offset = indices_shape.index(idx).

src/include/migraphx/op/scatternd.hpp
```diff
@@ -0,0 +1,74 @@
...
+                    auto* index_start = indices.data() + (offset * k);
+                    auto* index_end   = index_start + k;
+                    auto out_idx      = output_shape.multi(i);
```
scxiao, line 54: For the output index, it first k-dim is the k-tuple from the indices input as computed above. The last (r-k)-dim should be from the last (r-k) dimension from the update_idx computed above.

src/targets/gpu/device/scatternd.cpp
```diff
@@ -0,0 +1,56 @@
...
+            auto* output_ptr     = device_cast(output.data());
+            const auto* data_ptr = device_cast(data.data());
+            gs_launch(stream, output_shape.elements())(
+                [=](auto i) __device__ { output_ptr[i] = data_ptr[i]; });
+            arg1.visit([&](auto indices_view) {
+                hip_visit_views(indices_view)([&](auto indices) {
+                    const auto* updates_ptr = device_cast(updates.data());
+                    const auto* indices_ptr = device_cast(indices.data());
+                    gs_launch(stream, updates_shape.elements())([=](auto i) __device__ {
+                        auto offset  = updates_shape.multi(i).front();
```
scxiao, line 34: The same problem as that in the ref implementation.

test/ref_ops_test.cpp
```diff
@@ -4215,6 +4215,222 @@ TEST_CASE(scatter_test)
...
+    {
+        // r=5, q=1, k=1
+        migraphx::program p;
+        auto* mm   = p.get_main_module();
+        auto dtype = migraphx::shape::float_type;
+        auto itype = migraphx::shape::int64_type;
+        migraphx::shape ds{dtype, {2, 2, 2, 2, 2}};
+        migraphx::shape is{itype, {1}};
+        migraphx::shape us{dtype, {1, 2, 2, 2, 2}};
+
```
scxiao, line 4351: The number of dim (5) of the "update" input does not make sense. 
In this case, r = 5, q = 1, k = 1, according to the onnx spec, u-dim = r + q - k - 1 = 4, so the update input should be 4 dimensions.

src/targets/gpu/device/scatternd.cpp
```diff
@@ -0,0 +1,56 @@
...
+                            output[out_idx] += updates_ptr[i];
+                        else if(mul)
+                            output[out_idx] *= updates_ptr[i];
+                        else
+                            output[out_idx] = updates_ptr[i];
+                    });
+                });
+            });
+        });
+
```
scxiao, line 49: One implementation can be:

```
                auto updates_shape = updates.get_shape();
                auto indices_shape = indices.get_shape();
                // k = index length, r = rank(data)
                // k<r => update slices, k=r => update elements
                auto k = indices_shape.lens().back();
                auto r = output_shape.lens().size();
                auto q = indices_shape.lens().size();
                par_for(updates_shape.elements(), [&](const auto i) {
                    // updates and indices share the first dimension
                    auto uidx = updates_shape.multi(i);
                    std::vector<std::size_t> iidx(q, 0);
                    std::copy(uidx.begin(), uidx.begin() + q - 1, iidx.begin());
                    auto* index_start = indices.data() + indices_shape.index(iidx.begin(), iidx.end());
                    auto* index_end   = index_start + k;

                    std::vector<std::size_t> oidx(r, 0);
                    std::copy(index_start, index_end, oidx.begin());
                    std::copy(uidx.begin() + q - 1, uidx.end(), oidx.begin() + k);

                    if(reduction == "add")
                        output[output_shape.index(oidx)] += updates[i];
                    else if(reduction == "mul")
                        output[output_shape.index(oidx)] *= updates[i];
                    else
                        output[output_shape.index(oidx)] = updates[i];
                });
            });
```

turneram: Thanks you. Fixed it in both locations

test/ref_ops_test.cpp
turneram: Fixed by removing the leading 1

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
```diff
@@ -0,0 +1,74 @@
...
+template <class T, class U, class V, class W, class Settings>
+__device__ void
+scatternd(const T& data_t, const U& indices_t, const V& updates_t, const W& output_t, Settings s)
+{
+    auto index           = make_index();
+    auto i               = index.global;
+    auto updates_shape   = updates_t.get_shape();
+    auto output_shape    = output_t.get_shape();
+    const auto* data_ptr = data_t.data();
+    auto* output_ptr     = output_t.data();
```
pfultz2, line 34: Use the tensors directly instead of using raw pointers.

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+        for(std::size_t j = q - 1; j < updates_idx.size(); ++j)
+            out_idx[j + k] = updates_idx[j];
+
+        if(is_add)
+            output_ptr[output_shape.index(out_idx)] += updates_ptr[i];
+        else if(is_mul)
+            output_ptr[output_shape.index(out_idx)] *= updates_ptr[i];
+        else
+            output_ptr[output_shape.index(out_idx)] = updates_ptr[i];
```
pfultz2, line 69: These should be passed as a function object instead so you should just do:

```cpp
template <class T, class U, class V, class W, class F>
__device__ void
scatternd(const T& data_t, const U& indices_t, const V& updates_t, const W& output_t, F f)  
{
    ...
    f(output_t[out_idx], updates_ptr[updates_idx]);
}
```

And then you can create function objects like this for the different operators:

```cpp
struct assign_mul
{
    template<class T, class U>
    void operator()(T& x, U y) const
    {
        x *= y;
    }
};
```

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
```diff
@@ -0,0 +1,74 @@
...
+    auto index           = make_index();
+    auto i               = index.global;
+    auto updates_shape   = updates_t.get_shape();
+    auto output_shape    = output_t.get_shape();
+    const auto* data_ptr = data_t.data();
+    auto* output_ptr     = output_t.data();
+
+    if(i < output_shape.elements())
+        output_ptr[i] = data_ptr[i];
+    __syncthreads();
```
pfultz2, line 38: Is this really needed?

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
scxiao: This may also apply to the ref implementation

test/verify/0test_scatternd.cpp
```diff
@@ -0,0 +1,28 @@
+#include "verify_program.hpp"
```
scxiao, line 1: there is a prefix "0" in the file name. Is that intentional?

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
scxiao: The __syncthreads() here is not enough since the updates to an element can happen in another workgroup. In this case, if the copy from data happens after the update, then the output is incorrect.  I would suggest use a separate kernel for the copy from data to output since the output elements can be much larger than the update elements, in that case, a lot of workitems do nothing in the kernel.

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
```diff
@@ -0,0 +1,74 @@
...
+        for(std::size_t j = 0; j < q - 1; ++j)
+            indices_idx[j] = updates_idx[j];
+
+        auto* index_start = indices_ptr + indices_shape.index(indices_idx);
+        auto out_idx      = output_shape.multi(0);
+        for(std::size_t j = 0; j < k; ++j)
+            out_idx[j] = index_start[j];
+
+        for(std::size_t j = q - 1; j < updates_idx.size(); ++j)
+            out_idx[j + k] = updates_idx[j];
```
scxiao, line 62: The index here is incorrect. It should be out_idx[j + k - (q - 1)]

src/include/migraphx/op/scatternd.hpp
```diff
@@ -0,0 +1,81 @@
...
+                auto indices_shape = indices.get_shape();
+                auto k             = indices_shape.lens().back();
+                auto q             = indices_shape.lens().size();
+                auto r             = output_shape.lens().size();
+                par_for(updates_shape.elements(), [&](const auto i) {
+                    auto updates_idx = updates_shape.multi(i);
+                    std::vector<std::size_t> indices_idx(q, 0);
+                    std::copy(
+                        updates_idx.begin(), updates_idx.begin() + q - 1, indices_idx.begin());
+                    auto* index_start = indices.data() +
```
pfultz2, line 55: This should use `indices.begin()` otherwise this may not work with non-standard shape inputs.

src/include/migraphx/op/scatternd.hpp
```diff
@@ -0,0 +1,81 @@
...
+                                        indices_shape.index(indices_idx.begin(), indices_idx.end());
+                    auto* index_end = index_start + k;
+
+                    std::vector<std::size_t> out_idx(r, 0);
+                    std::copy(index_start, index_end, out_idx.begin());
+                    std::copy(updates_idx.begin() + q - 1, updates_idx.end(), out_idx.begin() + k);
+
+                    if(reduction == "add")
```
pfultz2, line 63: Similar to reductions, I think it would be better to have different operators for the different reduction modes and then have a base class for the common code.

src/include/migraphx/op/scatternd.hpp
```diff
@@ -0,0 +1,81 @@
...
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.reduction, "reduction"));
+    }
+
+    std::string name() const { return "scatternd"; }
+
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(3).standard();
```
pfultz2, line 35: We should not require `standard` shape for ref implementation. If we use the iterators for tensor_view(instead of pointers), then it should work for non-standard shapes.

src/include/migraphx/op/scatternd_op.hpp
```diff
@@ -0,0 +1,72 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+template <class Derived>
+struct scatternd_op : op_name<Derived>
+{
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(3);
+        return inputs.front();
```
scxiao, line 28: After removing the standard shape requirement, return shape should be {inputs.front().type(), inputs.front().lens()}.

scxiao (commented):
@pfultz2 Actuall

src/targets/gpu/include/migraphx/gpu/compile_scatternd.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <migraphx/config.hpp>
+#include <migraphx/operation.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+
+struct context;
+operation
+compile_scatternd(context& ctx, const std::vector<shape>& io_shapes, const std::string& reduction);
```
pfultz2, line 13: Just set a default parameter for `reduction` to `none` instead of having two overloads.

src/targets/gpu/lowering.cpp
```diff
@@ -505,6 +508,60 @@ struct miopen_apply
...
+    {
+        apply_map.emplace("scatternd_none", [=](instruction_ref ins) {
+            auto s      = ins->get_shape();
+            auto op_val = ins->get_operator().to_value();
+            auto output = insert_allocation(ins, s);
+            auto args   = ins->inputs();
+            args.push_back(output);
+
+            auto io_shapes = to_shapes(args);
+            check_shapes{io_shapes, ins->get_operator()}.standard();
```
pfultz2, line 521: The shapes should already be standard here, so this check should be unnecessary.

src/targets/gpu/lowering.cpp
```diff
@@ -505,6 +508,60 @@ struct miopen_apply
...
+            const std::string reduction = "none";
+            auto co                     = compile_scatternd(get_context(), io_shapes, reduction);
+            auto co_copy                = compile_scatternd_copy(get_context(), io_shapes);
```
pfultz2, line 524: Is it possible to use our copy operator instead of a custom copy function?

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+template <class T, class U, class V, class W, class F>
+__device__ void
+scatternd(const T& /* data_t */, const U& indices_t, const V& updates_t, const W& output_t, F f)
+{
+    auto index         = make_index();
+    auto i             = index.global;
+    auto updates_shape = updates_t.get_shape();
+
+    if(i < updates_shape.elements())
```
pfultz2, line 44: This probably needs to use a global strided loop. Otherwise this would miss elements if `nglobal < updates_shape.elements()`.

src/targets/gpu/lowering.cpp
scxiao: How the shape has been standard here?

pfultz2: Because auto_contiguous ensures everything is standard shape before lowering.

src/include/migraphx/op/scatternd_op.hpp
```diff
@@ -0,0 +1,72 @@
...
+        visit_all(result, args[0], args[2])([&](auto output, auto data, auto updates) {
+            std::copy(data.begin(), data.end(), output.begin());
+            args[1].visit([&](auto indices) {
+                auto updates_shape = updates.get_shape();
+                auto indices_shape = indices.get_shape();
+                auto k             = indices_shape.lens().back();
+                auto q             = indices_shape.lens().size();
+                auto r             = output_shape.lens().size();
+                par_for(updates_shape.elements(), [&](const auto i) {
+                    auto updates_idx = updates_shape.multi(i);
```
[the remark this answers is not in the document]
turneram: Could I fix this by adding make_contiguous to indices and updates in the onnx parser?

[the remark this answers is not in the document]
scxiao: I think we can add a compute_s as:

```
shape compute_s{update_shape.type(), update_shape.lens()}, and use the compute_s to compute the multi index.

File: src/targets/gpu/lowering.cpp (hunk 16 above)
Comment (scxiao) [reply src/targets/gpu/lowering.cpp]: But the eliminate_contiguous can remove the contiguous later, and then the computation is wrong.

Comment (pfultz2) [reply src/targets/gpu/lowering.cpp:521]: It wont because the code_object_op will throw an error if the shapes change(see [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/code_object_op.cpp#L21)).

File: src/targets/gpu/include/migraphx/gpu/compile_scatternd.hpp [hunk 20]
```diff
@@ -0,0 +1,20 @@
...
+
+operation compile_scatternd_copy(context& ctx, const std::vector<shape>& io_shapes);
```
pfultz2, line 15: This is not used anywhere and can be removed.

src/targets/gpu/kernels/include/migraphx/kernels/scatternd.hpp
```diff
@@ -0,0 +1,65 @@
...
+    template <class T, class U>
+    MIGRAPHX_DEVICE_CONSTEXPR void operator()(T& x, U y) const
+    {
+        x *= y;
+    }
+};
+
+template <class T, class U, class V, class W, class F>
+__device__ void
+scatternd(const T& /* data_t */, const U& indices_t, const V& updates_t, const W& output_t, F f)
```
pfultz2, line 38: If `data_t` is not used then we should not even pass it in then. This should let memory coloring reuse the buffer for something else.

src/include/migraphx/op/scatternd_op.hpp
```diff
@@ -0,0 +1,94 @@
...
+                auto updates_std   = shape{updates_shape.type(), updates_shape.lens()};
+                auto indices_shape = indices.get_shape();
+                auto k             = indices_shape.lens().back();
+                auto q             = indices_shape.lens().size();
+                auto r             = output_shape.lens().size();
+                par_for(updates_shape.elements(), [&](const auto i) {
+                    auto updates_idx = updates_std.multi(i);
+                    std::vector<std::size_t> indices_idx(q, 0);
+                    std::copy(
+                        updates_idx.begin(), updates_idx.begin() + q - 1, indices_idx.begin());
```
pfultz2, line 69: Will this modify the `indices` input parameter? I am not sure if these side effects could cause incorrect optimizations.

turneram: Could you elaborate? I'm not sure if I understand your concern

pfultz2 (approved):

src/include/migraphx/op/scatternd_op.hpp
```diff
@@ -0,0 +1,94 @@
+#ifndef MIGRAPHX_GUARD_OPERATORS_SCATTERND_OP_HPP
+#define MIGRAPHX_GUARD_OPERATORS_SCATTERND_OP_HPP
+
+#include <array>
+#include <migraphx/op/name.hpp>
+#include <migraphx/check_shapes.hpp>
+#include <migraphx/stringutils.hpp>
+#include <migraphx/streamutils.hpp>
+#include <migraphx/shape_for_each.hpp>
```
scxiao, line 9: some include files might be removed, for exampe, shape_for_each, normalize_attributes?

src/targets/gpu/lowering.cpp
```diff
@@ -3,6 +3,7 @@
 #include <migraphx/manage_ptr.hpp>
 #include <migraphx/instruction.hpp>
 #include <migraphx/make_op.hpp>
+#include <migraphx/check_shapes.hpp>
```
scxiao, line 6: The check_shapes.hpp can be removed.

test/verify/test_scatternd.cpp
```diff
@@ -0,0 +1,28 @@
+#include "verify_program.hpp"
```
scxiao, line 1: Can we change one or two of these verify tests to a non-standard input shape?, like one in transposed, another is in broadcasted?

scxiao (approved):

PR #1075, opened 2022-02-02
Add assign_to method for C++ API 

kahmed10:


tools/generate.sh
```diff
@@ -14,4 +14,6 @@ function api {
 }
 
 api $DIR/api/migraphx.h $SRC_DIR/api/include/migraphx/migraphx.h
+echo "Finished generating header migraphx.h"
```
umangyadav, line 17: Would it be possible to add comment strings inside `migraphx.h` that it is auto-generated, and shouldn't be changed?

test/api/test_assign.cpp
```diff
@@ -0,0 +1,18 @@
+#include <migraphx/migraphx.h>
+#include <migraphx/migraphx.hpp>
+#include "test.hpp"
+
+TEST_CASE(shape_assign)
```
umangyadav, line 5: Tests for all other types e.g. `Arguments`, would be similar, you can add tests for all / some of them to increase test coverage.

umangyadav (approved):

test/api/test_assign.cpp
```diff
@@ -0,0 +1,18 @@
...
+{
+    auto s1 = migraphx::shape{migraphx_shape_float_type, {1, 3}};
+    migraphx_shape_t s2;
+    std::vector<size_t> lens{2, 3};
+    // handle ptr is const, workaround to construct shape using C API
+    migraphx_shape_create(&s2, migraphx_shape_float_type, lens.data(), lens.size());
```
pfultz2, line 11: Move `migraphx_shape_t s2;` to right above this line.

test/api/test_assign.cpp
```diff
@@ -0,0 +1,18 @@
+#include <migraphx/migraphx.h>
+#include <migraphx/migraphx.hpp>
+#include "test.hpp"
+
+TEST_CASE(shape_assign)
+{
+    auto s1 = migraphx::shape{migraphx_shape_float_type, {1, 3}};
```
pfultz2, line 7: Call this `s1_cpp` to be consistent.

test/api/test_assign.cpp
```diff
@@ -0,0 +1,18 @@
...
+    migraphx_shape_t s2;
+    std::vector<size_t> lens{2, 3};
+    // handle ptr is const, workaround to construct shape using C API
+    migraphx_shape_create(&s2, migraphx_shape_float_type, lens.data(), lens.size());
+    auto s2_cpp = migraphx::shape(s2, migraphx::own{});
+    CHECK(bool{s1 != s2_cpp});
+    migraphx_shape_assign(s2, s1.get_handle_ptr());
```
pfultz2, line 14: Call `s1.assign_to(s2)` here instead to get coverage for the C++ API.

pfultz2 (approved):

PR #1078, opened 2022-02-03
File ext rename

CharlieL7:
Changed MessagePack file extensions to mxr.

The yolov4 example using Python notebooks looks for a .msgpack extension. Not sure if those should be changed.

### DIFF
--- a/examples/migraphx/cpp_parse_load_save/README.md
+++ b/examples/migraphx/cpp_parse_load_save/README.md
@@ -15,7 +15,7 @@ p = parse_onnx(input_file, options);
 ```
 
 ## Saving
-An instantiated migraphx::program object can then be serialized to MessagePack (.msgpack) format and saved so that it can be loaded for future uses. 
+An instantiated migraphx::program object can then be serialized to MessagePack (.mxr) format and saved so that it can be loaded for future uses. 
 
 A program can be saved with either of the following:
 ```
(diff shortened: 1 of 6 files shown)


kahmed10:
> Changed MessagePack file extensions to mxr.
> 
> The yolov4 example using Python notebooks looks for a .msgpack extension. Not sure if those should be changed.

Yes please rename them and any other files which are using different extensions.

CharlieL7:
> > The yolov4 example using Python notebooks looks for a .msgpack extension. Not sure if those should be changed.
> 
> Yes please rename them and any other files which are using different extensions.

Made the edits.

turneram (approved):

kahmed10:
I think there's one more place that needs changing:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/serialize_program.cpp#L51
The .dat extension should be changed here.

CharlieL7:
> I think there's one more place that needs changing: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/serialize_program.cpp#L51 The .dat extension should be changed here.

Made changes to ".dat" extensions. 

It's not clear to me if the "msgpack" format and default file format are different or the same. They are handled differently in the parse_load_save.cpp example.

kahmed10 (approved):

PR #1080, opened 2022-02-07
Operator test scripts added

bpickrel:
Created Python scripts to run the test executable "gpu-driver" in batches, and to plot output.  Also updated source code to parse variables "local", "global", and "strides" from JSON input.  The script test_broadcast_perf.py runs a pointwise kernel operator(s) repeatedly while iterating over multiple combinations of local, global, shape/size of tensor being processed, and stride (with or without broadcasting).  It plots the results as series of stacked line graphs, one per tensor shape, and also saves the results in JSON format for re-analysis with plot_pointwise_data.py.  

The purpose of the new scripts is to allow us to run large numbers of gpu-driver tests in a single batch to determine the optimum combinations of parameters; gpu-driver runs operators directly without the full model/inference pipeline, both saving run time and improving signal/noise ratio for measuring the operator's run speed.

**test_broadcast_perf.py** takes no input arguments because users are expected to edit the Python code directly to change parameters and plot style; trading user-friendliness for flexibility.  Users are also invited to edit and reuse **plot_pointwise_data.py** heavily, as it's intended partially as an example for reshaping data with Python libraries such as pandas and matplotlib.  **plot_pointwise_data.py** takes the name of a JSON file created by **test_broadcast_perf.py**  as its command line argument.  Example:
`python3 test/py/test_broadcast_perf.py`
`python3 test/py/plot_pointwise_data.py ./test/py/pointwise/pointwise_perf_results_747229.json`
**test_broadcast_perf.py** also makes a backup copy of itself each time it is run, with a name containing a unique string it shares with the plot file so future users can review its exact conditions.  The default name conventions (unless user changes it) are
`'plot_' + <input tensor size> + '_' + <unique number string>+ '.png'`
for the plot, 
`'test_broadcast_perf.py' + '_t' + <unique number string> + '.py'`
for the backup copy of itself, and 
`'pointwise_perf_results_' + <unique number string> + '.json'`
for the saved data.

A sample output plot is attached.
plot_245760_605056


src/targets/gpu/include/migraphx/gpu/compile_pointwise.hpp
```diff
@@ -11,11 +11,14 @@ struct module;
...
-                            const std::vector<shape>& inputs,
-                            const std::string& lambda,
-                            const std::string& preamble = "");
-
+operation
+compile_pointwise(context& ctx,
+                  const std::vector<shape>& inputs,
+                  const std::string& lambda,
+                  const std::string& preamble = "",
+                  int global_workitems = 1024, // Global (total) work items. todo: testing indicates
```
pfultz2, line 19: Use `std::size_t` for the integers.

src/targets/gpu/include/migraphx/gpu/compile_pointwise.hpp
```diff
@@ -11,11 +11,14 @@ struct module;
...
+                                               // values of global close to tensor size are best
+                  int local_workitems = 256);  // Local work items per CU
```
pfultz2, line 21: Dont provide these defaults here. Instead add a function overload with these parameters missing and call `compute_global` with local set to 1024. Otherwise, there will be a performance regression.

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
```diff
@@ -2,6 +2,7 @@
 #define MIGRAPHX_GUARD_RTGLIB_DEVICE_LAUNCH_HPP
 
 #include <hip/hip_runtime.h>
+#include <hip/hcc_detail/hip_prof_str.h>
```
pfultz2, line 5: Remove this header.

src/targets/gpu/driver/compile_pointwise.cpp
```diff
@@ -14,8 +14,14 @@ struct compile_pointwise : action<compile_pointwise>
     {
         context ctx;
         auto inputs = p.parse_shapes(v.at("inputs"));
-        auto op     = gpu::compile_pointwise(ctx, inputs, v.at("lambda").to<std::string>());
-        double t    = time_op(ctx, op, inputs, p.get(v, "iterations", 100));
+
+        int global = p.get(v, "global", 64);
+        int local  = p.get(v, "local", 64);
```
pfultz2, line 19: If global and local are not provided then it should be omitted from the function call so the defaults will get called. You can do `p.get(v, "global", 0)` and then if its equal to `0` you know it was not provided.

test/py/plot_pointwise_data.py
```diff
@@ -0,0 +1,132 @@
...
+from matplotlib import pyplot as plt
+import json
+import os
+import sys
+import numpy as np
+import pandas
+
+if len(sys.argv) < 2:
+    print('syntax: python3 ' + __file__ +
+          " ./test/py/pointwise/pointwise_perf_results_807270.json")
```
pfultz2, line 20: These scripts should not go under test. They should go under the tools/ directory. We probably should create a new directory under tools/ to put these. Maybe tools/perf or tools/perf_analysis?

test/py/test_broadcast_perf.py
```diff
@@ -0,0 +1,288 @@
...
+                      'w') as f:
+                f.write(json.dumps(top_node, indent=4))
+            # f.close() not necessary
+
+            ##################################################
+            #   run gpu-driver
+            ##################################################
+
+            # print('run the gpu-driver...')
+            stream = os.popen(os.path.join(base_dir,'build/bin/gpu-driver  ') +\
```
pfultz2, line 175: We can't assume the build directory is located under build/. We should probably take the gpu-driver as an input parameter.

src/targets/gpu/driver/compile_pointwise.cpp
bpickrel: I don't see how that is different than passing the default as an argument.  Can you clarify?

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
bpickrel: done

src/targets/gpu/include/migraphx/gpu/compile_pointwise.hpp
bpickrel: This change has been made, commit d66690b22f1.  Please review as it's rather mistake-prone.  Also, I had to rearrange argument list with preamble last.

src/targets/gpu/compile_pointwise.cpp
```diff
@@ -57,7 +59,17 @@ operation compile_pointwise(context&,
...
 }
-
+// Overload of compile_pointwise without global, local calculates global value at runtime
+operation compile_pointwise(context& ctx,
+                            const std::vector<shape>& inputs,
+                            const std::string& lambda,
+                            const std::string& preamble)
+{
+    size_t local(1024);
+    size_t n      = inputs.size() > 0 ? inputs[0].elements() : 1024;
```
pfultz2, line 69: The inputs will not ever be empty, we can probably just add an assert for that, and just use `inputs.front().elements()`.

bpickrel: done.  inputs.first().elements()

tools/perf_analysis/plot_pointwise_data.py
```diff
@@ -0,0 +1,132 @@
...
+from matplotlib import pyplot as plt
+import json
+import os
+import sys
+import numpy as np
+import pandas
+
+if len(sys.argv) < 2:
+    print('syntax: python3 ' + __file__ +
+          " .tools/perf_analysis/pointwise/pointwise_perf_results_807270.json")
```
kahmed10, line 20: this should not be hardcoded.

tools/perf_analysis/plot_pointwise_data.py
```diff
@@ -0,0 +1,132 @@
...
+# param_c string  is the independent variable (the x axis) (column label)
+#
+def plot_vs_parameter(the_df, param_a, param_b, param_c):
+    the_df = the_df.sort_values(by=[param_b, param_c])
+    # make a list of unique global values
+
+    # issue: we get a different range of param_b (global) for different param_a (tensor size)
+    # each tensorsize should get its own legend??
+    ss = np.unique(the_df[param_b])
+    # print('ss=',ss)
```
kahmed10, line 66: is this comment needed?

tools/perf_analysis/plot_pointwise_data.py
```diff
@@ -0,0 +1,132 @@
...
+#   otherwise do follow-up analysis of the data.
+#
+#   syntax:  python3 plot_pointwise_data.py <input JSON file>
+
+from matplotlib import pyplot as plt
+import json
+import os
+import sys
+import numpy as np
+import pandas
```
kahmed10, line 16: I suggest importing argparse and adding more flags since currently there are hardcoded values which the user may want to change (such as in L82, L92). You can see tools/roctx.py to see how it's done there. It's fine to have default values with the flags.

tools/perf_analysis/plot_pointwise_data.py
```diff
@@ -0,0 +1,132 @@
+#                 plot_pointwise_data.py
+#                 Author:  Brian Pickrell, AMD copyright 2022
```
kahmed10, line 2: I don't think we need this format of copyright within the file.

tools/perf_analysis/test_broadcast_perf.py
```diff
@@ -0,0 +1,299 @@
+from matplotlib import pyplot as plt
+import json
+import os
+import sys
+import numpy as np
+import math
+from datetime import datetime
+
+#                 test_broadcast_perf.py
+#                 Author:  Brian Pickrell, AMD copyright 2022
```
kahmed10, line 10: same as the other file.

tools/perf_analysis/test_broadcast_perf.py
```diff
@@ -0,0 +1,299 @@
...
+
+#
+# this script runs the test program gpu-driver while iterating over
+# ranges of
+# tensor lens; global work items; and local work items, and plots the results.
+# It saves output to a JSON File as well, so that it can be re-plotted by
+# other tools.
+
+# this script is simlar to test_pointwise_perf.py, except that its
```
kahmed10, line 19: where is thi sfile?

tools/perf_analysis/test_broadcast_perf.py
```diff
@@ -0,0 +1,299 @@
...
+# Json arguments include "strides".  Setting one element in a stride to 0
+# cause a tensor to have one axis broadcast in gpu-driver.
+#
+# This script version is for comparing speeds of operators when broadcasting
+# tensor data.
+#
+# Engineers using this script should expect to modify it frequently in the course of testing.
```
kahmed10, line 26: This file should also import `argparse` so that users can pass in modified values without editing the script each time.

tools/perf_analysis/test_broadcast_perf.py
bpickrel: obs. comment deleted

tools/perf_analysis/test_broadcast_perf.py
bpickrel: They'll end up editing it all the time anyway, but a placeholder arg parser has been added to both scripts.

tools/perf_analysis/plot_pointwise_data.py
bpickrel: I've considered this but I don't think it's appropriate here for two reasons:
- parameterizing code (such as reading values from an args list here) without being sure the code will be heavily reused adds a layer of overhead that makes it harder, not easier, to update for very casual use;
- the code in question has a strong coupling between logic and the hard-coded numbers.  It's unlikely anyone would want to run this script with a variety of workitems values without editing the boolean expression they're embedded in as well.

Finally, there are so many good arg parser tutorials available that adding it just for the sake of an example would not add much value.

If and when someone decides to use these scripts on a regular basis with a variety of input values, then it will be time to add arg parser functionality.  I've declared arg parsers in both scripts but for now, they're just placeholders.

tools/perf_analysis/plot_pointwise_data.py
bpickrel: Removed.  I've left several other print statements in comment form where they show syntax for list indexing and slicing that's not self-evident.

tools/perf_analysis/plot_pointwise_data.py
bpickrel: updated

bpickrel:
Closing this as no longer required.

PR #1082, opened 2022-02-09
Enable pointwise fusion by default

pfultz2:
There is now a `MIGRAPHX_DISABLE_POINTWISE_FUSION` to disable it, if we happen to find problems with this in the short-term.


scxiao (approved):

test/gpu/fast_math.cpp
```diff
@@ -1,63 +0,0 @@
...
-    migraphx::program p = create_gelu();
-    p.compile(migraphx::gpu::target{});
-    CHECK(any_of(*p.get_main_module(), [&](auto&& i) { return i.name() == "gpu::gelu"; }));
-}
-
-TEST_CASE(disable_fast_gelu)
-{
-    migraphx::program p = create_gelu();
-    migraphx::compile_options options;
-    options.fast_math = false;
```
umangyadav, line 58: Why this test is deleted ? it is testing fast_math.  Does it rely on fusion being disabled ?

pfultz2: We no longer matched gelu to replace it with a faster version. Either way, this should not be done in fuse_ops by the gpu backend. See #1083 which is to add this to simplify_algebra.

umangyadav (approved):

PR #1087, opened 2022-02-11
EyeLike Operator

### LINKED ISSUE #1064: Implement ONNX Operator EyeLike
Needed for some recent model support requests. The original models are from TensorFlow. The tf2onnx tool indicates that the TF Operator "CombinedNonMaxSuppression" has paths that call ONNX Operator EyeLike

https://github.com/onnx/tensorflow-onnx/blob/65aaa2ca1ce77ea69731878928898afa080138f2/tf2onnx/onnx_opset/tensor.py#L1962
https://github.com/onnx/onnx/blob/main/docs/Operators.md#eyelike

CharlieL7:
Adds EyeLike ONNX parser and unit tests.

Resolves #1064


src/onnx/parse_eyelike.cpp
```diff
@@ -0,0 +1,67 @@
...
+        {
+            output_type = get_type(info.attributes.at("dtype").i());
+        }
+
+        int k = 0;
+        if(contains(info.attributes, "k"))
+        {
+            k = info.attributes.at("k").i();
+        }
+        if(static_cast<size_t>(std::abs(k)) >= num_cols)
```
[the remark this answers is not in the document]
CharlieL7: Right, it should check the number of rows for negative k. Making the change.

src/onnx/parse_eyelike.cpp
```diff
@@ -0,0 +1,67 @@
...
+        std::vector<double> eyelike_mat(num_rows * num_cols, 0);
+        for(int i = 0; i < num_rows; ++i)
+        {
+            for(int j = 0; j < num_cols; ++j)
+            {
+                if(j == (i + k))
+                {
+                    eyelike_mat[num_cols * i + j] = 1.;
+                }
+            }
```
pfultz2, line 68: I think the nested for loop can be removed

```cpp
for(auto i:range(num_rows))
{
    if (i + k < num_cols)
        eyelike_mat[(num_cols + 1) * i + k] = 1;
}
```

CharlieL7: Made the change with an added check for >= 0.

src/onnx/parse_eyelike.cpp
```diff
@@ -0,0 +1,81 @@
...
+                oss << "EYELIKE: negative k out of bounds, k = " << k << " num_rows = " << num_cols;
+                MIGRAPHX_THROW(oss.str());
+            }
+        }
+
+        std::vector<char> eyelike_mat(num_rows * num_cols, 0);
+        for(size_t i = 0; i < num_rows; ++i)
+        {
+            size_t tmp = std::abs(k);
+            if(k >= 0)
```
pfultz2, line 63: This check should be moved outside of the loop.

test/onnx/onnx_test.cpp
```diff
@@ -1259,6 +1259,146 @@ TEST_CASE(external_data_diff_path_test)
...
+    auto output_type = migraphx::shape::half_type;
+    migraphx::shape s{input_type, input_lens};
+    mm->add_parameter("T1", s);
+
+    std::vector<double> eyelike_mat(num_rows * num_cols, 0);
+    for(size_t i = 0; i < num_rows; ++i)
+    {
+        if(i + k < num_cols)
+            eyelike_mat[(num_cols + 1) * i + k] = 1.;
+    }
```
pfultz2, line 1330: You should probably make a function to fill the matrix instead of repeating this everytime.

src/onnx/parse_eyelike.cpp
pfultz2: Actually I think this can be written like this:

```cpp
std::size_t k_abs = std::abs(k);
for(auto i:range(num_rows))
{
    std::size_t idx = i + k;
    if (idx < num_cols and idx > 0)
        eyelike_mat[(num_cols + 1) * i + k] = 1;
}
```

CharlieL7: > Actually I think this can be written like this:
> 
> ```c++
> std::size_t k_abs = std::abs(k);
> for(auto i:range(num_rows))
> {
>     std::size_t idx = i + k;
>     if (idx < num_cols and idx > 0)
>         eyelike_mat[(num_cols + 1) * i + k] = 1;
> }
> ```

I think you have a typo with k_abs, not sure. i+k can evaluate to be negative so the idx cast would wrap around to a large positive number. This would work for a while because of the idx < num_cols position but the idx > 0 would always be true. Should be able to move the k checks outside of the loop. I will think about this some more and refactor.

test/onnx/onnx_test.cpp
CharlieL7: Where should I put the function?

src/onnx/parse_eyelike.cpp
```diff
@@ -0,0 +1,84 @@
...
+                if(i + k_abs < num_cols)
+                    eyelike_mat[(num_cols + 1) * i + k_abs] = 1.;
+            }
+        }
+        else
+        {
+            for(size_t i = 0; i < num_rows; ++i)
+            {
+                if(i >= k_abs) // i - k_abs >= 0
+                    eyelike_mat[(num_cols + 1) * i - k_abs] = 1.;
```
pfultz2, line 74: So doing `i - k_abs` is the same as `i + k`, so why couldn't you do?

```cpp
for(size_t i = 0; i < num_rows; ++i)
{
    auto idx = i + k;
    if(idx < num_cols and idx >= 0)
        eyelike_mat[(num_cols + 1) * i + k] = 1.;
}
```

CharlieL7: * In `auto idx = i + k` i + k can evaluate to be negative (ex. i = 1, k = -2). Auto seems to evaluate to an unsigned type between int and size_t so it wraps around to a large number. This can be avoided by making idx a signed int but that doesn't pass CI.
* One of the checkers complains about adding between size_t and int.

pfultz2: I see, we can just declare `i` as `std::ptrdiff_t` instead of `std::size_t`. This should do all the math with signed integers.

```cpp
for(std::ptrdiff_t i = 0; i < num_rows; ++i)
{
    auto idx = i + k;
    if(idx < num_cols and idx >= 0)
        eyelike_mat[(num_cols + 1) * i + k] = 1.;
}
```

We will probably need to also declare `num_cols` and `num_rows` as `std::ptrdiff_t` as well.

src/onnx/parse_eyelike.cpp
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        int k = 0;
+        if(contains(info.attributes, "k"))
+        {
+            k = info.attributes.at("k").i();
+        }
+        if(k >= 0)
+        {
+            if(static_cast<size_t>(k) >= num_cols)
```
pfultz2, line 42: If we declare `num_cols` as `std::ptrdiff_t` then this `static_cast` can be removed as well.

pfultz2 (approved):

src/onnx/parse_eyelike.cpp
```diff
@@ -0,0 +1,73 @@
...
+                MIGRAPHX_THROW(oss.str());
+            }
+        }
+
+        std::vector<char> eyelike_mat(num_rows * num_cols, 0);
+        for(std::ptrdiff_t i = 0; i < num_rows; ++i)
+        {
+            auto idx = i + k;
+            if(idx < num_cols and idx >= 0)
+                eyelike_mat[(num_cols + 1) * i + k] = 1.;
```
scxiao, line 64: maybe assign char{1} to the eyelike_mat since 1.0 will be converted to char type anyway

test/onnx/onnx_test.cpp
```diff
@@ -46,6 +46,17 @@ migraphx::program optimize_onnx(const std::string& name, bool run_passes = false
     return prog;
 }
 
+std::vector<double> make_r_eyelike(size_t num_rows, size_t num_cols, size_t k)
```
scxiao, line 49: use static for this function

scxiao (approved):

src/onnx/parse_eyelike.cpp
```diff
@@ -0,0 +1,67 @@
...
+        {
+            k = info.attributes.at("k").i();
+        }
+        if(static_cast<size_t>(std::abs(k)) >= num_cols)
+        {
+            std::ostringstream oss;
+            oss << "EYELIKE: k out of bounds, k = " << k << " num_cols = " << num_cols;
+            MIGRAPHX_THROW(oss.str());
+        }
+
```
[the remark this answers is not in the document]
turneram: We can probably just add this later if we encounter a model that requires it.

turneram (approved):

Issue #983, opened 2021-10-14
change the mode attribute in pooling and roialign to an enum 


scxiao:
In the pooling and roialign ops, there is a "mode" attribute for the pooling. Current implementation uses a string for the mode. We had better to change it to an enum type.

bpickrel:
The only 2 enum values I see for mode in the  roialign code are "mode"and "avg"--what other values are possible?
Should attribute coord_trans_mode also become an enum?  (values:  half_pixel, output_half_pixel)

In pooling the same question applies.  The only value I see hard-coded is "average".

Should both of these uses of mode refer to the same enum, or use separate definitions?

PR #1091, opened 2022-02-14
Mode as enum for pooling and roi_align

bpickrel:
Changed the pooling values for two structures from strings to specialized enum classes.  Many test and operator parsing changes to support this.  Introduces one new source file, op_enums.cpp.

Parsing may be error prone even after this change since both parsing and string conversion of the enums is still coupled to hard-coded strings.

This answers [issue 983](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/983)


src/driver/alexnet.cpp
```diff
@@ -61,7 +61,7 @@ migraphx::program alexnet(unsigned batch) // NOLINT(readability-function-size)
     migraphx::op::relu relu19;
     auto mx19 = mm->add_instruction(relu19, mx18);
     migraphx::op::pooling pooling20;
-    pooling20.mode    = "max";
+    pooling20.mode    = migraphx::op::pooling_mode::kMax;
```
kahmed10, line 64: Looks like clang tidy doesn't accept the variable names `kMax` and `kAvg`. Since we are scoping it to pooling mode only, would it be fine to just have `Max` and `Avg`?

src/include/migraphx/op/common.hpp
```diff
@@ -15,6 +17,19 @@ enum padding_mode_t
     valid
 };
 
+// The pooling modes must correspond 1-1 to the operators defined for struct parse_pooling.
+enum class pooling_mode
+{
+    avg,
```
pfultz2, line 23: I think it would be better to name this `average` so it matches the printed output.

src/op_enums.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+// roi_align roialign_mode
+std::ostream& operator<<(std::ostream& os, roialign_mode v)
+{
+    static const std::vector<std::string> roi_str = {"average", "max"};
+    os << roi_str[static_cast<std::underlying_type<roialign_mode>::type>(v)];
+    return os;
+}
+
+std::string to_str(pooling_mode v)
```
pfultz2, line 38: We already have `to_string` in the stringutils.hpp that does this, so these function can be removed. I dont see them used anywhere either.

src/include/migraphx/op/common.hpp
```diff
@@ -15,6 +17,19 @@ enum padding_mode_t
...
+    max
+};
+
+enum class roialign_mode
```
pfultz2, line 27: This enum should be removed. The mode in roialign is describing the pooling_mode so it should use the same `pooling_mode` enum. This should also make it clearer in the code that the mode is the pooling mode and not some other type of mode.

src/include/migraphx/op/common.hpp
bpickrel: I'll pass on making that change--I already changed it once.

pfultz2: Whats the issue with making the change?

src/op_enums.cpp
bpickrel: Done; commit 8b0fbf7a

src/op_enums.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+std::ostream& operator<<(std::ostream& os, rnn_direction v)
+{
+    static const std::vector<std::string> rnn_direction_str = {
+        "forward", "reverse", "bidirectional"};
+    os << rnn_direction_str[static_cast<std::underlying_type<rnn_direction>::type>(v)];
+    return os;
+}
+
+std::string to_str(pooling_mode v)
```
pfultz2, line 30: So this was removed in the header, but it still needs to be removed from the .cpp file.

kahmed10:
I am not sure why `test_avg_pooling_ceil_3d` is failing on the clang ASAN pass. You might want to look into that.

kahmed10 (approved):

bpickrel:
Looks like the formatting version problem (PR #4942 should fix it)

pfultz2 (approved):

scxiao (approved):

PR #1093, opened 2022-02-16
Expose APIs for the MIGraphX program

umangyadav:
API includes following
`create_module`, 
`get_main_module` 
`add_instruction`  without module args
`add_instruction` with module args
`add_parameter` 
`add_return`

based on these API two testcases have been added in `test_module_construct.cpp`. 

Follow up work : `insert_instruction`, `add_literal` and expose `literal`.


src/api/include/migraphx/migraphx.hpp
```diff
@@ -493,12 +495,110 @@ struct shapes : MIGRAPHX_HANDLE_BASE(shapes), array_base<shapes>
...
+    {
+        std::array<char, 1024> out_name;
+        call(&migraphx_operation_name, out_name.data(), 1024, this->get_handle_ptr());
+        return {out_name.data()};
+    }
+};
+
+struct instruction_ref : MIGRAPHX_HANDLE_BASE(instruction_ref)
+{
+    instruction_ref(migraphx_instruction_ref* p, own) { this->set_handle(p, own()); }
```
pfultz2, line 520: There should still be a borrow constructor.

pfultz2, line 520: Remove the `own` parameter from the constructor and use `{}` to construct `own`: `instruction_ref(migraphx_instruction_ref* p) { this->set_handle(p, own{}); }`

src/api/migraphx.py
```diff
@@ -178,14 +178,67 @@ def shapes(h):
              returns='const migraphx::shape&')
 
 
+@auto_handle()
+def instruction_ref(h):
+    h.constructor('create')
```
pfultz2, line 183: This constructor can be removed. I think you need to use `pass` to make python happy.

src/api/migraphx.py
```diff
@@ -178,14 +178,67 @@ def shapes(h):
...
+    h.method('size', returns='size_t')
+    h.method('get',
+             api.params(idx='size_t'),
+             fname='at',
+             cpp_name='operator[]',
+             returns='const migraphx::instruction_ref&')
+
+
+@api.handle('migraphx_modules_refs', 'std::vector<migraphx::module*>')
+def modules_refs(h):
```
pfultz2, line 201: This should probably be called `modules` as the module class it already a `module_ref`.

src/api/migraphx.py
```diff
@@ -178,14 +178,67 @@ def shapes(h):
              returns='const migraphx::shape&')
 
 
+@auto_handle()
+def instruction_ref(h):
```
pfultz2, line 182: I think maybe we should call this `instruction` just to be consistent with `module`.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -493,12 +495,110 @@ struct shapes : MIGRAPHX_HANDLE_BASE(shapes), array_base<shapes>
...
+
+    instructions_refs(migraphx_instructions_refs* p, borrow) { this->set_handle(p, borrow{}); }
+
+    template <class... Ts>
+    instructions_refs(Ts... xs)
+    {
+        std::array<migraphx_instruction_ref_t, sizeof...(Ts)> a{xs.get_handle_ptr()...};
+        this->make_handle(&migraphx_instructions_refs_create, a.data(), a.size());
+    }
+};
```
pfultz2, line 536: This should add the `operator[]`, and `size` method as well as the `iterator_read` class. Same for module_refs.

src/api/include/migraphx/migraphx.hpp

src/api/migraphx.py
umangyadav: Done.

src/api/migraphx.py
umangyadav: Changed it to just `modules`.

src/api/include/migraphx/migraphx.hpp
umangyadav: Removed those methods and `array_base` inheritance.

kahmed10 (approved):

test/api/test_module_construct.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+TEST_CASE(add_op)
+{
+    migraphx::program p;
+    migraphx::module m = p.get_main_module();
+    auto x             = m.add_parameter("x", migraphx::shape(migraphx_shape_float_type, {3, 3}));
+    auto y             = m.add_parameter("y", migraphx::shape(migraphx_shape_float_type, {3, 3}));
+    auto add_op        = migraphx::operation("add");
+    m.add_instruction(add_op, {x, y});
+    m.print();
```
pfultz2, line 13: We should compile and run the program.

pfultz2: At least with the `ref` implementation.

umangyadav: added logic to run on GPU and verify result with Ref target

test/api/test_module_construct.cpp
```diff
@@ -1,16 +1,56 @@
...
+{
+    // make a copy of the program
+    migraphx_program_t p_ref_c;
+    migraphx_program_create(&p_ref_c);
+    migraphx_program_assign_to(p_ref_c, p.get_handle_ptr());
+    auto p_ref = migraphx::program(p_ref_c, migraphx::own{});
+    // compile on GPU
+    migraphx::compile_options options;
+    options.set_offload_copy();
+    p.compile(migraphx::target("gpu"), options);
```
pfultz2, line 16: We dont need to run it on the gpu(and this will break when gpu is not available). Just run it on the ref and then check the expected result.

test/api/test_module_construct.cpp
```diff
@@ -1,16 +1,56 @@
...
+    // set parameters_map
+    migraphx::program_parameters pp;
+    auto param_shapes = p.get_parameter_shapes();
+    for(auto&& name : param_shapes.names())
+    {
+        pp.add(name, migraphx::argument::generate(param_shapes[name]));
```
pfultz2, line 22: We cant use random numbers. Just fill it the same value. We aren't trying to do numeric validation here, just to need to check that the program ran as expected.

test/api/test_module_construct.cpp
```diff
@@ -1,16 +1,56 @@
...
+    p_ref.compile(migraphx::target("ref"));
+    // get results from Ref target
+    auto outputs_ref = p_ref.eval(pp);
+    auto output_ref  = outputs_ref[0];
+    lens             = output_ref.get_shape().lengths();
+    elem_num         = std::accumulate(lens.begin(), lens.end(), 1, std::multiplies<std::size_t>());
+    float* output_ref_data_ptr = reinterpret_cast<float*>(output_ref.data());
+    std::vector<float> output_ref_vec(output_ref_data_ptr, output_ref_data_ptr + elem_num);
+    // compare the results;
+    CHECK(output_vec == output_ref_vec);
```
pfultz2, line 41: There is no need to put the argument into a vector, you can use equality on the `argument` directly.

test/api/test_module_construct.cpp
umangyadav: Modified to run only on Ref.

test/api/test_module_construct.cpp
umangyadav: Done.

test/api/test_module_construct.cpp
umangyadav: Fixed.

test/api/test_module_construct.cpp
umangyadav: Did equality for `add_op` test.  But for the `if_op`, test i am still converting into a vector. For some reason the return argument has type of `bool` .  but the data of the return argument matches to the expected. Not able to figure out where this type mismatch is coming from.

umangyadav: there is type mismatch because i am using lambda which requires copy and assignment.

test/api/test_module_construct.cpp
```diff
@@ -0,0 +1,81 @@
...
+        return p;
+    };
+
+    std::vector<float> x_data(9, 1);
+    std::vector<float> y_data(9, -1);
+    auto run_prog = [&](bool cond) {
+        auto p = create_program();
+        p.compile(migraphx::target("ref"));
+        migraphx::program_parameters pp;
+        char ccond        = cond;
```
pfultz2, line 60: Cant you just pass `&cond` directly into argument?

test/api/test_module_construct.cpp
```diff
@@ -0,0 +1,81 @@
...
+        auto param_shapes = p.get_parameter_shapes();
+        pp.add("cond", migraphx::argument(cond_s, &ccond));
+        pp.add("x", migraphx::argument(param_shape, x_data.data()));
+        pp.add("y", migraphx::argument(param_shape, y_data.data()));
+        auto outputs    = p.eval(pp);
```
pfultz2, line 65: Perahps its better to use initializer lists instead: `p.eval({{"cond", migraphx::argument(cond_s, &cond)}, {"x", migraphx::argument(param_shape, x_data.data())}, {"y", migraphx::argument(param_shape, y_data.data())}}`

test/api/test_module_construct.cpp
```diff
@@ -0,0 +1,81 @@
...
+        auto output     = outputs[0];
+        float* data_ptr = reinterpret_cast<float*>(output.data());
+        std::vector<float> ret(data_ptr, data_ptr + 9);
+        return ret;
```
pfultz2, line 69: This should return `outputs[0]`. No need to return a vector.

test/api/test_module_construct.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+        auto if_ins = mm.add_instruction(migraphx::operation("if"), {cond}, {then_mod, else_mod});
+        auto get_tuple_op = migraphx::operation("get_tuple_elem", "{index: 0}");
+        auto ret          = mm.add_instruction(get_tuple_op, {if_ins});
+        mm.add_return({ret});
+        return p;
+    };
+
+    std::vector<float> x_data(9, 1);
+    std::vector<float> y_data(9, -1);
```
pfultz2, line 55: You should create the arguments for these here:

```cpp
    std::vector<float> x_data(9, 1);
    std::vector<float> y_data(9, -1);
    auto x = migraphx::argument(param_shape, x_data.data());
    auto y = migraphx::argument(param_shape, y_data.data());
    auto run_prog = [&](bool cond) {
        ...
    };
    // then branch
    auto then_res = run_prog(true);
    CHECK(run_prog(true) == x);
    // else branch
    CHECK(run_prog(false) == y);
```

test/api/test_module_construct.cpp
umangyadav: yes, but  for some reason, when `outputs[0]` argument is returned from the lambda function, its type changes from `float_type` to `bool_type`.  Underlying data remains the same.  Since type is mismatched, i am not able to directly use `==` on `migraphx::argument`.

test/api/test_module_construct.cpp
umangyadav: It worked for tidy and cppcheck.

test/api/test_module_construct.cpp
umangyadav: Done.

test/api/test_module_construct.cpp
umangyadav: Done

test/api/test_module_construct.cpp
pfultz2: That looks like a bug somewhere. Whats happens when running with ASAN? Also what does program look like after compiling for "ref"?

umangyadav:
```module: "main"
y = @param:y -> float_type, {3, 3}, {3, 1}
x = @param:x -> float_type, {3, 3}, {3, 1}
cond = @param:cond -> bool_type, {1}, {0}
main:@3 = if(cond), [If_0_if, If_0_else] -> [float_type, {3, 3}, {3, 1}]
main:@4 = ref::get_tuple_elem[index=0](main:@3) -> float_type, {3, 3}, {3, 1}
main:@5 = @return(main:@4)

module: "If_0_if"
If_0_if:@0 = ref::identity(x) -> float_type, {3, 3}, {3, 1}
If_0_if:@1 = @return(If_0_if:@0)

module: "If_0_else"
If_0_else:@0 = ref::identity(y) -> float_type, {3, 3}, {3, 1}
If_0_else:@1 = @return(If_0_else:@0)

This is after "ref" compilation. I ran ASAN but it seems to find unrelated memory leaks ( in libamdhip64).

umangyadav: Made required changes by returning `outputs` from lambda.

pfultz2 (approved):

PR #1094, opened 2022-02-17
Complete GPU implementation of CumSum op

### LINKED ISSUE #756: Unknown operator: CumSum for RoBERTa-base-11.onnx
**Model from ONNX Zoo:** /models/text/machine_comprehension/roberta/model/roberta-base-11.onnx

While executing this model with example `cpp_parse_load_save` and after fixing the `not` operator, it complains for operator `CumSum` :
```
./parse_load_save /cgMIGraphX/models/text/machine_comprehension/roberta/model/roberta-base-11.onnx --parse onnx
Parsing ONNX File
MIGraphX Error: /cgMIGraphX/notoperator/AMDMIGraphX/src/onnx/onnx_parser.cpp:281: parse_graph: Unknown operator: CumSum
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to call function
Aborted (core dumped)

```

### LINKED ISSUE #1039: Prefix scan for GPU
The inclusive mode is supported already for GPU. The work here is to support Reverse and Exclusive mode

turneram:
Add exclusive and reverse modes to gpu implementation of prefix_scan_sum, which completes support for ONNX op CumSum

Resolves #1039 
Resolves #756


src/targets/gpu/device/include/migraphx/gpu/device/scan.hpp
```diff
@@ -52,6 +52,51 @@ __device__ void block_scan(index idx, Op op, T init, index_int n, Input input, O
...
+            if(idx.local + s < idx.nlocal())
+            {
+                buffer[idx.local + s] = op(buffer[idx.local], buffer[idx.local + s]);
+            }
+            __syncthreads();
+        }
+        x = buffer[idx.local];
+        output(i, buffer[idx.local]);
+    });
+}
```
pfultz2, line 85: So a seperate implementation shouldn't be necessary, I think we can use a function combinator on the input and output to do this:

```cpp
template<class F>
constexpr auto reverse_scan(index_int n, F f)
{
    return [=](auto i, auto&&... xs) {
        return f(n-i-1, xs...);
    };
}
template <index_int N, class Op, class T, class Input, class Output>
__device__ void reverse_block_scan(index idx, Op op, T init, index_int n, Input input, Output output)
{
    block_scan<N>(idx,
                  op,
                  init,
                  n,
                  reverse_scan(input),
                  reverse_scan(output));
}
```

And `reverse_block_scan` may not be needed as you could do this directly in the `prefix_scan_sum` function. I think something similar could be done for exclusive scan as well, maybe something like:

```cpp
template<class F>
constexpr auto exclusive_scan(index_int n, F f)
{
    return [=](auto i, auto&&... xs) {
        auto j = i + 1;
        if (j < n)
            f(j, xs...);
    };
}
```

And apply this to the output.

turneram: Made the changes for reverse_scan, but I haven't been able to get exclusive_scan to work. When I attempt to use it in the same way that I've applied reverse_scan, I get this error:
```
AMDMIGraphX/src/targets/gpu/device/include/migraphx/gpu/device/scan.hpp:23:39: error: array has incomplete element type 'type' (aka 'void')
    MIGRAPHX_DEVICE_SHARED type buffer[N];
                                      ^
```

src/targets/gpu/device/prefix_scan_sum.cpp
```diff
@@ -24,12 +51,32 @@ void prefix_scan_sum(hipStream_t stream, const argument& result, const argument&
...
+                            idx,
+                            sum{},
+                            0,
+                            n,
+                            reverse_scan(n,
+                                         [&](auto j) {
+                                             return exclusive ? output[compute_idx(j)]
+                                                              : input[compute_idx(j)];
+                                         }),
+                            reverse_scan(n, [&](auto j, auto x) { output[compute_idx(j)] = x; }));
```
pfultz2, line 66: Isn't exclusive an adjustment to the output not the input? I would think it would be done something like this:

```cpp
block_scan<block_size>(
    idx,
    sum{},
    0,
    n,
    reverse_scan(n,
                 [&](auto j) {
                     input[compute_idx(j)];
                 }),
    reverse_scan(n, [&](auto j, auto x) {
        if (exclusive) {
            auto idx = j + 1;
            if (j == 0)
                output[compute_idx(0)] = 0;
            if (idx < n)
                output[compute_idx(idx)] = x;
        } else {
            output[compute_idx(j)] = x; 
        }
    }));
```

src/targets/gpu/device/prefix_scan_sum.cpp
```diff
@@ -24,12 +51,32 @@ void prefix_scan_sum(hipStream_t stream, const argument& result, const argument&
                         k[axis] = j;
                         return k;
                     };
-                    block_scan<block_size>(idx,
-                                           sum{},
-                                           0,
-                                           n,
-                                           [&](auto j) { return input[compute_idx(j)]; },
-                                           [&](auto j, auto x) { output[compute_idx(j)] = x; });
+                    if(reverse)
```
pfultz2, line 54: The `if` branches need to be moved outside of `gs_launch` or they need to use `if constexpr`. To use it in `if constexpr` you need to use a visitor:

```cpp
template<class F>
void visit_bool(bool b, F f)
{
    if (b)
        f(std::true_type{});
    else
        f(std::false_type{});
}
```

Then you can do something like:

```cpp
visit_bool(exclusive, [&](auto exclusive_c) {
    if constexpr(exclusive_c)
    {
        ...
    }
});
```

src/include/migraphx/op/prefix_scan_op.hpp
```diff
@@ -38,7 +38,15 @@ struct prefix_scan_op : op_name<Derived>
...
-        return inputs.at(0);
+        auto s = inputs.front();
+        if(s.broadcasted())
+        {
+            return {s.type(), s.lens()};
+        }
+        else
+        {
+            return s.with_lens(s.lens());
+        }
```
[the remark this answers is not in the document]
scxiao: Looks like the cpu target failed. Maybe the oneDNN implementation does not accept broadcasted input shape. You can install the oneDNN locally, then turn on the flag MIGRAPHX_ENABLE_CPU to debug it.

I checked the oneDNN documents, https://oneapi-src.github.io/oneDNN/group_dnnl_api.html, looks like it does not have the cumsum operator. @pfultz2 How do we handle this case? just fall back to use the ref implementation?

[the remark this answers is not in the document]
scxiao: Also, I printed out the input values with MIGRAPHX_TRACE_EVAL=2, by manually calculating the outputs, looks like both the ref and gpu implementations are correct.

kahmed10 (approved):

pfultz2 (approved):

Issue #1090, opened 2022-02-14
Commit e64b773f2 (PR#1068 Support nonstandard shapes for the Squeeze Op) caused an error the BERT squad large model


scxiao:
The model can be downloaded at:
https://zenodo.org/record/3733910/files/model.onnx

The command line to run the model:

```
./bin/migraphx-driver perf model.onnx --fill1 input_ids --fill1 input_mask --fill1 segment_ids --batch 1
```

Error message is:

terminate called after throwing an instance of 'migraphx::version_1::exception'
  what():  /workspace/host/projects/AMDMIGraphX/src/program.cpp:171: compile: Dangling reference in module main from instruction 1633
Aborted (core dumped)

scxiao:
It is caused by the following scenario:

Before removing standard input requirement for squeeze, the last few instructions are:

```
main:@2254 = gpu::code_object[code_object=13736,symbol_name=kernel,global=1024,local=1024,](main:@2250,main:@2252,main:@2253) -> float_type, {1, 384, 2}, {768, 2, 1}
main:@2255 = slice[axes={2},starts={1},ends={2}](main:@2254) -> float_type, {1, 384, 1}, {768, 2, 1}
main:#output_1 = @param:main:#output_1 -> float_type, {1, 384, 1}, {384, 1, 1}
main:@2256 = gpu::contiguous(main:@2255,main:#output_1) -> float_type, {1, 384, 1}, {384, 1, 1}
main:@2257 = slice[axes={2},starts={0},ends={1}](main:@2254) -> float_type, {1, 384, 1}, {768, 2, 1}
main:#output_0 = @param:main:#output_0 -> float_type, {1, 384, 1}, {384, 1, 1}
main:@2258 = gpu::contiguous(main:@2257,main:#output_0) -> float_type, {1, 384, 1}, {384, 1, 1}
main:@2259 = squeeze[axes={2}](main:@2256) -> float_type, {1, 384}, {384, 1}
main:@2260 = squeeze[axes={2}](main:@2258) -> float_type, {1, 384}, {384, 1}
main:@2261 = @return(main:@2260,main:@2259)
```
After removing, the last few instructions are:

```
main:@2254 = gpu::code_object[code_object=13736,symbol_name=kernel,global=1024,local=1024,](main:@2250,main:@2252,main:@2253) -> float_type, {1, 384, 2}, {768, 2, 1}
[...]

PR #1095, opened 2022-02-18
Alias 0 as ret value

scxiao:
This PR is to fix the problem that an op of alias_0 is used a return value, and offload_copy is false. 

In this case, the output parameter is provided by a user. If the output shape is different from the alias ins shape, then there is an error since the buffer provided by user cannot be used as the alias ins output. 

The fix is: if an op of alias 0 is used as a return value, and the op element_space is different from the element_space of its alias ins, we need to add a contiguous after the the op, so the alias output is allocated by the program, and user only needs to provide the output param to be used as the output to the contiguous.

This PR also refines the instruction index when printing a program. Current implementation is count only the instructions (param and literals are ignored), I changed it to count everything, so we can use the index of the instruction to get a better idea the instruction location in the program printed out.


scxiao:
This PR is to resolving the issue of dangling reference in issue #1090, but the solution is not what we like. Khalique is working on another PR to solve this problem. So close this PR.

PR #1096, opened 2022-02-21
Improve print ins

scxiao:
Right now, the module::debug_print(ins) is very slow, which make the trave_eval==1/2 very slow in print out instruction. The reason is printing an ins involves search the whole module to get the instruction, the print it. 

This change is to fix that by calling module::print() to get names of all instructions of a program, then print the instruction by getting its name from a hash map.


src/include/migraphx/program.hpp
```diff
@@ -76,7 +76,8 @@ struct program
     void from_value(const value& v);
 
     void debug_print() const;
-    void debug_print(instruction_ref ins) const;
+    void debug_print(instruction_ref ins,
+                     const std::unordered_map<instruction_ref, std::string>& ins_names) const;
```
pfultz2, line 80: We should not add a map as a parameter as it makes it very difficult(if not impossible) to call from a debugger.

src/program.cpp
```diff
@@ -353,13 +353,17 @@ std::vector<argument> program::eval(parameter_map params) const
 
     if(trace_level > 0)
     {
+        std::unordered_map<instruction_ref, std::string> ins_names;
+        // get instruction names
+        this->print(ins_names, [](auto, auto) {});
```
pfultz2, line 358: We should fill the map with the instruction print outs:

```cpp
std::unordered_map<instruction_ref, std::string> ins_out;
this->print([&](auto x, auto ins_names) {
    std::stringstream ss;
    instruction::print(ss, x, ins_names);
    ins_out[x] = ss.str();
});
```

Then when we want to print out we can do `std::cout << "Run instruction: " << ins_out.at(ins) << std::endl;`.

pfultz2: We also should add a `print` overload that doesnt take a map of names:

```cpp
void program::print(
    const std::function<void(instruction_ref, std::unordered_map<instruction_ref, std::string>)>&
        print_func) const
{
    std::unordered_map<instruction_ref, std::string> names;
    this->print(names, print_func);
}
```

src/include/migraphx/program.hpp
scxiao: @pfultz2 Maybe we can also keep the current function debug_print(instruction_ref ins) for debugging purpose. The reason I made this change is the current function needs to go through the whole program each time an ins is printed. This make the trace_eval=1 very slow, particularly for large model.

src/program.cpp
scxiao: fixed

pfultz2 (approved):

turneram (approved):

scxiao:
It looks like the missing lines can be called only with MIGRAPHX_TRACE_EVAL=1.

pfultz2:
Yea I dont think we need coverage tests for this, so it can be merged in as is.

PR #1103, opened 2022-02-25
Show the operator fields in the driver

pfultz2:
Show the fields and default values for an operator in the driver.


kahmed10 (approved):

src/driver/main.cpp
```diff
@@ -505,8 +505,10 @@ struct roctx : command<roctx>
 struct op : command<op>
 {
     bool show_ops = false;
+    std::string op_name{};
     void parse(argument_parser& ap)
     {
+        ap(op_name, {}, ap.metavar("<operator name>"));
```
scxiao, line 511: Looks like this time it works fine. I would suggest change the "operator name" to "migraphx operator name" or "migraphx op name" to explicitly mention it is not an ONNX op name.

scxiao (approved):

PR #1106, opened 2022-02-28
Clang format ver10

bpickrel:
Update the base version of clang-format from 5.0 to 10.0, which is the version that installs with Ubuntu 20.04. Observed differences were only in where lines break, but see [clang release notes](https://releases.llvm.org/10.0.0/tools/clang/docs/ReleaseNotes.html#clang-format) for more details.


kahmed10 (approved):

pfultz2:
We need to update the dockerfile to install clang-format-10 as well in Dockerfile and hip-clang.docker

pfultz2 (approved):

bpickrel:
Oh, there's another thing, @pfultz2 :  I just ran clang-format-10 on all the source files and I should also commit those that formatted differently due to this change.  But git is telling me that some file permissions were also altered.  Should I or should I not commit the files as-is, with the following difference:  
```
mode change 100755 => 100644 src/include/migraphx/matcher.hpp

 mode change 100755 => 100644 src/include/migraphx/op/deconvolution.hpp
```
etc.

pfultz2:
For source files the execution bit should not be set, so they should be 644

I had a problem with my NFS share a while back which was setting the execution bit, which may have been committed by accident.

pfultz2:
@causten Can you hold off merging other PRs until this is merged?

PR #1109, opened 2022-03-01
Add initial experimental custom op

pfultz2:
This creates a custom op which has `name()` and `compute_shape()` methods. This also can be registered and constructed later. This is just the initial work but most of the infrastructure is there to extend this further as we support `context` and serialization.


test/api/test_custom_op.cpp
```diff
@@ -0,0 +1,23 @@
+#include <migraphx/migraphx.h>
+#include <migraphx/migraphx.hpp>
+#include "test.hpp"
+
+struct my_custom_op final : migraphx::experimental_custom_op_base
```
kahmed10, line 5: should we rename this to `simple_custom_op`? Then similarly in L16 rename `cop` to `simple_op`?

src/api/api.cpp
```diff
@@ -218,12 +252,60 @@ void destroy(T* x)
...
+        : data(other.data), copier(other.copier), deleter(other.deleter)
+    {
+        other.data    = nullptr;
+        other.copier  = nullptr;
+        other.deleter = nullptr;
+    }
+
+    manage_generic_ptr& operator=(manage_generic_ptr rhs)
+    {
+        std::swap(data, rhs.data);
```
umangyadav, line 286: Why is it swap and not assignment?

src/api/api.cpp
```diff
@@ -373,12 +469,41 @@ extern "C" struct migraphx_quantize_int8_options;
...
     migraphx::quantize_int8_options object;
 };
 
+extern "C" struct migraphx_experimental_custom_op;
+struct migraphx_experimental_custom_op
+{
+    template <class... Ts>
+    migraphx_experimental_custom_op(void* p,
+                                    migraphx_experimental_custom_op_copy c,
+                                    migraphx_experimental_custom_op_delete d,
```
umangyadav, line 485: should it also have `compute_shape_f`  as one of the args ? or the does it have to set explicitly by calling `set_compute_f` ?

src/api/api.cpp
pfultz2: Copy-swap is necessary for the Strong Exception Guarantee. This also allows to implement move and copy assignment with same function.

src/api/api.cpp
pfultz2: Only the copy and delete are passed since those functions are necessary to manage the object. All other functions are set using the set functions like `migraphx_experimental_custom_op_set_compute_shape`.

Set functions are used(instead of structs or parameters) so we can extend the interface without breaking ABI. Furthermore to simplify code gen, all methods must use set methods except those required to manage the objects lifetime(ie copy and delete), rather than start with the initial methods as parameters and use set methods for additional methods added in the future since that requires code gen to support two type of generations.

kahmed10 (approved):

kahmed10:
I think there are some pyflakes and other warnings that need to be resolved first before being merged in.

PR #1114, opened 2022-03-02
Celu ONNX parser and tests

CharlieL7:
* How to handle division by zero and near zero?
* Will negative alpha ever be used?


kahmed10:
> * How to handle division by zero and near zero?
> * Will negative alpha ever be used?

If the spec doesn't define what should happen if alpha <= 0, I don't think we need to worry about it. However, you can always test using onnxruntime what the expected behavior would be if you parsed in an onnx file with such alpha values.

src/onnx/parse_celu.cpp
```diff
@@ -0,0 +1,53 @@
...
+        if(contains(info.attributes, "alpha"))
+        {
+            alpha = info.attributes.at("alpha").f();
+        }
+
+        auto input_lens = args[0]->get_shape().lens();
+        auto input_type = args[0]->get_shape().type();
+        if(input_type != migraphx::shape::float_type)
+        {
+            MIGRAPHX_THROW("CELU: input tensor not float type");
```
kahmed10, line 30: The onnx spec allows the user to have an input datatype that isn't float. However, alpha is always specified as float. I think we just need to convert the alpha datatype to whatever the input datatype is.

CharlieL7: It says on the operators.md that there is a type constraint for the tensor T:
Type Constraints
T : tensor(float)
Constrain input and output types to float32 tensors.

kahmed10: Oh never mind. I scrolled down further and was reading the wrong operator.

src/onnx/parse_celu.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    instruction_ref parse(const op_desc&,
+                          const onnx_parser&,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        float alpha = 1.0;
+        if(contains(info.attributes, "alpha"))
+        {
+            alpha = info.attributes.at("alpha").f();
```
turneram, line 23: We should check that alpha != 0. 
The white paper only discusses cases where alpha > 0, but I think alpha < 0 is still technically valid too.

CharlieL7: Yeah, alpha < 0 would be a pretty bad rectifier function as it quickly goes to large negative numbers.

CharlieL7: Checking for alpha != 0 doesn't pass CI because of comparing floating point numbers with == or !=. Could use an epsilon, but that doesn't really make much sense either. Maybe a check for alpha > 0 is the most reasonable?

turneram: You should be able to just use `float_equal(alpha, 0.0f)`. I think float_equal.hpp is already indirectly included by op_parser.hpp

turneram (approved):

kahmed10 (approved):

PR #1117, opened 2022-03-03
Update gitignore to other comparable ROCm projects

CharlieL7:
Small change to .gitignore to have it ignore compiled files. Should help prevent the addition of compiled files to the repository.


kahmed10:
I am okay with updating our .gitignore file to be more generic and in line with other ROCmSoftwarePlatform .gitignore files. Maybe you can take their examples and add more to the ignore list?
https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/.gitignore
https://github.com/ROCmSoftwarePlatform/pytorch/blob/master/.gitignore
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/.gitignore

It doesn't have to be super exhaustive since not all of the technologies in the other gitignore files are relevant to MIGraphX. I would at least include more of the code editors such as vscode and other swap temporary extensions.

.gitignore
```diff
@@ -1 +1,66 @@
...
+/GTAGS
+/ID
+.gitusers
+/compile_commands.json
+
+/CMakeSettings.json
+
+# Visual studio files
+!/.vscode/extensions.json
+!/.vscode/settings_recommended.json
```
pfultz2, line 53: What does the `!` do here?

CharlieL7: ! negates the next pattern. I think these lines would actually include these files explicitly. I'll remove them for now.

pfultz2 (approved):
Are you going to remove the .pyc files already committed at some point?

turneram (approved):

.gitignore
```diff
@@ -1 +1,63 @@
...
+/compile_commands.json
+
+/CMakeSettings.json
+
+
+#==============================================================================#
+# Directories to ignore (do not add trailing '/'s, they skip symlinks).
+#==============================================================================#
+# Nested build directory
+/build*
```
turneram, line 56: I think `/doc*` would also be good to add.

.gitignore
```diff
@@ -1 +1,63 @@
...
+
```
turneram, line 57: I think it would also be good to add `/doc/html*`

PR #1120, opened 2022-03-04
Increase max groups in kernel

scxiao:
Increase the max # of groups from 128 to 1B.


kahmed10 (approved):

scxiao:
@pfultz2 Can you take a look at this PR? Thanks

pfultz2 (approved):

PR #1122, opened 2022-03-04
Size ONNX op

CharlieL7:
Size ONNX parser and tests.

Looks like the onnx_backend_test for "Size" was already activated?


src/onnx/parse_size.cpp
```diff
@@ -0,0 +1,30 @@
...
+    std::vector<op_desc> operators() const { return {{"Size"}}; }
+
+    instruction_ref parse(const op_desc&,
+                          const onnx_parser&,
+                          const onnx_parser::node_info& info,
+                          std::vector<instruction_ref> args) const
+    {
+        auto input_lens = args[0]->get_shape().lens();
+        int64_t size    = std::accumulate(
+            input_lens.cbegin(), input_lens.cend(), int64_t{1}, std::multiplies<>());
```
turneram, line 22: The shape class actually has an elements method that does this.

test/onnx/gen_onnx.py
```diff
@@ -4418,6 +4418,54 @@ def sinh_test():
...
+    node = onnx.helper.make_node(
+        'Size',
+        inputs=['x'],
+        outputs=['y'],
+    )
+    return ([node], [x], [y])
+
+
+@onnx_test
+def size_verify_test():
```
umangyadav, line 4458: IMO, you don't really need a separate onnx file for a verify test. You can use any of the previous ones. It is okay if you want to keep it. LGTM.

test/onnx/onnx_test.cpp
```diff
@@ -4276,6 +4276,49 @@ TEST_CASE(sinh_test)
...
+TEST_CASE(size_float_test)
+{
+    migraphx::program p;
+    auto* mm               = p.get_main_module();
+    auto s                 = migraphx::shape{migraphx::shape::float_type, {2, 3, 4}};
+    const auto& input_lens = s.lens();
+    mm->add_parameter("x", s);
+    int64_t size =
+        std::accumulate(input_lens.cbegin(), input_lens.cend(), int64_t{1}, std::multiplies<>());
+    mm->add_literal(migraphx::literal{migraphx::shape::int64_type, {size}});
```
turneram, line 4288: We can also just use s.elements() instead of size here to keep it simple.

turneram (approved):

PR #1124, opened 2022-03-07
scatter operator refactoring to include reduction

bpickrel:
Change the "scatter" struct and op to a base/child set of three: scatter_none, scatter_add, scatter_mul to mirror Onnx' ScatterElements op. and its three reduction options.  (Onnx Scatter op is deprecated and is equivalent to scatter_none.)

Provides both a reference op. and update to Onnx parsing.  Tests updated and new test case added.


bpickrel:
I don't seem to have permissions to resolve conflicts.

kahmed10:
> I don't seem to have permissions to resolve conflicts.

Most of us don't have permissions, including myself. However, you should see the conflicts if you do a git pull. Then you can address those conflicts and manually merge.

kahmed10:
I would remove gen_onnx.pyc since we don't include python bytecode in our commits.

test/onnx/gen_onnx.py
```diff
@@ -4117,7 +4117,7 @@ def scatter_test():
     y = helper.make_tensor_value_info('y', TensorProto.FLOAT, [3, 4, 5, 6])
 
     node = onnx.helper.make_node(
-        'Scatter',
```
kahmed10, line 4120: Can you generate more onnx files with different reduction attributes? Then you can add the tests under onnx_test.cpp.

src/onnx/parse_scatter.cpp
```diff
@@ -0,0 +1,43 @@
...
+    std::vector<op_desc> operators() const { return {{"ScatterElements"}}; }
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& /*parser*/,
+                          const onnx_parser::node_info& info,
+                          const std::vector<instruction_ref>& args) const
+    {
+        operation op;
+
+        std::string op_name = "scatter_none";
```
kahmed10, line 21: I would prefer to add another string variable called `reduction_mode` and default it to "none." Then change `op_name` to "scatter_". In L36 where `make_op` is being called you can do something like `op_name + reduction_mode`.

src/onnx/parse_scatter.cpp
```diff
@@ -0,0 +1,43 @@
...
+        int axis            = 0;
+
+        if(contains(info.attributes, "axis"))
+            axis = info.attributes.at("axis").i();
+        if(contains(info.attributes, "reduction"))
+        {
+            if(info.attributes.at("reduction").s() == "add")
```
kahmed10, line 28: Referring to the comment above, I would assign the `reduction_mode` variable to `info.attributes.at("reduction"_.s())`.  This way you can just have a single conditional statement that checks if the string is a valid attribute.

So something like this:
```suggestion
            reduction_mode = info.attributes.at("reduction").s();
            if(reduction_mode != "none" or reduction_mode != "add" or reduction_mode != "mul")
                MIGRAPHX_THROW("PARSE_SCATTER: unsupported reduction mode " + reduction_mode);
```

test/onnx/gen_onnx.py
bpickrel: Then the *.pyc file should be in .gitignore, shouldn't it?

kahmed10: yes, there is a PR with that change. But for now, you can manually remove it.

src/onnx/parse_scatter.cpp
bpickrel: discussed this at length with Paul and concluded it's not appropriate to throw an error here.  The Onnx docs don't say an unsupported reduction attribute is an error; they just say that "none" is the default, so that's what I did.

test/onnx/gen_onnx.py
bpickrel: done, commit b36bda7a96

src/onnx/parse_scatter.cpp
bpickrel: If it's just a matter of style, I prefer to keep this the way it is.  I think it's easier to follow and less prone to errors in the future.

src/include/migraphx/op/scatter_add.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <migraphx/streamutils.hpp>
+#include <migraphx/shape_for_each.hpp>
+#include <migraphx/config.hpp>
+#include <migraphx/value.hpp>
+#include <migraphx/op/normalize_attribute.hpp>
+#include <cmath>
+#include <utility>
+#include <migraphx/op/scatter.hpp>
+
+// ScatterElement op. with "add" as the reduction attribute.
```
pfultz2, line 16: This should say "scatter with an add reduction". MIGraphX doesn't have a ScatterElement op nor a reduction attribute.

src/include/migraphx/op/scatter_add.hpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    value attributes() const
+    {
+        value normalize;
+        normalize["axis"] = value::array{normalize_attribute::include_min};
+        return {{"normalize_axes", normalize}};
+    }
+
+    std::string name() const { return "scatter_add"; }
```
pfultz2, line 37: These methods above can be removed: `name`, `attributes`, and `reflect`.

src/include/migraphx/op/scatter.hpp
```diff
@@ -38,30 +46,41 @@ struct scatter
...
     }
 
     argument compute(const shape& output_shape, std::vector<argument> args) const
     {
         argument result{output_shape};
+        auto& self = static_cast<const Derived&>(*this);
+
         // max dimension in axis
         auto axis_dim_size = output_shape.lens()[axis];
+        // iterate through all elements in output
```
pfultz2, line 60: This comment seems out of place. `visit_all` doesnt iterate through the elements.

src/include/migraphx/op/scatter.hpp
```diff
@@ -38,30 +46,41 @@ struct scatter
...
+
+                    // call reduction() method of derived struct
+                    self.reduction()(output[output_shape.index(out_idx)], update[ind_s.index(idx)]);
                 });
             });
         });
 
         return result;
     }
+    auto init() const {}
```
pfultz2, line 81: Where is this used?

src/include/migraphx/op/scatter.hpp
```diff
@@ -38,30 +46,41 @@ struct scatter
...
+    scatter() {}
+    scatter(int64_t ax) : axis(ax) {}
```
pfultz2, line 83: I cant comment on the line in github, but the `name()` method in this class should be removed.

src/onnx/parse_scatter.cpp
kahmed10: > We can still throw an error. I, personally, didnt think it was important, but its not bad to have some better mesages when something goes wrong.
> 
> However, I would suggest checking the error with `if(not contains({"none", "add", "mul"}, reduction_mode))`, its a little cleaner.

Yeah that's better.

src/include/migraphx/op/scatter.hpp
```diff
@@ -33,29 +44,38 @@ struct scatter
...
     }
 
-    std::string name() const { return "scatter"; }
-
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
         check_shapes{inputs, *this}.has(3).standard();
-        return inputs.front();
+        // If non-packed, this converts to a packed output while preserving permutation of tensor
+        return inputs.front().with_lens(inputs.front().lens());
```
scxiao, line 51: This is conflict with the standard input shape requirement. We can just return the inputs.front() here.

pfultz2: Actually, `.standard()` should be removed. The ref implementations should not require standard shape whenever possible.

bpickrel: Can you clarify how the input shape requirement is set (i.e. how would I have known?)  Does it have to do with scatter being listed in parse_generic_op::needs_contiguous()?

test/ref_ops_test.cpp
```diff
@@ -4257,6 +4260,195 @@ TEST_CASE(scatter_test)
...
+        auto r  = mm->add_instruction(migraphx::make_op("scatter_mul", {{"axis", 0}}), ld, li, lu);
+        mm->add_return({r});
+        p.compile(migraphx::ref::target{});
+        auto result = p.eval({}).back();
+        std::vector<float> results_vector;
+        result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+        std::vector<float> gold_mul2 = {3.3, 21.0, 3.0, 3.0, 21.6, 3.0, 3.6, 21.3, 3.0};
+
+        EXPECT(migraphx::verify_range(results_vector, gold_mul2));
+    }
```
pfultz2, line 4449: These should be split into multiple test cases instead putting it all into the same test case.

src/include/migraphx/op/scatter.hpp
scxiao: Considering the plan that we will refactor these operators, it is fine to put a standard shape here. Otherwise, we need to revisit the implementation to ensure they are correct.

src/include/migraphx/op/scatter_add.hpp
bpickrel: Several redundant or unused methods were added in trial-and-error efforts to get past build errors--removed.

src/include/migraphx/op/scatter.hpp
bpickrel: removed

src/include/migraphx/op/scatter.hpp
scxiao: Sorry, this is the line for the standard input shape requirement 

```
check_shapes{inputs, *this}.has(3).standard();
```

This means it needs to have 3 inputs, and each one of them is with standard input shape.  You can check the [file](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/check_shapes.hpp)
for different shape requirements.

src/include/migraphx/op/scatter.hpp
```diff
@@ -33,29 +44,38 @@ struct scatter
...
+            // copy all of data to output
             std::copy(data.begin(), data.end(), output.begin());
             args[1].visit([&](auto indices) {
                 auto ind_s = indices.get_shape();
+                // iterate through items in index
                 shape_for_each(ind_s, [&](const auto& idx) {
-                    auto out_idx  = idx;
-                    auto index    = indices[ind_s.index(idx)];
+                    auto out_idx = idx;
+                    auto index   = indices[ind_s.index(idx)];
```
pfultz2, line 70: Doing `indices[ind_s.index(idx)]` is incorrect(which was done previously). The problem is that `.index()` returns an offset, but the `[]` accepts the position(not the offset). This should do `indices(idx.begin(), idx.end())` instead.

Actually, it would be even better to add overloads to tensor_view to take a range:

```cpp
    template<class Range, MIGRAPHX_REQUIRES(not std::is_integral<Range>{})>
    T& operator[](Range&& r)
    {
        return (*this)(r.begin(), r.end());
    }

    template<class Range, MIGRAPHX_REQUIRES(not std::is_integral<Range>{})>
    const T& operator[](Range&& r) const
    {
        return (*this)(r.begin(), r.end());
    }
```

Then you could just write `indices[idx]` instead.

bpickrel: > Doing `indices[ind_s.index(idx)]` is incorrect(which was done previously). The problem is that `.index()` returns an offset, but the `[]` accepts the position(not the offset). This should do `indices(idx.begin(), idx.end())` instead.
> 

You'll have to clarify.  It passes the tests the way it is.



> Actually, it would be even better to add overloads to tensor_view to take a range:
> 
> ```c++
>     template<class Range, MIGRAPHX_REQUIRES(not std::is_integral<Range>{})>
>     T& operator[](Range&& r)
>     {
>         return (*this)(r.begin(), r.end());
>     }
> 
>     template<class Range, MIGRAPHX_REQUIRES(not std::is_integral<Range>{})>
>     const T& operator[](Range&& r) const
>     {
>         return (*this)(r.begin(), r.end());
>     }
> ```
> 
> Then you could just write `indices[idx]` instead.

I think that's out of scope for this issue.

pfultz2: > You'll have to clarify. It passes the tests the way it is.

I think the tests are only using standard shape(which the offset and position are the same), but if you use non-standard shape this will break.

Ideally, in the future, I would like to have a some kind of tagged type to catch these errors at compile-time because I think there are other places where we are doing this. Of course, a tagged type checking is not for this PR though.

> I think that's out of scope for this issue.

Sure, I'll open a PR for these overloads.

bpickrel: I'm not sure we're all on the same page.  Testing for nonstandard shapes conflicts with Shucai's comment [here, above](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1124#discussion_r824999498) that we should assume standard shape (and the current code does.)  Am I misreading this, or is this introducing a change to a specified behavior that we've been planning all along? 

It's not even clear to me that using a scatter operation on a nonstandard tensor shape even makes sense from a user perspective, or if it's only helpful when starting with contiguous memory.  

Maybe this would be a good discussion topic for this afternoon's team meeting, if we can keep it short.

bpickrel: After more discussion, I'll try adding the suggested [] overloads to the tensor_view definition and see if it still passes existing tests.  In the interest of limiting the scope and time of this PR, if it needs further work to pass then we'll back out the change and write a new issue.

Also, will postpone creating tests using nonstandard shape inputs, and write a separate issue to be addressed as part of the GPU implementation of scatter/gather operators.

bpickrel: Didn't add the [] overloads; created [issue 1132](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1132#issue-1168999696) instead

pfultz2: > I'll try adding the suggested []

When you add the [] overloads then you can change `indices[ind_s.index(idx)]` to `indices[idx]` so nonstandard index calculation should be correct.

src/include/migraphx/op/scatter.hpp
```diff
@@ -33,29 +44,41 @@ struct scatter
...
+
+                    // normalize negative indexes (may be redundant after using
+                    // normalize_compute_shape())
                     index         = (index < 0) ? index + axis_dim_size : index;
                     out_idx[axis] = index;
-                    output[output_shape.index(out_idx)] = update[ind_s.index(idx)];
+
+                    // look up the appropriate location in output, using index.
+                    // call reduction() method of derived struct to copy and reduce that element
+                    self.reduction()(output[output_shape.index(out_idx)], update[ind_s.index(idx)]);
```
pfultz2, line 81: The access for `output` and `update` needs to be updated as well.

bpickrel: Actually, I just pushed it to see if it would break any other tests.  But it's really not clear how to do the corresponding indexing into `output` and `update`.

pfultz2: It would be `output(out_idx.begin(), out_idx.end())` and `update(idx.begin(), idx.end())`

bpickrel: This is done and is in the process of passing tests, but 2 new questions:  

1. the iteration value `ind_s` seen in this snippet:

```
            // copy all of data to output
                auto ind_s = indices.get_shape(); <===== HERE
```
is no longer used in a formula, only as a looping range.  Does the changed indexing call mean that we now have an extra, unneeded layer of nested looping?

2. Is my comment `                    // normalize negative indexes (may be redundant after using
                    // normalize_compute_shape())` in error?  It's not clear to me if the array modified here is the same one that was normalized with `normalize_compute_shape()`.

src/onnx/parse_generic_op.cpp
```diff
@@ -10,42 +10,23 @@ struct parse_generic_op : op_parser<parse_generic_op>
...
+                {"Flatten", "flatten"},  {"Floor", "floor"},
+                {"Gather", "gather"},    {"Identity", "identity"},
+                {"IsNaN", "isnan"},      {"LeakyRelu", "leaky_relu"},
+                {"Log", "log"},          {"LRN", "lrn"},
+                {"Neg", "neg"},          {"NonMaxSuppression", "nonmaxsuppression"},
+                {"Reciprocal", "recip"}, {"Relu", "relu"},
+                {"Round", "round"},      {"Sigmoid", "sigmoid"},
+                {"Sign", "sign"},        {"Sin", "sin"},
+                {"Sinh", "sinh"},        {"Sqrt", "sqrt"},
+                {"Tan", "tan"},          {"Tanh", "tanh"},
```
pfultz2, line 29: I am not sure why clang-formatting change it to that. It should be a single list.

test/ref_ops_test.cpp
```diff
@@ -4257,6 +4260,195 @@ TEST_CASE(scatter_test)
...
+        auto result = p.eval({}).back();
+        std::vector<float> results_vector;
+        result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+        std::vector<float> gold_mul = {1.0, 2.2, 3.0, 8.4, 5.0};
+
+        EXPECT(migraphx::verify_range(results_vector, gold_mul));
+    }
+
+    {
+        migraphx::program p;
```
kahmed10, line 4319: As @pfultz2 mentioned, you can separate these into different unit tests. Also, you can make a common function above the first three tests that constructs a program and takes in the scatter mode as a parameter:

```
migraphx::program create_scatter_program(std::string scatter_mode) {
// ... add instructions and literals here
}
```

test/onnx/onnx_test.cpp
```diff
@@ -4233,7 +4233,8 @@ TEST_CASE(round_test)
     EXPECT(p == prog);
 }
 
-TEST_CASE(scatter_test)
+// the ScatterElements op has 3 reduction modes, which map to separate reference ops
+TEST_CASE(scatter_add_test)
```
kahmed10, line 4237: Like I mentioned in test/ref_ops_test.cpp, you can create a function that creates a program where the reduction mode is passed in as a parameter.

src/onnx/parse_generic_op.cpp
bpickrel: I'll wrap this in clang-format off

test/onnx/onnx_test.cpp
bpickrel: changed

test/ref_ops_test.cpp
bpickrel: changed

src/onnx/parse_scatter.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+            if(reduction_att == "add")
+                op_name = "scatter_add";
+            else if(reduction_att == "mul")
+                op_name = "scatter_mul";
+            else
+                op_name = "scatter_none";
+        }
+
+        op = migraphx::make_op(op_name, {{"axis", axis}});
```
kahmed10, line 41:
```suggestion
        }
        // merge scatter with reduction attribute to specify which scatter operation
        op = migraphx::make_op(join_strings({op_name, reduction_att}, "_"), op_name, {{"axis", axis}});
```

bpickrel: The suggestion didn't compile and didn't account for the default case., but I made a change similar in effect.

test/ref_ops_test.cpp
bpickrel: done

kahmed10 (approved):

pfultz2 (approved):

PR #1125, opened 2022-03-07
Refactor runtime compiled kernels to use the same compile_ops pipeline

pfultz2:
This adds the infrastructure so we can compile everything in parallel, whereas before only pointwise kernels were compiled in parallel. This will also directly integrate with lowering and the gpu-driver. The kernels for pointwise and roialign are using this infrastructure. Scatternd is not since it does require standard shape. 

This also makes it easier to add new runtime compiled kernels in the future.


pfultz2:
Actually, scatternd works with non-standard, but roialign does not. PR #1126 fixes that.

test/verify/test_roialign_nonstandard.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+
+struct test_roialign_nonstandard : verify_program<test_roialign_nonstandard>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        auto x_s = migraphx::shape::from_permutation(
```
umangyadav, line 13: Should we add test for slice and broadcast as well ?

pfultz2: I dont think its necessary, I dont know what other reviewers think.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -93,6 +93,30 @@ const std::vector<std::string>& compiler_warnings()
...
+    local = v.get("local", default_local);
+    if(v.contains("global"))
+        global = v.at("global").to<std::size_t>();
+    else
+        global = compute_global(local);
+}
+
+std::function<std::size_t(std::size_t local)> compute_global_for(context& ctx, std::size_t n)
+{
+    std::size_t max_global = ctx.get_current_device().get_cu_count() *
```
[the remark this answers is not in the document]
pfultz2: For broadcasted pointwise operators we dont want to oversubscribe workgroups as it runs slower(especially on vega20) due to the preloading. Maybe we can consider adding an oversubscribe parameter in the future.

scxiao (approved):

test/verify/test_roialign_nonstandard.cpp
umangyadav: @scxiao Do you think we will encounter slice or broadcast for ROIAlign ? I doubt it.

pfultz2: Well using a transpose is enough of a test to ensure we are using the shape class to compute the indices. We dont have special cases for transpose, slice, or broadcasted gpu kernels.

umangyadav (approved):

turneram (approved):

PR #1126, opened 2022-03-08
Add iterators to kernels tensor_view and fix roialign to work with non-standard shape

pfultz2:
This adds iterators to tensor_view, which can allow kernels to work with non-standard shapes like for roialign. 

To improve the performance of indexing when using the iterators, the shape class was updated to use integral_constants since the compiler doesn't always fold the const values. An integral_constant will at least enforce that in the AST.

Finally, since index calculations with single integers are improved, I also updated pointwise to use single index rather than multi index. There is about 4% improvement in some cases.


turneram (approved):

pfultz2:
@scxiao Any feedback?

PR #1129, opened 2022-03-09
Lp normalization op

CharlieL7:
LpNormalization ONNX parser

Should we be putting some type of DOxygen documentation for these operators? The ONNX spec is pretty sparse on explaining what this does.


src/onnx/parse_lpnormalization.cpp
```diff
@@ -0,0 +1,73 @@
...
+        // change them to 1 for the element-wise division
+        auto norms = info.add_instruction(migraphx::make_op("reduce_sum", {{"axes", axis}}), p_val);
+        if(p == 2)
+        {
+            norms = info.add_instruction(migraphx::make_op("sqrt"), norms);
+        }
+        // broadcast back to initial shape, negative axis option doesn't work with unidirectional
+        norms = info.add_instruction(
+            migraphx::make_op("multibroadcast", {{"out_lens", input_lens}}), norms);
+        auto zero_mb = info.add_instruction(
```
umangyadav, line 62: I think you should be able to declare literal with final broadcasted shape directly. e.g. `info.add_literal(migraphx::shape{type, broadcasted_shape}, values)`.

test/onnx/verify_onnx.cpp
```diff
@@ -467,6 +467,60 @@ TEST_CASE(lessorequal_test)
...
+                               1.f / 5.f,
+                               1.f / 10.f,
+                               -5.f / 10.f,
+                               3.f / 10.f,
+                               -1.f / 10.f,
+                               -4.f / 7.f,
+                               3.f / 7.f,
+                               0.f,
+                               0.f};
+}
```
umangyadav, line 495: add EXPECT

test/onnx/verify_onnx.cpp
```diff
@@ -467,6 +467,60 @@ TEST_CASE(lessorequal_test)
...
+                               1.f / 3.f,
+                               1.f / 6.f,
+                               -5.f / 6.f,
+                               3.f / 6.f,
+                               -1.f / 6.f,
+                               -4.f / 5.f,
+                               3.f / 5.f,
+                               0.f,
+                               0.f};
+}
```
umangyadav, line 522: add `EXPECT` here as well.

umangyadav (changes requested):

umangyadav:
> Should we be putting some type of DOxygen documentation for these operators? The ONNX spec is pretty sparse on explaining what this does.
For now you can include a doctring explaining LpNorm in parsing file itself based your understanding.

umangyadav:
Do we need to include tests related to this in the `onnx_backend_test` ?

CharlieL7:
> Do we need to include tests related to this in the `onnx_backend_test` ?

I didn't find any tests for LpNormalization in the onnx backend git repo.

src/onnx/parse_lpnormalization.cpp
CharlieL7: I'm not able to get this or `info.add_literal(migraphx::literal{migraphx::shape{input_type, input_lens}, {0.}})` to work.

umangyadav: What's the error ?

CharlieL7: I got it working using `info.add_literal(migraphx::literal{migraphx::shape{input_type, input_lens}, {0.}})` after fixing the previous bug. There's probably a good amount of ONNX parser code that could be simplified using this format.

CharlieL7: I'm now getting the same error from before about an assertion about distance. 
```
test_verify_onnx: ../../../src/include/migraphx/literal.hpp:85: void migraphx::version_1::literal::fill(Iterator, Iterator) [with Iterator = const double*]: Assertion `std::distance(start, end) == m_shape.elements()' failed.
```
This only seems to occur in debug mode.

src/onnx/parse_lpnormalization.cpp
```diff
@@ -0,0 +1,77 @@
...
+        {
+            norms = info.add_instruction(migraphx::make_op("sqrt"), norms);
+        }
+        // broadcast back to initial shape, negative axis option doesn't work with unidirectional
+        norms = info.add_instruction(
+            migraphx::make_op("multibroadcast", {{"out_lens", input_lens}}), norms);
+        auto zero_mb =
+            info.add_literal(migraphx::literal{migraphx::shape{input_type, input_lens}, {0.}});
+        auto one_mb =
+            info.add_literal(migraphx::literal{migraphx::shape{input_type, input_lens}, {1.}});
```
turneram, line 67: The buffer that you give to the literal constructor needs to have the same number of elements as the shape. So, `std::vector<float> zeros(0, input_shape.elements())
auto zero_mb = info.add_literal(migraphx::literal{migraphx::shape{input_type, input_lens}, zeros});` 
and the same for ones should make this work.

turneram: I think this is actually the reason why multibroadcasting a scalar literal has been used elsewhere.

pfultz2: Its better to broadcast the scalar. When fusion is done it will move the scalar into kernel rather than reading from global memory. Ideally `add_common_op` could be used to automatically broadcast the scalar(I am not sure if it works in this scenario but should definitely be extended to support it if not).

umangyadav (approved):

test/onnx/verify_onnx.cpp
```diff
@@ -467,6 +467,62 @@ TEST_CASE(lessorequal_test)
...
+    migraphx::shape s{migraphx::shape::float_type, {3, 4}};
+    std::vector<float> data{0.f, 2.f, -2.f, 1.f, 1.f, -5.f, 3.f, -1.f, -4.f, 3.f, 0.f, 0.f};
+    migraphx::parameter_map pp;
+    pp["x"] = migraphx::argument(s, data.data());
+
+    auto result = p.eval(pp).back();
+    std::vector<float> result_vector;
+    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
+
+    std::vector<float> correct{0.f,
```
kahmed10, line 483: For consistency, I think we should rename this variable to `gold`

test/onnx/onnx_test.cpp
```diff
@@ -2556,6 +2556,105 @@ TEST_CASE(loop_test)
     EXPECT(p == prog);
 }
 
+TEST_CASE(lpnormalization_double_test)
```
kahmed10, line 2559: Normally there isn't a need to create onnx tests for every supported datatype of the operator unless we have a specific reason to test them. By default, we set the datatype as float.

I would, however, create 2 tests for l1 and l2 norm cases. For the axis, you can specify the attribute in any one of the tests and get code coverage.

src/onnx/parse_lpnormalization.cpp
```diff
@@ -0,0 +1,84 @@
...
+            if(axis < -num_axes or axis >= num_axes)
+            {
+                // handled in normalize_attributes but throwing here might be clearer
+                MIGRAPHX_THROW("LPNORMALIZATION: selected axis out of bounds");
+            }
+        }
+        migraphx::instruction_ref p_val;
+        if(p == 1)
+        {
+            p_val = info.add_instruction(migraphx::make_op("abs"), args.front());
```
kahmed10, line 51: `args.front()` is used frequently. I would initialize a variable such as `input` and use that instead.

kahmed10 (approved):

PR #1133, opened 2022-03-15
upgrade docker images to ROCm 5.0.2

causten:
Move to CI containers to rocm 5.0.2


Dockerfile
```diff
@@ -6,7 +6,8 @@ ARG PREFIX=/usr/local
 RUN dpkg --add-architecture i386
 
 # Add rocm repository
-RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/4.5/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
+RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.0.2/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
+RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/amdgpu/21.50.2/ubuntu bionic main > /etc/apt/sources.list.d/amdgpu.list'
```
pfultz2, line 10: Is this needed? We only install the userspace packages in the docker, and I think this doesn't have any userspace packages.

pfultz2:
Besides fixing the compiler warnings, the CI nodes need to be upgraded to rocm 5.0 before we can merge this in.

.clang-tidy
```diff
@@ -113,3 +113,5 @@ CheckOptions:
     value:           UPPER_CASE
   - key:             readability-identifier-naming.MacroDefinitionPrefix
     value:           MIGRAPHX_
+  - key:             readability-identifier-length.MinimumVariableNameLength
```
pfultz2, line 116: I think it would be better to disable this check instead.

causten: Spec says to disable set to a 0 or a 1.

pfultz2: To disable a check you a put `-` in front of the name in the list [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/CMakeLists.txt#L149).

Dockerfile
causten: removed the amdgpu installer repo

.clang-tidy
causten: Done

kahmed10 (approved):
Someone can either make a PR for #1176 or commit changes here and merge this in?

CMakeLists.txt
```diff
@@ -93,11 +93,14 @@ rocm_enable_clang_tidy(
         modernize-*
         performance-*
         readability-*
-        -bugprone-signed-char-misuse
+        -bugprone-easily-swappable-parameters
+	-bugprone-implicit-widening-of-multiplication-result
```
pfultz2, line 97: This is not formatted correctly.

cmake/EnableCompilerWarnings.cmake
```diff
@@ -96,12 +96,15 @@ else()
                 -Wno-gnu-zero-variadic-macro-arguments
                 -Wno-missing-prototypes
                 -Wno-nested-anon-types
+		-Wno-option-ignored
```
pfultz2, line 99: This should be formatted.

.github/workflows/ci.yaml
```diff
@@ -15,7 +15,9 @@ jobs:
 
     steps:
     - name: Free space
-      run: sudo rm -rf /usr/local/android /usr/share/dotnet /usr/local/share/boost /opt/ghc /usr/local/share/chrom* /usr/share/swift /usr/local/julia*
+      run: |
+              sudo rm -rf /usr/local/android /usr/share/dotnet /usr/local/share/boost /opt/ghc /usr/local/share/chrom* /usr/share/swift /usr/local/julia* /usr/local/lib/android
```
pfultz2, line 19: This line should be the same as the other lines.

cmake/EnableCompilerWarnings.cmake
```diff
@@ -96,12 +96,15 @@ else()
...
                 -Wno-padded
                 -Wno-shorten-64-to-32
                 -Wno-sign-conversion
                 -Wno-unused-command-line-argument
                 -Wno-weak-vtables
                 -Wno-c99-extensions
+                -Wno-reserved-macro-identifier
+                -Wno-reserved-identifier
```
pfultz2, line 107: These warnings should not be disabled.

CMakeLists.txt
causten: Fixed

cmake/EnableCompilerWarnings.cmake
causten: fixed

cmake/EnableCompilerWarnings.cmake
causten: fixed

pfultz2 (approved):

PR #1135, opened 2022-03-16
Python binding for shape : Fix constructor for the shape and enable tests 

umangyadav:
Follow up to #1128


src/include/migraphx/value.hpp
```diff
@@ -122,6 +122,7 @@ struct value
 // clang-format off
 #define MIGRAPHX_VISIT_VALUE_TYPES(m) \
     m(int64, std::int64_t) \
+    m(uint8, std::uint8_t) \
```
pfultz2, line 125: Why is this added?

pfultz2: I would prefer to have only one signed and one unsigned integer type in the value class.

umangyadav: It was required to run the python test with type = "uint8".

umangyadav: without having that, it is not possible to convert value object of "uint8" type to string object.  string is then passed to 'shape::parse_type()`.

pfultz2: > convert value object of "uint8" type to string object

I am not sure I understand. How are you creating a value object of "uint8" when it doesn't exist?

umangyadav: For example : 
s = migraphx.shape(type='uint8_t')

in `migraphx_py.cpp` this is passed in as `py::kwargs` and those kwargs are converted into value object where type is one of the memeber. {{"type", "uint8"}}.  uint8 is converted into string from there and passed to parse_type(). In order to convert to string, it requires `uint8` to be added to one of the `MIGRAPHX_VISIT_VALUE_TYPES`.

pfultz2: >  uint8 is converted into string from there and passed to parse_type(). In order to convert to string, it requires uint8 to be added to one of the MIGRAPHX_VISIT_VALUE_TYPES.

It shouldn't need `uint8` to convert to string. It should already be converted to a string in `visit_py`.

pfultz2:
Hmm, I found some issues with value class handling string literals which I fixed in PR #1141. I dont know if this will fix the issue you are having.

umangyadav:
> Hmm, I found some issues with value class handling string literals which I fixed in PR #1141. I dont know if this will fix the issue you are having, but hopefully it will help.

Hi, yes that PR does fixes the problem with shape constructor here as well. I don't think we need this PR. #1141 is sufficient if you can just add lines to run python test there.

src/include/migraphx/value.hpp
umangyadav: > > uint8 is converted into string from there and passed to parse_type(). In order to convert to string, it requires uint8 to be added to one of the MIGRAPHX_VISIT_VALUE_TYPES.
> 
> It shouldn't need `uint8` to convert to string. It should already be converted to a string in `visit_py`.

"uint8" was converted into string in `visit_py`.  But for some reason during `visit` in `value.hpp`, `this->get_type()` resulted in `uint8_type` instead of `string_type`.  I am not sure why it was that way.  Adding `uint8` into `MIGRAPHX_VISIT_VALUE_TYPES` then added case for `uint8_type` and everything worked for some reason.

umangyadav: Removed that change from `value.hpp`.  Please disregard comments i made about the visit etc in value.hpp.

src/py/migraphx_py.cpp
```diff
@@ -214,7 +214,7 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
         .def(py::init([](py::kwargs kwargs) {
             auto v    = migraphx::to_value(kwargs);
             auto t    = migraphx::shape::parse_type(v.get("type", std::string{"float"}));
-            auto lens = v.get<std::size_t>("lens", {1});
+            auto lens = v.get<std::size_t>("lens", std::vector<size_t>{1});
```
pfultz2, line 217: I think with #1141 this is not necessary.

umangyadav: Yes, I rebased with respect to `value_string_literal`.  PR should be merged after #1141.

kahmed10 (approved):

pfultz2 (approved):

PR #1140, opened 2022-03-21
Update developer overview, fix doc CMakeLists

CharlieL7:
* Updates the developer overview in the docs
* Adds examples from the overview as a test
* Fixes compiling of the docs
   - Changes from #1088 made docs CMakeLists a separate project, include directory path was incorrect
* Make a DoxygenWarningLog.txt
* ~~Move compiled docs to <build_dir>/doc/[pdf, html]/~~ Reverted because github.io set to look at doc/ folder.


doc/src/dev_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+-------------------------
+
+The ``ref_dev_examples.cpp`` can be found in the test directory (``/test``).
+The executable file ``test_ref_dev_examples`` based on this file will be created in the ``bin/`` of the build directory after running ``make -j$(nproc) check``.
+
+
+Adding Two Literals
+--------------------
+
+A program consists of a set of instructions to be executed when calling `eval <migraphx::program::eval>`.
```
pfultz2, line 18: A program is a collection of modules which are a collection of instructions.

doc/src/dev_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+    pp["W"] = migraphx::argument(weights_shape, c.data());
+
+    // Evaluate and confirm the result
+    auto result = p.eval(pp).back();
+    std::vector<float> results_vector(64);
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    EXPECT(migraphx::verify_range(results_vector, sol));
+
+An `argument <migraphx::argument>` can handle memory buffers from either the GPU or the CPU, but when running the `program <migraphx::program>`, buffers should be allocated for the corresponding target.
+That is, when compiling for the CPU, the buffers should be allocated on the CPU, and when compiling for the GPU the buffers should be allocated on the GPU.
```
pfultz2, line 126: Actually, if you compile for the gpu with `offload_copy=true` then the buffers should be on the CPU.

doc/src/dev_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+
+MIGraphX provides an optimized execution engine for deep learning neural networks.
+In this introduction for developers of MIGraphX, we will cover some simple operations in the MIGraphX framework.
+
+
+Location of the Examples
+-------------------------
+
+The ``ref_dev_examples.cpp`` can be found in the test directory (``/test``).
+The executable file ``test_ref_dev_examples`` based on this file will be created in the ``bin/`` of the build directory after running ``make -j$(nproc) check``.
```
kahmed10, line 12: I don't think we should use make check if the user only wants to run the dev examples.

doc/src/dev_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+    auto two = mm->add_literal(2);
+    mm->add_instruction(migraphx::make_op("add"), x, two);
+    p.compile(migraphx::ref::target{});
+
+This adds a parameter of type ``int32``, and compiles it for the CPU.
+To run the program, we need to pass the parameter as a ``parameter_map`` when we call `eval <migraphx::program::eval>`.
+We create the ``parameter_map`` by setting the ``x`` key to an `argument <migraphx::argument>` object with an ``int`` data type::
+
+    std::vector<int> data = {4};
+    migraphx::parameter_map pp;
```
kahmed10, line 78: I would rather this variable be called `params`.

kahmed10:
Since you are changing doc/CMakeLists.txt, can you also modify L66 so that it says `MIGraphX` instead of `MIGraph`?

CharlieL7:
I tried to make it more clear that this documentation is meant for people who want to make changes to the codebase. Let me know if this works.

doc/src/dev_intro.rst
```diff
@@ -0,0 +1,152 @@
+MIGraphX Fundamentals
+======================
+
+MIGraphX provides an optimized execution engine for deep learning neural networks.
+We will cover some simple operations in the MIGraphX framework here.
+We created this documentation for contributors to the MIGraphX codebase.
+This documentation is not intended for users of the MIGraphX library.
```
kahmed10, line 7: I don't think we should say this. Instead, I would suggest to point those looking to quickly get started using MIGraphX to the examples: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/tree/develop/examples/migraphx

doc/src/dev_intro.rst
```diff
@@ -0,0 +1,152 @@
...
+
+
+Location of the Examples
+-------------------------
+
+The ``ref_dev_examples.cpp`` can be found in the test directory (``/test``).
+The executable file ``test_ref_dev_examples`` based on this file will be created in the ``bin/`` of the build directory after running ``make -j$(nproc) test_ref_dev_examples``.
+The executable will also be created when running ``make -j$(nproc) check``, alongside with all the other tests.
```
kahmed10, line 15: Why don't we add a sequence of commands such as:
```
cd build
```
```
make -j$(nproc) test_ref_dev_examples
./bin/test_ref_dev_examples
```
OR
```
make -j$(nproc) check
```

test/ref_dev_examples.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <migraphx/verify.hpp>
+#include <migraphx/ref/target.hpp>
+#include <migraphx/onnx.hpp>
+#include "test.hpp"
+
+/*!
+ * Example MIGraphX programs for following the Contributor's Guide.
+ */
+
+void add_two_literals()
```
kahmed10, line 16: We might not need to do this now, but it would be good to wrap this around the `TEST_CASE()` macro so that the user can pass in specific functions rather than running everything.

kahmed10:
In general, we are missing a lot of details that would make the developer documentation useful. I'm okay with this as a first pass, but we need at least descriptions of the following:

- current compiler passes within MIGraphX
- how to add compiler passes (subsequently, how to manipulate the current program and insert/replace instructions)
- how to use multiple modules
- how to add new operators
- how the onnx and tf parsers work
- how serialization works
- how mathcers work
- debugging


@pfultz2 may have more suggestions to add to this list.

CharlieL7:
Agreed, those sections are in the planning stages. This PR is for making the examples seen previously on the github.io page work.

doc/src/dev_intro.rst
CharlieL7: There might be a clarity issue where they don't have a build directory or it's named something else. May instead want to link back to the Docker or rbuild instructions.

kahmed10: You can link back to the build instructions if needed, but in L13-15 you already mention a build directory. I would just add the commands and say something like "assuming a build directory is already created:..."

kahmed10 (approved):

pfultz2 (approved):

doc/src/dev_intro.rst
```diff
@@ -0,0 +1,152 @@
+MIGraphX Fundamentals
+======================
+
+MIGraphX provides an optimized execution engine for deep learning neural networks.
+We will cover some simple operations in the MIGraphX framework here.
+For a quick start guide to using MIGraphX, look in the examples directory: ``https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/tree/develop/examples/migraphx``.
```
pfultz2, line 6: Not a change for this PR, but we should open an issue to move this example quick start guide to our documentation.

PR #1141, opened 2022-03-23
Improve handling of string literals in value class

pfultz2:
String literal which decay to `const char*` would get picked as a boolean in some cases(due to overload resolution rules in C++). This fixes these issues so it will be treated as a `string` type. Added more unit tests for these cases.

### DIFF
--- a/src/include/migraphx/value.hpp
+++ b/src/include/migraphx/value.hpp
@@ -189,8 +189,11 @@ struct value
     const cpp_type* if_##vt() const;
     MIGRAPHX_VISIT_VALUE_TYPES(MIGRAPHX_VALUE_GENERATE_DECL_METHODS)
 
-    template<class T>
-    using literal_to_string = std::conditional_t<(std::is_convertible<T, const char*>{} and std::is_convertible<T, std::string>{}), std::string, T>;
+    template <class T>
+    using literal_to_string = std::conditional_t<(std::is_convertible<T, const char*>{} and
+                                                  std::is_convertible<T, std::string>{}),
+                                                 std::string,
+                                                 T>;
 
     template <class T>
     using pick_numeric = std::conditional_t<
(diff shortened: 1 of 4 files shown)


src/include/migraphx/value.hpp
```diff
@@ -370,11 +378,11 @@ struct value
     }
 
     template <class To>
-    To value_or(const To& default_value) const
+    literal_to_string<To> value_or(const To& default_value) const
```
umangyadav, line 381: Unrelated but on line 360 in value.hpp inside `visit_value()`,  `MIGRAPHX_VALUE_GENERATE_CASE_VALUE` is unused in that method.  Do we need those cases to be generated as well ?

src/include/migraphx/value.hpp
```diff
@@ -188,6 +189,12 @@ struct value
     const cpp_type* if_##vt() const;
     MIGRAPHX_VISIT_VALUE_TYPES(MIGRAPHX_VALUE_GENERATE_DECL_METHODS)
 
+    template <class T>
+    using literal_to_string = std::conditional_t<(std::is_convertible<T, const char*>{} and
```
umangyadav, line 193: It is returning either string or T. I can't think of better alternatives. Should it be named `literal_to_type` ? or `literal_to_T`  or `lit_to_str_T` ?

scxiao (approved):

src/include/migraphx/value.hpp
pfultz2: If its a string literal its converted to std::string otherwise it just return T.

src/include/migraphx/value.hpp
pfultz2: Its unused because I pass `MIGRAPHX_VALUE_GENERATE_CASE` instead of ` MIGRAPHX_VALUE_GENERATE_CASE_VALUE`? That is definitely a copy-paste mistake that should be fixed.

PR #1142, opened 2022-03-24
parallelize the ref implementation of the gemm operator

scxiao:
ref implementation of the gemm op is sequential, this PR is to parallelize the gemm computation in the ref implementation.


pfultz2:
This might improve ref but it wont improve const propagation. You need to do this similar change in src/include/migraphx/gemm.hpp.

turneram (approved):

pfultz2 (approved):

PR #1145, opened 2022-03-28
Rename pointwise ops

kahmed10:
For clarity on kernel names found when profiling. The new names are set to the order of the ops being compiled. For example: add + relu = add_relu_kernel.


pfultz2:
This should be based on #1125.

pfultz2:
I think it would be better to just have a function to iterator over operators rather than do this in the generator:

```cpp
std::vector<std::string> get_op_names(module& m)
{
    std::vector<std::string> result;
    for(auto& ins:m)
    {
        if (starts_with(ins->name(), "@"))
            continue;
        result.push_back(ins->name());
    }
    return result;
}
```

causten:
we are doing more profiling lately.  I feel this will help.  You should push to get this in

kahmed10:
> we are doing more profiling lately. I feel this will help. You should push to get this in

Noted. Will prioritize and open for review soon.

kahmed10:
> I think it would be better to just have a function to iterator over operators in a module rather than do this in the generator:
> 
> ```c++
> std::vector<std::string> get_op_names(module& m)
> {
>     std::vector<std::string> result;
>     for(auto& ins:m)
>     {
>         if (starts_with(ins->name(), "@"))
>             continue;
>         result.push_back(ins->name());
>     }
>     return result;
> }
> ```

changed

src/targets/gpu/jit/pointwise.cpp
```diff
@@ -131,12 +143,20 @@ struct pointwise_compiler : compiler<pointwise_compiler>
...
+        else
+            op_name_string += "_kernel";
+        options.kernel_name = op_name_string;
         options.set_launch_params(v,
                                   compute_global_for(ctx,
                                                      options.output.elements() / vec_size,
                                                      oversubscribe_if(not is_preloading)));
         auto src = interpolate_string(pointwise_kernel,
-                                      {{"params", enum_params(inputs.size(), "void * private_p")},
+                                      {{"op_names", op_name_string},
```
pfultz2, line 158: Just name the key `kernel` such as `{"kernel", v.get("kernel", "kernel")}`.

src/targets/gpu/jit/pointwise.cpp
```diff
@@ -167,8 +187,11 @@ struct pointwise_compiler : compiler<pointwise_compiler>
         auto name = g.create_function(
             g.generate_module(*pm).set_attributes({"__device__"}).set_generic_types(*pm));
         std::string lambda = "MIGRAPHX_LIFT(" + name + ")";
+        auto op_names      = get_op_names(*pm);
```
pfultz2, line 190: Do the `join_strings` here and pass the kernel name in.

src/targets/gpu/jit/pointwise.cpp
pfultz2: And set the kernel name with `options.kernel_name = v.get("kernel", "kernel")` with the first `"kernel"` being the key and the second is the fallback value if its missing.

The reason for the fallback is that we may not pass this in when running from the gpu-driver.

src/targets/gpu/jit/pointwise.cpp
kahmed10: fixed

pfultz2 (approved):

turneram (approved):

PR #1146, opened 2022-03-28
Fix comparisons in migraphx::value class

pfultz2:


scxiao (approved):

umangyadav (approved):

src/value.cpp
```diff
@@ -417,33 +418,24 @@ value value::with_key(const std::string& pkey) const
...
 bool compare(const value& x, const value& y, F f)
 {
     bool result = false;
     x.visit_value([&](auto&& a) {
         y.visit_value([&](auto&& b) {
-            result = compare_common_impl(rank<1>{}, f, x.get_key(), a, y.get_key(), b);
+            if constexpr(std::is_same<decltype(a), decltype(b)>{})
+                result = f(std::forward_as_tuple(x.get_key(), compare_decay(a)),
+                           std::forward_as_tuple(y.get_key(), compare_decay(b)));
+            else
```
umangyadav, line 437: Would it be possible to add small test to cover this line ?

pfultz2: This line should never be reached that is why there is an `assert(false)`.

umangyadav: Yes i know, Just that, codecov is complaining. It should be okay otherwise.

Issue #1083, opened 2022-02-09
Use fast_gelu when fast_math enabled


pfultz2:
This should probably be done in simplify_algebra. There is already the [`gelu_tanh`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/match/gelu_tanh.hpp#L43) matcher which can find gelu from its decomposed operators.

causten:
First test to see if this change is now faster.

bpickrel:
I will take this one but can we have a description of the issue?

pfultz2:
So there are two versions of gelu: `gelu_tanh` and `gelu_erf`. Using `erf` should be faster(although less accurate), but we first need to investigate if its faster. If its faster we need to find `gelu_tanh` and replace it with `gelu_erf` only when `fast_math` is enabled.

To investigate the performance, it would be best to create onnx files with both versions and time them, but we need to make sure we are calling the float version of `erf` in [math.hpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/math.hpp). @kahmed10 Can you help with creating these onnx files?

bpickrel:
The real task for this issue:  Move the matcher from fuse_ops to simplify_algebra, update replacement too.  The matcher in fuse_ops.cpp already does the fast_math check.  line 400

pfultz2:
> update replacement too.

Yea, the replacement wont be done as a single operator. You will need to insert multiple operators. You can use the device function [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/device/gelu.cpp#L15) as reference for the formula.

bpickrel:
[...]

PR #1149, opened 2022-03-30
Fast_math matcher for gelu operation

bpickrel:
completes implementation of the gelu matcher in simplify_algebra.  The gelu function is a series of math instructions, not a built-in operator, and if the matcher detects the version of gelu based on the tanh function it replaces with an approximation based on erf.  This can be enabled/disabled by a calling flag fast_math.  [Issue 1083](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1083)


src/simplify_algebra.cpp
```diff
@@ -185,6 +187,41 @@ struct find_mul_add
...
+struct find_gelu
+{
+    // compile option fast_math controls whether to use math shortcuts like this
+    bool fast_math = true;
+
+    auto matcher() const { return match::gelu_tanh(); }
+
+    // apply the gelu_erf formula: return x * 0.5 * (1 + ::erf(x * M_SQRT1_2));
+    //  = x * (0.5 + 0.5 * ::erf(x * M_SQRT1_2)) after match/transform
+    void apply(module& p, match::matcher_result r) const
```
pfultz2, line 201: This should be `module& m` instead of `module& p`.

src/simplify_algebra.cpp
```diff
@@ -185,6 +187,41 @@ struct find_mul_add
     }
 };
 
+// find gelu (a multi-step formula), using the tanh implementation. Replace it with the gelu_erf
+// formula
+struct find_gelu
+{
+    // compile option fast_math controls whether to use math shortcuts like this
+    bool fast_math = true;
```
pfultz2, line 195: This variable is not used. It also should be `false` by default. You need to check this in `apply` and return early if its false.

src/include/migraphx/simplify_algebra.hpp
```diff
@@ -14,6 +14,8 @@ struct module;
  */
 struct simplify_algebra
 {
+    // flag to use shortcuts for some math functions
+    bool fast_math = true;
```
pfultz2, line 18: This should be `false` by default.

test/simplify_algebra_test.cpp
```diff
@@ -749,6 +751,67 @@ TEST_CASE(simplify_rsqrt)
...
+        auto point_5    = m2.add_literal(0.5F);
+        auto p5_erf_ins = migraphx::add_common_op(m2, migraphx::make_op("mul"), {point_5, erf_ins});
+        auto p5_add_ins =
+            migraphx::add_common_op(m2, migraphx::make_op("add"), {point_5, p5_erf_ins});
+        m2.add_instruction(migraphx::make_op("mul"), p5_add_ins, x_ins);
+    }
+    // TODO:  This only succeeds if the arguments in binary ops such as add and mul come in
+    // identical order; in addition, the result after run_pass is the result of multiple passes, not
+    // of find_gelu alone.  Result is that test cases are very fragile if you change anything in
+    // simplify_algebra at all.
```
pfultz2, line 811: Actually to avoid the fragility, it just occured to me, you could just check that the transformed program has `erf` and doesnt have `tanh`. We have done that in past for other tests like here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/eliminate_contiguous_test.cpp#L181

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/simplify_reshapes_test.cpp#L326

pfultz2:
You need a test with `fast_math` enabled and another with it not enabled.

src/simplify_algebra.cpp
bpickrel: changed

src/include/migraphx/simplify_algebra.hpp
bpickrel: fixed

test/simplify_algebra_test.cpp
bpickrel: Done, 1 rewritten test and 2 new ones

kahmed10:
Pending verification on bert model which uses gelu softmax variant.

bpickrel:
Not ready yet!  A close review of program created from a bert model file found that this matcher is substituting incorrect content!  Even though it passes the unit tests.  This would presumably lead to incorrect results any time a bert model is used with the fast_math option, so needs to be fixed.

bpickrel:
@pfultz2 and I concluded that matching the full gelu formula just couldn't be made reliable because the other simplify_matcher matchers rearrange the instructions in unpredictable ways.  So I added a standalone matcher within struct find_gelu which matches only a subset of the instruction sequence found in the earlier matcher, a sequence that hopefully won't ever be matched by accident.  I left the changes in gelu_tanh matcher because they included a fix to a math mistake I'd found earlier and are thoroughly tested, but that matcher is no longer the one invoked in simplify_algebra_test

The new matcher matches correctly both in the unit test and when processing the full bert model.

bpickrel:
clicked the wrong button--not closed!

src/simplify_algebra.cpp
```diff
@@ -185,6 +187,97 @@ struct find_mul_add
...
+            return;
+        auto ins   = r.result;
+        bool trace = enabled(MIGRAPHX_TRACE_MATCHES{});
+
+        if(trace)
+        {
+            std::cout << "module before gelu_tanh substitution:\n";
+            m.debug_print();
+            std::cout << "---end\n";
+        }
```
pfultz2, line 255: The debug statements can be removed.

src/onnx/parse_generic_op.cpp
```diff
@@ -29,6 +29,8 @@ struct parse_generic_op : op_parser<parse_generic_op>
                 {"Floor", "floor"},
                 {"Gather", "gather"},
                 {"GatherND", "gathernd"},
+                {"Gelu", "gelu"},
+                {"GatherND", "gathernd"},
```
pfultz2, line 33: These should be removed. GatherND is a duplicate, and the gelu is wrong as we dont have a gelu operator.

src/include/migraphx/match/gelu_tanh.hpp
```diff
@@ -39,6 +89,11 @@ auto gelu_tanh(F f)
 {
     return detail::gelu_tanh_matcher<F>{f}.matcher();
 }
+template <class F>
+auto gelu_tanh_inner(F f)
```
pfultz2, line 93: I think this function is no longer used so we can just remove it.

bpickrel: done

src/onnx/parse_generic_op.cpp
bpickrel: done

src/simplify_algebra.cpp
bpickrel: done

src/include/migraphx/match/gelu_tanh.hpp
```diff
@@ -13,23 +13,73 @@ template <class F>
...
-                        either_arg(0, 1)(any().bind("x"),
-                                         f("add")(any_arg(0, 1)(f("mul")(
-                                             either_arg(0, 1)(has_value(0.5f), tanh_fn()))))));
+        // clang-format off
+        // Formula used in gelu.cpp:   0.5 * x * (1 + tanh(sqrt(2 / pi) * (x + 0.044715 * pow(x, 3))))
+        //      but we only match the subexpression without the 0.5 factor:
+        //      x * (1 + tanh(sqrt(2 / pi) * (x + 0.044715 * pow(x, 3))))
+        //      because that's how it appears in common bert models.
+        //
+        // matcher grammar:
```
kahmed10, line 55: I would prefer these grammar comments go in matcher.hpp. Or if there are portions that are specific to this matcher, you can point them out. Otherwise, a simple comment to refer to matcher.hpp would suffice.

src/include/migraphx/simplify_algebra.hpp
```diff
@@ -14,6 +14,8 @@ struct module;
  */
 struct simplify_algebra
 {
+    // flag to use shortcuts for some math functions
```
kahmed10, line 17: I think it'll be better to refer to the specific math function we support currently, which is fast_gelu. Then if more are introduced, we can either comment on where to find all of the fast math functions we support or just enumerate them here.

src/include/migraphx/matcher.hpp
```diff
@@ -280,6 +280,20 @@ match::matcher_result find_match(module& modl, M&& m)
...
+// substituting
+//   it may not improve performance).
+// bind("x") makes an instruction "x" findable by substitution code such as x_ins =
+// r.instructions["x"];
+//   Behavior is undefined if you bind a name to more than 1 instruction..
+// any() matches anything; the only purpose is to get a handle for
+//   another instruction such as bind()
+// any_arg() finds a match in any argument position
+// either_arg() must match both arguments but in either order is OK
+// name() and name_contains() match if name matches
```
pfultz2, line 295: The documentation for each matcher should go with the matcher its documenting, not here.

src/include/migraphx/matcher.hpp
```diff
@@ -280,6 +280,20 @@ match::matcher_result find_match(module& modl, M&& m)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_MATCHES)
 
 /// Find matches for an instruction in the module
+//
+// matcher grammar:
+// used_once() means only match an instruction whose output is used only once (otherwise,
+// substituting
+//   it may not improve performance).
+// bind("x") makes an instruction "x" findable by substitution code such as x_ins =
+// r.instructions["x"];
```
pfultz2, line 289: This is very confusing. `.bind` doesn't make the instruction findable, it just associates a name to the instruction. It returns a matcher as well(so can be further composable). Also, there is no explanation what `r` is in this comment.

This documentation should go with the `bind` method in the `basic_matcher` class(there is also the `bindable_matcher` class but I plan to remove that in the future so dont worry about adding the documentation there).

bpickrel:
To test with and without the feature, use --disable-fast-math flag on the command line.  Run a bert model, since it's the one for which the gelu function was devised.  The following sample command line runs a bert model with some data fills that are necessary to keep bert models from crashing:

`bin/driver perf ../frozen_bert_mrpc.pb --fill1 input_ids_1 --fill1 input_mask_1 --fill1 segment_ids_1 --disable-fast-math`

PR #1152, opened 2022-03-30
Refactor Pooling and implement ONNX LpPool and GlobalLpPool

CharlieL7:
* Refactored the reference implementation of pooling to something like what was done for roialign. Moved the reference implementation of pooling from targets/ref/lowering.cpp to pooling.hpp.
* Removed cpu_pooling, instead using reference pooling in pooling.hpp
* Added reference implementation of Lp Norm pooling and the global version
* Added tests for the Lp Norm Pooling
* I did not find any ONNX backend tests for LpPool


src/include/migraphx/op/pooling.hpp
```diff
@@ -89,6 +92,135 @@ struct pooling
...
+                    int end = std::min(start + lengths[d_2], in_lens[dim]);
+                    start   = std::max(start, 0);
+                    win_start.push_back(start);
+                    win_size.push_back(end - start);
+                }
+
+                shape win_shape{output_shape.type(), win_size};
+                auto pool_size = win_shape.elements();
+                double output_val;
+                switch(mode)
```
pfultz2, line 197: The switch statement should be outside of the `par_for`.

src/include/migraphx/op/pooling.hpp
```diff
@@ -89,6 +92,135 @@ struct pooling
...
+            return std::numeric_limits<T>::lowest();
+        }
+
+        double operator()(double x, double y) { return std::max(x, y); }
+
+        double final(double x, std::size_t) { return (x); }
+    };
+
+    template <typename TOut, typename TIn, typename Op>
+    double calc_pooling(const TIn& input,
```
pfultz2, line 143: The `calc_pooling` should contain the entire `par_for` loop.

src/include/migraphx/op/pooling.hpp
```diff
@@ -89,6 +92,135 @@ struct pooling
...
+        {
+            return std::numeric_limits<T>::lowest();
+        }
+
+        double operator()(double x, double y) { return std::max(x, y); }
+
+        double final(double x, std::size_t) { return (x); }
+    };
+
+    template <typename TOut, typename TIn, typename Op>
```
pfultz2, line 142: Dont prefix template parameter with `T`. We know its a template parameter because it uses CamelCase.

I also prefer to use `class` just because its shorter, but you dont have to change that if you dont like it.

src/include/migraphx/op/pooling.hpp
pfultz2: This will reduce the number of parameters needed to pass. It should only need the input, output tensor and the `Op`. Everything else can be calculated in this function before `par_for`.

src/include/migraphx/op/pooling.hpp
CharlieL7: I will make the change to maintain consistency.

src/onnx/parse_pooling.cpp
```diff
@@ -19,22 +19,27 @@ struct parse_pooling : op_parser<parse_pooling>
...
+        std::string mode                     = opd.op_name;
+        if(std::find(valid_modes.begin(), valid_modes.end(), mode) == valid_modes.end())
         {
-            MIGRAPHX_THROW("onnx pooling mode must be \"max\" or \"average\"");
+            MIGRAPHX_THROW("onnx pooling mode must be [\"max\", \"average\", \"lpnorm\"]");
         }
-        operation op = make_op(
-            "pooling",
-            {{"mode", mode == "average" ? op::pooling_mode::average : op::pooling_mode::max}});
+        const std::unordered_map<std::string, op::pooling_mode> mode_map = {
```
umangyadav, line 38: You can just search over `mode_map` for valid_modes. That would eliminate need of `valid_modes`.

umangyadav (approved):

pfultz2 (changes requested):

src/onnx/parse_pooling.cpp
```diff
@@ -29,16 +29,15 @@ struct parse_pooling : op_parser<parse_pooling>
...
-        if(std::find(valid_modes.begin(), valid_modes.end(), mode) == valid_modes.end())
-        {
-            MIGRAPHX_THROW("onnx pooling mode must be [\"max\", \"average\", \"lpnorm\"]");
-        }
         const std::unordered_map<std::string, op::pooling_mode> mode_map = {
             {"max", op::pooling_mode::max},
             {"average", op::pooling_mode::average},
             {"lpnorm", op::pooling_mode::lpnorm}};
+        std::string mode = opd.op_name;
+        if(mode_map.find(mode) == mode_map.end())
```
pfultz2, line 37: I think you can write `if (not contains(mode_map, mode)`.

CharlieL7: That looks to be a C++20 addition: https://en.cppreference.com/w/cpp/container/unordered_map/contains. I think we're using C++17?

pfultz2: I am referring to the `contains` we have function here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/ranges.hpp#L95

pfultz2 (approved):

turneram (approved):

PR #1153, opened 2022-03-31
Change the doc to mention only gpu or ref as targets

umangyadav:


src/register_target.cpp
```diff
@@ -1,5 +1,6 @@
-#include <migraphx/register_target.hpp>
+#include <algorithm>
```
pfultz2, line 1: I dont think this additional header is needed.

src/register_target.cpp
```diff
@@ -11,7 +12,16 @@ std::unordered_map<std::string, target>& target_map()
 }
 
 void register_target(const target& t) { target_map()[t.name()] = t; }
-target make_target(const std::string& name) { return target_map().at(name); }
+
+target make_target(const std::string& name)
+{
+    if(target_map().find(name) == target_map().end())
+    {
+        MIGRAPHX_THROW("Requested target " + name + " is not enabled or not supported");
```
pfultz2, line 20: It would be good to put single quotes around the target name.

src/register_target.cpp
umangyadav: Removed.

src/register_target.cpp
```diff
@@ -11,7 +12,16 @@ std::unordered_map<std::string, target>& target_map()
...
+    }
+    return target_map().at(name);
```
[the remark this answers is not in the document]
umangyadav: Done.

doc/src/reference/py.rst
```diff
@@ -121,7 +121,10 @@ target
 
     Constructs the target.
 
-    :param str name: The name of the target to construct. This can either be 'cpu' or 'gpu'.
+    :param str name: The name of the target to construct. MIGraphx supports `cpu`, `gpu` and `ref` as targets. `ref` provides referecnce
```
pfultz2, line 124: I actually dont think we should try to iterate all the targets here. We should probably have a different page to describe all of our targets.

doc/src/reference/py.rst
```diff
@@ -121,7 +121,10 @@ target
 
     Constructs the target.
 
-    :param str name: The name of the target to construct. This can either be 'cpu' or 'gpu'.
+    :param str name: The name of the target to construct. MIGraphx supports `cpu`, `gpu` and `ref` as targets. `ref` provides referecnce 
+    implementation to compare results for correctness against `cpu` or `gpu` targets.  By default `MIGraphX` that ships with `ROCm`, `cpu` target is disabled. 
+    So, If you are using `MIGraphX` from `ROCm`, then parameter `name` can only be either `gpu` or `ref`.
```
pfultz2, line 126: I dont understand what you are trying to link to. I dont think many of them are valid sphinx targets. Maybe you could make the role explicit to make it clearer.

doc/src/reference/py.rst
pfultz2: For a quick fix, just list `gpu` and `ref`, and then we can open an issue to to add a page for backend targets.

src/register_target.cpp
umangyadav: Done

test/targets.cpp
```diff
@@ -9,6 +9,8 @@ TEST_CASE(make_target)
     {
         auto t = migraphx::make_target(name);
         CHECK(t.name() == name);
+        // make invalid target
+        EXPECT(test::throws([&] { migraphx::make_target(name + name); }));
```
pfultz2, line 13: Just put this in a seperate test case. It doesnt make sense inside of a loop.

doc/src/reference/py.rst
umangyadav: Opend #1154

test/targets.cpp
```diff
@@ -12,6 +12,11 @@ TEST_CASE(make_target)
     }
 }
 
+TEST_CASE(make_invalid_target) {
+    EXPECT(test::throws([&] { migraphx::make_target("mi100"); }));
+    EXPECT(test::throws([&] { migraphx::make_target("epyc"); }));
```
turneram, line 17: A single invalid target is probably sufficient

pfultz2 (approved):

test/targets.cpp
umangyadav: Okay. Only kept one.

turneram (approved):

PR #1155, opened 2022-04-01
fix a bug in create tensor_view with vec data type

scxiao:
When create a tensor_view with vector date type, the last dimension of the shape should be divided by the vec_size.


src/targets/gpu/device/include/migraphx/gpu/device/visit.hpp
```diff
@@ -176,7 +176,12 @@ template <index_int N, class T, class... Ts>
 auto hip_vec_visit_all(T&& x, Ts&&... xs)
 {
     return [&](auto f) {
-        hip_visit_all_impl(get_shape(x),
+        auto sx   = get_shape(x);
+        auto lens = sx.lens();
+        assert(lens.back() % N == 0);
```
pfultz2, line 181: Actually you need to check the stride is 1 as well.

scxiao: fixed

kahmed10 (approved):

pfultz2 (approved):

Issue #1127, opened 2022-03-08
Add half2 overloads to math.hpp


causten:
Also half4 /w split to call half2

Should also check what does hip provide in the hip/math_functions.h and hip/hip_fp16.h to see if additional overloads could be used.

bpickrel:
rocm's hip files eg. /opt/rocm/hip/include/hip/hip_fp16.h

pfultz2:
This is to update the [math.hpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/math.hpp) header used by the kernels. We need to add overloads for the math functions that have half2 versions in the first part.

The second part, we need to add overloads for half4 and split it into 2 half2 vectors and then call the half2 versions, and then merge it back to half4. For packing/splitting the vectors, we should have a `vec_packed_transform<N>` function(similar to the `vec_transform` functor), Here is quick prototype of that(you will need to implement the missing `vec_type` type trait):

```cpp
// Return a vector type of N from index i in another larger vector
// N will be 2 for half2 packing
template <index_int N, class T, class I>
constexpr auto vec_packed_at(T x, I i)
{
    if constexpr(vec_size<T>() == 0)
        return vec<T, N>{x};
    else
    {
        MIGRAPHX_ASSERT((i+N) < vec_size<T>());
        // TODO: vec_type type trait needs to be implemented
        vec<vec_type<T>, N> result;
        for(int j = 0; j < N; j++) {
            result[j] = x[i+j];
        }
        return result;
    }
}

template <index_int N, class... Ts>
constexpr auto vec_packed_transform(Ts... xs)
{
    return [=](auto f) {
        if constexpr(is_any_vec<Ts...>())
        {
            using type                  = decltype(f(vec_packed_at(xs, 0)...));
            constexpr auto size         = common_vec_size<Ts...>();
            safe_vec<type, size> result = {0};
            for(int i = 0; i < size / N; i++)
            {
                // Call the function with packed vectors
                safe_vec<type, N> r = f(vec_packed_at(xs, i)...);
                // Copy the packed vectors to the result
                for(int j = 0; j < N; j++)
                    result[i * N + j] = r[j];
            }
            return result;
        }
        else
        {
            return f(xs...);
        }
    };
}
```

This is part of the runtime compilation so using the gpu-driver can be useful for testing this.

PR #1157, opened 2022-04-01
Half2 overloads

bpickrel:
[Issue 1127](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1127)  Updates the math.hpp header file to perform overloads of various standard functions (ops) for the hip half2 type.  The half2 type is two 16-bit floats packed into a 32-bit number and therefore the overloads act on vectors of sizes that are multiples of 2.  They are invoked in runtime compilation any time one of the ops is called on a tensor declared with the data type `shape::half_type`.

Defined new template, made instances of the template for those math operations that the hip library contains, added verify tests for the sqrt operator for three cases:

1. tensor size not divisible by 2
2. tensor size divisible by 2 but not by 4
3. tensor size divisible by 4


test/math_sample.json
```diff
@@ -0,0 +1,10 @@
+{
+    settings: {
+        lens: [2048],
+        type: "half"
+    },
+    compile_pointwise: {
+        lambda: "[](vec<half, 4> x) { return migraphx::sqrt(x); }",
+        inputs: [{}, {}]
+    }
+}
```
pfultz2, line 10: This file can be removed.

bpickrel: done

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
```diff
@@ -46,6 +46,29 @@ constexpr T as_float(T x)
...
+        MIGRAPHX_RETURNS(migraphx::vec<migraphx::half, 2>{fname(x, xs...)});              \
+    template <class... Ts, index_int N, MIGRAPHX_REQUIRES(N % 2 == 0 && (N > 2))>         \
+    auto __device__ name(migraphx::vec<migraphx::half, N> x, Ts... xs)                    \
+    {                                                                                     \
+        return vec_packed_transform<2>(x, xs...)(                                         \
+            [](auto... ys) -> migraphx::vec<migraphx::half, 2> { return fname(ys...); }); \
+    }
+
+// NOLINTNEXTLINE
+#define MIGRAPHX_DEVICE_MATH_HALF_VEC(name)                                  \
```
kahmed10, line 65: Where is this macro being used?

bpickrel: @pfultz2 , can you answer this?  I put the definition in on your instructions, but in all the debugging I lost track of whether it's still needed or if half4 and larger vectors can all be handled by the ...HALF2 macro.  It isn't currently being referenced.

pfultz2: Yea this isnt used and can be removed.

bpickrel: I just removed it.

kahmed10 (approved):

pfultz2 (approved):

PR #1161, opened 2022-04-05
Expose get_queue method for context in API

umangyadav:


src/api/include/migraphx/migraphx.hpp
```diff
@@ -777,6 +777,13 @@ struct context
     migraphx_context_t ctx;
 
     void finish() const { call(&migraphx_context_finish, ctx); }
+
+    void* get_queue()
```
pfultz2, line 781: This should do the cast so we can type check this in the future:

```cpp
    template<class T>
    T get_queue() const
    {
        void* out;
        call(&migraphx_context_get_queue, &out, ctx);
        // TODO: Check type
        return reinterpret_cast<T>(out);
    }
```

test/api/CMakeLists.txt
```diff
@@ -1,10 +1,12 @@
-
 function(add_api_test TEST_NAME TEST_SRC TEST_DIR)
     set(NAME test_api_${TEST_NAME})
     add_executable(${NAME} EXCLUDE_FROM_ALL ${TEST_SRC})
     rocm_clang_tidy_check(${NAME})
     target_link_libraries(${NAME} migraphx_c migraphx)
     target_include_directories(${NAME} PUBLIC ../include)
+    if(MIGRAPHX_ENABLE_GPU)
+    target_link_libraries(${NAME} migraphx_gpu)
```
pfultz2, line 8: This should be done outside of this function.

src/api/include/migraphx/migraphx.hpp
umangyadav: made changes.

test/api/CMakeLists.txt
umangyadav: Done.

src/api/include/migraphx/migraphx.hpp
pfultz2: It needs to return `T` not `void*`.

kahmed10 (approved):

pfultz2 (approved):

PR #1162, opened 2022-04-06
Dump IR passes into seperate files and directories

umangyadav:
Currently, we dump all the IR passes into one giant file. The giant dump file would contain IR pass output for all the modules as well as the entire program. If quantization is required then, it would also go into the same dump file.  It is difficult to figure compilation process by looking at that dump file. 

After these changes, if `MIGRAPHX_TRACE_COMPILE` is enabled then, IR dumps would go into directory named `target_passes` where target could either be gpu, ref or cpu. 

All the modules passes will have their seperate directories.  All IR dump files are numbered. 

For example, for `if_literal_test` with three modules : main, if_0_else, if_0_if:

```
umayadav@rocm-rome-6:~/repo/AMDMIGraphX/build/ref_passes(umang_ir_dump)$ tree
.
├── 0_input_program.mxr
├── 10_lowering.mxr
├── 11_dead_code_elimination.mxr
├── 1_normalize_ops.mxr
├── 2_eliminate_pad.mxr
├── 3_dead_code_elimination.mxr
├── 4_insert_pad.mxr
├── 5_dead_code_elimination.mxr
├── 6_rewrite_rnn.mxr
├── 7_dead_code_elimination.mxr
├── 8_auto_contiguous.mxr
├── 9_dead_code_elimination.mxr
├── If_0_else
│   ├── 10_lowering.mxr
│   ├── 11_dead_code_elimination.mxr
│   ├── 1_normalize_ops.mxr
│   ├── 2_eliminate_pad.mxr
│   ├── 3_dead_code_elimination.mxr
│   ├── 4_insert_pad.mxr
│   ├── 5_dead_code_elimination.mxr
│   ├── 6_rewrite_rnn.mxr
│   ├── 7_dead_code_elimination.mxr
│   ├── 8_auto_contiguous.mxr
│   └── 9_dead_code_elimination.mxr
├── If_0_if
│   ├── 10_lowering.mxr
│   ├── 11_dead_code_elimination.mxr
│   ├── 1_normalize_ops.mxr
│   ├── 2_eliminate_pad.mxr
│   ├── 3_dead_code_elimination.mxr
│   ├── 4_insert_pad.mxr
│   ├── 5_dead_code_elimination.mxr
│   ├── 6_rewrite_rnn.mxr
│   ├── 7_dead_code_elimination.mxr
│   ├── 8_auto_contiguous.mxr
│   └── 9_dead_code_elimination.mxr
└── main
    ├── 10_lowering.mxr
    ├── 11_dead_code_elimination.mxr
    ├── 1_normalize_ops.mxr
    ├── 2_eliminate_pad.mxr
    ├── 3_dead_code_elimination.mxr
    ├── 4_insert_pad.mxr
    ├── 5_dead_code_elimination.mxr
    ├── 6_rewrite_rnn.mxr
    ├── 7_dead_code_elimination.mxr
    ├── 8_auto_contiguous.mxr
    └── 9_dead_code_elimination.mxr

3 directories, 45 files
```

If quantization is required then, compilation IR dump would go into seperate directory named "target_quantization" where target is again either `cpu`, `gpu` or `ref`. 

`0_input_program.mxr` is the raw input program.


mvermeulen:
One current use case we have with current MIGRAPHX_TRACE_COMPILE is that we can also turn it on as a way of monitoring a process.  For example, turn on the flag and one can see if propogate_constants is a pass that seems to be "stuck/hung".

Once this PR goes in, will this monitoring use case still work or is this focused on more of a "batch" mode to organize the output into files?

umangyadav:
> One current use case we have with current MIGRAPHX_TRACE_COMPILE is that we can also turn it on as a way of monitoring a process. For example, turn on the flag and one can see if propogate_constants is a pass that seems to be "stuck/hung".
> 
> Once this PR goes in, will this monitoring use case still work or is this focused on more of a "batch" mode to organize the output into files?

That's a good case. 

**Without this PR changes, as on the `develop` branch,**

`MIGRAPHX_TRACE_COMPILE`  does two things 1. Prints out IR pass's name before it starts executing that pass. 2. Prints out IR pass output as migraphx program after it finishes the execution. 

With that, you can directly figure out stuck/hung pass. 

**With this PR change :**

It dumps out IR pass output into a file named after IR_pass's name `after` it has finished executing.  So, it is in a way combining two separate things into one compared to the previous. 

To check for the hung pass,  users would have to manually inspect up to which pass, the IR dump files have been created. 

If user is running a script to run multiple test programs,  then, it is not possible to identify hung pass. Dump folder and files are not deleted for each fresh run by default. They are overwritten.  But user can delete the dump folder via script so that they are not overwritten.

src/include/migraphx/tracer.hpp
```diff
@@ -1,33 +1,43 @@
...
+#include <migraphx/filesystem.hpp>
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 struct tracer
 {
     tracer() {}
 
-    tracer(std::ostream& s) : os(&s) {}
+    tracer(std::string dump_directory) : dump_dir(std::move(dump_directory)), counter(0) {}
```
pfultz2, line 15: The tracer should be able to write to an `ostream`, otherwise we cant write to `cout` or `cerr`.

src/pass_manager.cpp
```diff
@@ -10,35 +10,34 @@
...
 
-void validate_pass(module& mod, const pass& p, tracer trace)
+void validate_pass(module& mod, const pass& p)
 {
     (void)mod;
     (void)p;
-    (void)trace;
 #ifndef NDEBUG
-    trace("Validate ...");
+    std::cout << "Validate..." << std::endl;
```
pfultz2, line 26: This should print to the tracer otherwise it will always print out when using debug builds.

src/pass_manager.cpp
```diff
@@ -72,20 +71,19 @@ struct module_pm : module_pass_manager
...
 };
 
 module& get_module(module_pass_manager& mpm) { return mpm.get_module(); }
 
 void run_passes(module& mod, const std::vector<pass>& passes, tracer trace)
 {
-    if(enabled(MIGRAPHX_TRACE_PASSES{}))
-        trace = tracer{std::cout};
+    if(enabled(MIGRAPHX_TRACE_PASSES{}) and not trace.enabled())
+        trace = tracer{mod.name() + "_passes"};
```
pfultz2, line 86: So this looks like `MIGRAPHX_TRACE_PASSES` will dump to a file instead of stdout.

pfultz2:
It would be better to have this under a different env variable name. Something like `MIGRAPHX_DUMP_PASSES_TO_FILE=<path-to-folder-to-save-dump>`.

pfultz2:
Also, the errors in the CI come from a default constructed tracer which should work as a pass-through.

src/pass_manager.cpp
umangyadav: Yes, by design.

kahmed10:
Is there a way we can also get timing information on each pass? And if a pass is hanging, if the user hits ctrl c, the signal is caught and the time at that interval is captured.

umangyadav:
> Is there a way we can also get timing information on each pass? And if a pass is hanging, if the user hits ctrl c, the signal is caught and the time at that interval is captured.

Great Idea. I can support that.

umangyadav:
Updated PR to take environment flag `MIGRAPHX_DUMP_PASSES_TO_FILE` to control whether to dump into files or not. 

`MIGRAPHX_TRACE_COMPILE` works as before. It dump on `std::cout`. 

Added timing information for each pass. For now, it counts wall time for successfully finishing passes.

Precedence : `MIGRAPHX_DUMP_PASSES_TO_FILE` > `MIGRAPHX_TRACE_COMPILE` > `MIGRAPHX_TRACE_PASSES`

umangyadav:
This can be reopened if necessary.

PR #1173, opened 2022-04-11
Expose `add_literal` in C and Python API

umangyadav:
This PR simply exposes the method.  It doesn't require exposing full literal class into the API.

### DIFF
--- a/doc/src/reference/py.rst
+++ b/doc/src/reference/py.rst
@@ -148,9 +148,9 @@ module
 
 .. py:method:: add_literal(data)
 
-    Adds constant or literal data of provided shape into the module from python array or numpy array.    
+    Adds constant or literal data of provided shape into the module from python buffer which includes numpy array.    
 
-    :param py::buffer data: Python array or numpy array 
+    :param py::buffer data: Python buffer or numpy array 
     :rtype instruction 
 
 .. py:method:: add_parameter(name, shape)
--- a/doc/src/reference/py.rst
+++ b/doc/src/reference/py.rst
@@ -146,6 +146,14 @@ module
     :param list[module] mod_args: optional list of module arguments to the operator.
     :rtype instruction
 
+.. py:method:: add_literal(s, data)
+
+    Adds constant or literal data of provided shape into the module from python array or numpy array.    
+
+    :param shape s: shape of the literal
+    :param py::buffer data: Python array or numpy array 
+    :rtype instruction 
+
 .. py:method:: add_parameter(name, shape)
     
     Adds a parameter to the module with provided name and shape.
--- a/doc/src/reference/py.rst
+++ b/doc/src/reference/py.rst
@@ -146,11 +146,10 @@ module
     :param list[module] mod_args: optional list of module arguments to the operator.
     :rtype instruction
 
-.. py:method:: add_literal(s, data)
+.. py:method:: add_literal(data)
 
     Adds constant or literal data of provided shape into the module from python array or numpy array.    
 
-    :param shape s: shape of the literal
     :param py::buffer data: Python array or numpy array 
     :rtype instruction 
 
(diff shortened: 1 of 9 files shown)


src/py/migraphx_py.cpp
```diff
@@ -272,6 +272,14 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
             py::arg("op"),
             py::arg("args"),
             py::arg("mod_args") = std::vector<migraphx::module*>{})
+        .def(
+            "add_literal",
+            [](migraphx::module& mm, const migraphx::shape& s, py::buffer data) {
```
pfultz2, line 277: There is no need to pass the shape as we can get it from the python buffer. You can use the `to_shape` function to convert the buffer to a shape.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -757,6 +757,13 @@ struct module
         return instruction(op_ins, own{});
     }
 
+    instruction add_literal(const migraphx::shape& s, const char* buffer)
```
pfultz2, line 760: Maybe it would be better to do:

```cpp
template<class T>
instruction add_literal(const migraphx::shape& s, const T* buffer)
```

And then do a `static_cast<const char*>` in the function.

test/py/test_module_construct.py
```diff
@@ -1,22 +1,34 @@
-import migraphx
+import migraphx, array, sys
+try:
+    import numpy as np
+except:
+    sys.exit()
```
pfultz2, line 5: I dont think its good to skip all the tests if numpy is missing. Ideally we should move all our numpy tests to a seperate file. You could start by adding this same test in a test_numpy.py file.

src/api/include/migraphx/migraphx.hpp
umangyadav: Good idea. but it requires reinterpret_cast instead of static_cast.

test/py/test_module_construct.py
umangyadav: Moved that numpy test to seperate file.

src/py/migraphx_py.cpp
umangyadav: Made that change. but only concern is that API signature for `add_literal` for both python and c++ would differ.

test/api/test_module_construct.cpp
```diff
@@ -8,18 +8,16 @@ TEST_CASE(add_op)
     migraphx::program p;
     migraphx::module m = p.get_main_module();
     migraphx::shape param_shape{migraphx_shape_float_type, {3, 3}};
-    auto x      = m.add_parameter("x", param_shape);
-    auto y      = m.add_parameter("y", param_shape);
+    std::vector<float> x_values(9, 1);
```
kahmed10, line 11: I would prefer we rename this test to `add_literals` so that it's clear why the test was modified.

umangyadav: Agreed. Changed named.

kahmed10 (approved):

doc/src/reference/py.rst
```diff
@@ -146,6 +146,13 @@ module
     :param list[module] mod_args: optional list of module arguments to the operator.
     :rtype instruction
 
+.. py:method:: add_literal(data)
+
+    Adds constant or literal data of provided shape into the module from python array or numpy array.
```
pfultz2, line 151: This should be: `Adds constant or literal data of provided shape into the module from python buffer which includes numpy arrays`.

doc/src/reference/py.rst
```diff
@@ -146,6 +146,13 @@ module
     :param list[module] mod_args: optional list of module arguments to the operator.
     :rtype instruction
 
+.. py:method:: add_literal(data)
+
+    Adds constant or literal data of provided shape into the module from python array or numpy array.    
+
+    :param py::buffer data: Python array or numpy array
```
pfultz2, line 153: This should say `Python buffer`.

umangyadav: Done

doc/src/reference/py.rst
umangyadav: Done

pfultz2 (approved):

PR #1174, opened 2022-04-12
Fix problem with incomplete types with older clang versions

pfultz2:
I also added the `PYTHON_DISABLE_VERSIONS` cmake variable to disable python versions.


src/py/migraphx_py.cpp
```diff
@@ -7,6 +7,7 @@
 #include <migraphx/operation.hpp>
 #include <migraphx/quantization.hpp>
 #include <migraphx/generate.hpp>
+#include <migraphx/instruction.hpp>
```
umangyadav, line 10: Was this necessary ?

umangyadav (approved):

kahmed10 (approved):

PR #1177, opened 2022-04-14
ReverseSequence op

CharlieL7:
Implements the ReverseSequence ONNX operator as a parser.

* This parser can only handle a constant sequence_lens input. This is the same as what is handled for TensorRT as far as I can tell.
* We could handle a variable sequence_lens input; that would require ref and GPU implementations of the operator.
* The ONNX backend tests are disabled because this does not handle variable sequence_lens.


src/onnx/parse_reversesequence.cpp
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        instruction_ref ret;
+
+        for(int b = 0; b < batch_size; ++b)
+        {
+            instruction_ref s0;
+            if(sequence_lens[b] > 1)
+            {
+                s0 = info.add_instruction(make_op("slice",
```
kahmed10, line 89: seems like most of the fields for the operator are the same with slight variations. Maybe you can make a helper function `add_slice()` where the starts and ends are modified for each call?

kahmed10, line 89: seems like most of the fields for the operator are the same with slight variations. Maybe you can make a helper function `add_slice()` where the starts and ends are modified for each call?

test/onnx/onnx_test.cpp
```diff
@@ -4173,6 +4173,133 @@ TEST_CASE(resize_upsample_pf_test)
...
+    int time_size  = sx.lens()[time_axis];
+
+    auto ret = mm->add_instruction(
+        migraphx::make_op(
+            "slice",
+            {{"axes", {batch_axis, time_axis}}, {"starts", {0, 0}}, {"ends", {1, time_size}}}),
+        input);
+    for(int b = 1; b < batch_size; ++b)
+    {
+        auto s0 = mm->add_instruction(migraphx::make_op("slice",
```
kahmed10, line 4200: same feedback as I mentioned in the parser.

kahmed10, line 4200: same feedback as I mentioned in the parser.

kahmed10 (approved):

turneram (approved):

src/onnx/parse_reversesequence.cpp
```diff
@@ -0,0 +1,125 @@
...
+            time_axis = info.attributes.at("time_axis").i();
+        }
+        if(time_axis != 0 and time_axis != 1)
+        {
+            MIGRAPHX_THROW("REVERSESEQUENCE: time axis not 0 or 1");
+        }
+
+        if(time_axis == batch_axis)
+        {
+            MIGRAPHX_THROW("REVERSESEQUENCE: time axis and batch axis are the same");
```
turneram, line 50: Any clue why these are still being reported as not covered?

CharlieL7: Figured it out, there was a typo in the gen_onnx.py.

PR #1178, opened 2022-04-14
Add method to insert multiple instructions

pfultz2:
This is an extension to `insert_module_instructions`, but instead of just inserting from a module, it can insert a range or a vector of instructions.


test/module_test.cpp
```diff
@@ -277,6 +277,73 @@ TEST_CASE(parameter_name_order)
...
+    auto x1 = m1.add_parameter("x1", s);
+    m1.add_instruction(migraphx::make_op("sqrt"), {x1});
+
+    migraphx::module m2("m2");
+    auto x2    = m2.add_parameter("x2", s);
+    auto sqrt2 = m2.add_instruction(migraphx::make_op("sqrt"), {x2});
+
+    m1.add_instructions(sqrt2, m2.end(), {{x2, x1}});
+    EXPECT(std::any_of(
+        m1.begin(), m2.end(), [&](auto&& ins) { return migraphx::contains(ins.inputs(), x1); }));
```
umangyadav, line 314: What is `m2.end()` here ? Aren't m1 and m2 separate modules ?  shouldn't we check to see if there are not `x2` in the `m1` ?

umangyadav:
Do we need these in the APIs ?

test/module_test.cpp
pfultz2: Yea thats a bug.

pfultz2:
> Do we need these in the APIs ?

No, this is to help with partitioning and optimizing the graph.

test/module_test.cpp
kahmed10: @pfultz2 is this resolved?

test/module_test.cpp
```diff
@@ -277,6 +277,73 @@ TEST_CASE(parameter_name_order)
     EXPECT(param_names == names1);
 }
 
+TEST_CASE(insert_instructions_module)
```
kahmed10, line 280: Even though we are implicitly calling insert_instructions() when using add_instructions(), the test name may be better as `add_instructions_module`. In other places in our source code, we directly call insert_instruction(). Maybe another example where we don't do the insert at the end of the module?

pfultz2: Done

test/module_test.cpp
umangyadav: @pfultz2  Does cppcheck complain about this case now ?

kahmed10 (approved):

umangyadav (approved):

PR #1180, opened 2022-04-17
Add lane reduction

pfultz2:
With reductions such as {2048, 2, 1456} on axes 1, this is 23x faster than using our new block_reduce, and its even over 100x faster than our original `reduce_sum`:

```
# lane
gpu::code_object[code_object=13736,symbol_name=kernel,global=2981888,local=1024,]: 0.0672928ms
# block
gpu::code_object[code_object=13800,symbol_name=kernel,global=39321600,local=64,]: 1.46072ms
# original
gpu::reduce_sum[axes={1}]: 6.73456ms
```

There is some basic logic to pick between lane and block reduce automatically.


src/targets/gpu/kernels/include/migraphx/kernels/tensor_view.hpp
```diff
@@ -29,11 +29,24 @@ struct tensor_view
...
+    };
+
+    constexpr T& operator[](MIGRAPHX_CAPTURE_SOURCE_LOCATION(index_to_offset) i) const
+    {
+        index_to_offset ito = i;
+        MIGRAPHX_WARN(ito.offset < get_shape().element_space(),
+                      i,
+                      "Out of bounds access at offset: ",
+                      ito.offset);
+        // MIGRAPHX_ASSERT(ito.offset < get_shape().element_space());
```
kahmed10, line 48: do we need this commented out?

pfultz2: Yea I removed it.

causten:
I tested this on an MI200. It passes `perf` and `verify`.  No comment on the code changes

src/targets/gpu/jit/reduce.cpp
```diff
@@ -57,6 +57,40 @@ static std::size_t get_reduce_elements(const std::vector<instruction_ref>& input
...
+    // The minimum stride
+    auto min_stride = std::inner_product(
+        rlens.begin(),
+        rlens.end(),
+        inputs.front().strides().begin(),
+        init,
+        [](auto x, auto y) { return std::min(x, y); },
+        [](auto len, auto stride) { return len == 1 ? init : stride; });
+    if(min_stride > 2)
+        return "lane";
```
kahmed10, line 90: is there a reason we don't want to make this into an enum?

kahmed10, line 90: is there a reason we don't want to make this into an enum?

turneram (approved):

src/targets/gpu/jit/reduce.cpp
pfultz2: It needs to return the name of the class, and right now we can't convert enums to strings directly(although I would like to add a macro to do this in the future).

kahmed10 (approved):

PR #1183, opened 2022-04-20
Create allocate op and replace_allocate pass

### LINKED ISSUE #1090: Commit e64b773f2 (PR#1068 Support nonstandard shapes for the Squeeze Op) caused an error the BERT squad large model

kahmed10:
Original design is intended to fix #1090 to ensure output parameters are not incorrectly removed from the program. Still running into issues with moving out loop operator hip::allocates out of lowering.

The flow is as follows:
- insert allocate op in lowering
- after eliminate_contiguous, a new pass called replace_allocate will determine if allocates should be replaced with hip::allocate (or cpu::allocate) or an output parameter should be created.


src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+            p.replace_instruction(ins, out_param);
+            continue;
+        }
+        else if(not main_offload_copy and ins_alias == last and tag.empty())
+        {
+            out_param = p.add_parameter("output", s);
+            p.replace_instruction(ins, out_param);
+            continue;
+        }
+        p.replace_instruction(
```
umangyadav, line 109: Perhaps put this in `else` clause and remove `continue` from previous two `if` blocks ?

src/dead_code_elimination.cpp
```diff
@@ -44,9 +44,10 @@ void dead_code_elimination::apply(module& m) const
         // Skip the last instruction
         if(i == last)
             break;
-        // Skip instruction with empty shape as output unless its a builtin or undefined or identity
+        // Skip instruction with empty shape as output unless its a builtin, undefined, identity, or
+        // allocate
         if(i->get_shape().elements() == 0 and i->name().front() != '@' and
-           i->name() != "undefined" and i->name() != "identity")
+           i->name() != "undefined" and i->name() != "identity" and i->name() != "allocate")
```
pfultz2, line 50: Why do we skip `allocate`?

src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+        else
+        {
+            output = mod.insert_instruction(ins, model.allocate(s));
+        }
+        inputs.push_back(output);
+    }
+    mod.replace_instruction(ins, ins->get_operator(), inputs, mod_args);
+}
+
+void replace_allocate::apply(module& p) const
```
pfultz2, line 67: It should be `module& m`.

test/replace_allocate.cpp
```diff
@@ -0,0 +1,155 @@
...
+{
+    migraphx::shape s{};
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return migraphx::pack(f(self.s, "shape"));
+    }
+
+    std::string name() const { return "cpu::allocate"; }
```
pfultz2, line 24: Dont use the same names for operators to avoid any potential conflict with names.

src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+        }
+        if(ins->get_operator().name() != "allocate")
+            continue;
+        auto op = ins->get_operator();
+        auto v  = op.to_value();
+        assert(v.contains("tag"));
+        auto tag = v.at("tag").get_string();
+        auto s   = ins->get_shape();
+
+        if(model_name == "cpu::allocate")
```
pfultz2, line 88: We shouldn't hardcode names here, ideally, such things should be done with the allocation_model.

Why is this behavior different?

src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+        if(not main_offload_copy and last->name() == "@return" and tag.empty() and
+           prog_output_names.count(ins_alias) > 0)
+        {
+            out_param = p.add_parameter(prog_output_names[ins_alias], s);
+            p.replace_instruction(ins, out_param);
+            continue;
+        }
+        else if(not main_offload_copy and ins_alias == last and tag.empty())
+        {
+            out_param = p.add_parameter("output", s);
```
pfultz2, line 105: We should only insert output parameters for the main module. Otherwise, the allocations for outputs should be moved to the parent module.

src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+    auto prog_output_names = create_output_names(p);
+    auto last              = instruction::get_output_alias(std::prev(p.end()));
+    bool main_offload_copy = p.name() == "main" ? this->offload_copy : false;
+    std::string model_name = model.name();
+    for(auto ins : iterator_for(p))
+    {
+        if(ins->get_operator().name() == "if")
+        {
+            insert_if_allocations(ins, p, model);
```
pfultz2, line 77: I dont think this should be needed.

src/dead_code_elimination.cpp
kahmed10: we aren't skipping, the first condition was passing, so it was then skipping. I had to add an extra condition because of that.

kahmed10: Essentially there are allocates where the number of elements is 0 (from what I saw this was coming from MIOpen convolution), so we need to still remove them as dead code.

src/replace_allocate.cpp
kahmed10: We have many passes where we name module `p`, but I'm okay with changing here. If that's fine, we can open a separate issue to replace the rest of them.

src/replace_allocate.cpp
kahmed10: In cpu lowering, we only ever insert `cpu::allocate`, so no output parameters are being added. Otherwise, the CPU unit tests fail. If there's a better way, I'd like to know.

test/replace_allocate.cpp
kahmed10: Will it make a difference here since the cpu/hip allocates are not registered as ops because we didn't link those libraries in?

src/replace_allocate.cpp
kahmed10: fixed

pfultz2: > We have many passes where we name module p

Those should be updated as well(not in this PR of course). The `p` comes from when it used to be a `program`, but when it was changed to `module` not all the variables were updated.

src/replace_allocate.cpp
pfultz2: We can add to the allocation model a method `insert_output` and then it can insert an allocation for cpu and insert an output parameter for the gpu.

src/dead_code_elimination.cpp
pfultz2: I see, we should probably use `contains({"undefined", "identity", "allocate"}, i->name())`.

test/replace_allocate.cpp
pfultz2: We could link them in the future. It also makes sure we dont rely on the name of the operator in replace_allocate. Either way, I dont think we need different allocation operators for the tests. At most we could have different allocation_models(although I am not sure if thats necessary either).

src/dead_code_elimination.cpp
kahmed10: changed

src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <migraphx/ranges.hpp>
+#include <migraphx/op/allocate.hpp>
+#include <map>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+std::unordered_map<instruction_ref, std::string> create_output_names(const module& mod)
+{
+    std::unordered_map<instruction_ref, std::string> prog_output_names{};
```
umangyadav, line 15: instead of prog_output_names should we name it `mod_output_names` ? since it is only module output names ?

src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+        auto ps = smod->get_parameter_shapes();
+        name_shapes.insert(ps.begin(), ps.end());
+    }
+
+    bool ins_output_allocated = false;
+    for(auto& pn : name_shapes)
+    {
+        const auto& s = pn.second;
+        instruction_ref output{};
+        if(s == ins->get_shape() and not ins_output_allocated)
```
umangyadav, line 53: Do we need this condition ? Regardless of `ins_output_allocated` we always `allocate` both in `then` and `else` condition. 

We are not using `ins_output_allocated` flag later either.

test/replace_allocate.cpp
```diff
@@ -0,0 +1,155 @@
...
+    m.add_return({std::prev(m.end())});
+    run_pass(m, gpu_allocation_model{});
+
+    EXPECT(std::none_of(m.begin(), m.end(), [](const migraphx::instruction& ins) {
+        return migraphx::contains(ins.name(), "allocate");
+    }));
+}
+
+TEST_CASE(hip_allocate)
+{
```
umangyadav, line 136: Can we add a test for the `if_allocation` as well?

src/replace_allocate.cpp
```diff
@@ -0,0 +1,116 @@
...
+            continue;
+        auto op = ins->get_operator();
+        auto v  = op.to_value();
+        assert(v.contains("tag"));
+        auto tag = v.at("tag").get_string();
+        auto s   = ins->get_shape();
+
+        if(model_name == "cpu::allocate")
+        {
+            m.replace_instruction(ins, m.insert_instruction(ins, model.allocate(s)));
```
umangyadav, line 90: I don't understand this, is it adding two allocates ? first by inserting and then by replacing ?

test/replace_allocate.cpp
kahmed10: added

src/replace_allocate.cpp
kahmed10: I believe there is a check to make sure there aren't too many nested conditional statements. I can make the change if it doesn't trigger.

src/replace_allocate.cpp
kahmed10: The unit tests were not passing if I didn't address the if operator.

src/replace_allocate.cpp
kahmed10: changed

src/replace_allocate.cpp
kahmed10: Yeah you're right, I transferred a lot of this code from lowering.cpp. Removed this condition.

src/replace_allocate.cpp
kahmed10: Since it's nested, first the cpu::allocate is inserted, then a replace_instruction() is called to remove the original allocate instruction.

kahmed10: I might be moving this check entirely based on Paul's feedback.

test/replace_allocate.cpp
```diff
@@ -0,0 +1,192 @@
...
+};
+
+void run_pass(migraphx::module& m, migraphx::allocation_model model, bool offload_copy = false)
+{
+    migraphx::run_passes(m,
+                         {migraphx::replace_allocate{std::move(model), offload_copy},
+                          migraphx::dead_code_elimination{}});
+}
+
+void run_pass(migraphx::program& p, migraphx::allocation_model model, bool offload_copy = false)
```
umangyadav, line 92: make this a templated function?

src/replace_allocate.cpp
umangyadav: > Since it's nested, first the cpu::allocate is inserted, then a replace_instruction() is called to remove the original allocate instruction.

After insert_instruction, wouldn't it be `ins -- > allocate_ins` and return argument from insert_instruction would be ins_ref to `allocate_ins`.  Replacement happens on `ins` with `allocate_ins` . So shouldn't that become `allocate_ins -- > allocate_ins` ?

umangyadav (approved):

aserio:
@kahmed10, what is the status of this PR? If I understand correctly it is required to close SWDEV-330397.

kahmed10:
> @kahmed10, what is the status of this PR? If I understand correctly it is required to close SWDEV-330397.

Still working on making changes based on feedback. The issue is that these aren't trivial changes, so I am making sure the unit tests don't break. However, I am making progress and hoping to update this PR soon.

src/replace_allocate.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <migraphx/op/allocate.hpp>
+#include <map>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+std::unordered_map<instruction_ref, std::string> create_output_names(const module& mod)
+{
+    std::unordered_map<instruction_ref, std::string> mod_output_names{};
+    auto last = instruction::get_output_alias(std::prev(mod.end()));
```
pfultz2, line 16: This shouldn't call `get_output_alias` here, it should just be `auto last = std::prev(mod.end());`.

src/replace_allocate.cpp
```diff
@@ -0,0 +1,107 @@
...
+                       prog_outputs.end(),
+                       outputs_alias.begin(),
+                       [](const auto& i) { return instruction::get_output_alias(i); });
+
+        std::size_t index = 0;
+        for(auto ins : outputs_alias)
+        {
+            mod_output_names[ins] = mod.name() + ":#output_" + std::to_string(index++);
+        }
+    }
```
pfultz2, line 32: There should be an else for when not using `@return`:

```cpp
else
{
    auto ins = instruction::get_output_alias(last);
    mod_output_names[ins] = "output";
}
```

src/replace_allocate.cpp
```diff
@@ -0,0 +1,107 @@
...
+        {
+            insert_submod_allocations(ins, m, model);
+            continue;
+        }
+        if(op_name != "allocate")
+            continue;
+
+        auto s = ins->get_shape();
+
+        if(not main_offload_copy and model.needs_out_params())
```
pfultz2, line 81: This should just do `if(not main_offload_copy and model.needs_out_params() and contains(mod_output_names, in))` and skip the inner `if` statement.

src/replace_allocate.cpp
kahmed10: fixed

src/replace_allocate.cpp
kahmed10: fixed

src/include/migraphx/allocation_model.hpp
```diff
@@ -136,6 +140,12 @@ struct allocation_model
         return (*this).private_detail_te_get_handle().preallocate(s, std::move(id));
     }
 
+    bool needs_out_params() const
```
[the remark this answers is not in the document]
pfultz2: Yes this should updated in `tools/allocation_model.hpp` and then run `make generate` to update `src/include/migraphx/allocation_model.hpp`.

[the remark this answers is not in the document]
kahmed10: fixed

pfultz2 (approved):

PR #1184, opened 2022-04-21
Refactor vectorization and preloading for pointwise fusions

pfultz2:
This improves performance. For add_gelu in bert its 4x faster and for mul_add its 50% faster than what is on develop:

```
// This PR
gpu::code_object[code_object=13736,symbol_name=kernel,global=14155776,local=1024,]: 0.243313ms
// Develop
gpu::code_object[code_object=13736,symbol_name=kernel,global=153600,local=1024,]: 0.380066ms
gpu::mul_add: 0.475723ms


// This PR
gpu::code_object[code_object=14176,symbol_name=kernel,global=9437184,local=1024,]: 0.201901ms
// Develop
gpu::code_object[code_object=14016,symbol_name=kernel,global=153600,local=1024,]: 0.89556ms
gpu::add_gelu: 3.04435ms
```

This matches the performance on branch_for_ort2 for these same operators.


causten:
This change does improve performance.  Closer to what is seen in branch_for_ort2.

src/targets/gpu/compile_hip.cpp
```diff
@@ -246,6 +247,16 @@ compile_hip_src(const std::vector<src_file>& srcs, std::string params, const std
             MIGRAPHX_THROW("Missing hsaco");
         };
 
+    if(enabled(MIGRAPHX_GPU_DUMP_SRC{}))
+    {
+        for(const auto& src : srcs)
+        {
+            if(src.path.extension() != ".cpp")
+                continue;
+            std::cout << std::string(src.content.first, src.len()) << std::endl;
```
turneram, line 256: Is there an advantage to doing it this way instead of 
```
if(src.path.extension() == ".cpp")
    std::cout << std::string(src.content.first, src.len()) << std::endl;
```

pfultz2: It just reduces the nested scopes.

kahmed10 (approved):

PR #1185, opened 2022-04-22
Expose `value` class for serialization 

umangyadav:
This PR exposes constructors and methods for basic data types of `int64`, `uint64`, `double`, `float`, `string`, and `bool`.

Constructor with key-value pair is also exposed. 

PR exposes `get_key`, `is_null` method for value object. 

PR adds tests for the same. Tests are copied from `value_test.cpp` with required modifications. 

We probably wouldn't require all the methods for custom_op serialization, just exposing constructors for cpp_types should do. In that case, later i can remove unused methods.  

Follow up work :
Support binary type, object type, array type with bracket operator `[]`, emplace, insert, push etc. methods.  
Support different types of constructors on `value`.


src/api/migraphx.py
```diff
@@ -220,6 +220,27 @@ def module(h):
...
+@auto_handle()
+def value(h):
+    h.constructor('create')
+    h.method('is_null', returns='bool', const=True)
+    h.method('get_key', returns='const std::string', const=True)
+    cpp_types = ['int64_t', 'uint64_t', 'double', 'std::string', 'bool']
+    vt = ['int64', 'uint64', 'float', 'string', 'bool']
+    for vt, cpp_type in zip(vt, cpp_types):
+        if (vt == 'string'):
+            h.constructor('create_' + vt, api.params(i='const char*'))
```
kahmed10, line 232: I would prefer using str.format() or fstrings here rather than concatenating `vt` with +.

kahmed10, line 232: I would prefer using str.format() or fstrings here rather than concatenating `vt` with +.

pfultz2: The api.py should automatically convert the `std::string` parameter to `const char*`, so I dont think the `if` here is necessary.

src/api/migraphx.py
```diff
@@ -220,6 +220,27 @@ def module(h):
...
+            h.constructor('create_' + vt + '_with_key',
+                          api.params(pkey='const char*', i='const char*'))
+        else:
+            h.constructor('create_' + vt, api.params(i=cpp_type))
+            h.constructor('create_' + vt + '_with_key',
+                          api.params(pkey='const char*', i=cpp_type))
+        h.method('if_' + vt, returns='const ' + cpp_type + '*', const=True)
```
pfultz2, line 239: I dont think we need to expose the `if` functions through the C API. We can technically implement them through C++ by calling the `is` and `get` methods.

src/api/migraphx.py
```diff
@@ -220,6 +220,27 @@ def module(h):
...
+        h.method('is_' + vt, returns='bool', const=True)
+        h.method('get_' + vt, returns=cpp_type, const=True)
```
pfultz2, line 241: Actually, I dont think we should add the get methods. Instead lets just add `to_string`, `to_int64`, `to_uint64`, and `to_double` method by calling `v.to<T>()`. Even for now, you could just add `to_string`, which should be enough for unit tests.

pfultz2, line 241: Actually, I dont think we should add the get methods. Instead lets just add `to_string`, `to_int64`, `to_uint64`, and `to_double` method by calling `v.to<T>()`. Even for now, you could just add `to_string`, which should be enough for unit tests.

src/api/migraphx.py
```diff
@@ -220,6 +220,27 @@ def module(h):
...
+        if (vt == 'string'):
+            h.constructor('create_' + vt, api.params(i='const char*'))
+            h.constructor('create_' + vt + '_with_key',
+                          api.params(pkey='const char*', i='const char*'))
+        else:
+            h.constructor('create_' + vt, api.params(i=cpp_type))
+            h.constructor('create_' + vt + '_with_key',
+                          api.params(pkey='const char*', i=cpp_type))
+        h.method('if_' + vt, returns='const ' + cpp_type + '*', const=True)
+        h.method('is_' + vt, returns='bool', const=True)
```
pfultz2, line 240: I think it would be better to use an enum instead of `is` methods. We can leave this off for now.

As an enum, we dont need to know if its a int or string, rather if its a single value, array or object, so it might be good to define the enum differently. We can discuss this for a later PR.

pfultz2, line 240: I think it would be better to use an enum instead of `is` methods. We can leave this off for now.

As an enum, we dont need to know if its a int or string, rather if its a single value, array or object, so it might be good to define the enum differently. We can discuss this for a later PR.

umangyadav:
Review comments are being duplicated. I've deleted duplicate ones.

src/api/migraphx.py
umangyadav: `if` methods return pointer to original data in the struct. It is not possible to return pointer if i compose `get` and `is` method. 
1. I would be returning pointer to temp variable. 2. I can return by value, but in that case, we don't have default value (e.g. nullptr in case of return by pointer).

src/api/migraphx.py
umangyadav: `to<T>()` does the implicit conversion, e.g. `int` can be casted to `float`.  It is not same as `get` methods which has type guarantee.  I am not sure if we want that.

src/api/migraphx.py
umangyadav: Addressed both comments. Using format strings and letting auto convert for string.

src/api/migraphx.py
umangyadav: Btw, I don't see `if` methods being used anywhere in the migraphx source.

src/api/migraphx.py
pfultz2: > to<T>() does the implicit conversion, e.g. int can be casted to float.

But during deserialization we would do this conversion anyways.

>  It is not same as get methods which has type guarantee. I am not sure if we want that in place of get

By having the exact type, then we need to write visitors for deserializations.

It would be simpler to just have a `.to<T>` method without visitors and such. By exposing, conversions to float, string, signed and unsigned ints, then we can efficiently implement `to<T>` in the C++ API(dont worry about implementing it for this PR though).

pfultz2: > then we need to write visitors

This might be a better choice since we dont have to deal with temporary strings.

test/api/test_value.cpp
```diff
@@ -0,0 +1,67 @@
...
+    migraphx::value v = 1.0;
+    EXPECT(v.is_float());
+    // TODO: add float_equal method
+    EXPECT(v.get_key().empty());
+}
+
+TEST_CASE(value_construct_bool)
+{
+    migraphx::value v = true;
+    EXPECT(v.is_bool());
```
kahmed10, line 62: do we have a test for if_bool()?

umangyadav:
> are you planning to still merge this?

I can close this for now, but eventually we need this.

umangyadav:
This can be revisited when necessary.

PR #1190, opened 2022-04-30
Prefuse layernorm for gpu

pfultz2:


src/targets/gpu/prefuse_ops.cpp
```diff
@@ -0,0 +1,76 @@
...
+        if(not x_ins->get_shape().standard())
+            x_ins = m.insert_instruction(ins, make_op("contiguous"), x_ins);
+
+        auto relements = x_ins->get_shape().lens().back();
+
+        if(relements > 1024 or (relements % 4 != 0 and relements > 256))
+            return;
+
+        auto a = m.insert_instruction(
+            ins, make_op("hip::allocate", {{"shape", to_value(x_ins->get_shape())}}));
```
kahmed10, line 28: I'm wondering if after the allocate PR this should just change to `"allocate"` since this pass is done before auto_contiguous.

pfultz2: Yes

src/include/migraphx/matcher.hpp
```diff
@@ -239,7 +252,39 @@ struct any_matcher : any_matcher_base
 
 struct matcher_result
 {
-    std::unordered_map<std::string, instruction_ref> instructions;
+    struct instruction_container
```
kahmed10, line 255: This change affects the CPU as well, so that's why the PR is failing on jenkins. https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/cpu/lowering.cpp#L355

kahmed10:
Looks like in debug mode, an assertion is failing for layernorm. @pfultz2 can you check this?

kahmed10 (approved):

turneram (approved):

Issue #1101, opened 2022-02-23
Add experimental custom op


kahmed10:
Assuming #1093 is merged, we need to create a prototype custom op that does not rely on the context or allocate.

umangyadav:
Closing this one. Solved with #1194

PR #1194, opened 2022-05-03
Add compute_method for the experimental custom op

umangyadav:
Adds `compute_method` for the experimental custom ops.

Adds a test for the same using HIP APIs.  

Depends on #1183

Solves #1101


test/api/CMakeLists.txt
```diff
@@ -11,7 +12,6 @@ endfunction()
 
 add_api_test(array_base test_array_base.cpp ${TEST_ONNX_DIR})
 add_api_test(assign test_assign.cpp ${TEST_ONNX_DIR})
-add_api_test(custom_op test_custom_op.cpp ${TEST_ONNX_DIR})
```
kahmed10, line 14: should we still make a test for custom op that doesn't rely on the GPU?

kahmed10, line 14: should we still make a test for custom op that doesn't rely on the GPU?

test/api/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+
```
kahmed10, line 1: extra newline?

kahmed10, line 1: extra newline?

test/api/test_custom_op.cpp
```diff
@@ -1,23 +1,61 @@
+
```
kahmed10, line 1: extra newline?

kahmed10, line 1: extra newline?

umangyadav:
Review comments are being duplicated. Removed duplicate ones.

test/api/CMakeLists.txt
umangyadav: Removed extra lines from both `test_custom_op.cpp` and `CMakeLists.txt`.

test/api/CMakeLists.txt
umangyadav: Added one `ref` target custom op test.

umangyadav:
Rebased with #1183  to add "allocate" arg to custom operators.

umangyadav:
Updated PR to have `add_allocation` method.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -445,9 +445,6 @@ struct shape : MIGRAPHX_CONST_HANDLE_BASE(shape)
 {
     shape() {}
 
-    MIGRAPHX_DEPRECATED("Contructor without lifetime annotation is deprecated.")
-    shape(const migraphx_shape* p) { this->set_handle(p, borrow{}); }
```
pfultz2, line 449: Why is this constructor removed? It is deprecated but I would like to wait until rocm 6.0 to remove it.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -786,28 +791,32 @@ struct module
...
 
     private:
     std::shared_ptr<migraphx_module> mm;
 };
 
-struct context
+struct context : MIGRAPHX_HANDLE_BASE(context)
 {
-    context(migraphx_context* p, borrow) : ctx(std::shared_ptr<migraphx_context*>(), p) {}
+    context() { this->make_handle(&migraphx_context_create); }
```
pfultz2, line 809: Whats the reason for adding a default constructor?

src/api/include/migraphx/migraphx.hpp
umangyadav: While setting function pointer for the compute_method, `auto_convert_param` matches with this constructor and `-Werror` converts warning into error failing the build. 

../../../src/api/include/migraphx/migraphx.hpp:389:10: error: ‘migraphx::api::shape::shape(const migraphx_shape*)’ is deprecated: Contructor without lifetime annotation is deprecated. [-Werror=deprecated-declarations]
     auto auto_convert_param(rank<1>, T x) -> decltype(as_handle<T>{x})
          ^~~~~~~~~~~~~~~~~~
../../../src/api/include/migraphx/migraphx.hpp:451:5: note: declared here
     shape(const migraphx_shape* p) { this->set_handle(p, borrow{}); }

src/api/include/migraphx/migraphx.hpp
umangyadav: Removed, it is not needed.

src/api/include/migraphx/migraphx.hpp
umangyadav: Added back that constructor, the thing that was causing it was c++ shape is a `const` handle base type and `C` API was passing non-const type. For auto_convert_param to match and invoke a constructor with lifetime at `rank<2>`, I had to pass `const_migraphx_shape` from C API.

test/api/test_custom_op.cpp
```diff
@@ -1,23 +1,62 @@
...
+    auto custom_kernel = m.add_instruction(migraphx::operation("sigmoid_custom_op"), {x, alloc});
+    p.compile(migraphx::target("ref"));
+    // run program
+    migraphx::program_parameters pp;
+    auto param_shapes            = p.get_parameter_shapes();
+    migraphx::argument input_arg = migraphx::argument::generate(param_shapes["x"]);
+    pp.add("x", input_arg);
+    auto results   = p.eval(pp);
+    auto result    = results[0];
+    auto input_vec = input_arg.as_vector<float>();
```
kahmed10, line 55: I would prefer to name this as `golden` or `output_golden` since technically sigmoid was already computed on it.

umangyadav: Done.

src/api/include/migraphx/migraphx.h
```diff
@@ -459,6 +469,13 @@ migraphx_status migraphx_quantize_int8(migraphx_program_t prog,
                                        migraphx_target_t target,
                                        migraphx_quantize_int8_options_t options);
 
+migraphx_status migraphx_context_destroy(migraphx_context_t context);
+
+migraphx_status migraphx_context_assign_to(migraphx_context_t output,
+                                           const_migraphx_context_t input);
+
+migraphx_status migraphx_context_create(migraphx_context_t* context);
```
pfultz2, line 477: Ideally, we shouldn't have a create/destroy for context as users shouldn't be creating and destroying the context. Whats the reason for adding this? Is this to support `auto_convert_param`? Maybe it would be better to inherit from `handle_lookup` or add an overload for `migraphx_adl_handle_lookup`.

causten:
Code Cov is indicating converge is very low.

src/api/include/migraphx/migraphx.h
umangyadav: Yes, it is for `auto_convert_param`.

umangyadav: Made changes for `context` to directly inherit from `handle_lookup` and removed constructors, deleters for context.

umangyadav:
> Code Cov is indicating converge is very low.

I am not too much worried about that. Looking at codecov diff, it shows lines that are not covered are mainly about sanity check 'asserts' that correct inputs are passed to functions. Also, for some reason, codecov shows coverage change for a bunch of files that are really not affected by these changes.

pfultz2:
> I am not too much worried about that.

Yea me neither. A lot of the code cov is coming from the auto generated API code.

test/api/test_op_construct.cpp
```diff
@@ -49,4 +49,11 @@ TEST_CASE(reduce_mean_with_format)
     EXPECT(rm.name() == "reduce_mean");
 }
 
+TEST_CASE(allocate_with_nested_json_str)
+{
+    auto alloc = migraphx::operation(
+        "allocate", R"({"shape":{"type":"float_type","lens":[3, 3], "strides":[3, 1]}})");
```
pfultz2, line 55: I dont think this is a good test. We shouldn't rely on how the shape class is serialized, as it will change in the future. I think we can remove this.

umangyadav: Removed that test.

pfultz2 (approved):

PR #1198, opened 2022-05-06
Upgrade CI environment to 5.3.0

causten:
Support rocm5.3 in the CI environment.
Drop support for 18.04, python 3.6
22.04 will be added with a different PR

Disabled the new `misc-confusable-identifiers`.  This is the case where I (upper case i) and l (lower case L), and even 1 (one) are confusable.  It also checks for odd ascii chars that look similar to each other.  Pretty useful but our code is littered with 1 and l and will take a developer to make those changes.

```
int fo; // Initial character is U+0066 (LATIN SMALL LETTER F).
int 𝐟o; // Initial character is U+1234 (SUPER COOL AWESOME UPPERCASE NOT LATIN F) not U+0066 (LATIN SMALL LETTER F)
```

Changed to clang-format instead of clang-format-10

Resolved Clang-Format attempts to simplify boolean expressions by DeMorgan's theorem.  Hence the code changes associated with removing


Dockerfile
```diff
@@ -6,13 +6,13 @@ ARG PREFIX=/usr/local
...
 # Add rocm repository
-RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.0.2/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
+RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.3/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
 
 # Install dependencies
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated \
     apt-utils \
     build-essential \
-    clang-format-10 \
+    clang-format \
```
pfultz2, line 15: We should use the package that is versioned.

.github/workflows/ci.yaml
```diff
@@ -136,13 +136,13 @@ jobs:
               -o -iname '*.cl' \
               -o -iname '*.c' \
           | grep -v 'build/' \
-          | xargs -n 1 -P 1 -I{} -t sh -c 'clang-format-10 -style=file {} | diff - {}'
+          | xargs -n 1 -P 1 -I{} -t sh -c 'clang-format -style=file {} | diff - {}'
```
pfultz2, line 139: We should use the executable with the version in the name. Also, if we are changing versions then the githook needs to be updated.

Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG PREFIX=/usr/local
 RUN dpkg --add-architecture i386
 
 # Add rocm repository
-RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.0.2/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
+RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.3/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
```
umangyadav, line 9: Dockefile installs `clang-format-10` is that needed?

umangyadav: Or is it keeping at `clang-format-10` instead of just `clang-format` ?

umangyadav (approved):

Dockerfile
causten: I've had this conversation with Paul already.  He wants to be explicit with the version.

umangyadav (approved):

src/targets/gpu/hip.cpp
```diff
@@ -183,8 +183,8 @@ argument register_on_gpu(const argument& arg)
 {
     auto arg_shared = arg.share();
     auto p          = register_on_gpu(arg_shared.data(), arg_shared.get_shape().bytes());
-    return {arg_shared.get_shape(),
-            [p, a = std::move(arg_shared)]() mutable { return get_device_ptr(p.get()); }};
+    auto s          = arg_shared.get_shape();
```
pfultz2, line 186: Why is this assigned to another variable?

test/gpu/literal.cpp
```diff
@@ -48,4 +48,4 @@ void gpu_literal_test()
     }
 }
 
-int main() { gpu_literal_test(); }
+int main() { gpu_literal_test(); } // NOLINT
```
pfultz2, line 51: What is the lint error here?

src/targets/gpu/hip.cpp
causten: Fixing issue caught by clang-tidy with possible ordering resolution of operands
due to std::move. Just captured the shape before arg_shared is used in the lambda

More details here.

https://releases.llvm.org/11.1.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-use-after-move.html

test/gpu/literal.cpp
causten: [ 60%] clang-tidy: Running clang-tidy on target /workspace/AMDMIGraphX/test/onnx/onnx_rnn_test.cpp...
62310 warnings generated.
::warning file=test/gpu/literal.cpp,line=51,col=5::an exception may be thrown in function 'main' which should not throw exceptions [bugprone-exception-escape,-warnings-as-errors]
int main() { gpu_literal_test(); }
    ^
note: expanded from here

CMake Error at tidy-target-test_gpu_literal-_workspace_AMDMIGraphX_test_gpu_literal_cpp.cmake:80 (message):
  Clang tidy failed.

pfultz2: Can you do `NOLINT(bugprone-exception-escape)` to make it clearer which warnings you are trying to suppress?

pfultz2:
This should probably be merged in after branching for 5.4. Otherwise, other PRs we need to merge in for 5.4 may need more fixes due to new warnings.

pfultz2 (approved):

PR #1199, opened 2022-05-06
Dyn shape update

CharlieL7:
Initial sketch for changes to shape to handle dynamic dimensions


CharlieL7:
The ambiguous constructor error seems to fail or not fail depending on the compiler used. Could change previous code to explicitly use std::vector<size_t> rather than braced-list-initialization or change the dynamic shape constructor.

src/include/migraphx/shape.hpp
```diff
@@ -59,6 +59,43 @@ struct shape
     {
     };
 
+    struct dynamic_dimension
+    {
+        std::size_t min = 0;
+        std::size_t max = 0;
+        std::size_t opt = 0;
```
pfultz2, line 66: Add a `reflect` method(like we do for operators) so we can easily serialize it.

src/shape.cpp
```diff
@@ -346,12 +468,24 @@ const std::vector<shape>& shape::sub_shapes() const { return impl->m_shapes; }
...
+        result["max_dyn_dims"] = migraphx::to_value(s.max_dyn_dims());
+        result["opt_dyn_dims"] = migraphx::to_value(s.opt_dyn_dims());
+        result["sub_shapes"]   = migraphx::to_value(s.sub_shapes());
+    }
+    else
+    {
+        result["lens"]       = migraphx::to_value(s.lens());
+        result["strides"]    = migraphx::to_value(s.strides());
+        result["sub_shapes"] = migraphx::to_value(s.sub_shapes());
+    }
```
pfultz2, line 485: I dont think you need a branch just serialize it with:

```cpp
        result["lens"]       = migraphx::to_value(s.lens());
        result["strides"]    = migraphx::to_value(s.strides());
        result["sub_shapes"] = migraphx::to_value(s.sub_shapes());
        result["dynamic_dimensions"]       = migraphx::to_value(s.dyn_dims());
```

This should work after adding the `reflect` method.

src/shape.cpp
```diff
@@ -309,20 +378,73 @@ std::size_t shape::element_space() const { return impl->element_space(); }
...
+const std::vector<shape::dynamic_dimension>& shape::dyn_dims() const { return impl->m_dyn_dims; }
+
+std::vector<std::size_t> shape::min_dyn_dims() const
+{
+    auto num_dims = dyn_dims().size();
+    std::vector<std::size_t> ret(num_dims);
+    for(int i = 0; i < num_dims; ++i)
+    {
+        ret.at(i) = dyn_dims().at(i).min;
+    }
```
pfultz2, line 392: Use `std::transform` for these functions.

src/include/migraphx/shape.hpp
```diff
@@ -59,6 +59,43 @@ struct shape
...
+        }
+        friend bool operator!=(const dynamic_dimension& x, const dynamic_dimension& y)
+        {
+            return !(x == y);
+        }
+        friend std::ostream& operator<<(std::ostream& os, const dynamic_dimension& x)
+        {
+            os << "[" << x.min << ", " << x.max << ", " << x.opt << "]";
+            return os;
+        }
```
pfultz2, line 89: We should put all these methods into the .cpp file eventually.

src/permutation.cpp
```diff
@@ -13,11 +13,18 @@ shape reorder_shape(const shape& s, const std::vector<int64_t>& permutation)
...
+ * Inverts the permutation using the less_than operator
+ */
 std::vector<int64_t> invert_permutation(const std::vector<int64_t>& permutation)
 {
     return sort_permutation(permutation, std::less<>{});
 }
 
+/*!
+ * Computes a permutation for the lengths based on decesending stride order.
+ * Compares the lengths if the strides are the same.
```
pfultz2, line 26: > Compares the lengths if the strides are the same

I dont think we need to put the implementation details like this in the documentation. I think it would be better to say "Finds the permutation most likely from a transpose operator that has been applied to the shape".

src/include/migraphx/shape.hpp
```diff
@@ -89,10 +128,30 @@ struct shape
...
+     * Return the size of the type of the main shape.
+     * Returns 0 if there are subshapes.
+     */
     std::size_t type_size() const;
 
+    const std::vector<dynamic_dimension>& dyn_dims() const;
+
+    std::vector<std::size_t> min_dyn_dims() const;
+    std::vector<std::size_t> max_dyn_dims() const;
+    std::vector<std::size_t> opt_dyn_dims() const;
```
pfultz2, line 153: These should probably be `max_lens`, `min_lens`, and `opt_lens` and they should return `lens()` when the shape is not dynamic.

src/include/migraphx/shape.hpp
```diff
@@ -59,6 +59,43 @@ struct shape
...
+    struct dynamic_dimension
+    {
+        std::size_t min = 0;
+        std::size_t max = 0;
+        std::size_t opt = 0;
+
+        dynamic_dimension() = default;
+
+        dynamic_dimension(std::size_t i_min, std::size_t i_max, std::size_t i_opt)
+            : min(i_min), max(i_max), opt(i_opt)
```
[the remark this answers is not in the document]
CharlieL7: I made the constructor because emplace_back() doesn't work with the aggregate constructor until C++20. I'll refactor and avoid the emplace_back() defect.

src/shape.cpp
```diff
@@ -150,11 +172,19 @@ shape shape::from_permutation(type_t t,
...
 const std::vector<std::size_t>& shape::strides() const { return impl->m_strides; }
+
 std::size_t shape::elements() const { return impl->elements(); }
+
 std::size_t shape::bytes() const
 {
+    if(this->dynamic())
+    {
+        MIGRAPHX_THROW("SHAPE: bytes() called on dynamic shape");
+    }
```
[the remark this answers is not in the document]
CharlieL7: I think we would have to assume that the maximum dynamic shape is packed to get the maximum number of bytes. Do we want to make this assumption?

CharlieL7:
Where do we test the ostream operator<< functions for code coverage?

src/shape.cpp
```diff
@@ -309,20 +413,95 @@ std::size_t shape::element_space() const { return impl->element_space(); }
 
 std::string shape::type_string() const { return name(this->type()); }
 
+bool shape::dynamic() const { return not impl->m_dyn_dims.empty(); }
+
+const std::vector<shape::dynamic_dimension>& shape::dyn_dims() const { return impl->m_dyn_dims; }
+
+std::vector<std::size_t> shape::min_lens() const
+{
+    if(not this->dynamic())
```
kahmed10, line 422: Would it be better to rewrite like this?
```
return this->dynamic() ? impl->min_lens() : this->lens();
```

src/include/migraphx/shape.hpp
```diff
@@ -248,6 +313,10 @@ struct shape
 
     const std::vector<shape>& sub_shapes() const;
 
+    /*!
+     * Returns size of the data buffer.
+     * Assuming a packed shape, returns maximum size of the data buffer for dynamic shape.
```
pfultz2, line 318: It only assumes packed shape for dynamic shapes, which is not completely clear in this comment.

pfultz2 (approved):
The code change is good, it would be good to update the comment to make it clearer what the `element_space` function does.

src/include/migraphx/shape.hpp
kahmed10: @CharlieL7 any update to what @pfultz2 suggested? Also, I would change the wording to something like "number of elements" in the data buffer, since size may imply something like bytes which this is not doing.

kahmed10 (approved):

PR #1202, opened 2022-05-11
Add vectorized reduce

pfultz2:
Reductions can be up to 30% faster when using half:

```
gpu::code_object[code_object=13816,symbol_name=reduce_kernel,global=25165824,local=256,]: 0.96653ms
gpu::code_object[code_object=13800,symbol_name=kernel,global=25165824,local=256,]: 1.30959ms
```


src/targets/gpu/compile_gen.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <migraphx/stringutils.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace gen {
+
+static std::vector<std::size_t> vector_sizes(const std::vector<shape>& inputs)
+{
+    // If all inputs is half then only use half2
```
kahmed10, line 13:
```suggestion
    // If all inputs are half then only use half2
```

turneram (approved):

kahmed10 (approved):

Issue #1193, opened 2022-05-03
Rename module variable in MIGraphX passes


kahmed10:
In our passes, we often write the following variable for the module as `module& p`. However, we should rename this to `module& m` where appropriate. These are remnants from when we converted the passes to be at the module level instead of the whole program.

For example, in [simplify_reshapes.cpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/simplify_algebra.cpp#L1014):
`module& p` should be changed to `module& m`.

PR #1204, opened 2022-05-12
renamed variables for module from p to m

shivadbhavsar:
Updated variable names according to #1193

### DIFF
--- a/src/targets/gpu/schedule_model.cpp
+++ b/src/targets/gpu/schedule_model.cpp
@@ -77,28 +77,28 @@ MIGRAPHX_REGISTER_OP(wait_event)
 MIGRAPHX_REGISTER_OP(set_stream)
 
 std::size_t schedule_model::concurrency() const { return streams; }
-void schedule_model::sched(module& p, instruction_ref ins, std::size_t n) const
+void schedule_model::sched(module& m, instruction_ref ins, std::size_t n) const
 {
     auto last_stream = std::find_if(std::make_reverse_iterator(ins),
-                                    std::make_reverse_iterator(p.begin()),
+                                    std::make_reverse_iterator(m.begin()),
                                     [&](auto&& i) { return i.name() == "gpu::set_stream"; });
-    if(last_stream != std::make_reverse_iterator(p.begin()))
+    if(last_stream != std::make_reverse_iterator(m.begin()))
     {
         auto&& op = any_cast<set_stream>(last_stream->get_operator());
         // If the same stream was set earlier then skip
         if(op.stream == n)
             return;
     }
-    p.insert_instruction(ins, set_stream{n});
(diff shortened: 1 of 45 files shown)


src/targets/gpu/schedule_model.cpp
```diff
@@ -77,28 +77,28 @@ MIGRAPHX_REGISTER_OP(wait_event)
 MIGRAPHX_REGISTER_OP(set_stream)
 
 std::size_t schedule_model::concurrency() const { return streams; }
-void schedule_model::sched(module& p, instruction_ref ins, std::size_t n) const
```
kahmed10, line 80: The clang-format error check looks to be unrelated to the changes you made to this file, but I think you should update it anyways. We must have missed the change when we upgraded versions awhile back.

kahmed10 (approved):

umangyadav:
I find couple of more places where change from`module& p` to `module& m` is required. 

See 
`src/analyze_stream.cpp`,  `src/include/analyze_stream.hpp`,  


In `src/rewrite_rnn.cpp` and `src/rewrite_pooling.cpp` and their header files,  It uses `module& prog` instead of `module& m`  we can change that too for uniformity.

umangyadav:
I find couple of more places where change from`module& p` to `module& m` is required. 

See 
`src/analyze_stream.cpp`,  `src/include/analyze_stream.hpp`,  


In `src/rewrite_rnn.cpp` and `src/rewrite_pooling.cpp` and their header files,  It uses `module& prog` instead of `module& m`  we can change that too for uniformity.

shivadbhavsar:
> I find couple of more places where change from`module& p` to `module& m` is required.
> 
> See `src/analyze_stream.cpp`, `src/include/analyze_stream.hpp`,
> 
> In `src/rewrite_rnn.cpp` and `src/rewrite_pooling.cpp` and their header files, It uses `module& prog` instead of `module& m` we can change that too for uniformity.

analyze_streams was left unchanged as the name "m" was already used for the stream_model parameter. Is there a commonly used variable name for stream_model objects?

umangyadav:
> > I find couple of more places where change from`module& p` to `module& m` is required.
> > See `src/analyze_stream.cpp`, `src/include/analyze_stream.hpp`,
> > In `src/rewrite_rnn.cpp` and `src/rewrite_pooling.cpp` and their header files, It uses `module& prog` instead of `module& m` we can change that too for uniformity.
> 
> analyze_streams was left unchanged as the name "m" was already used for the stream_model parameter. Is there a commonly used variable name for stream_model objects?

I don't think we use `stream_model` variable frequently.  I see it being used in those `analyze_stream` files only. It would be better to rename module var to `m` and `stream_model` to something else i.e. `strm` or `strmm`.

umangyadav (approved):

PR #1208, opened 2022-05-17
Custom Op example using MIOpen calls

umangyadav:
Depends on https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1183 and https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1194.

This PR only adds an example using MIOpen Calls. 

Files to look at are in examples/ folder. i.e. custom_op_miopen_kernel.cpp, CMakeLists.txt and README.md .


examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+int main(int argc, const char* argv[])
+{
+    abs_custom_op abs_op;
+    migraphx::register_experimental_custom_op(abs_op);
+    migraphx::program p;
+    migraphx::shape s{migraphx_shape_float_type, {32, 256}};
+    migraphx::module m = p.get_main_module();
+    auto x             = m.add_parameter("x", s);
+    auto neg_ins       = m.add_instruction(migraphx::operation("neg"), {x});
```
kahmed10, line 129: I don't think this instruction is needed if the input numbers range from negative to positive numbers.

examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
```diff
@@ -0,0 +1,160 @@
...
+struct abs_custom_op final : migraphx::experimental_custom_op_base
+{
+    virtual std::string name() const override { return "abs_custom_op"; }
+    virtual migraphx::argument compute(migraphx::context ctx,
+                                       migraphx::shape output_shape,
+                                       migraphx::arguments args) const override
+    {
+        float alpha = 1;
+        float beta  = 0;
+        auto y_desc = make_miopen_tensor(output_shape);
```
kahmed10, line 102: We should also add a comment to explain this step of making miopen tensors.

umangyadav: Added.

kahmed10 (approved):

examples/migraphx/custom_op_miopen_kernel/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+set (EXAMPLE custom_op_miopen_kernel)
+
+list (APPEND CMAKE_PREFIX_PATH /opt/rocm/miopen /opt/rocm/hip /opt/rocm)
+find_package (migraphx REQUIRED)
+find_package (miopen REQUIRED)
+
+message("source file: " ${EXAMPLE}.cpp " ---> bin: " ${EXAMPLE})
+add_executable(${EXAMPLE} ${EXAMPLE}.cpp)
+
+target_link_libraries(${EXAMPLE} migraphx::c MIOpen hip::device)
```
pfultz2, line 37: This shouldn't link `hip::device` since its host-only APIs.

examples/migraphx/custom_op_miopen_kernel/README.md
```diff
@@ -0,0 +1,10 @@
+# Custom MIOpen Kernel using MIGraphX API. 
+ This is an example of a custom operator implementation using MIGraphX's C/C++ APIs. It also demonstrates how to use this custom op in conjunction with rest of MIGraphX operators to build  and run MIGraphX program on GPU. 
+ Kernels can be written in either HIP, MIOpen, or by using RocBLAS library.  This particular example uses **MIOpen** library calls.
+
+ To build and run example, ensure ROCm is installed at `/opt/rocm`. 
+ 1.  `export LD_LIBRARY_PATH=/opt/rocm/lib:$LD_LIBRARY_PATH`
+ 2.  `cd $MIGRAPHX_SRC/examples/migraphx/custom_op_miopen_kernel/`
+ 3.  `mkdir build && cd build`
+ 4.  `CXX=/opt/rocm/llvm/bin/clang++ cmake ..  && make`
```
pfultz2, line 9: rocm clang should not be needed as this is using host-only APIs.

examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
```diff
@@ -0,0 +1,173 @@
...
+    auto alloc         = m.add_allocation(s);
+    auto custom_kernel = m.add_instruction(migraphx::operation("abs_custom_op"), {neg_ins, alloc});
+    auto relu_ins      = m.add_instruction(migraphx::operation("relu"), {custom_kernel});
+    m.add_return({relu_ins});
+
+    migraphx::compile_options options;
+    // set offload copy to true for GPUs
+    options.set_offload_copy();
+    p.compile(migraphx::target("gpu"), options);
+    migraphx::program_parameters pp;
```
TedThemistokleous, line 153: rename this to p_param, or something more descriptive hard to follow with program being p already.

examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include <migraphx/migraphx.hpp> // MIGraphX's C++ API
+#include <numeric>
+#include <stdexcept>
+
+#define MIGRAPHX_MIOPEN_ASSERT(x) (assert((x) == miopenStatusSuccess))
+#define MIGRAPHX_HIP_ASSERT(x) (assert((x) == hipSuccess))
+
+inline miopenTensorDescriptor_t make_miopen_tensor(const migraphx::shape& s, bool pack = false)
+{
+    // TODO: normalize_standard shape for scalar
```
TedThemistokleous, line 37: Get rid  of TODO.

examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
```diff
@@ -0,0 +1,173 @@
...
+        throw("MAKE_TENSOR: unsupported type");
+    }
+    miopenSetTensorDescriptor(t, d, s_lens.size(), lens.data(), strides.data());
+    return t;
+}
+
+inline auto make_miopen_handle(migraphx::context& ctx)
+{
+    MIGRAPHX_HIP_ASSERT(hipSetDevice(0));
+    auto* stream = ctx.get_queue<hipStream_t>();
```
TedThemistokleous, line 77: Can we handle the case if get_queue () fails or returns null?

TedThemistokleous (changes requested):
Overall good. Minor things with naming and a possible edge case if getting the queue fails.

examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
umangyadav: Type check needs to happen inside that `get_queue<>` call. It is part of TODO. For now, we are only focusing on AMD GPU so, we are assuming, get_queue with hipStream will always be valid.  In order to do type check, it requires additional piece of work which on the TODO list.

examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
umangyadav: Changed it to `prog_params`.  `pp` is used in many other tests, that's why i'd used `pp` here as well. `pp` name is mostly used in limited context whenever we need to run the program. So it should be fine otherwise.

pfultz2 (approved):

examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
TedThemistokleous: Understood, just wanted to ask. That makes sense though.

TedThemistokleous (approved):
Looks good.

Issue #1196, opened 2022-05-03
ONNX Mean doesn't work correctly for integral inputs


pfultz2:
Currently when we parse the Mean operator for onnx we divide the inputs first to avoid overflowing however this doesn't work for integral types which do division after summation.

PR #1209, opened 2022-05-17
Fix onnx mean parsing for integral inputs

shivadbhavsar:
As described in #1196, the ONNX mean parser does not work correctly for integral types. This update fixes the issue by handling integral types separately, where summation is performed before division. Additional test cases have also been added for handling integral types.


kahmed10:
You probably want to pin your python formatter to the one we use in Docker:
```
pip3 install yapf==0.28.0
```

src/onnx/parse_mean.cpp
```diff
@@ -24,17 +28,29 @@ struct parse_mean : op_parser<parse_mean>
...
-                               args.end(),
-                               info.add_broadcastable_binary_op("div", args[0], divisor),
-                               [&](auto mean, auto data_i) {
-                                   // Pre-divide each tensor element-wise by n to reduce risk of
-                                   // overflow during summation
-                                   auto div =
-                                       info.add_broadcastable_binary_op("div", data_i, divisor);
-                                   return info.add_broadcastable_binary_op("add", mean, div);
-                               });
+        if(contains(float_types, args[0]->get_shape().type()))
```
kahmed10, line 31: I wonder if we should add a method to the shape class `is_float_type()` instead of creating a set with the float types. Just because it might be useful elsewhere in the future. If others agree, we can make this a TODO and open a separate issue.

kahmed10 (approved):

causten:
Can you add some pull request comments.  https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/

src/onnx/parse_mean.cpp
pfultz2: Yea, I agree. We could use the visitor for this to check the type using type_traits instead of having raw lists. Thats what I did in the mlir-c branch to check for signedness:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/mlir-c/src/include/migraphx/shape.hpp#L171

We should open a seperate issue for this.

pfultz2 (approved):

PR #1212, opened 2022-05-18
Improve matching with has_value when there are convert operators

pfultz2:


kahmed10 (approved):
There may be a case where after a convert, a similar value falls within tolerance but isn't the actual value the matcher is looking for. But I doubt we'll see that behavior happen in practice.

umangyadav (approved):

PR #1213, opened 2022-05-18
Fuse gemm add with pointwise fusions

pfultz2:


src/targets/gpu/fuse_ops.cpp
```diff
@@ -931,6 +931,58 @@ struct find_gemm_add
...
+        auto gemm_ins = r.instructions["gemm"];
+        auto c_ins    = r.instructions["c"];
+
+        auto gemm = any_cast<rocblas_gemm<op::dot>>(gemm_ins->get_operator());
+
+        // Already fused gemm
+        if(not float_equal(gemm.beta, 0))
+            return;
+
+        if(std::any_of(ins->inputs().begin(), ins->inputs().end(), [](auto i) {
```
umangyadav, line 970: Do we need this ? We are checking for standard shapes of inputs in the matcher.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -931,6 +931,58 @@ struct find_gemm_add
...
+        auto n = std::count_if(pm->begin(), pm->end(), [&](auto& i) { return i.name() == s; });
+        if(n != 1)
+            return false;
+        return std::all_of(pm->begin(), pm->end(), [&](auto& i) {
+            return starts_with(i.name(), "@") or i.name() == s;
+        });
+    }));
+}
+
+struct find_gemm_pointwise
```
umangyadav, line 947: Do we have a test for this ?

pfultz2: We have verify tests for this.

src/targets/gpu/fuse_ops.cpp
pfultz2: Removed.

umangyadav: It can be removed from `find_gemm_add` as well.

umangyadav (approved):

kahmed10 (changes requested):
Can we remove L911-914 as well? Otherwise LGTM

kahmed10 (approved):

PR #1214, opened 2022-05-18
Improve applicable batched gemms

pfultz2:
This can help remove the need for contiguous for many cases when using BS 1 for bert.


causten:
I tested this branch and it does indeed boost performance for BS1

src/targets/gpu/gemm_impl.cpp
```diff
@@ -27,6 +28,22 @@ rocblas_datatype get_type(shape::type_t type)
     MIGRAPHX_THROW("ROCBLAS_GEMM: data type not supported!");
 }
 
+void blas_shape(const shape& s)
+{
+    if(s.lens().size() < 2)
```
umangyadav, line 33: It is checking for size() < 2 and size() < 3, returning both times, Should it only check size() < 3 ?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -27,6 +28,22 @@ rocblas_datatype get_type(shape::type_t type)
...
+        return;
+    if(std::none_of(s.strides().end() - 2, s.strides().end(), [&](auto i) { return i == 1; }))
```
umangyadav, line 35: Should be lengths instead of strides to be collapsible ? I think both works.

src/reduce_dims.cpp
```diff
@@ -37,17 +35,33 @@ bool reduce_dim(std::vector<shape>& shapes, std::size_t n)
     return true;
 }
 
+void reduce_dim1(std::vector<shape>& shapes)
```
umangyadav, line 38: This looks similar to `squeeze` op on last axis.

pfultz2: Yea, since the other `reduce_dim` doesn't work on the last dimension.

src/targets/gpu/gemm_impl.cpp
pfultz2: No it should return yet. If the size is at least two we need to check the memory layout still,

src/targets/gpu/gemm_impl.cpp
pfultz2: This isn't checking the batch dimensions. Blas needs one of the strides of the matrix dimension(ie either the row or column) to be `1`, so this checking for that.

umangyadav (approved):

kahmed10 (approved):

PR #1216, opened 2022-05-20
Enable BERT attention fusion in onnxruntime

turneram:


src/onnx/parse_layernorm.cpp
```diff
@@ -0,0 +1,47 @@
...
+        {
+            epsilon = parser.parse_value(info.attributes.at("epsilon")).at<float>();
+        }
+        if(contains(info.attributes, "axis"))
+        {
+            epsilon = parser.parse_value(info.attributes.at("axis")).at<int64_t>();
+        }
+
+        auto layernorm = info.add_instruction(
+            make_op("layernorm", {{"epsilon", epsilon}, {"axis", axis}}), args.front());
```
pfultz2, line 33: We shouldn't be adding a custom layernorm operator. We should just insert layernorm using our existing operators.

turneram: That is something I had done previously, but for some reason the fusion of those ops into gpu::layernorm and gpu::triaddlayernorm wasn't happening inside ORT even though it was happening in the driver. I can try it again and see if I can figure out why.

pfultz2: Yea we need to investigate why this fusion is not happening. #1212 did do some improvements for the matcher, I am not sure if that was related or not.

turneram:
Made unnecessary by #1354

PR #1218, opened 2022-05-24
Prioritizing int8 over int8x4 when it is applicable

jerryyin:
This PR is mostly inspired by @pfultz2's `conv_int8_pack` branch. I'm able to validate MIGraphx invoking into the int8 convolution using:

```bash
MIOPEN_FIND_MODE=1 MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvMlirIgemmFwdXdlops ./bin/test_verify quant_conv
```

### DIFF
--- a/src/targets/gpu/pack_int8_args.cpp
+++ b/src/targets/gpu/pack_int8_args.cpp
@@ -156,6 +156,12 @@ void pack_int8_args::apply(module& m) const
         }
         else if(ins->name() == "gpu::quant_convolution")
         {
+            auto val = ins->get_operator().to_value();
+            if(not val.at("int8_x4_format").to<bool>())
+            {
+                return;
+            }
+
             auto inputs   = ins->inputs();
             auto packed_x = m.insert_instruction(
                 ins,
--- a/src/targets/gpu/pack_int8_args.cpp
+++ b/src/targets/gpu/pack_int8_args.cpp
@@ -118,7 +118,7 @@ void pack_int8_args::apply(module& m) const
             assert(val.contains("int8_x4_format"));
             if(not val.at("int8_x4_format").to<bool>())
             {
-                return;
+                continue;
             }
             auto inputs = ins->inputs();
             auto lens   = inputs.at(0)->get_shape().lens();
@@ -159,7 +159,7 @@ void pack_int8_args::apply(module& m) const
             auto val = ins->get_operator().to_value();
             if(not val.at("int8_x4_format").to<bool>())
             {
-                return;
+                continue;
             }
 
             auto inputs   = ins->inputs();
--- a/src/targets/gpu/pack_int8_args.cpp
+++ b/src/targets/gpu/pack_int8_args.cpp
@@ -22,10 +22,10 @@ static instruction_ref pad_ins(module& m, instruction_ref ins, int offset)
     auto pad_k                     = (k + 3) / 4 * 4;
     auto pad_lens                  = lens;
     pad_lens[lens.size() + offset] = pad_k;
-    std::vector<int64_t> pad_dims(lens.size() * 2, 0);
-    auto ret_ins = ins;
+    auto ret_ins                   = ins;
     if(pad_k != k)
     {
+        std::vector<int64_t> pad_dims(lens.size() * 2, 0);
         pad_dims[lens.size() + offset] = pad_k - k;
         shape ps{s.type(), pad_lens};
         auto ins_out =
--- a/test/verify/quant_conv_default_mode.cpp
+++ b/test/verify/quant_conv_default_mode.cpp
@@ -10,9 +10,9 @@ struct quant_conv_default_mode : verify_program<quant_conv_default_mode>
     {
         migraphx::program p;
         auto* mm = p.get_main_module();
-        migraphx::shape a_shape{migraphx::shape::int8_type, {2, 3, 4, 4}};
+        migraphx::shape a_shape{migraphx::shape::int8_type, {16, 16, 4, 4}};
         auto pa = mm->add_parameter("a", a_shape);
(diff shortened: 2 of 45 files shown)


src/targets/gpu/pack_int8_args.cpp
```diff
@@ -156,6 +156,12 @@ void pack_int8_args::apply(module& m) const
         }
         else if(ins->name() == "gpu::quant_convolution")
         {
+            auto val = ins->get_operator().to_value();
+            if(not val.at("int8_x4_format").to<bool>())
+            {
+                return;
```
pfultz2, line 162: This should do `continue`(actually the other place where there is a `return` is a bug).

test/verify/quant_conv_default_mode.cpp
```diff
@@ -10,9 +10,9 @@ struct quant_conv_default_mode : verify_program<quant_conv_default_mode>
     {
         migraphx::program p;
         auto* mm = p.get_main_module();
-        migraphx::shape a_shape{migraphx::shape::int8_type, {2, 3, 4, 4}};
+        migraphx::shape a_shape{migraphx::shape::int8_type, {16, 16, 4, 4}};
```
pfultz2, line 13: Why are these tests changed?

src/targets/gpu/pack_int8_args.cpp
jerryyin: Ah yes, I think you mentioned about this when we walked through the code and I just forget about it in the implementation. I'll amend to `continue` in both places.

test/verify/quant_conv_default_mode.cpp
jerryyin: @pfultz2 `[2, 3, 4, 4]` dimensions will be converted to `[2, 4, 4, 4]` from `miopen_quant_convolution::pack_int8_shape()`, which happens to be supported by `int8x4`. 

On the other hand, neither `[2, 3, 4, 4]` nor `[2, 4, 4, 4]` are supported dimensions with `int8`, so I manually picked a dimension small enough and still supported.

src/targets/gpu/pack_int8_args.cpp
jerryyin: Addressed in d5eb524

test/verify/quant_conv_default_mode.cpp
pfultz2: We shouldn't change the original test as we need to ensure such sizes will work(or do you mean these tests will break?).

We can add a test with the `16, 16, 4, 4` input size to get test coverage.

jerryyin: Correct, those tests will break with previous dimensions: they will end with "MIOpen Error: /root/MIOpen/src/ocl/convolutionocl.cpp:518: Forward Convolution cannot be executed due to incorrect params".

Also I'm curious on why does MIGraphX care on a specific set of dimension? I'm assuming as long as it is supported by the library under the hood, it proves the MIGraphX correctness. Whatever that doesn't get supported is a MIOpen problem, not MIGraphX problem.

pfultz2: > Correct, those tests will break with previous dimensions 

Then we need to add more logic in the pack_int8_args to check for the dimensions supported by miopen. 

> Also I'm curious on why does MIGraphX care on a specific set of dimension?

We can have a model fail to run just because it didn't use dimensions that was nice for MIOpen.

jerryyin: > We cannot have a model fail to run just because it didn't use dimensions that was nice for MIOpen.

Fair enough.

> Then we need to add more logic in the pack_int8_args to check for the dimensions supported by miopen.

Are you talking about checking whether MIOpen supports int8 with current config, and use int8 only when it is supported? (otherwise fallback to int8x4). From what I can see that's the job of the `quant_convolution.cpp`, that it only knows whether the config can be supported after the `miopenFindConvolutionForwardAlgorithm()` call. But then `pack_int_args` seems to be just a lowering pass and making it to invoke the miopen API seems to be beyond its responsibility (or capability, because it mostly have just the shape information, if I understand it correctly.

pfultz2: > Are you talking about checking whether MIOpen supports int8 with current config, and use int8 only when it is supported? (otherwise fallback to int8x4).

Yes.

> From what I can see that's the job of the quant_convolution.cpp, that it only knows whether the config can be supported after the miopenFindConvolutionForwardAlgorithm() call.

Do you mean when we call `compute` method for quant_convolution? If find fails in compute(or finalize), then we cannot fallback on int8x4 since the input is a different layout.

> But then pack_int_args seems to be just a lowering pass and making it to invoke the miopen API seems to be beyond its responsibility (or capability, because it mostly have just the shape information, if I understand it correctly.

Ah thats right, the logic needs to be in lowering not in pack_int_args(which just checks the `int8_x4_format` flag). In lowering we already do call find so we can get the workspace:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/lowering.cpp#L369

We just need to put a try/catch around this and fallback on int8x4 if it throws an error.

jerryyin: @pfultz2 For sure. Test case wise, which of the following do you prefer:
 - Add an additional set of test with {16,4,4,4} to explicitly regress int8 shape; preserve the {2,3,4,4} to test the fallback
 - Mix both {16,4,4,4} and {2,3,4,4} between different test cases such that we have coverage for both (and it reduced duplication)

jerryyin: Done in 3d972e24d5f40201252765b5c8c6c611beda8349

turneram (approved):

Issue #1166, opened 2022-04-07
Optimized the propagate_constant pass..


scxiao:
Right now, the propagate_constant pass is very slow in some case, one possible reason is the recursive call inside this pass, which may be of a time complexity of O(n2), we need to optimize this pass to make it run faster...

causten:
This will be your 3rd assignment, slightly more challenging then the first two.  
Steps to reproduce

```
export MIGRAPHX_TRACE_PASSES=1
wget https://zenodo.org/record/3733910/files/model.onnx
/opt/rocm/bin/migraphx-driver perf ./model.onnx   --fill1 input_ids --fill1 segment_ids --input-dim @input_ids 1 384 --fp16
```
The 1 and 384 map to Batch Size and Sequence Length respectively.  

There is a lot of text that will print.  Eventually you will see the text below.  It will sit on the screen for a bit then move on...
    Module: main, Pass: propagate_constant

Now change the batch size to 16.  You will notice the `propagate_constant` pass will hang longer on the screen.

Note: You don't need to use the exact model, there are smaller ones that do the same thing.  I can share with you a smaller version of Bert. 

There is also a slightly less hang at "Eliminate Contiguous" Maybe the same efficiency can help with that one too, but probably you could use another issue for that one.

PR #1220, opened 2022-05-24
Parallelize evaluations in propagate_constant

shivadbhavsar:
Addressing issue #1166 - propagate_constant pass currently uses a recursive approach to find all instructions in a module that can be evaluated to a literal and performs the replacement in the same call. 

New approach:

1. Perform single pass though instructions in the module to determine which instructions can be evaluated
2. Evaluate selected instructions in parallel
3. Replace the selected instructions with the corresponding literal

Performance timed using:
`export MIGRAPHX_TRACE_PASSES=1`
`wget https://zenodo.org/record/3733910/files/model.onnx`
`time AMDMIGraphX/build/bin/migraphx-driver compile model.onnx --fill1 input_ids --fill1 input_mask --fill1 segment_ids --batch 16 --fp16`

Original:
real    26m10.002s
user    29m46.135s
sys     1m8.178s

New:
real    11m55.748s
user    24m19.602s
sys     1m28.457s


src/propagate_constant.cpp
```diff
@@ -20,33 +21,42 @@ bool skip_propogate(instruction_ref ins)
...
-            for(auto child : children)
-            {
-                if(child->name() == "@literal" or skip_propogate(child))
-                {
-                    self(child);
-                    continue;
-                }
-                auto r = child->eval();
-                if(not r.empty())
+        if(not is_const(i) or i == std::prev(m.end()))
```
pfultz2, line 33: Assign `auto last = std::prev(m.end())` outside of the loop, and then you can just compare `last`.

Also, use early return instead of nesting:

```cpp
for(auto i : iterator_for(m))
{
    if (is_const(i) and i != last)
        continue;
    ...
}
```

src/propagate_constant.cpp
```diff
@@ -20,33 +21,42 @@ bool skip_propogate(instruction_ref ins)
...
+        {
+            std::for_each(i->inputs().begin(), i->inputs().end(), [&](const instruction_ref ins) {
```
pfultz2, line 35: Use `std::copy_if` here.

pfultz2 (approved):

src/propagate_constant.cpp
```diff
@@ -20,33 +21,41 @@ bool skip_propogate(instruction_ref ins)
...
-                    auto l = m.add_literal(r.get_shape(), r.data());
-                    self(m.replace_instruction(child, l));
-                }
-            }
-        })(i);
+
+        std::copy_if(i->inputs().begin(),
+                     i->inputs().end(),
+                     std::inserter(const_instrs, const_instrs.begin()),
+                     [&](const instruction_ref ins) { return is_const(ins); });
```
pfultz2, line 40: Actually this needs to be `return is_const(ins) and ins->name() != "@literal"` otherwise we will try to compute literals.

pfultz2 (changes requested):

pfultz2 (approved):

kahmed10 (approved):

PR #1221, opened 2022-05-25
Used wrong path to download the bertsquad-10.onnx model

causten:
raw is the download for the file, blob is the url for the github page.


umangyadav (approved):

CharlieL7 (approved):
Examined the onnx file from wget, looks right.

PR #1222, opened 2022-05-25
Dynamic convolution draft

CharlieL7:
**Info**
* Make the reference implementation of Convolution work with dynamic shapes
* Depends on #1199
* Disables some reference compile passes when shape is dynamic
* Runs compute_shape() during eval() when the instruction has a dynamic output shape

**Issues/TODO**
* ONNX parser handle unknown shape to dynamic shape
* Rerun passes when shape given?
* Memoize the compute_shape()?
* Allow setting padding/strides/dilation at evaluation?
* Need to error out when kernel shape is incompatible (ex. kernel is larger than image)
* Other tests?


CharlieL7:
Made as regular PR by mistake.

PR #1223, opened 2022-05-25
Improve eliminate contiguous pass

shivadbhavsar:
Following up on issue #1166 and PR #1220.  Using the same approach as in #1220 for parallelizing the eval calls, we can significantly reduce the time spent on eliminate_contiguous pass.

Applying this on top of changes in #1220 yields:
`wget https://zenodo.org/record/3733910/files/model.onnx`
`time AMDMIGraphX/build/bin/migraphx-driver compile model.onnx --fill1 input_ids --fill1 input_mask --fill1 segment_ids --batch 16 --fp16`

real    9m43.252s
user    25m5.624s
sys     1m41.403s

### DIFF
--- a/src/eliminate_contiguous.cpp
+++ b/src/eliminate_contiguous.cpp
@@ -6,6 +6,7 @@
 #include <migraphx/stringutils.hpp>
 #include <migraphx/op/contiguous.hpp>
 #include <migraphx/op/identity.hpp>
+#include <migraphx/par_for.hpp>
 #include <utility>
 
 namespace migraphx {
@@ -71,6 +72,9 @@ static bool try_compute_shape(instruction_ref ins,
 
 void eliminate_contiguous::apply(module& m) const
 {
+    std::vector<instruction_ref> inputs;
+    std::vector<instruction_ref> prevs;
+
     for(auto ins : iterator_for(m))
     {
         // return instruction should have inputs with standard shape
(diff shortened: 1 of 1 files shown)


src/eliminate_contiguous.cpp
```diff
@@ -93,15 +98,25 @@ void eliminate_contiguous::apply(module& m) const
                 }
                 else if(prev->can_eval())
                 {
-                    auto c = op::contiguous{};
-                    auto r = c.compute(c.compute_shape({prev->get_shape()}), {prev->eval()});
-
-                    auto l = m.add_literal(r.get_shape(), r.data());
-                    m.replace_instruction(arg, l);
+                    inputs.push_back(arg);
+                    prevs.push_back(prev);
```
pfultz2, line 102: We really only need to store the `arg` since `prev` can be found using `arg->inputs().front()`.

src/eliminate_contiguous.cpp
```diff
@@ -71,6 +72,9 @@ static bool try_compute_shape(instruction_ref ins,
 
 void eliminate_contiguous::apply(module& m) const
 {
+    std::vector<instruction_ref> inputs;
```
pfultz2, line 75: I think it would be better to name this `const_instruction`.

kahmed10 (approved):

pfultz2 (approved):

PR #1224, opened 2022-05-25
Dynamic ref convolution op

CharlieL7:
**Info**
* Make the reference implementation of Convolution work with dynamic shapes
  * dynamic input tensor and weights
* Depends on #1199 and #1249 and #1295
* Disables some reference compiler passes when shape is dynamic
* ~~Runs compute_shape() during eval() when the instruction has a dynamic output shape~~
* Runs compute_shape() in the ref_convolution compute() function if the instruction shape is dynamic
* Dynamic auto_pad calculation implemented

**Issues/TODO**
* ~~ONNX parser handle unknown shape to dynamic shape~~ (implemented)
* ~~auto_pad ONNX option will need to be recomputed at evaluation time~~
* ~~Memoize the compute_shape() for multiple different sets of arguments?~~ (not doing for now)
* ~~Allow setting padding/strides/dilation at evaluation?~~ (Attributes will be set at compile-time)
* ~~Need to error out when kernel shape is incompatible, ex. kernel is larger than image~~ (Not handled in original code)


src/include/migraphx/op/convolution.hpp
```diff
@@ -69,43 +69,92 @@ struct convolution
...
 
-        if(input.lens().at(1) != (weights.lens().at(1) * group))
-            MIGRAPHX_THROW("CONVOLUTION: Mismatch channel numbers");
-
-        std::vector<size_t> output_lens{input.lens()[0], weights.lens()[0]};
-
-        for(size_t i = 0; i < kdims; i++)
+        if(not input.dynamic() and not weights.dynamic() and
+           input.lens().at(1) != (weights.lens().at(1) * group))
+            MIGRAPHX_THROW("CONVOLUTION: mismatched channel numbers");
```
turneram, line 92: Can you add tests for these in `test/shape_test.cpp`?

CharlieL7: I'll make some tests. I think the tests should be in `ref_ops_test.cpp` since `shape_test.cpp` looks to be testing the shape object only.

turneram: My bad, I meant `test/op_shape_test.cpp`

CharlieL7: Cool, I was wondering where these types of tests went. I found a bug with the shape calculation after making more shape calculation tests.

src/program.cpp
```diff
@@ -352,7 +355,10 @@ std::vector<argument> generic_eval(const module* mod,
                             }));
         }
         assert(results.find(ins) != results.end());
-        assert(results.at(ins).get_shape() == ins->get_shape());
+        if(not ins->get_shape().dynamic())
```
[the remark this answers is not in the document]
CharlieL7: They will not be the same since the instruction shape will be dynamic but the result shape will be fixed. With the current way the code is structured, any operator's `compute()` function will have a fixed output shape.

src/targets/ref/lowering.cpp
```diff
@@ -231,8 +231,14 @@ struct ref_convolution : auto_register_op<ref_convolution<Op>>
     {
         return op.normalize_compute_shape(inputs);
     }
+
     argument compute(context&, shape output_shape, std::vector<argument> args) const
     {
+        if(output_shape.dynamic())
```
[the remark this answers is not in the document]
CharlieL7: It's calculating the proper fixed shape output using the input arguments.

umangyadav (approved):

src/include/migraphx/op/convolution.hpp
```diff
@@ -69,43 +69,99 @@ struct convolution
...
-            auto padding_factor = 2 * padding[i];
-            if(padding_size == 2 * kdims)
-                padding_factor = padding[i] + padding[i + kdims];
-            output_lens.push_back(std::size_t(std::max<std::ptrdiff_t>(
-                1,
-                (input.lens()[i + 2] - (1 + dilation[i] * (weights.lens()[i + 2] - 1)) +
-                 padding_factor) /
-                        stride[i] +
-                    1)));
+            std::vector<shape::dynamic_dimension> output_dyn_dims = {};
```
kahmed10, line 125: currently if neither the input nor the weights are dynamic, we are populating this vector anyways. I suggest just calling it output_dims.

src/include/migraphx/op/convolution.hpp
```diff
@@ -69,43 +69,99 @@ struct convolution
...
+            for(size_t i = 0; i < num_spatial_dims; ++i)
+            {
+                output_dyn_dims.push_back(shape::dynamic_dimension{
+                    min_spatial_dims[i], max_spatial_dims[i], opt_spatial_dims[i]});
+            }
+            return shape{input.type(), output_dyn_dims};
+        }
+        else
+        {
+            std::vector<size_t> output_lens{input.lens()[0], weights.lens()[0]};
```
kahmed10, line 158: given my comment above, then do we need this?

src/include/migraphx/op/convolution.hpp
CharlieL7: That should not be the case. If both of the input shapes are fixed, the `output_lens` in the else block should be used.

kahmed10: yes, I saw the if statement before. But my second comment on the else block is asking if we need to do that or we can merge both into one calculation.

CharlieL7:
Made some significant changes to support dynamic shape auto_pad. Should be ready for re-review now.

src/include/migraphx/op/convolution.hpp
```diff
@@ -69,43 +71,141 @@ struct convolution
...
+                        1)));
+            }
+        }
+        return ret;
+    }
+
+    shape dynamic_compute_shape(shape x_shape, shape w_shape) const
+    {
+        std::vector<shape::dynamic_dimension> output_dyn_dims = {};
+        if(x_shape.dynamic())
```
kahmed10, line 147: I think you can rewrite this to a function:
```
void check_dynamic_shape(const shape& input_shape, std::vector<shape::dynamic_dimension>& output_dyn_dims)
{
    if(input_shape.dynamic())
    {
        output_dyn_dims.push_back(input_shape.dyn_dims().at(0));
    }
    else
    {
        auto l = input_shape.lens().at(0);
        output_dyn_dims.push_back({l, l, 0});
    }
}
...
check_dynamic_shape(x_shape, output_dyn_dims);
check_dynamic_shape(w_shape, output_dyn_dims);
```

kahmed10 (approved):

test/gpu/mlir.cpp
```diff
@@ -145,7 +145,7 @@ TEST_CASE(conv)
     const std::string mlir_output = R"__migraphx__(
 module {
   func @main(%arg0: tensor<2x8x3x3xf32>, %arg1: tensor<1x8x4x4xf32>) -> tensor<1x2x2x2xf32> attributes {kernel = "mixr"} {
-    %0 = migraphx.convolution(%arg1, %arg0) {dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1]} : (tensor<1x8x4x4xf32>, tensor<2x8x3x3xf32>) -> tensor<1x2x2x2xf32>
+    %0 = migraphx.convolution(%arg1, %arg0) {dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1], use_dynamic_auto_pad = 0} : (tensor<1x8x4x4xf32>, tensor<2x8x3x3xf32>) -> tensor<1x2x2x2xf32>
```
kahmed10, line 148:
```suggestion
    %0 = migraphx.convolution(%arg1, %arg0) {dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1], use_dynamic_auto_pad = 0 : i64} : (tensor<1x8x4x4xf32>, tensor<2x8x3x3xf32>) -> tensor<1x2x2x2xf32>
```

test/gpu/mlir.cpp
```diff
@@ -168,7 +168,7 @@ TEST_CASE(conv_add_relu)
     const std::string mlir_output = R"__migraphx__(
 module {
   func @main(%arg0: tensor<1x2x2x2xf32>, %arg1: tensor<2x8x3x3xf32>, %arg2: tensor<1x8x4x4xf32>) -> tensor<1x2x2x2xf32> attributes {kernel = "mixr"} {
-    %0 = migraphx.convolution(%arg2, %arg1) {dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1]} : (tensor<1x8x4x4xf32>, tensor<2x8x3x3xf32>) -> tensor<1x2x2x2xf32>
+    %0 = migraphx.convolution(%arg2, %arg1) {dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1], use_dynamic_auto_pad = 0} : (tensor<1x8x4x4xf32>, tensor<2x8x3x3xf32>) -> tensor<1x2x2x2xf32>
```
kahmed10, line 171:
```suggestion
    %0 = migraphx.convolution(%arg2, %arg1) {dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1], use_dynamic_auto_pad = 0 : i64} : (tensor<1x8x4x4xf32>, tensor<2x8x3x3xf32>) -> tensor<1x2x2x2xf32>
```

PR #1225, opened 2022-05-25
Upgrade to cppcheck 2.8 and fix new issues found

pfultz2:


src/onnx/parse_pooling.cpp
```diff
@@ -128,14 +128,14 @@ struct parse_pooling : op_parser<parse_pooling>
...
         // used to calculate the supposed output shape
-        std::vector<int64_t> orig_padding(paddings.begin(), paddings.end());
 
         std::vector<int64_t> slice_start;
         std::vector<int64_t> slice_end;
         tune_padding_size(values, paddings, count_include_pad, slice_start);
 
         if(!slice_start.empty())
         {
+            std::vector<int64_t> orig_padding(paddings.begin(), paddings.end());
```
umangyadav, line 138: Looks like change is making pooling tests fail. `padding` values changes during `tune_padding_size`, `orig_padding` should take original values.

umangyadav (changes requested):

umangyadav:
`module.cpp` has suppression for `redundantInitialization` and `cpu/target.cpp` has it for `constParameter`. Do you think we can remove them as well ?

pfultz2:
> Do you think we can remove them as well ?

No we cannot.

umangyadav (approved):

kahmed10 (approved):

PR #1228, opened 2022-05-30
Update protobuf version

kahmed10:
Fixes issue on clang_release build which calls into the onnx runtime test runner.

### DIFF
--- a/examples/nlp/python_bert_squad/requirements_bertsquad.txt
+++ b/examples/nlp/python_bert_squad/requirements_bertsquad.txt
@@ -1,3 +1,3 @@
-tensorflow==2.6.4
+tensorflow==2.7.2
 onnxruntime
 tokenizers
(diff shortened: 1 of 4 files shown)


kahmed10:
> In requirements.txt it specifically points to protobuf 3.11.0. Could you just change it to 3.20.0 there?

Trying that now.

pfultz2:
We should still pin the protobuf version for python.

kahmed10:
> We should still pin the protobuf version for python to match whatever version we use in requirements.txt.

Updated

kahmed10:
Looks like there's an issue with the newer version of protobuf when building with MIGraphX, so for now I am keeping it as is. We can resolve that issue in a different PR.

pfultz2 (approved):

causten (approved):

PR #1232, opened 2022-06-01
Horizontally fuse contiguous operators

pfultz2:
This reorders the transposes across slice to improve horizontal fusion for contiguous. This also improves eliminate_contiguous to remove contiguous better across splits.


src/eliminate_contiguous.cpp
```diff
@@ -70,53 +70,79 @@ static bool try_compute_shape(instruction_ref ins,
...
     {
         auto l = m.add_literal(literals[i].get_shape(), literals[i].data());
-        m.replace_instruction(const_instruction[i], l);
+        m.replace_instruction(const_instructions[i], l);
     }
 }
 
+void eliminate_contiguous::apply(module& m) const
+{
+    // Skip contiguous from splits first
```
umangyadav, line 138: It is not skipping, it is eliminating.

src/eliminate_contiguous.cpp
```diff
@@ -70,53 +70,79 @@ static bool try_compute_shape(instruction_ref ins,
...
+    auto last = std::prev(m.end());
+    std::vector<instruction_ref> const_instructions;
 
     for(auto ins : iterator_for(m))
     {
         // return instruction should have inputs with standard shape
         if(ins->name() == "@return")
             continue;
 
+        if(ins != last and ins->outputs().empty())
```
umangyadav, line 85: Shouldn't this be eliminated by deadcode elimination ?

src/eliminate_contiguous.cpp
```diff
@@ -70,53 +70,79 @@ static bool try_compute_shape(instruction_ref ins,
...
+                    const_instructions.push_back(arg);
                 }
             }
         }
     }
 
     // Perform evaluations in parallel
-    std::vector<argument> literals(const_instruction.size());
-    par_for(const_instruction.size(), 1, [&](const auto i) {
+    std::vector<argument> literals(const_instructions.size());
```
umangyadav, line 122: Why are we doing constant folding here and not in propagate constant pass?

src/eliminate_contiguous.cpp
pfultz2: It will be, but this will skip code that will be eliminated by dead code elimination.

src/eliminate_contiguous.cpp
pfultz2: No, its skipping elimination of contiguous that is being split. Then later it will try to eliminate all contiguous.

src/eliminate_contiguous.cpp
pfultz2: Propagate constant won't fold on a contiguous operator as it could be inefficient if contiguous can later be eliminated. So we const fold the contiguous here if it cant be removed.

src/eliminate_contiguous.cpp
```diff
@@ -70,53 +70,79 @@ static bool try_compute_shape(instruction_ref ins,
...
+            }
+            else if(prev->can_eval())
             {
-                auto prev = arg->inputs().front();
                 replace(new_args, arg, prev);
                 if(try_compute_shape(ins, new_args, mod_args))
                 {
                     instruction::replace_argument(ins, arg, prev);
                 }
                 else if(prev->can_eval())
```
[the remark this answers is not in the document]
pfultz2: We need to check `can_eval` otherwise we can't replace it with a literal.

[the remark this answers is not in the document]
umangyadav: else if (prev->can_eval()) is already checking it.

src/eliminate_contiguous.cpp
umangyadav: it does skip for contiguous that is being split, but it removes all other contiguous since `if not f(ins) then continue`.

src/eliminate_contiguous.cpp
umangyadav: I find it better to rather augment propagate constant to handle this to keep logic separated.

pfultz2: I agree, but I think we should have a seperate issue for that.

src/eliminate_contiguous.cpp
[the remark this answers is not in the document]
pfultz2: I see, it looks like this may have been duplicated during the merge. Let me try to fix it.

umangyadav (approved):

test/simplify_reshapes_test.cpp
```diff
@@ -1118,4 +1118,106 @@ TEST_CASE(transpose_contiguous_reshape_binary_broadcast)
...
+{
+    migraphx::module m1;
+    {
+        auto x      = m1.add_parameter("x", {migraphx::shape::float_type, {1, 384, 36, 64}});
+        auto slice1 = m1.add_instruction(
+            migraphx::make_op("slice", {{"axes", {2}}, {"starts", {0}}, {"ends", {12}}}), x);
+        auto sqrt1  = m1.add_instruction(migraphx::make_op("sqrt"), slice1);
+        auto slice2 = m1.add_instruction(
+            migraphx::make_op("slice", {{"axes", {2}}, {"starts", {12}}, {"ends", {24}}}), x);
+        auto transpose2 = m1.add_instruction(
```
kahmed10, line 1211: Should this be called `transpose` since there is only one transpose in the example?

pfultz2: Renamed.

kahmed10 (approved):

PR #1235, opened 2022-06-02
std::distance bugfix in instruction::valid

CharlieL7:
* Fixes hanging issue when an argument to an instruction is from another module.
  * Assumes the argument is ordered correctly (before the current one) and does not check the instruction distance.
* Allows SSD-Resnet34 to compile
  * Model still stuck on allocation due to extremely high max_iterations for the loop instruction

### DIFF
--- a/src/instruction.cpp
+++ b/src/instruction.cpp
@@ -74,14 +74,29 @@ bool operator==(const instruction& i, instruction_ref ref)
     return std::addressof(i) == std::addressof(*ref);
 }
 
+static void debug_name(std::ostream& os, const instruction& ins);
+
 bool instruction::valid(instruction_ref start, bool check_order) const
 {
+    // Need this lambda because std::distance has undefined behavior when comparing iterators of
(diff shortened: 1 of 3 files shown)


src/instruction.cpp
```diff
@@ -74,14 +74,18 @@ bool operator==(const instruction& i, instruction_ref ref)
     return std::addressof(i) == std::addressof(*ref);
 }
 
-bool instruction::valid(instruction_ref start, bool check_order) const
+static void debug_name(std::ostream& os, const instruction& ins);
```
kahmed10, line 77: Why does this need to be defined here?

kahmed10, line 77: Why does this need to be defined here?

CharlieL7: Leftover from debugging, removed.

CharlieL7:
Tried some tests, the hanging error is actually a little more nuanced. Making a new PR to explain it better.

PR #1237, opened 2022-06-06
Instruction distance check fix

CharlieL7:
* The hanging error occurs during validation when an instruction in a submodule has an argument from a parent module
* Normally, the check_order parameter will be set to false when the instruction is not found in the submodule. So std::distance will not be called between iterators from different containers.
* An error would occur if there is an instruction in the submodule that matches with the argument using the instruction operator==.
* The fix is to have the code for check_order compare the instruction_refs rather than the instructions.
* This bug does not occur in any compiled code that has run the eliminate_common_subexpression pass.
* Can revert the change to instruction::valid() to use the starting iterator instead of the module reference.

### DIFF
--- a/src/instruction.cpp
+++ b/src/instruction.cpp
@@ -74,14 +74,29 @@ bool operator==(const instruction& i, instruction_ref ref)
     return std::addressof(i) == std::addressof(*ref);
 }
 
+static void debug_name(std::ostream& os, const instruction& ins);
+
 bool instruction::valid(instruction_ref start, bool check_order) const
 {
+    // Need this lambda because std::distance has undefined behavior when comparing iterators of
+    // from different ranges
+    auto ins_dist = [](instruction_ref s, instruction_ref e) {
+        int dist = 0;
+        while((*s) != (*e))
+        {
+            ++s;
+            ++dist;
+        }
+        return dist;
+    };
+
     return valid() && std::all_of(arguments.begin(), arguments.end(), [&](instruction_ref i) {
                auto self = std::find(i->outputs().begin(), i->outputs().end(), *this);
                bool ret  = self != i->outputs().end();
                if(check_order)
                {
-                   ret = ret and (std::distance(start, i) < std::distance(start, *self));
+                   // arguments for this instruction before this instruction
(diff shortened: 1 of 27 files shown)


src/instruction.cpp
```diff
@@ -74,14 +74,15 @@ bool operator==(const instruction& i, instruction_ref ref)
...
+bool instruction::valid(const module& m, bool check_order) const
 {
     return valid() && std::all_of(arguments.begin(), arguments.end(), [&](instruction_ref i) {
                auto self = std::find(i->outputs().begin(), i->outputs().end(), *this);
                bool ret  = self != i->outputs().end();
                if(check_order)
                {
-                   ret = ret and (std::distance(start, i) < std::distance(start, *self));
+                   // check arguments for this instruction before this instruction
+                   ret = ret and (std::distance(m.begin(), i) < std::distance(m.begin(), *self));
```
pfultz2, line 85: This needs to check if the instruction is in the module here. I think we need to add an overload for `has_instruction` to take a pointer.

CharlieL7: I did something like `if (check_order and m.has_instruction(i))` in a previous commit. The change I made to `module.cpp:validate()` should make it such that check_order is only true if the correct instruction_ref is in the current module.

kahmed10 (approved):

test/module_test.cpp
```diff
@@ -312,4 +312,18 @@ TEST_CASE(module_without_bypass)
...
+    // Test when an instruction from a submodule depends on previous module
+    migraphx::program p;
+    auto* mm  = p.get_main_module();
+    auto* sub = p.create_module("sub");
+    auto l1   = mm->add_literal(migraphx::literal(3));
+    // second same literal to make sure instruction_ref is being compared, rather than the
+    // instructions
+    sub->add_literal(migraphx::literal(3));
+    sub->add_instruction(sum_op{}, l1, l1);
+    sub->validate();
```
pfultz2, line 326: This should do `EXPECT(sub->validate() == sub.end())`.

pfultz2 (approved):

PR #1245, opened 2022-06-14
Avoid registering host buffer ptr multiple times during hip copies

umangyadav:
Currently, while copying a host buffer to the device, it first registers/maps the host buffer pointer to address space of the device. 

If the host buffer has been allocated by the hipHostMalloc then, it is implicitly registered to the device's address space, and no need to register again. This PR adds a check for the same.


src/targets/gpu/hip.cpp
```diff
@@ -74,9 +74,15 @@ hip_ptr allocate_gpu(std::size_t sz, bool host = false)
 
 hip_host_ptr register_on_gpu(void* ptr, std::size_t sz)
 {
-    auto status = hipHostRegister(ptr, sz, hipHostRegisterMapped);
+    void* dev_ptr;
+    // check if ptr was allocated through hipHostMalloc
+    auto status = hipHostGetDevicePointer(&dev_ptr, ptr, 0);
```
kahmed10, line 79: This function returns 3 possible statuses. Should we be concerned with which error status is reported?

umangyadav: The other two error codes are 1. hipErrorInvalidValue which gets thrown whenever input `ptr` hipHostGetDevicePointer is not valid.  This is the error code that I get if try to fetch device pointer from host pointer that is not registered i.e. it is invalid. 

Second error code `hipErrorOutofMemory`, I am not sure in what situation it gets thrown for `hipHostGetDevicePointer`. Upon trying to find all references in ROCm software organization for that error code, it seems, it is used when `malloc`  fails.  

I changed the code and added explicit error status checks to be more precise.

kahmed10 (approved):

src/targets/gpu/hip.cpp
```diff
@@ -74,9 +74,19 @@ hip_ptr allocate_gpu(std::size_t sz, bool host = false)
 
 hip_host_ptr register_on_gpu(void* ptr, std::size_t sz)
 {
-    auto status = hipHostRegister(ptr, sz, hipHostRegisterMapped);
-    if(status != hipSuccess)
-        MIGRAPHX_THROW("Gpu register failed: " + hip_error(status));
+    void* dev_ptr;
```
pfultz2, line 77: Initialize the pointer: `void* dev_ptr = nullptr;`.

umangyadav:
> Actually, I dont think this can work, as one `hip_host_ptr` could unregister the memory while another `hip_host_ptr` is still using the memory(ie a submodule and parent module).
> 
> The only way to make it work is to use `shared_ptr`, but we need to track the pointers ourselves in a map of `weak_ptr`s. I am not sure if we need to track offsets into the buffer, hopefully not.

Added caching to avoid lifetime problems. But I don't find a way to test that without custom op.

umangyadav:
> > Actually, I dont think this can work, as one `hip_host_ptr` could unregister the memory while another `hip_host_ptr` is still using the memory(ie a submodule and parent module).
> > The only way to make it work is to use `shared_ptr`, but we need to track the pointers ourselves in a map of `weak_ptr`s. I am not sure if we need to track offsets into the buffer, hopefully not.
> 
> Added caching to avoid lifetime problems. But I don't find a way to test that without custom op.

@pfultz2 added a test for the lifetime.

src/targets/gpu/hip.cpp
```diff
@@ -77,29 +77,70 @@ void* get_device_ptr(void* hptr)
...
         else
             return allocate_gpu(sz, true);
     }
-    assert(result != nullptr);
-    return hip_ptr{result};
+    assert(alloc_ptr != nullptr);
+    std::shared_ptr<void> result = share(hip_ptr{alloc_ptr});
+    if(host)
+    {
+        get_malloc_host_ptrs()[alloc_ptr] = result;
```
pfultz2, line 103: This isnt thread-safe.

src/targets/gpu/hip.cpp
```diff
@@ -77,29 +77,70 @@ void* get_device_ptr(void* hptr)
     return result;
 }
 
-hip_ptr allocate_gpu(std::size_t sz, bool host = false)
+static auto& get_malloc_host_ptrs()
```
pfultz2, line 80: I dont think we need two seperate maps here since we search both of them anyways.

src/targets/gpu/hip.cpp
pfultz2: Since this is modified and read in two place we are going to need a class so the mutex stays with the map:

```cpp
struct host_ptr_cache
{
    std::unordered_map<void*, std::weak_ptr<void>> cache
    std::mutex m;

    std::shared_ptr<void> get(void* p)
    {
        std::lock_guard<std::mutex> lock(m);
        auto it = cache.find(ptr);
        if(it != cache.end())
            return it->second.lock();
        return nullptr;
    }

    void put(std::shared_ptr<void> p)
    {
        std::lock_guard<std::mutex> lock(m);
        cache[p.get()] = p;
    }
};

host_ptr_cache& get_host_ptr_cache()
{
    static host_ptr_cache cache;
    return cache;
}
```

src/targets/gpu/hip.cpp
umangyadav: Removed.

src/targets/gpu/hip.cpp
```diff
@@ -77,29 +77,71 @@ void* get_device_ptr(void* hptr)
     return result;
 }
 
-hip_ptr allocate_gpu(std::size_t sz, bool host = false)
+struct host_ptr_cache
+{
+    static auto& get_cache_map()
+    {
+        static std::unordered_map<void*, std::weak_ptr<void>> cache;
```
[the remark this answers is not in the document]
umangyadav: Yes, but for some reason, tidy was complaining about global variables. Running tidy again.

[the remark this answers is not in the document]
umangyadav: If i don't use function then, I have to define `cache` out of line in global scope to make it work. 
`std::unordered_map<void*, std::weak_ptr<void>> host_ptr_cache::cache = {};`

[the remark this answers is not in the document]
pfultz2: It shouldn't be static as the class is already constructed as a static variable:

```cpp
struct host_ptr_cache
{
    std::unordered_map<void*, std::weak_ptr<void>> cache{};
    std::mutex m;
    ...
};
```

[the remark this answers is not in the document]
pfultz2: > Yes, but for some reason, tidy was complaining about global variables. Running tidy again.

You might need to use `// NOLINT` for the global variable in the`get_host_ptr_cache()` function.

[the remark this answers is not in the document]
umangyadav: Done.

TedThemistokleous (approved):

PR #1249, opened 2022-06-16
Dynamic dimension input onnx parser

CharlieL7:
Depends on #1199 

* Adds ONNX parser functionality for dynamic input shapes.
* Uses `options` parameter in `parse_onnx()`


turneram (approved):

src/onnx/onnx.cpp
```diff
@@ -41,8 +41,17 @@ template <class... Ts>
 program parse_onnx_from(const onnx_options& options, Ts&&... xs)
 {
     onnx::onnx_parser parser;
-    parser.map_input_dims         = options.map_input_dims;
-    parser.default_dim_value      = options.default_dim_value;
+    parser.map_input_dims     = options.map_input_dims;
+    parser.map_dyn_input_dims = options.map_dyn_input_dims;
+    auto dim_val              = options.default_dim_value;
```
kahmed10, line 46: Since we haven't removed `default_dim_value` from our API, we should check that the user doesn't have conflicting values for `default_dim_value` and `default_dyn_dim_value`. Technically they could specify both, in which case we may want to throw an error.

CharlieL7: I've set it so `default_dim_value` has priority over `default_dyn_dim_value`. I will change it such that if both are set to something other than the defaults in `onnx.hpp` an error is thrown.

PR #1250, opened 2022-06-17
Fixing misspelled macro to enable MIOpen hidden find mode API

jerryyin:
Looks like an unexpected mis-spelling caused MIGraphx to by default use dynamic hybrid mode instead of find mode. According to below, it should use the cmake defined macro instead of a cmake variable.

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/c99be32c013a21c41f6ad31154fc12c03eac1124/src/targets/gpu/CMakeLists.txt#L349-L354


pfultz2 (approved):

kahmed10 (approved):

src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -11,7 +11,7 @@
 
 #include <sstream>
 
-#ifdef HAS_FIND_MODE_API
+#ifdef MIGRAPHX_HAS_FIND_MODE_API
 extern "C" miopenStatus_t miopenHiddenSetConvolutionFindMode(miopenConvolutionDescriptor_t convDesc,
```
pfultz2, line 15: You need to add `// NOLINT` to suppress the tidy warnings.

jerryyin: Oops you are right, done in 1cf219e.

PR #1255, opened 2022-06-21
Adding in check_stamped.py to tools/

TedThemistokleous:
Used to determine what files contain a license and are stamped. If not we exit and return an error code that can be later ingested by another script, as well as a list of the outstanding files in questions.

Currently baked in the list of files we should support or not support with licenses in them a well as some stuff to quickly ignore


umangyadav:
I think the idea was to add it as a CI check on Github Action. Otherwise LGTM.

.github/workflows/ci.yaml
```diff
@@ -160,6 +160,12 @@ jobs:
         mypy --version
         mypy tools/api.py
 
+  Licensing:
```
umangyadav, line 163: Can you add steps to free up the space like other CI checks? I remember in past we had some problems with available free space and had to include that step to free up more space.

tools/check_stamped.py
```diff
@@ -0,0 +1,151 @@
...
+
+    for key in fileTuple:
+        if key in filename:
+            supported = True
+            break
+
+    return supported
+
+
+debug = False
```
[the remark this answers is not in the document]
TedThemistokleous: Done

.github/workflows/ci.yaml
```diff
@@ -160,6 +160,19 @@ jobs:
...
+    - name: Free space
+      run: sudo rm -rf /usr/local/android /usr/share/dotnet /usr/local/share/boost /opt/ghc /usr/local/share/chrom* /usr/share/swift /usr/local/julia* /usr/local/lib/android
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    runs-on: ubuntu-18.04
```
umangyadav, line 171: Can you move runs-on on before all steps ?

umangyadav (approved):

.github/workflows/ci.yaml
pfultz2: can we make this lowercase?

TedThemistokleous: Sure!

tools/check_stamped.py
```diff
@@ -0,0 +1,148 @@
...
+
+    #add general stuff we shouldn't stamp and any exceptions here
+    unsupported_file_types = [
+        ".onnx", ".pb", ".rst", ".jpg", ".jpeg", ".proto", ".md", ".clang",
+        ".cmake", ".weight", ".ini", ".json", ".docker", ".git", ".rules",
+        ".yml"
+    ]
+
+    #Specifics to ignore
+    specificIgnores = ("digits.txt", "Dockerfile", "Jenkinsfile", "")
```
pfultz2, line 100: Maybe we should move these variables up to the top, since this is probably what we would change in the future.

tools/check_stamped.py
```diff
@@ -0,0 +1,148 @@
...
+def main():
+    # The filetypes we want to check for that are stamped
+    # LICENSE is included here as it SHOULD have a liscence in it otherwise flag it as unstamped
+    supported_file_types = (".cpp", ".hpp", ".h", ".ipynb", ".py", ".txt",
+                            ".sh", ".bsh", "LICENSE")
+
+    #add general stuff we shouldn't stamp and any exceptions here
+    unsupported_file_types = [
+        ".onnx", ".pb", ".rst", ".jpg", ".jpeg", ".proto", ".md", ".clang",
+        ".cmake", ".weight", ".ini", ".json", ".docker", ".git", ".rules",
```
pfultz2, line 95: I think we want to check for .cmake files since we do add license to CMakeLists.txt files. We can do that as a follow-up PR.

TedThemistokleous: Okay. That should be an easy change/PR after this one then.

tools/check_stamped.py
pfultz2: all three of them.

PR #1257, opened 2022-06-21
NMS refactor, enable nonstandard shape

CharlieL7:
* Allow for nonstandard input shapes
  * allows PyTorch converted version of SSD-resnet34 to work: [model link](https://zenodo.org/record/3228411/files/resnet34-ssd1200.onnx)
* Refactor `compute()`:
  * avoid raw pointers
  * use visitor pattern

### DIFF
--- a/examples/migraphx/custom_op_miopen_kernel/CMakeLists.txt
+++ b/examples/migraphx/custom_op_miopen_kernel/CMakeLists.txt
@@ -1,0 +1,37 @@
+#####################################################################################
+# The MIT License (MIT)
+#
+# Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
+cmake_minimum_required(VERSION 3.5)
+project (custom_miopen_kernel)
+
+set (CMAKE_CXX_STANDARD 14)
+set (EXAMPLE custom_op_miopen_kernel)
+
+list (APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)
+find_package (migraphx REQUIRED)
+find_package (miopen REQUIRED)
+
+message("source file: " ${EXAMPLE}.cpp " ---> bin: " ${EXAMPLE})
+add_executable(${EXAMPLE} ${EXAMPLE}.cpp)
+
+target_link_libraries(${EXAMPLE} migraphx::c MIOpen)
--- a/examples/migraphx/custom_op_miopen_kernel/README.md
+++ b/examples/migraphx/custom_op_miopen_kernel/README.md
@@ -1,0 +1,10 @@
+# Custom MIOpen Kernel using MIGraphX API. 
+ This is an example of a custom operator implementation using MIGraphX's C/C++ APIs. It also demonstrates how to use this custom op in conjunction with rest of MIGraphX operators to build  and run MIGraphX program on GPU. 
+ Kernels can be written in either HIP, MIOpen, or by using RocBLAS library.  This particular example uses **MIOpen** library calls.
+
+ To build and run example, ensure ROCm is installed at `/opt/rocm`. 
+ 1.  `export LD_LIBRARY_PATH=/opt/rocm/lib:$LD_LIBRARY_PATH`
+ 2.  `cd $MIGRAPHX_SRC/examples/migraphx/custom_op_miopen_kernel/`
+ 3.  `mkdir build && cd build`
+ 4.  `cmake ..  && make`
+ 5.  `./custom_op_miopen_kernel`
--- a/examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
+++ b/examples/migraphx/custom_op_miopen_kernel/custom_op_miopen_kernel.cpp
@@ -1,0 +1,172 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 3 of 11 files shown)


src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -128,100 +127,95 @@ struct nonmaxsuppression
...
+                    {
+                        // Check with existing selected boxes for this class, remove box if it
+                        // exceeds the IOU (Intersection Over Union) threshold
+                        const auto next_top_score = boxes_heap.top();
+                        bool not_selected         = std::any_of(
+                            selected_boxes_inside_class.begin(),
+                            selected_boxes_inside_class.end(),
+                            [&](auto selected_index) {
+                                return this->suppress_by_iou(
+                                    batch_box(boxes, batch_boxes_ind, next_top_score.second),
```
pfultz2, line 202: It would be better to pass in the start with `boxes.begin() + (batch_idx * num_boxes * 4)`.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -71,23 +70,23 @@ struct nonmaxsuppression
     };
 
     template <class T>
-    box batch_box(const T* boxes, std::size_t bidx) const
+    box batch_box(const T& boxes, std::size_t box_ind, std::size_t box_idx) const
     {
         box result{};
-        const T* start = boxes + 4 * bidx;
+        auto start = box_ind + 4 * box_idx;
```
pfultz2, line 76: This should be `auto start = boxes + 4 * box_idx` if we take the iterators.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -128,100 +127,95 @@ struct nonmaxsuppression
         return intersection_over_union > iou_threshold;
     }
 
+    // filter boxes below score_threshold
+    template <class T>
+    void filter_boxes_by_score(T scores,
+                               std::size_t score_offset_ind,
+                               std::size_t num_boxes,
+                               const float score_threshold,
+                               std::priority_queue<std::pair<float, int64_t>>& boxes_heap) const
```
pfultz2, line 136: Return this instead of using an out parameter.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -128,100 +127,95 @@ struct nonmaxsuppression
...
+
+        result.visit([&](auto output) {
+            visit_all(args[0], args[1])([&](auto boxes, auto scores) {
+                std::fill(output.begin(), output.end(), 0);
+                const auto& lens       = scores.get_shape().lens();
+                const auto num_batches = lens[0];
+                const auto num_classes = lens[1];
+                const auto num_boxes   = boxes.get_shape().lens()[1];
+                // boxes of a class with NMS applied [score, index]
+                std::vector<std::pair<float, int64_t>> selected_boxes_inside_class;
```
pfultz2, line 173: We usually use double in the ref implementation to get better accuracy, maybe this would help here too.

CharlieL7: The ONNX spec actually has both the boxes and scores limited to floats. https://github.com/onnx/onnx/blob/main/docs/Operators.md#NonMaxSuppression
I think that's why this operator didn't use the visitor before. I made the change to be consistent with the other operators and in case the spec changes later on.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -128,100 +127,93 @@ struct nonmaxsuppression
...
     }
 
+    // filter boxes below score_threshold
+    template <class T>
+    std::priority_queue<std::pair<float, int64_t>> filter_boxes_by_score(
+        T scores, std::size_t score_offset_ind, std::size_t num_boxes, float score_threshold) const
+    {
+        std::priority_queue<std::pair<float, int64_t>> boxes_heap;
+        auto insert_to_boxes_heap =
+            make_function_output_iterator([&boxes_heap](const auto& x) { boxes_heap.push(x); });
```
umangyadav, line 137: just keep `[&]` in lambda.

umangyadav:
Can you add a test with non_standard_shape ?

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -33,7 +33,6 @@ struct nonmaxsuppression
     shape compute_shape(std::vector<shape> inputs) const
     {
         // requires at least 2 inputs
-        check_shapes{inputs, *this}.standard();
```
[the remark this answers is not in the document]
umangyadav: What kind of non-standard shapes do you expect to see ? 

If they are non-standard i.e. let's say transpose then, boxes could be = [num_of_boxes, batches, 4]. We would need to update reference indices to these dimensions everywhere in the code.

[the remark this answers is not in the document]
CharlieL7: Yeah, this code is not going to handle non-standard shapes correctly. I think the reason why it works for the resnet34-ssd model is because our `auto_contiguous{}` pass makes the shape standard before it gets to this instruction. However, with the standard shape check the model doesn't get to that pass. I'll look into what I can do to use the strides rather than the fixed offsets.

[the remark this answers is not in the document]
umangyadav: > > If they are non-standard i.e. let's say transpose then, boxes could be = [num_of_boxes, batches, 4]
> 
> The dimensions should be the same order regardless of the order in memory. If they store the dimensions as `[num_of_boxes, batches, 4]` in memory, then a transpose will be needed to put the dimensions in the correct order of `[num_batches, spatial_dimension, 4]`. Thus we would need to support a transposed shape.

Agree, i am just saying current changes are not accounting for the fact that inputs could be transposed. 

e.g.  hardcoding following
```
        auto batch_num   = lens[0];
        auto class_num   = lens[1];
        auto box_num     = args.at(0).get_shape().lens()[1];
```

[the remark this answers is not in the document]
umangyadav: > > If they are non-standard i.e. let's say transpose then, boxes could be = [num_of_boxes, batches, 4]
> 
> The dimensions should be the same order regardless of the order in memory. If they store the dimensions as `[num_of_boxes, batches, 4]` in memory, then a transpose will be needed to put the dimensions in the correct order of `[num_batches, spatial_dimension, 4]`. Thus we would need to support a transposed shape.

Would that mean, MIGraphX would have to insert extra transposes to make shape requirement adhere to [num_batches, spatial_dimension, 4]` ? Do we do that currently ? I guess no.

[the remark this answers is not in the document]
pfultz2: > Agree, i am just saying current changes are not accounting for the fact that inputs could be transposed.

Yea, but when we say the shape is transposed we mean the memory layout is transposed not the dimensions.

> Would that mean, MIGraphX would have to insert extra transposes to make shape requirement adhere to [num_batches, spatial_dimension, 4] ?

Well the user would insert the transposes as necessary to get the dimensions in the correct order. This could happen right before the call but it could be precede several other shape manipulations as well.

test/ref_ops_test.cpp
```diff
@@ -3164,6 +3164,80 @@ TEST_CASE(nms_test)
...
+
+    p.compile(migraphx::ref::target{});
+    auto output = p.eval({}).back();
+    std::vector<int64_t> result;
+    output.visit([&](auto out) { result.assign(out.begin(), out.end()); });
+    std::vector<int64_t> gold = {0, 0, 3, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+    EXPECT(migraphx::verify_range(result, gold));
+}
+
+TEST_CASE(nms_transpose2_test)
```
umangyadav, line 3204: You can also add a verify test.  e.g. `verify/test_nms.cpp`.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -128,123 +135,136 @@ struct nonmaxsuppression
...
+    }
 
-        if(args.size() > 2)
-        {
-            max_output_boxes_per_class = args.at(2).at<std::size_t>();
-        }
-        // max_output_boxes_per_class is 0, no output
-        if(max_output_boxes_per_class == 0)
+    template <class H, class S>
+    void select_boxes(H& boxes_heap,
```
pfultz2, line 183: Is this a seperate function because of the nesting issue? I think it would be better to have a function with just the tensors right after the visit:

```cpp
template<class Output, class Boxes, class Scores>
void compute_nms(Ouput output, 
                 Boxes boxes,
                 Scores scores,
                 double iou_threshold,
                 double score_threshold) const
{
    std::fill(output.begin(), output.end(), 0);
    const auto& lens       = scores.get_shape().lens();
    const auto num_batches = lens[0];
    const auto num_classes = lens[1];
    const auto num_boxes   = lens[2];
    ...
}
```

CharlieL7: Made the change

pfultz2 (approved):

PR #1258, opened 2022-06-22
Improve help and error reporting in driver

pfultz2:
This improves the error reporting. If running with `./bin/driver read bert.onnx --batch max`, it will output:

```
error: Invalid input to '--batch [unsigned int]'
       Failed to parse 'max' as unsigned int

USAGE:
    ./bin/driver read --batch [unsigned int]

For more information try '--help'
```

Also, it will show errors if flags are passed in before the command, like `./bin/driver --batch 1 read bert.onnx`

```
error: command 'read' must be first argument

USAGE:
    ./bin/driver read <options>
```

Where as before it just ran silently.

Help is also updated to better show usage, and makes the flags more compact:

```
USAGE:
    ./bin/driver read <options> <input file>

FLAGS:

    -h, --help                     Show help
    --onnx                         Load as onnx
    --tf                           Load as tensorflow
    --migraphx                     Load as MIGraphX
    --migraphx-json                Load as MIGraphX JSON
    --nhwc                         Treat tensorflow format as nhwc
    --skip-unknown-operators       Skip unknown operators when parsing and continue to parse.
    --nchw                         Treat tensorflow format as nchw
    --optimize, -O                 Optimize when reading
    --graphviz, -g                 Print out a graphviz representation.
    --brief                        Make the output brief.
    --cpp                          Print out the program as cpp program.
    --json                         Print out program as json.
    --text                         Print out program in text format.
    --binary                       Print out program in binary format.

OPTIONS:

    <input file> [std::string]


    --model [resnet50|inceptionv3|alexnet]
        Load model

    --batch [unsigned int]
        Set batch size for model

    --trim, -t [unsigned int]
        Trim instructions from the end

    --input-dim [std::vector<std::string>]
        Dim of a parameter (format: "@name d1 d2 dn")

    --output-names [std::vector<std::string>]
        Names of node output (format: "name_1 name_2 name_n")

    --output, -o [std::string]
        Output to file.
```


kahmed10:
This is a big improvement, but here are a few suggestions:
If the user passes in `./driver read` with no file argument, then the error message doesn't point out that the user failed to pass in a filename. Similarly, if a flag such as `./driver read --onnx` is passed in, it won't give the same error message of "flag with no value".

pfultz2:
I think its a good suggestion but it will require more work to implement as we need to annotate which arguments are required. Can we open an issue to track these changes in the future?

pfultz2:
I updated it to support required arguments, so for `./bin/driver read --onnx` it now shows:

```
error:  The following required arguments were not provided:
       (<input file>|--model [resnet50|inceptionv3|alexnet])

USAGE:
    ./bin/driver read (<input file>|--model [resnet50|inceptionv3|alexnet])

For more information try '--help'
```

TedThemistokleous (approved):

TedThemistokleous (approved):
Add changes suggested by @kahmed10 into an issue so we can track future work pertaining to this improvement.

PR #1265, opened 2022-06-23
Verify load and save

pfultz2:
In the verification tests, check that saving and reloading the program is the same program. This also fixes serialization to always load instructions in the same order. There is also fixes for deconv and quant_conv which didn't save the solution id, and was broken for serialization.


bpickrel (commented):
Looks like *.mxr files from the save/load test aren't being cleaned up afterwards?

pfultz2:
> Looks like *.mxr files from the save/load test aren't being cleaned up afterwards?

Where are you seeing left over .mxr files? I dont see them:

```
$ echo /tmp/migraphx-migraphx_test-*
/tmp/migraphx-migraphx_test-*
```

src/targets/gpu/quant_convolution.cpp
```diff
@@ -114,25 +114,65 @@ shape miopen_quant_convolution::compile(context& ctx,
...
-        MIGRAPHX_THROW("Workspace has changed during finalization.");
+    if(cd == nullptr)
+        cd = make_conv(op);
+    if(solution_id == 0)
+    {
+        // Check that workspace hasn't changed
+        auto size = inputs.at(2).bytes();
+        auto ws   = find(ctx, output_shape, inputs);
+        if(ws.bytes() > size)
+            MIGRAPHX_THROW("MIOpen uant Convolution: workspace has changed during finalization.");
```
kahmed10, line 161:
```suggestion
            MIGRAPHX_THROW("MIOpen Quant Convolution: workspace has changed during finalization.");
```

kahmed10 (approved):

PR #1267, opened 2022-06-23
bug fix: register the miopen_fusion op.

bpickrel:
One-line fix to register the op miopen_fusion.  This error was causing loading of compiled model files (*.mxr) to fail.

[PR 11265](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1265) (currently in review) will add save/load tests for MXR files to `test_verify`, which should catch things like this in future.  

Need to do:  add a case to `test_verify` which will invoke `miopen_fusion` op.


pfultz2 (approved):

turneram (approved):

bpickrel:
A **similar-looking error** in *.mxr loads with a message **"Failed to parse: max"** (or avg) is due to a version mismatch.  It's caused by attempting to load a *.mxr model file created with a migraphx version older than (I think) commit a2e90b5dc13f2f190d50c53bc079f7136aeee89f from Mar 4, 2022 when pooling_mode for pooling ops was changed from a string to an enum.

Issue #1236, opened 2022-06-06
Add algebraic simplications for mul, div, add etc. 


umangyadav:
x * 1 = x
x / 1 = x
x * 0  = 0
x + 0 = x
x - 0 = x
x * 0 = 0 
x * -1 = -x 
x  / -1 = -x 
If x / 0 is detected anywhere throw errors. 

~~Consecutive const binary operations. 
y = (x + c1) + c2 --- > y = x + (c1+c2) -- > y = x + c3 where c3  = c1 + c2 const folded. 
y = ( x * c1 ) * c2 --- > y = x * (c1 * c2)~~
etc...

TedThemistokleous:
Got something in the works for x*1 = x. 

Getting the odd identity(x) operation popping out. Thanks to your recent help, I was able to handle the 

x*1 and 1*x cases with matchers. 

Currently breaking another test for this though.

TedThemistokleous:
Awesome. Thanks

umangyadav:
Fixed with https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1281

PR #1268, opened 2022-06-24
Simplify algebra for  x / 1 operations

TedThemistokleous:
Done to satisfy simplifications specified by #1236 .  Just replace every  parameter divided by 1 with itself. It's assumed that the eliminate_identity() pass will handle generated identity operators in our run_pass()

### DIFF
--- a/test/simplify_algebra_test.cpp
+++ b/test/simplify_algebra_test.cpp
@@ -740,14 +740,15 @@ TEST_CASE(simplify_unit_div_const)
     {
         auto x    = m1.add_parameter("x", {migraphx::shape::int32_type, {1}});
         auto unit = m1.add_literal(1);
-        m1.add_instruction(migraphx::make_op("div"), x, unit);
+        auto div  = m1.add_instruction(migraphx::make_op("div"), x, unit);
+        m1.add_return({div});
     }
     run_pass(m1);
 
     migraphx::module m2;
     {
         auto x = m2.add_parameter("x", {migraphx::shape::int32_type, {1}});
-        m2.add_instruction(migraphx::make_op("identity"), x);
+        m2.add_return({x});
     }
 
     EXPECT(m1 == m2);
--- a/test/simplify_algebra_test.cpp
+++ b/test/simplify_algebra_test.cpp
@@ -734,6 +734,25 @@ TEST_CASE(simplify_div_const)
     EXPECT(m1 == m2);
 }
 
+TEST_CASE(simplify_unit_div_const)
+{
+    migraphx::module m1;
+    {
+        auto x    = m1.add_parameter("x", {migraphx::shape::int32_type, {1}});
+        auto unit = m1.add_literal(1);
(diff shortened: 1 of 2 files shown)


umangyadav (approved):

umangyadav:
@pfultz2  We are not running `simplify_algebra` after propagate_constant, I think we may miss some optimizations in doing so.

test/simplify_algebra_test.cpp
```diff
@@ -734,6 +734,25 @@ TEST_CASE(simplify_div_const)
     EXPECT(m1 == m2);
 }
 
+TEST_CASE(simplify_unit_div_const)
+{
+    migraphx::module m1;
+    {
+        auto x    = m1.add_parameter("x", {migraphx::shape::int32_type, {1}});
+        auto unit = m1.add_literal(1);
+        m1.add_instruction(migraphx::make_op("div"), x, unit);
```
pfultz2, line 743: I think you can add a return or just add in an extra operator so it doesnt get replaced with identity:

```cpp
        auto unit = m1.add_literal(1);
        auto div = m1.add_instruction(migraphx::make_op("div"), x, unit);
        m1.add_instruction(migraphx::make_op("relu"), div);
```

```cpp
        auto unit = m1.add_literal(1);
        auto div = m1.add_instruction(migraphx::make_op("div"), x, unit);
        m1.add_return({div});
```

umangyadav:
> > Should we move these 'identity' transformations in another pass and call it after propagate_constant ?
> 
> No. Ideally, I would like to make a super pass that has several optimizations in a loop like simplify_algebra, simplify_reshapes, CSE, propagate_constant, etc.

That is a better idea, but ideally, we should loop into that until the program stop changing. I am not sure given the transformations we have, it is guaranteed to converge, (it could be, but not sure). Otherwise, loop trip count becomes a tuning parameter that would affect performance.

pfultz2:
> That is a better idea, but ideally, we should loop into that until the program stop changing.

Yea, we need a way to track changes to the module. Perhaps a signal? Or we could have finders return if they change the graph?

> Otherwise, loop trip count becomes a tuning parameter that would affect performance.

Well we already have loop counts already. Either way, we would want to have a max loop count , but we could have a compile_options flag to increase it.

TedThemistokleous:
> > That is a better idea, but ideally, we should loop into that until the program stop changing.
> 
> Yea, we need a way to track changes to the module. Perhaps a signal? Or we could have finders return if they change the graph?
> 
> > Otherwise, loop trip count becomes a tuning parameter that would affect performance.
> 
> Well we already have loop counts already. Either way, we would want to have a max loop count , but we could have a compile_options flag to increase it.

Possible another PR then? Seems like this would be more involved.

TedThemistokleous:
Total time:   3.37061ms ->   3.33602ms  with this change set off develop

TedThemistokleous:
Closing this out as I've moved the changes for a matcher to be shared between other operations.

PR #1269, opened 2022-06-24
Add is_supported and get_target_assignments

varunsh-xilinx:
As discussed in the UIF proposal, we're adding `is_supported` and `get_target_assignments` methods to the `target` and `program`, respectively, to eventually support multi-target compilation and execution.

`get_target_assignments` will assign the instructions of the program to run on one of the available targets and calls `is_supported` to determine which instructions are supported on which targets and to what degree.

For now, these functions are placeholders for more sophisticated algorithms in the future.


varunsh-xilinx:
@pfultz2 why is the test failing here? The existing `targets.cpp` test does something similar.

This test also runs fine on my local system.

pfultz2 (approved):

pfultz2:
> why is the test failing here? The existing targets.cpp test does something similar.

Try adding an include for `#include <migraphx/ref/target.hpp>`. For some reason, the targets dont always get registered until it references symbols from the library.

PR #1272, opened 2022-06-24
Simplify algebra x mult 1

TedThemistokleous:
As per #1236 , fixing our tests and adding the simplification for x*1 operations for each pass using matchers. 

Had to modify the simplify mul_add case since it was erroneously triggering a failure since the literal used was a 1 and was being prematurely optimized.


src/simplify_algebra.cpp
```diff
@@ -851,6 +851,31 @@ struct find_div_const
...
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins  = r.result;
+        auto args = ins->inputs();
+
+        auto c_in = r.instructions["x"];
+
+        auto res = args.front();
```
umangyadav, line 869: I don't think you need to check and replace for the order for `x`. `x` is bound to the non-identity argument to `ins`. You should be able to just replace `ins` with `c_in`.

TedThemistokleous: Oh thanks for pointing that out! Just rebuilt and tried that and you're correct since the test case I added checked for commutativity. So for future searches, because I'm using either_arg(0,1) the bind will always match the "other" thing then?

umangyadav: Yes, bind will match to instruction that is not `1`

TedThemistokleous:
Closing this for preference in changes that unify things to a single matcher.

PR #1273, opened 2022-06-27
Simplifiy Algebra for x*(-1) operations to simply negative x

TedThemistokleous:
Save a multiply operation with that of a negation of  input parameter x. Suggested improvement via #1236

Not sure if we want this in right away. Trying this change ontop of the previous multiply by 1 change results in not much of a change in total time running:

MIOPEN_FIND_ENFORCE=3 bin/driver perf    for resnet50-v2-7

### DIFF
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -870,6 +870,26 @@ struct find_unit_mult_const
     }
 };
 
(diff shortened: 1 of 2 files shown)


TedThemistokleous:
Closing this and putting things into one PR for with shared matchers between cases.

PR #1274, opened 2022-06-27
Improve error reporting in the API

umangyadav:
C++ API is not printing thrown exception string. this improves on it.

### DIFF
--- a/src/api/include/migraphx/migraphx.hpp
+++ b/src/api/include/migraphx/migraphx.hpp
@@ -346,19 +346,21 @@ struct interface_base : Base
 
     protected:
     template <class F>
-    static migraphx_status try_(F f, std::exception_ptr* eptr = nullptr) // NOLINT
+    static migraphx_status try_(F f, const char** ex_msg = nullptr) // NOLINT
     {
         try
         {
             f();
             return migraphx_status_success;
         }
-        catch(...)
+        catch(std::exception& ex)
         {
-            if(eptr)
-            {
-                *eptr = std::current_exception();
+            if(ex_msg) {
+                *ex_msg = ex.what();
             }
+            return migraphx_status_unknown_error;
+        }
+        catch(...) {
             return migraphx_status_unknown_error;
         }
     }
@@ -391,9 +393,8 @@ struct interface_base : Base
         (void)f; // avoid warning on gcc
         call(setter,
              this->get_handle_ptr(),
-             [](auto out, void* obj, std::exception_ptr* eptr, auto... xs) -> migraphx_status {
-                 auto status = try_([&] { call_cast_arg<T>(rank<1>{}, f, out, obj, xs...); }, eptr);
-                 return status;
+             [](auto out, void* obj, const char** ex_msg, auto... xs) -> migraphx_status {
+                return try_([&] { call_cast_arg<T>(rank<1>{}, f, out, obj, xs...); }, ex_msg);
              });
     }
 
--- a/src/api/include/migraphx/migraphx.hpp
+++ b/src/api/include/migraphx/migraphx.hpp
@@ -394,8 +394,10 @@ struct interface_base : Base
         (void)f; // avoid warning on gcc
         call(setter,
              this->get_handle_ptr(),
-             [](auto out, void* obj, char* ex_msg, size_t ex_msg_size, auto... xs) -> migraphx_status {
-                 return try_([&] { call_cast_arg<T>(rank<1>{}, f, out, obj, xs...); }, ex_msg, ex_msg_size);
+             [](auto out, void* obj, char* ex_msg, size_t ex_msg_size, auto... xs)
+                 -> migraphx_status {
+                 return try_(
+                     [&] { call_cast_arg<T>(rank<1>{}, f, out, obj, xs...); }, ex_msg, ex_msg_size);
              });
     }
 
--- a/src/api/include/migraphx/migraphx.hpp
+++ b/src/api/include/migraphx/migraphx.hpp
@@ -23,6 +23,7 @@
  */
 #ifndef MIGRAPHX_GUARD_API_RTGLIB_MIGRAPHX_HPP
 #define MIGRAPHX_GUARD_API_RTGLIB_MIGRAPHX_HPP
+#include "migraphx.h"
 #include <cxxabi.h>
 #include <initializer_list>
 #include <migraphx/migraphx.h>
--- a/src/api/include/migraphx/migraphx.hpp
+++ b/src/api/include/migraphx/migraphx.hpp
@@ -25,6 +25,7 @@
 #define MIGRAPHX_GUARD_API_RTGLIB_MIGRAPHX_HPP
 
 #include "migraphx.h"
+#include <cstring>
 #include <initializer_list>
 #include <migraphx/migraphx.h>
 #include <memory>
@@ -346,7 +347,7 @@ struct interface_base : Base
 
     protected:
     template <class F>
-    static migraphx_status try_(F f, const char** ex_msg = nullptr) // NOLINT
+    static migraphx_status try_(F f, char* ex_msg = nullptr, size_t ex_msg_size = 0) // NOLINT
     {
         try
         {
@@ -356,9 +357,7 @@ struct interface_base : Base
         catch(const std::exception& ex)
         {
             if(ex_msg)
-            {
-                *ex_msg = ex.what();
-            }
+                std::strncpy(ex_msg, ex.what(), ex_msg_size);
             return migraphx_status_unknown_error;
         }
         catch(...)
@@ -395,8 +394,8 @@ struct interface_base : Base
         (void)f; // avoid warning on gcc
         call(setter,
              this->get_handle_ptr(),
-             [](auto out, void* obj, const char** ex_msg, auto... xs) -> migraphx_status {
-                 return try_([&] { call_cast_arg<T>(rank<1>{}, f, out, obj, xs...); }, ex_msg);
+             [](auto out, void* obj, char* ex_msg, size_t ex_msg_size, auto... xs) -> migraphx_status {
+                 return try_([&] { call_cast_arg<T>(rank<1>{}, f, out, obj, xs...); }, ex_msg, ex_msg_size);
              });
     }
 
--- a/src/api/include/migraphx/migraphx.hpp
+++ b/src/api/include/migraphx/migraphx.hpp
@@ -23,8 +23,8 @@
  */
 #ifndef MIGRAPHX_GUARD_API_RTGLIB_MIGRAPHX_HPP
 #define MIGRAPHX_GUARD_API_RTGLIB_MIGRAPHX_HPP
+
 #include "migraphx.h"
-#include <cxxabi.h>
 #include <initializer_list>
 #include <migraphx/migraphx.h>
 #include <memory>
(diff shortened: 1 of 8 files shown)


kahmed10 (approved):

tools/api.py
```diff
@@ -1059,7 +1059,7 @@ def add_cpp_class(self) -> None:
         throw std::runtime_error("${name} function is missing.");
     auto api_error_result = ${fname}(${args});
     if (api_error_result != ${success})
-        throw std::runtime_error("Error in ${name}.");
+        throw std::runtime_error("Error in ${name} of: " + xobject.name);
```
pfultz2, line 1062: There is no guarantee that `xobject` will have a `name` member field. It works for custom op, but future interfaces may not have a `name` member field. You should use `typeid` instead: `typeid(object_ptr.data)` or add a method to `manage_generic_ptr`

src/api/include/migraphx/migraphx.hpp
```diff
@@ -317,6 +317,11 @@ struct interface_base : Base
             f();
             return migraphx_status_success;
         }
+        catch(const std::exception& ex)
+        {
+            std::cerr << "MIGraphX Error: " << ex.what() << std::endl;
```
pfultz2, line 322: I think printing the error might be quite noisy, especially since this is used in `eliminate_contiguous` to decide if `contiguous` can be removed. It could also be confusing since there could be printouts of "MIGraphX Error" when there was no error.

We probably need to think of a better to transport these messages. Perhaps we store them in a thread-local variable? Or should we generate a `store_error_message` methods for interface handles?

umangyadav: this is C++ API,  can you provide details on how it would catch exception from `eliminate_contiguous` ?

umangyadav: > It could also be confusing since there could be printouts of "MIGraphX Error" when there was no error.

It would print whenever there is exception, Wouldn't exception mean there was something wrong or error ?

tools/api.py
umangyadav: But wouldn't the `typeid` return some mangled name ?  for now i find using `xobject.name` clearer.  If we change interface in future, we can change this one too.

src/api/include/migraphx/migraphx.hpp
pfultz2: > can you provide details on how it would catch exception from eliminate_contiguous ?

It removes `contiguous` as an input and then calls `compute_shape` if it throws an exception it will keep the `contiguous` operator. We dont want an "MIGraphX Error" on every try. 

> Wouldn't exception mean there was something wrong or error ?

Yes, but it could have been recovered so not a fatal error, and its not something a user would need to fix.

It could also lead QA assigning issues to us because it sees "MIGraphX Error" when the error is not from migraphx at all.

tools/api.py
pfultz2: > But wouldn't the typeid return some mangled name ?

I think it is for gcc and clang unfortunately. We could try to store the type name in the `manage_generic_ptr` class(as computed from [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/type_name.hpp#L34) which will need to copy into migaphx.hpp), but we need to update api.py to add the typename parameters along with current copy constructor and destructor parameters:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/tools/api.py#L1106

And then pass the typename to `make_handle` here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/api/include/migraphx/migraphx.hpp#L344

> for now i find using xobject.name clearer.

Interfaces dont provide a `.name` so we can't use it even if it is clearer.

> If we change interface in future, we can change this one too.

`.name` donesn't come from the interface, its an implementation detail of custom op, which is the problem. This will prevent us from adding interfaces in the future as they will fail to compile since they would be missing `.name` in its implementation.

umangyadav: It prints `type` of the custom_op struct not really name of the custom_op though.

tools/api.py
```diff
@@ -790,6 +792,7 @@ def decorated(*args, **kwargs):
     }
 
     void* data = nullptr;
+    std::string ptr_typename = "";
```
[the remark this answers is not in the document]
umangyadav: That should work.  That's a better idea.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -310,13 +346,21 @@ struct interface_base : Base
...
         try
         {
             f();
             return migraphx_status_success;
         }
+        catch(std::exception& ex)
+        {
+            if(ex_msg)
+            {
+                *ex_msg = ex.what();
```
pfultz2, line 360: This will be a dangling reference upon returning. I think you will need to provide pointers to a char array that you copy into.

tools/api/api.cpp
```diff
@@ -262,14 +261,17 @@ struct custom_operation
 
     shape compute_shape(std::vector<shape> inputs) const
     {
-        return op.compute_shape(std::move(inputs));
+        const char* exception_msg = " ";
```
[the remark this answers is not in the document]
umangyadav: auto-generated.

src/api/include/migraphx/migraphx.hpp
umangyadav: I am using pointer to char_array. Moving `ex.what()` to `*ex_msg`

tools/api.py
```diff
@@ -1057,9 +1062,11 @@ def add_cpp_class(self) -> None:
     ${output_decls}
     if (${fname} == nullptr)
         throw std::runtime_error("${name} function is missing.");
+    const char* exception_msg = " ";
```
pfultz2, line 1065: This should be an array: `std::array<char, 256> exception_msg` so it can copy the message.

tools/api.py
```diff
@@ -1057,9 +1062,11 @@ def add_cpp_class(self) -> None:
...
     auto api_error_result = ${fname}(${args});
+    const std::string exception_str(exception_msg);
```
pfultz2, line 1067: We should only construct the string when there is an error. So this could be moved to the if branch below.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -310,13 +346,21 @@ struct interface_base : Base
...
     template <class F>
-    static migraphx_status try_(F f) // NOLINT
+    static migraphx_status try_(F f, const char** ex_msg = nullptr) // NOLINT
     {
         try
         {
             f();
             return migraphx_status_success;
         }
+        catch(std::exception& ex)
```
pfultz2, line 356: This should be const.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -310,13 +346,21 @@ struct interface_base : Base
...
+        {
+            if(ex_msg)
+            {
+                *ex_msg = std::move(ex.what());
```
pfultz2, line 360: This is still a dangling reference. Upon returning, `ex` lifetime ends which means any pointers to `ex` or its internal members(which is what `ex.what()` is returning) is no longer valid.

You need to copy the string to the buffer: `std::strncpy(&ex_msg, ex.what(), ex_msg_size)`

umangyadav: Done

tools/api.py
```diff
@@ -1079,7 +1086,10 @@ def generate_virtual_impl(f: Function, fname: str) -> str:
         largs += f.returns.virtual_output_args()
         output = f.returns.virtual_output()
     largs += [arg for p in f.params for arg in p.virtual_arg()]
-    lparams += [p.virtual_param() for p in f.params if not p.this]
+    lparams += [
+        p.virtual_param() for p in f.params
+        if not (p.this or p.name == "exception_msg")
```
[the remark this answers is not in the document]
umangyadav: Added `hidden` flag.

tools/api.py
```diff
@@ -1057,9 +1064,12 @@ def add_cpp_class(self) -> None:
     ${output_decls}
     if (${fname} == nullptr)
         throw std::runtime_error("${name} function is missing.");
+    std::array<char, 256> exception_msg;
     auto api_error_result = ${fname}(${args});
-    if (api_error_result != ${success})
-        throw std::runtime_error("Error in ${name}.");
+    if (api_error_result != ${success}) {
+        const std::string exception_str(exception_msg.data(), 256);
```
pfultz2, line 1070: This will make a string that is always size 256, which may included extra uninitialized or null values. Have you looked at the output from this?

It would be nice to have a test case for this. I think the easiest way would be to have a hidden api mode to disable exception catching in the C api so we can catch the exception and check if the message matches:

```cpp
static thread_local bool disable_exception_catch = false;

// We dont put this file in migraphx.hpp since its only for testing
extern "C" void migraphx_test_private_disable_exception_catch(bool b)
{
    disable_exception_catch = false;
}

namespace migraphx {

template <class F>
migraphx_status try_(F f, bool output = true) // NOLINT
{
    if (disable_exception_catch)
    {
        f();
    }
    else
    {
        try
        {
            f();
        }
        catch(const migraphx::exception& ex)
        {
            if(output)
                std::cerr << "MIGraphX Error: " << ex.what() << std::endl;
            if(ex.error > 0)
                return migraphx_status(ex.error);
            else
                return migraphx_status_unknown_error;
        }
        catch(const std::exception& ex)
        {
            if(output)
                std::cerr << "MIGraphX Error: " << ex.what() << std::endl;
            return migraphx_status_unknown_error;
        }
        catch(...)
        {
            return migraphx_status_unknown_error;
        }
    }
    return migraphx_status_success;
}
...
} // namespace migraphx
```

So in the test you could add the operator with an incorrect shape and use `test::throws` to check for the exception and error message:

```cpp
TEST_CASE(run_sigmoid_custom_op_incorrect_shape)
{
  migraphx::program p;
  ...

  EXPECT(test::throws([&] {
           m.add_instruction(migraphx::operation("sigmoid_custom_op"), {x, alloc});
         }),
         "Expected error message");
}
```

tools/api.py
```diff
@@ -1057,9 +1064,12 @@ def add_cpp_class(self) -> None:
     ${output_decls}
     if (${fname} == nullptr)
         throw std::runtime_error("${name} function is missing.");
+    std::array<char, 256> exception_msg;
```
pfultz2, line 1067: This should initialize `exception_msg.front()` to null so if a message is not written there will be valid empty string.

tools/api.py
umangyadav: >. Have you looked at the output from this? This should just do const std::string exception_str(exception_msg.data()). 

I had looked at the output, It was working allright, but yes, i need to add null character. 

Added method to disable exception and added a test for exception message.

tools/api.py
umangyadav: Done.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -310,13 +347,19 @@ struct interface_base : Base
...
     {
         try
         {
             f();
             return migraphx_status_success;
         }
+        catch(const std::exception& ex)
+        {
+            if(ex_msg)
+                std::strncpy(ex_msg, ex.what(), ex_msg_size);
```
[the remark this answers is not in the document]
umangyadav: Done.

turneram (approved):

PR #1278, opened 2022-06-29
Simplify Algebra of 0+x & x+0  = x

TedThemistokleous:
Added test case and code to simplify zero additions between paremeters and literals during simplifications.  In reference to issue #1236

### DIFF
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -851,6 +851,23 @@ struct find_div_const
     }
 };
 
+struct find_zero_add_const
+{
+    auto matcher() const
+    {
+        return match::name("add")(
+            match::either_arg(0, 1)(match::has_value(0.0f), match::any().bind("x")));
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins  = r.result;
+        auto x_in = r.instructions["x"];
+
+        m.replace_instruction(ins, x_in);
+    }
+};
+
 struct find_sub_const
 {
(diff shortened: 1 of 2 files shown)


src/simplify_algebra.cpp
```diff
@@ -851,6 +851,23 @@ struct find_div_const
...
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins  = r.result;
+        auto x_in = r.instructions["x"];
+
+        m.replace_instruction(ins, x_in);
+    }
+};
```
pfultz2, line 869: I think it would be better to combine x*1, x+0, x/1 into the same matcher, since the replacement is the same:

```cpp
    auto matcher() const
    {
        auto add_0 = match::name("add")(match::either_arg(0, 1)(match::has_value(0.0f), match::any().bind("x")));
        auto mul_1 = match::name("mul")(match::either_arg(0, 1)(match::has_value(1.0f), match::any().bind("x")));
        auto div_1 = match::name("div")(match::args(match::any().bind("x"), match::has_value(1.0f)));
        return match::any_of(add_1, mul_1, div_1);
    }
```

TedThemistokleous: I mean you're right they do resolve to the same thing and we'd just need some more test coverage? By having them as separate steps though doesn't that mean we can put them before and after each other pass with a separation of concerns?

TedThemistokleous:
Closing this in favor for rolling these changes into one single matcher instead.

PR #1281, opened 2022-06-29
Simplify unit algebraic  ops

TedThemistokleous:
Removing the other PRs for mul/div/add and rolling these into a single matcher. Reusing all the tests from the previous PRs

Ref to issue #1236


src/simplify_algebra.cpp
```diff
@@ -851,6 +851,50 @@ struct find_div_const
...
+        m.replace_instruction(ins, c_in);
+    }
+};
+
+struct find_neg_unit_mult_const
+{
+    auto matcher() const
+    {
+        return match::name("mul")(
+            match::either_arg(0, 1)(match::has_value(-1.0f), match::any().bind("x")));
```
umangyadav, line 882: You could then, include `div` here as well. `x / -1`.

umangyadav: probably because of `find_div_const`

TedThemistokleous: Thats what I think too, but oddly the other PR I haven't closed was passed its unit test and this would be run before find_div_const.

TedThemistokleous:
Rebased off develop and added support for x*0, 0*x , 0/x operations under one matcher

src/simplify_algebra.cpp
```diff
@@ -41,6 +41,8 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_PASSES);
```
umangyadav, line 44: Should it be removed ?

umangyadav: It is already defined somewhere else

TedThemistokleous: I have it in incase people want to debug the passes and get the debug info of what item is failing and causing the exception to be thrown

src/simplify_algebra.cpp
```diff
@@ -851,6 +853,92 @@ struct find_div_const
...
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins   = r.result;
+        auto x_ins = r.instructions["x"];
+        auto zero  = m.add_literal(0);
+        auto ret   = m.add_return({zero});
+
+        m.remove_instruction((x_ins));
```
umangyadav, line 937: you don't need to remove the instruction here. If it is not used anywhere later, it would get remove by `deadcodeElimination`.

src/simplify_algebra.cpp
```diff
@@ -851,6 +853,92 @@ struct find_div_const
...
+            match::name("div")(match::args(match::has_value(0.0f), match::any().bind("x")));
+        return match::any_of(mul_zero, div_zero);
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins   = r.result;
+        auto x_ins = r.instructions["x"];
+        auto zero  = m.add_literal(0);
+        auto ret   = m.add_return({zero});
```
umangyadav, line 935: You shouldn't add_return as well.

src/simplify_algebra.cpp
```diff
@@ -851,6 +853,92 @@ struct find_div_const
...
+        auto div_zero =
+            match::name("div")(match::args(match::has_value(0.0f), match::any().bind("x")));
+        return match::any_of(mul_zero, div_zero);
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins   = r.result;
+        auto x_ins = r.instructions["x"];
+        auto zero  = m.add_literal(0);
```
umangyadav, line 934: No need to add a literal as well.

src/simplify_algebra.cpp
```diff
@@ -851,6 +853,92 @@ struct find_div_const
...
+    auto matcher() const
+    {
+        auto mul_zero = match::name("mul")(
+            match::either_arg(0, 1)(match::has_value(0.0f), match::any().bind("x")));
+        auto div_zero =
+            match::name("div")(match::args(match::has_value(0.0f), match::any().bind("x")));
+        return match::any_of(mul_zero, div_zero);
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
```
umangyadav, line 930: You should just bind `zero` value in the matcher and replace `ins` with that `zero_ins` in the matcher.

src/simplify_algebra.cpp
umangyadav: We already have environment var by that name in `pass_manager`

src/simplify_algebra.cpp
```diff
@@ -851,6 +853,92 @@ struct find_div_const
     }
 };
 
+struct find_zero_div_const
+{
+    auto matcher() const { return match::name("div")(match::arg(1)(match::has_value(0.0f))); }
+
+    void apply [[noreturn]] (const module& m, const match::matcher_result& r) const
```
umangyadav, line 860: why does it require `[[noreturn]]` flag when return type is `void` ?

src/simplify_algebra.cpp
```diff
@@ -851,6 +853,92 @@ struct find_div_const
...
+    {
+        if(enabled(MIGRAPHX_TRACE_PASSES{}))
```
umangyadav, line 862: Can you check if `MIGRAPHX_TRACE_MATCHES` dumps the same information? I think it is, in that case we don't need another ENV_VAR

src/simplify_algebra.cpp
TedThemistokleous: what should I replace this with then if things resolve to zero? Just leave the add_literal?

src/simplify_algebra.cpp
TedThemistokleous: I know we do, but I can't get this to compile without adding that to the top of the file for the divide by zero error case

src/simplify_algebra.cpp
TedThemistokleous: Ah that makes sense. Ill try that.

TedThemistokleous: Still doesn't find it when I compile in here. I though we would have some param to gate off of for matchers no?

src/simplify_algebra.cpp
TedThemistokleous: Removed

src/simplify_algebra.cpp
TedThemistokleous: Done, removed.

TedThemistokleous:
Fixed the issues with the help of @umangyadav and got tests rolling. Note to self doing remove instructions upon replacement. Dead code elimination is your friend in our case.

src/simplify_algebra.cpp
TedThemistokleous: I think I need to remove the const here, not sure how that slipped in here. I'm also going to flip these params to [[maybe_unused]] instead of void incase that's causing issues since that appears to be more standard to the standard

https://stackoverflow.com/a/44106902

TedThemistokleous:
Not sure how to reproduce this on my machine right now. Seems like I'm consistently failing on test_layernorm_triad_large. Locally when running the test and I'm not getting the error seen on the unit tests. Cleared my ccache, rebuilt, and still getting the same result.

TedThemistokleous:
Seems like my add_0 matcher is doing something strange, its finding a match with something that is being prematurely optimized out in that test case. Running my branch without the add_0 matcher case, results in everything being fine and no code elimination. I've taken a few debug outputs with MIGRAPHX_TRACE_PASSES and MIGRAPHX_TRACE_MATCHES and seeing three lines missing when we start simplify_algebra altogether.

It appears we're removing the literal:
main:@1 = @literal{1e-12} -> float_type, {1}, {0}

TedThemistokleous:
[debug_add0_built_in.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9029460/debug_add0_built_in.txt)
[debug_no_add0_matcher.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9029461/debug_no_add0_matcher.txt)

test/simplify_algebra_test.cpp
```diff
@@ -734,6 +761,157 @@ TEST_CASE(simplify_div_const)
...
 }
 
+TEST_CASE(simplify_unit_mult_const)
+{
+    migraphx::module m1;
+    {
+        auto x    = m1.add_parameter("x", {migraphx::shape::int32_type, {1}});
+        auto unit = m1.add_literal(1);
+        m1.add_instruction(migraphx::make_op("mul"), x, unit);
+    }
```
pfultz2, line 771: It would be good to have tests using multidimensional tensors(which is much more common) rather than just scalars.

test/simplify_algebra_test.cpp
```diff
@@ -734,6 +761,157 @@ TEST_CASE(simplify_div_const)
...
+
+    migraphx::module m3;
+    {
+        auto unit = m3.add_literal(1);
+        auto x    = m3.add_parameter("x", {migraphx::shape::int32_type, {1}});
+        m3.add_instruction(migraphx::make_op("mul"), unit, x);
+    }
+    run_pass(m3);
+
+    EXPECT((m1 == m3) && (m1 == m2));
```
pfultz2, line 788: I would probably just split and repeat the tests.

src/simplify_algebra.cpp
```diff
@@ -851,6 +851,84 @@ struct find_div_const
...
+};
+
+struct find_unit_ops
+{
+    auto matcher() const
+    {
+        auto mul_1 = match::name("mul")(
+            match::either_arg(0, 1)(match::has_value(1.0f), match::any().bind("x")));
+        auto div_1 =
+            match::name("div")(match::args(match::any().bind("x"), match::has_value(1.0f)));
```
pfultz2, line 872: When you check `match::has_value` you need to use `skip_broadcasts` otherwise this only will match scalars: `match::skip_broadcasts(match::has_value(1.0f))`.

src/simplify_algebra.cpp
```diff
@@ -851,6 +851,84 @@ struct find_div_const
     }
 };
 
+struct find_zero_div_const
+{
+    auto matcher() const { return match::name("div")(match::arg(1)(match::has_value(0.0f))); }
+
+    [[noreturn]] void apply([[maybe_unused]] module& m,
+                            [[maybe_unused]] const match::matcher_result& r) const
```
pfultz2, line 859: For unused parameters you can just remove the parameter name: `void apply(module&, const match::matcher_result&)` instead of using `[[maybe_unused]]`.

src/simplify_algebra.cpp
```diff
@@ -851,6 +851,84 @@ struct find_div_const
...
+    {
+        MIGRAPHX_THROW("ERROR: Matched division by zero in pass");
```
pfultz2, line 861: I dont think we should throw an exception here. This could be code that would be eliminated by DCE so it doesn't make sense to warn. Also, simplify_algebra is for doing transformation.

I think we need to put such validations in a seperate pass or function somewhere, but we may not want to just throw an error. If this div by zero came from the user model then we would point to where the error came from in the IR.

pfultz2:
> It appears we're removing the literal:
> main:@1 = @literal{1e-12} -> float_type, {1}, {0}

The `has_value` matcher has a tolerance of `1e-6` by default (see [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/matcher.hpp#L789)) and this number is `1e-12` so it gets treated as equal to `0`.

The reason there is this really small number is to avoid division by zero when the mean and variance is also zero.  A short-term fix would be to lower the tolerance for the `add_0` case.

However, if this small number is not used to avoid division by zero then having a higher tolerance would be better. So it would be better to detect if the add with a smaller number is later used in a `div`, `recip`, or `rsqrt` by traversing up the graph, but we can do that for a later PR so I wouldn't worry about that for now.

src/simplify_algebra.cpp
TedThemistokleous: On it. I didn't even realize that. I

Unrelated but general question: Would there be a case where we don't want to cover that case, or as a general rule of them, we should be handling broadcasting always for our matchers?

src/simplify_algebra.cpp
TedThemistokleous: Okay, I wasn't sure which was better/our desired standard but I can change this. Shouldn't be an issue as long as we have an understanding of what's clear to the reader.

src/simplify_algebra.cpp
TedThemistokleous: I see, thats where my apprehension was with this case since this should be flagged as an error if say somehow our math/simplifications lead to a division by 0. 

Should I look at part of DCE to flag this instead then and remove this for now?

test/simplify_algebra_test.cpp
TedThemistokleous: Okay I can split these up.

test/simplify_algebra_test.cpp
TedThemistokleous: Makes sense, I suppose both would be valid then to verify that we cover all our bases then.

pfultz2:
> Wow, that makes sense then. I suppose I'll adjust the threshold as part of this changeset then and come up with the 2nd detection PR after?

Yea thats probably the best.

src/simplify_algebra.cpp
pfultz2: > or as a general rule of them, we should be handling broadcasting always for our matchers?

Well not all matchers. At least, for `has_value` matcher, most times we would want to skip broadcasts, but probably not always.

We could rename the current `has_value` matcher as `has_literal_value` and then make `has_value` do `skip_broadcasts_converts(has_literal_value(...))`. This would probably need some clean up work to remove the redundant `skip_broadcasts`.

TedThemistokleous: Yeah it would but seems doable/manageable and would simplify a lot of things. I think we would need to make sure we don't break other things though when we add that no?

For now I'm goin to add the skip_broadcasts statement since it isn't implicit in my has_value calls.

TedThemistokleous:
Applied a few of the changes. Still need to add the additional test cases for broadcasting. Want to make sure tolerance changes don't break anything else before proceeding forward.

TedThemistokleous:
Tried upping the threshold to be 10^-13 in latest fix.

src/simplify_algebra.cpp
pfultz2: We dont really have a place for this kind of checking but it probably needs to be in a seperate function we call. I think it would be best to put this in another PR.

src/include/migraphx/matcher.hpp
```diff
@@ -784,7 +784,7 @@ auto skip_broadcasts_converts(Ms... ms)
 }
 
 template <class T>
-inline auto has_value(T x, float tolerance = 1e-6)
+inline auto has_value(T x, float tolerance = 1e-12)
```
pfultz2, line 787: I dont think we should change the default parameter, only for the specific `add_0` matcher, you can pass a different tolerance `has_value(0.0f, 1e-12)`.

TedThemistokleous: Oh didnt realize that. Thanks I can fix this

TedThemistokleous:
removed global tolerance for floats, modified matcher for add_0 along with removed redudant skip_broadcast calls

TedThemistokleous:
TODO: Run bert-mrpc-tf on devbox with MI200 and see what happens

test/simplify_algebra_test.cpp
```diff
@@ -804,6 +1166,147 @@ TEST_CASE(simplify_sub_const)
...
+        m2.add_return({zero});
+    }
+
+    EXPECT(m1 == m2);
+}
+
+TEST_CASE(simplify_zero_div_const_vec)
+{
+    migraphx::shape inner{migraphx::shape::int32_type, {2}};
+    migraphx::shape outer{migraphx::shape::int32_type, {1, 2, 3, 3}};
```
umangyadav, line 1287: instead of `inner` and `outer` can you use `x_shape` and `zero_shape`? So that it is easier to follow when reading.

src/simplify_algebra.cpp
umangyadav: @TedThemistokleous Can you add `skip_broadcasts` ?

test/simplify_algebra_test.cpp
TedThemistokleous: Yup I can do that.

test/simplify_algebra_test.cpp
```diff
@@ -804,6 +1166,147 @@ TEST_CASE(simplify_sub_const)
...
+    migraphx::op::broadcast b{1, {1, 2, 3, 3}};
```
umangyadav, line 1288: use `add_instruction(make_op("broadcast"))`  syntax.

src/simplify_algebra.cpp
TedThemistokleous: Done I believe that's what I'm doing here

test/simplify_algebra_test.cpp
TedThemistokleous: Done

test/simplify_algebra_test.cpp
```diff
@@ -804,6 +1166,147 @@ TEST_CASE(simplify_sub_const)
...
+        m2.add_parameter("x", outer);
+        auto zerob = m2.add_instruction(b, zero);
+        m2.add_return({zerob});
+    }
+
+    EXPECT(m1 == m2);
+}
+
+TEST_CASE(simplify_zero_mul_const_vec2)
+{
```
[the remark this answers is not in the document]
TedThemistokleous: I think it makes sense to keep these separate actually. I know its one change between div/mul but in the future if something changes we won't need to break apart the template should something be added.

pfultz2 (approved):

PR #1286, opened 2022-07-01
Expose apis for load_buffer() and save_buffer()

aayushmaanjain:
Migraphx has cpp and Python APIs for saving and loading serialized migraphx models after compilation - save() and load(), respectively. These apis save and load the serialized model from a file and internally call functions to save and load it from a (string) buffer. In this PR, I have added those internal functions to the cpp and Python API. This is useful in cases where a user may need to append metadata to the migraphx model before saving it to disk.


src/api/migraphx.py
```diff
@@ -273,12 +273,25 @@ def operation(h):
                  fname='migraphx::load',
                  returns='migraphx::program')
 
+api.add_function('migraphx_load_buffer',
+                 api.params(buffer='const char*',
```
pfultz2, line 277: I assume `buffer='const std::vector<char>&` doesnt work?

aayushmaanjain: No, if I use `buffer='const std::vector<char>&` , it gives an error while building.
```
+        *out = allocate<migraphx_program_t>(migraphx::load_buffer(
+            (const std::vector<char>&(buffer, buffer + buffer_size)), (options->object)));
```
There might be an error with the code generation since the '&' shouldn't be here in my opinion. 

If instead I were to use `buffer='const std::vector<char>` , it changes the  buffer argument to `char *` instead of `const char *` , which causes type conversion issues when trying to pass the buffer using `vector<char> .data()` method
[vector-diff.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9041496/vector-diff.txt)

causten:
Please work on the CI failures.

src/api/migraphx.py
umangyadav: added fix.

umangyadav:
Updated PR with some changes and fixes.

umangyadav:
@aayushmaanjain Exposing load_buffer and save_buffer is a bit of more work  in API than expected. Do you still require this functionality exposed in API or have you found some alternate way?

umangyadav:
@pfultz2 @kahmed10 @aayushmaanjain This  PR is ready for review.

aayushmaanjain:
Hi @umangyadav, unfortunately I no longer have access to ROCm resources. You should contact Shomy Sanyal or Abhinav Vishnu at AMD to follow up regarding the priority/necessity of this. 

For context, I was using this branch to integrate MIGraphX into minigo. I was using the python API to save the model buffer and reading it using the C++ API. Both of these were working fine so I was able to integrate it. I could not find a workaround and was, thus, using my unmerged fork. I do not know if the condition has changed since. 
I found a bug during testing in saving the buffer in cpp iirc that prevented me from completing the PR. 

(cc:) @kahmed10 hopefully has more context based on our interactions.

PR #1288, opened 2022-07-06
Dynamic ref NMS

CharlieL7:
* Has NMS op output a dynamic shape (ONNX spec behavior)
* Allows for dynamic input shape to NMS op
* Depends on #1224 
* ~~Need to decide how to preserve previous zero output padding behavior~~ (done with member variable)
* ~~Unnecessary `compute_shape()` called at eval time for NMS as output shape will stay dynamic until after the calculation is complete~~ (refactored)


src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -45,39 +45,88 @@ namespace op {
...
+                }
+            }
+            else
+            {
+                fixed_shape_error_check();
+            }
+            std::vector<shape::dynamic_dimension> out_lens = {};
+            out_lens.push_back({0, max_num_boxes, 0});
+            out_lens.push_back({3, 3, 0});
+            return {shape::int64_type, out_lens};
```
umangyadav, line 122: I think we need to add tests in the `op_shape_test` for  dynamic cases.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -45,39 +45,88 @@ namespace op {
...
     }
 
     std::string name() const { return "nonmaxsuppression"; }
 
     shape compute_shape(std::vector<shape> inputs) const
     {
         // requires at least 2 inputs
-        check_shapes{{inputs.at(0), inputs.at(1)}, *this}.only_dims(3);
-        auto lens = inputs.front().lens();
+        check_shapes{{inputs.at(0), inputs.at(1)}, *this, true}.only_dims(3).same_ndims();
```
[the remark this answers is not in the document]
CharlieL7: `use_dynamic` refers to the output of NMS being a dynamic shape. Looking at it again, there's a possibility that someone could pass a dynamic shape input without `use_dynamic` set to true. I'm not sure what exactly we would want to occur in that situation. However, with how `check_shapes` currently works it would state that NMS does not support dynamic shapes if we do the check without the `true` parameter.

test/ref_ops_test.cpp
```diff
@@ -3120,6 +3403,40 @@ TEST_CASE(neg_test)
...
+    auto* mm = p.get_main_module();
+    migraphx::shape boxes_s{migraphx::shape::float_type, {1, 6, 4}};
+    std::vector<float> boxes_vec = {0.5, 0.5,  1.0, 1.0, 0.5, 0.6,  1.0, 1.0, 0.5, 0.4,   1.0, 1.0,
+                                    0.5, 10.5, 1.0, 1.0, 0.5, 10.6, 1.0, 1.0, 0.5, 100.5, 1.0, 1.0};
+
+    migraphx::shape scores_s{migraphx::shape::float_type, {1, 1, 6}};
+    std::vector<float> scores_vec = {0.9, 0.75, 0.6, 0.95, 0.5, 0.3};
+
+    auto boxes_l         = mm->add_literal(migraphx::literal(boxes_s, boxes_vec));
+    auto scores_l        = mm->add_literal(migraphx::literal(scores_s, scores_vec));
```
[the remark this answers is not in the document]
CharlieL7: Yes, we can. This test is for the case of dynamic output even with fixed shape input. I have made additional tests with what you suggested, will push up later.

CharlieL7:
I found a bug that I erroneously carried over from Shucai's code. The maximum output shape needs to change with the batch size. The `max_num_boxes` should actually be `number of batches * total number of boxes`. Made the fix in a recent commit. I'm somewhat surprised this operator has been working thus far?

umangyadav:
> I found a bug that I erroneously carried over from Shucai's code. The maximum output shape needs to change with the batch size. The `max_num_boxes` should actually be `number of batches * total number of boxes`. Made the fix in a recent commit. I'm somewhat surprised this operator has been working thus far?

Good fine. It probably worked because `number_of_batch` was only 1.

CharlieL7:
Noticed an issue: there's not going to be a `use_dynamic` attribute for NMS nodes in ONNX models. We need to be able to set `use_dynamic` through some other method. I think we would want to have another value in `struct onnx_options` of basically `use_dynamic_output` and to make a new `parse_nms.cpp` file.

CharlieL7:
I made a rather roundabout way to set `use_dyn_output` using the `onnx_options`. It sets or creates the attribute for any node with the same `op_type`. The `name` field for a node can be absent in the ONNX spec, so I could not use that.

In the future, I expect that we will change the code such that dynamic output operators always use the new dynamic output behavior. That would make the `use_dyn_output` stuff unnecessary.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -45,39 +45,90 @@ namespace op {
...
+        auto boxes_max_lens = inputs.at(0).max_lens();
+        // num batches * num boxes
+        const auto max_num_boxes = boxes_max_lens.at(0) * boxes_max_lens.at(1);
 
-        // check input shape
-        if(lens[1] != inputs.at(1).lens()[2])
+        auto fixed_shape_error_check = [&]() {
+            auto lens = inputs.front().lens();
+            if(lens[1] != inputs.at(1).lens()[2])
```
umangyadav, line 69: This should  additionally test that inputs are not dynamic, in case `use_dyn_output` is set to false mistakenly for dynamic inputs.

test/onnx/gen_onnx.py
```diff
@@ -3563,6 +3563,83 @@ def nms_test():
...
+
+    node = onnx.helper.make_node('NonMaxSuppression',
+                                 inputs=[
+                                     'boxes', 'scores',
+                                     'max_output_boxes_per_class',
+                                     'iou_threshold', 'score_threshold'
+                                 ],
+                                 outputs=['selected_indices'],
+                                 center_point_box=1,
+                                 use_dyn_output=1)
```
umangyadav, line 3588: Would this set additional attribute in Onnx::attributeproto for use_dyn_output ?

src/onnx/onnx_parser.cpp
```diff
@@ -334,9 +329,18 @@ void onnx_parser::parse_graph(module* mod, const onnx::GraphProto& graph)
...
         else
         {
+            auto node_attributes = get_attributes(node);
+            if(contains(map_use_dyn_output, node.op_type()))
+            {
+                onnx::AttributeProto ap;
+                ap.set_type(onnx::AttributeProto::INT);
+                ap.set_i(
+                    static_cast<google::protobuf::int64>(map_use_dyn_output.at(node.op_type())));
+                node_attributes["use_dyn_output"] = ap;
```
umangyadav, line 339: Should it check if node already doens't have it ?

test/onnx/gen_onnx.py
CharlieL7: Yes, the other two tests do not have this additional attribute.

umangyadav (approved):

test/onnx/onnx_test.cpp
```diff
@@ -3452,6 +3452,42 @@ TEST_CASE(nms_dynamic_classes_test)
...
+
+    migraphx::shape siou{migraphx::shape::float_type, {1}};
+    auto iou = mm->add_parameter("iou_threshold", siou);
+
+    migraphx::shape sst{migraphx::shape::float_type, {1}};
+    auto st = mm->add_parameter("score_threshold", sst);
+
+    auto ret = mm->add_instruction(
+        migraphx::make_op("nonmaxsuppression",
+                          {{"center_point_box", true}, {"use_dyn_output", true}}),
```
umangyadav, line 3496: this is not overwriting, both of them are using same value for `use_dyn_output`

CharlieL7: Right, need to generate a new ONNX file. Making the change.

turneram (approved):

src/onnx/parse_nonmaxsuppression.cpp
```diff
@@ -0,0 +1,55 @@
...
+                          std::vector<instruction_ref> args) const
+    {
+        if(parser.force_use_dyn_output)
+        {
+            onnx::AttributeProto ap;
+            ap.set_type(onnx::AttributeProto::INT);
+            ap.set_i(static_cast<google::protobuf::int64>(1));
+            info.attributes["use_dyn_output"] = ap;
+        }
+        auto op = parser.load(opd.op_name, info);
```
pfultz2, line 48: Just do:

```cpp
auto op = parser.load(opd.op_name, info);
op.from_value({"use_dyn_output", parser.force_use_dyn_output});
```

Instead of messing with `onnx::AttributeProto` classes.

src/include/migraphx/onnx.hpp
```diff
@@ -53,6 +53,10 @@ struct onnx_options
     bool print_program_on_error = false;
     /// Max iter num for the loop operator
     int64_t max_loop_iterations = 10;
+    /*!
+     * Set the use_dyn_output attribute to true for all pertinent operators
+     */
+    bool force_use_dyn_output = false;
```
pfultz2, line 59: I would just name the flag `use_dyn_output`.

src/include/migraphx/onnx.hpp
```diff
@@ -53,6 +53,10 @@ struct onnx_options
     bool print_program_on_error = false;
     /// Max iter num for the loop operator
     int64_t max_loop_iterations = 10;
+    /*!
+     * Set the use_dyn_output attribute to true for all pertinent operators
+     */
```
pfultz2, line 58: Use `///` like the rest of the comments(The comments for dynamic dimensions should probably be updated as well to match). And I would write something like `/// Use dynamic output for operators when available`, you dont need to mention an attribute as that is an implementation detail(we could use different operators instead of an attribute).

src/include/migraphx/operation.hpp
```diff
@@ -40,6 +40,7 @@
 #include <migraphx/auto_any_cast.hpp>
 #include <migraphx/lifetime.hpp>
 #include <migraphx/config.hpp>
+#include <migraphx/env.hpp>
```
pfultz2, line 43: This header can be removed here.

pfultz2:
We can merge this and then remove the header in a later PR if you want.

PR #1289, opened 2022-07-06
Show kernel time when using gpu-driver

pfultz2:


turneram (approved):

src/targets/gpu/driver/perf.cpp
```diff
@@ -42,22 +42,31 @@ std::vector<argument> generate_arguments(const std::vector<shape>& shapes, unsig
...
-    auto r   = range(n);
-    double t = std::accumulate(
-        r.begin(), r.end(), double{0.0}, [&](auto x, auto) { return x + time<milliseconds>(run); });
-    return t / n;
+    double host_time   = 0.0;
+    double device_time = 0.0;
+    for(auto i : range(n))
+    {
+        (void)i;
+        host_time += time<milliseconds>(run);
```
kahmed10, line 66: can we switch from milliseconds to microseconds? I know `hipEventElapsedTime` returns time in ms, but it's resolution is approximately 1 microsecond. So multiplying the result by 1000. This is more for readability as we could be dealing with smaller kernels or small improvements to a bigger kernel.

pfultz2: The host time is usually pretty large, and it would make the time inconsistent with the perf report.

kahmed10 (approved):

umangyadav:
@pfultz2 @causten can this be merged ?

PR #1296, opened 2022-07-08
Divide by zero check

TedThemistokleous:
Initial changes and idea on how to handle div by zero. 

The idea here is we detect and match div/zero instructions so we preserve the rest of the graph, and then during our final checks (similar to dangling reference checks) we detect any divzero instructions. This would mean we also create a divzero instruction operator, but I'm not sure the correctness of this approach. 

Reused and reverted the simplify algebra changeset's divide by zero changes to start this since we could reuse the unit test to validate that we are inserting a divzero in the correct spots.


src/program.cpp
```diff
@@ -186,13 +186,21 @@ void program::compile(const target& t, compile_options options)
...
         {
             auto index = std::distance(mod->begin(), dangling);
             MIGRAPHX_THROW("Dangling reference in module " + mod->name() + " from instruction " +
                            std::to_string(index));
         }
+
+        auto divide_by_zero = mod->flag_division_by_zero();
+        if(divide_by_zero != mod->end())
+        {
+            MIGRAPHX_THROW("Division by zero reference in module " + mod->name() + "");
```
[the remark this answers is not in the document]
TedThemistokleous: Yup I can do that

src/module.cpp
```diff
@@ -628,6 +637,17 @@ instruction_ref module::find_dangling_reference() const
     return end();
 }
 
+bool is_div_zero(instruction ins) { return ins.name() == "@divzero"; }
+
+instruction_ref module::flag_division_by_zero() const
+{
+    auto divzero = std::count_if(begin(), end(), [](auto x) { return is_div_zero(x); });
```
umangyadav, line 644: This should be using `find_if`

src/include/migraphx/module.hpp
```diff
@@ -182,6 +184,7 @@ struct module
 
     instruction_ref validate() const;
     instruction_ref find_dangling_reference() const;
+    instruction_ref flag_division_by_zero() const;
```
umangyadav, line 187: instead of `flag_division` let's use `find_division_by_zero` for consistency.

src/module.cpp
TedThemistokleous: don't we want to just see if one exists and halt compilation?

src/include/migraphx/module.hpp
TedThemistokleous: done

test/ref_ops_test.cpp
```diff
@@ -1330,6 +1339,56 @@ TEST_CASE(div_test)
...
+    auto* mm  = p.get_main_module();
+    auto zero = mm->add_literal(0.0f);
+    auto x    = mm->add_parameter("x", {migraphx::shape::float_type, {1}});
+    mm->add_instruction(migraphx::make_op("div"), x, zero);
+    run_pass(*mm);
+
+    bool result = false;
+    try
+    {
+        p.compile(migraphx::ref::target{});
```
umangyadav, line 1354: You can use `EXPECT(test_throws())`

test/ref_ops_test.cpp
```diff
@@ -1330,6 +1339,56 @@ TEST_CASE(div_test)
...
+    auto mul  = mm->add_instruction(migraphx::make_op("mul"), two, div0);
+    auto add  = mm->add_instruction(migraphx::make_op("add"), y, mul);
+    mm->add_instruction(migraphx::make_op("sub"), y, add);
+
+    run_pass(*mm);
+
+    bool result = false;
+    try
+    {
+        p.compile(migraphx::ref::target{});
```
umangyadav, line 1382: use `EXPECT(test_throws())`

test/ref_ops_test.cpp
TedThemistokleous: Will add this

test/ref_ops_test.cpp
TedThemistokleous: Tried that and EXPECT_THROW(test_throws()) seems to be breaking builds. What function did you mean?

TedThemistokleous:
Found a fun case and something to think about, not sure if just late and be chasing better coverage on compute_shape()/compute() is leading me to somewhere valid.

```
auto div0  = mm->add_instruction(div, x, 0)    
run_pass(*mm)   // <- this makes div0 replaced with @divzero

div0->do_something_else()
```

Not sure if this is contrived right now in a test case I created or a real issue. Got around it by just specifying the instruction directly using add_divzero and going to see if things don't break here. How do we know we're not creating stale references after every pass or is this something that only seems to come up when we write test cases?

test/ref_ops_test.cpp
umangyadav: `EXPECT(test::throws([&]{p.compile(migraphx::ref::target{});}));`

test/ref_ops_test.cpp
umangyadav: ` EXPECT(test::throws([&]{p.compile(migraphx::ref::target{});}));`

umangyadav: Similarly in other  places.

umangyadav:
> Found a fun case and something to think about, not sure if just late and be chasing better coverage on compute_shape()/compute() is leading me to somewhere valid.
> 
> ```
> auto div0  = mm->add_instruction(div, x, 0)    
> run_pass(*mm)   // <- this makes div0 replaced with @divzero
> 
> div0->do_something_else()
> ```
> 
> Not sure if this is contrived right now in a test case I created or a real issue. Got around it by just specifying the instruction directly using add_divzero and going to see if things don't break here in the latest fixup. How do we know we're not creating stale references after every pass or is this something that only seems to come up when we write test cases?

`div0` is instruction_ref. After `run_pass`, all its  references will get replaced by `divzero`  and  stale references to `div0` would be cleaned up by deadcode. For  the `div0->do_something_else()`,  Instruction_ref are more of abstract to  represent graph. It creates  sort of  edges. I've  not seen  cases where we directly operate on instruction_ref like `div0->do_something` rather we operate on memory buffers like `argument`, `literal` etc. through `visit()` or `visit_all()`.   I don't think we should worry about case like `div0->do_something`.

TedThemistokleous:
> 

Okay that eases my mind on that then.

test/ref_ops_test.cpp
TedThemistokleous: Done

test/ref_ops_test.cpp
TedThemistokleous: Done

src/module.cpp
```diff
@@ -187,6 +187,10 @@ void module::assign(const module& m)
             {
                 copy_ins = add_return(copy_inputs);
             }
+            else if(ins->name() == "@divzero")
```
umangyadav, line 190: Do we need to add the same check in the `insert_generic_instructions` ?

TedThemistokleous: Yeah I think so. Just adding this in and re-running tests.

TedThemistokleous:
Closing and will revisit later

PR #1298, opened 2022-07-11
Fix TF parsing for creating literals and Fix name lookups for input params

umangyadav:
Bug 1: `create_literal` was using `back_inserter` to copy vector with already allocated size, causing double the size of literal. 
Fix 1 :  not use `back_inserter`
Bug 2:  Input param to model can be from operation that has multiple output, in that case name of the input param would contain `:` e.g. `input_1:0` 
Fix 2: Look for `:` and take substring


pfultz2:
Are we able to add unit tests for these issues?

umangyadav:
> Are we able to add unit tests for these issues?

I can try tomorrow, not sure if newer versions of TF would generate `:0` for a node that has only one output edge.

umangyadav:
> > Are we able to add unit tests for these issues?
> 
> I can try tomorrow, not sure if newer versions of TF would generate `:0` for a node that has only one output edge.

I am unable to produce a model with TensorFlow with `:0` edge, I've mailed QA team on how they got the model. Waiting on it.

kahmed10 (approved):

umangyadav:
> Since this is passing CI, I am okay with merging and leaving the test generation as a TODO.

I looked at the code of the UIF model and how they generated `input:0`. They are using TF-1.15.2 which is very old. They have custom TF with Vitis AI Quantization transforms called `decent_q`.  I am not sure if we can generate such a test without pulling many dependencies into MIGraphX. I think it is fine for now to not add a test.

PR #1301, opened 2022-07-11
Add jit layernorm fusion

pfultz2:


src/targets/gpu/compile_gen.cpp
```diff
@@ -75,25 +82,26 @@ std::string vectorize::str() const
...
-        return {result};
-    // TODO: Try to partially preload items
-    std::fill(result.begin(), result.end(), false);
+    std::vector<bool> result(inputs.size());
+    std::vector<std::size_t> preloaded;
+    for(auto i : range(inputs.size()))
+    {
+        if(inputs[i].strides()[axis] == 0)
+            preloaded.push_back(i);
+    }
```
turneram, line 91: If we want to avoid the cppcheck failure, this could be changed to: 
```
auto idxs = range(inputs.size());
std::copy_if(idxs.begin(), idxs.end(), std::back_inserter(preloaded), [&](auto i){ return inputs[i].strides()[axis] == 0; });
```

@kahmed10 What are your thoughts? Is that less readable?

kahmed10: I thought it could be rewritten as:
```
std::copy_if(inputs.begin(), inputs.end(), std::back_inserter(preloaded), [&](auto input){ return input.strides()[axis] == 0; });
```

turneram (approved):

kahmed10 (approved):

PR #1302, opened 2022-07-12
Add onnx mod operator

TedThemistokleous:
Adding Onnx Mod operator and handling both cases when the fmod flag is set to 0 and 1. Solving part of the issue found in #1283


src/include/migraphx/op/mod.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <utility>
+#include <type_traits>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct mod : binary<mod>
+{
+    bool fmod_flag;
```
pfultz2, line 45: This can be removed(for both operators) since it's not used.

TedThemistokleous: Does that mean I can just reuse the binary attribute() call too then when I remove this?

pfultz2: Yes

src/onnx/parse_mod.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+    std::vector<op_desc> operators() const { return {{"Mod"}}; }
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& parser,
+                          onnx_parser::node_info info,
+                          std::vector<instruction_ref> args) const
+    {
+        int fmod_flag = 0;
+
```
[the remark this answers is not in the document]
TedThemistokleous: Oh I like that. I'm going to steal this.

TedThemistokleous:
> Make sure to add an onnx parser test as well. That will require modifying test/onnx/gen_onnx.py, test/onnx/onnx_test.cpp, and test/py/onnx_backend_test.py. You can see an example on how to do this in #1177

Oh awesome. Thanks!

TedThemistokleous:
Reused test values found in the onnx library for the simple mod operator that was brought up in issue #1283 by Paul. Looks like my previous changes broke functionality of the operator when using abs() instead of remainder() for this one. 

Comes down to how this handles negatives. Let me know if you need me to calculate some other values then the repeated example.

Primarily I used the known set repeatedin the larger 3x3x3 shape in test_verify_onnx.

``` Python
x = np.array([-4, 7, 5, 4, -7, 8]).astype(np.int64)
y = np.array([2, -3, 8, -2, 3, 5]).astype(np.int64)
z = np.mod(x, y)  # expected output [ 0, -2,  5,  0,  2,  3]
expect(node, inputs=[x, y], outputs=[z],
       name='test_mod_mixed_sign_int64')
```

and a smaller subset when testing the ref operator

``` Python
x = [3, 8.5, -7]
np.mod(x, 3)
# array([0. , 2.5, 2. ]) 
```

Found from here: https://www.skytowner.com/explore/difference_between_the_methods_mod_and_fmod_in_numpy#:~:text=Both%20mod(~)%20and%20fmod,capable%20of%20parsing%20floating%20numbers."

Which should be the expected behavior here and should now work for both set of test cases.

TedThemistokleous:
Converting to draft. Need to sort out CI issues and GPU related implementation that's breaking our builds

TedThemistokleous:
@kahmed10 even after pulling out the gpu related test_mod.cpp and test_fmod.cpp builds were breaking.

Pulled out lines from onnx_backend_test.py for  test_mod and test_fmod which seems to have fixed builds, I was still getting issues along the lines of "can't find migraphx::mod", is there somewhere else we should be defining our migraphx operators outside the CMakefile? Is there somewhere else I should be registering these?

kahmed10 (approved):

TedThemistokleous:
Fixed the issue I had with CI regarding the check for inputs. realized adding a second thrown exception in parse_mod was redundant since check_shapes() as part of binary<> handles the number of allowed input args to both mod operators.

test/onnx/gen_onnx.py
```diff
@@ -3231,6 +3232,89 @@ def min_test():
     return ([node], [a, b, c], [y])
 
 
+@onnx_test
+def mod_test():
+    a = helper.make_tensor_value_info('0', TensorProto.FLOAT, [3, 3, 3])
+    b = helper.make_tensor_value_info('1', TensorProto.FLOAT, [3, 3, 3])
+    y = helper.make_tensor_value_info('2', TensorProto.FLOAT, [3, 3, 3])
+
+    node = onnx.helper.make_node('Mod', inputs=['0', '1'], outputs=['2'])
```
umangyadav, line 3241: Shouldn't `fmod` be set to True always for `float` types ?

test/onnx/gen_onnx.py
```diff
@@ -3231,6 +3232,89 @@ def min_test():
...
+    return ([node], [a, b], [y])
+
+
+@onnx_test
+def mod_test_different_dtypes():
+    a = helper.make_tensor_value_info('0', TensorProto.FLOAT, [3, 3, 3])
+    b = helper.make_tensor_value_info('1', TensorProto.INT32, [3, 3, 3])
+    y = helper.make_tensor_value_info('2', TensorProto.FLOAT, [3, 3, 3])
+
+    node = onnx.helper.make_node(
```
umangyadav, line 3263: What is the expected behavior in this case ? `add_common_op` will promote this to be `float` type mod and according to Onnx spec, `fmod` should be set to true for `float` types..

test/onnx/gen_onnx.py
TedThemistokleous: No since it mirrors the functionality of pythons % or mod command.

test/onnx/gen_onnx.py
TedThemistokleous: Yes that's the point of this test. Promote to float and then operator accordingly. Fmod vs Mod works by how they handle negative values.

TedThemistokleous: Similar to above, we do want to type promote this to float and then run mod

test/onnx/gen_onnx.py
umangyadav: https://github.com/onnx/onnx/blob/main/docs/Operators.md#Mod 
"If the input type is floating point, then `fmod` attribute must be set to 1."

test/onnx/verify_onnx.cpp
```diff
@@ -631,6 +631,125 @@ TEST_CASE(mean_integral_test)
...
+TEST_CASE(mod_test)
+{
+    migraphx::program p = migraphx::parse_onnx("mod_test.onnx");
+    p.compile(migraphx::ref::target{});
+
+    migraphx::shape s{migraphx::shape::float_type, {3, 3, 3}};
+
+    std::vector<float> a = {-4.0, 7.0,  5.0, 4.0,  -7.0, 8.0, -4.0, 7.0,  5.0,
+                            4.0,  -7.0, 8.0, -4.0, 7.0,  5.0, 4.0,  -7.0, 8.0,
+                            -4.0, 7.0,  5.0, 4.0,  -7.0, 8.0, -4.0, 7.0,  5.0};
```
umangyadav, line 643: Can we use some values with fractions part nonzero ?

TedThemistokleous: done

test/onnx/onnx_test.cpp
```diff
@@ -2954,6 +2954,86 @@ TEST_CASE(min_test)
...
+    try
+    {
+        auto prog = optimize_onnx("mod_test_half.onnx");
+    }
+    catch(std::exception& e)
+    {
+        (void)e;
+        result = true;
+    }
+    EXPECT(result);
```
umangyadav, line 2982: EXPECT(test::throws([&] { migraphx::parse_onnx("mod_test_half.onnx"); }));

umangyadav, line 2982:
```suggestion
    EXPECT(test::throws([&] { migraphx::parse_onnx("mod_test_half.onnx"); }));
```

test/onnx/gen_onnx.py
```diff
@@ -2993,6 +2993,7 @@ def matmul_bmbm_test():
 
 @onnx_test
 def matmul_bmv_test():
+
```
umangyadav, line 2996: Undo this change.

test/ref_ops_test.cpp
```diff
@@ -3030,6 +3030,80 @@ TEST_CASE(min_test)
...
+    mm->add_instruction(migraphx::make_op("fmod"), curr_mod, l2);
+    p.compile(migraphx::ref::target{});
+    auto result = p.eval({}).back();
+    std::vector<float> results_vector(4);
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> gold{-1, 0, -3};
+    EXPECT(migraphx::verify_range(results_vector, gold));
+}
+
+TEST_CASE(fmod_floatingPoint_test)
```
umangyadav, line 3051:
```suggestion
TEST_CASE(fmod_float_test)
```

test/ref_ops_test.cpp
```diff
@@ -3030,6 +3030,80 @@ TEST_CASE(min_test)
...
+    mm->add_instruction(migraphx::make_op("mod"), curr_mod, l2);
+    p.compile(migraphx::ref::target{});
+    auto result = p.eval({}).back();
+    std::vector<float> results_vector(4);
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> gold{0, 0, 2};
+    EXPECT(migraphx::verify_range(results_vector, gold));
+}
+
+TEST_CASE(mod_floatingPoint_test)
```
umangyadav, line 3088:
```suggestion
TEST_CASE(mod_float_test)
```

src/onnx/onnx_parser.cpp
```diff
@@ -514,6 +514,26 @@ shape::type_t get_type(int dtype)
     }
 }
 
+bool is_type_float(int dtype)
```
umangyadav, line 517: Why we require `int` overload ?

TedThemistokleous: artifact of something I had before. Ill remove it

src/onnx/include/migraphx/onnx/onnx_parser.hpp
```diff
@@ -119,6 +119,8 @@ struct onnx_parser
 };
 
 shape::type_t get_type(int dtype);
+bool is_type_float(int dtype);
```
umangyadav, line 122:
```suggestion

```

TedThemistokleous: removed

test/ref_ops_test.cpp
```diff
@@ -3030,6 +3030,80 @@ TEST_CASE(min_test)
...
+    mm->add_instruction(migraphx::make_op("mod"), curr_mod, l2);
+    p.compile(migraphx::ref::target{});
+    auto result = p.eval({}).back();
+    std::vector<float> results_vector(4);
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> gold{0, 0, 2};
+    EXPECT(migraphx::verify_range(results_vector, gold));
+}
+
+TEST_CASE(mod_float_test)
```
umangyadav, line 3088: need to be careful about this one. With ONNX float should use `fmod`. As a test it is fine.

umangyadav (approved):

PR #1304, opened 2022-07-13
Add fpga target

varunsh-xilinx:
Add a placeholder FPGA target. It currently cannot be evaluated but it does compile. Once this is merged, we can work with other teams to implement the functionality to run on the FPGA defined by the function signatures here


src/targets/fpga/lowering.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    std::string name() const { return "fpga::vitis_ai"; }
+
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        (void)inputs;
+        return xmodel_.get_shape();
+    }
+
+    argument compute(context& ctx, const shape& output_shape, std::vector<argument> args) const
```
kahmed10, line 62: Can you add a const to context here?

varunsh-xilinx: Done

src/targets/fpga/subgraph.cpp
```diff
@@ -0,0 +1,123 @@
...
+            if(first == mod.end())
+            {
+                first = it;
+            }
+            last = it;
+        }
+    }
+
+    // assuming all FPGA instructions are in one contiguous range
+    pm->insert_instructions(pm->end(), first, last, {});
```
umangyadav, line 105: We may require additional checks here i guess, e.g. all the intermediate results between first and last are consumed within `pm` module, otherwise, it may need to return multiple outputs.

src/targets/fpga/subgraph.cpp
```diff
@@ -0,0 +1,123 @@
...
+    {
+        if(migraphx::starts_with(it->name(), "@return"))
+        {
+            placeholder_ins = mod.insert_instruction(
+                it, migraphx::make_op("fpga::vitis_placeholder"), literal_inputs, {pm});
+            break;
+        }
+    }
+
+    mod.replace_return({placeholder_ins});
```
umangyadav, line 118: Why is it replacing return ? I think it should replace the `last` instruction_ref

src/targets/fpga/lowering.cpp
kahmed10: Remember to push changes with updated clang-format.

src/targets/fpga/include/migraphx/fpga/context.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <migraphx/config.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace fpga {
+
+struct context
+{
+    int foo = 0;
```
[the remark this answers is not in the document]
varunsh-xilinx: Done

src/targets/fpga/vitis_ai_adapter.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+migraphx::shape XModel::get_shape() const { return shape_; };
+
+void XModel::set_shape(migraphx::shape shape) { shape_ = shape; }
+
+XModel create_xmodel(migraphx::module_ref mod)
+{
+    std::cout << "Calling an external function: create_xmodel!\n";
+    XModel xmodel;
+    xmodel.set_shape(std::prev(mod->end())->get_shape());
```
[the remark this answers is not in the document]
varunsh-xilinx: Done

src/targets/fpga/include/migraphx/fpga/vitis_ai_adapter.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace vitis_ai {
+
+class XModel
+{
+    public:
+    migraphx::shape get_shape() const;
+    void set_shape(migraphx::shape);
+
+    private:
```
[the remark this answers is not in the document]
varunsh-xilinx: Done

src/targets/fpga/vitis_ai_adapter.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "migraphx/module.hpp"
+
+#include "migraphx/stringutils.hpp"
+namespace vitis_ai {
+
+migraphx::shape XModel::get_shape() const { return shape_; };
+
+void XModel::set_shape(migraphx::shape shape) { shape_ = shape; }
+
+XModel create_xmodel(migraphx::module_ref mod)
```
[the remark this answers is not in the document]
varunsh-xilinx: Done

src/targets/fpga/subgraph.cpp
```diff
@@ -0,0 +1,123 @@
...
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.dummy, "dummy"));
+    }
+
+    std::string name() const { return "fpga::vitis_placeholder"; }
+
+    shape compute_shape(const std::vector<shape>& inputs, std::vector<module_ref> mods) const
+    {
+        (void)inputs;
```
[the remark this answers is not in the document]
varunsh-xilinx: Yes, I kept it named so people know what it is. It will probably be used so the void is just a placeholder

src/targets/fpga/subgraph.cpp
```diff
@@ -0,0 +1,123 @@
...
+            MIGRAPHX_THROW("Only one return");
+        return sm->get_output_shapes().front();
+    }
+};
+MIGRAPHX_REGISTER_OP(fpga_placeholder_op)
+
+bool is_fpga_instr(migraphx::instruction_ref it)
+{
+    // assuming all instructions that aren't @param, @literal, or input data are fpga instrs
+    return (!it->inputs().empty()) || (migraphx::starts_with(it->name(), "@"));
```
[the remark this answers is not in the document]
varunsh-xilinx: I ended up rewriting this logic because it was confusing me too

src/targets/fpga/subgraph.cpp
varunsh-xilinx: True, for now I think that's outside the scope

src/targets/fpga/subgraph.cpp
varunsh-xilinx: I don't know what the right way to do this is. I was just fumbling around with migraphx and this seemed to print the right instructions when I was testing the mnist graph but it's very possible that actually trying to run this won't work

test/verify/run_verify.cpp
```diff
@@ -168,7 +168,7 @@ void run_verify::verify(const std::string& name, const migraphx::program& p) con
     std::vector<std::string> target_names;
     for(const auto& tname : migraphx::get_targets())
     {
-        if(tname == "ref")
+        if(tname == "ref" || tname == "fpga")
```
TedThemistokleous, line 171: Is the plan to remove this exception once we integrate fpga implementations for now? I think this will remove the CI checks for test_compile

varunsh-xilinx: Yes, that is the eventual plan once it's actually functional. The test I added does pass but the verify tests try to run different ops on all targets and those don't work

TedThemistokleous: Similar to something we do for disabled_tests, it looks like you can stub out the other targets based on target name. Do this for your added test case to add it to CI, and that should be sufficient. We can slowly add further tests for the other targets as you build this out.

My concern with stubbing things out for all tests, is that we're not all testing the latest and greatest then should any issues arise with new changes from develop and your fpga changes as you integrate them in.

Add a todo comment as part of your disabled_tests added for stuff covered in "fpga" and then we can open an issue after this PR for the other targets/pieces to keep track of things.

TedThemistokleous (changes requested):

TedThemistokleous:
> Is there anything else that's gating this PR? I think I've addressed the comments

Just another comment and request from me. We can track/tag the other targets from this change set in later issues.

I think the other issue/questions with @umangyadav also need to be resolved.

test/verify/run_verify.cpp
varunsh-xilinx: Are you referring to [this ](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/verify/main.cpp#L70=)for disabled_tests? If so, I understood that as a blacklist to prevent specific tests from running. What I need is to prevent all these verify tests from running because none of these operations will work right now. The test case I added is already part of the CI since it's outside the verify/. At least, I can see it running when I do `make check` locally. Perhaps I'm misunderstanding your request.

I can add a TODO for this here, sure.

varunsh-xilinx: Got it, thanks for the reference. I think adding this to the CI may be a little premature at this point since we'll be working in the coming months with other teams in China to help create it. Once it's in a little more stable state, we can test it regularly. At this point, it's intended more as placeholder code so the teams in China have something to look at and fill in since you can't actually call `eval` on something targeting the FPGA. Do you feel otherwise?

I've updated the TODO.

umangyadav (approved):

test/verify/run_verify.cpp
varunsh-xilinx: My initial plan was to just let developers run any FPGA tests locally since they will not be making frequent PRs to this repo anyway but okay, I've added it to the CI.

varunsh-xilinx:
@TedThemistokleous I'm getting an error with the target not being registered in CI but `make check` runs and passes fine for me locally. How do I fix that?

Do we really need to add the `fpga/` tests to the CI to merge this PR? Ultimately, these tests will require hardware that the CI will not have so they won't run anyway. I don't think the GPU backend is enabled in these CI tests either because the HAS_HIP check fails

TedThemistokleous:
> @TedThemistokleous I'm getting an error with the target not being registered in CI but `make check` runs and passes fine for me locally. How do I fix that?
> 
> Do we really need to add the `fpga/` tests to the CI to merge this PR? Ultimately, these tests will require hardware that the CI will not have so they won't run anyway. I don't think the GPU backend is enabled in these CI tests either because the HAS_HIP check fails

Yes, but we don't have to make a requirement that it'll pass for things to work when the FPGA flag is on. What we should require though is the stock MIGraphx stuff doesn't break with new changes as we add a new target.

I've added another section to the set of CI builds called linux-fpga that will test debug builds for the fpga target with the flag on that should use. I'll ping in the comments and explain things. 

I can help you get the other checks passing since it won't actually let us submit a PR if any checks with the required flag next to them go in.

.github/workflows/ci.yaml
```diff
@@ -270,4 +270,98 @@ jobs:
         curl -s https://codecov.io/bash | bash
         echo "Uploaded"
 
+  linux-fpga:
```
TedThemistokleous, line 273: This is a new job created as part of the MIGraphx runtime, basically a copy of the existing linux job seen above with some modification.

Good resource for this is: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

.github/workflows/ci.yaml
```diff
@@ -270,4 +270,98 @@ jobs:
...
+
+    strategy:
+      matrix:
+        os:
+          - ubuntu-18.04
+          - ubuntu-20.04
+        configuration:
+          - debug
+          #- release  Uncomment when ready to test release builds
+          #- codecov  Uncomment when ready for codecov
```
TedThemistokleous, line 292: Once your team progresses, they can uncomment these sections to get similar coverage for our jobs.

.github/workflows/ci.yaml
```diff
@@ -270,4 +270,98 @@ jobs:
         curl -s https://codecov.io/bash | bash
         echo "Uploaded"
 
+  linux-fpga:
+    continue-on-error: true
```
TedThemistokleous, line 274: This flag is important since it means even if the steps in this job fails, you won't block the build. Remove this when ready to start requiring CI checks for your builds.

.github/workflows/ci.yaml
```diff
@@ -270,4 +270,98 @@ jobs:
...
+    #  env:
+    #    CODECOV_TOKEN: "8545af1c-f90b-4345-92a5-0d075503ca56"
+    #  run: |
+    #    sudo apt-get install -y lcov
+    #    cd build
+    #    lcov --directory . --capture --output-file $(pwd)/coverage.info
+    #    lcov --remove $(pwd)/coverage.info '/usr/*' --output-file $(pwd)/coverage.info
+    #    lcov --list $(pwd)/coverage.info
+    #    curl -s https://codecov.io/bash | bash
+    #    echo "Uploaded"
```
TedThemistokleous, line 367: Uncomment this section when you're ready to add codecov as above

Alternatively if you want additional steps, make sure to use the appropriate flags you wish to test in the above

.github/workflows/ci.yaml
```diff
@@ -270,4 +270,98 @@ jobs:
...
+        echo "leak:dnnl::impl::malloc" > suppressions.txt
+        export LSAN_OPTIONS="suppressions=$(pwd)/suppressions.txt"
+        rbuild build -d cget -s gh -T check \
+          -DCMAKE_BUILD_TYPE=${{matrix.configuration}} \
+          -DMIGRAPHX_ENABLE_PYTHON=${{matrix.configuration == 'release' && 'On' || 'Off'}} \
+          -DCMAKE_CXX_FLAGS_DEBUG="-g1 -Os -fdebug-prefix-map=$PWD=. -fdebug-types-section -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=undefined" \
+          -DCMAKE_CXX_FLAGS_CODECOV="-g1 -Og -fdebug-prefix-map=$PWD=. -fdebug-types-section -fprofile-arcs -ftest-coverage -fno-omit-frame-pointer" \
+          -DCMAKE_EXE_LINKER_FLAGS='-fuse-ld=gold' \
+          -DCMAKE_SHARED_LINKER_FLAGS='-fuse-ld=gold' \
+          -DMIGRAPHX_ENABLE_FPGA=On
```
TedThemistokleous, line 353: The new flag added to turn on FPGA builds.

TedThemistokleous (approved):
Looks like your latest CI job is now running the fpga tests you added. Feel free to modify that pipeline in future PRs if you need additional flags.

Good work

varunsh-xilinx:
Thanks for your help @TedThemistokleous

PR #1306, opened 2022-07-15
Add onnx mod operator gpu cpu

TedThemistokleous:
Next phase of changes to implement operator for CPU and  GPU implementations


src/targets/gpu/device/fmod.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+#include <migraphx/gpu/device/fmod.hpp>
+#include <migraphx/gpu/device/nary.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
+namespace device {
+
+void fmod(hipStream_t stream, const argument& result, const argument& arg1, const argument& arg2)
```
pfultz2, line 32: These device functions are not used and can be removed as this will get generated with the jit pipeline.

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
```diff
@@ -111,6 +111,8 @@ MIGRAPHX_DEVICE_MATH(sinh, ::sinh)
 MIGRAPHX_DEVICE_MATH(sqrt, ::sqrt)
 MIGRAPHX_DEVICE_MATH(tan, ::tan)
 MIGRAPHX_DEVICE_MATH(tanh, ::tanh)
+MIGRAPHX_DEVICE_MATH(mod, ::mod)
+MIGRAPHX_DEVICE_MATH(fmod, ::fmod)
```
pfultz2, line 115: I assume you need to expose `fmod` and `remainder`, but I am not sure if hip provides `remainder` so we might need provide our own implementation.

src/targets/gpu/lowering.cpp
```diff
@@ -146,6 +149,7 @@ struct miopen_apply
         add_generic_op("logical_xor");
         add_generic_op("max");
         add_generic_op("min");
+        add_generic_op("mod");
```
pfultz2, line 152: These can be removed as this will not be used. The fuse_pointwise will replace them before it gets here to lowering.

test/verify/test_mod.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+
+struct test_mod : verify_program<test_mod>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        migraphx::shape s{migraphx::shape::float_type, {3}};
```
pfultz2, line 36: It might be good to use a larger buffer so more random values are used. It also might be good to use neg/abs if there is specific behavior for negative numbers. Also, we may want to consider using our own set of values as well(but I think the onnx tests should test this).

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
TedThemistokleous: they do have a  remainderf 

https://rocmdocs.amd.com/en/latest/ROCm_API_References/HIP-MATH.html?highlight=remainder#remainderf

src/targets/gpu/device/fmod.cpp
TedThemistokleous: Oh okay.

src/targets/gpu/lowering.cpp
TedThemistokleous: OH didn't realize that. Thanks! I can remove

pfultz2:
Thirdly, the lines in [pointwise.cpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/jit/pointwise.cpp#L142) can be removed. This will override the `point_op` which we dont want. The pointwise.cpp does override some of the default behavior but that is only to support vector types.

test/verify/test_mod.cpp
TedThemistokleous: So move this over to onnx test then? I think I did have some tests in verify_onnx for this that I added with larger vectors and negative values as part of mod_test as well as one with differing types

TedThemistokleous:
> Thirdly, the lines in [pointwise.cpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/jit/pointwise.cpp#L142) can be removed. This will override the `point_op` which we dont want. The pointwise.cpp does override some of the default behavior but that is only to support vector types.

For just the mod, or both mod and fmod?  I'd think pointwise would be find for fmod since its just using the fmod call

pfultz2:
> For just the mod, or both mod and fmod? I'd think pointwise would be find for fmod since its just using the fmod call

For both. It already will generate a call to `fmod` so its not necessary here.

src/targets/gpu/device/fmod.cpp
TedThemistokleous: done

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
pfultz2: Well thats good.

src/targets/gpu/lowering.cpp
TedThemistokleous: done.

src/include/migraphx/op/mod.hpp
```diff
@@ -47,9 +47,10 @@ struct mod : binary<mod>
     {
         auto a           = base_attributes();
         a["commutative"] = false;
+        a["point_op"]    = "${function:fmod}((${function:remainder}(${0}, ${1})) + ${1}, ${1})";
         return a;
     }
-    std::string point_function() const { return "mod"; }
+    // std::string point_function() const { return "mod"; }
```
pfultz2, line 53: Yea this function can be removed. Also `name()` can be removed as well since the `binary<>` base class will use the name of the class by default. This could be done for `fmod` as well.

TedThemistokleous: All good. Done

TedThemistokleous:
rebased this off develop as this was originally based on the original onnx mod changes before being split out.

src/targets/cpu/lowering.cpp
```diff
@@ -43,6 +43,8 @@
 #include <migraphx/op/argmax.hpp>
 #include <migraphx/op/argmin.hpp>
 #include <migraphx/op/rnn_var_sl_last_output.hpp>
+#include <migraphx/op/mod.hpp>
+#include <migraphx/op/fmod.hpp>
```
pfultz2, line 47: These includes can be removed.

TedThemistokleous: Done

test/verify/test_mod.cpp
pfultz2: > So move this over to onnx test then? 

I meant the backend onnx tests(which you already enabled those tests).

> I think I did have some tests in verify_onnx for this that I added with larger vectors and negative values as part of mod_test as well as one with differing types

No, I dont mean verify_onnx, we can do this in verify directly.

src/targets/gpu/lowering.cpp
```diff
@@ -138,6 +138,7 @@ struct miopen_apply
         add_generic_op("erf");
         add_generic_op("exp");
         add_generic_op("floor");
+        add_generic_op("fmod");
```
pfultz2, line 141: This line can be removed.

src/targets/cpu/lowering.cpp
pfultz2, line 47: Are there commits to be pushed yet?

TedThemistokleous: No these should already be in the build.

kahmed10 (approved):

TedThemistokleous:
@pfultz2 this should be good to go, is there anything more to be done for this?

pfultz2:
> is there anything more to be done for this?

Can you make the tensor size larger for the verify tests? Maybe using something like 64 or something instead of 3.

TedThemistokleous:
> > is there anything more to be done for this?
> 
> Can you make the tensor size larger for the verify tests? Maybe using something like 64 or something instead of 3.

Sure. Done

pfultz2 (approved):

PR #1309, opened 2022-07-19
Add output_alias and runs_on_offload_target flags for the custom ops

umangyadav:
Depends on #1245. 

This PR adds two methods for the `custom_ops` virtual class. 
1. `bool runs_on_offload_target()`, if the custom op runs directly on the `gpu` then it should be set to `true`.  in this case, custom op expects its parameters to reside in GPU memory and writes output to the GPU memory.  If it is set to `false` then, custom op expects it's parameter to reside on the host and puts back the result into the host memory. 

2. `output_alias`, if output of the custom op is aliasing the input buffer. i.e. interpreting the same input buffer with differnet shape and strides. 

3. Update `as_vector()` in C++ API to handle non-standard shapes. It required exposing element_index to space_index conversion method for the shape class.


src/api/api.cpp
```diff
@@ -655,6 +662,47 @@ struct migraphx_experimental_custom_op
...
+    migraphx_experimental_custom_op_runs_on_offload_target runs_on_offload_target_f = nullptr;
+    bool runs_on_offload_target() const
+    {
+        std::remove_pointer_t<bool*> out;
+        if(runs_on_offload_target_f == nullptr)
+            throw std::runtime_error("runs_on_offload_target function is missing.");
+        std::array<char, 256> exception_msg;
+        exception_msg.front() = '\0';
+        auto api_error_result = runs_on_offload_target_f(
+            &out, object_ptr.data, exception_msg.data(), exception_msg.size());
```
TedThemistokleous, line 697: Small question, is the intent to use remove_pointer<bool *> since we want to reference the local context out as part of the offloaded call's result in runs_on_offload_target_f( &out, ....) and send it back out on the return?

src/api/api.cpp
```diff
@@ -655,6 +662,47 @@ struct migraphx_experimental_custom_op
...
+            throw std::runtime_error("Error in output_alias of: " +
+                                     std::string(object_ptr.obj_typename) + ": " + exception_str);
+        }
+        return std::ptrdiff_t(out);
+    }
+
+    migraphx_experimental_custom_op_runs_on_offload_target runs_on_offload_target_f = nullptr;
+    bool runs_on_offload_target() const
+    {
+        std::remove_pointer_t<bool*> out;
```
TedThemistokleous, line 691: Maybe create this var after the check if runs_on_offload_target_f is successful/doesn't throw.

src/api/api.cpp
```diff
@@ -655,6 +662,47 @@ struct migraphx_experimental_custom_op
         }
         return (&out)->object;
     }
+
+    migraphx_experimental_custom_op_output_alias output_alias_f = nullptr;
+    std::ptrdiff_t output_alias(std::vector<migraphx::shape> inputs) const
+    {
+        std::remove_pointer_t<int64_t*> out;
```
TedThemistokleous, line 669: Move creation of this until after the throw for output_alias_f being valid.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -1255,7 +1255,9 @@ struct experimental_custom_op_base
     virtual std::string name() const                                            = 0;
     virtual argument compute(context ctx, shape output, arguments inputs) const = 0;
     virtual shape compute_shape(shapes inputs) const                            = 0;
-    virtual ~experimental_custom_op_base()                                      = default;
+    virtual std::ptrdiff_t output_alias(shapes) const { return -1; }
+    virtual bool runs_on_offload_target() const { return true; }
```
TedThemistokleous, line 1259: So we're always assuming anything not setup will always run offloaded, and not have any alias?

src/api/migraphx.py
```diff
@@ -450,4 +450,9 @@ def experimental_custom_op(h):
     h.virtual('compute_shape',
               api.params(inputs='std::vector<migraphx::shape>'),
               returns='migraphx::shape')
+    # TODO(umang) : If output aliasing then, runs_on_offload_flag must be false
```
TedThemistokleous, line 453: Is this still a TODO or has this been added?

src/include/migraphx/operation.hpp
```diff
@@ -56,6 +56,8 @@ struct operation
     std::string name() const;
     /// An optional method that can be used to finalize the operator before running
     void finalize(context& ctx);
+    /// An optional method that can be used to set offload copy for a target
+    bool runs_on_offload_target();
```
TedThemistokleous, line 60: Make this either take bool arg to set the state of the offload target run, and rename this. The current function name reads like a getter (like "does this run on offload target") rather than setting the run state on the offloaded target.

src/include/migraphx/operation.hpp
```diff
@@ -1097,6 +1173,12 @@ struct operation
             return private_detail_te_default_has_finalize(char(0), private_detail_te_value);
         }
 
+        bool has_offload_copy() const override
+        {
+
```
TedThemistokleous, line 1178: Remove whitespace here

src/include/migraphx/operation.hpp
```diff
@@ -1123,6 +1205,13 @@ struct operation
                 char(0), private_detail_te_value, ctx, output, input);
         }
 
+        bool runs_on_offload_target() const override
+        {
+
```
TedThemistokleous, line 1210: remove whitespace

TedThemistokleous (changes requested):
change runs_on_offload_target naming since its not clear this sets the run state for the op

just answer the other comments/questions

src/api/include/migraphx/migraphx.hpp
umangyadav: Yes is defined as the default behavior.

src/api/migraphx.py
umangyadav: It is still a TODO

src/include/migraphx/operation.hpp
umangyadav: It is supposed to be a getter. and not a setter.

TedThemistokleous: remove the comment above then, it doesnt make sense to what that functions intent is

src/include/migraphx/operation.hpp
umangyadav: This code has been auto-generated. It has extra space because  common logic for the auto-gen was supposed to put namespace identifications e.g. `using ` clause there but in this method, it doesn't need any namespace identification and therefore emtpy. I can put a check for empty `using` clause somehow but would like to keep it for some other PR.

src/include/migraphx/operation.hpp
umangyadav: Same as before, it is auto-gen.

TedThemistokleous: all good as long as that's known that makes sense.

src/api/api.cpp
umangyadav: Done.

src/api/api.cpp
umangyadav: It is calling into C API. We pass everything by pointers so that if is compiled with a different C/C++ runtime library then,  it would have ABI compatibility.  This code is also auto-generated  `remove_pointer<bool*>` here is just an artifact of common logic for auto-gen code logic.

src/api/include/migraphx/migraphx.hpp
umangyadav: I think it is better if i keep `runs_on_offload_target` as a pure virtual method. made changes.

TedThemistokleous: I agree with that. I think its safe to assume that any custom op derived would need to have that defined otherwise throw an error at compile time.

src/api/api.cpp
TedThemistokleous: Okay makes sense.

TedThemistokleous (approved):

umangyadav:
@kahmed10 I've updated the test for the better. PR is ready for review.

TedThemistokleous (approved):

src/api/migraphx.py
```diff
@@ -450,4 +450,9 @@ def experimental_custom_op(h):
     h.virtual('compute_shape',
               api.params(inputs='std::vector<migraphx::shape>'),
               returns='migraphx::shape')
+    # TODO(umang) : If output aliasing then, runs_on_offload_flag must be true
```
kahmed10, line 453:
```suggestion
    # TODO : If output aliasing then, runs_on_offload_flag must be true
```

umangyadav: Done.

umangyadav:
updated logic for `as_vector()` in C++ API to work with non-standard shape.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -1255,7 +1273,9 @@ struct experimental_custom_op_base
     virtual std::string name() const                                            = 0;
     virtual argument compute(context ctx, shape output, arguments inputs) const = 0;
     virtual shape compute_shape(shapes inputs) const                            = 0;
-    virtual ~experimental_custom_op_base()                                      = default;
+    virtual std::ptrdiff_t output_alias(shapes) const { return -1; }
```
pfultz2, line 1276: `output_alias` should return a vector. We currently dont do this for our internal operators but we need to in order to support multiple aliases from tuples. Since we cant easily change APIs in the future, we should aim for the correct signature from the beginning.

src/api/migraphx.py
```diff
@@ -450,4 +451,9 @@ def experimental_custom_op(h):
     h.virtual('compute_shape',
               api.params(inputs='std::vector<migraphx::shape>'),
               returns='migraphx::shape')
+    # TODO : If output aliasing then, runs_on_offload_flag must be true
```
pfultz2, line 454: This is not true. You could have output aliasing and run on the host. A CPU custom op would run on the host and take CPU output buffers.

tools/include/operation.hpp
```diff
@@ -56,6 +56,9 @@ struct operation
     std::string name() const;
     /// An optional method that can be used to finalize the operator before running
     void finalize(context& ctx);
+    /// An optional method that indicates whether operation runs on the offload target device or on
+    /// the host
+    bool runs_on_offload_target();
```
pfultz2, line 61: I think it would be better to have attributes for this instead of a method.

src/api/migraphx.py
umangyadav: If compile target is GPU  and
If the output is aliasing and runs_on_offload_target is false, then MIGraphX would inject buffer copies to and from GPU memory. in that case, it is not clear if  custom_op would alias CPU buffer or GPU buffer.

pfultz2: > in that case, it is not clear if custom_op would alias CPU buffer or GPU buffer.

It would alias the CPU buffer since that is the input. Right now, we naively copy back and forth for CPU operators but we could in the future only copy once when there is a series of CPU operators, so the aliased output buffer would become input to another CPU operator.

umangyadav: Agree that buffer copy logic needs to be smarter. We would need that for FPGA partitioning also. 


> It would alias the CPU buffer since that is the input. 
Okay, but I guess wouldn't make sense if the consumer op of it is running on GPU. 

Do you want me to remove TODO?

umangyadav: Removed TODO

src/api/include/migraphx/migraphx.hpp
umangyadav: Added opaque object for `vector<ptrdiff>` similar to shapes, arguments, and instructions. 

I tried inferring the return type of virtual method but that didn't seem possible easily, without having type information of the input arguments for that method.  Idea was to switch between different methods for setting function pointers if the return type of vector.

src/api/api.cpp
```diff
@@ -284,6 +292,14 @@ struct custom_operation
     {
         return op.compute(std::move(ctx), std::move(output_shape), std::move(inputs));
     }
+
+    std::ptrdiff_t output_alias(std::vector<shape> inputs) const
+    {
+        // TODO: For now, only support one output alias
+        return op.output_alias(std::move(inputs)).front();
```
pfultz2, line 299: We should check the size of the vector. If its empty we should return `-1`, and if its greater then `1` we should throw.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -1255,7 +1299,9 @@ struct experimental_custom_op_base
     virtual std::string name() const                                            = 0;
     virtual argument compute(context ctx, shape output, arguments inputs) const = 0;
     virtual shape compute_shape(shapes inputs) const                            = 0;
-    virtual ~experimental_custom_op_base()                                      = default;
+    virtual ptrdiffs output_alias(shapes) const { return {-1}; }
```
pfultz2, line 1302: The default should be empty. Empty means it doesn't alias anything. We should also make this be `std::vector<std::size_t>` instead of using signed integers.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -1267,8 +1351,10 @@ struct experimental_custom_op : interface_base<MIGRAPHX_HANDLE_BASE(experimental
                              obj,
                              get_type_name(obj).c_str(),
                              obj.name().c_str());
-        MIGRAPHX_INTERFACE_LIFT(T, experimental_custom_op, compute_shape);
-        MIGRAPHX_INTERFACE_LIFT(T, experimental_custom_op, compute);
+        MIGRAPHX_INTERFACE_LIFT(T, experimental_custom_op, compute_shape, out_params<1>{});
```
pfultz2, line 1354: The out params should be passed as an integer: `MIGRAPHX_INTERFACE_LIFT(T, experimental_custom_op, compute_shape, 1)` then in the macro you can make it `out_params<1>{}`.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -469,20 +525,29 @@ struct interface_base : Base
...
     template <class T, class U>
     auto auto_assign(rank<1>, T* out, U x) -> decltype(x.assign_to_handle(out))
     {
         x.assign_to_handle(out);
     }
+
+    template <class T1, class T2, class U, class = std::enable_if_t<std::is_same<T2, size_t>{}>>
+    auto auto_assign(rank<2>, T1* out_ptr, T2* out_size, U x)
+    {
+        std::copy_n(x.begin(), x.size(), out_ptr);
```
pfultz2, line 540: If `out_ptr` is smaller than `x.size()` then this will write pass the end of the buffer. I would assume `out_size` would be the size of the buffer(passed by value instead of pointer), but then we would need a way to communicate the size back to as well so its probably better as a pointer.

Instead, we probably need to have `out_size` set to the size of `out_ptr` and then later we set it to the smaller size used:

```cpp
template <class T1, class T2, class U, class = std::enable_if_t<std::is_same<T2, size_t>{}>>
auto auto_assign(rank<2>, T1* out_ptr, T2* out_size, U x)
{
    *out_size = std::min(*out_size, x.size());
    std::copy_n(x.begin(), *out_size, out_ptr);
}
```

src/api/include/migraphx/migraphx.hpp
```diff
@@ -628,9 +701,18 @@ struct argument : MIGRAPHX_CONST_HANDLE_BASE(argument)
     template <typename T>
     std::vector<T> as_vector() const
     {
-        size_t vector_len = this->get_shape().bytes() / sizeof(T);
-        T* buffer_ptr     = reinterpret_cast<T*>(this->data());
-        return {buffer_ptr, buffer_ptr + vector_len};
+        auto ss   = this->get_shape();
+        auto lens = ss.lengths();
+        auto num_elements =
+            std::accumulate(lens.begin(), lens.end(), size_t{1}, std::multiplies<>());
```
pfultz2, line 707: Why not just expose `.elements` method?

src/api/include/migraphx/migraphx.hpp
```diff
@@ -628,9 +701,18 @@ struct argument : MIGRAPHX_CONST_HANDLE_BASE(argument)
...
+        std::vector<T> res(num_elements);
+        T* buffer_ptr = reinterpret_cast<T*>(this->data());
+        for(size_t i = 0; i < num_elements; i++)
+        {
+            size_t space_index = ss.index(i);
+            res[i]             = *(buffer_ptr + space_index);
```
pfultz2, line 713: I would write `res[i] = buffer_ptr[ss.index(i)]`.

src/api/include/migraphx/migraphx.hpp
pfultz2: Actually I would prefer to put this at the front though: `MIGRAPHX_INTERFACE_LIFT(1, T, experimental_custom_op, compute)`

tools/api.py
```diff
@@ -308,18 +311,36 @@ def virtual_param(self, prefix: Optional[str] = None) -> str:
...
 
     def virtual_output_args(self, prefix: Optional[str] = None) -> List[str]:
-        return [
-            '&{prefix}{n}'.format(prefix=prefix or '', n=n)
-            for t, n in self.cparams
-        ]
+        container_type = self.type.remove_generic().basic().str()
+        decl_list: List[str] = []
+        for t, n, in self.cparams:
+            if not decl_list and (container_type == "std::vector"
```
pfultz2, line 317: I would prefer this be done by the `vector_c_wrap` function instead of hardcoding `vector` here.

tools/api.py
```diff
@@ -308,18 +311,36 @@ def virtual_param(self, prefix: Optional[str] = None) -> str:
...
+                                  or container_type == "vector"):
+                inner_t = self.type.inner_type()
+                if inner_t:
+                    decl_list.append(
+                        'std::vector<{inner_t}> {prefix}{n}(1024);'.format(
```
pfultz2, line 337: This `1024` is a constant we could declare this as array instead: `std::array<{inner_t}, 1024> {prefix}{n};`

src/api/include/migraphx/migraphx.hpp
umangyadav: Done

src/api/include/migraphx/migraphx.hpp

src/api/include/migraphx/migraphx.hpp

tools/api.py
umangyadav: Done.

src/api/include/migraphx/migraphx.hpp
umangyadav: Exposed `elements()`.

tools/api.py
umangyadav: Yes, but I don't find an easier and nicer way to do it generically.

kahmed10 (approved):

src/include/migraphx/operation.hpp
```diff
@@ -733,12 +782,14 @@ struct operation
...
         virtual bool need_normalization() const                                    = 0;
         virtual bool has_finalize() const                                          = 0;
+        virtual bool has_offload_copy() const                                      = 0;
         virtual lifetime get_lifetime() const                                      = 0;
         virtual std::ptrdiff_t output_alias(const std::vector<shape>& input) const = 0;
         virtual value
         compile(context& ctx, const shape& output, const std::vector<shape>& input) = 0;
         virtual void
         finalize(context& ctx, const shape& output, const std::vector<shape>& input) = 0;
+        virtual bool runs_on_offload_target() const                                  = 0;
```
pfultz2, line 792: These methods should not be added to the operation class. They should be set as attributes.

src/targets/gpu/lowering.cpp
```diff
@@ -236,11 +238,39 @@ struct miopen_apply
             {
                 check_shape(s, insert_precompile_op(it));
             }
+            else if(it->get_operator().has_offload_copy()) // custom op
```
pfultz2, line 241: This should just be an `else`.

umangyadav: `else if` does assertion that it is indeed custom op.

src/include/migraphx/operation.hpp
umangyadav: I need a way to distinguish between regular MIGraphX op and custom op. If i just have an attribute, i need some other mechanism to classify regular MIGrpahX op v/s custom op.

src/targets/gpu/lowering.cpp
umangyadav: I can add `else` clause that throws, if it is better.

pfultz2: This logic should be applied to all cpu only operators not just custom ops.

src/include/migraphx/operation.hpp
pfultz2: I dont think we need to make a distinction. Either way, you can add a `custom_op` attribute.

src/targets/gpu/lowering.cpp
umangyadav: cpu operators you mean e..g `NMS` ? it is part of `apply_map` and handled in first `if` condition.

umangyadav: > This logic should be applied to all cpu only operators not just custom ops.

CPU operators require allocation while for the custom_op we are assuming they are provided.    
Also, can't keep just the else clause, this  `apply()` function is traversing through module, therefore contains @literals, @params etc. also.

src/include/migraphx/operation.hpp
umangyadav: Added logic to use `custom_op` attribute and `target` attribute.

src/targets/gpu/lowering.cpp
```diff
@@ -236,11 +240,37 @@ struct miopen_apply
             {
                 check_shape(s, insert_precompile_op(it));
             }
+            else if(attrs.contains("custom_op") and attrs["custom_op"].get_bool())
```
pfultz2, line 243: This should be `if(attrs.contains("target"))`.

src/targets/gpu/lowering.cpp
```diff
@@ -236,11 +240,37 @@ struct miopen_apply
...
+            {
+                check_shape(s, insert_custom_op(it));
```
pfultz2, line 245: Pass the `attrs` to the function instead of callng it again.

tools/api/api.cpp
```diff
@@ -271,6 +271,9 @@ struct custom_operation
         return pack();
     }
     CustomOp op;
+    bool custom_op     = true;
+    std::string target = op.runs_on_offload_target() ? "gpu" : "cpu";
```
pfultz2, line 275: This needs to be set as an attribute.

src/targets/gpu/lowering.cpp
pfultz2: > CPU operators require allocation while for the custom_op we are assuming they are provided.

I dont really follow this.

> Also, can't keep just the else clause, this apply() function is traversing through module, therefore contains @literals, @params etc. also.

Yea, just check for the `target` attribute.

tools/api/api.cpp
pfultz2: I am not sure how the tests are passing.

src/targets/gpu/lowering.cpp
umangyadav: Done

src/targets/gpu/lowering.cpp
umangyadav: Done

src/api/api.cpp
```diff
@@ -265,12 +265,17 @@ struct experimental_custom_op
...
     template <class Self, class F>
     static auto reflect(Self&, F)
     {
         return pack();
     }
+
+    value attributes() const { return {{"custom_op", true}, {"target", target}}; }
+
     CustomOp op;
+    std::string target = op.runs_on_offload_target() ? "gpu" : "cpu";
```
pfultz2, line 278: This should be moved into the `attributes` function.

src/targets/gpu/lowering.cpp
```diff
@@ -236,11 +240,37 @@ struct miopen_apply
             {
                 check_shape(s, insert_precompile_op(it));
             }
+            else if(attrs.contains("custom_op"))
```
pfultz2, line 243: This should be `attr.contains("target")`.

src/targets/gpu/lowering.cpp
```diff
@@ -236,11 +240,37 @@ struct miopen_apply
...
+            {
+                check_shape(s, insert_custom_op(it, attrs));
+            }
         }
-
         copy_params();
     }
 
+    instruction_ref insert_custom_op(instruction_ref ins, value& attrs) const
```
pfultz2, line 251: This should be `const value&`.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -1255,7 +1341,9 @@ struct experimental_custom_op_base
     virtual std::string name() const                                            = 0;
     virtual argument compute(context ctx, shape output, arguments inputs) const = 0;
     virtual shape compute_shape(shapes inputs) const                            = 0;
-    virtual ~experimental_custom_op_base()                                      = default;
+    virtual std::vector<size_t> output_alias(shapes) const { return {}; }
+    virtual bool runs_on_offload_target() const = 0;
```
pfultz2, line 1345: Can you add a `TODO` comment to return a `string`?

src/targets/gpu/lowering.cpp
umangyadav: yeah, but `operator[` doesn't work with `const` for now.

src/api/include/migraphx/migraphx.hpp
umangyadav: Done

src/targets/gpu/lowering.cpp
umangyadav: Done

src/api/api.cpp
umangyadav: Done

src/targets/gpu/lowering.cpp
umangyadav: made use of `.at()` to work with const.

src/targets/gpu/lowering.cpp
```diff
@@ -21,11 +21,14 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+#include "migraphx/errors.hpp"
```
pfultz2, line 24: This should use angle brackets.

umangyadav: Removed. that, not necessary.

pfultz2 (approved):

PR #1315, opened 2022-07-25
Add accuracy checker tool

kahmed10:
See README for more details on how to to use the accuracy checker.

Currently only works on:
onnx files
fp32 precision
dynamic batch info encoded into file

Tested with resnet50 and bert from mlperf repos.


tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+    sess = ort.InferenceSession(model_name)
+
+    ort_params = {}
+    for i, input in enumerate(sess.get_inputs()):
+        ort_params[input.name] = test_inputs[i]
+
+    pred_ort = sess.run(None, ort_params)[-1]
+
+    is_correct = check_correctness(pred_ort, pred_migx, args.tolerance,
+                                   args.tolerance, args.verbose)
```
umangyadav, line 142: This one is only comparing the final results, if I understood correctly. Is there a way to compare results  layer by layer ?

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            params[key] = migraphx.to_gpu(migraphx.generate_argument(value))
+
+    pred_migx = np.array(migraphx.from_gpu(model.run(params)[-1]))
+
+    sess = ort.InferenceSession(model_name)
+
+    ort_params = {}
+    for i, input in enumerate(sess.get_inputs()):
+        ort_params[input.name] = test_inputs[i]
```
umangyadav, line 137: Can we keep `test_inputs` as a map too instead of list? The list implies implicit order. It may always work in our case, but map is safer.

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+    model_name = args.onnx
+    batch = args.batch
+
+    model = migraphx.parse_onnx(model_name, default_dim_value=batch)
+
+    model.compile(migraphx.get_target('gpu'), offload_copy=False)
+
+    in_shape = []
+    params = {}
+    test_input = None
```
umangyadav, line 113: Don't need this line.

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+
+def main():
+    args = parse_args()
+
+    model_name = args.onnx
+    batch = args.batch
+
+    model = migraphx.parse_onnx(model_name, default_dim_value=batch)
+
+    model.compile(migraphx.get_target('gpu'), offload_copy=False)
```
TedThemistokleous, line 109: Do we always want to be testing on target GPU? What about options to check accuracy on CPU or future targets?

tools/accuracy/accuracy_checker.py
umangyadav: I guess it wouldn't be possible as ORT may be doing different optimizations. But would  be good to have results compared against Ref Target and GPU target within MIGraphX.

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+
+    in_shape = []
```
umangyadav, line 111: This line is not needed.

tools/accuracy/accuracy_checker.py
TedThemistokleous: I agree with this. That would be very useful if we want to detect vanishing or exploding gradients and trace back the cause to a certain input

tools/accuracy/README.md
```diff
@@ -0,0 +1,23 @@
+# AMD MIGraphX Accuracy checker
+## Instructions
+First ensure requirements and MIGraphX's python libary are installed. Refer to MIGraphX instructions at the root directory to install the python library.
```
umangyadav, line 3:
```suggestion
First ensure requirements and MIGraphX's python library are installed. Refer to MIGraphX instructions at the root directory to install the python library.
```

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+    test_input = None
+    test_inputs = []
+    for key, value in model.get_parameter_shapes().items():
+        if args.verbose:
+            print('Parameter {} -> {}'.format(key, value))
+        if not 'output' in key:
+            in_shape = value.lens()
+            in_type = value.type_string()
+            if not args.fill1:
+                test_input = np.random.rand(*(in_shape)).astype(
```
TedThemistokleous, line 122: Can you add an option to also add structured data as well from a datafile? Being able to put in a known working set of inputs will let us verify our migraphx model on our targets will be at parity with unlowered models with no loss to accuracy.

tools/accuracy/README.md
umangyadav, line 3: Add hyperlink to root directory instruction ? Also mention, it only works for GPU at the moment ?

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+
+    pred_ort = sess.run(None, ort_params)[-1]
+
+    is_correct = check_correctness(pred_ort, pred_migx, args.tolerance,
+                                   args.tolerance, args.verbose)
+    if is_correct:
+        print('PASSED: MIGraphX meets tolerance')
+    else:
+        print(
+            'FAILED: MIGraphX is not within tolerance. Rerun with --verbose for detailed information.'
```
TedThemistokleous, line 147: If there's a failure can we just automatically run a verbose output? Is the intent here so for release CI runs, this ensure's we're within some tolerance and the user would then run this locally to see any issues?

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+            else:
+                print('Outputs do not match')
+                break
+
+    return ret
+
+
+def get_np_datatype(in_type):
+    datatypes = {
+        'float64_type': np.float64,
```
umangyadav, line 91: need  to add more dtypes e.g. bool atleast.

tools/accuracy/accuracy_checker.py
```diff
@@ -0,0 +1,152 @@
...
+                        action='store_true',
+                        help='fill all arguments with a value of 1')
+    parser.add_argument('--verbose',
+                        action='store_true',
+                        help='show verbose information (for debugging)')
+    parser.add_argument('--tolerance',
+                        type=float,
+                        default=1e-3,
+                        help='accuracy tolerance (default = 1e-3)')
+    args = parser.parse_args()
```
TedThemistokleous, line 56: Can we add an option for multiple datatypes as well? I could see this being useful to catch issues when we try models with int8 and fp16

tools/accuracy/README.md
kahmed10: done

tools/accuracy/accuracy_checker.py
kahmed10: fixed

tools/accuracy/accuracy_checker.py
kahmed10: A proposed solution to this would be to modify the onnx file (using onnx's python API or external library), then have both runtimes execute the new onnx file. This is an enhancement that can be done later.

tools/accuracy/accuracy_checker.py
kahmed10: Yes, I planned on adding this as an enhancement in the future.

tools/accuracy/accuracy_checker.py
kahmed10: The output can be quite exhaustive if there are a lot of differences between the two. For that reason, I just wanted to start with a simple failure message, then the user can investigate further.

tools/accuracy/accuracy_checker.py
TedThemistokleous: All good.

tools/accuracy/accuracy_checker.py
TedThemistokleous: Makes sense. That's fair.

tools/accuracy/accuracy_checker.py
TedThemistokleous: I think this is better as a todo/phase this in. Disregard last comment.

TedThemistokleous (approved):

TedThemistokleous:
Approving this and think that any additions I was concerned of can be added after the fact. Don't want "perfect" to be the enemy of good enough here as we integrate this into our CI/daily workflow. I'm sure we'll find features add to this. Overall good work @kahmed10

umangyadav (approved):
We can improve this later based on requirements. For now i am good to merge to focus on other things.

tools/accuracy/accuracy_checker.py
kahmed10: done

PR #1317, opened 2022-07-26
Fix literal type in the instance_norm parsing 

umangyadav:
instancenorm parser  always creates literal of type `float` which would fail in type check  while creating binary ops if model  is `fp16`.


CharlieL7:
Should we add a ONNX test using `TensorProto.FLOAT16` to make sure this works for the data types?

umangyadav:
> Should we add a ONNX test using `TensorProto.FLOAT16` to make sure this works for the data types?

Yes, I can add that, thanks.

umangyadav:
> Should we add a ONNX test using `TensorProto.FLOAT16` to make sure this works for the data types?

Added tests.

CharlieL7 (approved):

turneram (approved):

src/onnx/parse_instancenorm.cpp
```diff
@@ -52,6 +52,7 @@ struct parse_instancenorm : op_parser<parse_instancenorm>
         auto scale = args[1];
         auto bias  = args[2];
         auto dims  = x->get_shape().lens();
+        auto dtype = x->get_shape().type();
```
kahmed10, line 55: you need to also check if the type is valid, similar to what we do in the parse_random ops:
```
const std::set<shape::type_t> valid_types = {
        shape::float_type, shape::half_type, shape::double_type};
...
if(not contains(valid_types, dtype))
            MIGRAPHX_THROW(opd.op_name + ": invalid input type: " + std::to_string(dtype) +
                           ". Valid types are float, half, or double.");
```

umangyadav: Added. Thanks.

kahmed10 (approved):

PR #1323, opened 2022-07-29
pybind updates for torch_migraphx library

shivadbhavsar:
This PR applies the following changes to the python api:

1. Add function `argument_from_pointer` to allow directly passing a migraphx.shape object and a memory address. This allows a user of this api to directly pass the memory location of a torch.tensor located on a device without first transferring to host and converting to a numpy array. This should resolve [SWDEV-349247](https://ontrack-internal.amd.com/browse/SWDEV-349247).

Example python usage:
```
def mgx_shape_from_tensor(tensor: torch.tensor) -> migraphx.shape:
    return migraphx.shape(lens=list(tensor.size()),
                          strides=list(tensor.stride()),
                          type=torch_dtype_to_mgx(tensor.dtype))


def mgx_argument_from_tensor(tensor: torch.tensor) -> migraphx.argument:
    shape = mgx_shape_from_tensor(tensor)
    return migraphx.argument_from_pointer(shape, tensor.data_ptr())
```

2.  Expose the `is_compiled()` method from `migraphx::program`. This is required to avoid recompiling a migraphx program when creating a MGXModule in the torch_migraphx library using an already compiled program,

3. Expose the enum types under `migraphx::op`. These are needed in converters written in the torch_migraphx library.


shivadbhavsar:
For enum types, we currently need to cast these to int on the python side when used in kwargs. Ideally, we want to be able to determine if a value is an enum type in the visit_py function and perform the conversion here. However, it is proving troublesome to figure out how to generically check (we can easily check for a specific enum type such as `migraphx::op::pooling_type`) for enum types due to how pybind creates its handles.

Here is some simple python code that can be used for testing @umangyadav :
```
import migraphx
import numpy as np

p = migraphx.program()
mm = p.get_main_module()

inp_np = np.random.randn(1, 3, 24, 24)
shape_mgx = migraphx.shape(lens=list(inp_np.shape), type='float_type')
inp_mgx = mm.add_parameter('x', shape_mgx)

mm.add_instruction(
    migraphx.op('pooling',
                mode=migraphx.op.pooling_mode.average,
                padding=[0, 0],
                stride=[1, 1],
                lengths=[3, 3],
                ceil_mode=False), [inp_mgx])

print(mm)
```

src/py/migraphx_py.cpp
```diff
@@ -358,18 +367,32 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
...
+          return migraphx::make_op(name, v);
+      }))
         .def("name", &migraphx::operation::name);
 
+    py::enum_<migraphx::op::pooling_mode>(op, "pooling_mode")
+        .value("average", migraphx::op::pooling_mode::average)
+        .value("max", migraphx::op::pooling_mode::max)
+        .value("lpnorm", migraphx::op::pooling_mode::lpnorm);
+
+    py::enum_<migraphx::op::padding_mode_t>(op, "padding_mode_t")
```
pfultz2, line 386: We shouldn't expose this enum. It doesnt do anything, and users should not be using it either. Ideally, I would like to remove it as well.

pfultz2:
> Ideally, we want to be able to determine if a value is an enum type in the visit_py function and perform the conversion here. However, it is proving troublesome to figure out how to generically check

To do it generically, we need to check for the `__index__` function and use that to convert it to an integer.

src/py/migraphx_py.cpp
```diff
@@ -268,6 +269,13 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
                  py::buffer_info info = b.request();
                  new(&x) migraphx::argument(to_shape(info), info.ptr);
              })
+        .def(
+            "__init__",
```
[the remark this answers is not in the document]
shivadbhavsar: So this type of overloading is specifically a pybind11 thing. From the pybind11 docs: 
>Note: To define multiple overloaded constructors, simply declare one after the other using the .def(py::init<...>()) syntax. The existing machinery for specifying keyword and default arguments also works.
If its cleaner to have a separate function we can easily go that route too

src/py/migraphx_py.cpp
umangyadav: why is that not useful for convolution?

umangyadav (approved):

kahmed10 (approved):

PR #1325, opened 2022-08-02
Fix parse if

TedThemistokleous:
Fixing condition to which parse if throws an error. using equals doesn't handle the case correctly as rank 1 tensors and scalar values will fail checks regardless of being semantically equivalent, equals if will check if the values of the dimensions themselves are identical.

Taken from the documentation of onnx

_Values that are live-out to the enclosing scope. The return values in the `then_branch` and `else_branch` must be of the same data type. The `then_branch` and `else_branch` may produce tensors with the same element type and different shapes. If corresponding outputs from the then-branch and the else-branch have static shapes S1 and S2, then the shape of the corresponding output variable of the if-node (if present) must be compatible with both S1 and S2 as it represents the union of both possible shapes.For example, if in a model file, the first output of `then_branch` is typed float tensor with shape [2] and the first output of `else_branch` is another float tensor with shape [3], If's first output should have (a) no shape set, or (b) a shape of rank 1 with neither `dim_value` nor `dim_param` set, or (c) a shape of rank 1 with a unique `dim_param`. In contrast, the first output cannot have the shape [2] since [2] and [3] are not compatible._

https://github.com/onnx/onnx/blob/main/docs/Operators.md#If


TedThemistokleous:
Feel free for any suggestions. I'm working on another change that would pad the sizes to the correct shape dimensions for scalars so we don't fail checks in if_op. Let me know what you think would be valid I can flip this change set to a draft and add that if you'd like to see that.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +65,39 @@ struct parse_if : op_parser<parse_if>
...
+                           std::to_string(else_out_shapes.at(0).type()));
+        }
+
+        // Need to check static shapes result
+        if(not then_out_shapes.at(0).dynamic() && not else_out_shapes.at(0).dynamic())
+        {
+            if(then_out_shapes.at(0).scalar())
+            {
+                if(then_out_shapes.at(0).lens().at(0) != else_out_shapes.at(0).lens().at(0) ||
+                   then_out_shapes.at(0).strides().at(0) != 1)
```
umangyadav, line 86: MIGraphX definition of scalar is that `strides` should be zero. 
Also if it is scalar then, it is not dynamic.

umangyadav: `then_out_shapes.at(0).strides().at(0) != 1` is always true if it is a scalar.

TedThemistokleous: Fixed

src/include/migraphx/op/if_op.hpp
```diff
@@ -54,7 +54,8 @@ struct if_op
         if(not std::equal(
                out_shapes1.begin(), out_shapes1.end(), out_shapes0.begin(), out_shapes0.end()))
```
CharlieL7, line 55: From the ONNX spec, the output shapes of the submodules do not have to be the same. But, getting a "compatible" output shape for `If` requires dynamic shape, so might keep this check for now. 

I see that the code in `parse_if.cpp` will make the output shapes the same. I'm not sure exactly what changes will be made when this operator is updated to support dynamic shape. May want to discuss the design.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +65,96 @@ struct parse_if : op_parser<parse_if>
...
+                if(then_out_shapes.at(0).type() != else_out_shapes.at(0).type() &&
+                   then_out_shapes.at(0).elements() < 1)
+                {
+                    convert_ins = then_mdl->insert_instruction(
+                        convert_ins,
+                        migraphx::make_op("convert",
+                                          {{"target_type", else_out_shapes.at(0).type()}}),
+                        convert_ins->inputs().back());
+                    // then_mdl->replace_return({convert_ins});
+                }
```
CharlieL7, line 96: Is this for the case where the output types are not the same but one output shape is scalar? Why would a scalar have the wrong type?

src/include/migraphx/op/if_op.hpp
TedThemistokleous: My assumption is that they should be compatible out of the parse_if. I don't intend on changing the if_op yet since the spec mentioned static shapes. I'm assuming dynamic shapes would be equivalent on output of this?

src/onnx/parse_if.cpp
TedThemistokleous: Empty shapes not being parsed in as the intended type. That may be an erroneous assumption since my other changes in the parse_constant PR seem to identify that shouldn't be the case

src/include/migraphx/op/if_op.hpp
CharlieL7: This looks to be one of the operators that can output a dynamic shape even if the inputs are static/fixed shapes. I'm basing this off of the spec writing:

"""For example, if in a model file, the first output of then_branch is typed float tensor with shape [2] and the first output of else_branch is another float tensor with shape [3], If's first output should have (a) no shape set, or (b) a shape of rank 1 with neither dim_value nor dim_param set, or (c) a shape of rank 1 with a unique dim_param. In contrast, the first output cannot have the shape [2] since [2] and [3] are not compatible."""

So the output shape lengths of the submodules can be different, but they should have the same number of dimensions and type. Per the example, the output of the `If` node could then have a shape of [2] or [3].

TedThemistokleous:
Rebased this off develop, and added additional set of tests that capture the error cases we're seeing with the operator in the current state (Trailing outputs shape sizes of static shapes, with matching iitial digits break) as well as another gotcha I found when testing with one of the input branches having an empty shape, but valid identical type. 

Updated prototypes and tests and will use this to validate changes as I debug and fix this operand. All input is welcome. 
I reused the existing checks for if_then_test, and if_else_test, for the protobufs and thus additional tests to verify since the results should be identical on parse and verification.

TedThemistokleous:
I've gotten past the errors with IF in the resnext model, but I seem to be seeing some oddity when running verification tests. I have a sense in what I've done with add_outline is a bit of a kludge and I think I still need to route that outlined shape to the output of each if/else block for the sub graph outputs to match. Not sure how to do that yet. 

Right now I'm now getting an issue with PARSE_SPLIT() which is a seperate issue on itself.

TedThemistokleous:
@umangyadav I'll need to just add test cases for empty then/else branches, what I was seeing earlier was operating as expected for scalar inputs. A view of netron helped verify this. Looks like codecov actually caught this too delving deeper so I'll be adding some more .onnx files, along with verify/parse tests. 

The re reset should be good to go.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,92 @@ struct parse_if : op_parser<parse_if>
 
         auto then_out_shapes = then_mdl->get_output_shapes();
         auto else_out_shapes = else_mdl->get_output_shapes();
-        if(not std::equal(then_out_shapes.begin(),
-                          then_out_shapes.end(),
-                          else_out_shapes.begin(),
-                          else_out_shapes.end()))
+
+        assert(then_out_shapes.size() == else_out_shapes.size());
```
umangyadav, line 71: instead of  assert, use THROW ?

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,92 @@ struct parse_if : op_parser<parse_if>
...
+            auto handle_empty_branch = [](module_ref& mdl, const shape& out_shape) {
+                auto outline_ins = mdl->add_outline(out_shape);
+                mdl->replace_return({outline_ins});
+            };
+
+            // Handle one empty branch by setting output identical to the other
+            // need to update the then_shape before we do further checks
+            if(then_shape.empty())
+            {
+                handle_empty_branch(then_mdl, else_out_shapes.at(0));
```
umangyadav, line 107: Make lambda return shape expected shape.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,92 @@ struct parse_if : op_parser<parse_if>
...
+                    auto unsqueeze_last_op = [](module_ref& mdl,
+                                                const std::vector<size_t>& out_shape) {
+                        auto convert_ins = mdl->add_instruction(
+                            make_op("unsqueeze", {{"axes", {out_shape.size() - 1}}}),
+                            --(--mdl->end()));
+                        mdl->replace_return({convert_ins});
+                        mdl->remove_instruction({--convert_ins});
+                    };
+
+                    auto last_then = *(--(then_shape.end()));
```
umangyadav, line 138: use `std::prev`.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,92 @@ struct parse_if : op_parser<parse_if>
...
+                       not equal(else_shape.begin(), else_shape.end(), then_shape.begin()))
+                    {
+                        throw_shapes();
+                    }
+
+                    auto unsqueeze_last_op = [](module_ref& mdl,
+                                                const std::vector<size_t>& out_shape) {
+                        auto convert_ins = mdl->add_instruction(
+                            make_op("unsqueeze", {{"axes", {out_shape.size() - 1}}}),
+                            --(--mdl->end()));
```
umangyadav, line 133: std::prev.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,92 @@ struct parse_if : op_parser<parse_if>
...
-                           " then and else sub_grahps must have same output shapes!");
+                           " then and else sub_grahps must have same output type! " +
+                           then_out_shapes.at(0).type_string() + " vs " +
+                           else_out_shapes.at(0).type_string());
+        }
+
+        if(not then_out_shapes.at(0).dynamic() && not else_out_shapes.at(0).dynamic())
+        {
+            auto then_shape   = then_out_shapes.at(0).lens();
+            auto else_shape   = else_out_shapes.at(0).lens();
```
umangyadav, line 86: rename to `then_lens` ? `then_shape` sounds like it refers to `migraphx::shape`.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,95 @@ struct parse_if : op_parser<parse_if>
...
         auto then_out_shapes = then_mdl->get_output_shapes();
         auto else_out_shapes = else_mdl->get_output_shapes();
-        if(not std::equal(then_out_shapes.begin(),
-                          then_out_shapes.end(),
-                          else_out_shapes.begin(),
-                          else_out_shapes.end()))
+
+        auto throw_shapes = [&]() {
+            MIGRAPHX_THROW("PARSE_IF: " + info.name +
+                            " then and else sub_graphs must compatible shapes ");
```
CharlieL7, line 73:
```suggestion
                            " then and else sub_graphs must have compatible shapes ");
```

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,95 @@ struct parse_if : op_parser<parse_if>
...
+
+                if(dim_delta <= 1)
+                {
+                    // make sure dims are equivalent in static shapes
+                    if(not equal(then_lens.begin(), then_lens.end(), else_lens.begin()) &&
+                       not equal(else_lens.begin(), else_lens.end(), then_lens.begin()))
+                    {
+                        throw_shapes();
+                    }
+
```
CharlieL7, line 131: I think this can seg fault:
ex: if then_lens.size() = 4 and else_lens.size() = 3 `equal(then_lens.begin(), then_lens.end(), else_lens.begin())` will check past the last value.

test/onnx/onnx_test.cpp
```diff
@@ -2328,6 +2328,87 @@ TEST_CASE(if_else_test)
...
+
+    std::ifstream ifs("if_else_empty_shape_test.onnx", std::ios::binary);
+    ifs.seekg(0, std::ios::end);
+    auto length = ifs.tellg();
+    ifs.seekg(0, std::ios::beg);
+    std::vector<char> onnx_buffer(length);
+    ifs.read(onnx_buffer.data(), length);
+    ifs.close();
+
+    auto prog = migraphx::parse_onnx_buffer(onnx_buffer.data(), length, {});
```
CharlieL7, line 2368: Why is this done for `if_else` in particular?

src/onnx/parse_if.cpp
TedThemistokleous: Yeah just updated this. Was trying to be smart here and assumed equal played nice with ranges. Added a custom lambda to handle this nicely.

test/onnx/onnx_test.cpp
TedThemistokleous: Not sure grabbed this from the other if_else tests as I reused code.

src/onnx/parse_if.cpp
TedThemistokleous: Done

test/onnx/onnx_test.cpp
```diff
@@ -2353,6 +2353,87 @@ TEST_CASE(if_else_test)
...
+    auto r   = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), ret);
+    mm->add_return({r});
+
+    std::ifstream ifs("if_else_trailing_one_shape_test.onnx", std::ios::binary);
+    ifs.seekg(0, std::ios::end);
+    auto length = ifs.tellg();
+    ifs.seekg(0, std::ios::beg);
+    std::vector<char> onnx_buffer(length);
+    ifs.read(onnx_buffer.data(), length);
+    ifs.close();
```
umangyadav, line 2431: I dont' think this would be necessary. You can make use of `parse_onnx` instead of `parse_onnx_buffer`.

test/onnx/verify_onnx.cpp
```diff
@@ -496,6 +496,69 @@ TEST_CASE(hardsigmoid_verify_test)
...
+
+    migraphx::parameter_map pp;
+    pp["x"] = migraphx::argument(s_data_x, data.data());
+    pp["y"] = migraphx::argument(s_data, data.data());
+
+    auto result = p.eval(pp).back();
+    std::vector<float> result_vector;
+    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
+
+    // protobuff adds ones so result should be just + 1.0
```
umangyadav, line 557: instead of protobuff, write `onnx`

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+                {
+                    throw_shapes();
+                }
+
+                auto unsqueeze_last_op = [](module_ref& mdl, const std::vector<size_t>& out_shape) {
+                    auto convert_ins = mdl->add_instruction(
+                        make_op("unsqueeze", {{"axes", {out_shape.size() - 1}}}),
+                        std::prev(std::prev(mdl->end())));
+                    mdl->replace_return({convert_ins});
+                    mdl->remove_instruction({std::prev(convert_ins)});
```
umangyadav, line 147: DeadCode elimination should take care of removing this.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+                // make sure dims are equivalent in static shapes
+                if(not all_but_last_dims_equal(then_lens, else_lens))
+                {
+                    throw_shapes();
+                }
+
+                auto unsqueeze_last_op = [](module_ref& mdl, const std::vector<size_t>& out_shape) {
+                    auto convert_ins = mdl->add_instruction(
+                        make_op("unsqueeze", {{"axes", {out_shape.size() - 1}}}),
+                        std::prev(std::prev(mdl->end())));
```
umangyadav, line 145: Input to `unsqueeze` should be `std::prev(mdl->end()).inputs().front()`. If return is only returning one argument.

test/onnx/onnx_test.cpp
TedThemistokleous: ok

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+                    mdl->replace_return({convert_ins});
+                    mdl->remove_instruction({std::prev(convert_ins)});
+                };
+
+                auto last_then = *(std::prev(then_lens.end()));
+                auto last_else = *(std::prev(else_lens.end()));
```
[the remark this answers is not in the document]
TedThemistokleous: Ah yes. I can change that

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+            // check equivilant length dims, and (x1,x2,.., xn, 1) == (x1,x2,..,xn)
+            int dim_delta = abs((static_cast<int>(then_lens.size() - else_lens.size())));
+
+            if(dim_delta == 1)
+            {
+                auto all_but_last_dims_equal = [](std::vector<size_t>& lens_a,
+                                                  std::vector<size_t>& lens_b) {
+                    if(lens_a.size() <= lens_b.size())
+                    {
+                        return equal(lens_a.begin(), lens_a.end(), lens_b.begin());
```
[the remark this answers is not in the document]
TedThemistokleous: Yup I can fix

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+                           " then and else sub_graphs must have compatible shapes ");
+        };
+
+        if(then_out_shapes.size() != else_out_shapes.size())
+        {
+            throw_shapes();
+        }
+
+        // Must have the same type for both if/else blocks by onnx spec
+        // Add exception for empty constant scalars
```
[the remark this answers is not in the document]
TedThemistokleous: Not a todo. Should fail because if both vectors are empty, we have no sense of their shape/ it wouldn't be valid onnx.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+        if(then_out_shapes.at(0).type() != else_out_shapes.at(0).type())
```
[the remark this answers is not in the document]
TedThemistokleous: Ah yes we do, and we also have to check the amount of shapes in each output is equal for then/else branches. I can update this

src/onnx/parse_if.cpp
TedThemistokleous: is there a better way to do this? It was odd since I was somehow getting two return statements surprisingly when using replace_return

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+                auto outline_ins = mdl->add_outline(out_shape);
+                mdl->replace_return({outline_ins});
+                return out_shape.lens();
+            };
+
+            // Handle one empty branch by setting output identical to the other
+            // need to update the then_shape before we do further checks
+            if(then_lens.empty())
+            {
+                then_lens = handle_empty_branch(then_mdl, else_out_shapes.at(0));
```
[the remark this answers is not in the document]
TedThemistokleous: Same as above. Ive got a fix i'm pushing that'll work for each output shape from each if/else branch.

src/onnx/parse_if.cpp
TedThemistokleous: Done

test/onnx/verify_onnx.cpp
TedThemistokleous: Done

test/onnx/onnx_test.cpp
TedThemistokleous: Done

src/onnx/parse_if.cpp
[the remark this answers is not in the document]
TedThemistokleous: Done

src/onnx/parse_if.cpp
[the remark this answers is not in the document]
TedThemistokleous: Done.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+                else_lens = handle_empty_branch(else_mdl, then_out_shapes.at(0));
+            }
+
+            // check equivilant length dims, and (x1,x2,.., xn, 1) == (x1,x2,..,xn)
+            int dim_delta = abs((static_cast<int>(then_lens.size() - else_lens.size())));
+
+            if(dim_delta == 1)
+            {
+                auto all_but_last_dims_equal = [](std::vector<size_t>& lens_a,
+                                                  std::vector<size_t>& lens_b) {
```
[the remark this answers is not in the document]
TedThemistokleous: Done

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,103 @@ struct parse_if : op_parser<parse_if>
...
+            if(then_lens.empty())
+            {
+                then_lens = handle_empty_branch(then_mdl, else_out_shapes.at(0));
+            }
+            else if(else_lens.empty())
+            {
+                else_lens = handle_empty_branch(else_mdl, then_out_shapes.at(0));
+            }
+
+            // check equivilant length dims, and (x1,x2,.., xn, 1) == (x1,x2,..,xn)
```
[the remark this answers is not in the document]
TedThemistokleous: Done

src/onnx/parse_if.cpp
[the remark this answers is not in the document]
TedThemistokleous: done

src/onnx/parse_if.cpp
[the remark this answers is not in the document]
TedThemistokleous: Removed the second comment since this is more confusing than it needs to be.

test/onnx/onnx_test.cpp
TedThemistokleous: I ended up removing this.

CharlieL7:
May want to add some ONNX tests to pass diff codecov. Should use something like `EXPECT(test::throws( [&] { migraphx::parse_onnx(...); }))`

TedThemistokleous:
> 

Yeah I know just testing and going to upload a test for the empty constant checks I do which seem to also be coming up realized there's a few branches that aren't covered correctly.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+
+        auto throw_shapes = [&]() {
             MIGRAPHX_THROW("PARSE_IF: " + info.name +
-                           " then and else sub_grahps must have same output shapes!");
+                           " then and else sub_graphs must have compatible shapes ");
+        };
+
+        if(then_out_shapes.size() != else_out_shapes.size())
+        {
+            throw_shapes();
```
pfultz2, line 78: Before we check for compatible shapes we should check that the condition is const, and then just inline the branch that is picked.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+
+                // Throw error if both branches have zero output shapes. Not possible for static
+                // inputs
+                if(then_lens.empty() && else_lens.empty())
+                {
+                    throw_shapes();
+                }
+
+                auto handle_empty_branch = [](module_ref& mdl, const shape& out_shape) {
+                    auto outline_ins = mdl->add_outline(out_shape);
```
pfultz2, line 106: Dont use `add_outline`. This will throw an error if the empty branch is picked. I also plan to remove `outline` in the future since its very problematic and is never helpful.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+                if(then_lens.empty())
+                {
+                    then_lens = handle_empty_branch(then_mdl, else_out_shapes.at(i));
+                }
+                else if(else_lens.empty())
+                {
+                    else_lens = handle_empty_branch(else_mdl, then_out_shapes.at(i));
+                }
+
+                auto all_but_last_dims_equal = [](const std::vector<size_t>& lens_a,
```
pfultz2, line 122: Make this a function.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+                        return std::equal(lens_a.begin(), lens_a.end(), lens_b.begin());
+                    }
+                    else
+                    {
+                        return std::equal(lens_b.begin(), lens_b.end(), lens_a.begin());
+                    }
+                };
+
+                // check equivalent length dims, and (x1,x2,.., xn, 1) == (x1,x2,..,xn)
+                int dim_delta = abs((static_cast<int>(then_lens.size() - else_lens.size())));
```
pfultz2, line 135: I would call this `rank_delta`.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+                    // make sure dims are equivalent in static shapes
+                    if(not all_but_last_dims_equal(then_lens, else_lens))
+                    {
+                        throw_shapes();
+                    }
+
+                    auto last_then = then_lens.back();
+                    auto last_else = else_lens.back();
+
+                    auto unsqueeze_last_op = [](module_ref& mdl,
```
pfultz2, line 148: Make this a function. Also dont pass by reference(`module_ref` is already a pointer).

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+
+        if(then_out_shapes.size() != else_out_shapes.size())
+        {
+            throw_shapes();
+        }
+
+        // Add checks for each output shape
+        for(int i = 0; i < then_out_shapes.size(); i++)
+        {
+            // Must have the same type for both if/else blocks by onnx spec
```
pfultz2, line 84: Index at the top of the loop and assign everything to a variable because `.at` does a bounds check everytime you access it:

```cpp
const auto& then_out_shape = then_out_shapes.at(i);
const auto& else_out_shape = else_out_shapes.at(i);
```

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+            if(then_out_shapes.at(i).type() != else_out_shapes.at(i).type())
+            {
+                MIGRAPHX_THROW("PARSE_IF: " + info.name +
+                               " then and else sub_grahps must have same output type! " +
+                               then_out_shapes.at(i).type_string() + " vs " +
+                               else_out_shapes.at(i).type_string());
+            }
+
+            if(not then_out_shapes.at(i).dynamic() && not else_out_shapes.at(i).dynamic())
```
pfultz2, line 93: Use `and` instead of `&&`.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+                    }
+
+                    auto last_then = then_lens.back();
+                    auto last_else = else_lens.back();
+
+                    auto unsqueeze_last_op = [](module_ref& mdl,
+                                                const std::vector<size_t>& out_shape) {
+                        auto convert_ins = mdl->add_instruction(
+                            make_op("unsqueeze", {{"axes", {out_shape.size() - 1}}}),
+                            std::prev(mdl->end())->inputs().front());
```
pfultz2, line 152: This should be `std::prev(mdl->end())->inputs().at(i)`.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+                    {
+                        throw_shapes();
+                    }
+
+                    auto last_then = then_lens.back();
+                    auto last_else = else_lens.back();
+
+                    auto unsqueeze_last_op = [](module_ref& mdl,
+                                                const std::vector<size_t>& out_shape) {
+                        auto convert_ins = mdl->add_instruction(
```
pfultz2, line 150: You need to do `mdl->insert_instruction(std::prev(mdl->end()), ...)` to insert the instruction before the return.

src/onnx/parse_if.cpp
TedThemistokleous: In this case it does in the input is an empty constant shape to one of the branches, otherwise we'll fail on the if_op of shapes being invalid. I'm using that to gate the size.

src/onnx/parse_if.cpp
TedThemistokleous: Inline as in just use that and ignore the if creation all together?

src/onnx/parse_if.cpp
TedThemistokleous: Will do.

src/onnx/parse_if.cpp
TedThemistokleous: Ah! Right. Good catch. Fixing this

src/onnx/parse_if.cpp
TedThemistokleous: Will do

src/onnx/parse_if.cpp
TedThemistokleous: Ah is that why I'm seeing two returns and need to remove the one then?

src/onnx/parse_if.cpp
pfultz2: > empty constant shape to one of the branches

What does this mean?

src/onnx/parse_if.cpp
pfultz2: yep

src/onnx/parse_if.cpp
pfultz2: Yes, we used to do that:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/e00479af6dc513d8f83e6da773014dee51ce1c2c/src/onnx/parse_if.cpp#L60-L81

We removed it when we added the inliner(in #802), but it seems like it is still better to do inlining in the parser to handle more cases.

src/onnx/parse_if.cpp
TedThemistokleous: Somehow the model generated had inputs to one of the then/else branches as an empty shape constant 0 with a type and then  specific type for the other part of the corresponding IF output. We have no way of determining the shape of that empty constant, only its type. By adding the outline using the other branch's shape is within the onnx spec for how those scenarios are handled.

pfultz2: What is an empty constant?

TedThemistokleous: One of the the inputs to the IF node in the onnx converted model of rexnext50. There isn't shape information outside an empty shape, and a type to use the other branch and that's what fails when running through that model.

src/onnx/parse_if.cpp
TedThemistokleous: More on this, its not just empty shapes but empty constants. We can get the type information but not anything about the shape, as it would register as an empty. Your fix for an empty literal will capture it correctly as having a shape::type {} {} 

In Netron it even shows the type, but has no information about the size.

src/onnx/parse_if.cpp
TedThemistokleous: Done

src/onnx/parse_if.cpp
TedThemistokleous: Done

src/onnx/parse_if.cpp
TedThemistokleous: Done, and removed the reference for module_ref

src/onnx/parse_if.cpp
TedThemistokleous: Done

src/onnx/parse_if.cpp
```diff
@@ -65,13 +67,108 @@ struct parse_if : op_parser<parse_if>
...
+                            make_op("unsqueeze", {{"axes", {out_shape.size() - 1}}}),
+                            std::prev(mdl->end())->inputs().front());
+                        mdl->replace_return({convert_ins});
```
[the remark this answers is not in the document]
TedThemistokleous: Done. Appreciate the insight. Thought I could just slap everything in a for loop and index at(i)

src/onnx/parse_if.cpp
TedThemistokleous: Done

src/onnx/parse_if.cpp
TedThemistokleous: Ah, again thank you for the insight. Thought it was a simpler than it is. This is done.  I guess this has to do with how we handle our returns at the end of each graph?

src/onnx/parse_if.cpp
TedThemistokleous: done

src/onnx/parse_if.cpp
TedThemistokleous: Resolved from another comment.

src/include/migraphx/op/if_op.hpp
TedThemistokleous: More on this, I suppose at a later date we should add dynamic shapes to this operator. 

Does having a dynamic output on one branch result in the static shape taking precedence of what's the valid output? How would that generally work?

src/onnx/parse_if.cpp
TedThemistokleous: image

This part of the Onnx model was added to one of the IF blocks and was a giant source of pain is what I mean. Originally this was being parsed in by default to as a float until @CharlieL7 added in a change to shape() constructor since we kept failing on type. Now the issue is how to infer shape, which there is none.

Without the outline copied from the other model I'm not sure how best to add the proper shape from the other branch. This is valid Onnx as this "empty" shape, but identical type to the other, non empty branch and the output should be the union of those.

Unless the better option is to just create a new literal with matching shapes and then remove the outline entirely?

src/onnx/parse_if.cpp
TedThemistokleous: Ah this seems to be nice if the IF is constant and would speed up our evaluation of this. I can add this in.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +89,91 @@ struct parse_if : op_parser<parse_if>
...
+                // inputs
+                if(then_lens.empty() and else_lens.empty())
+                {
+                    throw_shapes();
+                }
+
+                auto handle_empty_branch = [](module_ref& mdl, const shape& out_shape) {
+                    shape gen_shape(out_shape.type(), out_shape.lens(), out_shape.strides());
+                    auto literal_ins = mdl->add_literal(gen_shape, gen_shape.lens());
+                    mdl->replace_return({literal_ins});
```
umangyadav, line 133: For multiple returns, I guess we just want to modify one particular argument in this loop iteration.

test/onnx/gen_onnx.py
```diff
@@ -2462,6 +3014,111 @@ def if_pl_test():
...
+    node = onnx.helper.make_node('If',
+                                 inputs=['cond'],
+                                 outputs=['res'],
+                                 then_branch=then_body,
+                                 else_branch=else_body)
+
+    return ([node], [x, y], [res], [cond_tensor, xt_tensor, yt_tensor])
+
+
+def if_then_empty_shape_test():
```
umangyadav, line 3070: Would it be possible to add tests for multiple outputs from then and else branch ?

TedThemistokleous: I have one already to test differing shapes on output. You're saying adding more? Is there a particular test you want me to add?

umangyadav: I see this one : if_then_else_incompatible_output_shapes_test 
Without mutiple outputs from then and else branch. But it is testing for incompatibility. Can we have similar test but with compatible shapes ?

TedThemistokleous: Sure I can add a larger tensor and try that as well and retest the trailing case too

umangyadav: > Sure I can add a larger tensor and try that as well and retest the trailing case too

Why larger tensor ? we want to test multiple output from both `then` and `else` branch, and shapes of which should be compatible.

src/onnx/parse_if.cpp
TedThemistokleous: Looks like this will break a bunch of tests if I do it right now. I'll add this change to the end or even a separate PR since I think we'll need additional verification test changes as well to ensure we get the right value.

CharlieL7 (approved):

src/onnx/parse_if.cpp
pfultz2: But does this constant have a value?

I understand an "empty" shape would mean it was a scalar, but an empty constant would not be a valid ONNX, see from [`Constant`](https://github.com/onnx/onnx/blob/main/docs/Operators.md#constant) doc:

> Exactly one of the provided attributes, either value, sparse_value, or value_* must be specified.

And from there we should be able to unsqueeze the scalar to match the shape of the other branch.

TedThemistokleous: Well the unsqueeze doesn't seem to work in this instance. I've tried it and got errors with 

```UNSQUEZE: Input must be scalar.```

I've added a change that removes the use of outline and instead comes up with a new literal with the correct size instead.

Comment (pfultz2) [reply src/onnx/parse_if.cpp]: Ah wait, `value` attribute is for a tensor value, but this is a scalar. Either way we should have at least one value. ONNX requires at least one value, from [Tensor Definition](https://github.com/onnx/onnx/blob/main/docs/IR.md#tensor-definition)(where `S` is the shape and `V` is the values):

> If S has length 0, V must have length 1, since the empty product is defined to be 1. In this case, the tensor represents a scalar.

Comment (TedThemistokleous) [reply src/onnx/parse_if.cpp]: Does that mean squeeze is broken?

```
        if(input_shape.scalar())
        {
            if(old_lens.size() == 1 and old_lens.front() == 1)
                return shape{type, old_lens};
            else
                MIGRAPHX_THROW("UNSQUEEZE: Input must be a scalar");
        }
```
Is where I seem to be failing when I use unsqueeze.

Comment (pfultz2) [reply src/onnx/parse_if.cpp]: Why would it fail there? If the dimensions are empty then we should have a shape of `{1}, {0}`. It seems like we arent passing the shape correctly.

Comment (pfultz2) [reply src/onnx/parse_if.cpp]: Can you show me what this branch looks like in migraphx IR?

Comment (TedThemistokleous) [reply src/onnx/parse_if.cpp]: I think its due to how shape is handling empty inputs right now it defaults to the following

```
 /*!
     * Empty literal with a specific shape type
     */
    explicit literal(shape::type_t shape_type) : m_shape(shape_type, {}) {}
```

Should this be just 

```
 /*!
     * Empty literal with a specific shape type
     */
    explicit literal(shape::type_t shape_type) : m_shape(shape_type, {1}) {}
```
So 1, with no stride instead?

Comment (TedThemistokleous) [reply src/onnx/parse_if.cpp]:
```
[   RUN    ] if_then_empty_constant_test
If_4_if:@0 = @literal{2, 1} -> int64_type, {2, 1}, {1, 1}
If_4_if:@1 = @literal{} -> int64_type, {}, {}
If_4_if:@2 = @return(If_4_if:@0)

If_4_else:@0 = mul(?,?) -> int64_type, {2, 1}, {1, 1}
If_4_else:@1 = @return(If_4_else:@0)
```

Using MIGraphX IR

Comment (pfultz2) [reply src/onnx/parse_if.cpp]: > Trying with both zero len and stride seems toalso break if_literal_test since I assume we are expecting to get shape_type {}, {} for empty values?

Yea right now we dont represent scalars as empty dimensions, we have #519 to update this. Right now we represent the scalar as `{1}, {0}`.

Comment (pfultz2) [reply src/onnx/parse_if.cpp]: So `@literal{} -> int64_type, {}, {}` doesnt look like it was parsed correctly. A literal from onnx should always have a value. If the dimensions are empty in onnx it should look like this `@literal{5} -> int64_type, {1}, {0}`(assuming it has a scalar value of `5`). 

Since the shape has to go to `{2, 1}` then we will need to do a broadcast and a contiguous. We should probably call `common_shape` to get the common shape and then insert multibroadcasts+contiguous to get the common shape. And we could probably do this for all cases.

Comment (pfultz2) [reply src/onnx/parse_if.cpp]: > Should I be adding that as part of this PR now too?

No.

> Well I'd assume here then we can just say an empty value = 0? so it would resolve to

The value can't be empty. Onnx spec requires a value.

File: test/onnx/gen_onnx.py (hunk 34 above)
Comment (TedThemistokleous) [reply test/onnx/gen_onnx.py:3210]: Just added two tests for that, if_then_else_test_multi_outputs_test, if_then_else_test_multi_outputs_test2 in onnx_test for that. 

I've got the last case I'm working on for the empty multi output vector I'm gonna add as well.

File: src/onnx/parse_if.cpp [hunk 35]
```diff
@@ -65,13 +89,102 @@ struct parse_if : op_parser<parse_if>
...
+                // inputs
+                if(then_lens.empty() and else_lens.empty())
+                {
+                    throw_shapes();
+                }
+
+                auto handle_empty_branch = [](module_ref& mdl, int index, const shape& out_shape) {
+                    shape gen_shape(shape(out_shape.type(), {1}, {0}));
+                    auto literal_ins =
+                        mdl->insert_literal(std::prev(mdl->end()), literal(gen_shape, {0}));
```
pfultz2, line 133: You shouldn't be inserting a literal. If we are creating the literal wrong then this should be corrected in `parse_constant`.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +89,102 @@ struct parse_if : op_parser<parse_if>
...
+            }
+
+            if(not then_out_shape.dynamic() and not else_out_shape.dynamic())
+            {
+                auto then_lens = then_out_shape.lens();
+                auto else_lens = else_out_shape.lens();
+
+                // Throw error if both branches have zero output shapes. Not possible for static
+                // inputs
+                if(then_lens.empty() and else_lens.empty())
```
pfultz2, line 125: We should never have an empty shape when parsing onnx. This should be an `assert` instead: `assert(not then_lens.empty() and not else_lens.empty())`.

test/onnx/gen_onnx.py
```diff
@@ -2207,6 +2207,843 @@ def if_else_test():
...
+                                    dims=empty_val_y.shape,
+                                    vals=empty_val_y.flatten().astype(int))
+    shape_const_y = helper.make_node(
+        'Constant_y',
+        inputs=[],
+        outputs=['shape_const_y'],
+        value=empty_ts_y,
+    )
+
+    else_out = helper.make_tensor_value_info('shape_const', TensorProto.INT64,
```
pfultz2, line 2765: For unit tests we should just test the empty shape by itself without if/else branches. There is no reason to test this with if/else branches either as you already have tests for scalars which will cover this.

test/onnx/onnx_test.cpp
```diff
@@ -2454,9 +2519,387 @@ TEST_CASE(if_pl_test)
...
+    auto cond = mm->add_literal(migraphx::literal(sc, {1}));
+    migraphx::shape s{migraphx::shape::int64_type, {2, 1}};
+    std::vector<int> rand = {-1, 0};
+    mm->add_literal(migraphx::shape::int64_type);
+    auto l2 = mm->add_literal(s, rand);
+    auto y  = mm->add_parameter("y", s);
+
+    auto* then_mod = p.create_module("If_4_if");
+    auto then_lit  = then_mod->add_literal(migraphx::shape::int64_type);
+    then_mod->add_return({then_lit});
```
pfultz2, line 2603: Add the return at the end.

test/onnx/onnx_test.cpp
```diff
@@ -2454,9 +2519,387 @@ TEST_CASE(if_pl_test)
...
+
+    migraphx::shape gen_shape(migraphx::shape(s.type(), {1}, {0}));
+    auto literal_ins =
+        then_mod->insert_literal((std::prev(then_mod->end())), migraphx::literal(gen_shape, {0}));
```
pfultz2, line 2607: Dont use `insert_instruction` in the tests, just use `add_instruction`. We should be adding the instructions in the correct order.

test/onnx/onnx_test.cpp
```diff
@@ -2454,9 +2519,387 @@ TEST_CASE(if_pl_test)
...
+        then_mod->insert_instruction(std::prev(then_mod->end()),
+                                     migraphx::make_op("scalar", {{"scalar_bcst_dims", s.lens()}}),
+                                     literal_ins);
+    auto broad_ins =
+        then_mod->insert_instruction(std::prev(then_mod->end()),
+                                     migraphx::make_op("multibroadcast", {{"out_lens", s.lens()}}),
+                                     unsqueeze_ins);
+    auto contig_out = then_mod->insert_instruction(
+        std::prev(then_mod->end()), migraphx::make_op("contiguous"), broad_ins);
+    then_mod->replace_instruction(std::prev(then_mod->end())->inputs().at(0), contig_out);
```
pfultz2, line 2618: Dont do `replace_instructions` in the tests, we should be adding the correct final instructions to the tests. This makes the tests harder to follow.

test/onnx/gen_onnx.py
TedThemistokleous: I wanted to keep the branches in since they're more realistic examples of what could possibly occur as I've seen in the resnext model

test/onnx/onnx_test.cpp
TedThemistokleous: Okay will do.

test/onnx/onnx_test.cpp
TedThemistokleous: Understood. I'll fix this.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +89,102 @@ struct parse_if : op_parser<parse_if>
...
+            // Must have the same type for both if/else blocks by onnx spec
+            if(then_out_shape.type() != else_out_shape.type())
+            {
+                MIGRAPHX_THROW("PARSE_IF: " + info.name +
+                               " then and else sub_grahps must have same output type! " +
+                               then_out_shape.type_string() + " vs " +
+                               else_out_shape.type_string());
+            }
+
+            if(not then_out_shape.dynamic() and not else_out_shape.dynamic())
```
pfultz2, line 118: It would be better to do:

```cpp
if(then_out_shape.dynamic() or else_out_shape.dynamic())
    continue;
```

And avoid extra nesting.

TedThemistokleous:
Will revert parse_split changes out of this branch once other PR(#1429) is reviewed.

src/onnx/parse_if.cpp
TedThemistokleous: Got it. Done.

src/onnx/parse_if.cpp
TedThemistokleous: done

test/onnx/onnx_test.cpp
TedThemistokleous: Done

test/onnx/onnx_test.cpp
TedThemistokleous: Done

test/onnx/onnx_test.cpp
TedThemistokleous: Done

TedThemistokleous:
Updated some tests based on Pauls comments, and changed the assert to handle the case of both empty input literals. Going to be adding update to constant literal parsing as well.

src/onnx/parse_if.cpp
TedThemistokleous: Done.

test/onnx/gen_onnx.py
TedThemistokleous: This is more required than ever with the current set of changes as I'm using identity between branches for empty Then/Else cases. Will keep this in, but will be updated once I add const folding based on the conditional argument.

src/onnx/parse_if.cpp
```diff
@@ -65,13 +89,91 @@ struct parse_if : op_parser<parse_if>
...
+        for(int i = 0; i < then_out_shapes.size(); i++)
+        {
+            const auto& then_out_shape = then_out_shapes.at(i);
+            const auto& else_out_shape = else_out_shapes.at(i);
+
+            // Must have the same type for both if/else blocks by onnx spec
+            if(then_out_shape.type() != else_out_shape.type())
+            {
+                MIGRAPHX_THROW("PARSE_IF: " + info.name +
+                               " then and else sub_grahps must have same output type! " +
```
kahmed10, line 113:
```suggestion
                               " then and else sub_graphs must have same output type! " +
```

TedThemistokleous: Done. Thanks @kahmed10

zack-ch:
Perhaps could add a test case from MLPerf inference
website: https://github.com/mlcommons/inference/tree/master/vision/classification_and_detection
model: https://zenodo.org/record/6617879/files/resnext50_32x4d_fpn.onnx

I tried this branch, if/else error disappeared, but encountered another error:

======
root@mun-node-0:/code# migraphx-driver compile /datasets/resnext50_32x4d_fpn.onnx
Compiling ...
Reading: /datasets/resnext50_32x4d_fpn.onnx
  what():  /root/AMDMIGraphX/src/include/migraphx/check_shapes.hpp:179: same_type: logical_xor: Types do not match
Aborted (core dumped)

TedThemistokleous:
Hey all is there anything else we need on this?  @pfultz2 you mentioned common_shape(), is that really needed here still? 

If someone compiles onnx with an empty if section that goes to the empty section on eval, it should error out execution, which adding the identity operator does here while fixing shapes. 

Were you talking about another edge case?  Something like 

```
a_shape{1, 2, 3} // technically {1,2,3, 0,0,0}
b_shape {1,2,3,4,5,6} 

auto common_s = common_shape({a_shape, b_shape}) //should be {1, 2 ,3 4, 5, 6}
```

TedThemistokleous:
> @TedThemistokleous can you resolve the conflict with test/onnx/gen_onnx.py

Done

TedThemistokleous:
Isn't needed for now due to dynamic rank. Superseded by https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1533 and avoids the issue entirely.

Issue #1316, opened 2022-07-25
Fix typing for empty constants being parsed.


TedThemistokleous:
Related to issue #1283 , it appears we're incorrectly parsing in the type of our empty constant values, which results in the incorrect literal type to be used in the resulting migraphx program.

While trying https://zenodo.org/record/6617879/files/resnext50_32x4d_fpn.onnx with...
migraphx-driver perf ./resnext50_32x4d_fpn.onnx

We get the following output result.

```
module: "If_2644_if"
If_2644_if:@0 = @literal{} -> float_type, {}, {}
If_2644_if:@1 = @return(If_2644_if:@0)

module: "If_28_else"
If_28_else:@0 = identity(If_2644_else:@24) -> int64_type, {31702968, 1}, {1, 1}
If_28_else:@1 = @return(If_28_else:@0)
```

Using netron on the model, indicates that the constant returned in the first If_2644_if branch in migraphx should be of type int64, with an empty tensor.

umangyadav:
https://github.com/onnx/onnx/issues/1803 

Looks like onnx  has defined behaviour on how to interpret empty shapes . 

We may need to suppor that in the onnx parser, if we are  not already. https://github.com/onnx/onnx/blob/main/docs/IR.md

TedThemistokleous:
Important note too:

_A tensor representing a vector of length 1 has shape [1], while a tensor representing a scalar has shape []. They both have a single element, but scalars are not vectors of length 1._

also we only seem to have access to "value" and no shape attributes from the model. It appears that we have no type information from this iether so calls to things like value_int, won't work unless we convert.

TedThemistokleous:
Well oddly, I forced the type to see where this error would go away. Turns out doing the following:

```
module: "If_2644_if"
If_2644_if:@0 = @literal{ ... } -> int64_type, {31702968, 1}, {1, 1}
If_2644_if:@1 = @return(If_2644_if:@0)

module: "If_28_else"
If_28_else:@0 = identity(If_2644_else:@24) -> int64_type, {31702968, 1}, {1, 1}
If_28_else:@1 = @return(If_28_else:@0)

```

After probing the parse_value function further it seems to work as intended so I decided to look at parse_if further to make sure I was getting the right branch.  It looks like its the secondary branch not the initial constant that seems to be the issue as shapes are broken now.

```
module: "If_28_if"
If_28_if:@0 = @literal{1} -> int64_type, {1}, {1}
[...]

PR #1326, opened 2022-08-02
Imply type of literal returned based on input protobuff for zero elem…

TedThemistokleous:
This saves us the default behavior as the onnx parsing assumes that every zero value is float. This way we're still grabbing relevant type information from the protobuff instead and wont fail our data type checks for if them/else blocks from onnx.


TedThemistokleous:
Reverted change and uploaded isolated test case for this, looks like we're breaking even if we remove the return the default buffer.

test/onnx/gen_onnx.py
```diff
@@ -31,6 +31,7 @@
 
 
 def onnx_test(op_test):
+
```
kahmed10, line 34: Please make sure to run the same version of yapf as is found on the CI.

test/onnx/gen_onnx.py
```diff
@@ -626,6 +627,46 @@ def constant_scalar_test():
     return ([node], [], [y])
 
 
+@onnx_test
+def constant_scalar_test2():
```
kahmed10, line 631:
```suggestion
def constant_scalar_test_int64():
```

test/onnx/gen_onnx.py
```diff
@@ -626,6 +627,46 @@ def constant_scalar_test():
...
+            dims=x.shape,
+            vals=x.flatten().astype(np.int64),
+        ),
+    )
+
+    return ([node], [], [y])
+
+
+@onnx_test
+def constant_scalar_test3():
```
kahmed10, line 651: This test does not generate a scalar but instead a constant value with 1 element. Can you please rename this?

TedThemistokleous: Done

test/onnx/gen_onnx.py
TedThemistokleous: Done

test/onnx/gen_onnx.py
TedThemistokleous: Yup. Done

test/onnx/gen_onnx.py
umangyadav: `constant_scalar_test` is also not generating scalar in that case.

test/onnx/gen_onnx.py
```diff
@@ -626,6 +626,46 @@ def constant_scalar_test():
...
+def constant_one_val_int64_test():
+    x = np.array([1]).astype(np.int64)
+    y = helper.make_tensor_value_info('0', TensorProto.INT64, [0])
+
+    node = onnx.helper.make_node(
+        'Constant',
+        inputs=[],
+        outputs=['0'],
+        value=onnx.helper.make_tensor(
+            name='empty_tensor',
```
umangyadav, line 659: neat picking, it is not empty. Otherwise LGTM.

umangyadav (approved):

test/onnx/gen_onnx.py
TedThemistokleous: Sure give me a sec let me fix this.

kahmed10 (approved):

TedThemistokleous:
Fixes #1316

PR #1331, opened 2022-08-09
Explicitly set rocblas_pointer_mode  in examples

umangyadav:
Need to explicitly set pointer mode to device, otherwise for the `sscal`'s `alpha` params it  assumes it is on the Host. For some reason, this wasn't failing on Linux kernel 5.4.0.1xx but it did on 5.13.xx.xxx.


TedThemistokleous (approved):
Fix tidy comments. Apart from that looks good

kahmed10 (approved):

PR #1333, opened 2022-08-09
Enable switching to bare pointer ABI for MLIR

krzysz00:
Once
https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/690
lands, the ABI for MLIR-generated kernels will change. This commit
prepares MIGraphX for the change by conditionally selecting the new
ABI if MLIR reports a sufficiently high API version in its headers.

The new code here is untested.

### DIFF
--- a/src/targets/gpu/mlir.cpp
+++ b/src/targets/gpu/mlir.cpp
@@ -48,6 +48,10 @@
 #include <deque>
 #include <variant>
 
+#if defined(MLIR_MIGRAPHX_API_VERSION) && MLIR_MIGRAPHX_API_VERSION >= 2
+#define MIGRAPHX_MLIR_BARE_POINTER
+#endif
+
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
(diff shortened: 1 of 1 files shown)


pfultz2:
So this does make the kernel call much simpler. I do wonder how we would handle non-packed tensors.

pfultz2 (approved):

krzysz00:
What does "non-packed" mean, in this context?

If the strides and dimension lengths are known at the time the kernel is generated, the new ABI will still work - we'll just need to make sure to pass that information to the kernel generator (which is presumably already happening)

pfultz2:
> What does "non-packed" mean, in this context?

It means it skips over data in memory. Essentially, the offset to last element + 1(we call this the `element_space`) is not equal the number of elements. For example, a packed tensor would be `{2, 4}, {4, 1}`(ie the dims and strides), but a non-packed tensor could be `{2, 4}, {8, 1}`. Usually this comes from a slice or the output buffers for concat.

> we'll just need to make sure to pass that information to the kernel generator

Yea thats definitely the better way to go. 

> (which is presumably already happening)

Currently we can't pass the strides, but hopefully this can be improved in the future.

krzysz00:
Ok, yeah, we don't currently have handling for non-packed tensors in the MLIR interface, since we'd need to get the strides onto the tensor type at the point where we construct the input (which MLIR does have infrastructure for - we're just always setting that to the null attribute/identity map at the moment).

But the new bare pointer ABI will support non-packed tensors, so long as the stride vector is statically known. If it isn't, we'll need that old codepath around so we can pass in the strides.

jungpark-mlir (approved):

src/targets/gpu/mlir.cpp
```diff
@@ -48,6 +48,10 @@
 #include <deque>
 #include <variant>
 
+#if defined(MLIR_MIGRAPHX_API_VERSION) && MLIR_MIGRAPHX_API_VERSION >= 2
```
umangyadav, line 51: Should this be "MLIR_MIGRAPHX_DIALECT_API_VERSION" 2?

umangyadav (approved):

PR #1334, opened 2022-08-10
Update is_supported

varunsh-xilinx:
Based on some discussions with other teams, the original implementation of the `is_supported` API created some difficulties when converting MIGraphX IR to another IR (for the FPGA backend, this was XIR but they apply to any IR):

1. There is not necessarily a direct mapping between one MIGraphX instruction to one XIR instruction. It could be one-to-one, one-to-many, or many-to-one.
2. The original computation graph may need to undergo transformations to make it convertible to XIR. Fully mapping the transformed graph back to the original may not be possible.
3. Additional context is needed around an instruction to say whether it's supported. For example, an instruction may only be supported if followed/preceded by one specific other instruction but not otherwise.

As a result, the `is_supported` and `get_target_assignment` functions are changed in the following ways:

1. `is_supported` accepts a module as an argument, rather than an `instruction_ref`
2. Instead of returning a single numerical value corresponding to the quality of the instruction support, `is_supported` returns an object that contains a vector of discrete instruction_ref ranges that identify which ranges of instructions are supported by each target. Each range is accompanied by the value measuring its quality. This allows the backend to perform any transformations internally and just map the beginning and end to MIGraphX instructions.
3. `get_target_assignments` will call each targets' `is_supported` method, collect these ranges, and stitch together assignments that go from the beginning of the module to the end using the quality metric to help guide the process. With the assumption that a CPU/ref implementation can support all instructions, a legal solution should always exist.

Currently, there are some shortcuts built-in (e.g. the FPGA target assumes that all instructions are supported) but the `test_fpga_get_target_assignments` executable can be run manually to check the assignments.


pfultz2:
> an instruction may only be supported if followed/preceded by one specific other instruction but not otherwise.

"followed/preceded" would be in terms of dataflow not execution order, which means you wouldn't want to use `iterator_range<instruction_ref>`. I think we should have a `find_supported` method to return a vector of supported segments:

```cpp
struct supported_segment
{
    std::vector<instruction_ref> instructions;
    float metric;
};

struct target
{
    std::vector<supported_segment> find_supported(const module& m, support_metric metric) const;
    ...
};
```

src/program.cpp
```diff
@@ -167,13 +196,17 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+    target_subgraphs.reserve(targets.size());
+    for(const auto& t : targets)
     {
-        auto t = std::max_element(
-            targets.begin(), targets.end(), [it, m](const target& lhs, const target& rhs) {
-                return lhs.is_supported(it, m) < rhs.is_supported(it, m);
-            });
-        p.add_assignment(it, t->name());
+        target_subgraphs.emplace_back(t.is_supported(mod, m));
+    }
```
pfultz2, line 204: Use `std::transform`.

src/program.cpp
```diff
@@ -159,6 +160,34 @@ instruction_ref program::validate() const
...
+            if(range.begin() == start)
+            {
+                for(const auto ins : iterator_for(range))
+                {
+                    assignments->add_assignment(ins, target.name());
+                }
+                return range.end();
+            }
+        }
+    }
```
pfultz2, line 184: I dont think this approach will work with `std::vector<instruction_ref>`.

src/program.cpp
```diff
@@ -167,13 +168,37 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
     target_assignments p;
 
     const auto* mod = get_main_module();
-    for(auto it : iterator_for(*mod))
+    std::vector<supported_segments> target_subgraphs;
```
pfultz2, line 171: This should be `std::vector<std::pair<target, supported_segments>>`.

src/program.cpp
```diff
@@ -167,13 +168,37 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+    target_subgraphs.reserve(targets.size());
+    std::transform(targets.begin(),
+                   targets.end(),
+                   std::back_inserter(target_subgraphs),
+                   [mod, m](const auto& t) { return t.find_supported(mod, m); });
```
pfultz2, line 176: Use `[&]` for capture. And also return `make_pair(t, t.find_supported(mod, m))`.

src/include/migraphx/target_assignments.hpp
```diff
@@ -35,6 +35,10 @@ struct target_assignments
 {
     void add_assignment(instruction_ref ins, const std::string& target);
 
+    auto size() const { return assignments.size(); }
+    auto& at(instruction_ref ins) const { return assignments.at(ins); }
+    bool has(instruction_ref ins) const { return assignments.find(ins) != assignments.end(); }
```
pfultz2, line 40: You should just expose `find` so we can just use `contains` function. Also this will make it easier to swap out with a map in the future if we want to.

src/program.cpp
```diff
@@ -167,13 +168,37 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+
+    for(const auto ins : iterator_for(*mod))
     {
-        auto t = std::max_element(
-            targets.begin(), targets.end(), [it, m](const target& lhs, const target& rhs) {
-                return lhs.is_supported(it, m) < rhs.is_supported(it, m);
-            });
-        p.add_assignment(it, t->name());
+        if(p.has(ins))
```
pfultz2, line 180: This should be `if (contains(p, ins))`

src/program.cpp
```diff
@@ -167,13 +168,37 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+        }
+
+        for(auto i = 0U; i < target_subgraphs.size(); ++i)
+        {
+            const auto& subgraph = target_subgraphs[i];
+            const auto& target   = targets[i];
+            for(const auto& segment : subgraph)
+            {
+                const auto& instructions = segment.instructions;
+                if(instructions.find(ins) == instructions.end())
```
pfultz2, line 192: This should be `if (not contains(instructions, ins))`

src/program.cpp
```diff
@@ -167,13 +168,37 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+                {
+                    continue;
+                }
+                for(const auto instr : instructions)
+                {
+                    p.add_assignment(instr, target.name());
+                }
```
pfultz2, line 199: This should use `std::transform`:

```cpp
std::transform(instructions.begin(), instructions.end(), std::inserter(p, p.end()), [&](auto instr) {
    return std::make_pair(instr, target.name());
});
```

You need to add an `insert` method to the target assignments class.

src/program.cpp
```diff
@@ -167,13 +168,37 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
-            targets.begin(), targets.end(), [it, m](const target& lhs, const target& rhs) {
-                return lhs.is_supported(it, m) < rhs.is_supported(it, m);
-            });
-        p.add_assignment(it, t->name());
+        if(p.has(ins))
+        {
+            continue;
+        }
+
+        for(auto i = 0U; i < target_subgraphs.size(); ++i)
```
pfultz2, line 185: After changing to use `std::pair`, I think this can now be `for(const auto&[target, subgraph]:target_subgraphs)`.

tools/include/target.hpp
```diff
@@ -115,17 +116,17 @@ argument copy_from_target(T&, const argument& arg)
...
+    return {};
 }
 
 <%
 interface('target',
      virtual('name', returns='std::string', const=True),
      virtual('get_passes', ctx='context&', options='const compile_options&', returns='std::vector<pass>', const=True),
      virtual('get_context', returns='context', const=True),
-     virtual('is_supported', returns='float', ins='instruction_ref', m='support_metric', const=True, default='target_is_supported'),
+     virtual('find_supported', returns='supported_segments', mod='const module*', m='support_metric', const=True, default='target_find_supported'),
```
pfultz2, line 129: I would prefer `const module&` or `const_module_ref`(which is the same as `const module*`).

src/program.cpp
```diff
@@ -167,13 +168,36 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
-            });
-        p.add_assignment(it, t->name());
+        if(contains(p, ins))
+        {
+            continue;
+        }
+
+        for(const auto& [target, subgraph] : target_subgraphs)
+        {
+            const auto& t = target;
```
pfultz2, line 187: This could be removed and the loop variable could be either `t` or just use `target` down below.

varunsh-xilinx: The reason that's there is you can't pass a structured binding by reference into the lambda so I had to add this variable for that

pfultz2: I see, thats annoying. Maybe add a comment so we know in the future.

pfultz2 (approved):

src/program.cpp
varunsh-xilinx: Done. This can be removed with C++20

PR #1337, opened 2022-08-11
change onnx-rt master branch to main

umangyadav:


umangyadav:
@causten already has  a PR for the same.

PR #1341, opened 2022-08-15
Fix json strings in driver models

kahmed10:
The previous change to use json strings was throwing exceptions because the keys were not enclosed in quotes. This generates the correct json strings.

### DIFF
--- a/src/driver/alexnet.cpp
+++ b/src/driver/alexnet.cpp
@@ -25,7 +25,6 @@
 #include <migraphx/make_op.hpp>
 #include <migraphx/program.hpp>
 #include <migraphx/generate.hpp>
-#include <migraphx/json.hpp>
 #include "models.hpp"
 namespace migraphx {
 namespace driver {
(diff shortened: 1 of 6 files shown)


src/driver/alexnet.cpp
```diff
@@ -42,179 +40,181 @@ migraphx::program alexnet(unsigned batch) // NOLINT(readability-function-size)
...
+                          migraphx::from_json_string(
+                              "{\"dilation\":[1,1],\"group\":1,\"padding\":[2,2,2,2],\"padding_"
+                              "mode\":0,\"stride\":[4,4],\"use_dynamic_same_auto_pad\":0}")),
+        x_0,
+        x_main_module_19); // NOLINT(modernize-raw-string-literal)
     auto x_main_module_21 = mmain->add_instruction(
         migraphx::make_op("broadcast",
-                          migraphx::from_json_string("{axis:1,out_lens:[1,64,55,55]}")),
-        x_main_module_19);
+                          migraphx::from_json_string("{\"axis\":1,\"out_lens\":[1,64,55,55]}")),
```
pfultz2, line 86: Dont use escape quotes. This should call `convert_to_json` which will insert the quotes. Maybe we could add a function that will do `from_json_string(convert_to_json(s))` to make it cleaner.

pfultz2: Or maybe we could have a `make_json_op` function`. I think that would be cleaner.

src/driver/inceptionv3.cpp
```diff
@@ -27,1916 +27,2838 @@
...
+        x_main_module_299); // NOLINT(modernize-raw-string-literal)
     auto x_main_module_588 = mmain->add_instruction(migraphx::make_op("relu"), x_main_module_587);
     auto x_main_module_589 = mmain->add_instruction(
-        migraphx::make_op(
-            "convolution",
-            migraphx::from_json_string(
-                "{dilation:[1,1],group:1,padding:[0,1,0,1],padding_mode:0,stride:[1,1]}")),
+        migraphx::make_op("convolution",
+                          migraphx::from_json_string(
+                              "{\"dilation\":[1,1],\"group\":1,\"padding\":[3,0,3,0],\"padding_"
```
TedThemistokleous, line 1677: Just a naive check/question on this. 

Is it supposed to be  
```
.....bunch of stuff\"padding_""mode\" .........}"
                             ^no escape is deliberate here?
```

src/driver/resnet50.cpp
```diff
@@ -27,1031 +27,1564 @@
...
     auto x_main_module_20 = mmain->add_literal(
-        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {512}}, 19));
-    auto x_main_module_21 = mmain->add_literal(migraphx::generate_literal(
-        migraphx::shape{migraphx::shape::float_type, {512, 512, 3, 3}}, 20));
-    auto x_main_module_22 = mmain->add_literal(
-        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {512}}, 21));
+        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {2048}}, 19));
+    auto x_main_module_21 = mmain->add_literal(
+        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {2048}}, 20));
+    auto x_main_module_22 = mmain->add_literal(migraphx::abs(
```
TedThemistokleous, line 81: Why the sudden use of migraphx::abs() between the previous iteration?

pfultz2: These literal need to be positive numbers.

src/driver/alexnet.cpp
kahmed10: added this function and updated files accordingly

src/driver/inceptionv3.cpp
kahmed10: This is out of date with the new code change, but basically the compiler knows that this is a multi-line string, so it will just concatenate them together without inserting another quotation mark.

src/driver/resnet50.cpp
TedThemistokleous: Oh okay.

src/driver/inceptionv3.cpp
TedThemistokleous: Sounds good ok.

pfultz2 (approved):

TedThemistokleous (approved):

TedThemistokleous:
Odd not sure what happened there. Clicked on redoing my review and it somehow undid Paul's? @kahmed10 LGTM btw

PR #1343, opened 2022-08-17
Run performance benchmarks on PRs

causten:
Remove running during a Push but run the benchmark suite on a pull

### DIFF
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -1,11 +1,9 @@
 name: MIGraphX Performance Tests
 
 on:
-  push:
-    branches: [develop]
-
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/performance.yaml
```diff
@@ -1,11 +1,9 @@
 name: MIGraphX Performance Tests
 
 on:
-  push:
-    branches: [develop]
-
   pull_request:
     branches: [develop]
+    types: [opened, synchronize, closed]
```
umangyadav, line 6: Do we want to run on closed  PR ?

causten: there's another script that checks if closed+merged vs closed

Issue #1303, opened 2022-07-12
Store Tuning information with the MXR


causten:
Goal
Save tuning information and/or code objects with a model (MXR file) so that it can move to other physical servers and still have the tuning information.  If an MXR file has tuning information it should by default use it.

Design Ideas
- Invoke CK operations
- Save traditional MIOpen tuning plans 
- Initiate MIOpen tuning (and then capture it) 
- Limit tuning time (time budgeting) 
- Need to consider version of miopen vs migrapx mismatches
- Need to consider hardware model information.  Should actual MXR file have multiple HW/rocm versions ?  Could you replace /upgrade existing information already in the model
- Dump meta data labels (rocm version, hardware version, what things it has tuned)
- Consider remove tuning information by user request

Notes
- Currently MIGX uses an Immediate Mode [API](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/perfdatabase.html) from MIOpen
- MIOpen team has a new API to facilitate this call Find 2.0

Invocation (user perspective)
- use a new flag (--tune) to indicate to tune via the migraphx-driver command

Phases of development
- get it to work on a single model/rocm version first 
- support different hardware in same MXR
- support different version in MXR

Test Considerations
- Test incompatible execution such as MXR data contains gfx90A data but run on gfx908
- Tune multiple times

umangyadav:
First part : 

Update interface with MIOpen to use the new Find-2.0 API instead of Immediate mode. 

- MIOpen itself provides API to store and load miOpenSolutions struct. We can use that to save tunings information during our `Find()`  call. Load the same solution struct during the `Finalize` call. 
- Add additional fields into `miOpenConvolutionOp` object in MIGraphX to aid in serialization of `miOpenSolution` struct. 
- Store `miOpenSolution` struct using binary value type during serialization. 
-  Refactor or apply same sort of changes for quant_convolution. 
- TBD :  Backwards compatibility with a flag in MIGraphX to disable Find2.0.

Second part : 
Error check for  incompatibility : 
- TBD: Does MIOpen itself throws an error if there is mismatch in rocm version or miopen version or target arch mismatch ? If it does then perhaps no need for doing compatibility check through MIGraphX.

PR #1346, opened 2022-08-18
Use find_2.0 API for the convolution

umangyadav:
- ~Currently, this introduces an environment flag  `MIGRAHPX_MIOPEN_FIND_2` to switch between the older immediate mode and find_2.0 of MIOpen.~
- Introduces a compile-time flag `MIGRAPHX_HAS_FIND_2_API` to switch between two modes  in MIOpen

- ~This PR is a draft PR, because, I want to solicit some feedback on some of the decisions I've made while making changes.  Please Take a look at the comments inside the changes.~ Thaks for the comments. 

Testing:
- Tested inside MIOpen docker with rocm-5.1.  With find_2.0 API mode enabled, it passes all tests of MIGraphX.

Improvements/Additions to be made:
- changes for the quant_convolution, 
- changes for the deconvolution, 
- ~MIGRAPHX_MANAGE_PTRs  for  some of the new objects e.g. `miopenSolution_t`, `miopenFindOptions_t`~
- Macros for MIOpen status checks

TBD
- How should this be tested in CI ? 

Depends on 
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1503

Related issue
-  Store Tuning information with the MXR #1303


src/targets/gpu/convolution.cpp
```diff
@@ -21,9 +21,13 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+#include <migraphx/env.hpp>
 #include <migraphx/gpu/convolution.hpp>
 #include <migraphx/gpu/context.hpp>
 #include <migraphx/generate.hpp>
+#include <miopen/miopen.h>
+
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MIOPEN_FIND_2)
```
pfultz2, line 30: We cant use an env variable as this wont build with the older miopen. This should be a cmake flag.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -40,6 +40,7 @@ struct miopen_convolution
     shared<convolution_descriptor> cd = nullptr;
     miopenConvFwdAlgorithm_t algo{};
     uint64_t solution_id = 0;
+    value::binary solution_object{};
```
pfultz2, line 43: This needs to store the `miopenSolution_t` object(this actually needs to be a managed pointer).

src/targets/gpu/convolution.cpp
```diff
@@ -154,29 +248,32 @@ void miopen_convolution::finalize(context& ctx,
...
-        MIGRAPHX_THROW("MIOpen Convolution: compile solution failed");
+        auto status = miopenConvolutionForwardCompileSolution(ctx.get_stream().get_miopen(),
+                                                              w_desc.get(),
+                                                              x_desc.get(),
+                                                              cd.get(),
+                                                              y_desc.get(),
+                                                              solution_id);
+        if(status != miopenStatusSuccess)
+            MIGRAPHX_THROW("MIOpen Convolution: compile solution failed");
+    }
```
pfultz2, line 276: Finalize needs to call LoadSolution.

src/targets/gpu/convolution.cpp
```diff
@@ -121,6 +152,69 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
...
+        status = miopenGetSolutionSize(solution, &solution_size);
+
+        auto solution_binary = std::vector<char>{};
+        solution_binary.resize(solution_size);
+        status          = miopenSaveSolution(solution, solution_binary.data());
+        solution_object = value::binary{solution_binary.data(), solution_size};
+
+        status = miopenDestroySolution(solution);
+        // which  of the workspace sizes it should return ? from GetSolutionWorkspaceSize or from
+        // ForwardGetWorkspaceSize
```
pfultz2, line 212: It should return workspace from `GetSolutionWorkspaceSize`. `ForwardGetWorkspaceSize` returns the workspace needed to run Find(which will be the largest workspace for all solutions).

src/targets/gpu/convolution.cpp
```diff
@@ -121,6 +152,69 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
...
+        std::size_t found = 0;
+        solutions.resize(100);
+
+        // TODO: create managed pointer for find_options
+        miopenFindOptions_t options;
+
+        // add checks for return statues  later
+        status = miopenCreateFindOptions(&options);
+        // not sure which tuning option to select
+        status = miopenSetFindOptionTuning(options, 1);
```
pfultz2, line 186: Dont set this(also we dont want to enable tuning by default either).

src/targets/gpu/convolution.cpp
```diff
@@ -121,6 +152,69 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
...
+        // not needed by default it is ordered by time
+        status = miopenSetFindOptionResultsOrder(options, miopenFindResultsOrderByTime);
+        // not sure if i should be using workspace size from GetWorkSpaceSize() call from earlier
+        status = miopenSetFindOptionWorkspaceLimit(options, workspace_size);
```
pfultz2, line 190: Dont set this either, we dont want to limit workspace.

src/targets/gpu/convolution.cpp
pfultz2: Also, for find 2.0, you should need to call `ForwardGetWorkspaceSize` size either.

src/targets/gpu/convolution.cpp
```diff
@@ -121,6 +152,69 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
         MIGRAPHX_THROW("MIOpen Convolution: find convolution failed");
     algo = perf.fwd_algo;
 
+    if(enabled(MIGRAPHX_MIOPEN_FIND_2{}))
```
pfultz2, line 155: The lines above this, up(and including) to the `workspace` variable should not be ran for find 2.0. Those are the find 1.0 APIs.

src/targets/gpu/convolution.cpp
```diff
@@ -121,6 +152,69 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
...
+        auto set_tensor_descriptor = [conv_problem](miopenTensorArgumentId_t name,
+                                                    tensor_descriptor& desc) {
+            auto status = miopenSetProblemTensorDescriptor(conv_problem, name, desc.get());
+            if(status != miopenStatusSuccess)
+            {
+                MIGRAPHX_THROW("setting problem tensor description failed");
+            }
+        };
+
+        set_tensor_descriptor(miopenTensorConvolutionX, x_desc);
```
pfultz2, line 172: This function should be moved to miopen.hpp.

src/targets/gpu/convolution.cpp
```diff
@@ -121,6 +152,69 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
         MIGRAPHX_THROW("MIOpen Convolution: find convolution failed");
     algo = perf.fwd_algo;
 
+    if(enabled(MIGRAPHX_MIOPEN_FIND_2{}))
+    {
+        miopenProblem_t conv_problem;
+        status = miopenCreateConvProblem(&conv_problem, cd.get(), miopenProblemDirectionForward);
```
pfultz2, line 158: This should use a manage pointer and use `make_obj` to construct the object.

src/targets/gpu/convolution.cpp
DrizztDoUrden: One doesn't need `ForwardGetWorkspaceSize` for find 2.0. `miopenFindSolutions` doesn't require allocating anything from outside. After that point `miopenGetSolutionWorkspaceSize` would return the exact amount of workspace the solution would require to be executed. `ForwardGetWorkspaceSize` result may exceed what one needs to execute a specific solution. Using it implies that the backends behind find 2.0 and find 1.0 are the same, which may not be true in future.

pfultz2: Yea thanks for the explanation. Sorry I was missing a 'not' in my comment above.

src/targets/gpu/convolution.cpp
umangyadav: Done. But not sure if it would create compatibility issues between MIGraphX  and MIOpen.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
umangyadav: Didnt' find a way to create a managed_ptr since, `loadSolution` requires  pointer to pointer.

src/targets/gpu/convolution.cpp
umangyadav: Made changes,thanks.

src/targets/gpu/convolution.cpp
umangyadav: Removed.

src/targets/gpu/convolution.cpp
umangyadav: Removed.

src/targets/gpu/convolution.cpp
umangyadav: Done.

src/targets/gpu/convolution.cpp
umangyadav: Done

src/targets/gpu/convolution.cpp
umangyadav: Done.

src/targets/gpu/convolution.cpp
umangyadav: Done.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
umangyadav: Added `shared_ptr` for miopenSolution.

DrizztDoUrden (changes requested):

src/targets/gpu/convolution.cpp
```diff
@@ -148,35 +228,54 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
...
     {
-        // Check that workspace hasn't changed
-        auto size = inputs.at(2).bytes();
-        auto ws   = find(ctx, output_shape, inputs);
-        if(ws.bytes() > size)
-            MIGRAPHX_THROW("MIOpen Convolution: workspace has changed during finalization.");
+        (void)(ctx); // avoid warnings
+        (void)(output_shape);
+        (void)(inputs);
+        // load solution
```
DrizztDoUrden, line 243:
```suggestion
        if (solution_ptr != nullptr)
            return;
        // load solution
```

src/targets/gpu/convolution.cpp
```diff
@@ -71,29 +114,66 @@ argument miopen_convolution::compute(context& ctx,
...
+
+        std::size_t solution_size;
+        status = miopenGetSolutionSize(solution_ptr.get(), &solution_size);
+        if(status != miopenStatusSuccess)
+            MIGRAPHX_THROW("MIOpen Convolution: Failed to fetch solution size");
+
+        auto solution_binary = std::vector<char>{};
+        solution_binary.resize(solution_size);
+
+        status = miopenSaveSolution(solution_ptr.get(), solution_binary.data());
```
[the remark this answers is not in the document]
umangyadav: We  want to save tuning and problem information when we serialize the MIGraphX model after compiling the model using MIGraphX.  `saveSolution` would help do that.  

This compiled model then can passed around would be used by seperate processes. MIGraphX would do `loadSolution` during deserialization.

src/targets/gpu/convolution.cpp
```diff
@@ -55,14 +59,53 @@ argument miopen_convolution::compute(context& ctx,
...
+        miopenTensorDescriptor_t descriptors[3] = {x_desc.get(), w_desc.get(), y_desc.get()};
+
+        auto arguments = std::make_unique<miopenTensorArgument_t[]>(3);
+
+        for(auto i = 0; i < 3; ++i)
+        {
+            arguments[i].id         = names[i];
+            arguments[i].descriptor = descriptors[i] != nullptr ? &descriptors[i] : nullptr;
+            arguments[i].buffer     = buffers[i];
+        }
```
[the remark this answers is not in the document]
umangyadav: Is there way to retrieve descriptors from `miopenSolution`  object ? Ideally we would want to assert that they have not been changed.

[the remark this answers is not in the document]
DrizztDoUrden: I could do that, but I think that it would be better design-wise to either make your object effectively immutable or mark it as dirty on changes, rather than fetch data every time to compare.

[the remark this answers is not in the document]
umangyadav: Yes, i think it is unlikely case that descriptors would change. @pfultz2  Can you provide your thoughts ?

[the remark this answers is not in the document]
DrizztDoUrden: And if they have changed, you would need a new find for almost any solution even in the future as it would need a new compilation or at least a new invoker.

src/targets/gpu/convolution.cpp
umangyadav: > If your typical use case is prepare -> finalize in a new process -> not save second time, I would suggest releasing ptr after saving

So the process during compile and serialize is : `find()-->finalize()-->serialize()`.  During `find` it would create binary member variable for `miopen_convolution` class to save the solution as binary.  This would remain in memory until serialize is called. At that time, the contents of `miopen_convolution` gets written to some file. 

it will call `miopenDestroySolution` automatically if the `miopen_convolution` object goes out of scope.  If it is in the same process then, it wouldn't destroy and ptr would remain valid and during `finalize` it would avoid `Loading`. 

I am not sure if it is possible to clear/release binary solution_object member variable once it has been loaded in memory during finalize to avoid having duplicates.

DrizztDoUrden: Why won't you just create binary solution in serialize? And on deserialization you just load it. This way you don't need a member field for binary at all, only storing it as a temporary.

src/targets/gpu/convolution.cpp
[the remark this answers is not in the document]
pfultz2: We can store the descriptors in the convolution object and create them in finalize if we want to, but I dont see a problem with creating them on the fly. I haven't really seen a perf difference with it.

src/targets/gpu/convolution.cpp
```diff
@@ -148,35 +228,58 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
...
+        (void)(output_shape);
+        (void)(inputs);
+        // load solution
+        if(solution_ptr == nullptr)
+        {
+            miopenSolution_t ptr;
+            auto status  = miopenLoadSolution(&ptr,
+                                             reinterpret_cast<const char*>(solution_object.data()),
+                                             solution_object.size());
+            solution_ptr = miopen_solution(ptr);
```
pfultz2, line 250: Use `miopen_solution{ptr}`.

src/targets/gpu/convolution.cpp
pfultz2: We would need to create a custom serialized class to do this. I think what Umang is doing now is fine and we could address this in a future PR.

src/include/migraphx/streamutils.hpp
```diff
@@ -59,6 +60,11 @@ inline stream_range_container<Range> stream_range(const Range& r)
 
 namespace detail {
 
+inline void stream_write_value_impl(rank<3>, std::ostream& os, const migraphx::value::binary&)
+{
+    os << "{binary_object}";
```
pfultz2, line 65: This is nice doing it generically, I would just write a custom `operator<<` for the operator.

On `code_object_op` I usually show the size of the binary. Maybe we could do `os << "{binary_object:" << binary.size() << "}"`.

src/include/migraphx/streamutils.hpp
```diff
@@ -27,6 +27,7 @@
 #include <ostream>
 #include <algorithm>
 #include <migraphx/rank.hpp>
+#include <migraphx/value.hpp>
```
pfultz2, line 30: Actually instead of adding the value.hpp header here. Can we just add a `operator<<` to the `value::binary` class?

umangyadav: `value` class already has `friend operator <<()`.  For the  binary it prints entire binary blob.  in `streamutils`, If i  remove `rank<3>` method then, i think it would still match with `rank<2>` method as a container.

src/targets/gpu/convolution.cpp
umangyadav: Done.

src/include/migraphx/streamutils.hpp
umangyadav: Added `size`  in the prints.

src/include/migraphx/streamutils.hpp
pfultz2: I am saying to add `operator<<` to `value::binary` class not to the `value` class.

src/include/migraphx/streamutils.hpp
pfultz2: > > I would just write a custom operator<< for the operator.
> 
> I can do it, but unlike code_object_op, convolution has containers i.e. padding, dilations etc. in its fields.

No, dont do that. I was just saying what I have done in the past. I guess the "would" is a little ambiguous there. Doing this generically is much better.

src/include/migraphx/streamutils.hpp
umangyadav: Done.

src/include/migraphx/streamutils.hpp
```diff
@@ -59,6 +60,12 @@ inline stream_range_container<Range> stream_range(const Range& r)
 
 namespace detail {
 
+inline void
+stream_write_value_impl(rank<3>, std::ostream& os, const migraphx::value::binary& binary_object)
+{
+    os << binary_object;
+}
```
pfultz2, line 67: This can be removed, and also the value.hpp include as well.

umangyadav: I have it because, otherwise it would match with `rank<2>` method instead.

umangyadav: removed and changed the overloading and ranking.

DrizztDoUrden (approved):
LGTM if it is impossible to call save and load on actual serialization/deserialization. Otherwise I don't like the approach to have two copies of data. Binary solution may get massive at some point. But that may be fixed at some point in time. If postponed, you should make some reminder like issue to change that later.

src/targets/gpu/CMakeLists.txt
```diff
@@ -239,9 +239,14 @@ endif()
 include(CheckLibraryExists)
 get_target_property(MIOPEN_LOCATION MIOpen LOCATION)
 check_library_exists(MIOpen "miopenHiddenSetConvolutionFindMode" "${MIOPEN_LOCATION}" HAS_FIND_MODE_API)
-if(HAS_FIND_MODE_API)
+check_library_exists(MIOpen "miopenFindSolutions" "${MIOPEN_LOCATION}" HAS_FIND_2_API)
+
+if(HAS_FIND_2_API) 
+    target_compile_definitions(migraphx_gpu PUBLIC -DMIGRAPHX_HAS_FIND_2_API)
+    message(STATUS "MIGraphx is using Find-2.0 API of MIOpen")
+elseif(HAS_FIND_MODE_API)
```
pfultz2, line 247: So we still need this find mode API with find 2.0 now because MIOpen doesn't provide an API for it yet. We could address this with a later PR.

pfultz2 (approved):

src/targets/gpu/CMakeLists.txt
pfultz2: I think changing from `elseif` to `if` would fix this, since this is called on the convolution descriptor which find 2.0 will still use.

pfultz2 (changes requested):
I think this might be a simple change for the find mode API.

pfultz2 (approved):

PR #1348, opened 2022-08-23
Improve layernorm and reductions performance

pfultz2:
- Compute mean and variance in same reduction
- Set block size to numbers divisible by 32 instead powers of 2
    - Global is also set exactly instead of being divisible by block size
    - More exact matching of global/local can help get rid of branching/loops
- Reduce vectors first before doing dpp_reduce
- Explicitly vectorize array operators since the compiler doesnt always vectorize them
    - Still uses old for loop when its computing at compile-time since the `reinterpret_cast` nor the all the vector types is supported


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -61,13 +61,25 @@ struct mlir_conv
 MIGRAPHX_REGISTER_OP(mlir_conv);
 
 namespace {
+
+MIGRAPHX_PRED_MATCHER(is_mlir_conv, instruction_ref ins)
```
kahmed10, line 65: is this change related to the PR?

kahmed10: At first glance, I don't see why these MLIR changes are related to the original PR description. But CI was failing without these changes.

kahmed10 (approved):

src/targets/gpu/fuse_mlir.cpp
pfultz2: The test case added to check and issue with the PR changes caused a failure in MLIR, which is why it needed to be updated.

turneram (approved):

PR #1350, opened 2022-08-24
pad_calc fix: size_t -> std::size_t

CharlieL7:
Bug carried over from older code.


umangyadav (approved):
I don't think it makes any difference. 

If it for consistency, we can  make similar change in other places as well.

causten (approved):

CharlieL7:
With some changes made to ROCm stack for GCC compiler, QA found that the compiler produces an unknown type name "size_t" error.

PR #1351, opened 2022-08-24
Insert contiguous for  reshape as necessary

umangyadav:
`reshape` op requires standard shape. During simplify_algebra, it inserts reshapes without checking for this requirement. 
This PR adds contiguous if needed for reshape to avoid errors.


test/simplify_algebra_test.cpp
```diff
@@ -1847,16 +1903,19 @@ TEST_CASE(reorder_reshape_slice)
...
 
-    test(1);
-    test(4);
-    test(8);
+    test(1, true); // check if contiguous is added as necessary if input is transposed
+    test(1, false);
+    test(4, true);
+    test(4, false);
+    test(8, true);
+    test(8, false);
```
pfultz2, line 1918: I think it would be better to make this a template function and register each test case with `TEST_CASE_REGISTER`. Like what we do here: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/float_equal.cpp#L71. This makes debugging easier as we can just run one test case.

umangyadav: Done

test/simplify_algebra_test.cpp
```diff
@@ -1782,13 +1823,19 @@ TEST_CASE(simplify_mul_slice_conv_horiz_fusion)
     }
     EXPECT(m1.sort() == m2.sort());
 }
-TEST_CASE(reorder_reshape_slice)
+
+template <std::size_t BS, bool InTrans>
```
pfultz2, line 1827: Can we rename `InTrans` to `TransposeInput`?

test/simplify_algebra_test.cpp
```diff
@@ -1890,50 +1942,44 @@ TEST_CASE(reorder_reshape_slice_move_axis1)
...
-        run_pass(m1);
-        EXPECT(m1.sort() == m2.sort());
+        auto sum = m2.add_instruction(migraphx::make_op("add"), t0, t1);
+        auto ret = m2.add_instruction(migraphx::make_op("dot"), sum, t2);
+        m2.add_return({ret});
     };
 
-    test(4);
-    test(8);
+    run_pass(m1);
```
pfultz2, line 1973: This is missing a comparison at the end?

pfultz2: `EXPECT(m1.sort() == m2.sort());`

umangyadav: Added

test/simplify_algebra_test.cpp
umangyadav: Done

src/simplify_algebra.cpp
```diff
@@ -993,7 +1001,11 @@ struct find_split_reshape
         std::vector<int64_t> rsp_out_lens(rsp_lens.begin(), rsp_lens.end());
         rsp_out_lens[rsp_axis] = std::accumulate(vec_dims.begin(), vec_dims.end(), std::int64_t{0});
 
-        // insert the reshape instruction
+        // insert the reshape instruction and add contiguous if needed
+        if(!input->get_shape().standard())
```
pfultz2, line 1005: Use `not` instead of `!`.

pfultz2 (approved):

turneram (approved):

PR #1354, opened 2022-08-25
Use rocblas_gemm_ex for batched gemms with broadcasted B

turneram:
Improves performance for 4/6 GEMMs used by huggingface BERT models with batch_size>1 by using a non-batched rocBLAS call for GEMMs where the B input has a broadcasted batch dimension.
The four verify tests added reflect the actual configurations used by bert-base-cased, with varied batch sizes. 

Also adds a matcher to simplify_reshapes to move multibroadcasts after concats.


src/targets/gpu/lowering.cpp
```diff
@@ -316,10 +316,28 @@ struct miopen_apply
...
             assert(refs.size() == 2);
+            /// When batch size > 1 and input B is a literal broadcasted from a 2D literal
+            /// rocblas_gemm_ex can be used in place of rocblas_gemm_strided_batched_ex
+            /// by multiplying m by the number of batches.
+            /// The flag can_unbatch is set here and passed to gemm_impl, where the alternative
+            /// call is made.
+            auto can_unbatch = false;
+            if(refs[0]->get_shape().lens().size() > 2 and refs[0]->get_shape().lens().front() > 1)
+            {
+                if(refs[1]->can_eval())
```
pfultz2, line 327: I am not sure why this needs to be a constant. This should work with any input.

src/targets/gpu/lowering.cpp
```diff
@@ -316,10 +316,28 @@ struct miopen_apply
...
+                    can_unbatch = true;
+                else if(not refs[1]->inputs().empty())
+                {
+                    auto b_in = refs[1]->inputs().front();
+                    if(refs[1]->name() == "gpu::contiguous" and b_in->get_shape().broadcasted() and
```
pfultz2, line 332: You cant just check for broadcast, you need to check that the batch is broadcasted.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -95,7 +95,8 @@ void gemm_impl(context& ctx,
                T alpha,
                T beta,
                bool int8_x4_format,
-               bool compute_fp32)
+               bool compute_fp32,
+               bool can_unbatch)
```
pfultz2, line 99: There is no need to pass in a flag, you can just check that `num_matrices > 0 and get_batch_stride(batch_args[1]) == 0`, right?

turneram: At least in the case of the GEMMs this applies to in huggingface bert, the B matrices are all literals. After propagate_constants they're no longer broadcasted because they get replaced with a literal of the broadcasted shape. I wasn't able to figure out a way to tell if they were originally broadcasted from the state of the graph at gemm_impl, so I moved upstream to lowering, but that is still after propagate_constants, so I had to use the slightly less clear logic that you see there. 
I also wrote a version where we run a pass between simplify_algebra and propagate_constants to insert a reshape on A and skip the broadcast on B, but PR #1280 makes that less straight forward since B comes from another dot operator until after propagate_constants.

pfultz2: > At least in the case of the GEMMs this applies to in huggingface bert, the B matrices are all literals. After propagate_constants they're no longer broadcasted because they get replaced with a literal of the broadcasted shape.

We shouldn't be turning broadcast literals into literals in propagate_constants(unless there is a bug). We do turn broadcasted literal into literals in eliminate_contiguous here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/eliminate_contiguous.cpp#L137

Which happens when `compute_shape` throws an error without contiguous. This most likely comes from here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/include/migraphx/gpu/gemm.hpp#L77

Where we reject broadcasted inputs. However, in #1280 I remove this check, so it should remain a broadcasted input.

turneram: I was just looking back through it, and I was remembering a couple of things incorrectly. So for 3/4 of them, the B input does remain broadcasted at the lowering pass, so I can update that to check that the batch dim is broadcasted and remove the constraint that they're constants. 
The case I was thinking of is just the first GEMM, the horizontally fused one. Before propagate_constant: 
```
main:@51 = @literal{ ... } -> half_type, {768, 768}, {768, 1}
main:@52 = multibroadcast[out_lens={2, 768, 768}](main:@51) -> half_type, {2, 768, 768}, {0, 768, 1}
main:@53 = contiguous(main:@52) -> half_type, {2, 768, 768}, {589824, 768, 1}
main:@54 = @literal{ ... } -> half_type, {768, 768}, {768, 1}
main:@55 = multibroadcast[out_lens={2, 768, 768}](main:@54) -> half_type, {2, 768, 768}, {0, 768, 1}
main:@56 = contiguous(main:@55) -> half_type, {2, 768, 768}, {589824, 768, 1}
main:@57 = @literal{ ... } -> half_type, {768, 768}, {768, 1}
main:@58 = multibroadcast[out_lens={2, 768, 768}](main:@57) -> half_type, {2, 768, 768}, {0, 768, 1}
main:@59 = contiguous(main:@58) -> half_type, {2, 768, 768}, {589824, 768, 1}
main:@60 = concat[axis=2](main:@53,main:@56,main:@59) -> half_type, {2, 768, 2304}, {1769472, 2304, 1}
main:@61 = add(main:@48,main:@50) -> half_type, {2, 16, 768}, {12288, 768, 1}
main:@72 = dot(main:@61,main:@60) -> half_type, {2, 16, 2304}, {36864, 2304, 1}
```
And at the lowering pass:
```
main:@0 = @literal{ ... } -> half_type, {2, 768, 2304}, {1769472, 2304, 1}
main:@34 = gpu::precompile_op[op=pointwise](main:@29,main:@32,main:@33), [main:pointwise2] -> half_type, {2, 16, 768}, {12288, 768, 1}
main:@35 = dot(main:@34,main:@0) -> half_type, {2, 16, 2304}, {36864, 2304, 1}
```
So I think for that one, it's the concat that causes it to get replaced. 

I'll base a branch off of #1280 and see how I can improve it from there.

pfultz2: We can update simplify_reshapes to find a concat with broadcasted inputs to turn into a concat first and then a broadcast as long as the concat axis is not being broadcasted.

turneram: Sounds good. The other three can be checked from gemm_impl after merging in 1280, so I'll make that change to simplify_algebra and then we should be able to remove the flag that I had added.

pfultz2: Well I said simplify_reshapes not simplify_algebra. We have similar find_concat_transpose in simplify_reshapes.

turneram: Sorry, typed that incorrectly. Making the change to simplify_reshapes.

src/simplify_reshapes.cpp
```diff
@@ -271,6 +271,48 @@ struct find_nested_slice
...
+
+    void apply(module& m, const match::matcher_result& mr) const
+    {
+        auto ins        = mr.result;
+        auto op         = any_cast<op::concat>(ins->get_operator());
+        auto out_lens   = ins->get_shape().lens();
+        auto inputs     = ins->inputs();
+        auto in_strides = inputs.front()->get_shape().strides();
+
+        // Axis could be a negative value
```
pfultz2, line 289: Concat should be normalized, so there shouldn't be a negative axis.

pfultz2 (approved):

test/simplify_reshapes_test.cpp
```diff
@@ -337,6 +337,115 @@ TEST_CASE(nop_convert)
     EXPECT(std::distance(m.begin(), m.end()) == n - 1);
 }
 
+TEST_CASE(concat_multibroadcasts1)
+{
+    // Broadcasted batch dim, new axis < old axis
+    migraphx::module m;
+
+    auto s  = migraphx::shape{migraphx::shape::float_type, {3, 4}};
```
kahmed10, line 345: All of these tests have the same program pattern. I suggest a function like this:

```
migraphx::module make_concat_multibroadcast(std::vector<size_t> in_shape, std::vector size_t mbcast_shape, int axis)
{
    migraphx::module m;
    auto s  = migraphx::shape{migraphx::shape::float_type, in_shape};
    auto x  = m.add_parameter("x", s);
    auto y  = m.add_parameter("y", s);
    auto z  = m.add_parameter("z", s);
    auto xm = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mbcast_shape}}), x);
    auto ym = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mbcast_shape}}), y);
    auto zm = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mbcast_shape}}), z);
    auto concat = m.add_instruction(migraphx::make_op("concat", {{"axis", axis}}), xm, ym, zm);
    m.add_return({concat});
    return m;
}

kahmed10 (approved):

Issue #1112, opened 2022-03-02
Add stream syncing


pfultz2:
First we need to extend the `context` and add two methods:

```cpp
struct context
{
    // New methods to add:
    void wait_for(any_ptr queue);
    void finish_on(any_ptr queue);
    // Already included methods:
    any_ptr get_queue();
    void finish();
};
```

We can then add an `execution_environment` class which can be passed to `eval`:

```cpp
struct execution_environment
{
    any_ptr queue = nullptr;
    bool async = false;
};
```

The `eval` method will be update to call the context methods using this class. When `async` is `false`, it will call `ctx.finish()` at the start and end of running. When `async` is `true`, it will call `ctx.wait_for(env.queue)` when it starts running and then call `ctx.finish(env.queue)` when it finishes running.

We will then need to expose `execution_environment` to our API, and add a `run_async` method which takes the `execution_environment`. The original `run` method should become fully synchronous.

causten:
Once this is done ORT can be updated to support it

TedThemistokleous:
Completed as part of 5.3

PR #1358, opened 2022-08-26
Stream sync Changset

TedThemistokleous:
Initial Stream sync changes and associated API level changes.

Need to get this running with Onnx runtime and further tested with using execution environment.

Edit) Linking to issue #1112


TedThemistokleous:
Got this running with some changes I added in changes to a fork of onnxruntime 

found here: https://github.com/TedThemistokleous/onnxruntime
Changes are at SHA-1 : 525a46d9eb807a66ec3156b8e66cafa8519002f8

TedThemistokleous:
[benchmark_summary_20220902-212816.csv](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9481128/benchmark_summary_20220902-212816.csv)

Resulting benchmark with changes

TedThemistokleous:
Here's a before and after of running this with the above changes using the json traces . 
Renamed to .txt instead of .json because github seems to not like .json extensions


[stream_sync_bert_trace.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9481135/stream_sync_bert_trace.txt)
[stock_bert_trace.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9481137/stock_bert_trace.txt)

@pfultz2  What's odd about these is currently it appears like I'm doing the work after an upload as I see activity on the gpu but I'm not seeing any activity outside of copies.

TedThemistokleous:
[bert_post_fixes2.json.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9500718/bert_post_fixes2.json.txt)
Updated output using the most recent changes. Looks like I still have some tweaking to do on the onnxruntime side. 

Getting something like 135 QPS, vs stock 161 QPS. Think I need to remove create/delete of streams between calls.

TedThemistokleous:
Just got this working on my local Navi system.

Seeing now QPS of around 176-178 with this new changes on onnxruntime side linked here:
 https://github.com/TedThemistokleous/onnxruntime/tree/migraphx-stream-sync-changes

Here's the recent trace using the bert-cased model 
[bert_post_fixes_single_compute_stream.json.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9501136/bert_post_fixes_single_compute_stream.json.txt)


Will add csv with comparisons on other servers and cards for this change set as this is reviewed. I may target my docker changes to this for onnxruntime to also see how things play out for perf.

Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM ubuntu:20.04
+FROM pytorch/pytorch:latest
```
pfultz2, line 1: We shouldn't use the pytorch docker images. And we also should never use `latest` as the docker file wont be reproducible.

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 
 ARG PREFIX=/usr/local
 
 # Support multiarch
 RUN dpkg --add-architecture i386
 
 # Add rocm repository
-RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.0.2/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
+RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.2.3/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
```
pfultz2, line 9: We cannot upgrade to 5.2 in our dockerfile because of issue with our CI.

Dockerfile
TedThemistokleous: Funny, right now I'm chasing what looks like to be just that. I'll undo the docker changes and see how CI is. 

I ran that image and got things working locally but have been still having an issue reproducing things on one of our servers

Dockerfile
```diff
@@ -86,6 +90,9 @@ RUN git clone --single-branch --branch ${ONNXRUNTIME_BRANCH} --recursive ${ONNXR
 
 ADD tools/build_and_test_onnxrt.sh /onnxruntime/build_and_test_onnxrt.sh
 
+
+RUN cget -p /usr/local install kitware/cmake@v3.20.0
```
pfultz2, line 94: It would be best to make these docker file changes in a seperate PR though.

src/api/migraphx.py
```diff
@@ -308,6 +314,16 @@ def operation(h):
                  fname='migraphx::save')
 
 
+@api.handle('migraphx_execution_environment',
+            'migraphx::execution_environment')
+def execution_environment(h):
```
pfultz2, line 319: I dont think we should expose the `execution_environment` class.

src/api/migraphx.py
```diff
@@ -274,6 +274,12 @@ def program(h):
                  params='std::unordered_map<std::string, migraphx::argument>'),
              invoke='migraphx::run($@)',
              returns='std::vector<migraphx::argument>')
+    h.method('run_async',
+             api.params(
+                 params='std::unordered_map<std::string, migraphx::argument>',
+                 exec_env='migraphx::execution_environment'),
```
pfultz2, line 280: So `run_async` should take the stream as `void*` and its type name.

src/api/api.cpp
```diff
@@ -166,6 +167,24 @@ void set_output_names(tf_options& options, std::vector<const char*> names)
     options.output_node_names = std::vector<std::string>(names.begin(), names.end());
 }
 
+migraphx_status set_async_stream(execution_environment& e, void* stream)
+{
+    migraphx_status ret = migraphx_status_bad_param;
+    if(stream != nullptr)
+    {
+        e.queue = any_ptr(stream, "ihipStream_t");
```
pfultz2, line 175: We shouldn't hardcode the typename, instead it should be passed from the C++ API level, we have `get_type_name<T>()` function to get the type name as a string:

```cpp
template<class Stream>
arguments run_async(const program_parameters& pparams, Stream s) const
{
    migraphx_arguments_t pout;
    call(&migraphx_program_run_async,
         &pout,
         this->get_handle_ptr(),
         pparams.get_handle_ptr(),
         s,
         get_type_name<Stream>().c_str());
    return arguments(pout, own{});
}
```

src/api/include/migraphx/migraphx.hpp
```diff
@@ -979,6 +993,18 @@ struct program : MIGRAPHX_HANDLE_BASE(program)
         return arguments(pout, own{});
     }
 
+    /// Overloaded to allow for excecuction_environment input
+    arguments eval(const program_parameters& pparams, const execution_environment& e) const
```
pfultz2, line 996: This should be called `run_async`. Ideally, I would like to deprecate `eval` and replace it with `run` so it matches the python API.

src/api/api.cpp
TedThemistokleous: fair, that would make sense. I got a fun surprise during initial testing when what I expected hipStream_t to be the same string wise and was surprised it was "ihipStream_t". I suppose that'll future proof against any other surprises if hip changes the typename?

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -265,12 +267,43 @@ struct context
...
+    {
+        hipStream_t sPtr          = queue.get<hipStream_t>();
+        hip_device::stream ctx_q  = get_stream();
+
+        auto status = hipEventRecord(start_event.get(), sPtr);
+        if(status != hipSuccess)
+        {
+            std::stringstream ss;
+            ss << "wait_for: failed to record " << status;
+            MIGRAPHX_THROW(ss.str());
```
pfultz2, line 280: Just do `MIGRAPHX_THROW("failed to record: " + hip_error(status))`. `hip_error` converts the hip error to a string. Also, `MIGRAPHX_THROW` will add the function name to the message automatically.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -265,12 +267,43 @@ struct context
...
+        }
+        ctx_q.wait(start_event.get());
```
pfultz2, line 282: Just do `get_stream().wait(start_event.get());`.

src/targets/ref/include/migraphx/ref/context.hpp
```diff
@@ -25,13 +25,16 @@
...
+#include <migraphx/any_ptr.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace ref {
 
 struct context
 {
+    void wait_for(any_ptr) const {}
+    void finish_on(any_ptr) const {}
```
pfultz2, line 37: These methods shouldn't need to be added since we default the methods(with `default = 'wait_for_context'`).

src/targets/gpu/include/migraphx/gpu/context.hpp
TedThemistokleous: Will do. Didn't realize that we convert hip_error. That's really handy

src/targets/ref/include/migraphx/ref/context.hpp
TedThemistokleous: Okay I can remove these

src/api/include/migraphx/migraphx.hpp
TedThemistokleous: That's fine. So I'll just rename this and change the call chain appropriately. while leaving the old eval() intact till we deprecate it then?

src/targets/gpu/include/migraphx/gpu/context.hpp
TedThemistokleous: done

TedThemistokleous:
rebased changes ontop of develop

src/api/include/migraphx/migraphx.hpp
TedThemistokleous: done. Made this run_async, and put back eval() to the wait it was before these changes. If we want to rename this to run() to be consistent with the API I can do that as part of this changeset too.

src/api/api.cpp
TedThemistokleous: Done

src/api/migraphx.py
TedThemistokleous: removed

Dockerfile
TedThemistokleous: I think I'll wait for the newest one to come out. Seeing make check failures when running in this 5.2.x container vs the stock one, whereas when I run make check in the stock 5.0.x everything makes sense and is there.

The only thing I could see requiring a change is the sha-1 pointing to onnxruntime to point to the latest Execution Provider that leverages these changes

Dockerfile
TedThemistokleous: Removed. I saw some odd behavior locally as well.

src/include/migraphx/execution_environment.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define MIGRAPHX_GUARD_MIGRAPHLIB_EXECUTION_ENV_HPP
+
+#include <migraphx/any_ptr.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+struct execution_environment
+{
+    execution_environment(any_ptr q, bool is_async) : queue(q), async(is_async){};
```
umangyadav, line 34: default constructor would take care of this. Either way is fine.

src/include/migraphx/program.hpp
```diff
@@ -117,7 +119,7 @@ struct program
     void print_cpp(std::ostream& os) const;
 
     void dry_run(parameter_map params) const;
-
+    //
```
umangyadav, line 122: was this necessary ?

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -319,6 +339,7 @@ struct context
     bool measure_perf                 = false;
     shared<hip_event_ptr> start_event = nullptr;
     shared<hip_event_ptr> stop_event  = nullptr;
+    shared<hip_event_ptr> begin_event, finish_event;
```
umangyadav, line 342: Initialize with `nullptr` ?

Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG PREFIX=/usr/local
 RUN dpkg --add-architecture i386
 
 # Add rocm repository
-RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.0.2/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
+RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/5.2.3/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
```
umangyadav, line 9: Are you planning to keep this change ?

src/targets/gpu/include/migraphx/gpu/context.hpp
TedThemistokleous: Does it have to be? It's set on construction of the stream in the init list

Dockerfile
TedThemistokleous: no I'll remove this

src/include/migraphx/program.hpp
TedThemistokleous: That slipped in. can remove

TedThemistokleous: Slipped in let me remove

src/targets/gpu/include/migraphx/gpu/context.hpp
TedThemistokleous: Change this to be more explicit I suppose on initialization.

src/include/migraphx/execution_environment.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+    any_ptr queue;
```
pfultz2, line 36: This should be set to `nullptr` by default: `any_ptr queue = nullptr`.

src/include/migraphx/execution_environment.hpp
pfultz2, line 34: Remove this constructor, just use the aggregate constructor.

src/include/migraphx/program.hpp
```diff
@@ -78,6 +79,7 @@ struct program
 
     std::vector<argument> eval(parameter_map params) const;
 
+    std::vector<argument> run_async(parameter_map params, execution_environment exec_env) const;
```
pfultz2, line 82: For here, I would still call this `eval`, Only call it `run_async` for the API.

pfultz2: Actually we dont need another method. Just add `execution_environment` parameter to `eval` and have it with default parameter:

```cpp
std::vector<argument> run_async(parameter_map params, execution_environment exec_env = execution_environment{}) const
```

src/program.cpp
```diff
@@ -479,6 +479,100 @@ std::vector<argument> program::eval(parameter_map params) const
...
+                           }));
+    }
+
+    if(exec_env.async)
+    {
+        ctx.finish_on(exec_env.queue);
+    }
+
+    return ret;
+}
```
pfultz2, line 574: Using the same `eval` method will prevent us from copying all this code.

src/include/migraphx/program.hpp
TedThemistokleous: Fair I'll change this back then.

pfultz2:
I think this is pretty close to being done.

You need to add an API test for this. I would copy the `load_and_run` test in test/api/test_gpu.cpp and replace `eval` with `run_async` and remove `set_offload_copy`.

Also, we could add a unit test to ensure the syncing is happening correctly as well under test/gpu/, but I am not sure the best way to write it to avoid timing issues masking the test.

src/targets/ref/include/migraphx/ref/context.hpp
```diff
@@ -25,6 +25,7 @@
 #define MIGRAPHX_GUARD_RTGLIB_CONTEXT_HPP
 
 #include <migraphx/config.hpp>
+#include <migraphx/any_ptr.hpp>
```
pfultz2, line 28: This header can be removed.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -979,6 +979,23 @@ struct program : MIGRAPHX_HANDLE_BASE(program)
         return arguments(pout, own{});
     }
 
+// Needed to enable stream sync functionality at compile time with onnxruntime
+// Helps deal with compatability so we dont break execution provider.
+#define MIGRAPHX_STREAM_SYNC
+    template <class Stream>
+    /// Overloaded to allow for excecuction_environment input
```
umangyadav, line 986:
```suggestion
    /// Overloaded to allow for execution_environment input
```

src/include/migraphx/execution_environment.hpp
TedThemistokleous: Its defaulted to nullptr in any_ptr but I can set it to a default any_ptr() then

Front any_ptr()
```
    private:
    void* ptr                    = nullptr;
    optional<std::type_index> ti = nullopt;
    std::string_view name        = "";
```

TedThemistokleous:
> 

Maybe add some sort of out of order operation that requires that the results happen sequentially?

Also tried it without set_offload_copy() and I was getting gpu faults like:
`root@4fa93fc083f5:/code/AMDMIGraphX/test/onnx# ../../build/bin/test_api_gpu load_and_run_async

[   RUN    ] load_and_run_async
Memory access fault by GPU node-1 (Agent handle: 0xcc8210) on address 0xd89e000. Reason: Page not present or supervisor privilege.
Aborted (core dumped)
`

src/api/include/migraphx/migraphx.hpp
```diff
@@ -979,6 +979,23 @@ struct program : MIGRAPHX_HANDLE_BASE(program)
...
+    /// Overloaded to allow for execution_environment input
+    arguments run_async(const program_parameters& pparams, Stream s) const
+    {
+        migraphx_arguments_t pout;
+        call(&migraphx_program_run_async,
+             &pout,
+             this->get_handle_ptr(),
+             pparams.get_handle_ptr(),
+             s,
+             get_type_name<std::remove_cv_t<std::remove_pointer_t<Stream>>>().c_str());
```
pfultz2, line 995: Actually, I dont know if we want to be compatible with C++11 or not, but if we do then we can't use the `_t` overloads.

test/api/test_gpu.cpp
```diff
@@ -49,6 +49,42 @@ TEST_CASE(load_and_run)
     CHECK(bool{shapes_before.front() == outputs.front().get_shape()});
 }
 
+TEST_CASE(load_and_run_async)
+{
+    auto p             = migraphx::parse_onnx("conv_relu_maxpool_test.onnx");
+    auto shapes_before = p.get_output_shapes();
+    migraphx::compile_options options;
+    options.set_offload_copy();
```
pfultz2, line 57: We dont want offload copy set as that will copy back and forth from the host.

pfultz2:
> Also tried the test you suggested without set_offload_copy() and I was getting gpu faults.

For that test you need to allocate input buffers on the gpu.

src/api/include/migraphx/migraphx.hpp
TedThemistokleous: Oh shoot.  I'm assuming anyone as of late would be usings a more recent compiler C++(14/17/20) or are we targeting systems that still use C++11 as a default?

pfultz2: Yea I dont if we should target c++11 or c++14, I am not sure what onnxrt uses.

Actually, I just realized this could be simplified by changing the function to(notice the `*` next to `Stream`):

```cpp
    template <class Stream>
    arguments run_async(const program_parameters& pparams, Stream* s) const
```

As that will remove the need for `remove_pointer_t`. I dont think we need `remove_cv_t` either as this needs to be non-const.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -979,6 +979,23 @@ struct program : MIGRAPHX_HANDLE_BASE(program)
         return arguments(pout, own{});
     }
 
+// Needed to enable stream sync functionality at compile time with onnxruntime
+// Helps deal with compatability so we dont break execution provider.
+#define MIGRAPHX_STREAM_SYNC
```
pfultz2, line 984: We should remove this, onnxrt should just check for the function or migraphx version.

TedThemistokleous: It allows for onnxruntime to compile based on the migraphx. Do we want to gate this by version then on the ORT side then and it select which lamda to use at runtime?

test/api/test_gpu.cpp
TedThemistokleous: So we should be binding memory locations for that? Similar examples of that anywhere?

src/api/include/migraphx/migraphx.hpp
```diff
@@ -979,6 +979,23 @@ struct program : MIGRAPHX_HANDLE_BASE(program)
...
+    template <class Stream>
+    /// Overloaded to allow for execution_environment input
```
pfultz2, line 986: This comment should be updated, there is no `execution_environment` in the API.

src/program.cpp
```diff
@@ -423,6 +423,12 @@ std::vector<argument> program::eval(parameter_map params) const
 #endif
 
     auto trace_level = value_of(MIGRAPHX_TRACE_EVAL{});
+    std::vector<argument> ret;
+
+    if(exec_env.async)
+    {
+        ctx.wait_for(exec_env.queue);
```
umangyadav, line 430: Should this be asserting that `async` is true for this `exec_env` before calling `wait_for` ?

src/api/include/migraphx/migraphx.hpp
TedThemistokleous: Yup done.

src/program.cpp
TedThemistokleous: No since we want to reuse code between eval() synchronous runs so if the user provides a stream but the flag isn't set, things will run normally as if there was no execution environment. 

Right now the default is to supply an execution environment with a false flag and null stream but, technically the null stream is valid in HIP and can potentially be passed around for other calls. I think the more concerning case is if we get to this call, and the user things they're achieving stream syncing, but not because something erroneously set the async flag to false.

 @pfultz2 , more as a sanity on this one,  is it worth to add an assert if we see  non-null stream and false in the execution env? I've just assumed we'll always have things set correctly via the API/default assignment from eval() signature otherwise.

src/api/include/migraphx/migraphx.hpp
TedThemistokleous: I guess I can use MIGRAPHX_VERSION_MAJOR/MINOR? Is that something we want to expose in our API so we can do version checks for features at runtime then?

src/program.cpp
TedThemistokleous: Done, things are under the same eval() call

src/api/include/migraphx/migraphx.hpp
TedThemistokleous: I just removed this and added a hook in cmake for the API to pass on instead.

TedThemistokleous:
> 

To this point, I've made another PR #1404 with the tests related to this that I'll merge in here.

src/api/include/migraphx/migraphx.hpp
TedThemistokleous: Done.

TedThemistokleous: done

test/eval_test.cpp
```diff
@@ -36,6 +36,8 @@ struct id_target
     struct context
     {
         void finish() const {}
+        void finish_on(migraphx::any_ptr) const {}
+        void wait_for(migraphx::any_ptr) const {}
```
umangyadav, line 40: Is this required anywhere in test file ?

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -274,6 +276,24 @@ struct context
...
+        auto status = hipEventRecord(begin_event.get(), queue.get<hipStream_t>());
+        if(status != hipSuccess)
+            MIGRAPHX_THROW("failed to record " + hip_error(status));
+
+        get_stream().wait(begin_event.get());
+    }
+
+    void finish_on(any_ptr queue)
+    {
+        get_stream().record(finish_event.get());
```
umangyadav, line 290: Should this be 
```
auto status = hipEventRecord(begin_event.get(), queue.get<hipStream_t>());
```

test/const_eval_test.cpp
```diff
@@ -70,6 +70,8 @@ struct non_computable_cf
 struct test_context
 {
     void finish() const {}
+    void finish_on(migraphx::any_ptr) const {}
+    void wait_for(migraphx::any_ptr) const {}
```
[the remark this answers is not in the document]
TedThemistokleous: Can be removed.

test/eval_test.cpp
TedThemistokleous: Can also be removed.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -274,6 +276,24 @@ struct context
         this->current_device = std::make_shared<hip_device>(0, n_streams);
     }
 
+    void wait_for(any_ptr queue)
+    {
+        auto status = hipEventRecord(begin_event.get(), queue.get<hipStream_t>());
+        if(status != hipSuccess)
+            MIGRAPHX_THROW("failed to record " + hip_error(status));
+
+        get_stream().wait(begin_event.get());
```
umangyadav, line 285:
```      
      set_current_device();
      auto status = hipStreamWaitEvent(queue.get(), event, 0);
```

src/targets/gpu/include/migraphx/gpu/context.hpp
TedThemistokleous: Yes but I'm using the record function already implemented for the current stream this is running on.

src/targets/gpu/include/migraphx/gpu/context.hpp
TedThemistokleous: Similar to above, should I be not reusing the existing blocks from the current stream in this context?

test/gpu/stream_sync.cpp
```diff
@@ -0,0 +1,146 @@
...
+const std::string compare_numbers = R"__migraphx__(
+#include <hip/hip_runtime.h>
+
+extern "C" {
+__global__ void compare(float* data) 
+{
+    int i = threadIdx.x + blockDim.x * blockIdx.x;
+    if (data[i] != 1337) 
+    {
+        abort();
```
umangyadav, line 54: Instead of abort(), better case would be to have an if_then_else.

test/gpu/stream_sync.cpp
```diff
@@ -0,0 +1,146 @@
...
+    migraphx::gpu::kernel k1{binaries.front(), "compare"};
+
+    auto input =
+        migraphx::fill_argument({migraphx::shape::float_type, {128}}, stream_sync_test_val);
+
+    auto ginput = migraphx::gpu::to_gpu(input);
+
+    hip_stream_ptr pstream = get_stream();
+
+    k1.launch(pstream.get(), input.get_shape().elements(), 1024)(ginput.cast<float>());
```
umangyadav, line 97: So, for this kernel, it would abort and copy buffer from GPU to CPU ? is this expected ?

src/api/CMakeLists.txt
```diff
@@ -28,6 +28,17 @@ add_library(migraphx_c
 set_target_properties(migraphx_c PROPERTIES EXPORT_NAME c)
 rocm_set_soversion(migraphx_c 3.0)
 
+# Check miopen find mode api
+include(CheckLibraryExists)
+check_library_exists(MIGRAPHX "migraphx_program_run_async" "src/api/include/migraphx.h" HAS_STREAM_SYNC)
```
pfultz2, line 33: This should be in onnxrt. There is no point in us checking our headers as we already know what in them.

test/api/test_gpu.cpp
```diff
@@ -44,11 +46,75 @@ TEST_CASE(load_and_run)
...
 
+using hip_ptr = MIGRAPHX_MANAGE_PTR(void, hipFree);
+
+hip_ptr get_hip_buffer(size_t size)
+{
+    void* ptr;
+    auto err = hipMalloc(&ptr, size);
+    if(err != hipSuccess)
+    {
+        EXPECT(false);
```
pfultz2, line 63: Do `EXPECT(err == hipSuccess)` instead.

test/api/test_gpu.cpp
```diff
@@ -44,11 +46,75 @@ TEST_CASE(load_and_run)
...
+        args.push_back(migraphx::argument::generate(param_shapes[name]));
+        buffs.push_back(get_hip_buffer(args.rbegin()->get_shape().bytes()));
+
+        auto err = hipMemcpy(buffs.rbegin()->get(),
+                             args.rbegin()->data(),
+                             args.rbegin()->get_shape().bytes(),
+                             hipMemcpyHostToDevice);
+        if(err != hipSuccess)
+        {
+            EXPECT(false);
```
pfultz2, line 95: Same here.

test/api/test_gpu.cpp
```diff
@@ -44,11 +46,75 @@ TEST_CASE(load_and_run)
...
+    if(err != hipSuccess)
+    {
+        EXPECT(false);
+    }
+
+    auto outputs = p.run_async(pp, stream);
+    CHECK(shapes_before.size() == outputs.size());
+    CHECK(bool{shapes_before.front() == outputs.front().get_shape()});
+
+    err = hipStreamDestroy(stream);
```
pfultz2, line 111: Add a manage ptr for the stream.

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,77 @@
...
+    # Cleanup Stream
+    err = ctypes.c_long(hip.hipStreamDestroy(stream))
+    if err != hipSuccess:
+        print("hipStreamDestroy failed")
+        return err
+
+    print(result)
+
+
+test_conv_relu()
```
umangyadav, line 77:
```suggestion
if __name__ == "__main__":
    test_conv_relu()
```

src/api/CMakeLists.txt
TedThemistokleous: Ah I see, I can remove this and add to the EP side then.

test/api/test_gpu.cpp
TedThemistokleous: Ah makes sense. I can clean this up

test/api/test_gpu.cpp
TedThemistokleous: On it.

test/py/test_gpu_async.py
TedThemistokleous: Good point. Will add this.

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,77 @@
...
+
+    if err != hipSuccess:
+        print("hipStreamCreate failed")
+        return err
+
+    # Use to_gpu to push generated argument to the GPU before we perform a run
+    for key, value in p.get_parameter_shapes().items():
+        params[key] = migraphx.to_gpu(migraphx.generate_argument(value))
+
+    result = p.run_async(params, stream, "ihipStream_t")
```
umangyadav, line 60: Should it be `hipStream_t` ?

test/api/test_gpu.cpp
TedThemistokleous: Oh okay. Will do.

test/py/test_gpu_async.py
TedThemistokleous: hipStream_t actually resolves to ihipStream_t which I believe aliases a pointer.

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Wait for all commands in stream to complete
+    err = ctypes.c_long(hip.hipStreamSyncrhonize(stream))
+    if err != hipSuccess:
+        print("hipStreamSyncronize failed, invalid handle")
+        return err
```
umangyadav, line 66: Why does it require synchronization to complete the commands? I thought finish event on the stream would takecare of that.

TedThemistokleous: It's blocking this current thread based on the asynchronous stream's progress we created and fired off to do work. If we don't do this we destroy the async thread, from this thread because the command after is a hipStreamDestroy(). This ensures that all the work is complete on the thread and blocks this thread before we destroy the other.

src/api/CMakeLists.txt
```diff
@@ -28,6 +28,17 @@ add_library(migraphx_c
...
+if(HAS_STREAM_SYNC)
+    target_compile_definitions(migraphx_c PUBLIC -DMIGRAPHX_STREAM_SYNC)
+    message(STATUS "MIGRAPHX API has has stream synchronization enabled")
```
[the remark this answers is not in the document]
TedThemistokleous: Removing this and putting it on the EP side.

test/py/test_gpu_async.py
TedThemistokleous: Done

test/api/test_gpu.cpp
TedThemistokleous: Done

test/api/test_gpu.cpp
TedThemistokleous: Done

test/api/test_gpu.cpp
TedThemistokleous: Done

src/api/CMakeLists.txt
TedThemistokleous: Done. Saved copy locally and will add to EP changes

test/gpu/stream_sync.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+TEST_CASE(test_stream_sync)
+{
+    auto binaries = migraphx::gpu::compile_hip_src(
+        {make_src_file("check_stuff.cpp", compare_numbers)}, "", migraphx::gpu::get_device_name());
+    EXPECT(binaries.size() == 1);
+
+    migraphx::gpu::kernel k1{binaries.front(), "compare"};
+    constexpr unsigned int m = 128;
+    constexpr unsigned int k = 8192;
```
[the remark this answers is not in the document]
TedThemistokleous: No problem. Done

test/gpu/stream_sync.cpp
```diff
@@ -0,0 +1,146 @@
...
+    auto goutput = migraphx::gpu::to_gpu(output);
+
+    hip_stream_ptr pstream = get_stream();
+
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+
+    std::vector<float> test_data(k * m, stream_sync_test_val + 100);
+    auto x = mm->add_parameter("x", migraphx::shape{migraphx::shape::float_type, {m, k}});
+    auto y = mm->add_literal(migraphx::shape{migraphx::shape::float_type, {k, m}}, test_data);
```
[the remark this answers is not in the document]
TedThemistokleous: Done

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,77 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
+import migraphx
+import ctypes
+
+
+def test_conv_relu():
+    libname = "libamdhip64.so"
+    hip = ctypes.cdll.LoadLibrary(libname)
```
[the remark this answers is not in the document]
TedThemistokleous: All good. Done

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,77 @@
...
+    print("Compiling ...")
+    # Need to have offload_copy = False to avoid syncs() back to the host device
+    p.compile(migraphx.get_target("gpu"), offload_copy=False)
+    print(p)
+    params = {}
+
+    # Done to avoid parsing enums in ctypes. hipGetDevice always gives us
+    # hipSuccess as an output without modifying state of the current device
+    device_id = ctypes.c_void_p()
+    hipSuccess = ctypes.c_long(hip.hipGetDevice(device_id))
```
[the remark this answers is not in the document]
TedThemistokleous: I thought it follows this spec though: https://rocm-developer-tools.github.io/HIP/group__Device.html#ga7e0e2e8c5f78e3c7449764657c254e0a

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,78 @@
...
+        return err
+
+    # Use to_gpu to push generated argument to the GPU before we perform a run
+    for key, value in p.get_parameter_shapes().items():
+        params[key] = migraphx.to_gpu(migraphx.generate_argument(value))
+
+    result = p.run_async(params, stream, "ihipStream_t")
+
+    # Wait for all commands in stream to complete
+    err = ctypes.c_long(hip.hipStreamSyncrhonize(stream))
```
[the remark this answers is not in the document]
TedThemistokleous: Odd I have this in the CMakeLists.txt file for this change

src/targets/gpu/include/migraphx/gpu/context.hpp
pfultz2: `wait_for` is meant for `get_stream()` to wait on the user-queue, which means we would wait on `get_stream()` and record on the user-queue.

src/targets/gpu/include/migraphx/gpu/context.hpp
umangyadav: Okay, I think i understood things other way around. I was thinking `queue` passed as parameter would be the one that will get used to run the computation. But it is the `get_queue()` one. 

So, in the `eval()` method in `program.cpp`.  Shouldn't it call `wait_for` and `finish_on` one after the another ? 

e.g. right now, it is doing `wait_for()`  that make sure the user-queue has finished all its commands. 

Then, program computes its output using `get_stream()` queue. 

and then, it calls `finish_on` on the user-queue, which makes sure that the user-queue doesn't run until `get_queue` has finished all its commands. but Is there anything that is blocking issues commands to the user-queue when the program is being run with `get_queue`?

test/py/test_gpu_async.py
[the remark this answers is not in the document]
TedThemistokleous: The usual result I see is
289/305 Test #305: test_py_3.8_gpu_async .....................................................   Passed   11.58 sec

Oddly if I keep running make -j$(nproc) check over and over I can cause this.


288/305 Test #305: test_py_3.8_gpu_async .....................................................***Failed   10.33 sec
Segmentation fault

Something non obvious is happening,  I wasn't seeing any output when I ran with the 

```
if __name__ == "main":
```
at the start to run the test_conv_relu

umangyadav (approved):

test/py/test_gpu_async.py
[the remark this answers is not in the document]
TedThemistokleous: Soo bit of a fun one and a lot of non obvious things where.

```
a = ctypes.<some standard type>
b = ctypes.<some standard type>


print (a == b)   -> false.   # Fails since we need to compare value attributes
print(a.value == b.value) # True
```
This explains why sometimes some of the calls would fail, silently even though passing

Combined with how we pass by reference:
```
a = ctypes.c_void_p()
result = c_func(addressof(a))  # causes undefined behavior

#alternative
result = c_func(byref(a)) # works
```
Turns out ctypes has some overhead with both addressof() and byref() calls that handle pass by reference correctly. 

Removed these and then ran into the spelling error around with ` hipStreamSyncrhonize` finally triggering.

Bit of an onion but this runs all the way through now. Just for some reason using the output of run_async in a print results in a segfault still.

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,75 @@
...
+
+    p = migraphx.parse_onnx("conv_relu_maxpool_test.onnx")
+    print(p)
+    print("Compiling ...")
+    # Need to have offload_copy = False to avoid syncs() back to the host device
+    p.compile(migraphx.get_target("gpu"), offload_copy=False)
+    print(p)
+    params = {}
+
+    # Done to avoid parsing enums in ctypes. hipGetDevice always gives us
```
pfultz2, line 39: Need to update this comment.

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,75 @@
...
+    # hipSuccess as an output without modifying state of the current device
+    hipSuccess = ctypes.c_long(0)
+    migraphx.gpu_sync()
```
pfultz2, line 42: This shouldn't be necessary.

test/py/test_gpu_async.py
TedThemistokleous: Done

test/py/test_gpu_async.py
TedThemistokleous: Done

test/py/test_gpu_async.py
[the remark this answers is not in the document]
TedThemistokleous: Removed

pfultz2 (approved):

PR #1360, opened 2022-08-29
Add compute_fp32 flag for quant_gemm tests

umangyadav:
Currently. `test_gpu_pack_int8_args` fails on MI100 machine, because it doesn't set `compute_fp32` flag correctly. This PR fixes the test such that it checks for the device-name, and rocblas-versions and sets this flag accordingly.


test/gpu/pack_int8_args.cpp
```diff
@@ -52,6 +53,25 @@ void run_passes(migraphx::module& m)
                           migraphx::dead_code_elimination{}});
 }
 
+const std::unordered_set<std::string>& get_rocblas_fp32_archs()
+{
+    static std::unordered_set<std::string> supported_archs{"gfx908", "gfx90a"};
+    return supported_archs;
+}
+
+bool get_compute_fp32_flag()
```
pfultz2, line 62: We should move this function to rocblas.hpp header and reuse the function from there. It should take a `context` as a parameter.

umangyadav: Done.

src/targets/gpu/rocblas.cpp
```diff
@@ -41,6 +47,34 @@ rocblas_handle_ptr create_rocblas_handle_ptr(hipStream_t s)
...
+    static std::unordered_set<std::string> supported_archs{"gfx908", "gfx90a"};
+    return supported_archs;
+}
+
+bool get_compute_fp32_flag()
+{
+    bool compute_fp32 = false;
+#if ROCBLAS_VERSION_MAJOR >= 2 && ROCBLAS_VERSION_MINOR >= 38
+    const auto device_name =
+        migraphx::trim(migraphx::split_string(migraphx::gpu::get_device_name(), ':').front());
```
pfultz2, line 61: No need to use `migraphx::` namespace on these functions.

umangyadav: Done

test/gpu/pack_int8_args.cpp
```diff
@@ -81,7 +69,7 @@ TEST_CASE(quant_dot)
         return m;
     };
 
-    auto create_optimized_int8_x4 = [](bool int8_x4) {
+    auto create_optimized_int8_x4 = [](bool int8_x4, bool compute_fp32) {
```
kahmed10, line 72: Does it make sense to have this a parameter to the lambda? Wouldn't it be better to just to remove the parameter and refactor the function in L95 to something like this?
```
{{"int8_x4_format", int8_x4}, {"compute_fp32", migraphx::gpu::get_compute_fp32_flag()}}),
```

umangyadav: yes, 
Made changes.

kahmed10 (approved):

PR #1361, opened 2022-08-29
Enable cppcheck rule for 'not', 'or' keywords

pfultz2:
Using `not` and `or` improves readability. The cppcheck rule will help ensure we are doing it consistently.

This doesnt handle `&&` since they can be used as rvalues references and its hard to check for that with regexes. Hopefully, this can be done with a cppcheck addon in the future.


umangyadav (approved):

src/targets/gpu/kernels/include/migraphx/kernels/integral_constant.hpp
```diff
@@ -73,10 +73,10 @@ MIGRAPHX_INTEGRAL_CONSTANT_BINARY_OP(!=)
 MIGRAPHX_INTEGRAL_CONSTANT_BINARY_OP(&)
 MIGRAPHX_INTEGRAL_CONSTANT_BINARY_OP(^)
 MIGRAPHX_INTEGRAL_CONSTANT_BINARY_OP(|)
-MIGRAPHX_INTEGRAL_CONSTANT_BINARY_OP(&&)
-MIGRAPHX_INTEGRAL_CONSTANT_BINARY_OP(||)
+MIGRAPHX_INTEGRAL_CONSTANT_BINARY_OP(and)
```
CharlieL7, line 76: Are these supposed to be changed?

CharlieL7 (changes requested):

src/targets/gpu/kernels/include/migraphx/kernels/integral_constant.hpp
pfultz2: Yes, although the cppcheck rule will not catch this right now since it uses the preprocessed tokens. When I do this with the addon this will need to be changed since the macro will expand to using those operators. I dont see an issue with changing this now.

CharlieL7 (approved):

PR #1364, opened 2022-08-30
Fix accuracy bug when vectorizing slices

pfultz2:
This fixes the accuracy issues in nasnet models.


TedThemistokleous:
Looks good. guessing the performance dip is because we're doing the extra check and this is expected?

kahmed10 (approved):

PR #1365, opened 2022-08-30
Use workspace size returned from `getSolution()` call for convolutions

umangyadav:
https://rocmsoftwareplatform.github.io/MIOpen/doc/html/find_and_immediate.html

MIOpen calls should be using workspace  size returned from `getSolution()`


src/targets/gpu/convolution.cpp
```diff
@@ -145,9 +144,11 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
     if(status != miopenStatusSuccess)
         MIGRAPHX_THROW("MIOpen Convolution: get solution failed");
 
-    solution_id = solutions.front().solution_id;
+    solution_id    = solutions.front().solution_id;
```
pfultz2, line 147: So you could assign `solutions.front()` to a local variable. Actually, it could probably be refactored to only get 1 solution instead of a vector.

src/targets/gpu/convolution.cpp
```diff
@@ -145,9 +144,11 @@ shape miopen_convolution::find(context& ctx, const shape& output_shape, std::vec
...
+    algo           = solutions.front().algorithm;
+    workspace_size = solutions.front().workspace_size;
```
pfultz2, line 149: Dont reassign this variable. Just use it below directly.

src/targets/gpu/convolution.cpp
umangyadav: How can i make it return single `solution` ?

src/targets/gpu/convolution.cpp
umangyadav: Done.

umangyadav:
Removed checks for `solution_id == 0`.

kahmed10 (approved):

src/targets/gpu/convolution.cpp
umangyadav: Done, Removed call to `miopenGetSolutionCount`.

PR #1367, opened 2022-08-30
Parameterize epsilon for layernorm kernel

kahmed10:
Depends on #1348.

This PR allows for other values of epsilon to be matched when finding layernorm. Similarly, the calculation now uses the variable for epsilon (previously it was a fixed value).


src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
```diff
@@ -94,16 +94,17 @@ MIGRAPHX_DPP_REDUCE(op::max, v_max)
...
 {
+    MIGRAPHX_ASSERT(idx.max_nlocal() == idx.nlocal());
 #if __AMDGCN_WAVEFRONT_SIZE == 32
     constexpr index_int lanes_per_thread = 16;
 #else
     constexpr index_int lanes_per_thread = 64;
 #endif
     using type = decltype(f(0));
-    __shared__ type buffer[idx.nlocal() / lanes_per_thread];
+    __shared__ type buffer[idx.max_nlocal() / lanes_per_thread];
```
TedThemistokleous, line 107: why nlocal to max_nlocal? , if the assert holds above they should be identical no?

src/targets/gpu/kernels/include/migraphx/kernels/layernorm.hpp
```diff
@@ -37,46 +43,45 @@ template <index_int Axis,
...
+        auto variance      = mean_x2 - (mean_x * mean_x);
+        value_type eps_val = eps; // implicit conversion for eps
 
         r.inner([&](auto& y, auto x1, auto x2, auto... xs) {
-            auto m = op(x1, x2) - mean_x;
-            // m * rsqrt(mean(m ^ 2) + 1e-12)
-            y = compute(m * rsqrt(mean_m2 + value_type{1e-12}), xs...);
+            auto x = op(x1, x2);
+            auto m = x - mean_x;
+            // m * rsqrt(mean(m ^ 2) + epsilon)
```
TedThemistokleous, line 66: this intended to be commented out?

TedThemistokleous (commented):
Looks good, seems like a lot more moving parts from adding epsilon but looks like you merged in another branch. 

I'll approve once CI goes through.

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
kahmed10: This is not my change, it's #1348

TedThemistokleous (approved):
Pull in develop lbefore merge. Looks good to me though.

pfultz2 (approved):

PR #1370, opened 2022-08-31
Fix TF literal parsing for relu6

CharlieL7:
* previously always made a float type literal
  * for example, breaks for float16


kahmed10:
Can you add a test with  data type other than float?

kahmed10 (approved):

causten:
Add another reviewer

umangyadav (approved):

PR #1362, opened 2022-08-29
Rewrite ONNX parse batch norm

CharlieL7:
* Rewrites the BatchNormalization ONNX operator into other MIGX operators
  * Added handling of 1D input tensor case (edge case in ONNX spec)
* Removes the `spatial` and `per_activation` functionality (not in the ONNX spec)
* I did not remove the `batch_norm_inference` related code as the TensorFlow parser still uses it
  * Can remove that code when the TF version is updated

PR #1371, opened 2022-08-31
Rewrite TF batch norm; remove batch_norm_inference

CharlieL7:
* Depends on #1362
* Rewrites the TF batch norm like operators to other MIGX operators
* Removes the code related to batch_norm_inference
* Do we have an equivalent to `verify_onnx.cpp` for the TF parser?


TedThemistokleous:
@CharlieL7  can you target this on the branch it depents on from PR #1362 , it'll be easier to see changes relative to that branch

src/tf/parse_batchnorm.cpp
```diff
@@ -38,16 +39,44 @@ struct parse_batchnorm : op_parser<parse_batchnorm>
     instruction_ref parse(const op_desc& /*opd*/,
                           const tf_parser& /*parser*/,
                           tf_parser::node_info info,
-                          const std::vector<instruction_ref>& args) const
+                          std::vector<instruction_ref> args) const
     {
-        float epsilon  = 1e-5f;
-        float momentum = 0.9f;
+        // different default epsilon than from ONNX
+        float epsilon = 1e-4f;
```
TedThemistokleous, line 45: Why is the tolerance lower on MIGraphX vs Onnx? Was that a specific design call made or were we hitting some error without this?

test/tf/gen_tf_pb.py
```diff
@@ -142,19 +144,21 @@ def batchnormv3_test(g1):
...
                                                shape=(32),
-                                               name='4')
+                                               name='variance')
         tf.raw_ops.FusedBatchNormV3(x=g1_input,
                                     scale=g1_scale,
                                     offset=g1_offset,
                                     mean=g1_mean,
                                     variance=g1_variance,
-                                    epsilon=0.00001,
+                                    epsilon=1e-6,
```
TedThemistokleous, line 161: We're more precice for this test but less on the other test for batchnorm_test() above, why is that?

TedThemistokleous (commented):
Few comments on this one. Just curious mainly about the tolerance. Looks like you removed the batch_norm_inference and unrolled everything accordingly as part of the tests. which is nice.

src/tf/parse_batchnorm.cpp
CharlieL7: This is just a different between the TensorFlow spec and the ONNX spec.

test/tf/gen_tf_pb.py
CharlieL7: Testing that the epsilon value is being passed over correctly.

src/tf/parse_batchnorm.cpp
TedThemistokleous: Oh okay, so there's no way around it. Got it

test/tf/gen_tf_pb.py
TedThemistokleous: Oh that makes sense.

TedThemistokleous:
Not sure why CI is being odd on the perf test again. Everything else looks fine for this one.

TedThemistokleous (approved):

src/tf/parse_batchnorm.cpp
umangyadav: Are you using, this https://www.tensorflow.org/api_docs/python/tf/keras/layers/BatchNormalization 

or this : https://www.tensorflow.org/api_docs/python/tf/nn/batch_normalization

I assume, keras one. It uses epsilon = 0.001 == 1e-3f.

test/tf/tf_test.cpp
```diff
@@ -186,50 +185,87 @@ TEST_CASE(batchmatmul_test)
...
+    auto denom    = mm->add_instruction(migraphx::make_op("pow"), var_eps, mb_rt);
+    auto mb_denom = mm->add_instruction(
+        migraphx::make_op("multibroadcast", {{"out_lens", {1, 32, 16, 16}}}), denom);
+    auto div0     = mm->add_instruction(migraphx::make_op("div"), numer, mb_denom);
+    auto mb_scale = mm->add_instruction(
+        migraphx::make_op("multibroadcast", {{"out_lens", {1, 32, 16, 16}}}), usq_scale);
+    auto r0      = mm->add_instruction(migraphx::make_op("mul"), div0, mb_scale);
+    auto mb_bias = mm->add_instruction(
+        migraphx::make_op("multibroadcast", {{"out_lens", {1, 32, 16, 16}}}), usq_bias);
+    mm->add_instruction(migraphx::make_op("add"), r0, mb_bias);
```
umangyadav, line 223: This expansion of batchNorm appears in a few places. Would it be better if we add a helper method `add_bn` in `common.hpp` ?

test/tf/gen_tf_pb.py
```diff
@@ -120,19 +120,21 @@ def batchnorm_test(g1):
     with g1.as_default():
         g1_input = tf.compat.v1.placeholder(tf.float32,
                                             shape=(1, 16, 16, 32),
-                                            name='0')
-        g1_scale = tf.constant(1.0, dtype=tf.float32, shape=[32], name='1')
-        g1_offset = tf.compat.v1.placeholder(tf.float32, shape=(32), name='2')
-        g1_mean = tf.compat.v1.placeholder(tf.float32, shape=(32), name='3')
+                                            name='x')
+        g1_scale = tf.constant(1.0, dtype=tf.float32, shape=[32], name='scale')
```
umangyadav, line 124: Would be good to have `fp16` test and a test that would hit `unsqueeze` lines for batch norm parameters.

umangyadav (approved):

src/driver/resnet50.cpp
```diff
@@ -25,1453 +25,912 @@
...
-    auto x_main_module_12 = mmain->add_literal(migraphx::abs(
-        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {512}}, 11)));
-    auto x_main_module_13 = mmain->add_literal(migraphx::generate_literal(
-        migraphx::shape{migraphx::shape::float_type, {512, 512, 3, 3}}, 12));
-    auto x_main_module_14 = mmain->add_literal(migraphx::abs(
-        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {512}}, 13)));
-    auto x_main_module_15 = mmain->add_literal(
-        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {512}}, 14));
+        migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {512}}, 4));
+    auto x_main_module_6 = mmain->add_literal(migraphx::generate_literal(
```
umangyadav, line 49: Not for this PR, but i guess we can write these models similar to, 
https://github.com/keras-team/keras-applications/blob/master/keras_applications/resnet50.py

In a modularized way.

test/tf/gen_tf_pb.py
CharlieL7: Not sure I understand what you meant by hitting the `unsqueeze` lines.

test/tf/tf_test.cpp
CharlieL7: Simplified the tests with `add_common_op`

test/tf/gen_tf_pb.py
umangyadav: Never mind about `unsqueeze`.  But fp16, test is good to have.

test/verify/test_conv_bn.cpp
```diff
@@ -37,19 +38,52 @@ struct test_conv_bn : verify_program<test_conv_bn>
...
+        auto c_len   = bn_lens.at(1);
+        auto mb_mean = mm->add_instruction(
+            migraphx::make_op("multibroadcast", {{"out_lens", bn_lens}}), usq_mean);
+        auto numer  = mm->add_instruction(migraphx::make_op("sub"), conv, mb_mean);
+        auto mb_eps = mm->add_instruction(
+            migraphx::make_op("multibroadcast", {{"out_lens", {c_len, 1, 1}}}), eps);
+        auto var_eps = mm->add_instruction(migraphx::make_op("add"), usq_var, mb_eps);
+        auto mb_rt   = mm->add_instruction(
+            migraphx::make_op("multibroadcast", {{"out_lens", {c_len, 1, 1}}}), rt);
+        auto denom    = mm->add_instruction(migraphx::make_op("pow"), var_eps, mb_rt);
```
umangyadav, line 77: Use `add_common_op` similar to other test ?

src/driver/inceptionv3.cpp
```diff
@@ -35,2604 +36,1595 @@ migraphx::program inceptionv3(unsigned batch) // NOLINT(readability-function-siz
     migraphx::module_ref mmain = p.get_main_module();
     auto x_main_module_0       = mmain->add_literal(migraphx::abs(
         migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {1}}, 0)));
-    auto x_0                   = mmain->add_parameter(
-        "0", migraphx::shape{migraphx::shape::float_type, {batch, 3, 299, 299}});
+    auto x_input               = mmain->add_parameter(
```
umangyadav, line 39: Have you run each of these three driver models through `migraphx-driver verify` ?

CharlieL7: Looks like the models are good to go, no updates.

CharlieL7:
For the epsilon value I'm using the eps from this: https://www.tensorflow.org/api_docs/python/tf/raw_ops/FusedBatchNorm and this: https://www.tensorflow.org/api_docs/python/tf/raw_ops/FusedBatchNormV3. Since those are the TensorFlow ops used in the tests.

umangyadav (approved):
If you can add FP16 test, that would be good. Otherwise LGTM

PR #1372, opened 2022-09-06
Add MIGRAPHX_VALIDATE_MATCHES env variable to validate each matcher

pfultz2:


src/include/migraphx/matcher.hpp
```diff
@@ -371,7 +376,20 @@ void find_matches(Mod& mod, instruction_ref ins, Ms&&... ms)
                 std::cout << "Matched by " << get_type_name(m) << std::endl;
                 get_module(mod).debug_print(ins);
             }
+            // If its already invalid dont validate it again
+            bool invalidated = validate and get_module(mod).validate() != get_module(mod).end();
```
umangyadav, line 380: Why would it continue matching with other matchers if it previously generated an invalid module before from other matcher?

umangyadav: This would only find the first match that produced invalid instruction.

pfultz2: > This would only find the first match that produced invalid instruction.

Yes thats the matcher that caused the invalid instruction.

src/include/migraphx/matcher.hpp
```diff
@@ -356,7 +357,11 @@ void find_matches(Mod& mod, instruction_ref ins, Ms&&... ms)
     const
 #endif
         int trace = value_of(MIGRAPHX_TRACE_MATCHES{});
-    bool match    = false;
+#if !defined(__GNUC__) || defined(__clang__) || __GNUC__ > 5
+    const
+#endif
+        bool validate = enabled(MIGRAPHX_VALIDATE_MATCHES{});
```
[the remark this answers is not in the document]
pfultz2: There is too much output from the trace, the validation only prints when a matcher is doing something invalid.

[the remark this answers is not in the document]
TedThemistokleous: I think its better to keep this separate since itll be easier to toggle in our debug workflow.

TedThemistokleous (approved):

umangyadav (approved):

PR #1376, opened 2022-09-09
expose underlying migraphx::argument data pointer in pybind

shivadbhavsar:
Add python binding for the data method of migraphx::argument class. This is required by torch_migraphx to use the underlying data for creating torch tensors.


shivadbhavsar:
There is currently no use case for some of the other methods:
empty()
reshape(shape s)
copy()
share()

Do you guys think it may be worth exposing any of these through pybind as well with this change?

umangyadav:
> There is currently no use case for some of the other methods: empty() reshape(shape s) copy() share()
> 
> Do you guys think it may be worth exposing any of these through pybind as well with this change?

I think, we should add them as and when required based on use case.

umangyadav (approved):

pfultz2 (approved):

pfultz2:
> Do you guys think it may be worth exposing any of these through pybind as well with this change?

I dont think we will need those methods for now.

pfultz2 (changes requested):
Documentation needs updated.

pfultz2 (approved):

PR #1378, opened 2022-09-13
Add support in mlir for transposed and broadcasted shaped

pfultz2:


pfultz2:
So it looks like MLIR still doesn't work completely with broadcasts yet:

```
module {
  func @main(%arg0: tensor<1x1x1x1xf32>, %arg1: tensor<1x3x32x32xf32>, %arg2: tensor<1x3x5x5xf32>) -> tensor<1x1x28x28xf32> attributes {kernel = "mixr"} {
    %0 = migraphx.multibroadcast(%arg0) {out_lens = [1, 1, 28, 28]} : (tensor<1x1x1x1xf32>) -> tensor<1x1x28x28xf32>
    %1 = migraphx.convolution(%arg1, %arg2) {dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1], use_dynamic_same_auto_pad = 0 : i64} : (tensor<1x3x32x32xf32>, tensor<1x3x5x5xf32>) -> tensor<1x1x28x28xf32>
    %2 = migraphx.add(%1, %0) : (tensor<1x1x28x28xf32>, tensor<1x1x28x28xf32>) -> tensor<1x1x28x28xf32>
    %3 = migraphx.relu(%2) : (tensor<1x1x28x28xf32>) -> tensor<1x1x28x28xf32>
    return %3 : tensor<1x1x28x28xf32>
  }
}

error: Error constructing coordinate transformation: The transformation must define outputs
```

jungpark-mlir:
Fix for the above problem was merged yesterday, please try
6c9375a314da397b4161b5141a688c0eb166a12f

Or 5.4_rel once branched.

src/targets/gpu/mlir.cpp
```diff
@@ -589,8 +591,52 @@ std::string dump_mlir(const module& m)
...
+            auto perm  = find_permutation(input);
+            auto iperm = invert_permutation(perm);
+            lens       = reorder_dims(lens, iperm);
+            ops.push_back(make_op("transpose", {{"permutation", perm}}));
+        }
+        if(input.broadcasted())
+        {
+            std::transform(lens.begin(),
+                           lens.end(),
+                           input.strides().begin(),
```
jungpark-mlir, line 618: If input is transposed, don't we need to reorder strides as well?

Dockerfile
```diff
@@ -86,7 +86,7 @@ RUN git clone --single-branch --branch ${ONNXRUNTIME_BRANCH} --recursive ${ONNXR
 
 ADD tools/build_and_test_onnxrt.sh /onnxruntime/build_and_test_onnxrt.sh
 
-RUN cget -p /usr/local install ROCmSoftwarePlatform/llvm-project-mlir@d2cb9e580550e92ab75a0a417e7a4abd02a24edf -DBUILD_MIXR_TARGET=On
+RUN cget -p /usr/local install ROCmSoftwarePlatform/llvm-project-mlir@9cbe3ad5e5b2a1921de90b4f1c25d2218da18710 -DBUILD_MIXR_TARGET=On
```
jungpark-mlir, line 89: 5.4 has been branched
https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/tree/release/rocm-5.4
and its hash is 1252d5084784cb5c66c87c4770fc9e70b1743720

src/targets/gpu/mlir.cpp
jungpark-mlir: I mean, 
when it first goes into
  if(input.transposed())

lens is reordered as the given permutation.
And then if it goes into 
  if(input.broadcasted())

In std::transform, input.strides() needs to be reordered as well.

I'm not sure I understood correctly but please check this.

jungpark-mlir (changes requested):

src/targets/gpu/mlir.cpp
pfultz2: You are right this needs to be fixed.

sjw36 (approved):
Looks good.

jungpark-mlir (approved):

sjw36:
Failing with tidy errors.

src/targets/gpu/mlir.cpp
```diff
@@ -196,7 +196,6 @@ struct mlir_program
 
     MlirType make_tensor(const shape& s) const
     {
-        assert(s.standard());
```
pfultz2, line 197: This `assert` shouldnt be removed as we dont pass strides to the tensor and MLIR assumes standard shape.

PR #1379, opened 2022-09-13
Fix split_reshape for slice len of 1

umangyadav:
If the slice is of length 1 then, it would match with stride of 1 at the last dimension, which would cause reshape_lengths calculation to go out of  bounds in current logic. 
This fix checks for this case and calcualtes reshape axis accordingly.

### DIFF
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -985,32 +985,36 @@ struct find_split_reshape
         auto rsp_lens    = rsp->get_shape().lens();
         auto rsp_strides = rsp->get_shape().strides();
         rsp_strides.insert(rsp_strides.begin(), rsp_strides[0] * rsp_lens[0]);
-        
-        auto ait = std::find(rsp_strides.begin(), rsp_strides.end(), slc_dim_size);
+
+        auto ait     = std::find(rsp_strides.begin(), rsp_strides.end(), slc_dim_size);
         int rsp_axis = -1;
         if(ait == rsp_strides.end())
         {
             return;
-        } else if(ait == rsp_strides.end()-1) { 
+        }
+        else if(ait == rsp_strides.end() - 1)
+        {
             // edge case
-            // slice_dim == 1, in that case it would match with last stride of 1. 
-            // it should accumulate lengths from last dim in that case. discount 1 to avoid going out of bounds. 
-            rsp_axis = std::distance(rsp_strides.begin(), ait)-1;
-        } else {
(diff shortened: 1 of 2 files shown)


src/simplify_algebra.cpp
```diff
@@ -985,20 +985,34 @@ struct find_split_reshape
...
         if(ait == rsp_strides.end())
         {
             return;
         }
-        int rsp_axis = std::distance(rsp_strides.begin(), ait);
-
+        else if(ait == rsp_strides.end() - 1)
+        {
+            // edge case
+            // slice_dim == 1, in that case it could match with last stride of 1.
```
pfultz2, line 998: But you are not checking that the dimension is `1`. Can you add a test with slice and the last dimension is not `1`?

src/simplify_algebra.cpp
```diff
@@ -985,20 +985,34 @@ struct find_split_reshape
...
+            // it should accumulate lengths from last dim in that case. discount 1 to avoid going
+            // out of bounds.
```
pfultz2, line 1000: Where would we go out of bounds?

umangyadav: e.g.  slice output of lens `[1024, 128, 1]`

reshape output shape = `[1024, 128][128, 1]`
`slice_dim_size = 1`, 
`auto rsp_strides = rsp->get_shape().strides();`
`rsp_strides.insert(rsp_strides.begin(), rsp_strides[0] * rsp_lens[0]);` --> `rsp_strides = [1024 * 128, 128, 1]`
` auto ait     = std::find(rsp_strides.begin(), rsp_strides.end(), slc_dim_size);` ---> `ait = 2`. 
`rsp_axis = std::distance(rsp_strides.begin(), ait);` ---> `rsp_axis = 2`. 

Now, inside the following, it would go out of bounds since `lens()` would only be of length 2. 
`std::transform(vec_rsp.begin(), vec_rsp.end(), vec_dims.begin(), [&](auto is) {
            return is->get_shape().lens()[rsp_axis];
        });`

src/simplify_algebra.cpp
umangyadav: > Can you add a test with slice and the last dimension is not 1? 
Last dimension of which instruction?  
Almost all the existing tests are taking slices in the last dimension with `slice_length != 1`

pfultz2: Yea lets add an `assert`.

pfultz2 (approved):

shivadbhavsar (approved):

PR #1380, opened 2022-09-13
Refactor memory coloring pass

pfultz2:
This refactors memory coloring to only handle allocation instructions. It also handles allocations for tuple shapes.


src/CMakeLists.txt
```diff
@@ -57,14 +57,13 @@ add_library(migraphx
...
     make_op.cpp
+    memory_coloring.cpp
     module.cpp
     msgpack.cpp
     normalize_attributes.cpp
     normalize_ops.cpp
     op_enums.cpp
     operation.cpp
-    opt/memory_coloring.cpp
-    opt/memory_coloring_impl.cpp
```
umangyadav, line 67: These files needs to be deleted/removed.

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+                [&](auto child) { return *as.get_segment(child); });
+
+            // Get the segment for the parent
+            const auto* parent_segment = as.get_segment(parent);
+            // Add segment for the parent if there is none or segment overlaps with the children
+            if(parent_segment == nullptr or overlaps(segments, *parent_segment))
+                as.add_segment(parent, next_segment(segments, parent, alignment));
+        }
+        // Reduce the number of segments
+        for(std::size_t n = 0; n < 3; n++)
```
umangyadav, line 270: count of 3, is it arbitrary ?

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+            {
+                auto s2 = as.ins2segment.at(ins);
+                std::cout << s2.first << ", " << s2.second << ": ";
+                m.debug_print(ins);
+            }
+        }
+    }
+
+    // Total memory
+    std::size_t n = as.max() * alignment;
```
umangyadav, line 351: what if total required memory is out of bounds for given GPU ?

umangyadav: segmenting based on max_alignment wouldn't that lead to more fragmentation?  when there is only one instance of vectorized 4 and rest of the lengths are  not multiple of 4 or 2 (non-vectorized) ?

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+}
+
+struct allocation_segment
+{
+    using segment = std::pair<std::size_t, std::size_t>;
+    std::unordered_map<instruction_ref, segment> ins2segment;
+
+    const segment* add_segment(instruction_ref ins, segment s)
+    {
+        // this->remove(ins);
```
TedThemistokleous, line 106: remove commented code

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+        if(n > 4)
+        {
+            auto d = n % 4;
+            if(d == 0)
+                alignment *= 4;
+            if(d == 2)
+                alignment *= 2;
+        }
+        return alignment;
+    }
```
TedThemistokleous, line 197: Don't we want to handle the odd cases, there an assumption we add padding then?

 2, and 4 are powers of two, but are we setting the boundaries for this or is that dependent on how we read each block of memory?

test/memory_coloring_test.cpp
```diff
@@ -729,7 +729,7 @@ TEST_CASE(test39)
         run_pass(*smod);
     }
 
-    CHECK(mm->get_parameter_shape("scratch").bytes() == 4);
+    CHECK(mm->get_parameter_shape("scratch").bytes() == 1);
```
TedThemistokleous, line 732: Guessing this was changed because of the change to alignment then?

test/memory_coloring_test.cpp
```diff
@@ -3374,7 +3374,7 @@ TEST_CASE(rnn_dom)
     m.add_instruction(pass_op{}, moutput, mx250, mx249, mx248);
 
     run_pass(m);
-    CHECK(m.get_parameter_shape("scratch").bytes() == 1600);
+    CHECK(m.get_parameter_shape("scratch").bytes() == 1824); // Optimal is 1600
```
TedThemistokleous, line 3377: If 1600 is optimal, why are we now different? Is this a regression or needed due to the additional alignment/padding we've added?

TedThemistokleous (commented):
Few questions on this one. Do we have a design of the original design vs this refactor? Was there a perf boost with this change?

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+    // Replace zero allocation
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != allocation_op)
+            continue;
+        assert(ins->get_shape().bytes() == 0);
+        m.replace_instruction(ins, op::load{ins->get_shape(), 0}, mem);
+    }
+
+    // Remove scratch parameter if its not used
```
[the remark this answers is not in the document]
pfultz2: This matches the original behavior, there was some issue with changing it.

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+        m.replace_instruction(ins, op::load{s, offset}, mem);
+    }
+
+    // Replace zero allocation
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != allocation_op)
+            continue;
+        assert(ins->get_shape().bytes() == 0);
+        m.replace_instruction(ins, op::load{ins->get_shape(), 0}, mem);
```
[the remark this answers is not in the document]
pfultz2: The workspace for convolution has 0 byte allocations.

src/CMakeLists.txt
pfultz2: Yes, I plan to remove later to avoid merge conflicts.

src/memory_coloring.cpp
pfultz2: Yea, we loop three time, it seems to improve memory usage.

src/memory_coloring.cpp
pfultz2: All segments are multiples of the alignment otherwise we would need to deal with alignment when we are doing graph coloring.

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+    void remove(instruction_ref ins)
+    {
+        auto it = ins2segment.find(ins);
+        if(it != ins2segment.end())
+        {
+            ins2segment.erase(it);
+        }
+    }
+
+    std::size_t max()
```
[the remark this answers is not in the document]
pfultz2: I dont understand why you would want to put `end` on the name. This is to get the largest segment size, which corresponds to the max memory needed.

src/memory_coloring.cpp
pfultz2: Vectorized sizes that are odd are not very common, and we currently do not vectorize these sizes.

src/memory_coloring.cpp
```diff
@@ -0,0 +1,381 @@
...
+                    children.begin(),
+                    children.end(),
+                    std::inserter(segments, segments.begin()),
+                    [&](auto child) { return as.get_segment(child); },
+                    [&](auto child) { return *as.get_segment(child); });
+                // Get the segment for the parent
+                const auto* parent_segment = as.get_segment(parent);
+                assert(parent_segment != nullptr);
+
+                auto s = next_segment(segments, parent, alignment);
```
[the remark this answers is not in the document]
pfultz2: The alignment has to be the same for all segments.

test/memory_coloring_test.cpp
pfultz2: yes

test/memory_coloring_test.cpp
pfultz2: Graph coloring is NP-complete so it isnt always possible to get the best result. It depends on the order we process the nodes. This case uses more memory but other cases we use less memory.

pfultz2:
> Do we have a design of the original design vs this refactor?

There wasn't any design docs on the original. This is similar to register allocation in compilers if you are familiar with that(see [here](http://web.cecs.pdx.edu/~mperkows/temp/register-allocation.pdf) for an overview). This does liveness analysis to create a inference graph and then we do graph coloring on the inference graph. However, instead of using simple colors we are using segments.

I dont know really know exactly how the original one worked in detail. It also does some things that didnt really make sense(like why is it checking for param or check_context).

This simplifies what we were doing by using the same alignment for all segments. Most models use the same data type for 99% of its calculations. It also applies a loop afterward to try to adjust the graph to get better optimization as well.

This can also handle tuple types whereas the original cant.

> Was there a perf boost with this change?

This is a memory optimization not necessarily a perf optimization, but it does run faster during compile phase. For many models like inception, bert, etc there is less memory used as well. We may be able to tweak this easier to get better memory usage in the future.

src/memory_coloring.cpp
TedThemistokleous: That makes sense.

src/memory_coloring.cpp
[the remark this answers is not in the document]
TedThemistokleous: So it's done inplace then or is the allocation doing something to another state somewhere else?

src/memory_coloring.cpp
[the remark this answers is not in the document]
umangyadav: segment is `(start, end)` this returns max of all segments "ends", not necessarily largest or maximum segment size.

test/memory_coloring_test.cpp
TedThemistokleous: I don't think ordering of how the nodes are processed should matter here. You can reduce this into a decision problem + binary search to get the min amount of colors but I think the trade may be space and additional complexity.

Maybe as a later improvement but we can reduce the amount of colors if we use a SAT solver to determine the min amount of colors before finding the configuration of those colors. 

Like there's libraries like http://minisat.se/, that are under MIT license I've used to quickly find if the number of colors chosen there exists a solution, then again we'd need to represent things a certain way for the solver to work, and then use something like a binary search until we converged. 

I've got code for this I can dig up if this is something you want to explore further.

src/memory_coloring.cpp
umangyadav: okay, just by looking at the code, it seems loop is guaranteed to converge after a certain iterations (i could be wrong), if that is the case then,  we can do a while loop with some exit condition.

src/memory_coloring.cpp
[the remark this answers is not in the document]
umangyadav: Yes, I understand that, but I think this pass trying to do two things at once, which are alignment and memory allocation, which may not result in optimal allocation in terms of fragmentation. I was  thinking if it would be possible to just do (1-byte) alignment in this pass and adjust those alignments later in another pass separating two concerns.

TedThemistokleous:
> > Do we have a design of the original design vs this refactor?
> 
> There wasn't any design docs on the original. This is similar to register allocation in compilers if you are familiar with that(see [here](http://web.cecs.pdx.edu/~mperkows/temp/register-allocation.pdf) for an overview). This does liveness analysis to create a inference graph and then we do graph coloring on the inference graph. However, instead of using simple colors we are using segments.
> 
> I dont know really know exactly how the original one worked in detail. It also does some things that didnt really make sense(like why is it checking for param or check_context).

Okay, seems like we're trying to preserve state somewhere then.

> This simplifies what we were doing by using the same alignment for all segments. Most models use the same data type for 99% of its calculations. It also applies a loop afterward to try to adjust the graph to get better optimization as well.
> 
> This can also handle tuple types whereas the original cant.

Got it, so the added benefit of the additional type would be helpful. 

> > Was there a perf boost with this change?
> 
> This is a memory optimization not necessarily a perf optimization, but it does run faster during compile phase. For many models like inception, bert, etc there is less memory used as well. We may be able to tweak this easier to get better memory usage in the future.

Does our driver code have a way to measure compile time before an inference?

pfultz2:
> I still think we can always find the minimum set of colors possible for the graph without relying on spillover and statistics.

Well graph coloring is an NP-hard problem. I hadn't written a piece of code of iterate over all possible graph colorings at one point to find the minimum. It was able to finish for simple networks like resnet50, but for like RNNs it ran for months and still hadn't finished(I eventually just stopped it). 

> Does our driver code have a way to measure compile time before an inference? More for my current curiosity.

Thats why I added `MIGRAPHX_TIME_PASS` env variable to see the time for different passes here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1380/files#diff-0b6c8d83337d2bc89577573dcd7298fcbc76f205106bbce3993d522ff841ee83R100

TedThemistokleous:
On this

> > I still think we can always find the minimum set of colors possible for the graph without relying on spillover and statistics.
> 
> Well graph coloring is an NP-hard problem. I hadn't written a piece of code of iterate over all possible graph colorings at one point to find the minimum. It was able to finish for simple networks like resnet50, but for like RNNs it ran for months and still hadn't finished(I eventually just stopped it).

You wouldn't check every combination if you use a sat solver though. If you tie it to a binary search of "how many colors" where k is the number of colours, and start at something like  k = half the number of nodes, you can converge really quickly, by converting the search/check every combination into a simple decision problem.

I have code I did for my degree on this exact problem we can probably reuse to test this, and its why I was asking about stuff in our driver/toolchain to time this result. The code I wrote, generates a random graph, then finds the minimum amount of colors needed for said graph to get a proper valid coloring. There may be some tweaking needed for graph coloring with RNNs since I think we'll get a lot more self loops and other oddities but from my understanding these should simplify out.

Probably out of scope for this PR but this piqued my curiosity on the whole thing. Didn't realize we could use heuristics to solve this though.

test/memory_coloring_test.cpp
pfultz2: > I don't think ordering of how the nodes are processed should matter here.

Since we are using a greedy algorithm for coloring the order does matter. From Wikipedia's [Greedy Coloring](https://en.wikipedia.org/wiki/Greedy_coloring#Choice_of_ordering):

> Different orderings of the vertices of a graph may cause the greedy coloring to use different numbers of colors, ranging from the optimal number of colors to, in some cases, a number of colors that is proportional to the number of vertices in the graph. For instance, a [crown graph](https://en.wikipedia.org/wiki/Crown_graph) (a graph formed from two [disjoint sets](https://en.wikipedia.org/wiki/Disjoint_sets) of n/2 vertices {a1, a2, ...} and {b1, b2, ...} by connecting ai to bj whenever i ≠ j) can be a particularly bad case for greedy coloring. With the vertex ordering a1, b1, a2, b2, ..., a greedy coloring will use n/2 colors, one color for each pair (ai, bi). However, the optimal number of colors for this graph is two, one color for the vertices ai and another for the vertices bi.[[5]](https://en.wikipedia.org/wiki/Greedy_coloring#cite_note-5) There also exist graphs such that with high probability a randomly chosen vertex ordering leads to a number of colors much larger than the minimum.[[6]](https://en.wikipedia.org/wiki/Greedy_coloring#cite_note-6) Therefore, it is of some importance in greedy coloring to choose the vertex ordering carefully.

There are some adaptive orderings that could improve on what we are currently doing.

src/memory_coloring.cpp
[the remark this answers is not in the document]
pfultz2: >  but I think this pass trying to do two things at once, which are alignment and memory allocation

I dont see how you can seperate these two.

> which may not result in optimal allocation in terms of fragmentation

What do you mean? The only way to possibly get better memory for alignment is to use a different alignment for every segment. Since most models all use the same data type, I dont see the reason to introduce this complexity.

> I was thinking if it would be possible to just do (1-byte) alignment in this pass and adjust those alignments later in another pass separating two concerns.

This definitely wont work. If we change alignment then we could create overlaps of the segments since we need to shift the segments forward or backwards to be aligned. 

Also, segments are part of the internal data structures of memory coloring so I dont see how a later pass could adjust that, since it wont have access to it.

src/memory_coloring.cpp
pfultz2: It wont converge. Its NP-Hard so we wont know if its correct unless we try all options.

src/memory_coloring.cpp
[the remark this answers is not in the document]
pfultz2: Its the maximum memory needed to support all segments, the fact that it checks the `end` is an implementation detail.

test/memory_coloring_test.cpp
pfultz2: > if we use a SAT solver to determine the min amount of colors before finding the configuration of those colors.

That can tell use the minimum memory possible, but can it tell us the start and end of each of the segments as well?

src/memory_coloring.cpp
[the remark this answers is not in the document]
pfultz2: > So it's done inplace then or is the allocation doing something to another state somewhere else?

This is for the workspace needed for convolutions in MIOpen: https://rocmsoftwareplatform.github.io/MIOpen/doc/html/convolution.html#miopenconvolutionforward

pfultz2:
> You wouldn't check every combination if you use a sat solver though. If you tie it to a binary search of "how many colors" where k is the number of colours, and start at something like k = half the number of nodes, you can converge really quickly, by converting the search/check every combination into a simple decision problem.

Well we dont use colors, we use segments. Colors dont allow smaller allocations to reuse larger allocations. If you have three allocations: 4, 4, 8. The size 4s cannot reuse each memory and the 8 can reuse any memory. Then the two 4s can reuse the 8, requiring only 8 max memory. Whereas if they are assigned as seperate colors you would need 12 max memory.

src/memory_coloring.cpp
```diff
@@ -0,0 +1,377 @@
...
+        };
+        add_live_variables(ins->inputs());
+        add_live_variables(implicit_deps[ins]);
+        // Remove last usage
+        auto it = live_set.find(ins);
+        if(it != live_set.end())
+        {
+            live_set.erase(it);
+            f(ins, live_set);
+        }
```
kahmed10, line 55:
```suggestion
        if(contains(live_set, ins));
        {
            live_set.erase(it);
            f(ins, live_set);
        }
```

src/pass_manager.cpp
```diff
@@ -101,7 +102,16 @@ struct module_pm : module_pass_manager
         trace("Module: ", mod->name(), ", Pass: ", p.name());
         const double t1 = ts.record<seconds>();
         assert(mod->validate() == mod->end());
-        p.apply(*this);
+        if(enabled(MIGRAPHX_TIME_PASSES{}))
```
kahmed10, line 105: we probably want to consolidate this and update the previous timer code.

src/memory_coloring.cpp
pfultz2: I dont see how that change could compile.

src/pass_manager.cpp
pfultz2: I see, I will remove the extra timer code.

src/memory_coloring.cpp
kahmed10: Yeah you're right. I missed the erasing of `it`.

kahmed10: Actually you can erase by key, and this compiles. I think it should be equivalent (and cleaner) now?

pfultz2: > I think it should be equivalent (and cleaner) now?

But then it will search the map twice for the key whereas this only searches it once.

src/memory_coloring.cpp
```diff
@@ -0,0 +1,416 @@
...
+            // Add all segments for the children to the segments already processed
+            transform_if(
+                children.begin(),
+                children.end(),
+                std::inserter(segments, segments.begin()),
+                [&](auto child) { return as.get_segment(child); },
+                [&](auto child) { return *as.get_segment(child); });
+
+            // Get the segment for the parent
+            const auto* parent_segment = as.get_segment(parent);
```
umangyadav, line 299: During the first iteration, this is going to be nullptr always.

src/memory_coloring.cpp
```diff
@@ -0,0 +1,416 @@
...
+        // Add all allocations to the conflict_queue
+        std::transform(conflict_table.begin(),
+                       conflict_table.end(),
+                       std::back_inserter(conflict_queue),
+                       [](auto&& pp) { return pp.first; });
+
+        auto alloc_index = create_allocation_index(m, conflict_table);
+
+        // Sort the conflict queue so we process the allocation with the least
+        // number of adjacent allocations first
```
umangyadav, line 272: Most conflicting allocation is first in the queue. 
We should say that instead of least number of adjacent allocations first which sounds the opposite.

src/memory_coloring.cpp
umangyadav: > what if total required memory is out of bounds for given GPU ? 

Can we check for this ?

src/memory_coloring.cpp
```diff
@@ -0,0 +1,416 @@
...
+            if(i == ins)
+                continue;
+            // Skip variables that aren't allocations
+            if(i->name() != allocation_op)
+                continue;
+            // Skip zero allocations
+            if(i->get_shape().bytes() == 0)
+                continue;
+            conflict_table[i].insert(ins);
+            conflict_table[ins].insert(i);
```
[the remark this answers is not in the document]
pfultz2: This pass is ran on all modules not just the main module. See the output from the test_loop:

```
module: "main"
main:@0 = check_context::migraphx::version_1::gpu::context -> float_type, {}, {}
main:@1 = hip::hip_allocate_memory[shape=int8_type, {32}, {1},id=main:scratch] -> int8_type, {32}, {1}
ccond = @param:ccond -> bool_type, {1}, {0}
main:@3 = hip::copy_from_gpu(ccond) -> bool_type, {1}, {0}
main:@4 = hip::hip_copy_literal[id=main:@literal:0] -> int64_type, {1}, {1}
main:@5 = hip::hip_copy_literal[id=main:@literal:1] -> int64_type, {1}, {0}
main:@6 = hip::copy_from_gpu(main:@5) -> int64_type, {1}, {0}
main:@7 = hip::sync_stream(main:@6,main:@3) -> int64_type, {1}, {0}
main:@8 = load[offset=24,end=25](main:@1) -> bool_type, {1}, {0}
main:#output_1 = @param:main:#output_1 -> [int64_type, {1}, {1}, int64_type, {8, 1}, {1, 1}]
main:@10 = load[offset=8,end=16](main:@1) -> int64_type, {1}, {1}
main:@11 = load[offset=16,end=17](main:@1) -> bool_type, {1}, {0}
main:@12 = load[offset=0,end=8](main:@1) -> int64_type, {1}, {0}
main:@13 = gpu::loop[max_iterations=8](main:@7,main:@3,main:@4,main:@12,main:@11,main:@10,main:@8,main:#output_1), [loop_module] -> [int64_type, {1}, {1}, int64_type, {8, 1}, {1, 1}]
main:@14 = get_tuple_elem[index=0](main:@13) -> int64_type, {1}, {1}
main:@15 = get_tuple_elem[index=1](main:@13) -> int64_type, {8, 1}, {1, 1}
main:@16 = @return(main:@14,main:@15)

module: "loop_module"
loop_module:@0 = check_context::migraphx::version_1::gpu::context -> float_type, {}, {}
loop_module:@1 = hip::hip_allocate_memory[shape=int8_type, {4}, {1},id=loop_module:scratch] -> int8_type, {4}, {1}
cond = @param:cond -> bool_type, {1}, {0}
loop_module:@3 = load[offset=0,end=4](loop_module:@1) -> float_type, {1}, {0}
iter_num = @param:iter_num -> int64_type, {1}, {0}
loop_module:@5 = gpu::code_object[code_object=13632,symbol_name=convert_kernel,global=1,local=1024,](iter_num,loop_module:@3) -> float_type, {1}, {0}
loop_module:#output_2 = @param:loop_module:#output_2 -> int64_type, {1}, {1}
input = @param:input -> int64_type, {1}, {1}
loop_module:@8 = gpu::code_object[code_object=13664,symbol_name=add_convert_add_convert_kernel,global=1,local=1024,](loop_module:@5,input,loop_module:#output_2) -> int64_type, {1}, {1}
loop_module:#output_0 = @param:loop_module:#output_0 -> bool_type, {1}, {0}
loop_module:@10 = gpu::code_object[code_object=13664,symbol_name=equal_convert_convert_not_kernel,global=1,local=1024,](loop_module:@5,loop_module:#output_0) -> bool_type, {1}, {0}
loop_module:@11 = @return(loop_module:@10,loop_module:@8,loop_module:@8)
```

src/memory_coloring.cpp
```diff
@@ -0,0 +1,416 @@
...
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DEBUG_MEMORY_COLORING);
+
+using instruction_set     = std::unordered_set<instruction_ref>;
+using instruction_set_map = std::unordered_map<instruction_ref, instruction_set>;
+
+// This will do liveness analysis on the module, and it will call the
+// function `f` with the instruction and the set of the other instructions
+// that are live
+template <class F>
+void liveness(const module& m, F f)
```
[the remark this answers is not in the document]
pfultz2: We could move it to a `liveness.hpp` header(everyone who uses `module` doesnt need a liveness function), but I think we can do that when other passes need to use this function. Right now, its only used by memory_coloring,

src/memory_coloring.cpp
pfultz2: > what if total required memory is out of bounds for given GPU ?

We allocate pinned memory if the buffer is too large.

> Can we check for this ?

Well memory_coloring is used for both GPU and CPU(and possibly other backends). To support this we could need to create some kind of allocation model that can have function for max memory, scratch space allocation(on CPU or disk), and functions to copy to/from scratch space. Then when the memory overflows we can handle this gracefully.

We could handle this in a future PR.

src/memory_coloring.cpp
umangyadav: We can remove this assignment since this is going to nullptr always.

umangyadav (approved):

src/memory_coloring.cpp
umangyadav: Would be good to have an issue to track this.

kahmed10 (approved):

src/memory_coloring.cpp
```diff
@@ -0,0 +1,405 @@
...
+        });
+    }));
+
+    // Print out segments
+    if(enabled(MIGRAPHX_DEBUG_MEMORY_COLORING{}))
+    {
+        for(auto&& pp : conflict_table)
+        {
+            std::cout << "------- conflict -------" << std::endl;
+            auto s1 = as.ins2segment.at(pp.first);
```
TedThemistokleous, line 362: Add a test for the conflict_table

src/memory_coloring.cpp
```diff
@@ -0,0 +1,416 @@
...
+        // The base iterator is one ahead, so we need to use the previous iterator
+        auto ins = std::prev(rins.base());
+        // Add live variables
+        auto add_live_variables = [&](const auto& inputs) {
+            for(auto input : inputs)
+            {
+                auto i = instruction::get_output_alias(input);
+                // Skip if variable comes from parent
+                if(not m.has_instruction(i))
+                    continue;
```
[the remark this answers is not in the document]
umangyadav: here

PR #1384, opened 2022-09-14
Multibroadcast find_mul_conv

CharlieL7:
Change find_mul_conv to work with multibroadcast also. Checks the strides instead of the broadcast axis.

### DIFF
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -84,24 +84,26 @@ struct find_mul_conv
         // check broadcasted along channels
         auto a_lens    = a_ins->get_shape().lens();
         auto a_strides = a_ins->get_shape().strides();
-        // handle case of stride = 1 and len = 1
+        // handle len = 1 case
         auto invalid_sl = [&](std::size_t i) {
             if(a_strides[i] != 0)
             {
-                if(a_strides[i] == 1 and a_lens[i] == 1)
+                if(a_lens[i] == 1)
+                {
                     return false;
+                }
                 return true;
             }
             return false;
         };
-        auto check = false;
+        auto invalid_case = false;
         for(int i = 2; i < a_lens.size(); ++i)
         {
             if(invalid_sl(i))
-                check = true;
+                invalid_case = true;
         }
 
-        if(invalid_sl(0) or a_strides.at(1) != 1 or check)
+        if(invalid_sl(0) or a_strides.at(1) != 1 or invalid_case)
             return;
 
         auto sq    = m.insert_instruction(ins, make_op("squeeze"), a_ins->inputs().front());
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -74,15 +74,11 @@ struct find_mul_conv
         auto a_ins    = r.instructions["a"];
         auto w_ins    = r.instructions["w"];
 
-        // check that the channels stride is 1
         auto a_shape   = a_ins->get_shape();
-        auto a_lens    = a_shape.lens();
         auto a_strides = a_shape.strides();
-        if(a_strides.at(1) != 1)
-            return;
-        else if(a_strides.at(0) != 0 or std::any_of(a_strides.cbegin() + 2,
-                                                    a_strides.cend(),
-                                                    [](std::size_t i) { return i != 0; }))
+        if(a_strides.at(0) != 0 or a_strides.at(1) != 1 or
+           std::any_of(
+               a_strides.cbegin() + 2, a_strides.cend(), [](std::size_t i) { return i != 0; }))
             return;
 
         auto sq    = m.insert_instruction(ins, make_op("squeeze"), a_ins->inputs().front());
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -74,10 +74,34 @@ struct find_mul_conv
         auto a_ins    = r.instructions["a"];
         auto w_ins    = r.instructions["w"];
 
+        auto a_input_lens = a_ins->inputs().front()->get_shape().lens();
+
+        std::size_t num_not_one_dims = std::count_if(
+            a_input_lens.cbegin(), a_input_lens.cend(), [](auto dim) { return dim != 1; });
+        if(num_not_one_dims > 1)
+            return;
+
+        // check broadcasted along channels
+        auto a_lens    = a_ins->get_shape().lens();
         auto a_strides = a_ins->get_shape().strides();
-        if(a_strides.at(0) != 0 or a_strides.at(1) != 1 or
-           std::any_of(
-               a_strides.cbegin() + 2, a_strides.cend(), [](std::size_t i) { return i != 0; }))
+        // handle case of stride = 1 and len = 1
+        auto invalid_sl = [&](std::size_t i) {
+            if(a_strides[i] != 0)
+            {
+                if(a_strides[i] == 1 and a_lens[i] == 1)
+                    return false;
+                return true;
+            }
+            return false;
+        };
+        auto check = false;
+        for(int i = 2; i < a_lens.size(); ++i)
+        {
+            if(invalid_sl(i))
+                check = true;
+        }
+
+        if(invalid_sl(0) or a_strides.at(1) != 1 or check)
             return;
 
         auto sq    = m.insert_instruction(ins, make_op("squeeze"), a_ins->inputs().front());
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -57,12 +57,14 @@ auto conv_const_weights()
 
 auto reduction() { return match::name_contains("reduce"); }
 
+// conv(x, w) * a => conv(x, a * w)
 struct find_mul_conv
 {
     auto matcher() const
     {
-        return match::name("mul")(match::either_arg(0, 1)(conv_const_weights().bind("conv"),
-                                                          match::name("broadcast").bind("a")));
+        return match::name("mul")(
+            match::either_arg(0, 1)(conv_const_weights().bind("conv"),
+                                    match::name("broadcast", "multibroadcast").bind("a")));
     }
 
     void apply(module& m, const match::matcher_result& r) const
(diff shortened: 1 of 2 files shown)


src/simplify_algebra.cpp
```diff
@@ -72,14 +74,15 @@ struct find_mul_conv
...
+           std::any_of(
+               a_strides.cbegin() + 2, a_strides.cend(), [](std::size_t i) { return i != 0; }))
             return;
 
+        auto sq    = m.insert_instruction(ins, make_op("squeeze"), a_ins->inputs().front());
         auto new_a = m.insert_instruction(
-            ins,
-            make_op("broadcast", {{"axis", 0}, {"out_lens", w_ins->get_shape().lens()}}),
-            a_ins->inputs().front());
+            ins, make_op("broadcast", {{"axis", 0}, {"out_lens", w_ins->get_shape().lens()}}), sq);
```
umangyadav, line 85: Is it possible to use `multibroadcast` here ? `broadcast` is deprecated.

test/simplify_algebra_test.cpp
```diff
@@ -236,6 +236,30 @@ TEST_CASE(simplify_mul_conv1)
...
+    auto x = m.add_parameter("x", {migraphx::shape::int32_type, {1, 128, 28, 28}});
+    auto w =
+        m.add_literal(migraphx::generate_literal({migraphx::shape::int32_type, {256, 128, 3, 3}}));
+    auto conv = m.add_instruction(
+        migraphx::make_op("convolution",
+                          {{"padding", {1, 1}}, {"stride", {2, 2}}, {"dilation", {1, 1}}}),
+        x,
+        w);
+    auto a      = m.add_literal(migraphx::generate_literal({migraphx::shape::int32_type, {256}}));
+    auto unsq_a = m.add_instruction(migraphx::make_op("unsqueeze", {{"axes", {1, 2}}}), a);
```
umangyadav, line 251: I think multibroadcast can do unsqueeze itself.

src/simplify_algebra.cpp
```diff
@@ -72,14 +74,15 @@ struct find_mul_conv
...
 
-        auto broadcast_op = any_cast<op::broadcast>(a_ins->get_operator());
-        if(broadcast_op.axis != 1)
+        auto a_strides = a_ins->get_shape().strides();
+        if(a_strides.at(0) != 0 or a_strides.at(1) != 1 or
+           std::any_of(
+               a_strides.cbegin() + 2, a_strides.cend(), [](std::size_t i) { return i != 0; }))
             return;
 
+        auto sq    = m.insert_instruction(ins, make_op("squeeze"), a_ins->inputs().front());
```
umangyadav, line 83: squeeze would only remove axes with dims 1.  it is possible that axis has strides 0 but  dims != 1.

test/simplify_algebra_test.cpp
umangyadav: We can keep this, I think it tests the `squeeze` line in the matcher.

src/simplify_algebra.cpp
CharlieL7: Possible but would need to squeeze and then unsqueeze to something like a [C, 1, 1] shape to use multibroadcast. Paul mentioned that it might be better to combine `broadcast` and `multibroadcast` into one operation.

test/simplify_algebra_test.cpp
CharlieL7: I tried this out when rewriting batch_norm, it doesn't work without the unsqueeze. And AFAIK if we follow the numpy broadcasting rules it would try to broadcast to the last dimension rather than the channels (2nd dimension).

src/simplify_algebra.cpp
CharlieL7: Let me make sure I understand. The squeeze is applying on the input to previous `broadcast` or `multibroadcast` instruction. It's a `nop` for the `broadcast` case. For `multibroadcast` the `squeeze` should get rid of the single-dimensional entries. But there might be a problem when this gets matched for a case like this:
```
x = {256, 14, 14}, {1, 0, 0}
a = multibroadcast(x) => {4, 256, 14, 14}, {0, 1, 0, 0}
c = conv(x, w)
mul(c, a)
```
For this case the squeeze and broadcast won't work.

CharlieL7: I made some tests for this case, looks like the code on develop also fails for this case.

umangyadav: > x = {256, 14, 14}, {1, 0, 0}
a = multibroadcast(x) => {4, 256, 14, 14}, {0, 1, 0, 0}
c = conv(x, w)
mul(c, a)

I was thinking about similar case. In this case, `x` itself is broadcasted, you can keep traversing back in graph (with some checks) to find real input. 

We should keep some check (`squeeze_out.shape.lens().size() ==1 ?` )at least to avoid applying matcher in such case rather than failing at broadcast.

src/simplify_algebra.cpp
```diff
@@ -72,14 +74,39 @@ struct find_mul_conv
...
+                return true;
+            }
+            return false;
+        };
+        auto check = false;
+        for(int i = 2; i < a_lens.size(); ++i)
+        {
+            if(invalid_sl(i))
+                check = true;
+        }
```
CharlieL7, line 102: This should handle the cases discussed. The code I made is somewhat messy though.

src/simplify_algebra.cpp
```diff
@@ -72,14 +74,39 @@ struct find_mul_conv
...
+        // handle case of stride = 1 and len = 1
+        auto invalid_sl = [&](std::size_t i) {
+            if(a_strides[i] != 0)
+            {
+                if(a_strides[i] == 1 and a_lens[i] == 1)
+                    return false;
+                return true;
+            }
+            return false;
+        };
```
umangyadav, line 96:
```suggestion
        auto invalid_sl = [&](std::size_t i) {
            if(a_strides[i] > 1)
            {  
                return true;
            } else if(a_strides[i] == 0) { 
              if(a_lens[i] !=1) 
                  return true;
            } else {
            return false;  
            } 
            return false;
        };
```
Should it be something like this ?

umangyadav: I am not sure if it is possible but if there is case like `a_lens = [256, 1, 1]`, `a_strides = [256, 1 or 0, 1 or 0]`

src/simplify_algebra.cpp
```diff
@@ -72,14 +74,39 @@ struct find_mul_conv
...
+        auto check = false;
+        for(int i = 2; i < a_lens.size(); ++i)
+        {
+            if(invalid_sl(i))
+                check = true;
+        }
+
+        if(invalid_sl(0) or a_strides.at(1) != 1 or check)
```
umangyadav, line 104: rename `check` to `invalid_case` or `skip_match` ?

test/simplify_algebra_test.cpp
```diff
@@ -236,6 +236,105 @@ TEST_CASE(simplify_mul_conv1)
...
+    migraphx::module m;
+    auto x = m.add_parameter("x", {migraphx::shape::int32_type, {1, 128, 28, 28}});
+    auto w =
+        m.add_literal(migraphx::generate_literal({migraphx::shape::int32_type, {256, 128, 3, 3}}));
+    auto conv = m.add_instruction(
+        migraphx::make_op("convolution",
+                          {{"padding", {1, 1}}, {"stride", {2, 2}}, {"dilation", {1, 1}}}),
+        x,
+        w);
+    auto a = m.add_literal(
```
umangyadav, line 275: would this work if `a` had a shape with batch dims `1` ? e.g. `[1, 256]` ?

CharlieL7: Current code would skip because strides[0] is not 0 or 1.

src/simplify_algebra.cpp
umangyadav: > That case would be skipped in the current code because a_strides[0] is not 0 or 1. The tensor would also have pretty weird data, as it would not be contiguous.

Sorry, I meant to say this case : `a_lens = [256, 1, 1], a_strides = [1, 1 or 0, 1 or 0]`

umangyadav: > . I guess it is possible for a length 1 dimension to have any stride and still be valid? 

Yes i think, the broadcast would work on that.

test/simplify_algebra_test.cpp
CharlieL7: Updated to allow any stride for len=1.

test/simplify_algebra_test.cpp
```diff
@@ -236,6 +236,105 @@ TEST_CASE(simplify_mul_conv1)
...
+        migraphx::generate_literal({migraphx::shape::int32_type, {256, 14, 14}, {1, 0, 0}}));
```
umangyadav, line 326: Did you miss changing shape here compared to previous test ?

src/simplify_algebra.cpp
```diff
@@ -72,14 +74,37 @@ struct find_mul_conv
...
+                return a_lens[i] != 1;
+            }
+            return false;
+        };
+        auto invalid_case = false;
+        for(int i = 2; i < a_lens.size(); ++i)
+        {
+            if(invalid_sl(i))
+                invalid_case = true;
+        }
```
umangyadav, line 100: use `std::any_of` ?

test/simplify_algebra_test.cpp
CharlieL7: This one is using multibroadcast instead of broadcast

src/simplify_algebra.cpp
CharlieL7: I don't see a good way to use `std::any_of` as it needs to check `a_lens` and `a_strides`.

test/simplify_algebra_test.cpp
umangyadav: We should add a test for this case.

CharlieL7: Changed the simplify_mul_conv3 case to have it.

umangyadav (approved):

src/simplify_algebra.cpp
pfultz2, line 100: It would be simpler to write this as:

```cpp
auto is_broadcasted_axis = [](auto len, auto stride) {
    return len == 1 or stride == 0;
};

if (not is_broadcasted_axis(a_lens.front(), a_strides.front()))
    return;

if (not std::equal(a_lens.begin()+2, a_lens.end(), a_strides.begin()+2, a_strides.end(), is_broadcasted_axis))
    return;
```

src/simplify_algebra.cpp
```diff
@@ -72,14 +74,37 @@ struct find_mul_conv
...
+        auto a_input_lens = a_ins->inputs().front()->get_shape().lens();
+
+        std::size_t num_not_one_dims = std::count_if(
+            a_input_lens.cbegin(), a_input_lens.cend(), [](auto dim) { return dim != 1; });
+        if(num_not_one_dims > 1)
+            return;
+
+        // check broadcasted along channels
+        auto a_lens    = a_ins->get_shape().lens();
+        auto a_strides = a_ins->get_shape().strides();
```
pfultz2, line 86: These should be `const auto&`.

src/simplify_algebra.cpp
```diff
@@ -72,14 +74,37 @@ struct find_mul_conv
...
+            return false;
+        };
+        auto invalid_case = false;
+        for(int i = 2; i < a_lens.size(); ++i)
+        {
+            if(invalid_sl(i))
+                invalid_case = true;
+        }
+
+        if(invalid_sl(0) or a_strides.at(1) != 1 or invalid_case)
```
pfultz2, line 102: You should check `a_strides.at(1) != 1` before checking the broadcasted axis. Also, we should probably check the rank as well(ie `a_lens.size() > 2`) before we start checking everything else.

CharlieL7: Shouldn't `rank >= 4` to be a valid convolution output?

pfultz2 (approved):

Issue #1381, opened 2022-09-14
a mistake about fuse_ops.cpp


qianqing13579:
the following line in fuse_ops.cpp has a mistake:
(https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/827baeec9f99b76804b2ac3b70ef1a1786630ebe/src/targets/gpu/fuse_ops.cpp#L262)

I think it should be:
struct hip_add_sigmoid : binary_device<hip_add_sigmoid, &device::add_sigmoid>

umangyadav:
Yes that's a typo. Thanks for surfacing this.

umangyadav:
Fixed with #1385

PR #1385, opened 2022-09-16
Fix typo for add_sigmoid

umangyadav:
Solves https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1381

Surprisingly, this didn't affect functionality. Because of this typo, `add_sigmoid` would appear under the name `add_relu` but still do `add_sigmoid` operation.  Perf report would show `add_relu` instead of `add_sigmoid` as consequence. 
We didn't catch this earlier because we didn't have unit test for `fuse_ops` pass.


kahmed10 (approved):

test/gpu/fuse_ops.cpp
```diff
@@ -0,0 +1,66 @@
...
+        auto* mm = p2.get_main_module();
+        auto s   = migraphx::shape{migraphx::shape::float_type, {4, 3, 3, 3}};
+        auto x   = mm->add_parameter("x", s);
+        auto y   = mm->add_parameter("y", s);
+        auto buffer =
+            mm->add_instruction(migraphx::make_op("allocate", {{"shape", migraphx::to_value(s)}}));
+        mm->add_instruction(migraphx::make_op("gpu::add_sigmoid"), {x, y, buffer});
+    }
+    EXPECT(p1 == p2);
+}
```
pfultz2, line 64: Since this is deprecated and will be removed, I would prefer not to add a test for it, but I do like the idea of adding tests for fuse_ops.

umangyadav: Removed test for now.

PR #1387, opened 2022-09-16
Refactor dynamic padding mode

CharlieL7:
* Removes `use_dynamic_same_auto_pad`
* Change `padding_mode` to be used for dynamic padding
* Move `compute_padded_shape` to `pad_calc.cpp` as it will be used in other dynamic padding cases
* Fix `same_lower` `compute_padded_shape` bug and add a test.


src/include/migraphx/pad_calc.hpp
```diff
@@ -42,18 +43,21 @@ void calculate_padding(int64_t idx,
...
-                                           std::vector<std::size_t> dilations,
-                                           bool use_upper = true);
+std::vector<std::size_t> calc_dyn_auto_pad(const std::vector<std::size_t>& input_lens,
+                                           const std::vector<std::size_t>& wei_lens,
+                                           const std::vector<std::size_t>& strides,
+                                           const std::vector<std::size_t>& dilations,
+                                           bool use_upper);
+
+// Used for dynamic auto padding of convolution operators since padding needs to be computed at
+// evaulation time.
```
umangyadav, line 55: evaluation --> compilation time ?

src/onnx/parse_deconvolution.cpp
```diff
@@ -95,6 +95,8 @@ struct parse_deconvolution : op_parser<parse_deconvolution>
             check_attr_sizes(
                 kdims, values["dilation"].size(), "PARSE_CONV_TRANSPOSE: inconsistent dilations");
         }
+
+        // TODO: nothing is done with this?
```
umangyadav, line 99: what's the todo item here ?

test/verify/quant_conv_valid_mode.cpp
```diff
@@ -1,46 +0,0 @@
...
-    migraphx::program create_program() const
-    {
-        migraphx::program p;
-        auto* mm = p.get_main_module();
-        migraphx::shape a_shape{migraphx::shape::int8_type, {2, 3, 4, 4}};
-        auto pa = mm->add_parameter("a", a_shape);
-        migraphx::shape c_shape{migraphx::shape::int8_type, {2, 3, 3, 3}};
-        auto pc = mm->add_parameter("c", c_shape);
-        mm->add_instruction(
-            migraphx::op::quant_convolution{{{0, 0}}, {{1, 1}}, {{1, 1}}, migraphx::op::valid},
```
umangyadav, line 41: can this be made work with default_ padding ? instead of deleting the case ?

src/include/migraphx/pad_calc.hpp
CharlieL7: It's at evaluation time since the padding can change with different shape inputs.

src/onnx/parse_deconvolution.cpp
CharlieL7: The auto_pad option doesn't have any functionality for parse_deconvolution. It doesn't look like it's implemented. I will update the TODO comment to say this.

src/include/migraphx/op/convolution.hpp
```diff
@@ -93,13 +91,6 @@ struct convolution
            x_shape.lens().at(1) != (w_shape.lens().at(1) * group))
             MIGRAPHX_THROW("CONVOLUTION: mismatched channel numbers");
 
-        std::vector<op::padding_mode_t> dyn_pad_modes = {op::padding_mode_t::same_upper,
-                                                         op::padding_mode_t::same_lower};
-        if(use_dynamic_same_auto_pad and not contains(dyn_pad_modes, padding_mode))
-        {
-            MIGRAPHX_THROW("CONVOLUTION: use_dynamic_same_auto_pad set with invalid padding mode");
-        }
```
umangyadav, line 101: Do we need check that if inputs are dynamic then, padding must not be `default_` ?

test/onnx/onnx_test.cpp
```diff
@@ -1034,15 +1033,11 @@ TEST_CASE(conv_dynamic_batch_same_upper)
...
-                                               {"stride", {1, 1}},
-                                               {"dilation", {1, 1}},
-                                               {"padding_mode", migraphx::op::padding_mode_t::same},
-                                               {"use_dynamic_same_auto_pad", false}}),
-                            l0,
-                            l1);
+    auto c0 = mm->add_instruction(
+        migraphx::make_op("convolution",
+                          {{"padding", {1, 1, 1, 1}}, {"stride", {1, 1}}, {"dilation", {1, 1}}}),
+        l0,
```
TedThemistokleous, line 1039: So we're getting rid of all the padding_mode::same flags everywhere and assume everything dynamic or not is handled correctly here then?

src/pad_calc.cpp
```diff
@@ -86,5 +87,28 @@ std::vector<std::size_t> calc_dyn_auto_pad(std::vector<std::size_t> tensor_lens,
...
 }
 
+shape compute_padded_shape(const shape& input,
+                           const shape& weights,
+                           const std::vector<std::size_t>& padding,
+                           const std::vector<std::size_t>& stride,
+                           const std::vector<std::size_t>& dilation)
+{
+    const size_t num_spatial_dims = input.lens().size() - 2;
+
```
TedThemistokleous, line 97: can you add a check here incase somehow num_spatial_dims roll over? size_t I don't think goes negative.

```
if (incase num_spatial_dims > input.lens().size()) 
                  MIGRAPHX_THROW(...)
'''

File: src/onnx/parse_deconvolution.cpp [hunk 7]
```diff
@@ -95,6 +95,8 @@ struct parse_deconvolution : op_parser<parse_deconvolution>
             check_attr_sizes(
                 kdims, values["dilation"].size(), "PARSE_CONV_TRANSPOSE: inconsistent dilations");
         }
+
+        // TODO: auto padding needs to be implemented for this parser and operator
```
TedThemistokleous, line 99: Separate PR or will it be added in this one?

TedThemistokleous (changes requested):
Few questions. Main concern is possible roll over with the size_t when performing the padding. Add the MIGRAPHX_THROW when we take lens()- 2 so we don't loop more than we need to.

src/pad_calc.cpp
CharlieL7: The way this function is supposed to be used, the rank of `input` should always be `>= 4`. Should this be an assert or a throw?

src/include/migraphx/op/convolution.hpp
CharlieL7: No, the padding can be constant when the inputs are dynamic. The `padding_mode` is only not `default_` when the user wants `SAME_LOWER` or `SAME_UPPER` padding with dynamic shapes.

src/onnx/parse_deconvolution.cpp
TedThemistokleous: Got it, that's fine with me

test/onnx/onnx_test.cpp
TedThemistokleous: Oh okay. I think thats why I was a bit confused because it was just everywhere and then taken out but you're adding dynamic padding which that "seems" like what the old way was referring to.

Anyway more for my curiosity. Thanks!

TedThemistokleous (approved):

src/pad_calc.cpp
```diff
@@ -52,19 +52,21 @@ void calculate_padding(int64_t idx,
...
+std::vector<std::size_t> calc_dyn_auto_pad(const std::vector<std::size_t>& input_lens,
+                                           const std::vector<std::size_t>& wei_lens,
+                                           const std::vector<std::size_t>& strides,
+                                           const std::vector<std::size_t>& dilations,
                                            bool use_upper)
 {
     std::vector<std::size_t> padding;
-    padding.resize(2 * k_lens.size());
-    for(std::size_t i = 0; i < padding.size() / 2; i++)
+    assert(input_lens.size >= 4);
```
umangyadav, line 62: input_lens.size()

umangyadav: Also would it work for 1D convolutions ?

src/tf/parse_conv.cpp
```diff
@@ -75,7 +75,6 @@ struct parse_conv : op_parser<parse_conv>
             const std::string& pad_mode = info.attributes.at("padding").s();
             if(pad_mode.find("SAME") != std::string::npos)
             {
-                op.padding_mode                 = op::padding_mode_t::same;
```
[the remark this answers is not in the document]
CharlieL7: We don't handle dynamic shapes for the TF parser right now, so it will always be default.

CharlieL7:
`calc_dyn_auto_pad()` will work for 1D convolutions. Changed the assert to `>=3` instead.

umangyadav (approved):

TedThemistokleous (approved):

PR #1388, opened 2022-09-17
Concat pointwise fusions

pfultz2:


src/targets/gpu/fuse_ops.cpp
```diff
@@ -785,6 +785,34 @@ struct find_layernorm_pointwise
...
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins    = r.result;
+        auto concat = r.instructions["concat"];
+        auto* pm    = ins->module_inputs().front();
+
+        if(not concat->module_inputs().empty())
+            return;
+
```
TedThemistokleous, line 804: It doesn't look like pm depends on module inputs being empty. 

Swap the creation of PM with the early return here. No point in getting the front() if its not empty.

```
       if(not concat->module_inputs().empty())
            return;
       auto* pm    = ins->module_inputs().front();

```

pfultz2: Done

TedThemistokleous:
Not sure why your CI is still breaking. Fix that and I'll approve

turneram (approved):

TedThemistokleous:
> The failure might be related to #1418.

I guess it worth getting that in first then trying it on this. Outside CI I didnt see anything else that stood out.

TedThemistokleous (approved):
LGTM

PR #1390, opened 2022-09-17
Fix invalid program in debug mode from find_splits

pfultz2:


src/module.cpp
```diff
@@ -385,9 +385,13 @@ instruction_ref module::move_instruction(instruction_ref src, instruction_ref ds
 
 instruction_ref module::move_instructions(instruction_ref src, instruction_ref dst)
 {
-    this->move_instruction(src, dst);
     for(auto ins : src->inputs())
-        this->move_instruction(ins, src);
+    {
+        if(not contains(this->impl->instructions, ins))
+            continue;
+        this->move_instructions(ins, dst);
```
umangyadav, line 392:
```suggestion
        this->move_instructions(ins, src);
```

src/module.cpp
```diff
@@ -385,9 +385,13 @@ instruction_ref module::move_instruction(instruction_ref src, instruction_ref ds
 
 instruction_ref module::move_instructions(instruction_ref src, instruction_ref dst)
 {
-    this->move_instruction(src, dst);
     for(auto ins : src->inputs())
-        this->move_instruction(ins, src);
+    {
+        if(not contains(this->impl->instructions, ins))
+            continue;
```
umangyadav, line 391: why would we check for this inside a module, that seems like an impossible case in general.

turneram (approved):

src/module.cpp
pfultz2: No, because of lexical scope the input instruction could be from a parent module.

umangyadav (approved):

PR #1391, opened 2022-09-18
Add nhwc layout to gpu backend

pfultz2:
Right now this is enabled with `MIGRAPHX_ENABLE_NHWC` env variable.


src/include/migraphx/layout_nhwc.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+struct module_pass_manager;
+
+/**
+ * Transform convolutions to nhwc
+ */
+struct layout_nhwc
+{
+    std::string name() const { return "layout_nhwc"; }
+    void apply(module_pass_manager& m) const;
```
TedThemistokleous, line 19: Fix this to be either mpm. CI throwing errors over this

src/targets/cpu/binary.cpp
```diff
@@ -49,9 +49,21 @@ struct dnnl_binary : dnnl_op<dnnl_binary, dnnl::binary>
         auto s0 = inputs.at(0);
         auto s1 = inputs.at(1);
         auto r  = s0;
-        if(s0 != s1 or not s0.packed())
+        if(s0 == s1 and s0.packed())
```
TedThemistokleous, line 52: Nice to see Demorgan's law here.

TedThemistokleous (commented):
Overall looks good. Good test section. Just a few questions from my end. 

Fix the issues with CI (minor)

src/targets/cpu/binary.cpp
```diff
@@ -49,9 +49,21 @@ struct dnnl_binary : dnnl_op<dnnl_binary, dnnl::binary>
...
+        {
+            r = s0.packed() ? s0 : s1;
+        }
+        else if(s0.broadcasted() != s1.broadcasted())
+        {
+            r = s0.broadcasted() ? s1.with_lens(s0.lens()) : s0.with_lens(s0.lens());
+        }
+        else
+        {
+            r = {s0.type(), s0.lens()};
```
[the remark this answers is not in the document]
pfultz2: We want to use the permutation of the packed shape, if neither is packed then try to pick the permutation of the non-broadcasted shape otherwise we just return a standard shape.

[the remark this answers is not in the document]
TedThemistokleous: Ah I see.

src/targets/gpu/target.cpp
```diff
@@ -123,6 +125,9 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         dead_code_elimination{},
         simplify_algebra{},
         simplify_reshapes{},
+        enable_pass(enabled(MIGRAPHX_ENABLE_NHWC{}), layout_nhwc{}),
```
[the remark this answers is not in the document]
pfultz2: Right now this is for testing. Ideally we would probably enable this based on some analysis or have a compile option for it.

src/targets/gpu/lowering.cpp
```diff
@@ -238,34 +234,22 @@ struct miopen_apply
...
-            try
-            { // for the regular convolution and deconvolution, this try would always succeed
-                compile_conv_with_format(int8_x4_format);
-            }
-            catch(migraphx::exception&)
-            {
-                // In case no solver supports the default format, retry using the other format.
-                compile_conv_with_format(not int8_x4_format);
-            }
+            // auto&& op = any_cast<op::convolution>(ins->get_operator());
```
kahmed10, line 240: unused code

src/targets/gpu/lowering.cpp
```diff
@@ -238,34 +234,22 @@ struct miopen_apply
...
+
+            // TODO: Use make_op
```
kahmed10, line 242: Is this TODO needed? The conv uses make_op

src/targets/gpu/lowering.cpp
```diff
@@ -109,9 +104,10 @@ struct miopen_apply
         add_extend_op("scatter_none");
         add_extend_op("topk");
 
-        add_convolution_op<op::convolution>("convolution");
-        add_convolution_op<op::deconvolution>("deconvolution");
-        add_convolution_op<op::quant_convolution>("quant_convolution");
+        add_convolution_op("convolution");
+        add_convolution_op("convolution");
```
umangyadav, line 108: Is this duplication required ?

src/layout_nhwc.cpp
```diff
@@ -0,0 +1,94 @@
...
+        result.insert(m.replace_instruction(output, layout));
+    }
+    return result;
+}
+
+void transform_convolutions(module& m)
+{
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != "convolution")
```
umangyadav, line 51: Should this include deconv and quant_conv as well ? Or is that for later work ?

src/layout_nhwc.cpp
```diff
@@ -0,0 +1,94 @@
...
+            continue;
+        if(ins->get_shape().lens().size() != 4)
+            continue;
+        auto v = ins->get_operator().to_value();
+        if(v.at("group").to<int>() > 1)
+            continue;
+        auto args = ins->inputs();
+        std::transform(args.begin(), args.end(), args.begin(), [&](const auto& i) {
+            return m.insert_instruction(ins, make_op("layout", {{"permutation", {0, 2, 3, 1}}}), i);
+        });
```
umangyadav, line 61: Does it requires change of weights as well ? I thought only input activations would need to be permutated.

src/layout_nhwc.cpp
```diff
@@ -0,0 +1,94 @@
...
+            self(input);
+    })(std::prev(m.end()));
+    return result;
+}
+
+std::unordered_set<instruction_ref> preserve_output_layout(module& m)
+{
+    std::unordered_set<instruction_ref> result;
+    std::vector<instruction_ref> outputs =
+        find_lasts(m, [](auto ins) { return ins->get_shape().lens().size() == 4; });
```
umangyadav, line 36: Predicate should also check for `ins->name()` being convolutions, instead of just looking for lengths of 4.

src/targets/gpu/target.cpp
```diff
@@ -120,6 +123,9 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
...
+        dead_code_elimination{},
+        simplify_reshapes{},
```
umangyadav, line 128: Do we require simplify_reshapes ? I don't see layout_nhwc pass adding any reshapes.

src/targets/gpu/jit/pointwise.cpp
```diff
@@ -91,12 +91,12 @@ struct pointwise_compiler : compiler<pointwise_compiler>
...
     compiler_replace compile(context& ctx, instruction_ref ins, const operation& op) const
     {
-        if(op.name() == "contiguous")
+        if(contains({"layout", "contiguous"}, op.name()))
         {
             return replace(compile_op(
                 ctx,
                 to_shapes(ins->inputs()),
-                {{"lambda", "[](auto x) { return x; }"}, {"kernel", "contiguous_kernel"}}));
+                {{"lambda", "[](auto x) { return x; }"}, {"kernel", op.name() + "_kernel"}}));
```
umangyadav, line 99: So, `layout` op is just identity op at this point.
I am trying to understand how is this intended to work. Is the idea is to just keep it identity for now and later when MIOpen supports NHWC (if it doesn't already), we modify this `identity` to be something different ?

pfultz2: Its not identity, it changes the output shape to have a different permutation.

umangyadav: I see. But the buffer remains in the same order as previously. it just changes the MIGraphX shape.

src/layout_nhwc.cpp
pfultz2: yes, both the inputs and weights should have the channels first.

src/layout_nhwc.cpp
pfultz2: yea, I haven't updated those operators yet.

src/targets/gpu/target.cpp
pfultz2: Thats to remove any redundant transposes.

src/targets/gpu/jit/pointwise.cpp
pfultz2: Its not the same order in memory. Of course its the same element order otherwise it would be broken. Layout doesn't do a transpose.

src/layout_nhwc.cpp
kahmed10: This change had a side effect of inserting a layout immediately after the convolution. I am changing the tests to reflect this behavior.

kahmed10:
@umangyadav @TedThemistokleous please take a look at the proposed changes.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -83,9 +83,10 @@ struct miopen_convolution
 
     inline shape compute_shape(const std::vector<shape>& inputs) const
     {
-        check_shapes{inputs, op}.has(4).standard();
```
TedThemistokleous, line 86: This is intentional correct? We can't assume a standard shape since we need to do a transpose to get things in the nhwc format via our permutations?

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -144,12 +145,9 @@ struct miopen_convolution
 #endif
     }
 
-    inline void set_conv_descriptor()
+    void set_conv_descriptor()
```
TedThemistokleous, line 148: Why remove inline from here?

src/layout_nhwc.cpp
```diff
@@ -0,0 +1,118 @@
...
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != "layout")
+            continue;
+        if(ins->get_shape() != ins->inputs().front()->get_shape())
+            continue;
+        if(contains(output_layouts, ins))
+            continue;
+        m.replace_instruction(ins, ins->inputs().front());
+    }
```
TedThemistokleous, line 103: Probably worth doing another test here for coverage but also converting back from nhwc

TedThemistokleous (commented):
Some questions/comments. Will do another pass in the morning.

src/targets/gpu/compile_miopen.cpp
```diff
@@ -0,0 +1,101 @@
...
+        std::size_t ws = 0;
+        try
+        {
+            // for the regular convolution and deconvolution, this try would always succeed
+            ws = compile(op, ins, int8_x4_format);
+        }
+        catch(migraphx::exception&)
+        {
+            // In case no solver supports the default format, retry using the other format.
+            ws = compile(op, ins, not int8_x4_format);
```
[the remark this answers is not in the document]
umangyadav: This change is unrelated to NHWC and copied from convolution implementation.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
kahmed10: I'm not entirely sure why @pfultz2 did this, but maybe because changing the inline function added significant overhead to rebuilding or caused compatibility issues.

umangyadav (approved):

src/targets/gpu/include/migraphx/gpu/convolution.hpp
TedThemistokleous: All good. I just wasn't sure why we would remove that. Its minor enough that I dont think it breaks things.

TedThemistokleous (approved):

PR #1403, opened 2022-09-24
Swap 18.04 with 22.04 in CI

causten:
Github is dropping support for 18.04 in its Actions.  Drop 18.04 and add 22.04


hip-clang.docker
```diff
@@ -19,6 +19,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     gdb \
     git \
     lcov \
+    libhiredis-dev \
```
pfultz2, line 22: I would prefer not to require installing this package. Perhaps we could either us an option to disable it with ccache(if its available) or install the ccache package instead(as long as its version greater than 4 on ubuntu).

PR #1405, opened 2022-09-27
Stream sync python api changes

TedThemistokleous:
API side changes to get stream syncing working in python.


src/py/migraphx_py.cpp
```diff
@@ -83,7 +83,7 @@ void visit_py(T x, F f)
     {
         f(x.template cast<bool>());
     }
-    else if(py::isinstance<py::int_>(x) or py::hasattr(x, "__index__"))
+    else if(py::isinstance<py::int_>(x) || py::hasattr(x, "__index__"))
```
umangyadav, line 86: I think you should have kept it to `or` to avoid cppcheck.

src/py/migraphx_py.cpp
```diff
@@ -355,6 +355,23 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
...
+                 migraphx::parameter_map pm;
+                 for(auto x : params)
+                 {
+                     std::string key      = x.first.cast<std::string>();
+                     py::buffer b         = x.second.cast<py::buffer>();
+                     py::buffer_info info = b.request();
+                     pm[key]              = migraphx::argument(to_shape(info), info.ptr);
+                 }
+                 migraphx::execution_environment exec_env{
+                     migraphx::any_ptr(reinterpret_cast<void*>(stream), stream_name), true};
```
umangyadav, line 372: you want to cast it to `hipstream_t` if possible. for type checking.

test/py/test_gpu_async.py
```diff
@@ -0,0 +1,47 @@
...
+    p.compile(migraphx.get_target("gpu"))
+    #print(p)
+    params = {}
+
+    for key, value in p.get_parameter_shapes().items():
+        params[key] = migraphx.generate_argument(value)
+
+    stream = ctypes.c_void_p()
+    err = ctypes.c_int(hip.hipStreamCreate(ctypes.addressof(stream)))
+
```
shivadbhavsar, line 43: should call run_async() to execute the program on the stream

src/py/migraphx_py.cpp
TedThemistokleous: The type check happens when we try to get(), it checks typename of the type with stream_name that we pass in.

tools/accuracy/accuracy_checker.py
```diff
@@ -116,6 +116,7 @@ def main():
 
     model = migraphx.parse_onnx(model_name, default_dim_value=batch)
 
+    print(model)
```
shivadbhavsar, line 119: maybe only print this if verbose flag is passed?
`if args.verbose: ...`

TedThemistokleous: done

test/py/test_gpu_async.py
TedThemistokleous: Added. I also added checks to this as well. Curious how this will run in CI.

src/py/migraphx_py.cpp
TedThemistokleous: Fixed

PR #1406, opened 2022-09-29
Stream sync api test

TedThemistokleous:
Final test for stream sync to test API functionality.

### DIFF
--- a/test/api/test_gpu.cpp
+++ b/test/api/test_gpu.cpp
@@ -54,17 +54,6 @@ TEST_CASE(load_and_run)
 
 using hip_ptr = MIGRAPHX_MANAGE_PTR(void, hipFree);
 
-hip_ptr get_hip_host_buffer(size_t size)
-{
-    void* ptr;
-    auto err = hipHostMalloc(&ptr, size);
-    if(err != hipSuccess)
-    {
-        EXPECT(false);
-    }
-    return hip_ptr{ptr};
-}
-
 hip_ptr get_hip_buffer(size_t size)
 {
     void* ptr;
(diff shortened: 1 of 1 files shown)


TedThemistokleous:
@pfultz2 just ended up grabbing code from hip.cpp and putting that here and still running into problems. 

replacing the call to add the input buffer via to_gpu as a program parameter yields

```
root@d5f1b0c18dc9:/code/AMDMIGraphX/build# bin/test_api_gpu load_and_run_async
[   RUN    ] load_and_run_async
"Memobj map already has an en
```

test/api/test_gpu.cpp
```diff
@@ -45,27 +46,68 @@ TEST_CASE(load_and_run)
...
+
+    std::vector<hip_ptr> host_buffs;
+    std::vector<hip_ptr> buffs;
+    std::vector<migraphx::argument> args;
     for(auto&& name : param_shapes.names())
     {
-        pp.add(name, migraphx::argument::generate(param_shapes[name]));
+        args.push_back(migraphx::argument::generate(param_shapes[name]));
+        buffs.push_back(get_hip_buffer(args.rbegin()->get_shape().bytes()));
+        host_buffs.push_back(get_hip_buffer(args.rbegin()->get_shape().bytes()));
```
umangyadav, line 100: Doesnt' this `host_buffs` needs to point to `args` first before starting copy to GPU ?

TedThemistokleous: Ah got it. Ill fix that. Printing it out looks like I was getting all zeros.

TedThemistokleous: Done

umangyadav (approved):

PR #1407, opened 2022-09-29
Refactor dynamic compute; Dynamic ref unary functions

CharlieL7:
### Refactor dynamic compute
* add a `compute_output_shape` object that implicitly converts to a new `dyn_output` or `shape` object
* `dyn_output` object can handle computing the static output shape of an operator given the input arguments shapes
* change an operator's compute function to `argument compute(const dyn_output& dyn_out, std::vector<argument> args)` to use dyn_output object

### Dynamic ref unary functions
* Included these changes to have an example of the refactored dynamic compute being used
* Changes to unary base class to handle dynamic shapes
* Changed `elu` and `leaky_relu` to use unary base class and pointwise JIT
* Fixed typo error for `elu` verify test


src/targets/gpu/lowering.cpp
```diff
@@ -96,9 +96,9 @@ struct miopen_apply
 
         add_extend_op("argmax");
         add_extend_op("argmin");
-        add_extend_op("elu");
+        // add_extend_op("elu");
         add_extend_op("gather");
-        add_extend_op("leaky_relu");
+        // add_extend_op("leaky_relu");
```
pfultz2, line 101: These comments need to be removed.

src/include/migraphx/op/unary.hpp
```diff
@@ -30,6 +30,7 @@
 #include <migraphx/argument.hpp>
 #include <migraphx/stringutils.hpp>
 #include <migraphx/value.hpp>
+#include <migraphx/operation.hpp>
```
pfultz2, line 33: Dont include operation.hpp. We should just include the header for `dyn_output` struct.

src/targets/gpu/lowering.cpp
CharlieL7: The ops aren't used so these can be removed. Should I remove the miopen calls in this PR?

pfultz2: We could probably do that in a later PR.

src/include/migraphx/dyn_output.hpp
```diff
@@ -21,44 +21,55 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-#ifndef MIGRAPHX_GUARD_RTGLIB_ELU_HPP
```
kahmed10, line 24: not sure why this is showing up as changes to the elu.hpp file. If it will affect commit history, I would prefer the original elu.hpp file is removed, and the dyn_output.hpp file is added as a new file.

CharlieL7: I also have no idea. Does anyone know how to fix this? The original file should be deleted.

pfultz2: Its a git thing, if a file looks similar enough git assumes it was renamed, unfortunately.

pfultz2 (approved):

kahmed10 (approved):

PR #1408, opened 2022-09-30
Find2.0 changes for the Quant  and De-Convolution

umangyadav:
Requires following fixes in MIOpen before this can actually work with MIGraphX. 
https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1800
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1791

I've built MIOpen with fixes locally and it passes the MIGraphX's tests. 

Part of #1303 

Update : 6th October, 2022
- Refactored `convolution`, `quant_convolution` and `deconvolution` into single base class.

Update : 12th October, 2022,
- Making use of `compile` instead of `find`.


src/targets/gpu/deconvolution.cpp
```diff
@@ -55,14 +55,40 @@ argument miopen_deconvolution::compute(context& ctx,
...
+                                        solution_ptr.get(),
+                                        3,
+                                        tensor_args,
+                                        args[2].implicit(),
+                                        workspace_size);
+        if(status != miopenStatusSuccess)
+            MIGRAPHX_THROW("MIOpen Deconvolution: running convolution using find_2.0 failed");
+
+        return args[3];
+    }
```
pfultz2, line 85: These should all be combined into the same class. Similiar to how we combine rocblas gemms with `rcoblas_gemm`:

```cpp
template<class Op>
struct miopen_convolution
{
    Op op;
    bool int8_x4_format = false;
    shared<convolution_descriptor> cd;
#ifdef MIGRAPHX_HAS_FIND_2_API
    value::binary solution_object{};
    shared<miopen_solution> solution_ptr = nullptr;
#endif
    uint64_t solution_id = 0;

    {
        // TODO: Add algo
        return pack_join(migraphx::reflect(self.op, f),
                         pack(
#ifdef MIGRAPHX_HAS_FIND_2_API
                             f(self.solution_object, "solution_object"),
#endif
                             f(self.int8_x4_format, "int8_x4_format")));
    }
    ...
};
```

We could inline everything and then implement the [`compile`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/operation.hpp#L491) method to get the workspace, which means we can avoid including this header elsewhere(I think only lowering uses these headers). I also plan to implement this for NHWC.

umangyadav: Unified all into one single base class. 

`compile` method, are you saying i should rename `find` method to `compile` ?

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -24,20 +24,60 @@
...
+    check_shapes{inputs, op}.has(4).standard();
+    std::vector<shape> conv_inputs(inputs.begin(), inputs.begin() + 2);
+    check_shapes{conv_inputs, op}.max_ndims(5);
+    return op.compute_shape(conv_inputs);
+}
 
 struct miopen_convolution
 {
-    op::convolution op;
+    operation op                      = op::identity{};
```
kahmed10, line 78: Any reason to default this to identity instead of convolution?

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -48,29 +88,291 @@ struct miopen_convolution
     template <class Self, class F>
     static auto reflect(Self& self, F f)
     {
-        return pack(f(self.op.padding, "padding"),
```
kahmed10, line 51: I think we can retain these op fields since they're shared between conv, deconv, and quant conv.

umangyadav: `pack_join(pack(f(self.op, "op")),` would print all attributes for any of the conv types.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
umangyadav: Removed.

umangyadav: Identity is needed. It requires some reference to operation object without that it  fails on assertion in `operation.hpp`.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -48,29 +88,291 @@ struct miopen_convolution
...
+                                                        workspace_size,
+                                                        solution_id);
+
+        if(status != miopenStatusSuccess)
+            MIGRAPHX_THROW("MIOpen " + op.name() + ": running convolution failed");
+        return args[3];
+#endif
+    }
+
+    inline void make_conv_descriptor()
```
pfultz2, line 162: It would be better named `set_conv_descriptor`.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -48,29 +88,291 @@ struct miopen_convolution
...
 #ifdef MIGRAPHX_HAS_FIND_2_API
-                    f(self.solution_object, "solution_object"),
+                             f(self.solution_object, "solution_object"),
 #endif
-                    f(self.solution_id, "solution_id"));
+                             f(self.int8_x4_format, "int8_x4_format"),
+                             f(self.solution_id, "solution_id")));
+    }
+
+    std::string name() const { return "miopen_convolution"; }
```
pfultz2, line 100: Name should be prefixed with `gpu::`.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -25,19 +25,59 @@
...
+        std::vector<size_t> new_dims = dims;
+        new_dims.insert(new_dims.begin() + 2, 1);
+        new_shape = shape{input.type(), new_dims};
+    }
+    return new_shape;
+}
+
+template <typename T>
+auto conv_compute_shape(rank<2>, const T& op, const std::vector<shape>& inputs)
+    -> decltype(op.normalize_compute_shape(inputs))
```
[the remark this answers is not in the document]
umangyadav: Removed.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
umangyadav: Renamed.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -184,30 +184,32 @@ MIGRAPHX_PRED_MATCHER(fusable_conv, instruction_ref ins)
         return false;
     if(enabled(MIGRAPHX_DISABLE_MIOPEN_FUSION{}))
         return false;
-    if(ins->name() != "gpu::convolution")
+    if(ins->name() != "gpu::miopen_convolution")
+        return false;
+    auto miopen_conv_op = any_cast<miopen_convolution>(ins->get_operator());
```
pfultz2, line 189: Use `ins->get_operator().to_value()`, and remove the include for `migraphx/gpu/convolution.hpp`.

src/targets/gpu/pack_int8_args.cpp
```diff
@@ -177,8 +178,13 @@ void pack_int8_args::apply(module& m) const
                 m.replace_instruction(ins, ins->get_operator(), inputs);
             }
         }
-        else if(ins->name() == "gpu::quant_convolution")
+        else if(ins->name() == "gpu::miopen_convolution")
```
pfultz2, line 181: Templating the class will make it simpler to check the operator name.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
pfultz2: This is similar to how it is done for gemms and rocblas.

umangyadav: Made changes to make it templated.

src/targets/gpu/pack_int8_args.cpp
umangyadav: Done.

kahmed10 (approved):

src/targets/gpu/fuse_ops.cpp
umangyadav: Two things are preventing me from doing so:
1. `miopen_conv_bias` construction requries `op::convolution`.  I need to cast `miopen_convolution<op::convolution>`. to get that `op::convolution` object.  I can make it work if i can some how reflect "op" object from `miopen_convolution`. which is not working with templated `miopen_convolution`. 

2. `fuse_ops` pass refers to `algo` (e.g  miopenConvolutionFwdWinograd) in many places. which is not reflected currently.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
umangyadav: I am running into errors when trying to reflect entire `"op"` with templated code.  Errors says it doesn't have matching `stream_write_value_impl(rank<XY>, std::ostream& os, const T& x) ` for either `op::deconvolution`, `op::convolution` or `op::quant_convolution`.

src/targets/gpu/fuse_ops.cpp
pfultz2: > miopen_conv_bias construction requries op::convolution. 

You can deserialize the operator:

```cpp
auto v = ins->get_operator().to_value();
auto conv_op = from_value<op::convolution>(v["op"]);
```

> fuse_ops pass refers to algo (e.g miopenConvolutionFwdWinograd) in many places. which is not reflected currently.

Lets add that to the reflect method then.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -48,27 +70,275 @@ struct miopen_convolution
     template <class Self, class F>
     static auto reflect(Self& self, F f)
     {
-        return pack(f(self.op.padding, "padding"),
-                    f(self.op.stride, "stride"),
-                    f(self.op.dilation, "dilation"),
-                    f(self.op.group, "group"),
-                    f(self.op.padding_mode, "padding_mode"),
+        return pack_join(migraphx::reflect(self.op, f),
```
pfultz2, line 73: Dont do `migraphx::reflect(self.op, f)`, just do `f(self.op, "op")`.

umangyadav: Done.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
umangyadav: Made  required changes.

src/include/migraphx/streamutils.hpp
```diff
@@ -61,34 +62,50 @@ inline stream_range_container<Range> stream_range(const Range& r)
...
 template <class T>
-auto stream_write_value_impl(rank<1>, std::ostream& os, const T& x) -> decltype(os << x, void())
+auto stream_write_value_impl(rank<2>, std::ostream& os, const T& x) -> decltype(os << x, void())
 {
     os << x;
 }
 
 template <class T>
-void stream_write_value_impl(rank<1>, std::ostream& os, const std::vector<T>& r)
+void stream_write_value_impl(rank<2>, std::ostream& os, const std::vector<T>& r)
```
pfultz2, line 71: This should be `rank<1>`.

src/include/migraphx/streamutils.hpp
```diff
@@ -61,34 +62,50 @@ inline stream_range_container<Range> stream_range(const Range& r)
...
-auto stream_write_value_impl(rank<0>, std::ostream& os, const Range& r)
+auto stream_write_value_impl(rank<1>, std::ostream& os, const Range& r)
     -> decltype(r.begin(), r.end(), void())
 {
     os << "{";
     os << stream_range(r);
     os << "}";
 }
 
+template <class T>
```
pfultz2, line 87: This should do `MIGRAPHX_REQUIRES(is_reflectable<T>{})`

src/include/migraphx/streamutils.hpp
```diff
@@ -61,34 +62,50 @@ inline stream_range_container<Range> stream_range(const Range& r)
...
+auto stream_write_value_impl(rank<0>, std::ostream& os, const T& x)
```
pfultz2, line 88: Use `void` instead of `auto`.

src/include/migraphx/streamutils.hpp
```diff
@@ -61,34 +62,50 @@ inline stream_range_container<Range> stream_range(const Range& r)
...
+    -> decltype(os << x.name(), void())
```
pfultz2, line 89: Remove the trailing return type. In fact this is not even correct. There doesnt need to be a requirement for `x.name()`.

src/include/migraphx/streamutils.hpp
```diff
@@ -61,34 +62,50 @@ inline stream_range_container<Range> stream_range(const Range& r)
...
+{
+    os << x.name();
```
pfultz2, line 91: Remove this line.

src/include/migraphx/streamutils.hpp
```diff
@@ -61,34 +62,50 @@ inline stream_range_container<Range> stream_range(const Range& r)
...
+    char delim = '[';
```
pfultz2, line 92: Use `{` and `}` like we do for the other overloads.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -190,11 +189,12 @@ MIGRAPHX_PRED_MATCHER(fusable_conv, instruction_ref ins)
         return false;
     auto wei = ins->inputs().at(1)->get_shape();
     assert(wei.lens().size() == 4);
-    auto miopen_conv_op = any_cast<miopen_convolution<op::convolution>>(ins->get_operator());
-    auto conv_op        = miopen_conv_op.op;
+    auto miopen_conv_op = ins->get_operator().to_value();
+    auto algo           = miopen_conv_op.at("algo").to<int>();
```
pfultz2, line 193: Do `miopen_conv_op.at("algo").to<miopenConvFwdAlgorithm_t>()`.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -528,8 +528,8 @@ struct find_conv_pointwise
         auto ins         = r.result;
         auto input_ins   = conv_ins->inputs().at(0);
         auto weights_ins = conv_ins->inputs().at(1);
-        auto conv_op   = any_cast<miopen_convolution<op::convolution>>(conv_ins->get_operator()).op;
-        auto alloc_ins = ins->inputs().back();
+        auto conv_op     = from_value<op::convolution>((conv_ins->get_operator()).to_value()["op"]);
```
pfultz2, line 531: I think you can remove the extra parenthesis around `conv_ins->get_operator()`.

src/include/migraphx/reflect.hpp
```diff
@@ -111,8 +111,17 @@ auto reflect(T& x, Selector f)
 template <class T>
 auto reflect_tie(T& x)
 {
-    return reflect(x, [](auto&& y, auto&&...) { return detail::wrap<decltype(y)>(y); })(
-        [](auto&&... xs) { return detail::auto_tuple(xs.get()...); });
+    return reflect(x, [](auto&& y, auto&&...) {
+        if constexpr(is_reflectable<decltype(y)>{}) // cppcheck-suppress UnnecessaryElseStatement
```
pfultz2, line 115: Does removing the else not compile?

Also I think you can put the `cppcheck-suppress UnnecessaryElseStatement` on the line above.

src/include/migraphx/streamutils.hpp
pfultz2: That is `template <class T, MIGRAPHX_REQUIRES(is_reflectable<T>{})>`

src/include/migraphx/reflect.hpp
umangyadav: yes, If i remove `else` it doesn't compile.

src/include/migraphx/streamutils.hpp
umangyadav: Changed ranking

src/include/migraphx/streamutils.hpp
umangyadav: Done

src/include/migraphx/streamutils.hpp
umangyadav: Done

src/include/migraphx/streamutils.hpp
umangyadav: Done

src/targets/gpu/fuse_ops.cpp
umangyadav: Done

src/targets/gpu/fuse_ops.cpp
umangyadav: Done.

src/include/migraphx/streamutils.hpp
umangyadav: Done.

src/include/migraphx/reflect.hpp
umangyadav: Moved `cppcheck-suppress` to line above.

pfultz2 (approved):

PR #1411, opened 2022-10-10
Add JIT pad

kahmed10:
- updated GPU pad to now use JIT version.
- added range functions for JIT kernels.
- tested with larger input size to ensure padding was still working correctly.


src/targets/gpu/kernels/include/migraphx/kernels/pad.hpp
```diff
@@ -0,0 +1,55 @@
...
+        auto multi = output_shape.multi(i);
+
+        for(std::size_t j = 0; j < offsets.size(); j++)
+        {
+            if(multi[j] < offsets[j] or multi[j] >= bounds[j])
+            {
+                output[multi] = pad_val;
+                return;
+            }
+        }
```
pfultz2, line 49: I dont really understand this. I would do something like this(assuming a lexical comparison):

```cpp
auto input_idx = multi - offsets;
if (multi < offsets or input_idx >= input.get_shape().lens)
    output[multi] = pad_val;
else
    output[multi] = input[input_idx];
```

Does that not work?

src/targets/gpu/kernels/include/migraphx/kernels/pad.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <migraphx/kernels/clamp.hpp>
+#include <migraphx/kernels/shape.hpp>
+#include <migraphx/kernels/index.hpp>
+#include <migraphx/kernels/print.hpp>
+
+namespace migraphx {
+
+template <class T, class U, class V, class W, class X, class Y>
+__device__ void
+pad(const T& idx, const U& offsets, const V& bounds, const W& input, X& output, const Y& pad_val)
```
pfultz2, line 36: Dont template the `idx` parameter, And name the template parameter types as `Offsets`, `Input`, `Output`, and `PadVal`.

src/targets/gpu/kernels/include/migraphx/kernels/pad.hpp
kahmed10: I got the above to work if I had the following comparison:
```
friend constexpr bool operator<(const array& x, const array& y)
    {
        for(index_int i = 0; i < N; i++)
        {
            if(x[i] >= y[i]) return false;
        }
        return true;
    }
friend constexpr bool operator>=(const array& x, const array& y) { return not( x < y); }

```

kahmed10: Actually this should be logically identical to what is in develop currently. Just the operator >= needed to be changed.

pfultz2: Yea thinking about this more and lexical ordering wont work. We do want to use product order as we want to check if any index is out of bounds. So the `input_idx >= input.get_shape().lens` check will work without changed to array.hpp, but the `multi < offsets` wont work. We really want to check if `multi - offsets` has negative indices(or rather `input_idx < 0`), which wont work for unsigned integers. We could check that `input_idx > multi`.

src/targets/gpu/kernels/include/migraphx/kernels/array.hpp
```diff
@@ -218,14 +218,14 @@ struct array
     {
         for(index_int i = 0; i < N; i++)
         {
-            if(not(x[i] < y[i]))
+            if((x[i] >= y[i]))
```
pfultz2, line 221: I think its better to write this in terms of `<`.

src/targets/gpu/kernels/include/migraphx/kernels/array.hpp
```diff
@@ -218,14 +218,14 @@ struct array
...
                 return false;
         }
         return true;
     }
     friend constexpr bool operator>(const array& x, const array& y) { return y < x; }
     friend constexpr bool operator<=(const array& x, const array& y) { return (x < y) or (x == y); }
-    friend constexpr bool operator>=(const array& x, const array& y) { return (y < x) or (x == y); }
+    friend constexpr bool operator>=(const array& x, const array& y) { return not(x < y); }
```
pfultz2, line 228: I think `(x >= y) == not(x < y)` is not valid for product order since its not totally ordered. 

https://en.cppreference.com/w/cpp/concepts/totally_ordered
https://en.wikipedia.org/wiki/Product_order

src/targets/gpu/kernels/include/migraphx/kernels/pad.hpp
kahmed10: I've tried the following which still doesn't work:
`if(input_idx > multi or input_idx >= input.get_shape().lens`
I think it's because the `>` overload is just doing `y < x`. If I have a simple example like this:
```
pads:
0, 0, 1, 1, 0, 0, 1, 1 // pad before and after on last 2 dims
offsets:
0, 0, 1, 1
multi:
0, 0, 0, 0
input_idx:
0, 0, 4294967295, 4294967295
input.get_shapes().lens:
1, 1, 2, 2
```
What happens is that it will iterate through the first 2 dimensions and identify that `0 < 0` is false, so the `input_idx > multi` condition fails.

src/targets/gpu/kernels/include/migraphx/kernels/pad.hpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    auto output_shape = output.get_shape();
+    idx.global_stride(output_shape.elements(), [&](auto i) {
+        auto multi  = output_shape.multi(i);
+        auto bounds = input.get_shape().lens + offsets;
+        for(auto j = 0; j < offsets.size(); j++)
+        {
+            // offsets accounts for padding at the beginning
+            // bounds accounts for padding at the end
+            if(multi[j] < offsets[j] or multi[j] >= bounds[j])
```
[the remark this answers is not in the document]
kahmed10: I've refactored this part to remove the extra bounds calculation.

src/targets/gpu/kernels/include/migraphx/kernels/array.hpp
kahmed10: I've reverted this file.

src/targets/gpu/kernels/include/migraphx/kernels/array.hpp
kahmed10: Reverted back to as it was before.

src/targets/gpu/kernels/include/migraphx/kernels/pad.hpp
pfultz2: Yea we should probably just do:

```cpp
bool use_pad = false;
for(index_int i = 0; i < multi.size(); i++)
{
    if (input_idx[i] > multi[i] or input_idx[i] >= input.get_shape().lens[i])
    {
        use_pad = true;
        break;
    }   
}
```

Although it would be cleaner to use `range(multi.size())` with `any_of` instead.

CharlieL7 (approved):
Looks good. May want to add a asymmetrical ref padding test, since it looks like only the symmetrical case is tested right now.
```
TEST_CASE(pad_asym_test)
{
    migraphx::program p;
    migraphx::shape s{migraphx::shape::float_type, {2, 2}};
    auto l0 = mm->add_literal(migraphx::literal{s, {1, 2, 3, 4}});
    mm->add_instruction(migraphx::make_op("pad", {{"pads", {0, 0, 1, 1}}}), l0);
    p.compile(migraphx::ref::target{});
    auto result = p.eval({}).back();
    std::vector<float> results_vector(16);
    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
    std::vector<float> gold{1, 2, 0, 3, 4, 0, 0, 0, 0};
}
```

kahmed10:
> Looks good. May want to add a asymmetrical ref padding test, since it looks like only the symmetrical case is tested right now.
> 
> ```
> TEST_CASE(pad_asym_test)
> {
>     migraphx::program p;
>     auto* mm = p.get_main_module();
>     migraphx::shape s{migraphx::shape::float_type, {2, 2}};
>     auto l0 = mm->add_literal(migraphx::literal{s, {1, 2, 3, 4}});
>     mm->add_instruction(migraphx::make_op("pad", {{"pads", {0, 0, 1, 1}}}), l0);
>     p.compile(migraphx::ref::target{});
>     auto result = p.eval({}).back();
>     std::vector<float> results_vector(16);
>     result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
>     std::vector<float> gold{1, 2, 0, 3, 4, 0, 0, 0, 0};
>     EXPECT(migraphx::verify_range(results_vector, gold));
> }
> ```

We have existing pad tests that have asymmetrical pads. I mainly wanted to include the large one which would fail on develop.

kahmed10:
> TEST_CASE(pad_asym_test)
> {
>     migraphx::program p;
>     auto* mm = p.get_main_module();
>     migraphx::shape s{migraphx::shape::float_type, {2, 2}};
>     auto l0 = mm->add_literal(migraphx::literal{s, {1, 2, 3, 4}});
>     mm->add_instruction(migraphx::make_op("pad", {{"pads", {0, 0, 1, 1}}}), l0);
>     p.compile(migraphx::ref::target{});
>     auto result = p.eval({}).back();
>     std::vector<float> results_vector(16);
>     result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
>     std::vector<float> gold{1, 2, 0, 3, 4, 0, 0, 0, 0};
>     EXPECT(migraphx::verify_range(results_vector, gold));
> }

Added

src/targets/gpu/jit/pad.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+        auto padding    = v.at("pads").to_vector<int64_t>();
+        auto input_lens = inputs.front().lens();
+        std::vector<size_t> offsets(input_lens.size());
+        std::copy(padding.begin(), padding.begin() + offsets.size(), offsets.begin());
+
+        auto src =
+            interpolate_string(pointwise_kernel,
+                               {{"pad_val", to_string(pad_val_string)},
+                                {"offsets", to_string_range(offsets.begin(), offsets.end())}});
```
pfultz2, line 90: Does `to_string_range(offsets)` work?

kahmed10: yeah, updated

pfultz2 (approved):

PR #1413, opened 2022-10-13
Add script to convert models based on onnx model conversion example

TedThemistokleous:
Work in progress. Seems to break on a few different models.

Should be used so we can get all models into a certain opset before testing.

This came up in issue #1283 for the resnext50 model after some changes on #1325  were using add_outline to fix empty case of if operands.. 

Realized the split operator uses opset 13 even though we're coded for opset 12, causing some odd behavior when we parse in the graph.

As @pfultz2  suggested, this would be useful for us to be able to standardize our models before we run them in migraphx to get rid of opset gotchas as per the onnx spec, new opsets "should be backwards compatible."

https://onnxruntime.ai/asddocs/reference/compatibility.html


tools/convert_onnx_version.py
```diff
@@ -0,0 +1,79 @@
...
+
+    model_path     = args.model
+    out_model_path = args.output
+    target_opset   = args.opset
+    verbose        = args.verbose
+
+
+    original_model = onnx.load(model_path)
+    if verbose:
+        print(f"The model before conversion:\n{original_model}")
```
umangyadav, line 67: Is it possible to print onnx_model in text format ?

umangyadav (approved):

tools/convert_onnx_version.py
```diff
@@ -0,0 +1,79 @@
...
+
+     # A full list of supported adapters can be found here:
+    # https://github.com/onnx/onnx/blob/main/onnx/version_converter.py#L21
+    # Apply the version conversion on the original model
+    converted_model = version_converter.convert_version(original_model, target_opset)
+
```
umangyadav, line 73: Does it  only do backward conversion on the opset ? or can it handle forward as well ?

TedThemistokleous: I can handle both directions from my understanding.

TedThemistokleous:
It appears my issues aren't on the MIGraphX side but on the ONNX side with the errors I'm seeing. There's a support issue with how they convert models. 

I've opened an issue and PR on that side for these script.

kahmed10 (approved):

PR #1416, opened 2022-10-19
Add relaxed standard shape assertion

jungpark-mlir:
Reiterate the assertion on the standard shape but relax it for the multibroadcast ops deliberately inserted to explicit the broadcast.


src/targets/gpu/mlir.cpp
```diff
@@ -194,19 +194,20 @@ struct mlir_program
         return result;
     }
 
-    MlirType make_tensor(const shape& s) const
+    MlirType make_tensor(const shape& s, bool bStandard) const
```
pfultz2, line 197: Follow C++ naming for variable names.

src/targets/gpu/mlir.cpp
```diff
@@ -498,7 +499,7 @@ struct mlir_program
             auto ops  = create_operation_state(name);
             ops.add_attribute_value(get_operator_value(ins->get_operator()));
             if(ins->name() != "@return")
-                ops.add_results({get_shape(ins)});
+                ops.add_results({get_shape(ins)}, ins->name() != "multibroadcast");
```
pfultz2, line 502: We would need to this for more then just broadcast. Transpose will have a non-standard shape as well and output of many operators will propagate the permutation as well so they all will be non-standard.

src/targets/gpu/mlir.cpp
```diff
@@ -369,9 +370,9 @@ struct mlir_program
             return *this;
         }
 
-        mlir_operation_state& add_results(const std::vector<shape>& outputs)
+        mlir_operation_state& add_results(const std::vector<shape>& outputs, bool bStandard)
         {
-            auto x = prog->make_tensors(outputs);
+            auto x = prog->make_tensors(outputs, bStandard);
```
pfultz2, line 375: I would rather convert the shape to standard instead of trying to pass a boolean through all of this.

jungpark-mlir: Do you mean, having something like
shape::get_standardize()
which recalculates stride for the lens, ignoring the inferred shape?

That could workaround this. little hacky.

pfultz2 (approved):

PR #1418, opened 2022-10-19
 rearrange default pass list; adjust_allocation must be run after rep…

bpickrel:
…lace_allocate for transpose operations.  The role of adjust_allocation is to check that the allocated shape is the size given in the instruction, and replace it with a new one if necessary.

This fixes an observed regression error in an internal issue ticket, with failed compilation of model frozen_resnet_v2_50.pb.  The problem was apparently that a convolution <correction: a transpose op before a convolution> created an instruction with a nonstandard shape attached (i.e. the stride pattern was not in the default order), and a subsequent automatic check threw an exception.  The extra `adjust_allocation` pass cleans up this shape.  The issue originates with a reordering of passes in commit 6f7ee0b788b6241eb973c49d384e45d7b3858006 (jul 7) as part of fuse_ops work in [PR 1280](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1280/files) but is not directly related to fuse_ops.


pfultz2:
Yea it would be good to have a verification test for this though.

bpickrel:
[AMD Official Use Only - General]

It’s easy to add this test to the PR.  But would it do any good, or does it need to be modified?

test/shape_test.cpp
```diff
@@ -64,6 +64,15 @@ TEST_CASE(test_shape_standard)
...
 }
 
+TEST_CASE(test_shape_standard_stride)
+{
+    migraphx::shape s{migraphx::shape::float_type, {1, 2048, 1, 1}, {2048, 1, 2048, 2048}};
+    EXPECT(not s.standard());
+    EXPECT(s.packed());
+    EXPECT(s.transposed());
+    EXPECT(not s.broadcasted());
+}
```
pfultz2, line 74: This test is bot related to this PR. It should go in a different PR.

bpickrel: ok, reverted

CharlieL7 (approved):
LGTM. By verify test I think Paul means a test in `test/verify` that would have a minimal example of the bug.

kahmed10 (approved):

bpickrel:
New verification test added.  Verified that the test throws an exception if called with the pre-pull request pass list.

test/verify/test_shape_alloc.cpp
```diff
@@ -0,0 +1,62 @@
...
+ *   the 2nd index, a non-standard layout that should be reallocated by adjust_allocation.
+ */
+struct test_shape_alloc : verify_program<test_shape_alloc>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+
+        migraphx::shape shape135{migraphx::shape::float_type, {11, 8, 1, 1}, {8, 1, 1, 1}};
```
pfultz2, line 43: It would be better to create the shape inline.

test/verify/test_shape_alloc.cpp
```diff
@@ -0,0 +1,62 @@
...
+        auto lit135  = migraphx::generate_literal(shape135);
+        auto main135 = mm->add_literal(lit135);
```
pfultz2, line 45: Do `mm->add_literal(migraphx::generate_literal(migraphx::shape{migraphx::shape::float_type, {11, 8, 1, 1}, {8, 1, 1, 1}}))`. 

Also name the variable `weights`.

test/verify/test_shape_alloc.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+        migraphx::shape s413{migraphx::shape::float_type, {1, 8, 7, 7}};
+        auto main413 = mm->add_parameter("x413", s413);
```
pfultz2, line 48: Just name the parameter `x`(and the variable as well). And pass the shape inline `mm->add_parameter("x", migraphx::shape{migraphx::shape::float_type, {1, 8, 7, 7}})`.

test/verify/test_shape_alloc.cpp
```diff
@@ -0,0 +1,62 @@
...
+        auto main414 =
+            mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 2, 3, 1}}}),
+                                main413); //  -> float_type, {1, 7, 7, 8}, {392, 7, 1, 49}
+        auto main416 = mm->add_instruction(migraphx::make_op("reduce_mean", {{"axes", {1, 2}}}),
+                                           main414); //  -> float_type, {1, 1, 1, 8}, {8, 8, 8, 1}
+        auto main417 =
+            mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {0, 3, 1, 2}}}),
+                                main416); //  -> float_type, {1, 8, 1, 1}, {8, 1, 8, 8}
+        auto convo_op419 = migraphx::make_op("convolution");
```
pfultz2, line 57: Just name this `conv_op`, no need for random numbers in the variable names.

test/verify/test_shape_alloc.cpp
bpickrel: done, all cosmetic comments

pfultz2 (approved):

PR #1422, opened 2022-10-25
Regenerate driver models

kahmed10:
`use_dynamic_same_auto_pad` was removed from convolution, but the driver models still retain the fields. This PR regenerates the files so that they are compatible again.


CharlieL7:
Would we need to have the ONNX files in the repository? I've been using the same script, but it doesn't look like we have it in `tools/`.

pfultz2:
The onnx files are big so we shouldn't store them in the repo, but we could have a script to download them.

pfultz2 (approved):

CharlieL7 (approved):

PR #1423, opened 2022-10-25
Dyn ref multibroadcast; dyn binary

CharlieL7:
* Updated Multibroadcast op to have a two input version for dynamic shapes
  * Can pass a vector of `dynamic_dimension` to the two input version in case it was precalculated like in `add_common_op()`
* Current dynamic shape broadcasting logic
  * `dynamic_dimensions` must be the same or one of them is `{1, 1, 0}` or `{1, 1, 1}`
    * If the `dynamic_dimension` is fixed, don't check the `opt` in `operator==`
    * Set `opt` to 0 in broadcasted shape
    * Note: Probably going to change how `opt` works in the future to allow for multiple optimal values
  * Works for dyn-dyn, dyn-static, and static-static shape combinations
* Changed `common.cpp` for multibroadcasting for binary ops with dynamic shapes
* Extended `binary.hpp` for dynamic shapes to test the new `common.cpp` stuff 
* Misc. changes:
  * `parse_batchnorm()` prepared to handle dynamic shapes
  * ~small changes to `broadcast` before having dynamic two input version~ (moved to other PR)


src/shape.cpp
```diff
@@ -464,15 +504,11 @@ bool shape::dynamic_dimension::is_fixed() const { return this->min == this->max;
...
-{
-    return pack(f(self.min, "min"), f(self.max, "max"), f(self.opt, "opt"));
-}
-
 bool operator==(const shape::dynamic_dimension& x, const shape::dynamic_dimension& y)
 {
-    return (x.min == y.min and x.max == y.max and x.opt == y.opt);
+    // don't check opt if both are fixed
+    bool check_opt = not(x.is_fixed() and y.is_fixed());
+    return (x.min == y.min and x.max == y.max and (check_opt ? x.opt == y.opt : true));
```
bpickrel, line 511: could use     ...and ((x.is_fixed() and y.is_fixed()) or (x.opt == y.opt))

src/common.cpp
```diff
@@ -43,32 +44,75 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+    if(not s0.dynamic())
+        s0 = s0.to_dynamic();
+    if(not s1.dynamic())
+        s1 = s1.to_dynamic();
+
+    if(s0.ndim() > s1.ndim())
+    {
+        std::swap(s0, s1);
+    }
+    auto offset = s1.ndim() - s0.ndim();
```
pfultz2, line 82: Do we need to handle tensors with different ndim?

src/common.cpp
```diff
@@ -43,32 +44,75 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+        s0.dyn_dims().cbegin(),
+        s0.dyn_dims().cend(),
+        s1.dyn_dims().cbegin() + offset,
+        out_dims.begin() + offset,
+        [&](auto a, auto b) {
+            if(a == b)
+            {
+                return a;
+            }
+            else if(contains(one_dyn_dims, a) or contains(one_dyn_dims, b))
```
pfultz2, line 95: Maybe we should add equality operators with `std::size_t` so we can just write `a == 1`:

```cpp
friend bool operator==(const dynamic_dimension& x, const std::size_t& y)
{
    return x.min == y and x.max == y;
}
friend bool operator==(const std::size_t& x, const dynamic_dimension& y)
{
    return y == x;
}
friend bool operator!=(const dynamic_dimension& x, const std::size_t& y)
{
    return !(x == y);
}
friend bool operator!=(const std::size_t& x, const dynamic_dimension& y)
{
    return !(x == y);
}
```

test/ref_ops_test.cpp
```diff
@@ -1860,6 +1884,32 @@ TEST_CASE(div_test)
     EXPECT(migraphx::verify_range(results_vector, gold));
 }
 
+TEST_CASE(div_dynamic_test)
```
bpickrel, line 1887: style: other tests have names with "_dyn_"

src/common.cpp
```diff
@@ -43,32 +44,75 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+            {
+                // setting opt to 0, may need to be changed
+                return shape::dynamic_dimension{std::max(a.min, b.min), std::max(a.max, b.max), 0};
```
pfultz2, line 98: The `std::max(a.min, b.min)` seems like a typo.

We should first take the dimensions from the fixed dimension if available:

```cpp
if (a.is_fixed())
    return a;
else if (b.is_fixed())
    return b;
else
    shape::dynamic_dimension{std::min(a.min, b.min), std::max(a.max, b.max), 0}
```

But I wonder if the last `else` should maybe be an error instead.

test/op_shape_test.cpp
```diff
@@ -1114,6 +1114,162 @@ TEST_CASE(multibroadcast)
...
+            migraphx::make_op("multibroadcast"),
+            a_shape,
+            b_shape);
+        expect_shape(
+            migraphx::shape{migraphx::shape::float_type, {{1, 4, 0}, {4, 4, 0}, {4, 4, 0}}},
+            migraphx::make_op("multibroadcast"),
+            b_shape,
+            a_shape);
+    }
+    {
```
pfultz2, line 1135: Make seperate test cases for each of these.

src/shape.cpp
```diff
@@ -244,6 +265,15 @@ const std::vector<std::size_t>& shape::lens() const { return impl->m_lens; }
 
 const std::vector<std::size_t>& shape::strides() const { return impl->m_strides; }
 
+std::size_t shape::ndim() const
```
pfultz2, line 268: Add a shape unit test for this method.

src/shape.cpp
```diff
@@ -437,6 +467,16 @@ shape shape::with_type(type_t t) const
     return {c};
 }
 
+shape shape::to_dynamic() const
```
pfultz2, line 470: Add a shape unit test for this.

src/common.cpp
CharlieL7: Probably, I know I use this property in for the `BatchNormalization` ONNX parser.

src/common.cpp
CharlieL7: This is the case where one of the shapes (after it is converted to a dynamic shape if it was a static shape) has a fixed dynamic dimension of 1. In that case, it should just broadcast to the other dynamic dimension. Which `std::max` should do. I don't think it should prefer a fixed dynamic_dimension.

ex:
a = {1, 4, 2}   b = {1, 1, 0}   ->  {1, 4, 0}

bpickrel (approved):
LGTM

src/common.cpp
```diff
@@ -114,20 +158,63 @@ instruction_ref insert_common_op(module& m,
...
         }
-        return input;
-    });
+        if(inputs[1]->get_shape().dyn_dims() != c_dyn_dims)
+        {
+            inputs[1] = m.insert_instruction(
+                ins,
+                make_op("multibroadcast", {{"out_dyn_dims", to_value(c_dyn_dims)}}),
+                inputs[1],
+                inputs[0]);
```
umangyadav, line 190: These two `IFs` can be combined into single one. Just check if either of the inputs is not the same as `c_dyn_dims`. 

`multibroadcast` would broadcast both of them to same common shape.

test/op_shape_test.cpp
```diff
@@ -1114,6 +1114,182 @@ TEST_CASE(multibroadcast)
...
+{
+    // doesn't match on first dimension
+    migraphx::shape a_shape{migraphx::shape::float_type, {3, 6}};
+    std::vector<migraphx::shape::dynamic_dimension> b{{1, 2, 0}, {6, 6, 0}};
+    migraphx::shape b_shape{migraphx::shape::float_type, b};
+    throws_shape(migraphx::make_op("multibroadcast"), a_shape, b_shape);
+    throws_shape(migraphx::make_op("multibroadcast"), b_shape, a_shape);
+}
+
+TEST_CASE(multibroadcast_2in_dyn_dyn0)
```
umangyadav, line 1177: Add a test where `multibroadcast` has two inputs and also `out_dyn_dims` attribute.

test/shape_test.cpp
```diff
@@ -185,6 +185,31 @@ TEST_CASE(test_shape_packed)
     EXPECT(not s.broadcasted());
 }
 
+TEST_CASE(test_shape_ndim_static)
```
umangyadav, line 188: Add a unit-test for constructor of `shape` which takes `type`, `max`,`min` and `opts`.

src/onnx/parse_binary_op.cpp
```diff
@@ -57,6 +57,12 @@ struct parse_binary_op : op_parser<parse_binary_op>
                 parser.parse_value(info.attributes.at("broadcast")).at<uint64_t>();
             if(broadcasted != 0)
             {
+                if(std::any_of(
+                       args.cbegin(), args.cend(), [](auto a) { return a->get_shape().dynamic(); }))
+                {
+                    MIGRAPHX_THROW(
```
umangyadav, line 63: If you can add a test to cover this case, that would be great too.

umangyadav (changes requested):
Looks good to me, except just adding having better code coverage

src/common.cpp
CharlieL7: So there are two `if` statements right now to avoid having two `multibroadcast` calls if one of inputs is already correct. This is to mimic the behavior for the static shape stuff. The two input version of `multibroadcast` only reshapes the first input because it can only have one output.

umangyadav: Okay, but then we shouldn't be making use of  two input version call i guess.It is confusing.

CharlieL7: It has to because it needs to broadcast to the other shape at model evaluation time.
example:
shape a is dynamic [{1, 4, 0}, {4, 4, 0}, {4, 4, 0}]
shape b is fixed [4, 4]

* multibroadcast(a -> b) is not required, because it would do nothing
* only multibroadcast(b -> a) is required
* however b could broadcast to any shape in dynamic range of a: [1, 4, 4] to [4, 4, 4]

umangyadav (approved):

src/common.cpp
```diff
@@ -43,32 +44,75 @@ inline namespace MIGRAPHX_INLINE_NS {
...
 
+std::vector<shape::dynamic_dimension> compute_broadcasted_dyn_dims(shape s0, shape s1)
+{
+    assert(s0.dynamic() or s1.dynamic());
+    // change both shapes to dynamic_dimension representation
+    if(not s0.dynamic())
+        s0 = s0.to_dynamic();
+    if(not s1.dynamic())
+        s1 = s1.to_dynamic();
+
```
kahmed10, line 77:
```suggestion
    // change both shapes to dynamic_dimension representation
    s0 = s0.to_dynamic();
    s1 = s1.to_dynamic();
```

kahmed10: This returns a vector of dynamic dimensions, so I don't think the assert is needed. Assuming the user knows when to use the intended function anyways.

kahmed10 (approved):

src/common.cpp
```diff
@@ -43,32 +44,71 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+        s0.dyn_dims().cbegin(),
+        s0.dyn_dims().cend(),
+        s1.dyn_dims().cbegin() + offset,
+        out_dims.begin() + offset,
+        [&](auto a, auto b) {
+            if(a == b)
+            {
+                return a;
+            }
+            else if(a == one_dyn_dim or b == one_dyn_dim)
```
pfultz2, line 91: I still think this should be written as `a == 1 or b == 1`.

PR #1424, opened 2022-10-27
Dynamic broadcast

CharlieL7:
* Two input version of the broadcast operator to handle dynamic shapes
* Added comments to describe the versions of the broadcast operator
* Dynamic broadcast only handles broadcasting a static 1D shape tensor into the other input shape
* Unsure where this operator is used in the code
  * Seems to be used to broadcast a 1D input literal


umangyadav (approved):

src/include/migraphx/op/broadcast.hpp
```diff
@@ -55,37 +61,78 @@ struct broadcast
...
+        else
         {
-            MIGRAPHX_THROW("BROADCAST: when broadcasting, succeeding sizes must match");
-        }
-        std::copy(s0.strides().begin(), s0.strides().end(), bcast_strides.begin() + axis);
+            // two inputs
+            auto s1 = inputs.at(1);
+            if(s0.dynamic())
+            {
+                MIGRAPHX_THROW("BROADCAST_2in: s0 is a dynamic shape, does not handle broadcasting "
```
pfultz2, line 100: Why `BROADCAST_2in`?

CharlieL7: Marks that it's the two input version that's throwing the error. Not sure if it matters.

src/include/migraphx/op/broadcast.hpp
```diff
@@ -55,37 +61,78 @@ struct broadcast
...
+            }
+            if(axis >= s1.ndim())
+            {
+                MIGRAPHX_THROW("BROADCAST_2in: axis is out of range");
+            }
+            if(s1.dynamic())
+            {
+                s0 = s0.to_dynamic();
+                if(s0.dyn_dims()[0] != s1.dyn_dims()[axis])
+                    MIGRAPHX_THROW("BROADCAST_2in: s0 length doesn't match with dynamic s1 axis "
```
bpickrel, line 116: If this and all the similar exception messages included the bad dimension lengths, it would make it much easier for users to troubleshoot the error.

src/include/migraphx/op/broadcast.hpp
```diff
@@ -55,37 +61,86 @@ struct broadcast
...
-            MIGRAPHX_THROW("BROADCAST : axis is out of range");
-        }
+            // the ONNX broadcast op is deprecated now, so not handling the negative
+            // value of axis anymore
+            if(axis >= broadcast_lens.size())
+            {
+                MIGRAPHX_THROW("BROADCAST : axis " + migraphx::to_string(axis) +
+                               " is out of range");
+            }
+            if(broadcast_lens.size() - axis < s0.lens().size())
```
bpickrel, line 76: Will this ever give a negative result?  Looks like you're doing subtraction between unsigned integers

test/op_shape_test.cpp
```diff
@@ -118,6 +118,69 @@ TEST_CASE(broadcast)
...
+    throws_shape(migraphx::make_op("broadcast", {{"axis", 2}}), a_input, b_input);
+}
+
+TEST_CASE(broadcast_2in_dyn_s0_ndim_greater_than_1_error)
+{
+    migraphx::shape a_input{migraphx::shape::float_type, {4, 2}};
+    migraphx::shape b_input{migraphx::shape::float_type, {{1, 4, 0}, {4, 4, 0}, {2, 2, 0}}};
+    throws_shape(migraphx::make_op("broadcast", {{"axis", 0}}), a_input, b_input);
+}
+
```
bpickrel, line 183: Add test case for negative value of axis (should throw).

src/include/migraphx/op/broadcast.hpp
CharlieL7: The previous check should prevent that.

test/op_shape_test.cpp
CharlieL7: Not handled in the previous code, should just be invalid code. Could add an assert.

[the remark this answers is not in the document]
CharlieL7: It's possible to do that for the 1 input version. The two input version is limited to broadcasting 1D things.

bpickrel (approved):
LGTM.  My two comments that are still unresolved are optional, as far as I'm concerned.

PR #1427, opened 2022-10-27
Use minimum block size of 64 threads 

umangyadav:
LocalThreads that are not multiple of 64 are causing correctness issues.

### DIFF
--- a/test/verify/test_avg_pooling_2d.cpp
+++ b/test/verify/test_avg_pooling_2d.cpp
@@ -1,0 +1,42 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


test/verify/test_avg_pooling_2d.cpp
```diff
@@ -0,0 +1,42 @@
...
+struct test_avg_pooling_2d : verify_program<test_avg_pooling_2d>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        auto input =
+            mm->add_parameter("x", migraphx::shape{migraphx::shape::float_type, {1, 336, 20, 20}});
+        auto op = migraphx::op::pooling{
+            migraphx::op::pooling_mode::average, {0, 0, 0, 0}, {1, 1}, {20, 20}, 0, 2};
```
pfultz2, line 38: It would be better to write the test using reduce_mean, since thats the operator that is failing.

umangyadav: Done.

pfultz2 (approved):

TedThemistokleous (approved):

PR #1429, opened 2022-10-28
Add opset-13 support for parse_split

TedThemistokleous:
Changes cherry picked from parse_if PR #1325 . 

This fixes parse_split so that we can support opset-13 inputs for split where there's two parameters (data and split vector)

Added additional unit tests for this to test the new cases

### DIFF
--- a/src/onnx/parse_split.cpp
+++ b/src/onnx/parse_split.cpp
@@ -58,7 +58,10 @@ struct parse_split : op_parser<parse_split>
         }
         else if(args.size() == 2)
         {
-            assert(args[1]->can_eval());
+            if(not args[1]->can_eval())
+            {
+                MIGRAPHX_THROW("PARSE_SPLIT: invalid split input arg");
+            }
 
             auto s = args[1]->eval();
             s.visit([&](auto v) { vec_splits.assign(v.begin(), v.end()); });
--- a/src/onnx/parse_split.cpp
+++ b/src/onnx/parse_split.cpp
@@ -58,6 +58,8 @@ struct parse_split : op_parser<parse_split>
         }
         else if(args.size() == 2)
         {
+            assert(args[1]->can_eval());
+
             auto s = args[1]->eval();
             s.visit([&](auto v) { vec_splits.assign(v.begin(), v.end()); });
         }
--- a/src/onnx/parse_split.cpp
+++ b/src/onnx/parse_split.cpp
@@ -26,6 +26,7 @@
 #include <migraphx/ranges.hpp>
 #include <migraphx/make_op.hpp>
 #include <migraphx/tune_axis.hpp>
+#include <migraphx/onnx/checks.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
@@ -58,12 +59,8 @@ struct parse_split : op_parser<parse_split>
         }
         else if(args.size() == 2)
         {
-            if(not args[1]->can_eval())
-            {
-                MIGRAPHX_THROW("PARSE_SPLIT: invalid split input arg");
-            }
-
             auto s = args[1]->eval();
+            check_arg_empty(s, "Split: dynamic shape is not supported");
             s.visit([&](auto v) { vec_splits.assign(v.begin(), v.end()); });
         }
         // no split attribute, input is equally divided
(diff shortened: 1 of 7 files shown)


umangyadav (dismissed):

src/onnx/parse_split.cpp
```diff
@@ -55,12 +55,11 @@ struct parse_split : op_parser<parse_split>
...
-
-            if(std::accumulate(vec_splits.begin(), vec_splits.end(), int64_t(0)) !=
-               static_cast<int64_t>(lens[tuned_axis]))
-            {
-                MIGRAPHX_THROW("PARSE_SPLIT: sum of split attribute unequal to dim size of axis!");
-            }
+        }
+        else if(args.size() == 2)
+        {
+            auto s = args[1]->eval();
```
umangyadav, line 61: I think it should first check by calling `can_eval()` otherwise it would return invalid argument in case it is not compile time constant.

TedThemistokleous: valid. I guess adding it as part of the if check if there's a second argument?

src/onnx/parse_split.cpp
```diff
@@ -55,12 +55,13 @@ struct parse_split : op_parser<parse_split>
         {
             literal s = parser.parse_value(info.attributes.at("split"));
             s.visit([&](auto v) { vec_splits.assign(v.begin(), v.end()); });
+        }
+        else if(args.size() == 2)
+        {
+            assert(args[1]->can_eval());
```
umangyadav, line 61: instead of `assert`, i think`MIGRAPHX_THROW` is better. `assert` would only kick if it is debug build.

umangyadav (approved):

src/onnx/parse_split.cpp
TedThemistokleous: I guess we want to signal that this is invalid onnx, I can change to throw

src/onnx/parse_split.cpp
```diff
@@ -55,12 +55,16 @@ struct parse_split : op_parser<parse_split>
...
+        }
+        else if(args.size() == 2)
+        {
+            if(not args[1]->can_eval())
             {
-                MIGRAPHX_THROW("PARSE_SPLIT: sum of split attribute unequal to dim size of axis!");
+                MIGRAPHX_THROW("PARSE_SPLIT: invalid split input arg");
             }
+
+            auto s = args[1]->eval();
```
pfultz2, line 66: Use `check_arg_empty(s, "Split: dynamic shape is not supported")`. Like what we do in `parse_reshape`.

src/onnx/parse_split.cpp
```diff
@@ -55,12 +55,16 @@ struct parse_split : op_parser<parse_split>
         {
             literal s = parser.parse_value(info.attributes.at("split"));
             s.visit([&](auto v) { vec_splits.assign(v.begin(), v.end()); });
-
-            if(std::accumulate(vec_splits.begin(), vec_splits.end(), int64_t(0)) !=
-               static_cast<int64_t>(lens[tuned_axis]))
+        }
+        else if(args.size() == 2)
+        {
+            if(not args[1]->can_eval())
```
pfultz2, line 61: Dont need for `can_eval` when calling `check_arg_empty`.

TedThemistokleous: done

src/onnx/parse_split.cpp
TedThemistokleous: done

src/onnx/parse_split.cpp
```diff
@@ -74,6 +75,23 @@ struct parse_split : op_parser<parse_split>
             vec_splits.resize(info.num_outputs, dl);
         }
 
+        if(std::accumulate(vec_splits.begin(), vec_splits.end(), int64_t(0)) !=
+           static_cast<int64_t>(lens[tuned_axis]))
+        {
+            std::string output;
+            for(const auto& index : vec_splits)
```
kahmed10, line 82: can you include stringutils.hpp and try `to_string_range(vec_splits)` instead of the for loop?

TedThemistokleous: On it

TedThemistokleous: Done

kahmed10 (approved):

PR #1432, opened 2022-10-28
Dynamic reduce

bpickrel:
Implements dynamic shapes in reduce_op and all its child operator classes (reduce_max etc.)

Created 1 test in op_shape_test.cpp and 1 test in ref_ops_test.cpp

- what other tests are needed?
- are there any new parameter checks or other housekeeping required in normalize_compute_shape() where most of the implementation happens?


src/include/migraphx/op/reduce_op.hpp
```diff
@@ -107,16 +108,40 @@ struct reduce_op : op_name<Derived>
...
-        for(auto axis : tuned_axes)
+        check_shapes{inputs, *this, true}.has(1);
+        auto s = inputs.at(0);
+        if(s.dynamic())
         {
-            lens[axis] = 1;
+            std::vector<shape::dynamic_dimension> output_dyn_dims = {};
+
+            // create a dynamic dimensions vector that leaves out any axis named in this->axes.
+            for(size_t index = 0; index < s.dyn_dims().size(); index++)
```
pfultz2, line 118: Use `std::copy_if`.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -107,16 +108,40 @@ struct reduce_op : op_name<Derived>
...
+            {
+                auto name_it = std::find_if(this->axes.begin(), this->axes.end(), [&](auto& axis) {
+                    return (axis == index); // if the dim is in this op's axes list, don't include
+                                            // it
+                });
+
+                if(name_it == this->axes.end())
```
pfultz2, line 125: Use `contains(this->axes, index)`.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -107,16 +108,40 @@ struct reduce_op : op_name<Derived>
...
-        auto tuned_axes = tune_axes(lens.size());
-        for(auto axis : tuned_axes)
+        check_shapes{inputs, *this, true}.has(1);
+        auto s = inputs.at(0);
+        if(s.dynamic())
         {
-            lens[axis] = 1;
+            std::vector<shape::dynamic_dimension> output_dyn_dims = {};
+
+            // create a dynamic dimensions vector that leaves out any axis named in this->axes.
```
pfultz2, line 117: We shouldn't leave out the axes. The output should be the same rank as the input.

pfultz2: I assume you would do:

```cpp
auto dyn_dims = s.dyn_dims();
for(auto axis : tuned_axes)
{
    dyn_dims[axis] = {1, 1};
}
```

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -107,16 +108,40 @@ struct reduce_op : op_name<Derived>
...
+            for(size_t index = 0; index < s.dyn_dims().size(); index++)
+            {
+                auto name_it = std::find_if(this->axes.begin(), this->axes.end(), [&](auto& axis) {
```
pfultz2, line 120: This should use the `tuned_axes` Otherwise negative axis wont work.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -107,16 +108,40 @@ struct reduce_op : op_name<Derived>
...
+                }
+            }
+
+            // compare with what src/include/migraphx/op/convolution.hpp does:
+            return shape{s.type(), output_dyn_dims};
+        }
+        else
+        {
+            auto lens       = s.lens();
+            auto tuned_axes = tune_axes(lens.size());
```
pfultz2, line 137: This should be moved outside of the `if` statement. And it should do `tune_axes(s.ndim())` so it works for both dynamic and non-dynamic shapes.

test/op_shape_test.cpp
```diff
@@ -1396,6 +1396,19 @@ void test_reduce_ops()
         migraphx::shape input{migraphx::shape::float_type, {2, 3, 4, 5}};
         throws_shape(T{{4}}, input);
     }
+    // dynamic shape
+    {
```
pfultz2, line 1400: Make a seperate test case for each test.

src/include/migraphx/op/reduce_op.hpp
bpickrel: The onnx specification allows for both, with a flag to choose.  Are we sure we want to keep the axis (and set it to {1,1,1})?

test/op_shape_test.cpp
```diff
@@ -1396,6 +1396,19 @@ void test_reduce_ops()
...
+        migraphx::shape input{migraphx::shape::float_type, {{2, 3, 4}, {2, 4, 4}}};
+        migraphx::shape::dynamic_dimension dd0{2, 3, 4};
+        expect_shape(migraphx::shape{migraphx::shape::float_type,
+                                     std::vector<migraphx::shape::dynamic_dimension>({{2, 3, 4}})},
```
pfultz2, line 1404: The output dimension should be `{2, 3, 4}, {1, 1}`

src/include/migraphx/op/reduce_op.hpp
pfultz2: The current behavior for static shapes always keeps the dimensions, and the dynamic version should work the same. We dont support removing the dimensions.

Our operators are not the same as onnx operators. For ONNX, we insert a `squeeze` when `keepdims` is false: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/onnx/parse_reduce_op.cpp#L90

test/op_shape_test.cpp
```diff
@@ -1396,6 +1396,19 @@ void test_reduce_ops()
         migraphx::shape input{migraphx::shape::float_type, {2, 3, 4, 5}};
         throws_shape(T{{4}}, input);
     }
+    // dynamic shape
+    {
+        migraphx::shape input{migraphx::shape::float_type, {{2, 3, 4}, {2, 4, 4}}};
```
pfultz2, line 1401: This is an odd dynamic dimension. The opt is not even in the range of dynamic values. You should do `{2, 3, 2}` or `{2, 3, 3}`.

src/include/migraphx/op/argmax.hpp
```diff
@@ -56,7 +56,7 @@ struct argmax
 
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(1);
+        check_shapes{inputs, *this, true}.has(1);
```
pfultz2, line 59: This is set to support dynamic shapes, but there isn't any code to support them.

src/include/migraphx/op/reduce_op.hpp
bpickrel: It does work, even if accidentally.  A test of a negative axis is at https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/dynamic_reduce/test/op_shape_test.cpp#L1403

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -105,18 +106,39 @@ struct reduce_op : op_name<Derived>
...
-        auto lens       = s.lens();
-        auto tuned_axes = tune_axes(lens.size());
-        for(auto axis : tuned_axes)
+        check_shapes{inputs, *this, true}.has(1);
+        auto s = inputs.at(0);
+        if(s.dynamic())
         {
-            lens[axis] = 1;
+            auto output_dyn_dims = s.dyn_dims();
+            auto tuned_axes      = tune_axes(output_dyn_dims.size());
```
pfultz2, line 123: This can be moved above the `if(s.dynamic())` check. You can do `tune_axes(s.ndim())`, but I think you will need #1423.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -105,18 +106,39 @@ struct reduce_op : op_name<Derived>
...
+            for(const auto& axis : tuned_axes)
+            {
+                output_dyn_dims[axis] = {1, 1};
```
CharlieL7, line 126:
```suggestion
                output_dyn_dims[axis] = {1, 1, 0};
```
Consistency

CharlieL7 (commented):
LGTM, some cppcheck errors. 
EDIT: Looked at ONNX parser for reduce, will also need updates

src/include/migraphx/op/reduce_op.hpp
bpickrel: Paul suggested the code with only 2 arguments.  Does it make a difference?

CharlieL7: No, it does not make a difference since `opt` is initialized to 0.

src/include/migraphx/op/reduce_op.hpp
bpickrel: Not sure what you mean here...there is no ndim() in branch dyn_ref_multibroadcast either.  In any case, the idference would be insignificant.

CharlieL7: He meant that using `shape::ndims()` from my PR could simplify the code by moving the `tuned_axes` call outside of the if-else statement. Not a major issue either way.

bpickrel: I couldn't find the definition of that function in the referenced branch, so let's just leave it as is for now.

CharlieL7:
The ONNX parser for `reduce`, `parse_reduce_op.cpp` will likely need some changes to handle a dynamic shape input. Try making an ONNX test with the reduce operator and a dynamic shape input.

test/ref_ops_test.cpp
```diff
@@ -4992,6 +4992,27 @@ TEST_CASE(reduce_max_axis0)
     EXPECT(results_vector == gold);
 }
 
+TEST_CASE(reduce_max_dynamic_axis0)
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape s{migraphx::shape::float_type, {{2, 4, 2}, {3, 5, 3}}};
+    auto input = mm->add_parameter("X", s);
+    std::vector<float> input_data{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12};
```
umangyadav, line 5001: just as a convention, since this is not used to build and compile the program, move this declaration of `input_data` where you are setting `params` later.

test/ref_ops_test.cpp
```diff
@@ -4992,6 +4992,27 @@ TEST_CASE(reduce_max_axis0)
...
+    auto reduce_max_op = migraphx::make_op("reduce_max", {{"axes", {0}}});
+    mm->add_instruction(reduce_max_op, input);
+    p.compile(migraphx::ref::target{});
+
+    migraphx::parameter_map params0;
+    migraphx::shape input_fixed_shape0{migraphx::shape::float_type, {2, 5}};
+    params0["X"] = migraphx::argument(input_fixed_shape0, input_data.data());
```
umangyadav, line 5008: remove suffix `0` where unnecessary.

umangyadav:
@bpickrel  are you planning to push changes for the onnx parser of reduce ops in this PR itself ?

bpickrel:
> 

No, because I found that onnx parsing of reduce ops uses the squeeze operator which doesn't support dynamic shapes yet.  Charlie has a PR in progress for that now.  I will revert my onnx parsing changes for now, since they won't work.

bpickrel:
Onnx parsing update and test are complete, and this PR now passes all tests.    Requesting someone to review the content of the two ONNX tests I added:  are they sufficient?

test/onnx/gen_onnx.py
```diff
@@ -4569,6 +4569,22 @@ def reduce_log_sum_exp_test():
...
+                                 axes=axes,
+                                 keepdims=0)
+
+    return ([node], [x], [y])
+
+
+@onnx_test
+def reducemax_dyn_test():
+    x = helper.make_tensor_value_info('x', TensorProto.FLOAT, [None, 4, 5, 6])
+    y = helper.make_tensor_value_info('y', TensorProto.FLOAT, [3, 4, 6])
```
CharlieL7, line 4587: Missing reducel1_dyn_test?

test/onnx/onnx_test.cpp
```diff
@@ -4311,6 +4311,26 @@ TEST_CASE(reducel1_test)
...
+    auto sq_l0  = mm->add_instruction(migraphx::make_op("squeeze", {{"axes", {-2}}}), sum_l0);
+    mm->add_return({sq_l0});
+
+    migraphx::onnx_options options;
+    options.map_dyn_input_dims["x"] = {{3, 3}, {3, 5}, {4, 6, 5}, {5, 7, 6}};
+    auto prog                       = migraphx::parse_onnx("reducel1_dyn_test.onnx", options);
+
+    EXPECT(p == prog);
+}
+
```
CharlieL7, line 4333: Why is there an `abs` instruction?

test/onnx/gen_onnx.py
CharlieL7, line 4587: This output should have the shape `[None, 4, 6]`. Doesn't matter to MIGX, however.

test/onnx/onnx_test.cpp
bpickrel: I don't know.  The instructions were copied from the previous test, `reducel1_test`

bpickrel: Removed the abs--wait to see if test result is the same

bpickrel: Abs turns out to be a necessary part of L1_reduce.  Restored.

CharlieL7 (approved):

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -105,26 +106,47 @@ struct reduce_op : op_name<Derived>
...
+        check_shapes{inputs, *this, true}.has(1);
+        auto s = inputs.at(0);
+        if(s.dynamic())
         {
-            lens[axis] = 1;
+            auto output_dyn_dims = s.dyn_dims();
+            auto tuned_axes      = tune_axes(output_dyn_dims.size());
+            for(const auto& axis : tuned_axes)
+            {
+                output_dyn_dims[axis] = {1, 1, 0};
```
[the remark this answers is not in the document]
CharlieL7: Zero for `opt` does mean no optimal value right now. It doesn't make a difference for a fixed dynamic dimension. I'm thinking we'll probably want to refactor how `opt` works in the future to allow for multiple. The other option is to allow for multiple dynamic dimension ranges like what TRT does with optimization profiles.

test/ref_ops_test.cpp
```diff
@@ -5731,6 +5731,27 @@ TEST_CASE(reduce_max_axis0)
...
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape s{migraphx::shape::float_type, {{2, 4, 2}, {3, 5, 3}}};
+    auto input         = mm->add_parameter("X", s);
+    auto reduce_max_op = migraphx::make_op("reduce_max", {{"axes", {0}}});
+    mm->add_instruction(reduce_max_op, input);
+    p.compile(migraphx::ref::target{});
+
+    migraphx::parameter_map params;
+    migraphx::shape input_fixed_shape0{migraphx::shape::float_type, {2, 5}};
```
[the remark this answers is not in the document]
bpickrel: Will push the change.

test/onnx/gen_onnx.py
```diff
@@ -4623,6 +4623,21 @@ def reducel1_test():
     return ([node], [x], [y])
 
 
+@onnx_test
+def reducel1_dyn_test():
```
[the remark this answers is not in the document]
bpickrel: What result is expected in this case?  I tried it with the result:
`main:@1 = abs(x) -> dynamic, float_type, {[3, 3, 0], [3, 5, 0], [4, 6, 5], [5, 7, 6]}
main:@2 = reduce_sum[axes={0, 1, 2, 3}](main:@1) -> dynamic, float_type, {[1, 1, 0], [1, 1, 0], [1, 1, 0], [1, 1, 0]}
main:@3 = squeeze[axes={0, 1, 2, 3}](main:@2) -> float_type, {}, {}
main:@4 = @return(main:@3)`

In other words, the onnx parser creates an instruction that explicitly names all the axes when it parses a file with no axes arguments.  (It does the same whether an empty list is specified, or no "axes" argument at all is given.)  This does not match a test case program that actually has no axes argument.  Is that what we want to see?

test/op_shape_test.cpp
```diff
@@ -1779,9 +1779,42 @@ void test_reduce_ops()
...
+        migraphx::shape input{migraphx::shape::float_type, {{2, 3, 3}, {2, 4, 4}}};
+        expect_shape(migraphx::shape{migraphx::shape::float_type,
+                                     std::vector<migraphx::shape::dynamic_dimension>(
+                                         {{1, 1, 0}, {2, 4, 4}})},
+                     T{{0}},
+                     input);
+    }
+    {
+        migraphx::shape input{migraphx::shape::float_type, {{2, 3, 3}, {2, 4, 4}}};
+        throws_shape(T{{4}}, input);
```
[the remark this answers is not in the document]
bpickrel: see next conversation.

test/onnx/gen_onnx.py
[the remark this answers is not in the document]
bpickrel: We agreed it's correct to make the test match the `axes={0, 1, 2, 3}` argument, which is what the Onnx parser inserts by default when no axes are given.

umangyadav (approved):

pfultz2 (approved):

PR #1433, opened 2022-10-28
Fix parse if inline if

TedThemistokleous:
Inline if based on eval input

Fix tests that break with this. Bit of a pain but allows us to fold the appropriate branch at compile time instead of having to evaluate it at inference.


TedThemistokleous:
Rebase of latest changes.

TedThemistokleous:
Based off #1325 which isn't needed for now due to dynamic rank. Superseded by #1533

PR #1442, opened 2022-11-02
Include timestamp while tracing

causten:
Adds a timestamp to tracing of passes.  Prints both seconds and microseconds to catch rollovers

Example...
 MIGRAPHX_TRACE_COMPILE=1 migraphx-driver perf resnet50_v1.onnx |grep Pass

Module: main, Pass: dead_code_elimination
Pass: dead_code_elimination completed in (s): 0.0072007
Pass: dead_code_elimination
Module: main, Pass: gpu::compile_miopen
Pass: gpu::compile_miopen completed in (s): 17.9621
Pass: gpu::compile_miopen
Module: main, Pass: dead_code_elimination
Pass: dead_code_elimination completed in (s): 0.00857062
Pass: dead_code_elimination
Module: main, Pass: gpu::pack_int8_args
Pass: gpu::pack_int8_args completed in (s): 0.00850617


shivadbhavsar:
Is the main use case for this to measure the time it's taking for each pass?
If so, it might be easier if we capture the time at the beginning and end of run_pass and just do the math to output the time taken so that it doesn't need to be done manually from these timestamps.

src/pass_manager.cpp
```diff
@@ -30,6 +30,7 @@
 #include <migraphx/ranges.hpp>
 #include <migraphx/time.hpp>
 #include <migraphx/iterator_for.hpp>
+#include <sys/time.h>
```
kahmed10, line 33:
```suggestion
#include <sys/time.h>
```

src/pass_manager.cpp
```diff
@@ -94,11 +95,18 @@ struct module_pm : module_pass_manager
     virtual void run_pass(const pass& p) override
     {
         assert(mod);
+
```
kahmed10, line 98:
```suggestion
timer t{};
```

src/pass_manager.cpp
```diff
@@ -94,11 +95,18 @@ struct module_pm : module_pass_manager
...
         trace("Module: ", mod->name(), ", Pass: ", p.name());
+        const auto start = std::chrono::system_clock::now();
```
kahmed10, line 100:
```suggestion
        const auto start = t.record<seconds>();
```

src/pass_manager.cpp
```diff
@@ -94,11 +95,18 @@ struct module_pm : module_pass_manager
...
         assert(mod->validate() == mod->end());
         p.apply(*this);
         trace(*mod);
         validate_pass(*mod, p, *t);
+
+        const auto end = std::chrono::system_clock::now();
```
kahmed10, line 106:
```suggestion
        const auto end = t.record<seconds>();
```

kahmed10:
Paul had implemented a timer for this feature that you might want to try. I've added suggested code changes

kahmed10 (approved):

shivadbhavsar (approved):

pfultz2:
See: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1380/files#diff-0b6c8d83337d2bc89577573dcd7298fcbc76f205106bbce3993d522ff841ee83R100

PR #1443, opened 2022-11-02
Refactor non-standard literal construction

CharlieL7:
* I was working on the contiguous operator and found oddness with constructing non-standard shape literals
  * a non-standard literal will almost never be used, since a literal is known at compile time

* Added some comments on the intended behavior:
  * `literal{shape, vector}` constructor with a non-standard shape is intended to keep the same ordering as the given vector. The data buffer will be populated such that when the non-standard indexing is used the original order is as given.
  * `literal{shape, argument}` constructor directly copies the data buffer from the argument 

* Changed non-standard literal `fill()` to use `tensor_view` iterators as it handles non-standard shapes now

* Changed the contiguous `ref_ops_test` to be more helpful

* ~I think the previous code wanted to make a non-standard shape literal into a standard shape literal during construction of the literal~
* ~The conversion did not work as expected:~
  * ~a change to `shape_for_each()` made it such that the buffer was indexed with standard strides~
  * ~the literal's shape was not changed after it's buffer was changed, keeping the non-standard strides~
* ~This PR removes the attempted conversion to a standard shape and changes some tests~
  * ~I think the compiler passes should handle the non-standard shape literal case, making it a standard shape~
  * ~If the passes do not, then it is possible to implement the conversion correctly~

* ~When using `add_parameter()`  and `argument` instead of `literal` with a non-standard shape the `contiguous_test` works as I expected. Where the original buffer `{0, 1, 2, ..., 11}` is changed to `{0, 3, 6, 9, ...}`. This is instead of the weirdness in the test where the buffer doesn't change at all with the contiguous call.~


src/include/migraphx/literal.hpp
```diff
@@ -111,21 +111,7 @@ struct literal : raw_data<literal>
...
-        }
-        else
-        {
-            auto it = start;
-            m_shape.visit_type([&](auto as) {
-                auto output = make_view(m_shape, as.from(buffer.get()));
-                shape_for_each(output.get_shape(), [&](const auto& idx) {
-                    output(idx.begin(), idx.end()) = *it; // NOLINT(bugprone-signed-char-misuse)
-                    it++;
-                });
```
umangyadav, line 126: From what i understand, this is not trying to make it a standard shape . Input vector from users would be provided in a standard shape. This is trying to fill the buffer such that, it puts standard shape vector input elements into non-standard shape migraphx_buffer in a correct order.  

`shape_for_each` is iterating using standard shape because input vector values are read in standard shape order. 
tensor_view `output` is assigning them into non-standard shape according to `m_shape`. tensor_view has `operator()` which calculates the correct element index using non-standard shape strides.

kahmed10: I can't remember why this needed to be implemented a few years ago. But many changes have been made to the use of literals in GPU kernel implementations, so it could be fine simplifying to what Charlie has implemented. Unless certain models are failing, then we can revisit.

kahmed10 (approved):

src/include/migraphx/literal.hpp
TedThemistokleous: thats been my apprehension since its a somewhat large block being removed with the standard case. Maybe its worth running this with DeepLearningModels library to see if we have any regressions?

TedThemistokleous:
@CharlieL7  I've opened a PR on DLM thats going to run CI for those models. Just changed target MIGraphX branch for this

Relevant PR: https://github.com/ROCmSoftwarePlatform/DeepLearningModels/pull/499

Will approve once this completes

TedThemistokleous (approved):
Ran this through DLM, don't see anything breaking.

test/literal_test.cpp
```diff
@@ -49,6 +49,20 @@ TEST_CASE(literal_test)
...
 }
 
+TEST_CASE(literal_nstd_shape)
+{
+    migraphx::shape nstd_shape{migraphx::shape::float_type, {1, 3, 2, 2}, {12, 1, 6, 3}};
+    std::vector<float> nstd_data(12);
+    std::iota(nstd_data.begin(), nstd_data.end(), 0);
+
+    migraphx::shape std_shape{migraphx::shape::float_type, {1, 3, 2, 2}};
+    std::vector<float> std_data = {0, 3, 6, 9, 1, 4, 7, 10, 2, 5, 8, 11};
```
umangyadav, line 59: I think you want to modify this data a bit. 
1 should come on index 3 (starting from 0th index)

CharlieL7: Here's my work for the striding:
input buffer: [0, 1,  2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
lengths: [1, 3, 2, 2]
strides: [12, 1, 6, 3]

[0, 0, 0, 0] dot [12, 1, 6, 3] = 0 buffer index
[0, 0, 0, 1] dot [12, 1, 6, 3] = 3 buffer index
[0, 0, 1, 0] dot [12, 1, 6, 3] = 6 buffer index
[0, 0, 1, 1] dot [12, 1, 6, 3] = 9 buffer index
[0, 1, 0, 0] dot [12, 1, 6, 3] = 1 buffer index
etc...

The buffer index matches with the data so it's the same.

umangyadav (approved):

CharlieL7:
Updating the literal_nstd_shape test to make it better.

umangyadav (approved):

pfultz2:
> I think the previous code wanted to make a non-standard shape literal into a standard shape literal during construction of the literal

No, thats not how that works. The literal is constructed with the elements in order. The order of elements in a tensor do not change when the strides change(only the order in memory changes).

> a change to shape_for_each() made it such that the buffer was indexed with standard strides

The `shape_for_each` iterates the indices in order. Strides are completely irrelevant here.

src/include/migraphx/literal.hpp
pfultz2: > This is trying to fill the buffer such that, it puts standard shape vector input elements into non-standard shape migraphx_buffer in a correct order.

Umang is correct here.

src/include/migraphx/literal.hpp
```diff
@@ -111,21 +111,7 @@ struct literal : raw_data<literal>
...
-            });
-        }
+        m_shape.visit_type([&](auto as) { std::copy(start, end, as.from(buffer.get())); });
```
pfultz2, line 114: If you are using this for standard and non-standard shape then it should use the iterators: 

```cpp
this->visit([&](auto v) { std::copy(start, end, v.begin()); });
```

src/include/migraphx/shape_for_each.hpp
```diff
@@ -31,6 +31,10 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+/**
+ * Iterates the given function over the standard shape indices.
+ * Will iterate using standard strides if given a non-standard shape.
```
pfultz2, line 36: This comment is very confusing. It should not say say anything about standard shape as that is irrelevent. This should just say: `Iterates the given function over the indices from the shape in order.`.

pfultz2 (changes requested):

src/include/migraphx/literal.hpp
pfultz2: > Why was literal construction made to function differently than when using a parameter?

An `argument`(which I assume is what you mean by a parameter) takes a raw buffer, not a list of values. There is a constructor to literal which will take a raw buffer as well and it will just do a simple copy:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/literal.hpp#L86

So you can convert an `argument` to a literal with `literal{arg.get_shape(), arg.data()}`.

src/include/migraphx/literal.hpp
CharlieL7: I wasn't able to use the `tensor_view` made by the `visit` call because it was `const`. I used `shape.visit_type` and `make_view` instead.

src/include/migraphx/literal.hpp
```diff
@@ -107,25 +108,15 @@ struct literal : raw_data<literal>
...
-                shape_for_each(output.get_shape(), [&](const auto& idx) {
-                    output(idx.begin(), idx.end()) = *it; // NOLINT(bugprone-signed-char-misuse)
-                    it++;
-                });
-            });
-        }
+        m_shape.visit_type([&](auto as) {
+            auto output = make_view(m_shape, as.from(buffer.get()));
+            std::copy(start, end, output.begin());
+        });
```
pfultz2, line 119: Just curious, why did `this->visit` not work?

test/ref_ops_test.cpp
```diff
@@ -848,11 +848,42 @@ TEST_CASE(contiguous_test)
     p.compile(migraphx::ref::target{});
     auto result = p.eval({}).back();
 
+    result.visit([&](auto output) {
+        std::vector<size_t> new_strides = {12, 4, 2, 1};
+        EXPECT(bool{output.get_shape().strides() == new_strides});
```
pfultz2, line 853: Why do you do a visit here? Why not just `EXPECT(result.get_shape().strides() == new_strides)`?

src/include/migraphx/literal.hpp
CharlieL7: Looks like the `basic_iota_iterator` object was made constant so I could not update the values with `std::copy`.

pfultz2 (approved):

PR #1445, opened 2022-11-03
Dynamic ref contiguous

CharlieL7:
* Extends the ref contiguous operator to handle dynamic shapes
* Updates the `eliminate_contiguous` pass to use the `dyn_output` struct
  * Also should skip over dynamic shapes (not tested on ref, will revisit when working on GPU)


kahmed10 (approved):

src/include/migraphx/op/contiguous.hpp
```diff
@@ -42,19 +43,31 @@ namespace op {
...
+            return s0;
+        }
+        else
+        {
+            if(s0.standard())
+            {
+                return inputs.front();
+            }
+            auto lens = inputs.at(0).lens();
+            auto t    = inputs.at(0).type();
```
pfultz2, line 62: you could use `s0` here as well instead of `inputs.at(0)`.

pfultz2 (approved):

PR #1449, opened 2022-11-08
Dynamic ref pooling

CharlieL7:
* Extends the `pooling` operators for dynamic shape inputs
  * `AveragePooling`
  * `GlobalAveragePooling`
  * `MaxPooling`
  * `GlobalMaxPooling`
  * `LpNormPooling`
  * `GlobalLpNormPooling`
* Updates the `pooling` ONNX parser
* Dynamic spatial data with the "Global" type of pooling operator requires a new flag attribute: `dyn_global`

* Does not handle `auto_pad` attribute for dynamic shapes at this time
* Does not handle asymmetrical padding for dynamic shapes at this time
  * asymmetrical padding uses some stuff with the `slice` operator that hasn't been updated


bpickrel (commented):
LGTM

test/ref_ops_test.cpp
```diff
@@ -565,111 +565,161 @@ TEST_CASE(atanh_dynamic_test)
...
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> gold{
+        0.908,     0.250625,  0.795,     0.40425, 0.711875,  0.194875,  0.014125,  0.09425,
+        -0.078375, 0.139375,  0.46075,   0.0285,  -0.188125, -0.085,    0.378125,  -0.085375,
+        -0.04,     0.304125,  0.40775,   0.2835,  0.112375,  -0.073375, 0.4355,    -0.187,
+        -0.392625, -0.258375, -0.485875, -0.0345, 0.16125,   -0.131875, -0.228375, 0.068625};
+    EXPECT(migraphx::verify_range(results_vector, gold));
+}
+
+TEST_CASE(avgpool_rank3_dyn_test)
```
umangyadav, line 700: Isn't this same as `avgpool_dyn_test` ?

umangyadav: You could do similar test as fixed shape either with strides 2 or a rank 5 input test.

src/onnx/parse_pooling.cpp
```diff
@@ -159,6 +198,11 @@ struct parse_pooling : op_parser<parse_pooling>
 
         if(not slice_start.empty())
         {
+            if(in_shape.dynamic())
```
umangyadav, line 201: Would it be able to handle if input is dynamic but fixed shape ?

pfultz2: If all dimensions are fixed then we should probably make a static shape during construction.

src/include/migraphx/op/pooling.hpp
```diff
@@ -57,59 +61,115 @@ struct pooling
...
+        else
+        {
+            auto input_lens = input.lens();
 
-    size_t kdims() const
-    {
-        check_attribute_size();
-        return stride.size();
+            std::vector<std::size_t> output_lens(input_lens.begin(), input_lens.begin() + 2);
+            if(dyn_global)
```
umangyadav, line 157: So is this flag exclusively for the  dynamic shapes ? It doesn't seem so.  In that case, we should name if `global_pool`

CharlieL7: Yes, it is only for global pooling with dynamic input shapes. Since we have to use the given shape at model eval time.

umangyadav: Thanks.

src/include/migraphx/op/pooling.hpp
```diff
@@ -108,16 +107,16 @@ struct pooling
...
+                    std::size_t padding_factor = 2 * padding[i];
                     if(padding_size == 2 * kdims)
                         padding_factor = padding[i] + padding[i + kdims];
-                    dim_size = input_lens[i + 2] + padding_factor - lengths[i];
-                    assert(dim_size >= 0);
-                    std::size_t len = (ceil_mode)
-                                          ? ceil_divide<std::ptrdiff_t>(dim_size, stride[i])
-                                          : floor_divide<std::ptrdiff_t>(dim_size, stride[i]);
-                    output_lens.push_back(std::size_t(std::max<std::ptrdiff_t>(1, len + 1)));
+                    assert(input_lens[i + 2] + padding_factor > lengths[i]);
```
umangyadav, line 113: it should be `>=`

src/include/migraphx/op/pooling.hpp
```diff
@@ -57,59 +60,116 @@ struct pooling
...
-        if(input_size != padding_size / 2 + 2 and input_size != padding_size + 2)
+        auto padding_size  = padding.size();
+        size_t kdims       = input.ndim() - 2;
+        if(input.ndim() != padding_size / 2 + 2 and input.ndim() != padding_size + 2)
         {
             MIGRAPHX_THROW("POOLING: input and attribute size mismatch!");
         }
 
-        std::vector<std::size_t> output_lens(input_lens.begin(), input_lens.begin() + 2);
+        auto calc_spatial_dim_out = [&](std::vector<std::size_t> input_lens) {
```
pfultz2, line 99: Use const ref for parameter. Also I think you make it a member function. You only need to pass `kdims`. However, if this adds a lot of parameters it might be better to keep it as a lambda, and move under the `if(input.dynamic())` branch.

src/include/migraphx/op/pooling.hpp
```diff
@@ -57,59 +60,116 @@ struct pooling
...
+        else
+        {
+            auto input_lens = input.lens();
 
-    size_t kdims() const
-    {
-        check_attribute_size();
-        return stride.size();
+            std::vector<std::size_t> output_lens(input_lens.begin(), input_lens.begin() + 2);
+            if(dyn_global)
```
pfultz2, line 157: Why is this checked when the input is not dynamic? If its not dynamic then we should know if the pooling is global based on window size.

CharlieL7: Mistake, shouldn't be checked here.

CharlieL7: I remember why it's here now. It's for when the `compute_shape()` is called again at model evaluation time by the `dynamic_output` object. The kernel size is not used for that case.

PR #1450, opened 2022-11-09
fix issues with compiling lstm ops in fp16 mode

shivadbhavsar:
Currently, quantizing a program with rnn layers to fp16 results in segmentation faults due to a "convert" operation being applied to an "undefined" instruction.

The following changes are implemented to fix this issue:
- Added `is_undefined` method to the instruction class that returns true if all inputs to the instruction are from an undefined op.
- Updated rewrite_rnn pass to use the new `is_undefined` method rather than checking `ins->name()` 
- Updated the dead_code_elimination pass to also use this new method rather than only checking the instruction name

Following onnx file can be used to reproduce the issue in question:
[lstm_test.zip](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/9966297/lstm_test.zip)


src/dead_code_elimination.cpp
```diff
@@ -51,8 +51,8 @@ void dead_code_elimination::apply(module& m) const
         // Skip instruction with empty shape as output unless its [dynamic, builtin, undefined,
         // identity, allocate]
         if((not i->get_shape().dynamic() and i->get_shape().elements() == 0) and
-           i->name().front() != '@' and
-           not contains({"undefined", "identity", "allocate"}, i->name()))
+           i->name().front() != '@' and not contains({"identity", "allocate"}, i->name()) and
```
TedThemistokleous, line 54: make "!="    not (i->name().front() == '@')

TedThemistokleous (changes requested):

src/instruction.cpp
```diff
@@ -302,6 +302,24 @@ void instruction::replace_mod_argument(module_ref old, module_ref new_mod)
...
+    if(op.name() == "undefined")
+    {
+        return true;
+    }
+    else if(is_context_free(op))
+    {
+        return std::all_of(this->inputs().begin(), this->inputs().end(), [](auto arg) {
+            return arg->is_undefined();
+        });
+    }
```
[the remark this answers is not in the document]
TedThemistokleous: talked about this offline. Understand context here now.

TedThemistokleous (approved):

kahmed10:
Can you make a test for fp16 mode?

src/instruction.cpp
```diff
@@ -302,6 +302,24 @@ void instruction::replace_mod_argument(module_ref old, module_ref new_mod)
     std::replace(module_args.begin(), module_args.end(), old, new_mod);
 }
 
+bool instruction::is_undefined() const
+{
+    if(op.name() == "undefined")
+    {
+        return true;
+    }
+    else if(is_context_free(op))
```
pfultz2, line 311: Remove the `is_context_free` check here. And just do an `else` and the check all of the inputs.

src/instruction.cpp
```diff
@@ -302,6 +302,24 @@ void instruction::replace_mod_argument(module_ref old, module_ref new_mod)
     std::replace(module_args.begin(), module_args.end(), old, new_mod);
 }
 
+bool instruction::is_undefined() const
```
pfultz2, line 305: Can we add a unit test for this function?

pfultz2:
Can a rnn test be added? There should also be some unit tests for the `is_undefined` function.

kahmed10 (approved):

causten:
Is all of Paul's change request resolved?

test/ref_rnn_ops_test.cpp
```diff
@@ -932,6 +933,83 @@ TEST_CASE(rnn_bidirectional)
...
+        migraphx::make_op("rnn",
+                          {{"hidden_size", hidden_size},
+                           {"actv_func", {}},
+                           {"direction", migraphx::to_value(migraphx::op::rnn_direction::forward)},
+                           {"clip", clip}}),
+        seq,
+        w,
+        r);
+    mm->add_instruction(migraphx::make_op("rnn_last_hs_output"), out_hs);
+    migraphx::quantize_fp16(p);
```
pfultz2, line 1001: Can you just insert the converts directly instead of relying on `quantize_fp16`?

shivadbhavsar: Updated in latest commit

PR #1451, opened 2022-11-09
Update MLIR integration

jungpark-mlir:
- Update dialect registration interface
- Update 2nd build pipeline call and use full arch name
- Update hash


jungpark-mlir:
@causten , can you find someone to help fixing a link failure here please? It's not related to mlir and not sure why they fail.

```[2022-11-10T20:07:09.020Z] make[3]: Leaving directory '/var/jenkins/workspace/MLLibs_AMDMIGraphX_mlir-tot-5.5/build'

[2022-11-10T20:07:09.020Z] [ 95%] Built target header_src_targets_gpu_include_migraphx_gpu_reduce_op_hpp

[2022-11-10T20:07:09.020Z] /usr/bin/ld: /var/jenkins/workspace/MLLibs_AMDMIGraphX_mlir-tot-5.5/build/lib/libmigraphx_gpu.so.2: undefined reference to `ZSTD_getErrorName'

[2022-11-10T20:07:09.020Z] /usr/bin/ld: /var/jenkins/workspace/MLLibs_AMDMIGraphX_mlir-tot-5.5/build/lib/libmigraphx_gpu.so.2: undefined reference to `ZSTD_decompress'

[2022-11-10T20:07:09.020Z] /usr/bin/ld: /var/jenkins/workspace/MLLibs_AMDMIGraphX_mlir-tot-5.5/build/lib/libmigraphx_gpu.so.2: undefined reference to `ZSTD_isError'

[2022-11-10T20:07:09.020Z] /usr/bin/ld: /var/jenkins/workspace/MLLibs_AMDMIGraphX_mlir-tot-5.5/build/lib/libmigraphx_gpu.so.2: undefined reference to `ZSTD_compressBound'

[2022-11-10T20:07:09.020Z] /usr/bin/ld: /var/jenkins/workspace/MLLibs_AMDMIGraphX_mlir-tot-5.5/build/lib/libmigraphx_gpu.so.2: undefined reference to `ZSTD_compress'

[2022-11-10T20:07:09.020Z] collect2: error: ld returned 1 exit status

[2022-11-10T20:07:09.020Z] make[3]: *** [test/api/CMakeFiles/test_api_c_op.dir/build.make:86: bin/test_api_c_op] Error 1```

**Comment by krzysz00** [comment]
@jungpark-mlir I think LLVM has a zlib dependency that I thought we were turning off but maybe not. I'd go crawling through our build files for `ZLIB` or `LIBZ` and seeing what the deal is there.

I'll also note that this could, in theory, become less of a problem once https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/847 lands on our and and so we don't need to include `RegisterEverything` at all.

File: src/targets/gpu/perfdb.cpp [hunk 1]
```diff
@@ -27,6 +27,9 @@
 #include <migraphx/stringutils.hpp>
 #include <migraphx/permutation.hpp>
 #include <fstream>
+#include <mutex>
+
+static std::mutex g_db_mutex; // NOLINT
```
pfultz2, line 32: This variable should go in the `query_miopen_db` function.

src/targets/gpu/mlir.cpp
```diff
@@ -168,10 +180,19 @@ struct mlir_program
...
     {
+#ifdef MIGRAPHX_MLIR_REGISTER_ROCMLIR
+        MlirDialectRegistry registry = mlirDialectRegistryCreate();
+        mlirRegisterRocMLIRDialects(registry);
+        mlirRegisterAllDialects(registry);
+        mlirContextAppendDialectRegistry(ctx.get(), registry);
+        mlirContextLoadAllAvailableDialects(ctx.get());
+        mlirDialectRegistryDestroy(registry);
+        mlirContextSetAllowUnregisteredDialects(ctx.get(), true /*allow*/);
+#else
```
pfultz2, line 191: Why do we need the ifdef? We aren't supporting different MLIR versions.

jungpark-mlir: You're right, I'll just check the version and undefine `MIGRAPHX_MLIR` with warning message.

src/targets/gpu/perfdb.cpp
jungpark-mlir: Makes sense, it's already `static`.

krzysz00:
@jungpark-mlir Can you bump the commit to include @giuseros's PR that makes the library smaller?

src/targets/gpu/mlir.cpp
krzysz00: I figured we should have the ifdef to match MIOpen's practice of being able to build with slightly older librockcompiler

jungpark-mlir: This code directly integrates librockcompiler, not using MIOpen. Officially, corresponding release version is the only correct version to use.

src/targets/gpu/perfdb.cpp
```diff
@@ -88,6 +89,9 @@ std::string generate_miopen_config(const problem_params& pp)
 
 auto query_miopen_db(const std::string& query)
 {
+    static std::mutex g_db_mutex; // NOLINT
+    const std::lock_guard<std::mutex> lock(g_db_mutex);
+
```
umangyadav, line 94: why does it require lock for reading into db ?

jungpark-mlir: Reads don't have any dependency but it complains the db file is locked when a thread tries to open the  db file while it's being used.

But also, that's a good point, db  should be able to handle multiple thread, I'll try and see if it works.

jungpark-mlir:
Currently investigating ubsan error,

[   RUN    ] test_conv_add_relu
WARNING: Instruction needs normalization, performance may be affected.
[ COMPLETE ] test_conv_add_relu
[==========] 1 tests ran
/usr/lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/future:213:52: runtime error: member call on address 0x7f3ed0368010 which does not point to an object of type 'std::__future_base::_Result_base'
0x7f3ed0368010: note: object has invalid vptr
 00 00 00 00  c0 06 5b 22 3f 7f 00 00  00 00 00 00 00 00 00 00  20 00 00 00 00 00 00 00  41 4e 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/future:213:52 in

Which might be somehow related to the usage of future in test/verify/run_verify.cpp
Maybe mlir is not destroying something or .., not sure yet.
### Appreciate if anyone has helpful idea about this.

The new build failure in building mlir with cget could be caused by the recent changes, was hidden because it's not reproducible by normal Ninja build. Should be fixed easily.

src/targets/gpu/perfdb.cpp
jungpark-mlir: Tried using flags on sqlite call explicitly with various combination but no luck.

umangyadav: okay, that's fine.  I was just concerned in case it increased run time since it would do sequential reads.

jungpark-mlir:
CI passes but now I'm investigating a new crash on running resnet50 with driver

Compiling ...
Reading: /home/resnet50i1.onnx
driver: /usr/local/cget/build/tmp-62caff1dad1140498c40558248f4e141/rocMLIR-b8c0a51c7bd1b4d5a70c8389c40fb451fe8652a3/external/llvm-project/llvm/lib/Support/PrettyStackTrace.cpp:230: virtual llvm::PrettyStackTraceEntry::~PrettyStackTraceEntry(): Assertion `PrettyStackTraceHead == this && "Pretty stack trace entry destruction is out of order"' failed.
Aborted (core dumped)

Unfortunately, this happens when llvm thread support is disabled, which was a potential WA for the ubsan error.
Hope thread sanitizer has something to report.

jungpark-mlir:
The crash I mentioned has been fixed, thanks Simon.
Now, everything is in and can be merged.

I've checked the compilation time of resnet50 batch size 256 with different options.

Normal run
real    0m27.5s
user    0m51.4s
sys     0m13.6s

MIOpen fusion disabled
real    0m28.2s
user    1m13.5s
sys     0m17.7s

MLIR enabled (No difference from LLVM_ENABLE_THREADS because multl-threading has been disabled already)
real    0m19.6s
user    0m23.7s
sys     0m1.14s

src/targets/gpu/jit/mlir.cpp
```diff
@@ -41,6 +41,9 @@ struct mlir_compiler : compiler<mlir_compiler>
     {
         auto* smod = ins->module_inputs().front();
         assert(smod->get_parameter_names().size() == ins->inputs().size() - 1);
+        // set mutex while llvm thread support is disabled.
+        static std::mutex g_mlirc_mutex; // NOLINT
+        const std::lock_guard<std::mutex> lock(g_mlirc_mutex);
```
pfultz2, line 46: The lock should probably go into `compile_mlir` function.

jungpark-mlir:
There was a minor regression in the recent mlir code but fixed in the last commit.
Now everything passes and resnet50 verifies.

pfultz2 (approved):

umangyadav (approved):

jungpark-mlir:
@causten Can you please merge this? I don't have a merge permission

jungpark-mlir:
Tests failed with CI issue, 'No space left on device'.

causten:
> Tests failed with CI issue, 'No space left on device'.

I'll restart to see if there are auto cleanup processes.  If not I'll talk to the DevOps team

sjw36:
@causten , @pfultz2 looks like this is ready to go. Please commit.

PR #1452, opened 2022-11-09
Fix parse if shapes and literals

TedThemistokleous:
Fixes to get shape and literal working correclty for parse if.

- removed literal creation from parse_if and handle empty literal correctly.
- Changed literal construction such that we do this as per initializer list
- Updated unit tests which expected a variant of an empty scalar used when parsing in onnx
- Updated parse_if to scan for empty  and or scalar inputs that d

### DIFF
--- a/src/onnx/parse_constant.cpp
+++ b/src/onnx/parse_constant.cpp
@@ -43,7 +43,7 @@ struct parse_constant : op_parser<parse_constant>
         // return empty literal
         if(v.get_shape().elements() == 0)
         {
-            return info.add_literal(literal{v.get_shape().type()});
+            return info.add_literal(literal{migraphx::shape{v.get_shape().type(), {1}, {0}}, {0}});
         }
 
         auto dim_size = info.attributes.at("value").t().dims_size();
--- a/src/onnx/parse_constant_of_shape.cpp
+++ b/src/onnx/parse_constant_of_shape.cpp
@@ -68,7 +68,9 @@ struct parse_constant_of_shape : op_parser<parse_constant_of_shape>
             // empty input tensor, output is a scalar
             if(args[0]->get_shape().elements() == 0)
             {
-                s = migraphx::shape{type, {1}, {0}};
+                std::vector<size_t> lens(1, 1);
+                std::vector<size_t> strides(1, 0);
+                s = migraphx::shape(type, lens, strides);
             }
             else
             {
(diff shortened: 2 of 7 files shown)


src/onnx/parse_constant.cpp
```diff
@@ -43,7 +43,7 @@ struct parse_constant : op_parser<parse_constant>
         // return empty literal
         if(v.get_shape().elements() == 0)
         {
-            return info.add_literal(literal{v.get_shape().type()});
+            return info.add_literal(literal{migraphx::shape{v.get_shape().type(), {1}, {0}}, {0}});
```
CharlieL7, line 46: Is this is using the deduced template literal constructor? I think it would be using the literal constructor that gets a shape type.

src/onnx/parse_constant_of_shape.cpp
```diff
@@ -68,7 +68,9 @@ struct parse_constant_of_shape : op_parser<parse_constant_of_shape>
             // empty input tensor, output is a scalar
             if(args[0]->get_shape().elements() == 0)
             {
-                s = migraphx::shape{type, {1}, {0}};
+                std::vector<size_t> lens(1, 1);
+                std::vector<size_t> strides(1, 0);
+                s = migraphx::shape(type, lens, strides);
             }
```
CharlieL7, line 74: Why change to making a rank 2 shape?

src/onnx/parse_constant_of_shape.cpp
```diff
@@ -84,8 +86,16 @@ struct parse_constant_of_shape : op_parser<parse_constant_of_shape>
...
+                if(s.elements() > 0)
+                {
+                    std::vector<val_type> out_vec(s.elements(), val.front());
+                    l_out = literal(s, out_vec);
+                }
+                else
+                {
+                    std::vector<val_type> out_vec{val.front()};
+                    l_out = literal(s, out_vec);
+                }
```
CharlieL7, line 98: Why do we want to have a buffer with one value when there are no elements?

test/onnx/onnx_test.cpp
```diff
@@ -761,7 +761,9 @@ TEST_CASE(constant_empty_scalar_int64_test)
 {
     migraphx::program p;
     auto* mm = p.get_main_module();
-    mm->add_literal(migraphx::literal{migraphx::shape::int64_type});
+    migraphx::shape empty_shape{migraphx::shape::int64_type, {1}, {0}};
```
CharlieL7, line 764: This doesn't look like an empty shape to me. It looks like a scalar, as it has a length of 1.

TedThemistokleous: Correct the intention was to make empty shapes -> scalar of length 1 with value 0

src/onnx/parse_constant_of_shape.cpp
TedThemistokleous: To hold a zero value, so that every empty constant resolves to a literal with value {0}.

test/onnx/onnx_test.cpp
CharlieL7: I think you want an empty shape to have a length of 0 if you're avoiding having `shape.lens() == {}`? With this how would I represent a scalar zero?

TedThemistokleous:
Closing this out. Came to a much better solution after discussing this issue further and some of the avenues I took with @CharlieL7 . Will upload a change to the target branch directly without modifying shape.cpp and messing with "What is a valid input for empty literals" That opens up a can of worms and breaks existing functionality of other ops.

PR #1458, opened 2022-11-16
Fix logical_xor type checking

TedThemistokleous:
Fix to stop types failing for logical_xor during our fusions. Quick and dirty. Will post more changes here once I sort this out to get resnext50 working.

### DIFF
--- a/src/fuse_pointwise.cpp
+++ b/src/fuse_pointwise.cpp
@@ -45,7 +45,14 @@ static literal get_scalar(instruction_ref ins)
         return {};
     auto e = ins->eval();
     literal r{};
-    e.visit_at([&](auto x) { r = literal{x}; });
+    if(e.get_shape().type() == shape::bool_type)
+    {
+        r = literal{e.at<bool>()};
+    }
+    else
+    {
+        e.visit_at([&](auto x) { r = literal{x}; });
+    }
     return r;
 }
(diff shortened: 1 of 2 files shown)


src/fuse_pointwise.cpp
```diff
@@ -45,7 +45,14 @@ static literal get_scalar(instruction_ref ins)
         return {};
     auto e = ins->eval();
     literal r{};
-    e.visit_at([&](auto x) { r = literal{x}; });
+    if(e.get_shape().type() == shape::bool_type)
+    {
+        r = literal{e.at<bool>()};
```
umangyadav, line 50: Just add comment here that during `visit` bool type are treated as `int8` and that's why this requires `if` condition to handle this case separately.

umangyadav (approved):

src/fuse_pointwise.cpp
TedThemistokleous: will do. I'll rebase this off develop too.

TedThemistokleous: done

TedThemistokleous:
> Yea this looks good to merge into develop if you can address the comments from @umangyadav.

@pfultz2 done

pfultz2 (approved):

PR #1459, opened 2022-11-16
Dynamic ONNX Gemm

CharlieL7:
* Depends on #1457
* Extends ONNX `Gemm` parser to handle dynamic input shapes
* Limits ONNX `Gemm` parsing to 2D input tensors for `A` and `B` inputs
  * As per the ONNX specifications 
  * No idea why it was testing 4D inputs before, commit history is from >2 years ago
* Changed `Gemm` ONNX tests to 2D input versions
* Add `onnx_verify` tests for `Gemm` 
  * Parsing ONNX `Gemm` links to more than one operator, checking that it produces the correct result


src/onnx/parse_gemm.cpp
```diff
@@ -53,61 +62,70 @@ struct parse_gemm : op_parser<parse_gemm>
...
+            trans_a = parser.parse_value(info.attributes.at("transA")).at<bool>();
         }
         if(contains(info.attributes, "transB"))
         {
-            transb = parser.parse_value(info.attributes.at("transB")).at<bool>();
+            trans_b = parser.parse_value(info.attributes.at("transB")).at<bool>();
         }
 
-        std::vector<int64_t> perm(args[0]->get_shape().lens().size());
+        std::vector<int64_t> perm(2);
```
pfultz2, line 72: I think you could probably just write it as `std::vector<int64_t> perm = {1, 0}` since its always size 2.

src/onnx/parse_gemm.cpp
```diff
@@ -53,61 +62,69 @@ struct parse_gemm : op_parser<parse_gemm>
...
         }
         if(contains(info.attributes, "transB"))
         {
-            transb = parser.parse_value(info.attributes.at("transB")).at<bool>();
+            trans_b = parser.parse_value(info.attributes.at("transB")).at<bool>();
         }
 
-        std::vector<int64_t> perm(args[0]->get_shape().lens().size());
-        std::iota(perm.begin(), perm.end(), int64_t{0});
+        std::vector<int64_t> perm = {0, 1};
```
umangyadav, line 72: You can remove `std::swap` below and just initialize perm to {1, 0}.

src/onnx/parse_gemm.cpp
```diff
@@ -53,61 +62,69 @@ struct parse_gemm : op_parser<parse_gemm>
...
+                    ? info.add_instruction(make_op("transpose", {{"permutation", perm}}), args[1])
+                    : args[1];
 
-        auto ret = info.add_instruction(make_op("dot"), l1, l2);
+        auto ret = info.add_instruction(make_op("dot"), a_arg, b_arg);
 
         if(args.size() == 3)
         {
-            if(not float_equal(beta, 0.0f) && args[2]->get_shape().elements() > 0)
+            // TODO: support dynamic C input
```
umangyadav, line 101: what's preventing us to consider this case ? AFAIK, we support multibroadcast and mul for the dynamic shapes.

CharlieL7: Should be doable, but not used right now so I didn't implement it yet.

umangyadav (approved):

pfultz2 (approved):

PR #1461, opened 2022-11-18
Add disable tests for one case that breaks in onnx v1.12.0

TedThemistokleous:
currently breaking for CI but we're not relying on CPU right now. Seems to be on the onnx side of this error.


pfultz2:
But we dont use onnx 1.12 in CI?

causten (changes requested):
Do not include the 22.04 work in to this PR.  It is good to include 3.10 and the 1.12.0 disable tests.   The CI runs of debug builds fail on 22.04 due to a compiler issue and that is taking longer then expected to get resolved.

TedThemistokleous:
@causten all good. Just reverted that change then

causten (approved):

PR #1462, opened 2022-11-18
Temporarily disable Find 2.0 API for now

umangyadav:
Disabling it untill `int8` fix is in mainline from MIOpen and also so that QA tests could run `migraphx-driver` and unittests from MIGraphX.


causten (approved):
I understand this disables Find 2.0 from the develop branch for now.

umangyadav:
> I understand this disables Find 2.0 from the develop branch for now.

Yes, I can retarget to master  if you want to keep it alive on develop.

TedThemistokleous (approved):

pfultz2:
There should be a cmake variable to disable or enable it, so we dont have to edit CMakeLists.txt to enable or disable.

PR #1465, opened 2022-11-21
Add tupleVisitor for from_gpu

TedThemistokleous:
Need this for when we debug and use MIGRAPHX_TRACE_EVAL() to show tuples

Without this we break when reading our buffer due to the use of visit()

Credit for Paul Fultz helping on this.

This came up as part of #1283 debugging.

### DIFF
--- a/src/targets/gpu/hip.cpp
+++ b/src/targets/gpu/hip.cpp
@@ -196,12 +196,21 @@ argument to_gpu(const argument& arg, bool host)
 argument from_gpu(const argument& arg)
 {
     argument result;
-    arg.visit([&](auto x) {
-        using type = typename decltype(x)::value_type;
-        auto v     = read_from_gpu<type>(arg.data(), x.get_shape().bytes() / sizeof(type));
-        // cppcheck-suppress returnDanglingLifetime
-        result = {x.get_shape(), [v]() mutable { return v.data(); }};
-    });
+    arg.visit(
+        [&](auto x) {
+            using type = typename decltype(x)::value_type;
+            auto v     = read_from_gpu<type>(arg.data(), x.get_shape().bytes() / sizeof(type));
+            // cppcheck-suppress returnDanglingLifetime
+            result = {x.get_shape(), [v]() mutable { return v.data(); }};
+        },
+        [&](const auto& xs) {
+            std::vector<argument> args;
+            std::transform(xs.begin(), xs.end(), std::back_inserter(args), [&](auto x) {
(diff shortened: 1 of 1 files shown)


umangyadav (approved):

src/targets/gpu/hip.cpp
```diff
@@ -196,12 +196,21 @@ argument to_gpu(const argument& arg, bool host)
...
-    });
+    arg.visit(
+        [&](auto x) {
+            using type = typename decltype(x)::value_type;
+            auto v     = read_from_gpu<type>(arg.data(), x.get_shape().bytes() / sizeof(type));
+            // cppcheck-suppress returnDanglingLifetime
+            result = {x.get_shape(), [v]() mutable { return v.data(); }};
+        },
+        [&](const auto& xs) {
+            std::vector<argument> args;
```
umangyadav, line 207: Is it possible to add a unit test ?

src/targets/gpu/hip.cpp
```diff
@@ -196,12 +196,21 @@ argument to_gpu(const argument& arg, bool host)
...
 {
     argument result;
-    arg.visit([&](auto x) {
-        using type = typename decltype(x)::value_type;
-        auto v     = read_from_gpu<type>(arg.data(), x.get_shape().bytes() / sizeof(type));
-        // cppcheck-suppress returnDanglingLifetime
-        result = {x.get_shape(), [v]() mutable { return v.data(); }};
-    });
+    arg.visit(
+        [&](auto x) {
```
umangyadav, line 200: Can this be made `const auto& x` ?

umangyadav:
> Without this we break when reading our buffer due to the use of visit()
Is there any more places where we don't pass `TupleVisitor` for `visit()` ?

src/targets/gpu/hip.cpp
pfultz2: This is a tensor_view so it should be cheap to copy.

TedThemistokleous:
> > Without this we break when reading our buffer due to the use of visit()
> 
> Are there any more places where we don't pass `TupleVisitor` for `visit()` ?

Id need to check.

src/targets/gpu/hip.cpp
TedThemistokleous: done

umangyadav:
Added test for hip to/from_gpu for tuple case.

umangyadav (approved):

pfultz2 (approved):

PR #1477, opened 2022-11-28
remove extra adjust allocation pass

kahmed10:
Merging #1391 caused an extra adjust allocation pass for GPU targets. This removes that merge error.


pfultz2 (approved):

kahmed10:
> Did this PR cause a breakage (NAN) with slim-inceptionv4_1 ?

I don't see a NAN in the report above. We moved the adjust allocation pass in #1418, but the previous position was also included with a merge.

causten:
> 

oh nice, the comment block was updated.  At the time I asked the question there was a NAN.

PR #1480, opened 2022-11-30
Dynamic gathernd 

bpickrel:
Dynamic shape support for gathernd op.

No changes to compute logic (other than replacing variable types) were required.  No changes to parsing were required, as this op. uses generic parsing.

Two additional validation conditions were added to compute_shape--one is a preexisting algebraic relationship that's implied in the Onnx documentation; the other is a requirement that the last dimension, if dynamic, must be fixed (i.e. uses the dynamic type but can't really change).  This is because this value is used in computing the output shape's rank.  MIGraphX supports dynamic dimensions, but not dynamic NUMBER of dimensions.


src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +47,137 @@ struct gathernd
...
+            }
+            auto r               = data_shape.ndim();
+            auto q               = i_shape.ndim();
+            int output_dims_size = int(q) + r - k - batch_dims - 1;
+            if(k > r - batch_dims)
+            {
+                MIGRAPHX_THROW("GATHERND: Indices of length " + std::to_string(k) +
+                               " cannot be used to access data of rank " +
+                               std::to_string(r - batch_dims));
+            }
```
CharlieL7, line 115: The errors have repeated code between the dynamic and static shape versions. Should be more unified.

test/op_shape_test.cpp
```diff
@@ -2283,6 +2283,198 @@ TEST_CASE(test_scalar_nelemnts)
...
+        auto dtype = migraphx::shape::float_type;
+        auto itype = migraphx::shape::int64_type;
+        migraphx::shape is{itype, {2, 4}};
+        std::vector<migraphx::shape::dynamic_dimension> b{{8, 8, 0}};
+        migraphx::shape ds{dtype, b};
+
+        int batch_dims(1);
+        throws_shape(migraphx::make_op("gathernd", {{"batch_dims", batch_dims}}), ds, is);
+    }
+}
```
CharlieL7, line 2358: Remove inner braces

test/op_shape_test.cpp
```diff
@@ -2283,6 +2283,198 @@ TEST_CASE(test_scalar_nelemnts)
...
+        auto itype = migraphx::shape::int64_type;
+        migraphx::shape is{itype, {2, 2}};
+        std::vector<migraphx::shape::dynamic_dimension> b{{2, 2, 0}, {2, 2, 0}};
+        migraphx::shape ds{dtype, b};
+
+        migraphx::shape::dynamic_dimension ddout{2, 2, 0};
+        migraphx::shape s0{dtype, {ddout}};
+        expect_shape(s0, migraphx::make_op("gathernd"), ds, is);
+    }
+}
```
CharlieL7, line 2403: This is Example 1 from the ONNX spec, not Example 4. The `dynamic_dimension` is also fixed. We wouldn't see this case because the `dynamic_dimension` would be simplified during parsing to a static shape.

test/op_shape_test.cpp
```diff
@@ -2283,6 +2283,198 @@ TEST_CASE(test_scalar_nelemnts)
...
+        migraphx::shape is{itype, {2, 1}};
+        std::vector<migraphx::shape::dynamic_dimension> b{{2, 2, 0}, {2, 2, 0}, {2, 2, 0}};
+        migraphx::shape ds{dtype, b};
+
+        std::vector<migraphx::shape::dynamic_dimension> ddout{{2, 2, 0}, {2, 2, 0}};
+        int batch_dims(1);
+        migraphx::shape s0{dtype, {ddout}};
+        expect_shape(s0, migraphx::make_op("gathernd", {{"batch_dims", batch_dims}}), ds, is);
+    }
+}
```
CharlieL7, line 2421: Same issues as above

test/op_shape_test.cpp
```diff
@@ -2283,6 +2283,198 @@ TEST_CASE(test_scalar_nelemnts)
...
+        std::vector<migraphx::shape::dynamic_dimension> b{{2, 2, 0}, {1, 1, 0}};
+        migraphx::shape is{itype, b};
+        migraphx::shape ds{dtype, {2, 2, 2}};
+
+        std::vector<migraphx::shape::dynamic_dimension> ddout{{2, 2, 0}, {2, 2, 0}};
+        int batch_dims(1);
+        migraphx::shape s0{dtype, {ddout}};
+        expect_shape(s0, migraphx::make_op("gathernd", {{"batch_dims", batch_dims}}), ds, is);
+    }
+}
```
CharlieL7, line 2439: Same issues as above

test/op_shape_test.cpp
```diff
@@ -2283,6 +2283,198 @@ TEST_CASE(test_scalar_nelemnts)
...
+
+TEST_CASE(test_gathernd_dynamic5)
+{
+    {
+        // See Example 5 at https://github.com/onnx/onnx/blob/main/docs/Operators.md#GatherND
+        // index dynamic shape, data static
+        auto dtype = migraphx::shape::float_type;
+        auto itype = migraphx::shape::int64_type;
+        std::vector<migraphx::shape::dynamic_dimension> b{{2, 2, 0}, {1, 1, 0}};
+        migraphx::shape is{itype, b};
```
CharlieL7, line 2431: Need to test when indices other than the last are not fixed

test/op_shape_test.cpp
```diff
@@ -2283,6 +2283,198 @@ TEST_CASE(test_scalar_nelemnts)
...
+TEST_CASE(test_gathernd_dynamic4)
+{
+    {
+        // See Example 5 at https://github.com/onnx/onnx/blob/main/docs/Operators.md#GatherND
+        // index static shape, data dynamic
+        auto dtype = migraphx::shape::float_type;
+        auto itype = migraphx::shape::int64_type;
+        migraphx::shape is{itype, {2, 1}};
+        std::vector<migraphx::shape::dynamic_dimension> b{{2, 2, 0}, {2, 2, 0}, {2, 2, 0}};
+        migraphx::shape ds{dtype, b};
```
CharlieL7, line 2414: Need to test when the data dimensions are not fixed.

test/op_shape_test.cpp
```diff
@@ -2283,6 +2283,198 @@ TEST_CASE(test_scalar_nelemnts)
...
+        migraphx::shape is{itype, {2, 2, 1}};
+        std::vector<migraphx::shape::dynamic_dimension> b{{2, 2, 2}, {3, 3, 0}, {1, 1, 0}};
+        migraphx::shape ds{dtype, b};
+
+        std::vector<migraphx::shape::dynamic_dimension> ddout{{2, 2, 0}, {2, 2, 0}, {1, 1, 0}};
+        int batch_dims(1);
+        migraphx::shape s0{dtype, {ddout}};
+        expect_shape(s0, migraphx::make_op("gathernd", {{"batch_dims", batch_dims}}), ds, is);
+    }
+}
```
CharlieL7, line 2476: Same problems

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,165 @@ TEST_CASE(gathernd_test)
...
+        params["X"] = migraphx::argument(input_fixed_shape0, data_vec.data());
+        params["I"] = migraphx::argument(input_fixed_shape1, indices_vec.data());
+
+        auto result = p.eval(params).back();
+        std::vector<float> res_data{};
+        std::vector<float> gold{3, 4, 5, 0, 1, 2, 0, 1, 2, 3, 4, 5};
+        result.visit([&](auto output) { res_data.assign(output.begin(), output.end()); });
+
+        EXPECT(migraphx::verify_range(res_data, gold));
+    }
```
CharlieL7, line 2712: This case shouldn't happen because it can be easily simplified to a static shape.

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,165 @@ TEST_CASE(gathernd_test)
     }
 }
 
+TEST_CASE(gathernd_dynamic)
```
CharlieL7, line 2677: Split these tests into different cases

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,165 @@ TEST_CASE(gathernd_test)
...
+    { // dynamic data, dims not fixed
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+
+        migraphx::shape ds{migraphx::shape::float_type, {{2, 5, 2}, {1, 5, 0}, {1, 5, 0}}};
+        migraphx::shape is{migraphx::shape::int64_type, {2, 2, 1}};
+
+        std::vector<float> data_vec(2 * 3 * 1);
+        std::iota(data_vec.begin(), data_vec.end(), 0);
+        std::vector<int64_t> indices_vec{1, 0, 0, 1};
```
CharlieL7, line 2723: Move these lines to after the compile call, makes more sense that way.

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,165 @@ TEST_CASE(gathernd_test)
...
+    { // dynamic both index and data
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+
+        migraphx::shape ds{migraphx::shape::float_type, {{2, 5, 2}, {1, 5, 0}, {1, 5, 0}}};
+        migraphx::shape is{migraphx::shape::int64_type, {{2, 5, 3}, {2, 2, 3}, {1, 1}}};
+
+        std::vector<float> data_vec(2 * 3 * 1);
+        std::iota(data_vec.begin(), data_vec.end(), 0);
+        std::vector<int64_t> indices_vec{1, 0, 0, 1};
```
CharlieL7, line 2758: Same thing

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,165 @@ TEST_CASE(gathernd_test)
...
+
+        EXPECT(migraphx::verify_range(res_data, gold));
+    }
+
+    { // dynamic both index and data
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+
+        migraphx::shape ds{migraphx::shape::float_type, {{2, 5, 2}, {1, 5, 0}, {1, 5, 0}}};
+        migraphx::shape is{migraphx::shape::int64_type, {{2, 5, 3}, {2, 2, 3}, {1, 1}}};
```
CharlieL7, line 2754: {2, 2, 3} is not a valid `dynamic_dimension`

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,165 @@ TEST_CASE(gathernd_test)
...
+
+        EXPECT(migraphx::verify_range(res_data, gold));
+    }
+
+    { // dynamic index, static data
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+
+        migraphx::shape ds{migraphx::shape::float_type, {2, 3, 1}};
+        migraphx::shape is{migraphx::shape::int64_type, {{2, 5, 3}, {2, 2, 3}, {1, 1}}};
```
CharlieL7, line 2789: {2, 2, 3} problem

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,165 @@ TEST_CASE(gathernd_test)
...
+
+        migraphx::shape ds{migraphx::shape::float_type, {{2, 2}, {3, 3}, {1, 1}, {3, 3}}};
+        migraphx::shape is{migraphx::shape::int64_type, {2, 3, 3}};
+
+        const int batch_dims = 2;
+        auto xdata           = mm->add_parameter("X", ds);
+        auto xindex          = mm->add_parameter("I", is);
+
+        auto gathernd_op = migraphx::make_op("gathernd", {{"batch_dims", batch_dims}});
+        EXPECT(test::throws([&] { mm->add_instruction(gathernd_op, xdata, xindex); }));
```
CharlieL7, line 2832: The same test is already in op_shape_test.cpp

test/onnx/onnx_test.cpp
```diff
@@ -2099,6 +2099,24 @@ TEST_CASE(gathernd_test)
...
+    auto l0  = mm->add_parameter("data",
+                                migraphx::shape{migraphx::shape::float_type, {{2, 2, 2}, {2, 2}}});
+    auto l1  = mm->add_parameter("indices",
+                                migraphx::shape{migraphx::shape::int64_type, {{2, 2}, {2, 2}}});
+    auto r   = mm->add_instruction(migraphx::make_op("gathernd"), l0, l1);
+    mm->add_return({r});
+
+    migraphx::onnx_options options;
+    options.map_dyn_input_dims["data"]    = {{2, 2, 2}, {2, 2}};
+    options.map_dyn_input_dims["indices"] = {{2, 2}, {2, 2}};
```
CharlieL7, line 2115: This test is not dynamic, it has all fixed `dynamic_dimensions`. I think it works because `map_dyn_input_dims` overrides to a `dynamic_dimension` even if it's fixed.

CharlieL7 (changes requested):
Several problems with checking that the `compute_shape` works correctly. The ref_ops_test will work even if the compute_shape is incorrect because the shape is replaced at model eval time. Possibly incorrect dynamic output shapes will break the rest of the model.

test/onnx/onnx_test.cpp
bpickrel: Isn't the default dimension typing a functionality that also needs to be tested?

As for fixed vs. non-fixed dimensions, they all use the dynamic_dimension type.  This test does force the code to get the mins, the maxes, and the opts right.  Is there something else about non-fixed dimensions that needs to be checked at this point?  If so, maybe we should clarify what we want to test for.

Also, should we include multiple tests, including fixed and non-fixed dynamic_dimensions?  Does that add any value, or is it overkill?

CharlieL7: I don't understand the first question.

The tests for the output_shape should be in op_shape_test.cpp. If we only test when the dynamic_dimension is fixed we won't know if it works when it's not fixed.

Testing all fixed dynamic_dimensions doens't make much sense because it's equivalent to a static shape at that point.

test/ref_ops_test.cpp
bpickrel: It isn't supposed to be, but the existing code allows it.  I've included dimension specifications in this test that initialize in different ways.  Why not include tests of what should be invalid input?

CharlieL7: This won't error out though. If you wanted to test this or make a new assert to prevent this I would make a new PR doing this for the `shape` class and related tests rather than in `gathernd`.

test/op_shape_test.cpp
bpickrel: I'll update the comment.  

Even if we wouldn't see the case, don't we still want to test the internal behavior?  Given that it's a unit and not end-to-end test.  Do we have, or need, tests of the conversion to static?

bpickrel: Also, if I change all the fixed dynamic shapes I've got in my tests to be non-fixed, does it matter for purposes of testing whether I specify an opt value; set it to 0; or leave it blank?

src/include/migraphx/op/gathernd.hpp
bpickrel: The code is the same but the variables are in a different scope.  Trying to shoehorn them together would cause more problems than it solves.

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +47,137 @@ struct gathernd
```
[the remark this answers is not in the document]
bpickrel: will change.

bpickrel:
> 

Please clarify

test/ref_ops_test.cpp
bpickrel: This change made in several places.

test/ref_ops_test.cpp
bpickrel: done

test/op_shape_test.cpp
bpickrel: done in several places.

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +47,137 @@ struct gathernd
...
-        auto k = inputs.back().lens().back();
-        if(k > r - batch_dims)
+        check_shapes{inputs, *this, true}.has(2);
+        if(migraphx::none_of(inputs, [](auto v) { return v.dynamic(); }))
         {
-            MIGRAPHX_THROW("GATHERND: Indices of length " + std::to_string(k) +
-                           " cannot be used to access data of rank " +
-                           std::to_string(r - batch_dims));
+            auto r = inputs.front().lens().size();
+            auto q = inputs.back().lens().size();
```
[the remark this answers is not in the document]
bpickrel: Done, for extremely marginal payoff in readability

test/op_shape_test.cpp
CharlieL7: You specify the opt value so that you're testing whatever code you added for calculating it. Ex: if you coded it that the opt for the output shape should always be 0, test that with different inputs.

test/ref_ops_test.cpp
bpickrel: removed

test/onnx/onnx_test.cpp
bpickrel: Changed several dimensions in several tests to be non-fixed.  

We agreed that adding the functionality  of checking (and testing) for invalid dynamic dimension values may be a good idea but would lie outside the scope of this operator and its tests.

test/ref_ops_test.cpp
bpickrel: changed

test/ref_ops_test.cpp
bpickrel: changed, we agreed this op. is not the place to check for invalid dynamic dimension inputs

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,151 @@ TEST_CASE(gathernd_test)
...
+    params["X"] = migraphx::argument(input_fixed_shape0, data_vec.data());
+    params["I"] = migraphx::argument(input_fixed_shape1, indices_vec.data());
+
+    auto result = p.eval(params).back();
+    std::vector<float> res_data{};
+    std::vector<float> gold{3, 4, 5, 0, 1, 2, 0, 1, 2, 3, 4, 5};
+    result.visit([&](auto output) { res_data.assign(output.begin(), output.end()); });
+
+    EXPECT(migraphx::verify_range(res_data, gold));
+}
```
CharlieL7, line 2712: This test case is not particularly useful since the shape is equivalent to a static shape.

CharlieL7 (changes requested):

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +48,136 @@ struct gathernd
...
+            // If one or both inputs are dynamic shapes, the output is dynamic
+            auto i_shape    = inputs.back();
+            auto data_shape = inputs.front();
+            size_t k;
+            if(i_shape.dynamic())
+            {
+                // the rank of the output is a function of k, so it must be fixed.
+                if(not i_shape.dyn_dims().back().is_fixed())
+                {
+                    MIGRAPHX_THROW(
```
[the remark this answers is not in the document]
bpickrel: added case

test/ref_ops_test.cpp
bpickrel: removed

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +48,136 @@ struct gathernd
...
+                               " cannot be used to access data of rank " +
+                               std::to_string(r - batch_dims));
+            }
+            if(batch_dims >= q or batch_dims >= r)
+            {
+                MIGRAPHX_THROW("GATHERND: rank of an input cannot be less than batch_dims=" +
+                               std::to_string(batch_dims));
+            }
+            auto indices_lens_iter = inputs.back().lens().begin();
+            int output_lens_size   = int(q) + r - k - batch_dims - 1;
```
umangyadav, line 69:
```suggestion
            int output_lens_size   = q + r - k - batch_dims - 1;
```

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +48,136 @@ struct gathernd
...
+            if(output_lens_size <= 0)
+            {
+                MIGRAPHX_THROW("GATHERND: Indices too large for static data input: k=" +
+                               std::to_string(k));
+            }
```
umangyadav, line 74: Don't really need this since we already assert that `k > r -b` is NOT true. 
Therefore `1 <= k <= r-b` ---> `1 <= (r -b  -k)` and given `q >=1` ----> `output_lens_size >=1`.

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +48,136 @@ struct gathernd
...
+                }
+                k = i_shape.dyn_dims().back().min;
+            }
+            else
+            {
+                k = i_shape.lens().back();
+            }
+            auto r          = data_shape.ndim();
+            auto q          = i_shape.ndim();
+            int output_ndim = int(q) + r - k - batch_dims - 1;
```
umangyadav, line 109:
```suggestion
            int output_ndim = q + r - k - batch_dims - 1;
```

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +48,136 @@ struct gathernd
...
+            if(batch_dims >= q or batch_dims >= r)
+            {
+                MIGRAPHX_THROW("GATHERND: rank of an input cannot be less than batch_dims=" +
+                               std::to_string(batch_dims));
+            }
+            if(output_ndim <= 0)
+            {
+                MIGRAPHX_THROW("GATHERND: Indices too large for data input: k=" +
+                               std::to_string(k));
+            }
```
umangyadav, line 125: Same as  previous comment, this can't happen if we already check for `k >  r - b`

src/include/migraphx/op/gathernd.hpp
umangyadav: Yeah, we should combine error messages and some part of the code. To start with , 
we don't need to redefine  `q`, `r` and `output_dims` and `k`'s definition can be If-Then-Else to cover dynamic case.
Thereafter we can merge all the `MIGRAPHX_THROWS`

umangyadav (changes requested):

test/op_shape_test.cpp
```diff
@@ -2325,6 +2325,218 @@ TEST_CASE(test_scalar_nelemnts)
...
+
+    {
+        // int(q) + r - k - batch_dims - 1 = 0
+        auto dtype = migraphx::shape::float_type;
+        auto itype = migraphx::shape::int64_type;
+        migraphx::shape is{itype, {2, 4}};
+        std::vector<size_t> b{2};
+        migraphx::shape ds{dtype, b};
+
+        throws_shape(migraphx::make_op("gathernd"), is, ds);
```
umangyadav, line 2371: Shouldn't rank `0` output be valid? that means it is a scalar value. I don't find any documentation on ONNX for that. 
++++ @CharlieL7

test/ref_ops_test.cpp
```diff
@@ -2674,6 +2674,151 @@ TEST_CASE(gathernd_test)
...
+
+TEST_CASE(gathernd_dynamic3)
+{
+    // dynamic index, static data
```
umangyadav, line 2788: You can edit one of these tests to include `batch_dims` for more inclusive tests.

test/op_shape_test.cpp
CharlieL7: To get a scalar, indices tensor must be rank 1 `q = 1`. Let's say `r = 2`, `batch_dims = 0`. Therefore `q + r - k - batch_dims - 1 = 0` goes to `1 + 2 - k - 0 - 1 = 0`. So `k = 2`. That looks like it could work. Would need to change how the code works to handle this case.

src/include/migraphx/op/gathernd.hpp
bpickrel: The signed int is used to check for incompatible input values which would create a negative result.  The suggested change would yield uint64 with an underflowed value, not a negative.

CharlieL7: Looks like this might be implementation specific for `std::size_t`. `int + uint` will convert the int to an unsigned. Weird stuff can happen with subtracting `std::size_t` values. Option to convert everything to `std::ptrdiff_t` before or checking the subtractions before.

Also I think this situation is still called overflow, since underflow is for floating point values

bpickrel: I appreciate the thought given, but let's not overthink this.  This is a situation that only comes up with erroneous inputs anyway, and you just yesterday gave me a lecture on not adding excessive error checking to model compilation.  This math works with the operating system we use; with the signed int it leads to an exception message and without it we'd get a segmentation fault.  That's enough.  We have no requirement to future-proof against every conceivable integer over- or underflow.

test/op_shape_test.cpp
CharlieL7: Just add a throw for this case. I do not think we will see it often.

bpickrel: I added this feature and it tests out OK.  A calculation that leads to rank 0 returns, instead, a 1 dimensional shape with a size of 1.

test/ref_ops_test.cpp
bpickrel: added

src/include/migraphx/op/gathernd.hpp
```diff
@@ -47,33 +48,103 @@ struct gathernd
...
+                           std::to_string(k));
+        }
+
+        if(migraphx::none_of(inputs, [](auto v) { return v.dynamic(); }))
+        {
+            auto indices_lens_iter = i_shape.lens().begin();
+
+            // A rank 0 output is a scalar
+            if(output_ndim == 0)
+                return shape{data_shape.type(), {1}};
```
umangyadav, line 99: This can be moved out, Since it appears for both static and dynamic shapes.

umangyadav (approved):
LGTM, except just do one time check for scalar in `compute_shape()`.

src/include/migraphx/op/gathernd.hpp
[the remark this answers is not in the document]
bpickrel: No, in one case it returns a static shape and in the other, a dynamic shape.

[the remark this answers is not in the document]
umangyadav: Other case, it is dynamic but fixed shape.

CharlieL7 (approved):

PR #1482, opened 2022-12-01
Dynamic ref flatten

CharlieL7:
* Changes `flatten`'s `compute_shape()` to handle dynamic shapes
  * Calculates the flattened shape with the min, max, and opt


test/onnx/onnx_test.cpp
```diff
@@ -1914,6 +1914,23 @@ TEST_CASE(flatten_nonstd_test)
...
 }
 
+TEST_CASE(flatten_dyn_test)
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    auto l0  = mm->add_parameter(
+        "0",
+        migraphx::shape{migraphx::shape::float_type, {{1, 4, 0}, {3, 3, 0}, {4, 4, 0}, {5, 5, 0}}});
+    auto c0  = mm->add_instruction(migraphx::make_op("contiguous"), l0);
```
umangyadav, line 1924: Where does it add contiguous on parsing ?

src/include/migraphx/op/flatten.hpp
```diff
@@ -55,17 +55,47 @@ struct flatten
...
+        check_shapes{inputs, *this, true}.has(1);
+        auto s = inputs[0];
+        if(s.dynamic())
+        {
+            auto min_lens = s.min_lens();
+            auto max_lens = s.max_lens();
+            auto opt_lens = s.opt_lens();
+            // If any of the opt values is 0, output opt will be 0
+            shape::dynamic_dimension x = {
+                std::accumulate(
```
umangyadav, line 67: It is using `std::accumulate` in same fashion in multiple places, can we add lambda to shorten that ?

test/onnx/onnx_test.cpp
CharlieL7: `parse_generic_op.cpp` has `flatten` requiring contiguous

umangyadav (approved):

src/include/migraphx/op/flatten.hpp
```diff
@@ -55,17 +55,47 @@ struct flatten
...
+                                opt_lens.begin() + axis,
+                                std::size_t{1},
+                                std::multiplies<>{})};
+            shape::dynamic_dimension y = {
+                std::accumulate(
+                    min_lens.begin() + axis, min_lens.end(), std::size_t{1}, std::multiplies<>{}),
+                std::accumulate(
+                    max_lens.begin() + axis, max_lens.end(), std::size_t{1}, std::multiplies<>{}),
+                std::accumulate(
+                    opt_lens.begin() + axis, opt_lens.end(), std::size_t{1}, std::multiplies<>{}),
```
bpickrel, line 81: Does this formula for opt_lens actually make any sense?  If not, you could probably safely set them all to 0.  Not sure if that makes any functional difference.

CharlieL7: Yeah it makes sense for opt_lens. If any of the opt is 0, it will evaluate to zero. Otherwise it will evaluate the same as the others.

bpickrel (approved):
LGTM

PR #1487, opened 2022-12-08
Dynamic ref pad

CharlieL7:
* extends `pad` operator to handle dynamic input shapes
* only handles computing the shape for adding constant padding to a dynamic shape
  * adds the padding to the min, max, and opt values (unless opt is 0, where it keeps it 0)
  * does not handle `reflect` padding with dynamic shapes
* I don't think the `pad_mode` is used, may want to refactor later


src/include/migraphx/shape.hpp
```diff
@@ -107,6 +107,11 @@ struct shape
         friend bool operator==(const std::size_t& x, const dynamic_dimension& y);
         friend bool operator!=(const dynamic_dimension& x, const std::size_t& y);
         friend bool operator!=(const std::size_t& x, const dynamic_dimension& y);
+
+        // add and subtract fixed std::size_t dimension
+        friend dynamic_dimension operator+(const dynamic_dimension& x, const std::size_t& y);
+        friend dynamic_dimension operator+(const std::size_t& x, const dynamic_dimension& y);
+        friend dynamic_dimension operator-(const dynamic_dimension& x, const std::size_t& y);
```
TedThemistokleous, line 114: No subtract dynamic_dimension from size_t?  Are we assuming this result is just a size_t then and handled somewhere else?

CharlieL7: I don't use that case in the current code. I will add it and some asserts for the subtraction operators to catch if the subtraction goes below 0. (a dimension of 0 won't make sense for most cases either?)

CharlieL7: It's a somewhat weird case however.
Example:
10 - {2, 5, 3} = {5, 8, 7} where the `min` output is `10 - max` and `max` output is `10 - min`

CharlieL7: Holding off on adding that case. I don't see where we would use it.

TedThemistokleous: Thats fine. Just wanted to ask the question since you had the other cases with the addition case.

TedThemistokleous (approved):
Looks good.  Should be good to go in since you added asserts if we get into those cases i mentioned above.

src/include/migraphx/op/pad.hpp
```diff
@@ -59,18 +59,29 @@ struct pad
...
-        for(std::size_t i = 0; i < num_dims; i++)
+        check_shapes{inputs, *this, true}.has(1);
+        const auto& s0 = inputs.front();
+        if(s0.dynamic())
         {
-            rdims[i] += pads[i] + pads[i + num_dims];
+            auto out_dyn_dims = s0.dyn_dims();
+            for(std::size_t i = 0; i < s0.ndim(); ++i)
+            {
+                out_dyn_dims[i] = out_dyn_dims[i] + pads[i] + pads[i + s0.ndim()];
```
umangyadav, line 69: We can also add `+=` friend method.

umangyadav:
>I don't think the pad_mode is used, may want to refactor later 
Can you open an issue to track?

src/shape.cpp
```diff
@@ -529,6 +529,29 @@ bool operator==(const std::size_t& x, const shape::dynamic_dimension& y) { retur
...
+{
+    assert(x.min >= y);
+    assert(x.max >= y);
+    if(x.opt == 0)
+    {
+        return {x.min - y, x.max - y, 0};
+    }
+    else
+    {
+        assert(x.opt >= y);
```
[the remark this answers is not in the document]
CharlieL7: `opt` is implied to have to be within the range of `min` and `max`, but we currently don't prevent it being otherwise. The assert should catch if we're going to wraparound on `size_t`.

src/include/migraphx/op/pad.hpp
CharlieL7: `+=` function is mutating so I don't think it would benefit from being implemented outside of the class declaration. Adding the operators as class functions.

src/shape.cpp
```diff
@@ -504,6 +504,28 @@ bool shape::dynamic_dimension::is_fixed() const { return this->min == this->max;
...
+
+shape::dynamic_dimension& shape::dynamic_dimension::operator-=(const std::size_t& x)
+{
+    assert(this->min >= x);
+    assert(this->max >= x);
+    this->min -= x;
+    this->max -= x;
+    if(this->opt != 0)
+    {
+        assert(this->opt >= y);
```
umangyadav: Should this by `x` instead of `y` ?

CharlieL7: Should've just fixed that typo in commit `54ccaea`.

umangyadav (approved):

src/shape.cpp
```diff
@@ -504,6 +504,28 @@ bool shape::dynamic_dimension::is_fixed() const { return this->min == this->max;
 
 bool shape::dynamic_dimension::has_optimal() const { return opt != 0; }
 
+shape::dynamic_dimension& shape::dynamic_dimension::operator+=(const std::size_t& x)
+{
+    this->min += x;
+    this->max += x;
+    this->opt == 0 ? this->opt = 0 : this->opt += x;
```
pfultz2, line 511: Use an `if` statement: `if (this->opt != 0) this->opt += x`.

pfultz2 (approved):

PR #1489, opened 2022-12-12
Copy into registers first when doing reductions with layernorm and softmax

pfultz2:
This avoids double global loads.

Strided loops are unrolled which lets store results in array which compiler will use registers for since the index access is constant.


pfultz2:
So this is ready to go. The problem was the test was using a much larger reduction than I realized. This should work for all the sizes in stable diffusion. We can revisit this later if we run into even larger reductions.

pfultz2:
@kahmed10 @turneram are you able to review this?

src/targets/gpu/kernels/include/migraphx/kernels/debug.hpp
```diff
@@ -178,5 +178,9 @@ MIGRAPHX_HIP_NORETURN inline __host__ __device__ void assert_fail(const source_l
 #define MIGRAPHX_WARN(...)
 #endif
 
+#define MIGRAPHX_STATIC_ASSERT_FOR(...) \
+    static_assert(__VA_ARGS__);         \
+    if constexpr(__VA_ARGS__)
```
turneram, line 183: Two questions about this: 
What is the reason for using FOR in the name? 
Why isn't static_assert sufficient on its own?

pfultz2: >  What is the reason for using FOR in the name?

I am not sure if this is the best name, but it means we are applying a `static_assert` for the block.

>  Why isn't static_assert sufficient on its own?

When the `static_assert` is `false` the compiler will still continue compiling, but this macro will prevent the block from compiling if the assert is wrong. The problem is that when `max_stride_iterations` grows too large, the `static_assert` will hit, but it will contine to compile, which is very slow. So this will allow to skip the loop unroll if the `static_assert` is `false`.

kahmed10 (approved):

turneram (approved):

pfultz2:
So I updated this to handle large reductions so this should work fine with stable diffusion now. So I think it should be safe to merge in when CI passes.

PR #1491, opened 2022-12-12
Add a general optimize pass

pfultz2:
This will run several of the passes in a loop. I currently named the pass `optimize`, but perhaps there is a better name for this.


kahmed10:
maybe something like optimize_arithmetic, optimize_expressions, etc

turneram (approved):
I think `optimize` should be fine since the passes it runs each have more descriptive names.

kahmed10:
> I think `optimize` should be fine since the passes it runs each have more descriptive names.

Only reason I think we should be specific in the name is that it becomes more clear to an external user looking into our codebase. Technically we are optimizing the graph when applying other transformations under target.cpp too.

causten:
team ok with optimize_module

kahmed10 (approved):
This PR has changes to matchers, prefuse_ops, and fuse_ops which I'm assuming was done in order to simplify `target.cpp`. Would be good just to sanity check the regressions on performance just in case the changes result in some unoptimized portions.

pfultz2:
> This PR has changes to matchers, prefuse_ops, and fuse_ops which I'm assuming was done in order to simplify target.cpp.

Those changes was to enable layernorm fusion when there is no epsilon. On some bert models, the epsilon is 0 and so gets remove with the more aggressive optimizations.

PR #1495, opened 2022-12-16
Add serialization of tuples and optional types

pfultz2:


src/include/migraphx/serialize.hpp
```diff
@@ -187,28 +205,35 @@ void from_value_impl(rank<4>, const value& v, T& x)
...
 template <class T, MIGRAPHX_REQUIRES(std::is_arithmetic<T>{})>
-void from_value_impl(rank<5>, const value& v, T& x)
+void from_value_impl(rank<7>, const value& v, T& x)
 {
     x = v.to<T>();
 }
 
 template <class T, MIGRAPHX_REQUIRES(std::is_enum<T>{})>
-void from_value_impl(rank<6>, const value& v, T& x)
+void from_value_impl(rank<8>, const value& v, T& x)
```
umangyadav, line 222: I think this two can be  combined by logical or inside `MIGRAPHX_REQUIRES`.

test/serialize_test.cpp
```diff
@@ -125,6 +130,21 @@ TEST_CASE(serialize_empty_struct)
     EXPECT(v.at("a").to<int>() == 1);
 }
 
+TEST_CASE(serialize_empty_optional)
```
umangyadav, line 133: Add unittest for tuple as well

pfultz2: The tuple is already tested in the `serialize_reflectable_type` test since it was added to the `reflectable_type` class.

umangyadav (approved):

kahmed10 (approved):

Issue #1496, opened 2022-12-16
simplify algebra - conv_dot_horiz_fusion bug


shivadbhavsar:
Attached mxr contains a small set of instructions found in the unet model used within the stable diffusion pipeline. 
[conv_dot_fuse.zip](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/10249631/conv_dot_fuse.zip)

The bug can be reproduced by running driver verify on this mxr:
```
migraphx-driver verify conv_dot_fuse.mxr 
```
This verify test passes if we comment out the `find_conv_dot_horiz_fusion{}` matcher in simplify_algebra.cpp. The issue is likely within this function.

shivadbhavsar:
Upon further investigation, the bug is actually in the `find_split_reshape` pass. Attached is an even simpler program to reproduce the issue:
[slice_reshape.zip](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/10280443/slice_reshape.zip)

Program for reference:
```
module: "main"
x0 = @param:x0 -> float_type, {2, 4096, 640}, {2621440, 640, 1}
main:@1 = slice[axes={2},starts={0},ends={320}](x0) -> float_type, {2, 4096, 320}, {2621440, 640, 1}
main:@2 = contiguous(main:@1) -> float_type, {2, 4096, 320}, {1310720, 320, 1}
main:@3 = reshape[dims={2, 4096, 8, 40}](main:@2) -> float_type, {2, 4096, 8, 40}, {1310720, 320, 40, 1}
main:@4 = transpose[permutation={0, 2, 1, 3}](main:@3) -> float_type, {2, 8, 4096, 40}, {1310720, 40, 320, 1}
main:@5 = slice[axes={2},starts={320},ends={640}](x0) -> float_type, {2, 4096, 320}, {2621440, 640, 1}
main:@6 = contiguous(main:@5) -> float_type, {2, 4096, 320}, {1310720, 320, 1}
main:@7 = reshape[dims={2, 4096, 8, 40}](main:@6) -> float_type, {2, 4096, 8, 40}, {1310720, 320, 40, 1}
main:@8 = transpose[permutation={0, 2, 1, 3}](main:@7) -> float_type, {2, 8, 4096, 40}, {1310720, 40, 320, 1}
main:@9 = @return(main:@4,main:@8)
```

PR #1499, opened 2022-12-23
Transpose slice fix

shivadbhavsar:
This PR resolves the bug addressed in #1496. The issue points to several matchers in the simplify_algebra pass, however, upon further investigation, the simplify_algebra pass results in the correct program output. The issue was determined to be in the simplify_reshapes pass. 

List of updates:
1. find_split_reshape currently causes a segmentation fault if the reshaped split is the output of the program.
2. The transpose_slice simplification had a bug in its implementation causing the output of the compiled instructions to be a jumbled version of the true output.
3. Updated the existing simplify_reshape test to expect the correct modified program after the pass
4. Added a verify test for the same program found in simplify_reshapes_test.cpp (existing implementation caused this program to fail verify)

These changes allow us to run the full stable diffusion pipeline in fp32 mode correctly.


src/simplify_algebra.cpp
```diff
@@ -1066,10 +1066,16 @@ struct find_split_reshape
         }
 
         std::vector<instruction_ref> vec_rsp(split_outputs.size());
-        std::transform(split_outputs.begin(), split_outputs.end(), vec_rsp.begin(), [](auto i) {
+        std::list<instruction> undef_inss{instruction(make_op("undefined"), {}, {})};
```
pfultz2, line 1069: This is the wrong way to insert undefined operator(as the instruction doesnt belong to any module in the program. Why are you inserting undefined operators in the first place? Why dont we just exit when output size is not 1?

test/simplify_reshapes_test.cpp
```diff
@@ -1405,9 +1405,9 @@ TEST_CASE(transpose_slice_non_packed_axis)
     {
         auto x = m2.add_parameter("x", {migraphx::shape::float_type, {2, 384, 36, 64}});
         auto unsqueeze =
-            m2.add_instruction(migraphx::make_op("unsqueeze", {{"axes", {2}}, {"steps", {12}}}), x);
+            m2.add_instruction(migraphx::make_op("unsqueeze", {{"axes", {2}}, {"steps", {3}}}), x);
```
pfultz2, line 1408: Why is this changed?

src/simplify_algebra.cpp
shivadbhavsar: The check for output size happens in the lambda function for each item in split_outputs and so adding the undefined instruction is to ensure that the following `same_ops` check fails causing it to exit

pfultz2: I think it would be better to do `std::any_of` and check for the output size and then exit.

src/simplify_algebra.cpp
```diff
@@ -1065,6 +1065,13 @@ struct find_split_reshape
             return;
         }
 
+        if(std::any_of(split_outputs.begin(), split_outputs.end(), [](auto i) {
+               return i->outputs().front()->outputs().size() != 1;
```
kahmed10, line 1069: As it's written right now, it's kind of confusing. The `std::transform` in L1076 helps explain better what these variables are supposed to represent (and also some of the asserts shouldn't be needed anymore with the addition you made). `i->outputs().front()` is a contiguous op. You may want to explicitly make a variable `cont` like in L1078. Also, we have an assert for `i->outputs().size() == 1` in L1077. I'm not sure if you want to make the function return where the size is != 1 and remove that assert as well.

shivadbhavsar: Updated to remove both asserts and also add a comment to make the intent behind the check more clear

kahmed10 (approved):

pfultz2 (approved):

PR #1500, opened 2022-12-28
Dynamic ref reshape (one non-fixed case)

CharlieL7:
* Extends `reshape` to handle the case of a single non-fixed `dynamic_dimension`
* Example:
input dynamic shape = [{1, 4, 0}, {24, 24, 0}, {1, 1, 0}, {1, 1, 0}]
dims attribute = [-1, 8, 3, 1]
output dynamic shape = [{1, 4, 0}, {8, 8, 0}, {3, 3, 0}, {1, 1, 0}]
* Created to handle how the operator is used in a Resnet50 ONNX file


src/include/migraphx/op/reshape.hpp
```diff
@@ -46,14 +47,66 @@ struct reshape
...
+                    MIGRAPHX_THROW("Reshape: Only support one non-fixed dynamic_dimension");
+                }
+            }
+        }
+        if(num_dims_ele != num_dd_ele)
+        {
+            MIGRAPHX_THROW("Reshape: Number of fixed elements must match. Input: " +
+                           std::to_string(num_dd_ele) + " Output: " + std::to_string(num_dims_ele));
+        }
+        if(dims[not_fixed_index] != 0 and dims[not_fixed_index] != -1)
```
umangyadav, line 82:
```suggestion
        if(dims[not_fixed_index] != 0 or dims[not_fixed_index] != -1)
```

src/include/migraphx/op/reshape.hpp
```diff
@@ -46,14 +47,66 @@ struct reshape
...
+            else
+            {
+                std::size_t d = dims[i];
+                output_dyn_dims.push_back({d, d, 0});
+            }
+        }
+        return {s0.type(), output_dyn_dims};
+    }
+
+    template <class T>
```
umangyadav, line 104: Can you just use std::size_t, Instead of template?

CharlieL7: It looks like std::count() returns the iterator difference type, which is probably std::ptrdiff_t. I didn't want to deal with it so I made it a template. Probably could convert it to a size_t since it's going to be >= 0.

src/include/migraphx/op/reshape.hpp
```diff
@@ -46,14 +47,66 @@ struct reshape
...
+        for(std::size_t i = 0; i < dims.size(); ++i)
+        {
+            if(i == not_fixed_index)
+            {
+                output_dyn_dims.push_back(dyn_dims[not_fixed_index]);
+            }
+            else
+            {
+                std::size_t d = dims[i];
+                output_dyn_dims.push_back({d, d, 0});
```
pfultz2, line 98: Leave the `0` off of the constructor, it will already default to `0`.

src/include/migraphx/op/reshape.hpp
```diff
@@ -46,14 +47,66 @@ struct reshape
...
+            else
+            {
+                if(not_fixed_index == -1)
+                {
+                    not_fixed_index = i;
+                }
+                else
+                {
+                    MIGRAPHX_THROW("Reshape: Only support one non-fixed dynamic_dimension");
+                }
```
pfultz2, line 74: Use `count_if` to check for multiple non-fixed dynamic dimensions.

src/include/migraphx/op/reshape.hpp
```diff
@@ -46,14 +47,66 @@ struct reshape
...
+        {
+            if(i == not_fixed_index)
+            {
+                output_dyn_dims.push_back(dyn_dims[not_fixed_index]);
+            }
+            else
+            {
+                std::size_t d = dims[i];
+                output_dyn_dims.push_back({d, d, 0});
+            }
```
pfultz2, line 99: Use `std::transform` for this loop.

src/include/migraphx/op/reshape.hpp
CharlieL7: I changed it, but essentially added a loop over the `dyn_dims`. Doing everything in the `count_if` loop seemed hard to read.

src/include/migraphx/op/reshape.hpp
```diff
@@ -46,14 +47,58 @@ struct reshape
...
+            MIGRAPHX_THROW("Reshape: Non-fixed dynamic_dimension doesn't match with 0 or -1 "
+                           "output dimension");
+        }
+        // construct output dynamic shape from dims attribute
+        std::vector<shape::dynamic_dimension> output_dyn_dims(dims.size());
+        std::transform(dims.cbegin(), dims.cend(), output_dyn_dims.begin(), [](auto in_d) {
+            std::size_t d = in_d;
+            return shape::dynamic_dimension{d, d};
+        });
+        output_dyn_dims[not_fixed_index] = dyn_dims[not_fixed_index];
```
pfultz2, line 92: This should be done in the `std::transform` as well:

```cpp
std::transform(dims.cbegin(), dims.cend(), dyn_dims.begin(), std::back_inserter(output_dyn_dims), [](std::size_t dim, auto dyn_dim) {
    if (dyn_dim.is_fixed())
        return dyn_dim;
    return shape::dynamic_dimension{dim, dim};
});
```

CharlieL7: I've been wondering about `transform` and `back_inserter` since `transform` doesn't guarantee in-order traversal of the iterators. It works for our setup, but might not be up to the standard spec?

src/include/migraphx/op/reshape.hpp
umangyadav: It should get converted implictly.

src/include/migraphx/op/reshape.hpp
pfultz2: The `back_inserter` is an `OutputIterator` so it only allows traversing it one way. Although it could traverse the inputs in a different order, but this is not a problem because we dont have any side effects in the binary_op we pass.

src/include/migraphx/op/reshape.hpp
```diff
@@ -46,14 +47,58 @@ struct reshape
...
+        for(std::size_t i = 0; i < dyn_dims.size(); ++i)
+        {
+            if(dyn_dims[i].is_fixed())
+            {
+                num_dims_ele *= dims[i];
+                num_dd_ele *= dyn_dims[i].min;
+            }
+            else
+            {
+                not_fixed_index = i;
```
pfultz2, line 73: Remove this variable and just check `dim[i] != 0 and dim[i] != -1` here.

src/include/migraphx/op/reshape.hpp
umangyadav: Did you try removing template ?

umangyadav (approved):

pfultz2 (approved):

PR #1503, opened 2023-01-03
Dyn slice

bpickrel:
Add dynamic shape support to slice operator.

First draft of this feature doesn't support ops slicing non-fixed, dynamic axes.  Resulting shape in such cases is not guaranteed.*  Also, onnx parsing doesn't support any arguments other than "axes".

*We haven't specified the behavior yet if you try to slice a variable dimension.  What if part of the variable range is smaller than the requested slice?  Presumably this should be legal, since the Onnx spec allows a no-op "slicing" of dims that are smaller than the given slice size.  Perhaps we could allow a dynamic max. andmin. that are smaller than the slice size.


bpickrel:
A more detailed comment about Onnx parsing:  An Onnx Slice instruction always translates to an Migraphx ref. Slice instruction, but in addition a Step instruction will be added if (non-default) steps attribute is given, and a Reverse instruction is added if any negative steps are given.  The step and reverse ops don't currently support dynamic shapes, so these will cause an exception that might be hard for the user to trace to the Slice instruction.  This implementation does work with the Resnet model Charlie has been testing with, because it doesn't ever use attributes that require step or reverse.

CharlieL7:
> 

Slice is not used in Resnet50, it's in Yolov4 and Bert

src/include/migraphx/op/slice.hpp
```diff
@@ -120,15 +142,33 @@ struct slice
...
 
-    argument compute(shape output_shape, std::vector<argument> args) const
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
     {
-        auto input  = args[0];
-        auto offset = compute_offset(input.get_shape()) * output_shape.type_size();
-        return {std::move(output_shape), [=] { return input.data() + offset; }};
+        auto input = args[0];
+        if(dyn_out.computed_shape.dynamic())
+            return {dyn_out.computed_shape, [=] { return input.data(); }};
```
umangyadav, line 166: this  line is not covered by codecov, Do we have Ref test to cover this ?

test/ref_ops_test.cpp
```diff
@@ -7188,6 +7188,68 @@ TEST_CASE(slice_test)
...
+            x);
+
+        migraphx::shape s2{migraphx::shape::int32_type, {{2, 2, 0}, {2, 2, 0}, {2, 2, 0}}};
+        EXPECT(p.get_output_shapes().back() == s2);
+        p.compile(migraphx::ref::target{});
+        migraphx::shape sresult{migraphx::shape::int32_type, {2, 2, 2}, {6, 3, 1}};
+
+        migraphx::shape input_fixed_shape0{migraphx::shape::int32_type, {2, 2, 3}};
+        migraphx::parameter_map params0;
+        params0["x"] = migraphx::argument(input_fixed_shape0, data.data());
```
umangyadav, line 7242: don't suffix `0` unless necessary.

test/ref_ops_test.cpp
```diff
@@ -7188,6 +7188,68 @@ TEST_CASE(slice_test)
...
 
+TEST_CASE(slice_dyn_test)
+{
+    {
+        // Slice a single dynamic dimension
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        std::vector<int> data(2 * 2 * 3);
+        std::iota(data.begin(), data.end(), 0);
+        migraphx::shape s{migraphx::shape::int32_type, {{2, 2, 0}, {2, 2, 0}, {3, 3, 0}}};
```
umangyadav, line 7199: This is fixed dynamic shape. Can you make dims other than axes dynamic ?

test/onnx/gen_onnx.py
```diff
@@ -5790,6 +5790,132 @@ def slice_test():
...
+                                 axes=[0, 1],
+                                 starts=[1, 0],
+                                 ends=[2, 2],
+                                 outputs=['1'])
+
+    return ([node], [x], [y])
+
+
+@onnx_test
+def slice_step_dyn_test():
```
umangyadav, line 5809: Since this is only used inside onnx_test, you could have just kept using literal values (e.g. as in slice_dyn_test). instead of making `nodes` for `axes`, `steps`, `end` etc. It doesn't hurt though.  You can keep this suggestion for future to minimize code changes. No need to change for this PR.

test/op_shape_test.cpp
```diff
@@ -2305,6 +2305,40 @@ TEST_CASE(slice_shape)
...
+    // Todo:  When support is added for non-fixed inputs, test  +/- start/end; clipping values too
+    // large or small;
+    //    what happens if min is less than slice size?
+
+    // // Slice axis 1 from dyn. range 2..7 to range 2..3
+    // // Non-fixed input leads to either fixed shape or clipping; shouldn't change the other axes
+    // input = {migraphx::shape::int32_type, {{2, 5, 3}, {2, 7, 3}, {2, 4, 0}}};
+    // expect_shape(migraphx::shape{migraphx::shape::int32_type, {{2, 5, 3}, {2, 3, 3}, {2, 4, 0}}},
+    //              migraphx::make_op("slice", {{"axes", {1}}, {"starts", {1}}, {"ends", {4}}}),
+    //              input);
```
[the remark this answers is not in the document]
bpickrel: I think I commented this out because even though it basically works, I haven't coded all the out-of-bounds and special case handling.  I added a check that disallows non-fixed shapes for now.

src/include/migraphx/op/slice.hpp
```diff
@@ -98,10 +100,30 @@ struct slice
 
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        auto input_shape        = inputs[0];
-        auto t                  = input_shape.type();
-        const auto& old_lens    = input_shape.lens();
-        const auto& old_strides = input_shape.strides();
+        auto input_shape = inputs[0];
```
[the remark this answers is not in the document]
bpickrel: done

src/include/migraphx/op/slice.hpp
```diff
@@ -98,10 +100,30 @@ struct slice
...
+        {
+            old_lens = input_shape.lens();
+            // For static shape (including eval after a dynamic input) the strides are indexed into
+            // the pre-slice array, so they are larger than the apparent size of the resulting
+            // shape.
+            old_strides = input_shape.strides();
+        }
 
         if(std::any_of(
                axes.begin(), axes.end(), [&](auto i) { return (i >= old_lens.size() and i < 0); }))
```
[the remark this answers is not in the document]
bpickrel: done

test/ref_ops_test.cpp
```diff
@@ -7188,6 +7188,68 @@ TEST_CASE(slice_test)
...
+        auto result  = p.eval(params0).back();
+
+        std::vector<int> gold = {1, 2, 4, 5, 7, 8, 10, 11};
+        std::vector<int> results_vector(2 * 2 * 2);
+        result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+        EXPECT(migraphx::verify_range(results_vector, gold));
+        EXPECT(result.get_shape() == sresult);
+    }
+
+    {
```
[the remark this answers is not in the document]
bpickrel: done

test/ref_ops_test.cpp
```diff
@@ -7188,6 +7188,68 @@ TEST_CASE(slice_test)
...
+        migraphx::shape s2{migraphx::shape::int32_type, {{2, 2, 0}, {2, 2, 0}, {2, 2, 0}}};
+        EXPECT(p.get_output_shapes().back() == s2);
+        p.compile(migraphx::ref::target{});
+
+        //  the strides of sresult are those of the original shape, not
+        // reduced to sliced size.
+        migraphx::shape sresult{migraphx::shape::int32_type, {2, 2, 2}, {6, 3, 1}};
+        migraphx::shape input_fixed_shape0{migraphx::shape::int32_type, {2, 2, 3}};
+        migraphx::parameter_map params0;
+        params0["x"] = migraphx::argument(input_fixed_shape0, data.data());
```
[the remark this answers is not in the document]
bpickrel: done

test/ref_ops_test.cpp
```diff
@@ -7188,6 +7188,68 @@ TEST_CASE(slice_test)
...
 }
 
+TEST_CASE(slice_dyn_test)
+{
+    {
+        // Slice a single dynamic dimension
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        std::vector<int> data(2 * 2 * 3);
+        std::iota(data.begin(), data.end(), 0);
```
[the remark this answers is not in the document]
bpickrel: done

test/ref_ops_test.cpp
bpickrel: done.  Makes the test more comprehensive.

test/ref_ops_test.cpp
bpickrel: done

bpickrel:
> > Onnx spec allows a no-op "slicing" of dims that are smaller than the given slice size.
> 
> If we don't allow `no op` slicing then, we should enforce that.

I think it's OK.  By no-op, I mean that, for instance, a slice of a dimension size 3 with index of 0:7 passes through the full 3 dimension.  A quick check shows that a large slice end gives sensible results, but to check for (and test) all the permutations would require a lot of new code.  I think we should allow this, but leave careful checking as a "to-do" item.

src/include/migraphx/op/slice.hpp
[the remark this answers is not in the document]
umangyadav: Where is this ? I don't see it.

src/include/migraphx/op/slice.hpp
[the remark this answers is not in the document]
umangyadav: I don't see change here as well.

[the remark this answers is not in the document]
bpickrel: I found a bigger problem when I tried this:  the test I wrote for an (invalid) negative axis doesn't work because the negative value wasn't passed along to the Slice operator.  I added debug code to print the axes and a -1 test input gets converted to 1.  Does anyone know why that would be?  It's not because of the axes vector itself, which contains signed integers.

src/include/migraphx/op/slice.hpp
bpickrel: That entire _if_ block shouldn't even be there--removing.

bpickrel:
> @bpickrel I dont' see any commits related to comments pushed here but comments have been resolved. Can you check if you have pushed changes or not ?

Sorry, I made comments and then decided the commit wasn't ready to push.  I just pushed a commit that answers most of the resolved comments.  The biggest change is that I decided it __is__ legal to slice a non-fixed dimension, and it wasn't really that difficult to implement.  Basically, if the sliced size is larger than the max it's allowed (Numpy allows that too); if it's smaller than the max then the max is reduced but min isn't; and if it's smaller than the min then both max and min are reduced, leading to a fixed dimension.

Please let me know if you agree with this implementation.

test/ref_ops_test.cpp
```diff
@@ -7188,6 +7188,67 @@ TEST_CASE(slice_test)
...
+    EXPECT(migraphx::verify_range(results_vector, gold));
+    EXPECT(result.get_shape() == sresult);
+}
+
+TEST_CASE(slice_dyn_test1)
+{
+    // Slice all three dynamic dimensions
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape s{migraphx::shape::int32_type, {{2, 2, 0}, {2, 2, 0}, {3, 3, 0}}};
```
umangyadav, line 7226: Can you make this not fixed lengths ? 
also between these two `slice_dyn_test0` and `slice_dyn_test1` make one of the test such where slice_lengths is greater than`max` allowed length and in another axis or test  slice length is lower than `min` length.

umangyadav:
> Please let me know if you agree with this implementation. 

Looks right to me.

src/include/migraphx/op/slice.hpp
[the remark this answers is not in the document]
umangyadav: It could be due to `normalize_attributes()`. before calling to `normalize_compute_shape()`, it first calls into `normalize_attributes()`.

test/ref_ops_test.cpp
bpickrel: Done.  Note that for the min case the result is perfectly sensible:  the sliced tensor is smaller than the original and the output shape is a smaller dimension than the input shape.

test/op_shape_test.cpp
```diff
@@ -2303,6 +2303,65 @@ TEST_CASE(slice_shape)
...
+                 input);
+}
+
+TEST_CASE(slice_dyn_shape1)
+{
+    migraphx::shape input{migraphx::shape::int32_type, {{2, 3, 0}, {7, 8, 0}, {2, 3, 0}}};
+    // Slice axis 1 to size 4-1=3, negative index
+    expect_shape(migraphx::shape{migraphx::shape::int32_type, {{2, 3, 0}, {3, 3, 0}, {2, 3, 0}}},
+                 migraphx::make_op("slice", {{"axes", {1}}, {"starts", {1}}, {"ends", {-4}}}),
+                 input);
```
CharlieL7, line 2327: Doesn't look right. -4 index for a 7 length vector is index 3. So `start = 1` and `end = 3`. Since the end, is exclusive the slice  should have length 2.

src/include/migraphx/op/slice.hpp
```diff
@@ -98,37 +100,75 @@ struct slice
...
         std::vector<std::size_t> new_lens = old_lens;
         for(std::size_t i = 0; i < axes.size(); i++)
         {
             auto axis = axes[i];
-            new_lens[axis] =
+            auto sliced_length =
                 fix_index(old_lens, axis, ends[i]) - fix_index(old_lens, axis, starts[i]);
+            // A Numpy indexing convention: a slice size larger than the actual dimension
+            // is legal and is clipped to the axis size
+            new_lens[axis] = std::min(new_lens[axis], sliced_length);
```
CharlieL7, line 148: Doesn't look right.
Say `a` is a vector of length `n`.
In numpy you can do a slice `a[n-2:n+5]`. This would get you just the last value of `a`, so `length = 1`. The `sliced_length` calculated for this example would evaluate to `n+5 - n-2 = 3 -> length = 3`

CharlieL7 (changes requested):

src/include/migraphx/op/slice.hpp
CharlieL7: I think there's a bug with `fix_index()`. It's using `lens[axis]` rather than `lens[axis]-1` for the maximum index.

src/include/migraphx/op/slice.hpp
```diff
@@ -98,37 +100,75 @@ struct slice
...
+            if(input_shape.dynamic())
+            {
+                new_mins[axis] = std::min(new_lens[axis], new_mins[axis]);
+                if(new_opts[axis] < new_mins[axis] or new_opts[axis] > new_lens[axis])
+                    new_opts[axis] = 0;
```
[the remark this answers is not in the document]
bpickrel: What would you suggest?  This says that if the opt is still within the min/max range, leave it alone; if not, set it to 0.  It's scarcely even a calcualtion.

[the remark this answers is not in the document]
CharlieL7: Here's what I'm thinking:
data shape: [{1, 4}, {4, 10, 8}]
axes: [1]
starts:[2]
ends:[7]

output: [{1, 4}, {2, 5, 5}]

src/include/migraphx/op/slice.hpp
bpickrel: Rechecked this, and the existing logic stands.  The maximum value in `fix_index()` is lens[axis] because it's really the max. "ends" value which is 1 more than the last index in the slice.  I will clarify in a comment.  In your example, the slice `a[n-2:n+5]` would result in `(a[n-2], a[n-1])` so 2 is the correct length.

bpickrel: Adding a related fix: the existing code only calculates the maxes of the new output shape.  Mins need to be calculated separately too.

test/op_shape_test.cpp
bpickrel: You are correct.  This test is giving the wrong result.  The underlying cause is that negative values in the attribute `ends` aren't being passed correctly.  This debug output shows that when values of {-4, -3} were set, the array received by the operator was [4, 0]:  (ignore other incorrect values)

`bpickrel@home-tower:~/AMDMIGraphX/build$ bin/test_op_shape_test slice_dyn_shape1
[   RUN    ] slice_dyn_shape1
ends:  4 0 
ends:  4 0 
FAILED: Incorrect shape for slice[axes={1, 2},starts={1, 1},ends={-4, -3}]: dynamic, int32_type, {[2, 3, 0], [2, 3, 0], [2, 3, 0]} != dynamic, int32_type, {[2, 3, 0], [3, 3, 0], [2, 3, 0]}
    dynamic, int32_type, {[2, 3, 0], [7, 8, 0], [2, 3, 0]}
[ COMPLETE ] slice_dyn_shape1`

src/include/migraphx/op/slice.hpp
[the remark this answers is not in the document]
bpickrel: In other words, when the new max is less than the opt; you set the opt to the new max value.  To extend the logic, whenever the old opt is outside the new min/max range you clamp it?  That makes sense too, but I don't know if it's what a user would want because I don't know what opts are used for.

src/include/migraphx/op/slice.hpp
```diff
@@ -66,7 +66,8 @@ struct slice
 
     std::string name() const { return "slice"; }
 
-    // Convert negative index to a value counting backwards from the max index, per numpy convention
+    // Convert negative index to a value counting backwards from the max index, per numpy slicing
+    // convention, and clip large values to 1 more than the largest index.
```
umangyadav, line 70: This comment is a bit confusing.
It is clipping larger values to the largest possible value. not 1 more. 

also we should check if `index` is within bounds for negative values. 
e.g. for the dimensions `7` if the index is -8.

src/include/migraphx/op/slice.hpp
```diff
@@ -144,11 +146,12 @@ struct slice
...
                 fix_index(old_lens, axis, ends[i]) - fix_index(old_lens, axis, starts[i]);
             // A Numpy indexing convention: a slice size larger than the actual dimension
-            // is legal and is clipped to the axis size
+            // is legal and the "ends" value is clipped to the axis size
             new_lens[axis] = std::min(new_lens[axis], sliced_length);
             if(input_shape.dynamic())
             {
-                new_mins[axis] = std::min(new_lens[axis], new_mins[axis]);
+                auto sliced_min_length = fix_index(new_mins, axis, ends[i]) - fix_index(new_mins, axis, starts[i]);
+                new_mins[axis] = std::min(sliced_min_length, new_mins[axis]);
```
umangyadav, line 154: I would find it better if we used `old_mins`, `old_maxs` etc. as names. 
Here we are using `new_mins` to calculuate and values to `new_mins` which defeats the `new` meaning

src/include/migraphx/op/slice.hpp
```diff
@@ -98,37 +100,75 @@ struct slice
...
-               axes.begin(), axes.end(), [&](auto i) { return (i >= old_lens.size() and i < 0); }))
+               axes.begin(), axes.end(), [&](auto i) { return (i >= old_lens.size() or i < 0); }))
         {
             MIGRAPHX_THROW("SLICE: input axis " + to_string_range(axes) + " out of range");
         }
 
         if(starts.size() != axes.size() or axes.size() != ends.size())
         {
-            MIGRAPHX_THROW("SLICE: inconsistent sizes");
+            MIGRAPHX_THROW("SLICE: inputs have inconsistent sizes");
```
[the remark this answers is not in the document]
bpickrel: changed

src/include/migraphx/op/slice.hpp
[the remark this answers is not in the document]
umangyadav: axes values are parsed negative from ONNX, but it is during evaluation time before calling `compute_shape()` for slice operator, MIGraphX would call `normalize_attributes()` which would convert those negative value to positive ones.

src/include/migraphx/op/slice.hpp
bpickrel: The comment may be a bit confusing, but it's accurate.  The largest possible value is 1 more than the largest index.

CharlieL7: I understand what fix_index() is trying to do now. The comment could say that the end is exclusive.

src/include/migraphx/op/slice.hpp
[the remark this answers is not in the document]
CharlieL7: `opt` is essentially just a suggested value in the range. It should be calculated like a single static dimension; which is often what happens for min and max too.

test/op_shape_test.cpp
bpickrel: Learned that negative inputs are being intercepted by the normalize_attributes op.  We'd like the allowed range to be [-L, L] where L is the size of the N'th dimension.  Unfortunately L needs two different values for max and min lengths, and normalize_attributes doesn't (yet) handle dynamic dimensions.  I think it won't handle L having different values for different axis numbers either.

Options for dealing with this are:

- cut "starts" and "ends" out of the attribute() method code and handle range checking explicitly in normalize_compute_shape().  This makes sense if the requirements are specific to this op. only and generic normalize_axes handling shouldn't be expected to handle them (is this true?)
- add dynamic shape support to normalize_axes .
- don't support negative length inputs in the Slice operator for now; wait until normalize_axes is updated as a separate task, and remove test case with negative input for now.

bpickrel: Another issue that came up during debugging:  our notation can't handle slices that include the end of a non-fixed dynamic dimension array.  Example:  suppose array A has size 5 (indexes from 0 to 4).  In Numpy, A[1:] is a slice from the 1th to last element, slicing off the 0'th element.  We don't support leaving a value blank in "ends", so the only way to make this slice is with starts={1}, ends={5}.  But if we're slicing an array with dynamic dimension with a min and max length, no constant "ends" value will fit both.  Note that negative indexing doesn't solve this, because ends={-1} refers to A[4] and the notation requires that the index be one greater than the last element you want--in other words, A[1 : -1] gives a slice with elements A(1), A(2), and A(3) only.

src/include/migraphx/op/slice.hpp
bpickrel: I'd like to leave this as is.  I think it's already about as clear as any alternative I can think of.  The new_mins array is copied from the inputs, then one or a few of its values are updated.

test/op_shape_test.cpp
umangyadav: `start`, `end` and `axes` are fixed literal  values. `normalize_attributes` is only called on those three attributes. 
Therefore, it doesn't need to handle dynamic shapes.

umangyadav: > But if we're slicing an array with dynamic dimension with a min and max length, no constant "ends" value will fit both
Wouldn't `end[i]` value equal to `max_lens[i]` where `i` is axis, work ?

src/include/migraphx/op/slice.hpp
```diff
@@ -83,52 +76,100 @@ struct slice
...
         for(std::size_t i = 0; i < axes.size(); i++)
         {
-            auto axis = axes[i];
-            new_lens[axis] =
-                fix_index(old_lens, axis, ends[i]) - fix_index(old_lens, axis, starts[i]);
+            auto axis            = axes[i];
+            size_t sliced_length = ends[i] - starts[i];
+            // A Numpy indexing convention: a slice size larger than the actual dimension
+            // is legal and the "ends" value is clipped to the axis size
+            new_lens[axis] = std::min(new_lens[axis], sliced_length);
```
CharlieL7, line 143: With how normalize_attributes works is this still needed?

test/ref_ops_test.cpp
```diff
@@ -7188,6 +7188,69 @@ TEST_CASE(slice_test)
...
+    // Slice a single dynamic dimension. ax1 slice limits are smaller than min; ax2 "ends" is too
+    // large
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape s{migraphx::shape::int32_type, {{2, 3, 0}, {2, 2, 0}, {3, 3, 0}}};
+    auto x = mm->add_parameter("x", s);
+    mm->add_instruction(
+        migraphx::make_op("slice", {{"axes", {1, 2}}, {"starts", {0, 1}}, {"ends", {1, 6}}}), x);
+    migraphx::shape s2{migraphx::shape::int32_type, {{2, 3, 0}, {1, 1, 0}, {2, 2, 0}}};
+    EXPECT(p.get_output_shapes().back() == s2);
```
CharlieL7, line 7202: This looks like an op_shape_test, why are we testing this again here?

CharlieL7 (changes requested):
Minor comments

bpickrel:
The latest test failure, in CLANG ASAN, is down to a normalize_attributes problem.  This code snippet at line 53 of [normalize_attributes.cpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/dyn_slice/src/normalize_attributes.cpp)

shows lens[i] being used without making sure of the number of axes.
```
    std::vector<int64_t> max_vals(vec.size(), n_rank);
    if(contains(vec_attrs, op::normalize_attribute::use_len))
    {
        std::transform(axes.begin(), axes.end(), max_vals.begin(), [&](auto i) { return lens[i]; });
    }
```
This failed in test slice_dyn_shape6, which deliberately sends mismatched attributes.
**How do we want to fix it?**

bpickrel:
> 

I can add code to throw an exception if(i > lens.size()), but I'm wondering--shouldn't axis values have been normalized to the right range before this code is hit?  This occurs in  tune_attribute().

bpickrel:
> 

@CharlieL7 the question about normalize_attributes is really for you.

test/ref_ops_test.cpp
bpickrel: Following the pattern of the tests that were there before.

bpickrel:
Deleted the shape test case slice_dyn_shape6, which tested attribute vectors of mismatched lengths ("axes" was bigger than the others).  The sanitizer test showed this caused a memory access problem in normalize_attributes before normalize_compute_shape() was reached, so the check I put there was ineffective--removed.  I strongly recommend revising normalize_attributes  to handle this situation gracefully, but it's outside the scope of this PR.

src/include/migraphx/op/slice.hpp
```diff
@@ -83,52 +80,94 @@ struct slice
...
-        const auto& old_lens    = input_shape.lens();
-        const auto& old_strides = input_shape.strides();
+        check_shapes{inputs, *this, true}.has(1);
+        auto input_shape = inputs[0];
+        auto t           = input_shape.type();
 
-        if(std::any_of(
-               axes.begin(), axes.end(), [&](auto i) { return (i >= old_lens.size() and i < 0); }))
+        // TODO:  When support for dynamic shapes is added to normalize_attributes,
+        //  remove this restriction.
```
umangyadav, line 103: I don't see how normalize attributes would solve this limitation.

src/include/migraphx/op/slice.hpp
```diff
@@ -83,52 +80,94 @@ struct slice
...
-            new_lens[axis] =
-                fix_index(old_lens, axis, ends[i]) - fix_index(old_lens, axis, starts[i]);
+            auto axis            = axes[i];
+            size_t sliced_length = ends[i] - starts[i];
+            // A Numpy indexing convention: a slice size larger than the actual dimension
+            // is legal and the "ends" value is clipped to the axis size
+            new_lens[axis] = std::min(new_lens[axis], sliced_length);
+            if(input_shape.dynamic())
+            {
+                std::size_t sliced_min_length = ends[i] - starts[i];
```
umangyadav, line 144: you can re-use `sliced_length` calculated before.

test/op_shape_test.cpp
```diff
@@ -2374,6 +2374,65 @@ TEST_CASE(slice_shape)
...
+    // Sliced range max bigger than dimension; is clipped
+    expect_shape(migraphx::shape{migraphx::shape::int32_type, {{2, 3, 0}, {6, 6, 0}, {2, 3, 0}}},
+                 migraphx::make_op("slice", {{"axes", {1}}, {"starts", {1}}, {"ends", {10}}}),
+                 input);
+}
+
+// TEST_CASE(slice_dyn_shape3)
+// {
+//     migraphx::shape input{migraphx::shape::int32_type, {{2, 3, 0}, {7, 8, 0}, {2, 3, 0}}};
+//     // TODO: When variable dimension slicing is allowed, Slice to a size smaller than min.
```
umangyadav, line 2413: enable this test

bpickrel: The test is expected to cause an exception until the restriction is removed.  I'll enable a 
"throws_shape" test with a note as to what should be expected when it works.

src/include/migraphx/op/slice.hpp
bpickrel: I think it's more logical to keep this in place .  It's correct behavior and there's nothing to be gained by introducing one more dependency on code that's still in flux.

CharlieL7 (approved):

umangyadav (approved):

PR #1506, opened 2023-01-05
Add trap operator

TedThemistokleous:
use this to handle stuff that's parsed in but invalid. Allows us to continue parsing in code, before passes and optimizations to determine if the onnx resolves to a degenerate configuration during evaluation.

### DIFF
--- a/src/include/migraphx/trap.hpp
+++ b/src/include/migraphx/trap.hpp
@@ -1,0 +1,52 @@
+/*
+ * The MIT License (MIT)
+ *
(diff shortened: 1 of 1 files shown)


pfultz2:
Are you able to add in this PR the `parse_constant` changes that will use the `trap`?

TedThemistokleous:
@pfultz2 do we still want an operator like this for running into invalid configs? Looks like the original target for this was closed (parse_if related stuff)

TedThemistokleous:
Closing and to revisit later

PR #1509, opened 2023-01-07
Add basic Trilu op support

kahmed10:
- supports upper and lower trilu
- not able to support dynamic k values
- not able to support negative k values


src/onnx/parse_trilu.cpp
```diff
@@ -0,0 +1,86 @@
...
+        shape::type_t output_type = args[0]->get_shape().type();
+
+        // when creating the mask, if upper == 1,
+        // the inner triangle will have values set to 0
+        std::vector<bool> mask_mat(num_rows * num_cols, upper);
+        for(size_t i = 0; i < num_rows; i++)
+        {
+            for(size_t j = 0; j < k; j++)
+            {
+                mask_mat[i * num_cols + j] = not upper;
```
umangyadav, line 72: Value of k can be larger than available number of diagnonals. This can go out of bounds.

src/onnx/parse_trilu.cpp
```diff
@@ -0,0 +1,86 @@
...
+        size_t num_rows = *(input_lens.rbegin() + 1);
+        size_t num_cols = input_lens.back();
+        size_t k        = 0;
+        bool upper      = true;
+
+        if(args.size() > 1)
+        {
+            auto arg_k = args[1]->eval();
+            check_arg_empty(arg_k, "PARSE_TRILU: dynamic k not supported");
+            k = arg_k.at<size_t>();
```
umangyadav, line 55: Can you assert and throw if k is negative?

test/onnx/verify_onnx.cpp
```diff
@@ -1370,4 +1370,41 @@ TEST_CASE(where_test)
...
+
+    migraphx::parameter_map pp;
+    pp["x"] = migraphx::argument(sx, x_data.data());
+
+    auto result = p.eval(pp).back();
+    std::vector<float> result_vector;
+    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
+    return result_vector;
+}
+TEST_CASE(trilu_test)
```
umangyadav, line 1388: Can we add more tests ? like number of rows == 1 and with batch size ? 
ONNX doc has extensive list of examples. We can borrow tests from there. https://github.com/onnx/onnx/blob/main/docs/Operators.md#Trilu

umangyadav (changes requested):

test/onnx/gen_onnx.py
```diff
@@ -6551,6 +6551,28 @@ def transpose_gather_test():
...
+    )
+    return ([node], [x], [y])
+
+
+@onnx_test()
+def trilu_lower_test():
+    x = helper.make_tensor_value_info('x', TensorProto.FLOAT, [3, 4])
+    y = helper.make_tensor_value_info('y', TensorProto.FLOAT, [3, 4])
+
+    node = onnx.helper.make_node('Trilu', inputs=['x'], outputs=['y'], upper=0)
```
umangyadav, line 6572: These tests are not using `k`.  For the Lower triangle value of `k` should be negative if to get diagnoals below main diagnoal. We need to support negative k.

src/onnx/parse_trilu.cpp
```diff
@@ -0,0 +1,86 @@
...
+            {
+                mask_mat[i * num_cols + j] = not upper;
+            }
+            k++;
+        }
+
+        auto mask =
+            info.add_literal(migraphx::literal{migraphx::shape{output_type, input_lens}, mask_mat});
+
+        return info.add_instruction(make_op("mul"), mask, args[0]);
```
CharlieL7, line 80: I think you want to use `add_broadcastable_binary_op()` since there can be batches.

CharlieL7 (changes requested):
* The mask to a element-wise multiplication method looks fine for this op for a constant `k`
* I think we could support negative `k` if needed, your call if we want that
* The ONNX spec on this op is a little out of whack. It says the input can be of rank >= 2, but it is only defined for rank 2 or 3.

src/onnx/parse_trilu.cpp
kahmed10: done

test/onnx/gen_onnx.py
kahmed10: going to support negative k later, for now I made a test that throws if k is negative

test/onnx/verify_onnx.cpp
kahmed10: I added a few other tests, but I will add more at a future date once we get more support for the operator.

src/onnx/parse_trilu.cpp
kahmed10: done

test/onnx/verify_onnx.cpp
umangyadav: Current logic would work with batch dimension and also if row ==1. IMO we should add tests in this PR than later, it shouldn't be much work to add those tests.

CharlieL7:
I would like to see Trilu over batches test, will then approve.

test/onnx/verify_onnx.cpp
kahmed10: added

kahmed10:
> I would like to see Trilu over batches test, will then approve.

done

umangyadav (approved):

CharlieL7 (approved):

PR #1512, opened 2023-01-09
add-to-project workflow

CharlieL7:
* Error on token
* Got to be a better way to test these
  * Can test on push


CharlieL7:
Testing opening PR

PR #1514, opened 2023-01-11
[Experimental] Update accuracy checker for tf support and map input dims (custom input shape)

kahmed10:
May not work with all tf protobufs. Will need to test and implement additional features.


TedThemistokleous (approved):
Make sense. LGTM

The CI errors here don't make sense though due to this being changes to python.

tools/accuracy/accuracy_checker.py
```diff
@@ -111,10 +119,40 @@ def get_np_datatype(in_type):
...
+        for input in custom_inputs:
+            input_dim = ''.join(input.split(':')[:-1])
+            dims = [int(dim) for dim in input.split(':')[-1].split(',')]
+            input_dims[input_dim] = dims
+
+    if use_onnx:
+        if any(input_dims):
+            model = migraphx.parse_onnx(model_name,
+                                        default_dim_value=batch,
+                                        map_input_dims=input_dims)
```
umangyadav, line 146: This is only for static shapes and doesn't work for dynamic shape

tools/accuracy/accuracy_checker.py
```diff
@@ -111,10 +119,40 @@ def get_np_datatype(in_type):
...
+
+    input_dims = {}
+    if custom_inputs != None:
+        for input in custom_inputs:
+            input_dim = ''.join(input.split(':')[:-1])
+            dims = [int(dim) for dim in input.split(':')[-1].split(',')]
+            input_dims[input_dim] = dims
+
+    if use_onnx:
+        if any(input_dims):
```
umangyadav, line 143:
```suggestion
        if not input_dims:
```

tools/accuracy/accuracy_checker.py
```diff
@@ -111,10 +119,40 @@ def get_np_datatype(in_type):
...
+                                        map_input_dims=input_dims)
+        else:
+            model = migraphx.parse_onnx(model_name, default_dim_value=batch)
+    else:
+        model_name = args.tf
+        if any(input_dims):
+            model = migraphx.parse_tf(model_name,
+                                      batch_size=batch,
+                                      map_input_dims=input_dims)
+        model = migraphx.parse_tf(model_name, batch_size=batch)
```
umangyadav, line 155: Else is missing

tools/accuracy/accuracy_checker.py
```diff
@@ -125,28 +163,74 @@ def main():
...
                 get_np_datatype(in_type))
-        else:
+        elif not args.fill0:
             test_input = np.ones(in_shape).astype(get_np_datatype(in_type))
+        else:
+            test_input = np.zeros(in_shape).astype(get_np_datatype(in_type))
         test_inputs[name] = test_input
         params[name] = migraphx.to_gpu(migraphx.argument(test_input))
 
     pred_migx = np.array(migraphx.from_gpu(model.run(params)[-1]))
```
umangyadav, line 179: I think you can just set `offload_copy=True`

tools/accuracy/accuracy_checker.py
```diff
@@ -125,28 +163,74 @@ def main():
...
 
-    sess = ort.InferenceSession(model_name, providers=[args.provider])
+    if use_onnx:
+        sess = ort.InferenceSession(model_name, providers=[args.provider])
+
+        ort_params = {}
+        for input in sess.get_inputs():
+            ort_params[input.name] = test_inputs[input.name]
```
umangyadav, line 186: looks like it is just copying . Is there possibility that inputs could differ otherwise,
you can set `ort_params = test_inputs`.

tools/accuracy/accuracy_checker.py
```diff
@@ -125,28 +163,74 @@ def main():
...
+
+        try:
+            pred_fw = sess.run(None, ort_params)[-1]
+        except:
+            if any(input_dims):
+                print(
+                    'Error: custom input dim not compatible with onnx runtime')
```
umangyadav, line 193: Add `else` clause in case it fails without custom inputs dims ?

tools/accuracy/accuracy_checker.py
```diff
@@ -125,28 +163,74 @@ def main():
...
+        import tensorflow as tf
+
+        def load_tf_graph(model_name):
+            with tf.io.gfile.GFile(model_name, 'rb') as f:
+                graph_def = tf.compat.v1.GraphDef()
+                graph_def.ParseFromString(f.read())
+
+            with tf.compat.v1.Graph().as_default() as graph:
+                tf.graph_util.import_graph_def(graph_def)
+            return graph
```
umangyadav, line 205: Why not use TFv2 ?

tools/accuracy/accuracy_checker.py
```diff
@@ -125,28 +163,74 @@ def main():
...
+        for name in test_inputs.keys():
+            # graph.get_operations() adds 'import/' to the op name
+            tf_name = f'import/{name}'
+            if tf_name not in graph_ops_set:
+                continue
+            x = graph.get_tensor_by_name(f'{tf_name}:0')
+            tf_input = test_inputs[name]
+            # lazy check for image model (NHWC layout)
+            if tf_input.ndim == 4:
+                tf_dict[x] = np.transpose(tf_input, (0, 2, 3, 1))
```
umangyadav, line 221: We should check some other way, this could fail easily.

tools/accuracy/accuracy_checker.py
```diff
@@ -125,28 +163,74 @@ def main():
...
+        graph_ops = [op.name for op in graph.get_operations()]
+        graph_ops_set = set(graph_ops)
+        tf_dict = {}
+
+        for name in test_inputs.keys():
+            # graph.get_operations() adds 'import/' to the op name
+            tf_name = f'import/{name}'
+            if tf_name not in graph_ops_set:
+                continue
+            x = graph.get_tensor_by_name(f'{tf_name}:0')
```
umangyadav, line 217: I remember for one of the model earlier, tensor name was not appended by `:0`. This could  fail for such models.

umangyadav, line 217: I remember for one of the models earlier, tensor name was not appended by `:0`. This could  fail for such models.

umangyadav, line 217: I remember for one of the models earlier, tensor name was not appended by `:0`. This could  fail for such models.

umangyadav (commented):
is this supposed to handle dynamic shapes ?

tools/accuracy/accuracy_checker.py
kahmed10: will fix

tools/accuracy/accuracy_checker.py
kahmed10: will reverse order of checks

tools/accuracy/accuracy_checker.py
kahmed10: I tried that originally, but for some reason I was getting errors on larger batch sizes.

tools/accuracy/accuracy_checker.py
kahmed10: This works with TF2 as well.

tools/accuracy/accuracy_checker.py
umangyadav: okay, PR title mentions `dynamic input shape` so i thought that.  

by dynamic you mean users can specify batch size and input dims then ?

tools/accuracy/accuracy_checker.py
kahmed10: If you can find that model again, please let me know. This call should be different in that it does not require the `:0` when first checking if the name is the set of graph models. The `:0` is then later appended as convention before passing into `tf_dict`.

tools/accuracy/accuracy_checker.py
kahmed10: This should throw earlier when parsing MIGraphX if that was the case.

tools/accuracy/accuracy_checker.py
umangyadav: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1298

umangyadav: From that PR that it seems we used to have models without `:0`.  I think we can keep `:0` for now and modify as required later.

tools/accuracy/accuracy_checker.py
kahmed10: updated with better check

tools/accuracy/accuracy_checker.py
kahmed10: updated PR title

tools/accuracy/accuracy_checker.py
umangyadav: what were the errors ?

tools/accuracy/accuracy_checker.py
```diff
@@ -125,28 +168,80 @@ def main():
...
+                tf.graph_util.import_graph_def(graph_def)
+            return graph
+
+        graph = load_tf_graph(model_name)
+        is_nhwc = False
+        graph_ops = []
+        for op in graph.get_operations():
+            graph_ops.append(op.name)
+            if 'Conv' in op.node_def.op and not args.nchw:
+                is_nhwc = True
```
umangyadav, line 219: https://www.tensorflow.org/api_docs/python/tf/Operation#get_attr 

you can use `get_attr('data_format")`

tools/accuracy/accuracy_checker.py
kahmed10: I can't seem to reproduce anymore, so I've just updated to have offload_copy=True.

umangyadav (approved):

PR #1515, opened 2023-01-11
Use float accumulator when reduction size is too large for half

pfultz2:


src/targets/gpu/jit/reduce.cpp
```diff
@@ -156,16 +156,25 @@ struct reduce_compiler : compiler<reduce_compiler>
...
         else if(op.name() == "reduce_mean")
         {
-            v["reduction"] = "op::sum{}";
-            v["write"]     = "op::mean{" + std::to_string(reduce_elements) + "}";
+            auto reduce_elements = get_reduce_elements(ins->inputs());
+            auto reduce_type     = ins->inputs().front()->get_shape().type();
+            v["reduction"]       = "op::sum{}";
+            std::string mean     = "op::mean{" + std::to_string(reduce_elements) + "}";
+            // Use float accumulator when reduction size is too large for half
+            if(reduce_type == shape::half_type and reduce_elements > 16384)
```
kahmed10, line 171: How was 16384 chosen?

pfultz2: The max value is 65535, so this is 65536/4 so it should allow 2 bits of precision, but I didnt try out where the threshold would be to get accurate results.

pfultz2: So I checked and 16383 does pass verify so it looks like a good threshold.

kahmed10 (approved):

PR #1519, opened 2023-01-11
Add flag for tuning in migraphx-driver

umangyadav:
Adds `--exhaustive-tune` flag in `migraphx-driver` to enable MIOpen tuning. 
Changes C/C++ and Python API to also support this. 

~It does this by setting the environment flag MIOPEN_FIND_ENFORCE internally when compiling instead of relying on users.~

It does this by setting `exhaustive-search` flag in MIOpen's Find API.


pfultz2:
> Adds --tune flag in migraphx-driver to enable MIOpen tuning.

I think this flag should be called `--exhaustive-search` as it is more descriptive of what it is doing.

> It does this by setting the environment flag MIOPEN_FIND_ENFORCE internally when compiling instead of relying on users.

We should not be setting the env variable. MIOpen caches the env variables so this can be wrong if it does two compiles with different settings in the same process.

We should be using MIOpen's API. For Find 1.0, they have a [`exhaustiveSearch`](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/include/miopen/miopen.h#L1622) flag that could be passed to enable tuning. For find 2.0, I believe there is the [`miopenSetFindOptionTuning`](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/include/miopen/miopen.h#L4840) to enable tuning in MIOpen similar to `exhaustiveSearch`.

src/driver/main.cpp
```diff
@@ -328,6 +328,7 @@ struct compiler
     compiler_target ct;
     bool offload_copy  = false;
     bool fast_math     = true;
+    bool tune          = false;
```
pfultz2, line 331: I think it would be better to store `compiler_options` struct here, and then in `parse` just do `ap(co.offload_copy, ...)`(assuming it was named `co`).

umangyadav: Done.

umangyadav:
> I think this flag should be called `--exhaustive-search` as it is more descriptive of what it is doing.

I find `--tune` more intuitive for the users to understand. Also In the future, we may want to use this flag for rocBLAS/CK/MLIR tuning.

doc/src/driver/compile.rst
```diff
@@ -28,6 +28,10 @@ Enable implicit offload copying
 
 Disable fast math optimization
 
+.. option:: --tune
+
+Enable tuning for the GPU kernels
```
pfultz2, line 33: This should say "Perform an exhaustive search to find the fastest version.", this flag doesnt need to apply to just the GPU kernels.

umangyadav: Done.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -59,6 +59,7 @@ struct miopen_convolution
 {
     Op op;
     bool int8_x4_format               = false;
+    bool tune                         = false;
```
pfultz2, line 62: This should go into the `context` and not the operator.

pfultz2:
> I find --tune more intuitive for the users to understand.

Also, we already do some tuning even without this flag. This flags enables a more comprehensive set of tunings. If you want the `tune` word then maybe this flag should be `--exhaustive-tuning`?

src/targets/gpu/include/migraphx/gpu/convolution.hpp
umangyadav: Agree. Added to context. but for that, i  had to add additional method in `context` virtual class.

umangyadav:
> > I find --tune more intuitive for the users to understand.
> 
> Also, we already do some tuning even without this flag. This flags enables a more comprehensive set of tunings. If you want the `tune` word then maybe this flag should be `--exhaustive-tuning`?

renamed to `--exhaustive-tune`

kahmed10 (approved):

src/include/migraphx/context.hpp
```diff
@@ -172,6 +188,18 @@ struct context
         (*this).private_detail_te_get_handle().from_value(v);
     }
 
+    void set_exhaustive_tune_flag(bool v)
```
pfultz2, line 191: This shouldn't be added to the `context` interface. The flag should just be added to the GPU context, and then we can set it in `get_passes`.

doc/src/driver/compile.rst
```diff
@@ -28,6 +28,10 @@ Enable implicit offload copying
 
 Disable fast math optimization
 
+.. option:: --exhaustive-tune
```
pfultz2, line 31: What about calling this flag `--exhaustive-optimize`? I think that sounds better.

doc/src/driver/compile.rst
```diff
@@ -28,6 +28,10 @@ Enable implicit offload copying
...
+
+Perform an exhaustive search to find the fastest version of generated kernels for selected backend
```
pfultz2, line 33: We could also use this flag to time different graph transformations as well, not just kernels.

umangyadav: `-tune` still sounds better to me. If we are timing different graph transformations/optimization then `tune` relates to auto-tuning.  IMO, also during the discussions, tuning has lesser overloaded meaning compared to optimize.

src/include/migraphx/context.hpp
umangyadav: Done.

TedThemistokleous (commented):
Add test case for the CI coverage. Overall looks good though

umangyadav:
> Add test case for the CI coverage. Overall looks good though

Modified API tests to set tuning flag.

TedThemistokleous (approved):

PR #1526, opened 2023-01-13
Dynamic shape support for concat op.

bpickrel:
Add dynamic shape support for the Concat operator (ref).  Includes new tests for op_shape, ref_ops, and Onnx.

For discussion:  this PR requires the shapes in the input list to either be all static, or all dynamic (and to all have identical dimensions in all but the selected axis).  Do we want to keep that restriction?  Is there any use case where the code needs to deal with inputs where different inputs have different mins or maxes, or where some inputs are static and dynamic inputs need to match them?


src/include/migraphx/op/concat.hpp
```diff
@@ -73,49 +74,103 @@ struct concat
         }
         return offsets;
     }
+
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
+        // Todo:: not using check_shapes struct; it
+        // doesn't have a method to check for at least one shape.
```
umangyadav, line 81: It can be used `.has(1, 2)`.

test/op_shape_test.cpp
```diff
@@ -2695,4 +2695,46 @@ TEST_CASE(roialign_test)
...
+{
+    migraphx::shape sx{migraphx::shape::float_type, {{1, 3, 3}, {4, 4}, {1, 5, 5}, {6, 6}}};
+    migraphx::shape sy{migraphx::shape::float_type, {{1, 3, 3}, {4, 4}, {1, 4, 4}, {6, 6}}};
+    migraphx::shape sout{migraphx::shape::float_type, {{1, 3, 3}, {4, 4, 0}, {2, 9, 0}, {6, 6}}};
+
+    expect_shape(sout, migraphx::make_op("concat", {{"axis", 2}}), sx, sy);
+
+    // rank doesn't match
+    migraphx::shape sbi1{migraphx::shape::int64_type, {2, 3}};
+    throws_shape(migraphx::make_op("concat", {{"axis", 0}}), sx, sbi1);
```
CharlieL7, line 2730: This looks like it could also be mixed static and dynamic. Could make `sbi1` a dynamic shape to be clearer.

test/onnx/onnx_test.cpp
```diff
@@ -840,6 +840,33 @@ TEST_CASE(concat_test)
...
 }
 
+TEST_CASE(concat_dyn_test)
+{
+    // migraphx::program p;
+    // auto* mm = p.get_main_module();
+    // auto l0  = mm->add_parameter("0", migraphx::shape{migraphx::shape::float_type, {2, 4, 3}});
+    // auto l1  = mm->add_parameter("1", migraphx::shape{migraphx::shape::float_type, {7, 4, 3}});
+    // mm->add_instruction(migraphx::make_op("concat", {{"axis", 0}}), l0, l1);
+    // auto prog = optimize_onnx("concat_dyn_test.onnx");
```
CharlieL7, line 850: extra comments

test/onnx/gen_onnx.py
```diff
@@ -759,6 +759,22 @@ def concat_test():
     return ([node], [x, y], [z])
 
 
+@onnx_test()
+def concat_dyn_test():
+    x = helper.make_tensor_value_info('0', TensorProto.FLOAT, [None, None, 3])
+    y = helper.make_tensor_value_info('1', TensorProto.FLOAT, [None, None, 3])
+    z = helper.make_tensor_value_info('2', TensorProto.FLOAT, [9, 4, 3])
```
CharlieL7, line 766: The output doesn't matter to MIGX, but the output shape should probably be `[None, None, 3]`

test/ref_ops_test.cpp
```diff
@@ -953,6 +953,41 @@ TEST_CASE(concat_test)
...
+
+    auto input0 = mm->add_parameter("X", s0);
+    auto input1 = mm->add_parameter("Y", s1);
+    auto input2 = mm->add_parameter("Z", s2);
+    mm->add_instruction(migraphx::make_op("concat", {{"axis", axis}}), input0, input1, input2);
+    p.compile(migraphx::ref::target{});
+
+    migraphx::shape static_shape0{migraphx::shape::int32_type, {2, 2}, {2, 1}};
+    migraphx::shape static_shape1{migraphx::shape::int32_type, {3, 2}, {2, 1}};
+    migraphx::shape static_shape2{migraphx::shape::int32_type, {1, 2}, {2, 1}};
```
CharlieL7, line 973: These look like standard strides, do we need to specify them?

CharlieL7 (changes requested):
Looks good, some minor comments. We could add a check to `check_shapes{}` for atleast one input.

src/include/migraphx/op/concat.hpp
CharlieL7: no, that won't work. It can have [1, inf] inputs.

src/include/migraphx/op/concat.hpp
```diff
@@ -73,49 +74,103 @@ struct concat
...
+                    }
                 }
             }
+            std::size_t new_dim_axis = 0;
+            for(const auto& input : inputs)
+            {
+                const auto& lens = input.lens();
+                new_dim_axis += lens[axis];
+            }
+            std::vector<std::size_t> new_lens;
```
umangyadav, line 122: You can initialize new_lens directly with `first_shape_lens` instead of first declaring and then copying.

bpickrel: done

test/onnx/gen_onnx.py
bpickrel: changed

test/op_shape_test.cpp
bpickrel: changed

src/include/migraphx/op/concat.hpp
```diff
@@ -73,49 +74,101 @@ struct concat
...
+            }
+
+            auto new_dims  = inputs[0].dyn_dims();
+            new_dims[axis] = migraphx::shape::dynamic_dimension{new_min, new_max, 0};
+            return {inputs[0].type(), new_dims};
+        }
+        else
+        {
+            MIGRAPHX_THROW("CONCAT: Cannot mix static and dynamic input shapes");
         }
```
bpickrel, line 157: I could make this legal by calling to_dynamic() on all the static shapes, but is it safe with nonstandard shapes?

test/onnx/onnx_test.cpp
bpickrel: deleted

bpickrel:
> Looks good, some minor comments. We could add a check to `check_shapes{}` for atleast one input.

I looked into doing that and found that the program never got as far as the check anyway, because add_instruction() had a segmentation fault.

bpickrel:
> Looks good, some minor comments. We could add a check to `check_shapes{}` for atleast one input.

I looked into doing that and found that the program never got as far as the check anyway, because add_instruction() had a segmentation fault.

test/ref_ops_test.cpp
bpickrel: Not strictly necessary, but it does no harm and test coverage is more robust if we don't set up every case in exactly the same way.

umangyadav: not specifying strides here, shouldn't affect test coverage for this PR.  Specifying them doesn't do any harm, but my convention if  strides are  standard shape then, in MIGraphX we would let the constructor of shape handle it.

src/include/migraphx/op/concat.hpp
umangyadav: I think it is better to not handle case of mixing them in this  PR. IF need arises we can do it later as follow up.

umangyadav:
> I looked into doing that and found that with no inputs the program never got as far as the check anyway, because add_instruction() had a segmentation fault first.

You can directly construct contact op using `make_op` without using `add_instruction` and then call `compute_shape()`  on it to trigger the error.

bpickrel:
> > I looked into doing that and found that with no inputs the program never got as far as the check anyway, because add_instruction() had a segmentation fault first.
> 
> You can directly construct contact op using `make_op` without using `add_instruction` and then call `compute_shape()` on it to trigger the error.

Charlie and I determined there was a real bug in operation.hpp.  Latest commit fixes the bug and adds the check and a test for zero input shapes.

tools/include/operation.hpp
```diff
@@ -141,7 +141,8 @@ auto compute_shape_op(rank<2>, const T& x, const std::vector<shape>& inputs)
     -> decltype(x.normalize_compute_shape(inputs))
 {
     dependent_type<operation, T> y = x;
-    normalize_attributes(y, inputs[0].max_lens());
+    if(inputs.size() > 0)
+        normalize_attributes(y, inputs[0].max_lens());
```
CharlieL7, line 145: I'm not sure if we should be allowing this to continue to normalize_compute_shape() or throwing here. @pfultz2

bpickrel: The one op. anyone identified that takes 0 arguments is Allocate--it may be allowed to continue.  Its compute_shape returns a default shape.

CharlieL7: right, but allocate doesn't use normalize_attributes.

CharlieL7: Paul and I agree that it should instead throw here in operation.hpp if there are no inputs. This will make your atleast one check_shapes not pass code coverage, so just remove the check for now.

bpickrel:
[AMD Official Use Only - General]

Ok, I'll do that.  Did you see the other, similar issue I found in the Slice PR?

src/include/migraphx/check_shapes.hpp
```diff
@@ -100,6 +100,23 @@ struct check_shapes
         return *this;
     }
 
+    /*!
+     * Require the number of shape objects to equal at least a given amount.  Use this
+     * method for ops that can take any number (variadic) of inputs.
+     * \param n min. number of shapes
+     */
+    const check_shapes& has_at_least(int64_t n) const
```
pfultz2, line 108: Use `std::size_t` like the other overloads do.

src/include/migraphx/op/concat.hpp
```diff
@@ -73,49 +74,96 @@ struct concat
...
+
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        if(inputs.empty())
+        // inputs can contain 1 or more shapes (variadic).  compute_shape_op ensures there must
+        // be at least 1.
+
+        if(not std::all_of(inputs.begin(), inputs.end(), [&](const shape& s) {
+               return (s.ndim() == inputs[0].ndim() and s.type() == inputs[0].type());
+           }))
```
pfultz2, line 85: This should be checked with `check_shapes{inputs, *this}.same_ndim().same_type()`.

src/include/migraphx/operation.hpp
```diff
@@ -141,6 +141,8 @@ auto compute_shape_op(rank<2>, const T& x, const std::vector<shape>& inputs)
     -> decltype(x.normalize_compute_shape(inputs))
 {
     dependent_type<operation, T> y = x;
+    if(inputs.empty())
+        MIGRAPHX_THROW("No input shapes: " + x.name());
```
pfultz2, line 145: Probably say `"At least one input is required for " + x.name()`.

test/ref_ops_test.cpp
pfultz2: Yea it would better to not specify them for standard shapes, so it can make clear to those reading this that it is using standard shape for the test.

src/include/migraphx/op/concat.hpp
```diff
@@ -73,49 +74,96 @@ struct concat
...
         {
-            MIGRAPHX_THROW("CONCAT: Number of input tensors should exceed 0");
+            MIGRAPHX_THROW("CONCAT: input tensors are not all the same rank or type");
         }
 
-        const auto& first_shape_lens = inputs.front().lens();
-        const auto& type             = inputs.front().type();
-        for(std::size_t l = 0; l < first_shape_lens.size(); l++)
+        if(axis >= inputs[0].ndim() or axis < 0)
+            MIGRAPHX_THROW("CONCAT: axis attribute out of range");
```
bpickrel, line 91: I'm removing this check because codecov flags it; there's a test but this case is rejected by TUNE_VECTOR and this line is never reached.  Our unit tests are not so pure in this case, since they assume a dependency of one module on another.

src/include/migraphx/op/concat.hpp
bpickrel: changed

umangyadav (approved):

src/include/migraphx/op/concat.hpp
```diff
@@ -73,49 +74,87 @@ struct concat
         }
         return offsets;
     }
+
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        if(inputs.empty())
+        // inputs can contain 1 or more shapes (variadic).  compute_shape_op ensures there must
+        // be at least 1.
```
CharlieL7, line 81: update comment

CharlieL7 (approved):

src/include/migraphx/op/concat.hpp
bpickrel: That ***is*** an update.  That check is performed in operation.hpp:143 and means that if we get to this point in the code, there's nothing to check for.

PR #1528, opened 2023-01-17
Dynamic shape support for Where op.

bpickrel:
Add dynamic shape support for Where operator.

Where requires all 3 inputs to have the same shape.  We might some day want to let it "adjust" dynamic shapes with min/max ranges that don't match but overlap, or make dynamic shapes match static shapes, but this PR doesn't do that--it requires all dynamic dimensions to be identical.

Onnx parsing has a dependency on dynamic shape broadcasting, which isn't implemented yet.(Update:  dynamic broadcasting works for binary ops but not ternary ops like this one.) The Onnx Where operator allows inputs to have different numbers of dimensions, which then requires broadcasting operators to be executed before Where.  This PR's Onnx parsing skips these broadcasting ops for dynamic shapes.  This should be updated when dynamic shape broadcasting becomes available.  See [Onnx Where spec](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Where)


src/include/migraphx/op/where.hpp
```diff
@@ -42,9 +42,19 @@ struct where
...
-        check_shapes{inputs, *this}.has(3).same_dims();
+        check_shapes{inputs, *this, true}.has(3).same_dims();
         auto s1 = inputs.at(1);
         auto s2 = inputs.at(2);
+        if(s1.dynamic() or s2.dynamic())
+        {
+            if(s1 == s2)
+                return s1;
+            // MIGRAPHX_THROW("BINARY: " + point_function() + ": fixed-dyn shape for inputs");  what
+            // is point_function() for?
```
umangyadav, line 53: `point_op` is used for JIT. I dont' think we need to mention in THROW message here.

src/include/migraphx/op/where.hpp
```diff
@@ -42,9 +42,19 @@ struct where
...
+            MIGRAPHX_THROW("WHERE: fixed-dyn shape for inputs");
```
umangyadav, line 54: Can you change this throw message ? It doesn't seem helpful. 
IMO. It should be more related to  "shape mismatch"

src/include/migraphx/op/where.hpp
```diff
@@ -42,9 +42,19 @@ struct where
 
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(3).same_dims();
+        check_shapes{inputs, *this, true}.has(3).same_dims();
         auto s1 = inputs.at(1);
         auto s2 = inputs.at(2);
+        if(s1.dynamic() or s2.dynamic())
```
umangyadav, line 48: This would mean it handles one static and one dynamic. we don't know which one is which and it always return `s1`.

src/include/migraphx/op/where.hpp
```diff
@@ -42,9 +42,19 @@ struct where
 
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(3).same_dims();
+        check_shapes{inputs, *this, true}.has(3).same_dims();
```
umangyadav, line 45: `same_dims()`. If one input is static and one is dynamic that imply dynamics is fixed because they both need to have `same_dims()`.

umangyadav:
Does this support mix of static and dynamic inputs? It is not clear from the changes.

bpickrel:
> Does this support mix of static and dynamic inputs? It is not clear from the changes.

No, it requires that if either of shape1 and shape2 is dynamic, the other must be exactly the same.  There's no attempt to reconcile ranges that partially overlap, etc.  I can't think of a use case where a user would need to start with two shapes that aren't the same.

For two static shapes, it's different:  the code reconciles packed and broadcasted shapes while using the lens from Shape 1.  I didn't question whether it makes sense not to check that the lens of the two shapes are the same, or whether the 0'th input (the Boolean values) should be checked too.

src/include/migraphx/op/where.hpp
bpickrel: That would get rejected in Lines  50-54 anyhow.  `if(s1 == s2)...`

src/include/migraphx/op/where.hpp
bpickrel: ref. my answers below.  If either shape is dynamic, then the other has to be exactly equal (including being dynamic)

src/include/migraphx/op/where.hpp
bpickrel: I'll reword the message

bpickrel:
Discussion today concluded that all three inputs should have the exact same shape, except that 0'th input has boolean data type and that the existing code accounts for packed and broadcasted static shapes.  When users want to call Where on different shaped tensors, the inputs should be broadcasted to matching shapes before this operator is called.

test/onnx/onnx_test.cpp
```diff
@@ -6693,4 +6693,43 @@ TEST_CASE(where_test)
...
+    //     0}}});
+    // auto ins = mm->add_instruction(migraphx::make_op("argmax", {{"axis", 2}}), l0);
+    // auto ret = mm->add_instruction(migraphx::make_op("squeeze", {{"axes", {2}}}), ins);
+    // mm->add_return({ret});
+
+    // migraphx::onnx_options options;
+    // options.default_dyn_dim_value = {1, 4, 0};
+    // auto prog                     = parse_onnx("argmax_dyn_test.onnx", options);
+
+    // EXPECT(p == prog);
```
umangyadav, line 6732: Do we need this ? You can remove this otherwise.

umangyadav (approved):
Looks good, just clean up `onnx_test.cpp`.

test/onnx/onnx_test.cpp
bpickrel: done

src/onnx/parse_where.cpp
```diff
@@ -40,28 +40,41 @@ struct parse_where : op_parser<parse_where>
...
     {
-        auto lens =
-            compute_broadcasted_lens(args[0]->get_shape().lens(), args[1]->get_shape().lens());
-        lens = compute_broadcasted_lens(lens, args[2]->get_shape().lens());
-        if(args[0]->get_shape().lens() != lens)
+        // This op. doesn't currently support a mix of static and dynamic shapes.
+        // Also, TODO: broadcasting for dynamic shapes isn't implemented at time of writing.
+        //   When it becomes available, add broadcasting steps in the dynamic shape case.
+        // For now for dynamic shapes, just insert the Where op.  All shapes must be the
+        // same for it to succeed.
```
CharlieL7, line 47: We do support broadcasting dynamic shapes

CharlieL7 (commented):
We do support broadcasting dynamic shapes. It should be possible to get it to work with a ternary operator. Would mainly be changes in parse_where.cpp.

src/onnx/parse_where.cpp
```diff
@@ -40,28 +40,42 @@ struct parse_where : op_parser<parse_where>
...
+
+            if(args[2]->get_shape().lens() != lens)
+            {
+                args[2] =
+                    info.add_instruction(make_op("multibroadcast", {{"out_lens", lens}}), args[2]);
+            }
 
-        return info.add_instruction(make_op("where"), args[0], args[1], args[2]);
+            return info.add_instruction(make_op("where"), args[0], args[1], args[2]);
+        }
```
CharlieL7, line 78: We should also catch and throw when it's mixed static and dynamic shapes. It looks like it would throw something about lens being empty in the else branch right now. Can probably just change to std::any_of rather than std::all_of. Add an onnx_test to make sure.

CharlieL7 (changes requested):
parse static-dynamic onnx change

src/onnx/parse_where.cpp
bpickrel: Isn't that an unwarranted level of detail?  As written, it won't allow mixed shape/dimension types and will tell the user that they don't match.  That's all that should be needed.  I'd assume that if we make this op. more elaborate in the future, there will be some logic to fit dynamic shapes to static ones wherever possible, in which case a separate check for that condition becomes obsolete anyway.

bpickrel:
> 

?

src/onnx/parse_where.cpp
CharlieL7: No, as written it will try to make a multibroadcast on mixed shape inputs.

bpickrel: ok, done

src/onnx/parse_where.cpp
bpickrel: comment updated, commit 9b43b0499a

CharlieL7 (approved):

PR #1504, opened 2023-01-04
Update MIGraphx Docker to point to latest working onnxrt branch.

TedThemistokleous:
Will use this as a staging branch between ORT builds and updating CI on our end to catch any integration issues between our EP and MIGraphX

PR #1530, opened 2023-01-18
Migx onnxrt action

TedThemistokleous:
Adding action to setup CI for weekly commits for Friday 9AM

Sorting out last minute issues with cron. (doesn't seem to trigger)

Opening this early in case we may have other concerns/issues with this workflow


causten (approved):

.github/workflows/sync-onnxrt-main.yaml
```diff
@@ -0,0 +1,45 @@
...
+  PR:
+    needs: Add_commit
+    steps:
+      - name: GitHub Action for creating Pull Requests
+        uses: devops-infra/action-pull-request@v0.5.3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          title: Sync Onnxruntime main
+          reviewer: TedThemistokleous, pfultz2, causten
+          assignee: TedThemistokleous, pfultz2, causten
```
causten, line 43: If you need to touch the yaml again, just make you the assignee, and then pfultz2 & causten as the reviewers

TedThemistokleous: Will do. What's odd though is I'm still not seeing this trigger even though I've adjusted things for UTC. Hm..My concern right now is suddenly waking up Friday and we have a bunch of PRs...

TedThemistokleous: Actually just found this:

"Scheduled workflows run on the latest commit on the default or base branch."

https://stackoverflow.com/questions/63436541/github-action-workflow-schedule-not-working-on-non-default-branch

Looks like I should be targeting this with master instead of develop with cron

CharlieL7:
Some questions to make sure I understand what this is doing:
1. Does this workflow updates the onnxrt used to the latest commit every week?
2. Why has the Dockerfile been updated?

TedThemistokleous:
> Some questions to make sure I understand what this is doing:
> 
>     1. Does this workflow updates the onnxrt used to the latest commit every week?
> 
>     2. Why has the Dockerfile been updated?

1. Yes the idea is to continually integrate onnxrt builds into our CI to catch future issues via their scripts and unit tests. This arose from use using previously older commits just to test builds but were missing changes that appeared in other repos like DeepLearningModels.

2. This changeset was built off PR #1504 originally

CharlieL7 (approved):

PR #1531, opened 2023-01-20
hipRTC fixes 

umangyadav:
~This is just a test PR, in which hipRTC is enabled by default through CMakeListsts.  Workarounds for hipRTC are also not gated by any env flags or build flag, which is a TODO.~

Added CMakeFlag for hipRTC. `MIGRAPHX_USE_HIPRTC`.
Added stages in Jenkins for hipRTC.
Fixes for some of the  pending  issues from hipRTC.


src/targets/gpu/kernels/include/migraphx/kernels/gathernd.hpp
```diff
@@ -83,15 +79,15 @@ __device__ void gathernd(const T& data_t, const U& indices_t, const V& output_t,
             int64_t index                   = slice_indices[idx];
             const std::size_t input_dim_idx = batch_dims + idx;
             const auto input_dim            = data_shape_lens[input_dim_idx];
-            assert(index >= -static_cast<int64_t>(input_dim) and
-                   index < static_cast<int64_t>(input_dim));
+            // assert(index >= -static_cast<int64_t>(input_dim) and
```
TedThemistokleous, line 82: remove comment

TedThemistokleous (approved):
Just cleanup comments. Looks good

src/targets/gpu/kernels/include/migraphx/kernels/gathernd.hpp
umangyadav: Done.

src/targets/gpu/CMakeLists.txt
```diff
@@ -33,11 +33,14 @@ if(NOT TARGET MIOpen)
...
 endif()
 
+set(MIGRAPHX_USE_HIPRTC OFF CACHE BOOL "Use hipRTC APIs")
+
 include(Embed)
 file(GLOB KERNEL_FILES ${CONFIGURE_DEPENDS}
     ${CMAKE_CURRENT_SOURCE_DIR}/kernels/include/migraphx/kernels/*.hpp)
 message(STATUS "KERNEL_FILES: ${KERNEL_FILES}")
 add_embed_library(migraphx_kernels ${KERNEL_FILES})
+target_compile_definitions(migraphx_kernels PRIVATE -DMIGRAPHX_USE_HIPRTC)
```
pfultz2, line 43: This flag shouldnt be added to the `migraphx_kernels`.

src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp
```diff
@@ -35,7 +35,12 @@ inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 namespace device {
 
+#ifdef MIGRAHPX_USE_HIPRTC
+#define MIGRAPHX_NO_DPP
+#endif
```
pfultz2, line 40: These changes should be removed. This file is not ever compiled with hiprtc.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -176,8 +177,17 @@ operation compile_hip_code_object(const std::string& content, hip_compile_option
     options.params += " -DMIGRAPHX_NLOCAL=" + std::to_string(options.local);
     options.params += " " + join_strings(compiler_warnings(), " ");
     options.params += " -ftemplate-backtrace-limit=0";
+#ifdef MIGRAPHX_USE_HIPRTC
+    options.params += " -DMIGRAPHX_USE_HIPRTC";
+    // remove following three compilation flags for HIPRTC once fixes from hipRTC are available in
+    // rocm release, add -Werror flag for hipRTC when fixes are available
+    options.params += " -Wno-reserved-identifier";
+    options.params += " -Wno-gnu-line-marker";
+    options.params += " -Wno-old-style-cast";
```
pfultz2, line 186: These flags need to be added to compile_hip.cpp.

src/targets/gpu/compile_hip.cpp
```diff
@@ -176,14 +174,15 @@ compile_hip_src(const std::vector<src_file>& srcs, std::string params, const std
 {
     hiprtc_program prog(srcs);
     auto options = split_string(params, ' ');
+    options.push_back("-DMIGRAPHX_HIPRTC");
```
pfultz2, line 177: Also `-DMIGRAPHX_HAS_DPP=0` needs to be added.

src/targets/gpu/kernels/include/migraphx/kernels/dpp.hpp
```diff
@@ -30,11 +30,11 @@
 
 namespace migraphx {
 
-#ifndef MIGRAPHX_HAS_DPP
-#define MIGRAPHX_HAS_DPP 1
+#ifdef MIGRAPHX_USE_HIPRTC
+#define MIGRAPHX_NO_DPP
```
pfultz2, line 34: This should still use `MIGRAPHX_HAS_DPP` as before and we just pass `-DMIGRAPHX_HAS_DPP=0` when compiling hiprtc.

src/targets/gpu/kernels/include/migraphx/kernels/algorithm.hpp
```diff
@@ -44,6 +44,12 @@ struct greater
     }
 };
 
+template <class T>
+struct multiplies
+{
+    constexpr T operator()(const T& lhs, const T& rhs) { return lhs * rhs; }
+};
```
pfultz2, line 51: There is already a `product` class in ops.hpp that can be used instead.

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
```diff
@@ -31,7 +31,11 @@
 
 namespace migraphx {
 
-#if MIGRAPHX_HAS_DPP
+#ifdef MIGRAPHX_USE_HIPRTC
+#define MIGRAPHX_NO_DPP
+#endif
+
+#ifndef MIGRAPHX_NO_DPP
```
pfultz2, line 38: Should be `#if MIGRAPHX_HAS_DPP`. ANd the above can be removed as its already handled by dpp.hpp.

src/targets/gpu/kernels/include/migraphx/kernels/types.hpp
```diff
@@ -25,11 +25,16 @@
...
 
 #include <migraphx/kernels/hip.hpp>
+#include <initializer_list>
 
 namespace migraphx {
 
-using index_int = std::uint32_t;
-using diff_int  = std::int32_t;
+using index_int = uint32_t;
+using diff_int  = int32_t;
```
pfultz2, line 33: Same here, need a typedef for `int32_t`.

src/targets/gpu/kernels/include/migraphx/kernels/types.hpp
```diff
@@ -25,11 +25,16 @@
...
+using int8_t    = signed char;
+using uint8_t   = unsigned char;
+using int16_t   = signed short int;
+using uint16_t  = unsigned short int;
```
pfultz2, line 37: Does `__hip_uint16_t` work?

test/gpu/jit.cpp
```diff
@@ -56,9 +59,11 @@ int main() {}
 // NOLINTNEXTLINE
 const std::string add_2s_binary = R"__migraphx__(
 #include <hip/hip_runtime.h>
+#include <migraphx/kernels/types.hpp>
+using namespace migraphx;
 
 extern "C" {
-__global__ void add_2(std::int8_t* x, std::int8_t* y) 
+__global__ void add_2(int8_t* x, int8_t* y)
```
pfultz2, line 66: Just used `char` and skip including migraphx header.

test/gpu/jit.cpp
```diff
@@ -35,11 +35,14 @@
...
 #include <migraphx/gpu/compile_hip_code_object.hpp>
 #include <migraphx/gpu/compiler.hpp>
+#include <migraphx_kernels.hpp>
 
 // NOLINTNEXTLINE
 const std::string write_2s = R"__migraphx__(
 #include <hip/hip_runtime.h>
+#include <migraphx/kernels/types.hpp>
 
+using namespace migraphx;
```
pfultz2, line 45: Just used `char` and skip including migraphx header for this test,

test/gpu/jit.cpp
```diff
@@ -158,8 +164,20 @@ migraphx::src_file make_src_file(const std::string& name, const std::string& con
...
-    auto binaries = migraphx::gpu::compile_hip_src(
-        {make_src_file("main.cpp", write_2s)}, "", migraphx::gpu::get_device_name());
+    std::vector<migraphx::src_file> srcs;
+    std::transform(migraphx_kernels().begin(),
+                   migraphx_kernels().end(),
+                   std::back_inserter(srcs),
+                   [](auto&& p) {
+                       auto&& name = p.first;
+                       auto&& c    = p.second;
+                       auto path   = migraphx::fs::path{"migraphx"} / "kernels" / name;
```
pfultz2, line 174: The paths can be removed if you use `char` instead,

src/targets/gpu/compile_hip_code_object.cpp
pfultz2: Also the flags should only be added when `MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS` is set.

src/targets/gpu/CMakeLists.txt
umangyadav: Done

test/gpu/jit.cpp
umangyadav: Done

test/gpu/jit.cpp
umangyadav: Done

test/gpu/jit.cpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/types.hpp
umangyadav: It doesn't just yet. Compiler team has fix merged in and I tested on a docker provided by other team, it works with those fixes, but with released rocm versions, it doesn't work yet.

src/targets/gpu/kernels/include/migraphx/kernels/types.hpp
```diff
@@ -25,11 +25,16 @@
 #define MIGRAPHX_GUARD_AMDMIGRAPHX_KERNELS_TYPES_HPP
 
 #include <migraphx/kernels/hip.hpp>
+#include <initializer_list>
 
 namespace migraphx {
 
-using index_int = std::uint32_t;
-using diff_int  = std::int32_t;
+using index_int = uint32_t;
```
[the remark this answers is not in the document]
umangyadav: `__hip_uint32_t` is undefined in rocm-5.4.2

src/targets/gpu/kernels/include/migraphx/kernels/algorithm.hpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/dpp.hpp
umangyadav: Done.

src/targets/gpu/compile_hip.cpp
umangyadav: Done.

src/targets/gpu/compile_hip_code_object.cpp
umangyadav: Moved those flags inside `compile_hip.cpp`. 

Why do we need `MIGRAPHX_ENABLE_HIPRATC_WORKAROUNDS` ?  How that would be different compared to `MIGRAPHX_USE_HIPRTC` ?  We can use `MIGRAPHX_USE_HIPRTC` to gate workarounds instead  of an extra flag.

src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/types.hpp
umangyadav: Added required typedefs.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -177,7 +178,7 @@ operation compile_hip_code_object(const std::string& content, hip_compile_option
     options.params += " " + join_strings(compiler_warnings(), " ");
     options.params += " -ftemplate-backtrace-limit=0";
     options.params += " -Werror";
-    auto cos = compile_hip_src(srcs, std::move(options.params), get_device_name());
+    auto cos = compile_hip_src(srcs, options.params, get_device_name());
```
pfultz2, line 181: Why is `std::move` being removed?

src/targets/gpu/compile_hip.cpp
```diff
@@ -143,25 +143,28 @@ struct hiprtc_program
...
         MIGRAPHX_HIPRTC(hiprtcGetProgramLogSize(prog.get(), &n));
-        if(n < 2)
+        if(n == 0)
             return {};
-        std::vector<char> buffer(n);
+        std::string buffer(n, '\0');
         MIGRAPHX_HIPRTC(hiprtcGetProgramLog(prog.get(), buffer.data()));
-        assert(buffer.back() == 0);
-        return {buffer.begin(), buffer.end() - 1};
+        return buffer;
```
pfultz2, line 164: We need to make sure the last character is not null. We could do `assert(buffer.back() != 0)` or we could pop it off if it is null: `if(buffer.back() == 0) buffer.pop_back()`.

src/targets/gpu/kernels/include/migraphx/kernels/dpp.hpp
```diff
@@ -68,11 +68,15 @@ __device__ T dpp_mov(T& x)
     input.data = x;
     for(index_int i = 0; i < n; i++)
     {
+#if defined(__HCC__)
+        output.reg[i] = __llvm_amdgcn_move_dpp(input.reg[i], DppCtrl, RowMask, BankMask, BoundCtrl);
+#else
```
pfultz2, line 73: I dont think we need to add support for HCC.

src/targets/gpu/kernels/include/migraphx/kernels/hip.hpp
```diff
@@ -24,11 +24,17 @@
...
 // Workaround macro redefinition issue with clang tidy
 #if defined(__HIP_PLATFORM_HCC__) && defined(MIGRAPHX_USE_CLANG_TIDY)
 #undef __HIP_PLATFORM_HCC__ // NOLINT
 #endif
-
 #include <hip/hip_runtime.h>
+#include <hip/hip_fp16.h>
+#include <hip/math_functions.h>
+#else
+#include <hip/hip_common.h>
```
pfultz2, line 36: We shouldn't include headers from hip when using hiprtc. If we need the functionality from those headers then we should open an issue about this.

For this PR, just include it inside `#ifdef MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS` and then add that define when the env variable is set.

src/targets/gpu/compile_hip.cpp
```diff
@@ -48,9 +47,10 @@ MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_GPU_OPTIMIZE);
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_GPU_DUMP_ASM);
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_GPU_DUMP_SRC);
 
-#if MIGRAPHX_USE_HIPRTC
+#ifdef MIGRAPHX_USE_HIPRTC
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_HIPRTC)
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS);
```
pfultz2, line 53: Missing a semicolon on the above line.

src/targets/gpu/kernels/include/migraphx/kernels/hip.hpp
umangyadav: `hip_common.h` is only so that we get defines for `HIP_PLATFORM_AMD` which is required for `hip_math_constants.h`. 

`hiprtc.h` includes `hip_common.h` [internally](https://github.com/ROCm-Developer-Tools/HIP/blob/436ba9757b965ef967e74f2a9688be5ee0ea863b/include/hip/hiprtc.h#L24).

src/targets/gpu/compile_hip.cpp
umangyadav: internally, it keep  log as `std::string`.  i don't think it will have null character. I can add an assert anyway.

src/targets/gpu/compile_hip.cpp
umangyadav: Fixed.

src/targets/gpu/kernels/include/migraphx/kernels/dpp.hpp
umangyadav: Why is that ? It would fail to compile with HCC then.

src/targets/gpu/compile_hip_code_object.cpp
umangyadav: Put back `std::move`.  It was the remainder from earlier changeset .

src/targets/gpu/kernels/include/migraphx/kernels/hip.hpp
pfultz2: > hip_common.h is only so that we get defines for HIP_PLATFORM_AMD which is required for hip_math_constants.h.

Yes but hiprtc should provide those defines for us without including the file, as those file may not be available during runtime.

> hiprtc.h includes hip_common.h [internally](https://github.com/ROCm-Developer-Tools/HIP/blob/436ba9757b965ef967e74f2a9688be5ee0ea863b/include/hip/hiprtc.h#L24).

Yea but thats on the host side, we need this for device compilation.

src/targets/gpu/kernels/include/migraphx/kernels/hip.hpp
```diff
@@ -24,11 +24,17 @@
...
+#endif
+#include <hip/hip_math_constants.h>
```
pfultz2, line 38: This include needs a `#ifdef MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS` as well.

src/targets/gpu/kernels/include/migraphx/kernels/hip.hpp
umangyadav: > Yes but hiprtc should provide those defines for us without including the file, as those files may not be available during runtime. 

Aren't both HIPRTC and hip-clang are shipping under the same library `libamdhip64.so`.  If so, those headers files would be available.

src/targets/gpu/kernels/include/migraphx/kernels/types.hpp
```diff
@@ -25,11 +25,49 @@
 #define MIGRAPHX_GUARD_AMDMIGRAPHX_KERNELS_TYPES_HPP
 
 #include <migraphx/kernels/hip.hpp>
+#include <initializer_list>
```
pfultz2, line 28: This should not be included when using hiprtc.

src/targets/gpu/kernels/include/migraphx/kernels/dpp.hpp
pfultz2: Yea it would fail to compile, but hcc was deprecated 3 years ago, so there is no reason to try to support it. We dont even test using hcc either so there is no coverage for this code either.

umangyadav: In that case, we also need to clean up HCC from other  places.

src/targets/gpu/kernels/include/migraphx/kernels/hip.hpp
pfultz2: > Aren't both HIPRTC and hip-clang are shipping under the same library libamdhip64.so.

Thats not a header thats a .so. The .so is included in the runtime package, but the headers need to be installed using the dev package. We cant require dev packages when using migraphx's runtime package.

umangyadav: okay, 
I see also see comment in hipRTC header that, eventually it will be shipped as separate library. 

I only need hip_common.h and hip_math_constants.h for `PI_2`.  we can also just define that our side for better workaround.

src/targets/gpu/kernels/include/migraphx/kernels/hip.hpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/types.hpp
umangyadav: Removed.

src/targets/gpu/kernels/include/migraphx/kernels/dpp.hpp
umangyadav: Removed HCC for now, but later need to clean up other references from entire codebase.

pfultz2 (approved):

Issue #1344, opened 2022-08-18
an error about parse_if


qianqing13579:
I want to use migraphx-driver to test a model but an error occurs in migraphx::parse_onnx. The migraphx is from commit f2531606c266dd548fee59f242849615cab7f613
the command I used:/opt/rocm/bin/migraphx-driver perf TextLinePred_0722_v1.onnx
the outputs: src/onnx/parse_if.cpp:72: parse: PARSE_IF: then and else sub_grahps must have same output shapes!
the model link: https://drive.google.com/file/d/18p1AAdjCA3kr1zd7lkRfmfcFOC7layrW/view?usp=sharing

CharlieL7:
This looks related to a similar issue we've been seeing for ResNext50. There's a PR in progress for the If operator that needs more work. @TedThemistokleous

TedThemistokleous:
Potentially. This looks like this was before we added the fix Charlie and I did with our input constants as well. Let me look at the model on netron and see why this is failing.

TedThemistokleous:
Confirmed its another issue with IF being parsed in but shapes not interpreting outputs based on stride

```
module: "If_641_else"
If_641_else:@0 = identity(main:@637) -> float_type, {1, 3, 1, 1}, {3, 1, 1, 1}
If_641_else:@1 = @return(If_641_else:@0)

module: "If_641_if"
If_641_if:@0 = squeeze[axes={2}](main:@637) -> float_type, {1, 3, 1}, {3, 1, 1}
If_641_if:@1 = @return(If_641_if:@0)


terminate called after throwing an instance of 'migraphx::version_1::exception'
  what():  /code/AMDMIGraphX/src/onnx/parse_if.cpp:74: parse: PARSE_IF: If_641 then and else sub_grahps must have same output shapes!
Aborted (core dumped)
root@c285b5e687d8:/code/AMDMIGraphX/build# 
```
The branch you tried this on @qianqing13579 checks strict shapes. It's an issue we're currently working on with our shapes IF operator as the output case you're hitting requires dynamic output. We've seen this similar issues in a few other model's we've tested (#1283).

qianqing13579:
[...]

PR #1533, opened 2023-01-24
Parse if inline constant args

TedThemistokleous:
This allows us to inline the IF operator when we run into a IF that can be evaluated at compile time, thus avoiding us injecting IF and just inserting the instructions directly.

These changes were spawned from the work of #1325 and related to #1283 

Onnx tests have been updated for the original IF operator then/else cases to include a non constant parameter as well as an "inlined" version which reduces the operations to the correct branch of instructions without inserting a sub module.


test/onnx/onnx_test.cpp
```diff
@@ -2665,6 +2667,34 @@ TEST_CASE(if_else_test)
...
+    auto re = mm->add_instruction(migraphx::make_op("mul"), y, l2);
+    mm->add_return({re});
+
+    std::ifstream ifs("if_else_test_inlined.onnx", std::ios::binary);
+    ifs.seekg(0, std::ios::end);
+    auto length = ifs.tellg();
+    ifs.seekg(0, std::ios::beg);
+    std::vector<char> onnx_buffer(length);
+    ifs.read(onnx_buffer.data(), length);
+    ifs.close();
```
umangyadav, line 2692: why are we doing it this way instead of just calling `parse_onnx()` ?  I think we can just use `parse_onnx()`.

test/onnx/onnx_test.cpp
```diff
@@ -2665,6 +2667,34 @@ TEST_CASE(if_else_test)
...
+TEST_CASE(if_else_test_inlined)
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape sc{migraphx::shape::bool_type, {1}};
+    mm->add_literal(migraphx::literal(sc, {0}));
+    migraphx::shape s{migraphx::shape::float_type, {2, 3}};
+    std::vector<float> ones(s.elements(), 1.0f);
+    mm->add_literal(s, ones);
+    std::vector<float> rand = {0.811412, -0.949771, -0.169276, 0.36552, -0.14801, 2.07061};
```
umangyadav, line 2679: how did you get these values ? did you print before `numpy` values while generating onnx ?

umangyadav, line 2679: I like how program is structured in `if_else_test` or `if_literal_test`, separating literal values'  declaration and construction of MIGraphX program. If you can re-arrange it would be good.

test/onnx/onnx_test.cpp
TedThemistokleous: Sure I can change this. Grabbed this from a previous changeset

test/onnx/onnx_test.cpp
TedThemistokleous: Yeah basically

TedThemistokleous: Sure I'll clean this up

TedThemistokleous: Done

test/onnx/onnx_test.cpp
TedThemistokleous: Done

src/onnx/parse_if.cpp
```diff
@@ -33,6 +33,28 @@ namespace migraphx {
...
 namespace onnx {
 
+static std::vector<instruction_ref> cleanup_arguments(module* mod)
+{
+    auto num_outputs = mod->get_output_shapes().size();
+
+    if(num_outputs > 1)
+    { // TODO: Add support for make_tuple operator. This isn't trivial
+        //      Function that does our output aliasing has
+        MIGRAPHX_THROW("PARSE_IF: Multi-output if not supported for IF operator const folding");
```
umangyadav, line 43: Instead of throwing, would it be possible to fallback on non-const fold version ?

pfultz2: I think non-const version doesnt support multi-output either. So the error could be moved.

src/onnx/parse_if.cpp
```diff
@@ -33,6 +33,28 @@ namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace onnx {
 
+static std::vector<instruction_ref> cleanup_arguments(module* mod)
```
pfultz2, line 36: A better name would be `remove_return`.

TedThemistokleous: okay i'll update

src/onnx/parse_if.cpp
TedThemistokleous: I did add that in to the previous PR #1325 but not inlining. Do we want to add that in then?

src/onnx/parse_if.cpp
TedThemistokleous: Done

src/onnx/parse_if.cpp
TedThemistokleous: wouldn't that already be handled by mapping inputs to the output vector though?

TedThemistokleous: Removed check and added tests for multi output for inline and non inline cases.

src/onnx/onnx_parser.cpp
```diff
@@ -375,8 +375,11 @@ void onnx_parser::parse_graph(module* mod, const onnx::GraphProto& graph)
     // add the return instuction
     mod->add_return(output_ins);
 
-    // remove instructions added in this mod
-    erase_if(instructions, [&](auto&& p) { return mod->has_instruction(p.second); });
+    if(not inlining)
```
umangyadav, line 378: we can also not just add `add_return` here if we are inlining a subgraph.

test/onnx/onnx_test.cpp
```diff
@@ -2734,6 +2753,70 @@ TEST_CASE(if_param_test)
...
+
+    auto* else_mod = p.create_module("If_5_else");
+    auto re        = else_mod->add_instruction(migraphx::make_op("mul"), y, l2);
+    auto re2       = else_mod->add_instruction(migraphx::make_op("sub"), y, l2);
+    else_mod->add_return({re, re2});
+
+    auto ret = mm->add_instruction(migraphx::make_op("if"), {cond}, {then_mod, else_mod});
+    auto r   = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), ret);
+    mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 1}}), ret);
+    mm->add_return({r});
```
umangyadav, line 2814: this should also return other tuple element

test/onnx/verify_onnx.cpp
```diff
@@ -590,6 +590,28 @@ TEST_CASE(if_else_test)
...
+
+    auto result = p.eval(pp).back();
+    std::vector<float> result_vector;
+    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
+
+    std::vector<float> gold = {0.0866565, -0.371067, 0.017719, 0.0250614, 0.0612539, -0.744683};
+    EXPECT(migraphx::verify_range(result_vector, gold));
+}
+
+TEST_CASE(if_else_test_inlined)
```
umangyadav, line 609: Do we support returning multiple args from MIGraphX program during evaluation ? 
I have vague memory that we don't. 
If we do, we should also add atleast add one of the multi-outputs onnx tests here.

pfultz2: > Do we support returning multiple args from MIGraphX program during evaluation ?

Yes we do. For example, bert modles have 2 outputs(which we return).

What we dont support is inlining a subgraph(from `if`) with multiple outputs because we need to have a `make_tuple` operator which requires us to refactor `get_output_alias`.

pfultz2: > What we dont support is inlining a subgraph(from if) with multiple outputs because we need to have a make_tuple operator which requires us to refactor get_output_alias.

Ah wait, we can support this by mapping the return to the vector we return from `parse_if`.

test/onnx/gen_onnx.py
```diff
@@ -2513,6 +2565,146 @@ def if_else_test():
...
+
+    cond_tensor = onnx.helper.make_tensor_value_info("cond",
+                                                     onnx.TensorProto.BOOL,
+                                                     [1])
+
+    res = onnx.helper.make_tensor_value_info('res', TensorProto.FLOAT, [])
+
+    node = onnx.helper.make_node('If',
+                                 inputs=['cond'],
+                                 outputs=['res'],
```
pfultz2, line 2701: We should do `outputs=['res1', 'res2']` so we test returning both outputs from the `if` statement.

test/onnx/onnx_test.cpp
pfultz2: Its technically correct with the way the onnx was written. Only the first element is returned, but the test should be updated to return both so we are actually testing multiple outputs.

test/onnx/gen_onnx.py
```diff
@@ -2513,6 +2565,146 @@ def if_else_test():
...
+                                 then_branch=then_body,
+                                 else_branch=else_body)
+
+    return ([node], [x, y, cond_tensor], [res], [xt_tensor, yt_tensor])
```
pfultz2, line 2705: We should return both outputs: `return ([node], [x, y, cond_tensor], [res1, res2], [xt_tensor, yt_tensor])`

src/onnx/onnx_parser.cpp
pfultz2: But to do this, we need to have `parse_graph` return the vector of returned instructions.

test/onnx/gen_onnx.py
TedThemistokleous: All good. Done

test/onnx/gen_onnx.py
TedThemistokleous: Done

test/onnx/verify_onnx.cpp
TedThemistokleous: Yeah I have it running in an test case right now. Going to upload this.

test/onnx/onnx_test.cpp
TedThemistokleous: Just updated

src/onnx/onnx_parser.cpp
TedThemistokleous: Done. Just gonna let this run till overnight. Used (void) to signal ignored returns in our code since we want to support earlier versions of modern C++ and only since C+17  [[maybe_unused]] was added.

test/onnx/verify_onnx.cpp
umangyadav: Can we add one multi output test in `onnx_test` ?

TedThemistokleous: Sure. Adding that now

src/onnx/parse_if.cpp
```diff
@@ -51,17 +51,35 @@ struct parse_if : op_parser<parse_if>
...
+
         std::string then_name = info.name + "_if";
         module_ref then_mdl   = parser.prog.create_module(then_name);
 
         std::string else_name = info.name + "_else";
         module_ref else_mdl   = parser.prog.create_module(else_name);
 
         // parse the then sub_graph
-        parser.parse_graph(then_mdl, then_graph);
+        (void)parser.parse_graph(then_mdl, then_graph);
```
pfultz2, line 79: Is the `(void)` to fix a warning?

TedThemistokleous: Yes and to signal that we know there's a return but we're choosing not to use it in this case. Since I believe we're not using C++17, we wouldn't have the [[maybe_unused]] attribute to indicate similar.

umangyadav (approved):

pfultz2 (approved):

TedThemistokleous:
confirmed to also fix #1344

PR #1535, opened 2023-01-24
Fix nonmaxsuppression max boxes

TedThemistokleous:
Add sanity tests + bound max_output_boxes_per_class to the scores dimension


src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -308,6 +308,13 @@ struct nonmaxsuppression
         {
             return result;
         }
+
+        //#bound the max amount of boxes allowed to the max of boxes
+        if(max_output_boxes_per_class > args[1].get_shape().lens()[2])
+        {
+            max_output_boxes_per_class = args[1].get_shape().lens()[2];
+        }
```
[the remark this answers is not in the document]
TedThemistokleous: yea they are. Came up in resnext and thought it was odd since it renders part of the loop useless

image

test/onnx/verify_onnx.cpp
```diff
@@ -1080,6 +1080,72 @@ TEST_CASE(nonzero_test)
...
+    pp["max_output_boxes_per_class"] = migraphx::argument(mob_shape, mob_data.data());
+    pp["boxes"]                      = migraphx::argument(boxes_shape, boxes_data.data());
+    pp["scores"]                     = migraphx::argument(score_shape, score_data.data());
+    pp["score_threshold"]            = migraphx::argument(score_thresh_shape, score_thresh.data());
+    pp["iou_threshold"]              = migraphx::argument(iou_thresh_shape, iou_thresh.data());
+
+    p.eval(pp).back();
+
+    EXPECT(true);
+}
```
CharlieL7, line 1147: I think this would be better tested in ref_ops_test.cpp. Since you're checking the nonmaxsuppression.hpp compute function

test/onnx/verify_onnx.cpp
```diff
@@ -1080,6 +1080,72 @@ TEST_CASE(nonzero_test)
...
```
CharlieL7, line 1114: I don't understand what this is testing

TedThemistokleous: Just that it runs. We don't have a test which runs the "normal expected" state of non max suppression

test/onnx/verify_onnx.cpp
TedThemistokleous: Sure I can move that

src/include/migraphx/op/nonmaxsuppression.hpp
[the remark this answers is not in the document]
umangyadav: > it renders part of the loop useless
It doesn't. 

` while(not boxes_heap.empty() &&
                  selected_boxes_inside_class.size() < max_output_boxes_per_class)`. 

AND condition will  make sure we don't processes more than what we have.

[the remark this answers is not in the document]
TedThemistokleous: If the max is much larger than the amount of boxes that exist in the batch, select_boxes_inside_class.size() will never trigger.  That's why I added the change to bound this, otherwise in that case we'd need to process every value in the heap until it's empty() regardless of if we select every box.

test/ref_ops_test.cpp
```diff
@@ -4612,6 +4612,72 @@ TEST_CASE(neg_dyn_test)
...
+    pp["boxes"]                      = migraphx::argument(boxes_shape, boxes_data.data());
+    pp["scores"]                     = migraphx::argument(score_shape, score_data.data());
+    pp["score_threshold"]            = migraphx::argument(score_thresh_shape, score_thresh.data());
+    pp["iou_threshold"]              = migraphx::argument(iou_thresh_shape, iou_thresh.data());
+
+    p.eval(pp).back();
+
+    EXPECT(true);
+}
+
```
CharlieL7, line 4680: We don't normally parse an onnx file in ref_ops_test.cpp. You can copy and paste changing the `max_out_l` from TEST_CASE(nms_test) to test your change. I don't think the sanity test is necessary since nms_test should be fine. I'm pretty sure verify_onnx is only used for ONNX operators that we essentially map to other MIGX operators rather than have it's own separate MIGX op. For example, the BatchNormalization or Celu ONNX operators.

TedThemistokleous:
Closing this out. I don't think the change actually fixes since shape_for_each doesn't share vars between each call. I wrote this late without realizing that. To @umangyadav comment, all this is doing is the same as if boxes_heap.pop() is empty since we at most put one next_top_score into  selected_boxes_inside_class.

PR #1540, opened 2023-01-27
use .onnxrt-commit in docker image

causten:
Fix to use the contents of the .onnxrt-commit file.  Docker was not assigning the contents to the dockerfile argument.  Changed the technique to be used as part of the git clone download.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -99,15 +99,12 @@ COPY ./test/onnx/.onnxrt-commit /
 
 ARG ONNXRUNTIME_REPO=https://github.com/Microsoft/onnxruntime
 ARG ONNXRUNTIME_BRANCH=main
-
-# Let us know which commit where're using for CI
-RUN echo "Onnxruntime Commit:" $(cat /.onnxrt-commit)
+ARG ONNXRUNTIME_COMMIT
 
 RUN git clone --single-branch --branch ${ONNXRUNTIME_BRANCH} --recursive ${ONNXRUNTIME_REPO} onnxruntime && \
     cd onnxruntime && \
-    git checkout $(cat /.onnxrt-commit) && \
-    /bin/sh dockerfiles/scripts/install_common_deps.sh
-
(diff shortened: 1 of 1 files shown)


TedThemistokleous (approved):

umangyadav (approved):

Dockerfile
```diff
@@ -95,18 +95,17 @@ RUN cget -p $PREFIX install facebook/zstd@v1.4.5 -X subdir -DCMAKE_DIR=build/cma
 RUN cget -p $PREFIX install ccache@v4.1 -DENABLE_TESTING=OFF
 RUN cget -p /opt/cmake install kitware/cmake@v3.24.3
 
-RUN export ONNXRT_COMMIT=$(cat test/onnx/.onnxrt-commit)
+COPY ./test/onnx/.onnxrt-commit /
 
 ARG ONNXRUNTIME_REPO=https://github.com/Microsoft/onnxruntime
 ARG ONNXRUNTIME_BRANCH=main
-ARG ONNXRUNTIME_COMMIT=$ONNXRT_COMMIT
```
pfultz2, line 102: The `ARG` should not be removed. This is so we can build a dockerfile with a different onnxrt commit without modifying the dockerfile. The same feedback was mentioned here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1504#discussion_r1084285301

pfultz2 (changes requested):

Dockerfile
causten: The argument doesn't do anything anymore. That's because of line 108.  It uses "git checkout $(cat /.onnxrt-commit)" not a variable.  And that's because Docker doesn't assign values from RUN commands.  The only way your request would work would be to assign ONNXRUNTIME_COMMIT when you issue the Docker build command in the Jenkins file.

causten: docker build -t test:yesarg . --build-arg ONNXRUNTIME_COMMIT=3b6d551c3563738d6396e9e44715ecc3ffc4678b
docker build -t test:noarg .

This worked.  I tested with different commits both in the .onnx-commit file as well as the build arg CLI.

pfultz2 (approved):

PR #1546, opened 2023-01-30
Remove HCC 

umangyadav:
HCC has been deprecated. MIGraphX is not using HCC either. Removing HCC related logic from codebase


src/include/migraphx/compile_src.hpp
```diff
@@ -43,12 +43,11 @@ struct src_file
 
 struct src_compiler
 {
-    std::string compiler                      = "c++";
-    std::string flags                         = "";
-    std::string output                        = "";
-    std::string launcher                      = "";
-    std::string out_ext                       = ".o";
-    std::function<fs::path(fs::path)> process = nullptr;
+    std::string compiler = "c++";
```
pfultz2, line 46: `process` should not be removed., its not hcc-specific. It could be used to invoke a linker.

src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp
```diff
@@ -184,17 +170,8 @@ __device__ auto block_reduce(index idx, Op op, T init, ForStride fs, F f)
...
-template <index_int N, class Op, class T, class F>
-__device__ auto block_reduce(index idx, Op op, T init, index_int n, F f)
-{
-    auto midx = make_multi_index(idx.local, idx.nlocal());
-    // Workaround hcc, create a local array
-    auto fs = midx.id;
-    fs[0]   = n;
-    return block_reduce<N>(
-        idx, op, init, midx.for_stride(fs), [&](auto mi) __device__ { return f(mi[0]); });
-}
```
pfultz2, line 197: Multi-index version is no hcc-specific. This should be kept.

umangyadav: It mentioned `hcc` workround in the comment.

src/include/migraphx/compile_src.hpp
umangyadav: It wasn't used anywhere that's why i removed it.

umangyadav: Reverted change.

src/targets/gpu/device/include/migraphx/gpu/device/reduce.hpp
umangyadav: Reverted change.

install_deps.cmake
```diff
@@ -63,21 +65,24 @@ if(NOT CMakeGet_FOUND)
...
         message(FATAL_ERROR "Download for install.cmake failed: ${RESULT_MESSAGE}")
     endif()
+
     execute_process(COMMAND ${CMAKE_COMMAND} -P ${FILENAME} ${PREFIX})
     file(REMOVE ${FILENAME})
     find_package(CMakeGet REQUIRED PATHS ${PREFIX})
 endif()
 
-# Set compiler to hcc if not set
+# Set compiler to clang++ if not set
```
kahmed10, line 78: with this change, it seems like we can update the README to remove `CXX=/opt/rocm/llvm/bin/clang++` from the cmake command?

umangyadav: I don't think so. `install_deps.cmake` is another way to install dependencies. 

We are not using or don't have documented way of using `install_deps.cmake` to install dependencies. 

If users haven't installed dependencies using this `install_deps.cmake` then, they won't have `CXX` environment variable.  

For example, MIOpen uses this : https://github.com/ROCmSoftwarePlatform/MIOpen#installing-the-dependencies

TedThemistokleous (approved):

PR #1549, opened 2023-01-31
Optimize add convolution

pfultz2:
This will replace `conv(x+a, w)` with `conv(x, w) + conv(a, w)` where `a` is a constant so `conv(a, w)` can be replaced with a constant. 

This still needs unit tests.


test/verify/test_add_conv.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+
+struct test_add_conv : verify_program<test_add_conv>
```
kahmed10, line 30:
```suggestion
struct test_add_conv_constant : verify_program<test_add_conv_constant>
```
Maybe to make it more clear that it's not just add + conv, but the simplification is expected because of the constants

test/verify/test_add_conv.cpp
kahmed10, line 1: Looks like the CPU implementation is failing.

umangyadav (changes requested):
Changes look good to me, 
but i don't get how is this beneficial.  

`conv(x+a, w)` is doing one add and one conv.
`conv(x, w) + constant` is also doing one add and one convolution. 

Can we put some explanation on `find_conv_add` matcher on why it  makes sense ?

src/simplify_algebra.cpp
```diff
@@ -268,6 +269,32 @@ struct find_dot_add
     }
 };
 
+struct find_conv_add
+{
+    auto matcher() const
+    {
+        auto add = match::name("add")(
+            match::either_arg(0, 1)(match::any().bind("x"),
+                                    match::any_of(match::is_constant()).bind("a")),
```
umangyadav, line 278: constant `a` needs to be of same shape as `x`  or shape should be such that atleast `conv(x, w)` shouldn't fail.

umangyadav: Nevermind, pointwise add would have both `a` and `x` guaranteed to be of the same type and dims.

pfultz2:
> conv(x+a, w) is doing one add and one conv.
> conv(x, w) + constant is also doing one add and one convolution.

Usually you have convolution followed by a bias add: `conv(x+a, w) + b` which then becomes `conv(x, w) + conv(a, w) + b`. The `conv(a, w) + b` gets const folded and you have `conv(x, w) + c`. 

So when you have two adds and a convolution, it becomes one add and a convolution.

Also, even if you didnt have convolution followed by an add, `conv(x, w) + c` can be fused in one kernel whereas `conv(x+a, w)` cannot.

We do this same transformation for gemms.

umangyadav:
> > conv(x+a, w) is doing one add and one conv.
> > conv(x, w) + constant is also doing one add and one convolution.
> 
> Usually you have convolution followed by a bias add: `conv(x+a, w) + b` which then becomes `conv(x, w) + conv(a, w) + b`. The `conv(a, w) + b` gets const folded and you have `conv(x, w) + c`.
> 
> So when you have two adds and a convolution, it becomes one add and a convolution.
> 
> Also, even if you didnt have convolution followed by an add, `conv(x, w) + c` can be fused in one kernel whereas `conv(x+a, w)` cannot.
> 
> We do this same transformation for gemms.

Can we put such explanation over the matcher `find_conv_add` for quick reference?

kahmed10:
@pfultz2 to add description on optimization then will approve

kahmed10 (approved):

src/simplify_algebra.cpp
```diff
@@ -204,7 +204,12 @@ struct find_mul_slice_conv
     }
 };
 
-// a * (x + b) => a * x + a * b
+// ******************************
```
umangyadav, line 207: we need similar comment on `find_conv_add`.

pfultz2: This applies to all the `find_*_add` classes that follow.

umangyadav (approved):

PR #1557, opened 2023-02-10
[quick fix] Change neg unit ops to use insert instruction

kahmed10:
There seems to be a bigger issue present when neg unit ops in `simplify_algebra` matches a param as the input instruction. For example, after replacing a `mul` instruction with `@param` and -1 with a `neg(@param)`, dead code elimination results in an empty `neg` instruction.

### DIFF
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -982,6 +982,9 @@ struct find_neg_unit_ops
         auto ins  = r.result;
         auto c_in = r.instructions["x"];
 
+        if(c_in->name() == "@param")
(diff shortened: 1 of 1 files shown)


umangyadav:
> a neg(@param), dead code elimination results in an empty neg instruction. 

Is it possible to fix it in the DCE then ?

kahmed10:
> > a neg(@param), dead code elimination results in an empty neg instruction.
> 
> Is it possible to fix it in the DCE then ?

I will try, and looks like I will need another fix anyways since the original tests are failing on CI.

PR #1561, opened 2023-02-10
Set device to current hip device when loading programs

shivadbhavsar:
Currently, we default to device 0 when loading programs. Updating this to use hipGetDevice to set the device for the loaded program.


umangyadav (approved):

TedThemistokleous (approved):

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -273,7 +273,11 @@ struct context
         auto v_streams        = v.at("streams");
         std::size_t n_streams = v_streams.without_key().to<std::size_t>();
 
-        this->current_device = std::make_shared<hip_device>(0, n_streams);
+        int device;
+        auto status = hipGetDevice(&device);
+        if(status != hipSuccess)
+            MIGRAPHX_THROW("Failed to get hip device");
```
umangyadav, line 279: You can actually include `gpu/device_name.hpp` and call `get_device_id()`. which does same thing.

umangyadav (approved):

PR #1564, opened 2023-02-13
Requirements.txt obselete, now split into different requirements files

TedThemistokleous:
Use requirements-dev.txt instead of the other instances

### DIFF
--- a/tools/build_and_test_onnxrt.sh
+++ b/tools/build_and_test_onnxrt.sh
@@ -22,7 +22,7 @@
 # THE SOFTWARE.
 #####################################################################################
 cd /onnxruntime
-pip3 install -r requirements.txt
+pip3 install -r requirements-dev.txt
 # Add newer cmake to the path
 export PATH="/opt/cmake/bin:$PATH"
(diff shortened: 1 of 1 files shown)


umangyadav (approved):

tools/build_and_test_onnxrt.sh
```diff
@@ -22,7 +22,7 @@
 # THE SOFTWARE.
 #####################################################################################
 cd /onnxruntime
-pip3 install -r requirements.txt
+pip3 install -r requirements-dev.txt
```
causten, line 25: I don't think this file exists in /onnxruntime/requirements-dev.txt.  The Dockerfile copies it to /requirements-dev.txt.  But isn't this suppose to be the ORT requirements?  If no, and it is suppose to be the migx requirements then a /requirements-dev.txt would work.

causten (changes requested):

tools/build_and_test_onnxrt.sh
causten: oh nevermind I see the ONNX based requirements-dev.txt file.

causten (approved):

PR #1573, opened 2023-02-15
Add Python 3.10 modules

causten:


Dockerfile
```diff
@@ -55,7 +55,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     rm -rf /var/lib/apt/lists/*
 
 # add this for roctracer dependancies
-RUN pip3 install CppHeaderParser packaging==22.0
+RUN pip3 install CppHeaderParser packaging==23.0
```
umangyadav, line 58: I think we can move this inside install_prereqs.sh

PR #1580, opened 2023-02-24
Split single dynamic dimension compiler pass

CharlieL7:
### INFO
* Adds a new GPU compiler pass `split_single_dyn_dim` that handles when one input parameter has a single non-fixed `dynamic_dimension`.
  * commonly occurs for dynamic batch or BERT sequence length
* Splits the dynamic shape into several submodules will static input parameters to handle all of the cases in the `dynamic_dimension` range. 
  * Essentially does what I manually did for the `select_module` verify tests
* Adds a compile option `split_single_dyn_dim` that toggles the pass on/off. Defaults to false.
* Updates `verify_program.hpp` and `run_verify.cpp` to allow for the tests to change the `compile_options`

### QUESTIONS / ISSUES
* `offload_copy` will not work with this pass as the `load`, `hip::copy_to_gpu` and `hip::copy_from_gpu` instructions do not work with dynamic shape
* I was not able to link `rank.hpp` properly for `verify_program.hpp` to use the rank template. How do I change the Cmake to get this done?
* It looks like we require `get_tuple_elem` instructions to dereference tuple types. Have I done this correctly for multiple outputs?


src/split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+bool has_one_dyn_dim(const std::unordered_map<std::string, shape>& param_shapes,
+                     std::string& dyn_param_str,
+                     int& dyn_index,
+                     int& min_dim,
+                     int& max_dim)
+{
+    // True if parameters contain exactly one dynamic shape with exactly one non-fixed
+    // dynamic_dimension.
+    if(std::none_of(
```
kahmed10, line 42: I don't think this is needed because the for loop would go through `param_shapes` and not process any further if there are no dynamic dimensions. Maybe we can just return `num_dynamic == 1` at the end.

src/include/migraphx/compile_options.hpp
```diff
@@ -32,9 +32,14 @@ inline namespace MIGRAPHX_INLINE_NS {
 
 struct compile_options
 {
-    bool offload_copy    = false;
+    /// Have MIGX allocate memory for the parameters and copy to and from the GPU
```
umangyadav, line 35: This needs to be re-worded. 

It allocates memory for Parameters (need to check if it is pinned or not) and injects instructions that copies parameters and final output to and from GPU back to the Host.

src/include/migraphx/module.hpp
```diff
@@ -54,10 +54,6 @@ using ins_dep_map   = std::unordered_map<instruction_ref, std::unordered_set<ins
  */
 struct module
 {
-    // used by replace_allocate pass
```
umangyadav, line 57: Need to rebase and resolve conflicts.

src/split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,137 @@
...
+    if(has_one_dyn_dim(param_shapes, dyn_param_name, dyn_index, min_dim, max_dim))
+    {
+        const auto& dyn_param = mm->get_parameter(dyn_param_name);
+        auto dyn_param_shape  = mm->get_parameter_shape(dyn_param_name);
+        std::vector<module_ref> submodules;
+        // create submodules for each dimension size
+        for(int dim_size = min_dim; dim_size <= max_dim; ++dim_size)
+        {
+            auto* submod = mpm.create_module("dim_" + std::to_string(dim_size));
+            // instruction map for new static submodule parameters
```
umangyadav, line 101:
```suggestion
            // instruction map for new static shaped submodule parameters
```

src/split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <migraphx/make_op.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+bool has_one_dyn_dim(const std::unordered_map<std::string, shape>& param_shapes,
+                     std::string& dyn_param_str,
+                     size_t& dyn_index,
+                     size_t& min_dim,
+                     size_t& max_dim)
```
pfultz2, line 38: Dont use output parameters. Instead make this a struct and return `optional<...>` of that struct.

src/split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+    // True if parameters contain exactly one dynamic shape with exactly one non-fixed
+    // dynamic_dimension.
+    size_t num_dynamic = 0;
+    for(const auto& ps : param_shapes)
```
pfultz2, line 43: Use `find_if` for this loop.

src/split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,137 @@
...
+        if(ps.second.dynamic())
+        {
+            num_dynamic += 1;
+            if(num_dynamic > 1)
+            {
+                return false;
+            }
+            size_t num_nf = 0;
+            auto dds      = ps.second.dyn_dims();
+            for(size_t i = 0; i < dds.size(); ++i)
```
pfultz2, line 54: Use `find_if` for this loop.

src/split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,137 @@
...
+        // create submodules for each dimension size
+        for(size_t dim_size = min_dim; dim_size <= max_dim; ++dim_size)
+        {
+            auto* submod = mpm.create_module("dim_" + std::to_string(dim_size));
+            // instruction map for new static shaped submodule parameters
+            std::unordered_map<instruction_ref, instruction_ref> map_ins;
+            // create static shape using dim_size
+            auto static_lens          = dyn_param_shape.max_lens();
+            static_lens.at(dyn_index) = dim_size;
+            auto static_param         = submod->add_parameter(
```
pfultz2, line 106: Just do `map_ins[dyn_param] = submod->add_parameter(...)`.

src/split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,137 @@
...
+    size_t dyn_index;
+    size_t min_dim;
+    size_t max_dim;
+    if(has_one_dyn_dim(param_shapes, dyn_param_name, dyn_index, min_dim, max_dim))
+    {
+        const auto& dyn_param = mm->get_parameter(dyn_param_name);
+        auto dyn_param_shape  = mm->get_parameter_shape(dyn_param_name);
+        std::vector<module_ref> submodules;
+        // create submodules for each dimension size
+        for(size_t dim_size = min_dim; dim_size <= max_dim; ++dim_size)
```
pfultz2, line 98: I would prefer `for(size_t dim_size:range(min_dim, max_dim+1)).

src/include/migraphx/compile_options.hpp
```diff
@@ -32,9 +32,17 @@ inline namespace MIGRAPHX_INLINE_NS {
 
 struct compile_options
 {
-    bool offload_copy    = false;
+    /**
+     * Have MIGX allocate memory for parameters and add instructions
+     * to copy parameters and output to/from the GPU and host
```
pfultz2, line 37: This should say `to copy parameters and output to/from an offload device like a GPU`. Since this is not a GPU-specific flag.

test/verify/verify_program.hpp
```diff
@@ -24,20 +24,47 @@
...
+auto get_co(rank<1>, const T& t) -> decltype(t.get_compile_options())
+{
+    return t.get_compile_options();
+}
+
+template <class T>
+auto get_co(rank<0>, const T&) -> decltype(migraphx::compile_options{})
+{
+    return migraphx::compile_options{};
+}
```
pfultz2, line 65: Just add `get_compile_options` to the `verify_program` base class instead of this. Then you wont need `rank<>` class.

test/verify/test_split_single_dyn_dim.cpp
```diff
@@ -0,0 +1,54 @@
...
+        auto broadcast_lit =
+            mm->add_instruction(migraphx::make_op("multibroadcast"), literal_ins, input);
+        auto add_ins = mm->add_instruction(migraphx::make_op("add"), input, broadcast_lit);
+        mm->add_return({add_ins});
+        return p;
+    }
+
+    migraphx::compile_options get_compile_options() const
+    {
+        return migraphx::compile_options{false, true, false, true};
```
pfultz2, line 52: Use the name:

```cpp
migraphx::compile_options options;
options.split_single_dyn_dim = true;
return options;
```

src/include/migraphx/compile_options.hpp
```diff
@@ -32,9 +32,17 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+     */
+    bool offload_copy = false;
+
     bool fast_math       = true;
     bool exhaustive_tune = false;
+
+    /// Use the split_single_dyn_dim pass
+    bool split_single_dyn_dim = false;
```
pfultz2, line 45: We can use this flag for now for testing, but ideally we should automatically split the dynamic dimension.

src/split_single_dyn_dim.cpp
CharlieL7: Same issue as above. I think I can use `accumulate` instead?

pfultz2: No need to use `std::accumulate`, just do a second `find_if` call:

```cpp
auto not_fixed = [](const auto& dd) {
    return not dd.is_fixed();
};
auto dd_it = std::find_if(dds.begin(), dds.end(), not_fixed);
// Maybe this should be an assert?
if (dd_it == dds.end())
    return std::nullopt;
// Check for second non-fixed dimension
if (std::find_if(dd_it, dds.end(), not_fixed) != param_shapes.end())
    return std::nullopt;
result.max = dd_it->max;
result.min = dd_it->min;
result.index = dd_it - dds.begin();
return result;
```

pfultz2: Also `any_of` can be used here as well for the second search.

src/include/migraphx/compile_options.hpp
CharlieL7: I tried getting rid of the flag, but it's probably going to be needed for when we actually want things to run dynamically on the GPU. Removing it also breaks the `select_module` verify tests. Instead I just made it default to true for now.

CharlieL7: Going to change it back to defaulting to false. The `promote_literals` pass breaks too many tests otherwise.

pfultz2 (approved):
This can be merge in for now, but we should try to remove the compile flag in a future PR.

Issue #1560, opened 2023-02-10
Update rocm-cmake used in MIGraphX


umangyadav:
Current SHA is more than  6 months older. 

CI is complaining about `-config-file` which had been fixed in rocm-cmake but MIGraphX is not using latest rocm-cmake.

PR #1581, opened 2023-02-27
Update rocm-cmake and half package URL

umangyadav:
`half` (header only) package in ROCm now ships under `/opt/rocm/include/half.hpp`.  Half package doesn't really change that much. It is pinned to rocm-5.4.2 same as CI rocm-version. 

Updated rocm-cmake to latest SHA which provides utility to wrap `half.hpp` such that it is backward compatible. which also Fixes https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1560

Developers would need to build dependencies again locally after this change.


umangyadav:
Perf CI test needs to build with `rbuild` and that requires changes in other repo. It will fail for now.

TedThemistokleous (approved):

kahmed10 (approved):

causten (approved):

Issue #1553, opened 2023-02-03
Support user-driven tuning for rocBLAS inside  MIGraphX


umangyadav:
rocBLAS has released [beta API](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/library/include/internal/rocblas-beta.h) to allow [users to find out](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/clients/samples/example_user_driven_tuning.cpp) the best solver/implementation for the given GEMM problem. 

Currently, MIGraphX relies on the default solution from rocBLAS. MIGraphX can make use of BETA API during the compile time and can serialize the solution index to use it in later inference runs. 

This could potentially yield better performance on users' hardware where rocBLAS's default solution may not work as efficiently or if rocBLAS doesn't have tuned parameters for the user's problem or their hardware setup. 

[How it works in rocBLAS : ](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/clients/samples/example_user_driven_tuning.cpp)

1. Get maximum applicable rocBLAS solutions to users' problem 
2. Get solution_index of those solutions. 
3. Profile each rocBLAS solution to  find out best solution 
4. Save best solution_index for later use. 

TO-DOs in MIGraphX : 

- [ ] rocBLAS function calls requires a long list of arguments to pass. Refactor current code such that, we encapsulate building all those args into a class/Struct and it can be reused in other places. 
- [ ] Add a member in migraphx' rocblas_gemm to store solution index. 
       Store best solution index if available otherwise default value (<=0) 
- [ ] Add `tune()` method in MIGraphX's  rocblas_gemm class, which would build rocBLAS args and would follow steps 1-3 above and return solution_index. Assert that the best solution is valid. 
  - [ ] How to Profile?
    - [ ] 	Directly run for loops  inside `tune()` for now, which should be simpler to implement. 
    - [ ] 	Later, Make use of `gpu-driver`.  Link `migraphx_gpu` to `gpu-driver` to call  into  that. 
    - [ ] 	Not  sure if we can directly supply json file, otherwise directly call into `time_op()`. May need to avoid calling into `finalize()` in that case. 
- [ ] Call into `tune()` during lowering for GPU if tuning is enabled. 
[...]

PR #1585, opened 2023-02-28
Auto GEMM tuning with rocBLAS

bpickrel:
Implements a new feature in the rocBLAS API to tune GEMM solutions automatically.  The finalize() method calls a new tune() method which simply iterates through the list of available solutions, timing them all and choosing the fastest.

This PR is incomplete at this time--changes are still in progress.

At the time of creating this PR, these API calls are a Beta-only feature and are not available in regular ROCm releases.  It is enabled with the preprocessor value ```ROCBLAS_BETA_FEATURES_API```, which is set by CMakeLists.txt if the needed rocBLAS version is installed at compile time.

Update: verified that the PR can be built using the Rocm ver 5.5 API and the internal_testing image is no longer needed.  Set the following line in Dockerfile: `RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/.apt_5.5/ ubuntu main > /etc/apt/sources.list.d/rocm.list'`

See [issue 1553](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1553)


src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -132,6 +133,43 @@ struct rocblas_gemm
...
+        rocblas_initialize();
+        rocblas_status_ check_valid;
+        if(ctx.get_exhaustive_tune_flag() && solution_idx == 0)
+        {
+            if(this->name() == "gpu::gemm")
+            {
+                auto gemmImpl = gemm_impl<float>(
+                    output_shape, input_shapes, alpha, beta, int8_x4_format, compute_fp32);
+                solution_idx = gemmImpl.tune(ctx, input_shapes);
+                check_valid  = gemmImpl.validate(ctx, input_shapes, solution_idx);
```
umangyadav, line 148: Only need to validate if we are de-serializing. No need to validate just after tuning since it would be valid for sure.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -132,6 +133,43 @@ struct rocblas_gemm
...
+            {
+                auto gemmImpl = gemm_impl<int32_t>(
+                    output_shape, input_shapes, alpha, beta, int8_x4_format, compute_fp32);
+                solution_idx = gemmImpl.tune(ctx, input_shapes);
+                check_valid  = gemmImpl.validate(ctx, input_shapes, solution_idx);
+            }
+        }
+        else if(solution_idx != 0)
+        {
+            auto gemmImpl = gemm_impl<int32_t>(
```
umangyadav, line 160: This should check first if it is `quant_gemm` or regular gemm . If it quant-Gemm then only use `int32_t` otherwise use `float`.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -132,6 +133,43 @@ struct rocblas_gemm
...
+                output_shape, input_shapes, alpha, beta, int8_x4_format, compute_fp32);
+            check_valid = gemmImpl.validate(ctx, input_shapes, solution_idx);
+        }
+
+        // Check if solution is valid for problem (could be invalid if this model was
+        // tuned with a different rocBLAS version)
+        if(check_valid == rocblas_status_invalid_value)
+        {
+            std::cerr << "WARNING:  tuned solution is invalid; reverting to default" << std::endl;
+            solution_idx = 0;
```
umangyadav, line 170: This check is better done inside `validate()`.  `validate()` can return `solution_index` if it is valid else can return `0`, the default.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,492 @@
...
+    bool transa, transb;
+    T alpha, beta;
+    void* alpha_v = nullptr;
+    void* beta_v  = nullptr;
+    rocblas_gemm_flags int8_flag;
+    rocblas_int lda, ldb, ldc, ldd;
+    rocblas_int a_stride, b_stride, c_stride, d_stride;
+    rocblas_datatype compute_type, arg_type, output_type;
+    bool strided_batched = true, is_3inputs = true, compute_fp32 = true;
+    uint32_t flags = 0; //  optional gemm flags.
```
umangyadav, line 508: is this used?

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,492 @@
...
+                ctx.finish();
+            };
+            for(rocblas_int hc = 0; hc < hot_calls; ++hc)
+            {
+                ctx.finish();
+                host_time += time<microseconds>(run_func);
+            }
+            // todo:  Measured time dropped from 20 us to about 6.7 us when I raised hot_calls from
+            // 1 to 11. The higher the hot_calls value, the faster per-call time up to at least 25,
+            // and increasing cold_calls makes little or no difference.  Why?
```
umangyadav, line 480: Are you still seeing this ? 
Idea of cold calls is to fill up caches.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -132,6 +133,43 @@ struct rocblas_gemm
     {
         return shapes.size() - 1;
     }
+
+    void finalize(context& ctx, const shape& output_shape, const std::vector<shape>& input_shapes)
+    {
+        rocblas_initialize();
```
umangyadav, line 139: We should call `rocblas_initialize` in get_rocblas in the context.

src/targets/gpu/include/migraphx/gpu/rocblas.hpp
```diff
@@ -25,6 +25,9 @@
 #define MIGRAPHX_GUARD_MIGRAPHLIB_ROCBLAS_HPP
 #include <migraphx/manage_ptr.hpp>
 #include <migraphx/config.hpp>
+// for development:
+#define ROCBLAS_BETA_FEATURES_API
```
umangyadav, line 29: We can check if BETA API exists  in  installed rocBLAS or not inside CMake and define ROCBLAS_BETA_FEATURES_API from there. Similar to this : 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a63ee2e0a81aa3a59ff65d7aa67b522993977274/src/targets/gpu/CMakeLists.txt#L221

umangyadav: Since it is BETA api, all the logic related to tuning should be guarded based on whether #ROCBLAS_BETA_FEATURES_API is defined or not.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
bpickrel: Can you be more specific? How do we know if we are deserializing, and where in the code?

umangyadav: After deserializing `finalize()` is called for each operator. 
I mean you can place `validate()`  inside `else_if(solution_idx !=0)`. So that it only validates for the case where tuning is not requested and solution_index is non-default.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
bpickrel: Is that exactly the same as the if-else block above, then?  The other kinds of gemm I see defined in README.MD are gpu::int8_gemm_pack_a and gpu::int8_gemm_pack_b -- how to handle them?

umangyadav: same as earlier if_then_else.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,492 @@
...
+                  << " output_type: " << output_type << "\n";
+        std::cout << "int8_flag: " << int8_flag << "\n";
+    }
+
+    auto create_strided_batched_gemm_get_solutions_args(context& ctx,
+                                                        const std::vector<argument>& args,
+                                                        rocblas_int* list,
+                                                        rocblas_int* list_size) const
+    {
+        return pack(ctx.get_stream().get_rocblas(),
```
umangyadav, line 356: You can refactor this using `pack_join()`which can merge two packs. 

e.g you can have function for `create_gemm_args()` which returns common args for running gemm and for calling get_solutions. 

You can then, extend gemm_args using pack_join to make it run with get_solutions.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,492 @@
...
+        }
+        else
+        {
+            auto gemm_args = create_gemm_args(
+                ctx, input_args, rocblas_gemm_flags_check_solution_index, solution_idx);
+            return rocblas_invoke(&rocblas_gemm_ex, gemm_args);
+        }
+    }
+
+    void print_args() const
```
umangyadav, line 336: This is mostly for debugging. You can also check if rocBLAS has any environment flags that would dump args for `rocBLAS-bench`. In that  case, we wouldn't need this. Or we can make `print_args()` to print args such that we can directly pass to `rocBLAS-bench`. 
This is just an improvement, You can work on it if you wish after everything else is fixed/done.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,492 @@
...
+            // todo:  Measured time dropped from 20 us to about 6.7 us when I raised hot_calls from
+            // 1 to 11. The higher the hot_calls value, the faster per-call time up to at least 25,
+            // and increasing cold_calls makes little or no difference.  Why?
+            host_time /= hot_calls;
+            std::cout << "Solution index " << sol << ": " << host_time << " us" << std::endl;
+
+            // track current best
+            if(host_time < bestTime)
+            {
+                printf(" current best index  %d\n", sol);
```
umangyadav, line 487: This should use `std::cerr` or `std::clog` instead of plain `printf`.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
bpickrel: done

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: It's been replaced with individual flag names for the various create...() calls.  I'll hold off on deleting it, though, until I'm sure I'm not going to refactor those yet again.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: debug code--I don't think we're going to want that line in the delivered product at all.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: I don't see any examples of similar usage.  Can pack_join put the arguments in the right order?  The different API calls contain different arguments in the middle of the lists, not just concatenated at the end.  Is it possible to insert extra arguments in a specified location in the list?  (And is it even worth it, if it's that much bother?)

umangyadav: pack_join would concatenate two different packs. It can't insert in the middle. 
I was saying we could keep strided_batched_gemm and regular gemm separate, but for each one separately, we could have function `create_common_gemm_args()` and `create_common_strided_batched_gemm_args()` and then we can extend them using pack_join for creating args for either running gemm or for fetching candidate solutions.

src/targets/gpu/CMakeLists.txt
```diff
@@ -235,6 +235,16 @@ else()
     message(STATUS "MIOpen does not have find mode api")
 endif()
 
+check_library_exists(MIOpen "rocblas_gemm_strided_batched_ex" "${MIOPEN_LOCATION}" HAS_ROCBLAS_BETA_FEATURES_API)
```
umangyadav, line 238: This should check for existence of BETA API e.g. `rocblas_gemm_get_solution()`.

And replace MIOpen with rocBLAS wherever relevant.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -168,6 +169,7 @@ struct rocblas_gemm
                 solution_idx = gemmImpl.validate(ctx, input_shapes, solution_idx);
             }
         }
+#endif
```
umangyadav, line 172: We also need to consider cases where MIGraphX wasn't compiled with rocBLAS BETA but MXR passed MiGraphX was created with BETA.  In that case, we can't use solution_index other than `0` since it wouldn't be supported by rocBLAS that was linked with MIGraphX during compile time.

src/targets/gpu/include/migraphx/gpu/rocblas.hpp
```diff
@@ -26,8 +26,9 @@
 #include <migraphx/manage_ptr.hpp>
 #include <migraphx/config.hpp>
 // for development:
-#define ROCBLAS_BETA_FEATURES_API
+#ifdef ROCBLAS_BETA_FEATURES_API
 #define ROCBLAS_NO_DEPRECATED_WARNINGS
```
umangyadav, line 30: Are you seeing deprecated warnings if you don't define this?

src/targets/gpu/include/migraphx/gpu/gemm.hpp
bpickrel: Done.  Refactored finalize() to call either tune() or validate() but not both.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
bpickrel: done.

src/targets/gpu/include/migraphx/gpu/rocblas.hpp
bpickrel: The actual warning I see reads like this:  `"rocblas_gemm_ex_get_solutions is a beta feature and is subject to change in future releases"`

src/targets/gpu/CMakeLists.txt
bpickrel: This has been fixed and now appears to get it right.

src/targets/gpu/include/migraphx/gpu/rocblas.hpp
umangyadav: You can set this inside Cmakelists.txt then along with `ROCBLAS_BETA_FEATURES_API`

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: ok, done.

src/targets/gpu/include/migraphx/gpu/rocblas.hpp
bpickrel: done

bpickrel: I assume this variable will have to be renamed some day when rocBLAS commits these features to production.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
bpickrel: done.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -1,280 +0,0 @@
...
-void gemm(context& ctx,
-          const shape& output_shape,
-          const std::vector<argument>& args,
-          int32_t alpha,
-          int32_t beta,
-          bool int8_x4_format,
-          bool compute_fp32)
-{
-    gemm_impl(ctx, output_shape, args, alpha, beta, int8_x4_format, compute_fp32);
-}
```
pfultz2, line 276: All this should stay in the .cpp file.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,488 @@
...
+            // Define the function to be timed
+            auto run_func = [&]() {
+                run(ctx, input_args, sol);
+                ctx.finish();
+            };
+            for(rocblas_int hc = 0; hc < hot_calls; ++hc)
+            {
+                ctx.finish();
+                host_time += time<microseconds>(run_func);
+            }
```
pfultz2, line 471: This should just use the `time_op` function from the driver. We just need to move it to the gpu lib so it can be reused here.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: There **are** environment flags like that; see [API_Reference_Guide logging-in-rocblas](https://rocblas.readthedocs.io/en/rocm-5.2.3/API_Reference_Guide.html?highlight=rocBLAS-bench#logging-in-rocblas).  I think that takes care of this with no action required in code.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,502 @@
...
+}
+
+static inline bool is_transposed(const shape& s)
+{
+    return (s.transposed() and s.strides().back() != 1);
+}
+
+static inline rocblas_int get_batch_stride(const shape& a)
+{
+    return a.strides()[a.strides().size() - 3];
```
bpickrel, line 109: This is giving an error from a sanitizer, presumably because a shape with less than 3 strides is being passed.  How to handle that situation?  Check and return zero?  Assert?  Is it ever correct for this to receive a lower rank shape?

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: You'll have to specify what you have in mind for that.  gemm_impl doesn't use instances of the Migraphx operation, which is what time_op takes as an argument.  Instead, we're making the rocblas calls directly.  So we can't reuse it directly--what would we substitute here?  (Also, does this change belong in the same PR?)

pfultz2:
These are the tidy warnings I get using `make tidy-target-migraphx_gpu-gemm_impl_cpp`(which will only run tidy on the gemm_impl.cpp):

```
/data/src/targets/gpu/gemm_impl.cpp:421:16: error: invalid case style for local variable 'bestTime' [readability-identifier-naming,-warnings-as-errors]
        double bestTime   = std::numeric_limits<double>::max();
               ^~~~~~~~
               best_time
/data/src/targets/gpu/gemm_impl.cpp:424:21: error: invalid case style for local variable 'bestSol' [readability-identifier-naming,-warnings-as-errors]
        rocblas_int bestSol = 0;
                    ^~~~~~~
                    best_sol
```

src/targets/gpu/CMakeLists.txt
```diff
@@ -236,6 +240,13 @@ else()
     message(STATUS "MIOpen does not have find mode api")
 endif()
 
+if(HAS_ROCBLAS_BETA_FEATURES_API)
+    target_compile_definitions(migraphx_gpu PUBLIC -DROCBLAS_BETA_FEATURES_API -DROCBLAS_NO_DEPRECATED_WARNINGS)
+    message(STATUS "MIGraphx is using Beta API of MIOpen")
```
pfultz2, line 245: This is the beta API of rocblas not miopen.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -99,180 +115,477 @@ static bool is_transposed(const shape& s)
...
+        double first_time = -1;
+        // Initialize to default solution index
+        rocblas_int bestSol = 0;
+        for(auto sol : solution_indices)
+        {
+            // Warmup: the first few calls to an op. may not be representative since there is
+            // more time taken initializing caches, etc. so we won't time them.
+            for(rocblas_int cc = 0; cc < cold_calls; ++cc)
+            {
+                run(ctx, input_args, sol);
```
pfultz2, line 431: Just call this once, dont use a loop for cold calls. Also use `run_func` not `run`.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -99,180 +115,477 @@ static bool is_transposed(const shape& s)
...
+            }
+            double host_time = 0.0;
+            // Define the function to be timed
+            auto run_func = [&]() {
+                run(ctx, input_args, sol);
+                ctx.finish();
+            };
+            for(rocblas_int hc = 0; hc < hot_calls; ++hc)
```
pfultz2, line 439: Dont use `rocblas_int`

src/targets/gpu/gemm_impl.cpp
```diff
@@ -99,180 +115,477 @@ static bool is_transposed(const shape& s)
...
+    void* beta_v  = nullptr;
+    flag_type int8_flag;
+    rocblas_int lda, ldb, ldc, ldd;
+    rocblas_int a_stride, b_stride, c_stride, d_stride;
+    rocblas_datatype compute_type, arg_type, output_type;
+    bool strided_batched = true, is_3inputs = true, compute_fp32 = true;
+#ifdef ROCBLAS_BETA_FEATURES_API
+    // tuning meta parameters
+    rocblas_int cold_calls = 18;
+    rocblas_int hot_calls  = 40;
```
pfultz2, line 481: This should be defined as a constant in the `tune` method.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -99,180 +115,477 @@ static bool is_transposed(const shape& s)
...
+            host_time /= hot_calls;
+
+            // debug only: track time for first solution.
+            if(first_time < 0)
+                first_time = host_time;
+
+            // track current best
+            if(host_time < bestTime)
+            {
+                printf(" current best index  %d, time %g\n", sol, host_time);
```
pfultz2, line 456: Use `std::cout`.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,75 @@
...
 #define MIGRAPHX_GUARD_RTGLIB_GEMM_IMPL_HPP
 
+#include <iterator>
 #include <migraphx/shape.hpp>
 #include <migraphx/argument.hpp>
+#include <migraphx/generate.hpp>
 #include <migraphx/gpu/context.hpp>
+#include <migraphx/reduce_dims.hpp>
+#include <migraphx/gpu/hip.hpp>
+#include <migraphx/time.hpp>
```
pfultz2, line 34: These extra headers can be removed or moved to the .cpp file.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,75 @@
...
+
+using milliseconds = std::chrono::duration<double, std::milli>;
+using microseconds = std::chrono::duration<double, std::micro>;
```
pfultz2, line 37: These can be removed from the header.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
```diff
@@ -24,28 +24,75 @@
...
-          const std::vector<argument>& args,
-          int32_t alpha,
-          int32_t beta,
-          bool int8_x4_format,
-          bool compute_fp32);
+#if ROCBLAS_VERSION_MAJOR >= 2 && ROCBLAS_VERSION_MINOR >= 38
+using flag_type = rocblas_gemm_flags;
+#else
+using flag_type = int;
+#endif
```
pfultz2, line 47: This can be removed from the header and moved to cpp file.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -24,6 +24,8 @@
 #ifndef MIGRAPHX_GUARD_RTGLIB_GPU_GEMM_HPP
 #define MIGRAPHX_GUARD_RTGLIB_GPU_GEMM_HPP
 
+#include <algorithm>
+#include <iterator>
```
pfultz2, line 28: These headers can be removed.

.github/workflows/add-to-project.yaml
```diff
@@ -15,5 +15,5 @@ jobs:
     steps:
       - uses: actions/add-to-project@v0.4.0
         with:
-          project-url: https://github.com/orgs/ROCmSoftwarePlatform/projects/26
+          project-url: https://github.com/orgs/ROCmSoftwarePlatform/projects/20
```
umangyadav, line 18: Is this change necessary ? undo this change otherwise.

.github/workflows/ci.yaml
```diff
@@ -39,8 +39,10 @@ jobs:
 
     - name: Prepare timestamp
       id: cache_timestamp
-      shell: bash 
-      run: echo timestamp="$(date +'%Y-%m-%dT%H:%M:%S')" >> $GITHUB_OUTPUT
+      shell: cmake -P {0}
```
umangyadav, line 42: You need to rebase with respect to develop properly.

Jenkinsfile
```diff
@@ -28,7 +28,7 @@ def rocmtestnode(Map conf) {
             rm -rf build
             mkdir build
             cd build
-            cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_DEV=On ${flags} ..
+            cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ${flags} ..
```
umangyadav, line 31: Undo this change as well.

install_deps.cmake
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/cmake -P
```
umangyadav, line 1: This file was deleted in develop .

bpickrel (commented):
@pfultz2 this PR code no longer checks the status returned by rocblas_invoke().  Do we need to retain the ability to check for degraded performance, etc?

install_deps.cmake
bpickrel: ok; deleted

src/targets/gpu/include/migraphx/gpu/gemm.hpp
bpickrel: done

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: done

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: done

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: done

src/targets/gpu/gemm_impl.cpp
bpickrel: done.  Not sure if the message is needed at all.

src/targets/gpu/gemm_impl.cpp
bpickrel: done

src/targets/gpu/gemm_impl.cpp
bpickrel: changed to int

src/targets/gpu/gemm_impl.cpp
bpickrel: Changing the parameter to 1 but leaving the loop in place, for now.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -99,180 +115,477 @@ static bool is_transposed(const shape& s)
...
+        //
+        rocblas_int list_size = 0;
+        std::vector<rocblas_int> solution_indices;
+        if(strided_batched)
+        {
+            auto common_args = create_strided_batched_args_common(ctx, input_args);
+            auto ded_args = pack(rocblas_gemm_algo_solution_index, int8_flag, nullptr, &list_size);
+
+            rocblas_invoke(&rocblas_gemm_strided_batched_ex_get_solutions,
+                           pack_join(common_args, ded_args));
```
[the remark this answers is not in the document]
bpickrel: done

src/targets/gpu/CMakeLists.txt
bpickrel: changed

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -163,6 +167,20 @@ struct miopen_convolution
         auto w_desc                = make_tensor(reshape_if_1d(inputs[1]), int8_x4_format);
         auto y_desc                = make_tensor(reshape_if_1d(output_shape));
         std::size_t workspace_size = 0;
+
+        // TODO:  Using an environment variable to disable convolution tuning while still allowing
+        // other (i.e. GEMM) tuning with --exhaustive-tune, is a workaround for a problem in which
+        // convolution tuning takes multiple hours.
+        const bool convo_tune_enabled = enabled(MIGRAPHX_ENABLE_CONV_TUNING{});
+        if(ctx.get_exhaustive_tune_flag() and not convo_tune_enabled)
+        {
```
bpickrel, line 176: This causes a flood of warning messages.  Do we have a tidy way to limit this message to only output once per compilation?

.github/workflows/add-to-project.yaml
bpickrel: merge error; fixed

.github/workflows/ci.yaml
bpickrel: merge error; fixed

Jenkinsfile
bpickrel: fixed

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -111,19 +110,21 @@ struct rocblas_gemm
...
     {
+        std::vector<shape> input_shapes;
+        std::transform(args.begin(),
+                       args.end(),
+                       std::back_inserter(input_shapes),
+                       [](const argument& x) { return x.get_shape(); });
         if(this->name() == "gpu::gemm")
         {
-            gemm(ctx, output_shape, args, alpha, beta, int8_x4_format, compute_fp32);
+            gemm_impl<float>(output_shape, input_shapes, alpha, beta, int8_x4_format, compute_fp32)
```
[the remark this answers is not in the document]
bpickrel: at long last, worked out a way to implement a visit lambda with the refactored code and the `half` verification test now passes.

src/targets/gpu/include/migraphx/gpu/gemm_impl.hpp
bpickrel: resolved it by returning 0 in the bad case (should only happen when get_batch_stride isn't needed)

bpickrel:
Closed because of corrupted merge; replaced by PR 1668

Issue #1579, opened 2023-02-23
Stable Diffusion Decoder Compilation Error


shivadbhavsar:
When compiling stable diffusion 2.0 decoder, the simplify_reshapes and simplify_algebra steps lead to reshape instructions on non-contiguous inputs. 
Need to fix onnx parser to add contiguous operators for such cases even when the original input shapes are standard.

shivadbhavsar:
Reproduce using: 

```
import onnx
from onnx import helper
from onnx import TensorProto
from onnx.numpy_helper import from_array
import numpy as np


name = 'reshape_test'
x1 = helper.make_tensor_value_info('x1', TensorProto.FLOAT, [1, 4096, 512])
y1 = helper.make_tensor_value_info('y1', TensorProto.FLOAT, [1, 4096, 512])
y2 = helper.make_tensor_value_info('y2', TensorProto.FLOAT, [1, 4096, 512])
y3 = helper.make_tensor_value_info('y3', TensorProto.FLOAT, [1, 4096, 512])

M1 = helper.make_tensor('M1', onnx.TensorProto.FLOAT, [512, 512], np.random.rand(512, 512))
M2 = helper.make_tensor('M2', onnx.TensorProto.FLOAT, [512, 512], np.random.rand(512, 512))
M3 = helper.make_tensor('M3', onnx.TensorProto.FLOAT, [512, 512], np.random.rand(512, 512))

mm1 = onnx.helper.make_node(
    'MatMul',
    inputs = ['x1', 'M1'],
    outputs= ['m1'],
)

mm2 = onnx.helper.make_node(
    'MatMul',
    inputs = ['x1', 'M2'],
    outputs= ['m2'],
)

mm3 = onnx.helper.make_node(
    'MatMul',
    inputs = ['x1', 'M3'],
    outputs= ['m3'],
)

res1 = onnx.helper.make_node(
    'Reshape',
    inputs=['m1'],
    outputs=['y1'],
    shape=[1, 4096, 1, 512]
)

res2 = onnx.helper.make_node(
    'Reshape',
    inputs=['m2'],
    outputs=['res_m2'],
    shape=[1, 4096, 1, 512]
)

res3 = onnx.helper.make_node(
    'Reshape',
    inputs=['m3'],
    outputs=['y3'],
    shape=[1, 4096, 1, 512]
)

trans = helper.make_node(
    'Transpose',
    inputs=['res_m2'],
[...]

PR #1594, opened 2023-03-03
fix stable diffusion decoder non standard shape issue

shivadbhavsar:
Addressing issue described in #1579 
In simplify algebra, there are certain edge cases where we manipulate instructions in a graph where the resulting shape becomes non-standard (mostly by transpose ops). Example shown in linked issue.

Updates:
1. Always add contiguous in reshape parser to be safe. Unnecessary contiguous instructions are removed early on by the simply_reshapes pass when it looks for nop reshapes. Remaining ones are taken care of by eliminate_contiguous.
2. It's possible for strides corresponding to dim size 1 to be an arbitrary number for standard shapes (since it is still contiguous). These are the instructions that cause problems in simplify_algebra and we do not want nop_reshapes pass to remove contiguous in this case. So even if the input is technically standard, this update forces contiguous op to recompute the strides.


src/include/migraphx/op/contiguous.hpp
```diff
@@ -48,7 +48,7 @@ struct contiguous
     {
         check_shapes{inputs, *this, true}.has(1);
         auto s0 = inputs.front();
-        if(s0.dynamic() or s0.standard())
+        if(s0.dynamic() or s0.scalar())
```
pfultz2, line 51: Contiguous after scalar should be a standard shape nor a scalar shape.

shivadbhavsar: ok reverted. However this test case in op_shape_test.cpp suggests that contiguous after scalar is scalar:

TEST_CASE(contiguous_shape_scalar)
{
    migraphx::shape output{migraphx::shape::float_type};
    migraphx::shape input{migraphx::shape::float_type};
    expect_shape(output, migraphx::make_op("contiguous"), input);
}

test/op_shape_test.cpp
```diff
@@ -431,7 +431,7 @@ TEST_CASE(contiguous_dyn_shape)
 
 TEST_CASE(contiguous_shape_scalar)
 {
-    migraphx::shape output{migraphx::shape::float_type};
+    migraphx::shape output{migraphx::shape::float_type, {1}};
```
umangyadav, line 434: Can we add a test inside `shape_test` for standard() where one of the dims is 1 and stride for for that dim is a random value? It should assert that it is a standard shape.

umangyadav: Also test add test for `contiguous` in `op_shape_test` where it converts such standard shape to "standard"-standard shape.

shivadbhavsar: it seems like contiguous does not convert modify the strides for such a shape. Is that the expected behaviour?

umangyadav (approved):

causten:
@pfultz2  anything remaining?  I'd like to merge this one soon

pfultz2 (approved):

Issue #1588, opened 2023-03-01
Const fold `literals + convert` pattern


umangyadav:
`literals + convert` can be folded at compile time. 

Potential issue : Increase in size of MXR file.

TedThemistokleous:
Comes up as part of  #1283 during verify runs. Two literals get converted into literal + convert ops and breaks verify.

TedThemistokleous:
It looks like this exists already regardless of convert or gather operands

```
@0 = @literal{ ... } -> float_type, {3, 3, 3}, {9, 3, 1}
@1 = convert[target_type=8](@0) -> int32_type, {3, 3, 3}, {9, 3, 1}
@2 = @return(@1)

@0 = @literal{ ... } -> int32_type, {3, 3, 3}, {9, 3, 1}
@1 = @return(@0)

[ COMPLETE ] literal_convert_vec
[   RUN    ] literal_gather_scalar
@0 = @literal{1} -> int32_type, {1}, {0}
@1 = @literal{3, 800, 800} -> int32_type, {3}, {1}
@2 = gather[axis=0](@1,@0) -> int32_type, {1}, {0}
@3 = @return(@2)

@0 = @literal{800} -> int32_type, {1}, {0}
@1 = @return(@0)

```

Which is handled by propagate constant.

PR #1597, opened 2023-03-03
Add matcher to remove convert from literals and just create the desir…

TedThemistokleous:
…ed type output

Creates a matcher for convert(literal) and then creates something of the correct output type instead of requiring the need for a convert.

Related to issue #1588


src/simplify_reshapes.cpp
```diff
@@ -185,6 +185,25 @@ struct find_transpose
...
+    {
+        auto ins   = mr.result;
+        auto x     = ins->inputs().front();
+        auto input = x->inputs().front();
+
+        auto old_shape = input->get_shape();
+        auto lens      = old_shape.lens();
+        auto stride    = old_shape.strides();
+
+        m.add_literal(migraphx::literal(migraphx::shape(ins->get_shape().type(), lens, stride),
```
umangyadav, line 202: better thing would be similar to how we do const-folding, You can call eval on convert and add that as literal.

src/propagate_constant.cpp
```diff
@@ -36,6 +36,8 @@ bool skip_propogate(instruction_ref ins)
 {
     if(ins->name() == "contiguous")
         return skip_propogate(ins->inputs().front());
+    if(ins->name() == "convert" and ins->inputs().front()->can_eval())
```
pfultz2, line 39: `ins->inputs().front()->can_eval()` should always be true.

However, we dont want to fold the convert if its broadcasted though(unless its a scalar with 1 element). All that is already checked below.

test/propagate_constant_test.cpp
```diff
@@ -163,4 +164,42 @@ TEST_CASE(const_dot)
...
+                          {{"target_type", migraphx::to_value(migraphx::shape::int32_type)}}),
+        lit);
+    m.add_return({t});
+
+    auto out_shape = m.get_output_shapes().back();
+    run_pass(m);
+
+    auto target_lit_out_shape = migraphx::shape{migraphx::shape::int32_type, {3, 3, 3}};
+    EXPECT(target_lit_out_shape == out_shape);
+}
```
pfultz2, line 203: Just curious, do these tests fail before your updates?

TedThemistokleous: They don't

```
[   RUN    ] const_add
[ COMPLETE ] const_add
[   RUN    ] const_add_parameter
[ COMPLETE ] const_add_parameter
[   RUN    ] const_multiadd
[ COMPLETE ] const_multiadd
[   RUN    ] const_add_mul
[ COMPLETE ] const_add_mul
[   RUN    ] const_add_scalar
[ COMPLETE ] const_add_scalar
[   RUN    ] const_scalar
[ COMPLETE ] const_scalar
[   RUN    ] const_dot
[ COMPLETE ] const_dot
[   RUN    ] literal_convert_scalar
[ COMPLETE ] literal_convert_scalar
[   RUN    ] literal_convert_vec
[ COMPLETE ] literal_convert_vec
[   RUN    ] literal_gather_scalar
@0 = @literal{0.4375} -> float_type, {1}, {0}
@1 = @literal{0.4375} -> float_type, {1}, {0}
@2 = gather[axis=0](@1,@0) -> float_type, {1}, {0}
@3 = @return(@2)

@0 = @literal{0.4375} -> float_type, {1}, {0}
@1 = @return(@0)


```

src/propagate_constant.cpp
TedThemistokleous: So I should move this lower then and just check for convert?

test/propagate_constant_test.cpp
```diff
@@ -163,4 +164,60 @@ TEST_CASE(const_dot)
...
+    auto lit2 = m.add_literal(
+        migraphx::literal{migraphx::shape(migraphx::shape::int32_type, {1}, {0}), {1}});
+    auto t = m.add_instruction(migraphx::make_op("gather"), lit, lit2);
+    m.add_return({t});
+
+    auto out_shape = m.get_output_shapes().back();
+    run_pass(m);
+
+    auto target_lit_out_shape = migraphx::shape{migraphx::shape::int32_type, {1}, {0}};
+    EXPECT(target_lit_out_shape == out_shape);
```
umangyadav, line 220: `out_shape` would match even if it didn't const-fold.  
It should assert that after running pass on your module, there is no `convert' instructions left.

umangyadav: Or you can check for `EXPECT(m1 == m2)` where `m1` is  output of run_pass(m) and m2 is expected module.

TedThemistokleous: Done

src/propagate_constant.cpp
TedThemistokleous: @pfultz2  retesting this , the change doesn't do anything. Removed it.

test/propagate_constant_test.cpp
```diff
@@ -163,4 +164,64 @@ TEST_CASE(const_dot)
...
+                          {{"target_type", migraphx::to_value(migraphx::shape::int32_type)}}),
+        lit);
+    m.add_return({t});
+
+    auto out_shape = m.get_output_shapes().back();
+    run_pass(m);
+
+    auto target_lit_out_shape = migraphx::shape{migraphx::shape::int32_type, {1}, {0}};
+    EXPECT(target_lit_out_shape == out_shape);
+}
```
umangyadav, line 184: This also shouldn't compare output shapes.

TedThemistokleous: Done

PR #1605, opened 2023-03-08
[MLIR] Adds a runtime switch to trigger MLIR

manupak:
Currently, in order to use MLIR, we need to rebuild the MIGraphX with a CMake flag.

This commit changes this have an switched polarity in the cmake flag -- i.e. MIGRAPHX_DISABLE_MLIR_BUILD. Then adds ENV var MIGRAPHX_MLIR to enable MLIR offloads in the fuse_mlir pass.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/810

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -118,7 +118,7 @@ rocmtest clang_debug: rocmnode('vega') { cmake_build ->
     stage('MLIR Debug') {
         def sanitizers = "undefined"
         def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers}"
-        cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}'")
+        cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}'")
     }
 }, clang_asan: rocmnode('nogpu') { cmake_build ->
     stage('Clang ASAN') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -118,7 +118,7 @@ rocmtest clang_debug: rocmnode('vega') { cmake_build ->
     stage('MLIR Debug') {
         def sanitizers = "undefined"
         def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers}"
-        cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}'")
+        cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}'")
     }
 }, clang_asan: rocmnode('nogpu') { cmake_build ->
     stage('Clang ASAN') {
--- a/src/targets/gpu/fuse_mlir.cpp
+++ b/src/targets/gpu/fuse_mlir.cpp
@@ -27,6 +27,7 @@
 #include <migraphx/pass_manager.hpp>
 #include <migraphx/make_op.hpp>
 #include <migraphx/register_op.hpp>
+#include <migraphx/env.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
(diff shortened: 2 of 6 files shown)


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -35,7 +36,9 @@ struct module;
 
 namespace gpu {
 
-#ifdef MIGRAPHX_MLIR
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR);
```
pfultz2, line 39: Should be called `MIGRAPHX_ENABLE_MLIR`.

Jenkinsfile
```diff
@@ -118,7 +118,7 @@ rocmtest clang_debug: rocmnode('vega') { cmake_build ->
     stage('MLIR Debug') {
         def sanitizers = "undefined"
         def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers}"
-        cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}'")
+        cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}'")
```
pfultz2, line 121: MLIR still needs to be enabled at runtime.

pfultz2:
This will also break packages being built since it wont build MLIR. MLIR needs to be added to requirements.txt(and removed from the dockerfile). However, we need to make sure we dont build MLIR on GH actions.

manupak:
> This will also break packages being built since it wont build MLIR. MLIR needs to be added to requirements.txt(and removed from the dockerfile). However, we need to make sure we dont build MLIR on GH actions. It would be better to just download prebuilt packages from repo.radeon.com.

I see. I think our packaging process is a bit WIP.

How about I re-instate the explicit enabled cmake flag and also keep the runtime switch ? (so that we can do measurements w/o needing to rebuild).

cc: @sjw36

src/targets/gpu/fuse_mlir.cpp
manupak: Ack

pfultz2:
> How about I re-instate the explicit enabled cmake flag and also keep the runtime switch ?

Sounds good to me.

manupak:
@pfultz2 , I ve made the changes now. PTAL when you have some time.

manupak:
Im trying to get this PR past CI...

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/actions/runs/4373827517

Does anyone know what the following mean:

[release / MIGraphX Performance](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/actions/runs/4373827517/jobs/7654288286#step:5:28)
Input required and not supplied: token

test/gpu/mlir.cpp
```diff
@@ -187,4 +187,38 @@ module {
...
+    auto* m   = p.get_main_module();
+    auto x    = m->add_parameter("x", {migraphx::shape::float_type, {1, 8, 4, 4}});
+    auto w    = m->add_parameter("w", {migraphx::shape::float_type, {2, 8, 3, 3}});
+    auto b    = m->add_parameter("b", {migraphx::shape::float_type, {1, 2, 2, 2}});
+    auto conv = m->add_instruction(migraphx::make_op("convolution"), x, w);
+    auto add  = m->add_instruction(migraphx::make_op("add"), conv, b);
+    auto relu = m->add_instruction(migraphx::make_op("relu"), add);
+    m->add_return({relu});
+    EXPECT(verify_mlir_e2e(p));
+}
```
pfultz2, line 222: We already have verify tests for this.

test/gpu/mlir.cpp
```diff
@@ -187,4 +187,38 @@ module {
...
+{
+    migraphx::compile_options options;
+    options.offload_copy = true;
+    p.compile(migraphx::gpu::target{}, options);
+    return p.eval(inputs).front();
+}
+
+bool verify_mlir_e2e(const migraphx::program& p)
+{
+    setenv("MIGRAPHX_ENABLE_MLIR", "1", 1);
```
pfultz2, line 202: This should be set in the Jenkinsfile.

test/gpu/mlir.cpp
manupak: Sorry, I don't follow. Where would those test be?
I wanted to add full case which will also check the fuse_mlir pass

manupak:
(Here is an attempt to read between the lines)

> https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/actions/runs/4373827517
> 
> Does anyone know what the following mean:
> 
> [release / MIGraphX Performance](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/actions/runs/4373827517/jobs/7654288286#step:5:28) Input required and not supplied: token

This might potentially be coming from the fact the PR originates from my fork. I found that all other PRs in this project and pushed to this repo. So I had to create a new PR : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1610

> We already have verify tests for this.
> This should be set in the Jenkinsfile.

OK, I think I dont know/understand how the CI works here. So the comments kind of mean there is an CI (other GH Actions) that uses the jenkinsfile and runs the all tests 'MLIR Debug'. It would be great to know where to look for this CI.
On the assumption, I just removed e2e test I've added because I was thinking fuse_mlir is not tested.

Finally, this PR is supersceded by : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1610 on the assumption the CI checkout failure is due to the PR branch being in fork.

If my current assumptions are wrong, please do let me know.

manupak:
> OK, I think I dont know/understand how the CI works here. So the comments kind of mean there is an CI (other than GH Actions) that uses the jenkinsfile and runs the all tests 'MLIR Debug'. It would be great to know where to look for this CI.

ok got it -- somehow, I didn't see them being run in the PR. I can see the jenkins runs now.

Issue #1595, opened 2023-03-03
[RFC] Backend targets as plug-in shared objects


umangyadav:
### **Background:**

MIGraphX currently supports three different target backends e.g. CPU, GPU, and FPGA. In the future, it could potentially support more backends through different mechanisms. 
MIGraphX has a modular architecture and is built as a collection of shared libraries. 
•	`libmigraphx.so` provides core functionality and infrastructure for MIGraphX IR, interface with different backends and common IR passes. 
•	`libmigraphx_tf.so` provides functionality related to parsing TensorFlow/Protobuf models. 
•	`libmigraphx_onnx.so` provides functionality related to parsing onnx models. 
•	`libmigraphx_gpu.so`: GPU backend target and links to core `libmigraphx.so`. 
•	`libmigraphx_ref.so`: Ref target which provides gold standard CPU-based reference implementation to compare results against for correctness and verification. 
•	`libmigraphx_cpu.so`: CPU target that uses ZenDNN and links to core `libmigraphx.so`. 
•	`libmigraphx_fpga.so`: FPGA target 
•	`libmigraphx_c.so` is a separate C/C++ API library that external applications links against to make use of MIGraphX. C/C++ API library internally links to `libmigraphx.so`, `libmigraphx_tf/onnx`, and to all the target .so files.  

All backend target libraries links to `libmigraphx.so`. 

### **Packaging :** 

By default MIGraphX ships with only ref and GPU targets enabled. 
By default, MIGraphX is a non-relocatable package. But the probability is that it would change in the future. 
MIGraphX is available on Linux platforms and windows support is in development. 

### **Problem :** 

As the number of targets increases, the number of dependencies that it requires also increases. Currently, if any of the dependencies are not present for any of the targets then, MIGraphX would just fail to load.  For example, MIOpen is a dependency for GPU target. If MIOpen is not present then ideally MIGraphX should still be able to execute Ref target, but it fails to do so. 

MIGraphX in future would allow for multi-target execution. In such cases, it should have ability to scan for available backend targets on the machine and make them available for compilation. 

### **MIGraphX backend targets :**

All targets in MIGraphX implements a common [interface ](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/97940f81f848b42775cacc7ea6a05f4c2c4efeed/src/include/migraphx/target.hpp#L51) that includes `get_passes()`, `name()`, `allocates()` etc.  

[...]

PR #1608, opened 2023-03-10
Dynamically plug-in backend target libs

umangyadav:
Fixes https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1595


CMakeLists.txt
```diff
@@ -97,162 +104,169 @@ endif()
...
+    -analyzer-config cfg-lifetime=true
+    -analyzer-config cfg-scopes=true
 )
 include(ROCMCppCheck)
 rocm_enable_cppcheck(
     CHECKS
-        warning
-        style
-        performance
-        portability
```
pfultz2, line 213: The spaces should not be removed here. It would be good to put formatting changes in a seperate PR.

src/include/migraphx/dynamic_loader.hpp
```diff
@@ -24,15 +24,29 @@
...
+struct dlclose_wrapper
+{
+    void operator()(void* handle)
+    {
+        if(handle)
+        {
+            dlclose(handle);
+        }
+    }
+};
```
pfultz2, line 48: This should go into the .cpp file.

src/include/migraphx/dynamic_loader.hpp
```diff
@@ -45,6 +59,8 @@ struct dynamic_loader
 
     dynamic_loader(const std::vector<char>& buffer);
 
+    static void check_load_error(bool flush = false);
```
pfultz2, line 62: This function should just be in the .cpp file.

src/include/migraphx/register_op.hpp
```diff
@@ -33,15 +33,33 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+// unregister all ops for specified target, useful when unloading dynamically plugged-in target lib
+void unregister_op(const std::string& op_name);
```
pfultz2, line 37: This should take an `operation` to be consistent.

src/include/migraphx/register_op.hpp
```diff
@@ -33,15 +33,33 @@
...
+
+struct op_handler
+{
+    operation op;
+    std::string name;
+    op_handler(const operation& op_r) : op(op_r), name(op.name()){};
+    ~op_handler() { unregister_op(name); }
+};
+
+std::unordered_map<std::string, operation>& op_map();
```
pfultz2, line 47: The map should not be exposed.

src/register_op.cpp
```diff
@@ -34,6 +34,13 @@ std::unordered_map<std::string, operation>& op_map()
     return m;
 }
 void register_op(const operation& op) { op_map()[op.name()] = op; }
+
+void unregister_op(const std::string& op_name)
+{
+    assert(op_map().count(op_name));
```
pfultz2, line 40: We shouldn't assert if its missing. It should be a no-op if it isn't register.

src/include/migraphx/register_op.hpp
pfultz2: Also, the map needs to be changed to `std::unordered_map<std::string, const operation*>`.

src/register_target.cpp
```diff
@@ -21,22 +21,42 @@
...
 #include <unordered_map>
 #include <migraphx/register_target.hpp>
+#include <migraphx/ranges.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+std::unordered_map<std::string, dynamic_loader>& target_lib_loader_map()
+{
+    static std::unordered_map<std::string, dynamic_loader> m; // NOLINT
```
pfultz2, line 34: This can just be a `std::vector<dynamic_loader>`.

src/register_target.cpp
```diff
@@ -21,22 +21,42 @@
...
  */
+#include <string>
 #include <unordered_map>
 #include <migraphx/register_target.hpp>
+#include <migraphx/ranges.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+std::unordered_map<std::string, dynamic_loader>& target_lib_loader_map()
```
pfultz2, line 32: This function should be `void store_target_lib(dynamic_loader lib)`. And the global variable should not be exposed outside of the function.

src/dynamic_loader.cpp
```diff
@@ -26,18 +26,18 @@
...
-
 struct dynamic_loader_impl
 {
     dynamic_loader_impl() = default;
+
     dynamic_loader_impl(const fs::path& p, std::shared_ptr<tmp_dir> t = nullptr)
-        : handle(dlopen(p.string().c_str(), RTLD_LAZY), &dlclose), temp(std::move(t))
+        : handle(dlopen(p.string().c_str(), RTLD_LAZY), dlclose_wrapper{}), temp(std::move(t))
     {
+        dynamic_loader::check_load_error();
```
pfultz2, line 40: We should be checking the error for `dlopen` and then throw.

CMakeLists.txt
```diff
@@ -46,14 +47,19 @@ else()
     set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "")
 endif()
 
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
```
pfultz2, line 50: We shouldn't set `CMAKE_INSTALL_RPATH`.

src/targets/fpga/include/migraphx/fpga/target.hpp
```diff
@@ -43,14 +43,11 @@ struct target
     std::vector<pass> get_passes(migraphx::context& ctx, const compile_options&) const;
     migraphx::context get_context() const { return context{}; }
     supported_segments find_supported(const_module_ref mod, support_metric m) const;
-
     argument copy_to(const argument& arg) const { return arg; }
     argument copy_from(const argument& arg) const { return arg; }
     argument allocate(const shape& s) const;
 };
 
-MIGRAPHX_REGISTER_TARGET(target);
```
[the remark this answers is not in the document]
umangyadav: They are duplicated in both `.hpp` and `.cpp` files.

src/include/migraphx/dynamic_loader.hpp
umangyadav: Done.

src/include/migraphx/dynamic_loader.hpp
umangyadav: done.

src/register_op.cpp
umangyadav: `unregister_op` would only be called if the operator is registered first.  Therefore, IMO it is okay  to have an assert.

CMakeLists.txt
umangyadav: RPATH is so that it looks for target libraries first inside `/opt/rocm/lib` itself. Otherwise we need to rely on other mechanisms. 
https://linux.die.net/man/3/dlopen

src/dynamic_loader.cpp
umangyadav: When constrctor gets called, it calls into `dlopen` and then, this checks for error on `dlopen` and throws if necessary. So it is happening.

src/include/migraphx/register_op.hpp
umangyadav: That map is exposed because it needs to be instantiated first as `static` object.  `static` objects gets destructed in the order they are defined.  I would need to instantiate `static` map before registering any `static` operator, during `register_op`.

umangyadav: If we make `operation*` it would require more changes in `load_op` and `make_op` etc. 

Why does it need to be a `operation*` ? with the current changes it seems alright.

src/register_target.cpp
umangyadav: It could be but we want to `dlclose()` when target library gets unloaded.  We can erase loader for that lib, and that would handle `dlcose()` properly.

src/include/migraphx/register_op.hpp
umangyadav: maps key is `op_name` that's why it is std::string.

src/include/migraphx/register_op.hpp
pfultz2: > I would need to instantiate static map before registering any static operator, during register_op.

`register_op(op_h.op)` already calls the static map so it should be created already. I dont see why it needs to be in `register_op<T>()`.

> Why does it need to be a operation* ?

So the memory will be owned by target .so, rather than libmigraphx.so.

> with the current changes it seems alright.

Yea it may not be an issue since we always unregister it.

src/register_target.cpp
pfultz2: Ah yea, its probably faster to use a map to erase it.

src/include/migraphx/register_op.hpp
umangyadav: > register_op(op_h.op) already calls the static map so it should be created already. I dont see why it needs to be in register_op<T>().

During unloading I found sometimes otherwise it destructs static `op_map` first and then tries to call `unregister_op()`which requires `op_map()` to be present.   
Therefore calling `op_map()` first inside `register_op<T>()` gurantees destruction order.

CMakeLists.txt
pfultz2: First the target libraries should be found since libmigraphx_c.so was also found. Secondly, this sets RPATH to an absolute path, which we shouldn't be doing. Also, `make package` will add the path to ldconf so it will be found. The release packages will set the RPATH/RUNPATH(using relative paths) when building our packages.

If we want to always first search in the same directory as libmigraphx.so, then we can get the path for the .so(using dladdr) and then first check if the target lib exists in that directory. But I think we can do that in a later PR though.

src/dynamic_loader.cpp
```diff
@@ -26,18 +26,37 @@
...
+        MIGRAPHX_THROW("Dynamic loading or symbol lookup failed with " + std::string(error_msg));
+}
+
 struct dynamic_loader_impl
 {
     dynamic_loader_impl() = default;
+
     dynamic_loader_impl(const fs::path& p, std::shared_ptr<tmp_dir> t = nullptr)
-        : handle(dlopen(p.string().c_str(), RTLD_LAZY), &dlclose), temp(std::move(t))
+        : handle(dlopen(p.string().c_str(), RTLD_LAZY), dlclose_wrapper{}), temp(std::move(t))
```
pfultz2, line 57: Instead of defining a class you could just do `manage_deleter<decltype(&dlclose), &dlclose>`.

src/include/migraphx/check_context.hpp
```diff
@@ -24,39 +24,42 @@
...
+        return {};
+    }
+    void finalize(context& ctx, const shape&, const std::vector<shape>&) const { this->check(ctx); }
+    void check(context& ctx) const
+    {
+        T* x = any_cast<T>(&ctx);
+        if(x == nullptr)
+            MIGRAPHX_THROW(std::string("Unexpected context type: ") + ctx.type_id().name());
+    }
+};
```
pfultz2, line 52: Why is this class moved?

src/include/migraphx/config.hpp
```diff
@@ -24,10 +24,17 @@
...
 
 #if !defined(MIGRAPHX_USE_CLANG_TIDY) && !defined(DOXYGEN)
-#define MIGRAPHX_INLINE_NS version_1
+
+#define MIGRAPHX_CONCAT_1(x, y) x##y
+#define MIGRAPHX_CONCAT(x, y) MIGRAPHX_CONCAT_1(x, y)
+
+#define MIGRAPHX_VERSION MIGRAPHX_CONCAT(MIGRAPHX_VERSION_MAJOR, MIGRAPHX_VERSION_MINOR)
+
+#define MIGRAPHX_INLINE_NS MIGRAPHX_CONCAT(VERSION, MIGRAPHX_VERSION)
```
pfultz2, line 37: There should be a `BUILD_DEV` flag in cmake to disable this.

src/include/migraphx/config.hpp
```diff
@@ -24,10 +24,17 @@
...
+#include <migraphx/version.h>
 namespace migraphx {
 
 #if !defined(MIGRAPHX_USE_CLANG_TIDY) && !defined(DOXYGEN)
-#define MIGRAPHX_INLINE_NS version_1
+
+#define MIGRAPHX_CONCAT_1(x, y) x##y
+#define MIGRAPHX_CONCAT(x, y) MIGRAPHX_CONCAT_1(x, y)
+
+#define MIGRAPHX_VERSION MIGRAPHX_CONCAT(MIGRAPHX_VERSION_MAJOR, MIGRAPHX_VERSION_MINOR)
```
pfultz2, line 35: This should include the patch and tweak version.

pfultz2: Ah wait that needs to be added to the version.h header. This should probably be done in a seperate PR.

src/include/migraphx/register_op.hpp
pfultz2: > it destructs static op_map first and then tries to call unregister_op()which requires op_map() to be present.

So does it construct the `op_map` again? Or is there an error about missing the operator in the map?

> Therefore calling op_map() first inside register_op<T>() gurantees destruction order.

We dont want to expose these global variables in the header.

pfultz2: I see the issue, the `op_map` needs to be called before the static variable, but I think we can encapsulate that in a function.

src/include/migraphx/config.hpp
pfultz2: To improve ccache in the dev environment.

src/include/migraphx/register_op.hpp
pfultz2: Yea we can add a `register_op_init` function we call that can then calls `op_map` to construct it.

src/register_target.cpp
pfultz2: Ah wait, but you never erase from the map. Also, we cant remove when we unregister because it might not be in the map if the user links the lib or uses `dlopen`. Better to store as a vector and hide the global by making the function as `void store_target_lib(dynamic_loader lib)`.

src/include/migraphx/register_op.hpp
umangyadav: Added `op_init()`

src/dynamic_loader.cpp
umangyadav: Done.

src/include/migraphx/config.hpp
umangyadav: I'll keep this and adding `tweak` and `patch` for another PR. For now, If it's okay, i'll not make any further changes in `config.hpp`.

CMakeLists.txt
umangyadav: Okay. Makes sense. removed `RPATH`.

src/register_target.cpp
umangyadav: Done.

src/register_target.cpp
umangyadav: Done.

src/include/migraphx/check_context.hpp
umangyadav: reverted back. Not necessary anymore.

test/api/CMakeLists.txt
```diff
@@ -56,8 +56,10 @@ add_api_test(custom_op test_custom_op.cpp ${TEST_ONNX_DIR})
...
 # GPU-based tests
 if(MIGRAPHX_ENABLE_GPU)
+list(APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)
+find_package(hip)
 add_api_test(gpu test_gpu.cpp ${TEST_ONNX_DIR})
-target_link_libraries(test_api_gpu migraphx_gpu)
+target_link_libraries(test_api_gpu hip::device)
 add_api_test(custom_op_gpu test_custom_op_gpu.cpp ${TEST_ONNX_DIR})
-target_link_libraries(test_api_custom_op_gpu migraphx_gpu)
+target_link_libraries(test_api_custom_op_gpu hip::device)
```
pfultz2, line 64: Just use `hip::host` not `hip::device`.

CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ else()
     set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "")
 endif()
 
+set(CMAKE_BUILD_RPATH "${CMAKE_BINARY_DIR}/lib")
```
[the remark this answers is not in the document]
umangyadav: no it is not doing it.

src/include/migraphx/register_target.hpp
```diff
@@ -27,20 +27,35 @@
...
+#include <migraphx/dynamic_loader.hpp>
 #include <cstring>
 #include <vector>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+void store_target_lib(const dynamic_loader& lib);
+
+void target_map_init();
```
pfultz2, line 39: Should be called `register_target_init`.

src/include/migraphx/register_target.hpp
```diff
@@ -27,20 +27,35 @@
...
 #include <migraphx/target.hpp>
 #include <migraphx/auto_register.hpp>
+#include <migraphx/dynamic_loader.hpp>
 #include <cstring>
 #include <vector>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+void store_target_lib(const dynamic_loader& lib);
```
pfultz2, line 37: This doesnt need to be in the header.

src/include/migraphx/register_op.hpp
```diff
@@ -33,15 +33,33 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+// unregister all ops for specified target, useful when unloading dynamically plugged-in target lib
+void unregister_op(const std::string& op_name);
+
+struct op_handler
+{
+    operation op;
```
pfultz2, line 41: This variable is never used.

umangyadav: It is used later as when we call `register_op()`

src/include/migraphx/register_target.hpp
umangyadav: Done.

src/include/migraphx/register_target.hpp
umangyadav: Moved.

src/include/migraphx/register_target.hpp
```diff
@@ -27,20 +27,35 @@
...
+void target_map_init();
+
 void register_target(const target& t);
+void unregister_target(const std::string& name);
 target make_target(const std::string& name);
 std::vector<std::string> get_targets();
 
+struct target_handler
+{
+    target t;
```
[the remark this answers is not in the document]
umangyadav: Done.

test/api/CMakeLists.txt
umangyadav: Done.

src/dynamic_loader.cpp
```diff
@@ -21,24 +21,38 @@
...
+    char* error_msg = dlerror();
+    if(not flush and error_msg != nullptr)
+        MIGRAPHX_THROW("Dynamic loading or symbol lookup failed with " + std::string(error_msg));
+}
+
 struct dynamic_loader_impl
 {
     dynamic_loader_impl() = default;
+
+#pragma GCC diagnostic push
```
pfultz2, line 46: This needs `ifdef` for gcc.

src/include/migraphx/register_op.hpp
```diff
@@ -33,15 +33,33 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+// unregister all ops for specified target, useful when unloading dynamically plugged-in target lib
+void unregister_op(const std::string& op_name);
+
+struct op_handler
```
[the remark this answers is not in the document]
umangyadav: Done.

src/include/migraphx/config.hpp
pfultz2: It would be better to remove this for this PR otherwise everything needs to be rebuilt, and then rebuilt again when we update this with `BUILD_DEV` flag.

umangyadav: Okay. https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1627

umangyadav: Removed.

src/include/migraphx/register_target.hpp
```diff
@@ -27,20 +27,35 @@
 #include <migraphx/config.hpp>
 #include <migraphx/target.hpp>
 #include <migraphx/auto_register.hpp>
+#include <migraphx/dynamic_loader.hpp>
```
pfultz2, line 30: This header can be removed.

test/CMakeLists.txt
```diff
@@ -228,4 +217,4 @@ test_headers(migraphx/ref ${CMAKE_SOURCE_DIR}/src/targets/ref/include/migraphx/r
 
 if(MIGRAPHX_ENABLE_GPU)
     test_headers(migraphx/gpu ${CMAKE_SOURCE_DIR}/src/targets/gpu/include/migraphx/gpu/*.hpp)
-endif()
+endif()
```
pfultz2, line 220: Missing newline at the end.

umangyadav: Done.

src/include/migraphx/register_target.hpp
umangyadav: Done.

test/context_test.cpp
```diff
@@ -1,40 +0,0 @@
...
-{
-    migraphx::context ctx = migraphx::ref::context{};
-    migraphx::value v     = ctx.to_value();
-    EXPECT(v.empty());
-
-    migraphx::context cpu_ctx = migraphx::ref::context{};
-    cpu_ctx.from_value(v);
-}
-
-int main(int argc, const char* argv[]) { test::run(argc, argv); }
```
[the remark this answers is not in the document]
umangyadav: I removed `migraphx_ref` link requirement from all tests. So it is doing lazy loading for all tests  `/test/*.cpp`.

pfultz2 (approved):

TedThemistokleous (approved):

causten:
Slim performance dropped

umangyadav:
> Slim performance dropped

These changes shouldn't affect the performance. This should be safe to merge.

umangyadav:
> Slim performance dropped

@causten ready for merge now.

PR #1614, opened 2023-03-10
Add reduction fusion

pfultz2:
This will automatically fuse multiple reductions and pointwise operations.


src/fuse_reduce.cpp
```diff
@@ -0,0 +1,331 @@
...
+    auto matcher() const { return match::name("fused_reduce")(match_reduce_input()); }
+
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
+        auto reduce1 = r.result;
+        auto reduce2 = r.instructions["reduce"];
+        auto input   = r.instructions["input"];
+
+        if(reduce1->get_operator() != reduce2->get_operator())
+            return;
```
pfultz2, line 286: Test case needs to be added for this.

src/fuse_reduce.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+        std::unordered_map<instruction_ref, instruction_ref> map_ins;
+        // Copy reduce1 instructions
+        insert_module_in_submodule(rm, reduce2, map_ins);
+        if(contains(r.instructions, "broadcast"))
+        {
+            auto broadcast                       = r.instructions["broadcast"];
+            map_ins[broadcast->inputs().front()] = rm->get_returns().front();
+            auto bout                            = insert_ins_in_submodule(rm, broadcast, map_ins);
+            map_ins[input]                       = bout.front();
```
pfultz2, line 301: A test case needs to be added for this.

src/targets/gpu/target.cpp
```diff
@@ -128,6 +130,8 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         auto_contiguous{},
         optimize_module{},
         enable_pass(not enabled(MIGRAPHX_DISABLE_POINTWISE_FUSION{}), fuse_pointwise{}),
+        enable_pass(not enabled(MIGRAPHX_DISABLE_REDUCE_FUSION{}), fuse_reduce{}),
+        dead_code_elimination{},
         dead_code_elimination{},
```
TedThemistokleous, line 135: Why do we need two dead code eliminations? Shouldint there be a dead_code_elimination after the fuse_pointwise_pass?

src/targets/gpu/compile_gen.cpp
```diff
@@ -184,8 +184,132 @@ std::string generate_pointwise(const module& pm, const std::string& name)
     // Add explict conversions
     g.fresult(
         [](const shape& s) { return "migraphx::convert<" + shape::cpp_type(s.type()) + ">"; });
-    g.create_function(
-        g.generate_module(m).set_attributes({"__device__"}).set_generic_types(m).set_name(name));
+    gg.create_function(g.generate_module(m)
```
TedThemistokleous, line 187: this supposed to be  `g` instead of `gg`?

src/targets/gpu/compile_gen.cpp
```diff
@@ -184,8 +184,132 @@ std::string generate_pointwise(const module& pm, const std::string& name)
...
+    else if(op.name() == "reduce_prod")
+    {
+        reduction = "op::product{}";
+        init      = "1";
+    }
+    else
+    {
+        MIGRAPHX_THROW("Unsupported reduce");
+    }
+}
```
TedThemistokleous, line 243: Do we want to make this a map with a functor lookup based on op.name() instead of an if/else here? Do we intend to add more reductions down the road?

src/targets/gpu/compile_gen.cpp
```diff
@@ -184,8 +184,132 @@ std::string generate_pointwise(const module& pm, const std::string& name)
...
+{
+    if(ins->outputs().size() != 1)
+        return false;
+    auto output = ins->outputs().front();
+    return contains(output->name(), "reduce") or output->name() == "@return";
+}
+
+std::string generate_reduce(const module& rm, const std::string& name)
+{
+    module m = rm;
```
TedThemistokleous, line 261: Remove the copy here

src/targets/gpu/compile_gen.cpp
pfultz2: > Do we want to make this a map with a functor lookup based on op.name() instead of an if/else here?

This is not a functor, there is no parametrized types.

> Do we intend to add more reductions down the road?

Down the road I would like have just one reduce operator.

src/fuse_reduce.cpp
```diff
@@ -0,0 +1,331 @@
...
+static void insert_params(module_ref sm,
+                          instruction_ref ins,
+                          std::unordered_map<instruction_ref, instruction_ref>& map_ins)
+{
+    auto n = sm->get_parameter_shapes().size();
+    for(auto input : ins->inputs())
+    {
+        if(contains(map_ins, input))
+            continue;
+        auto s         = shape{input->get_shape().type(), input->get_shape().lens()};
```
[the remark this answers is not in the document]
pfultz2: We just use the `lens` from the submodule to compute the output shape as we propagate the permutation of the inputs. This also allows for the shape of the inputs to change as long as they have the same dimensions and type(which can happen with `eliminate_contiguous`).

src/fuse_reduce.cpp
```diff
@@ -0,0 +1,331 @@
...
+                                    std::unordered_map<instruction_ref, instruction_ref>& map_ins)
+{
+    insert_params(sm, ins, map_ins);
+    return sm->add_instructions({ins}, map_ins);
+}
+
+static auto insert_ins_in_submodule(module_ref sm, instruction_ref ins)
+{
+    std::unordered_map<instruction_ref, instruction_ref> map_ins;
+    return insert_ins_in_submodule(sm, ins, map_ins);
```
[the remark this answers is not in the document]
pfultz2: No, the map needs to be an lvalue.

src/fuse_reduce.cpp
```diff
@@ -0,0 +1,331 @@
...
+}
+
+static std::vector<instruction_ref>
+find_inputs(module_ref sm,
+            const module& parent,
+            const std::unordered_map<instruction_ref, instruction_ref>& map_ins)
+{
+    std::vector<instruction_ref> result;
+    std::map<std::string, instruction_ref> names;
+    for(auto&& [input, param] : map_ins)
```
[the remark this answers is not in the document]
pfultz2: yes.

src/include/migraphx/cpp_generator.hpp
```diff
@@ -105,6 +106,10 @@ struct cpp_generator
 
     std::string create_function(const function& f);
 
+    static std::vector<std::string>
```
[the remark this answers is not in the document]
pfultz2: This function is also used in `generate_reduce` in compile_gen.cpp. So it needs to be exposed in the header.

src/fuse_reduce.cpp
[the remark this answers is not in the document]
umangyadav: I thought we needed to be compatible with c++14 and earlier as well or was that only for the API ?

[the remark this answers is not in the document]
pfultz2: I thought it was C++11 for the API headers. Internally we can used whichever version is available on gcc.

src/targets/gpu/compile_gen.cpp
TedThemistokleous: > > Do we want to make this a map with a functor lookup based on op.name() instead of an if/else here?
> 
> This is not a functor, there is no parametrized types.
> 
I know its not right now. I meant instead of having every branch for each reduction, we create a custom functor for each reduction case instead so all we have to do is call the appropriate op instead of doing a bunch of if/else compares.

> > Do we intend to add more reductions down the road?
> 
> Down the road I would like have just one reduce operator.
That's fair and I guess makes what I was wondering a moot point.

umangyadav (approved):

TedThemistokleous (approved):

pfultz2:
All test cases have been added, so this is ready to be merged.

PR #1615, opened 2023-03-10
select_module refactor

CharlieL7:
### Description
* Changes `select_module` to use correctly use output parameters on the gpu.
* Makes one tuple output parameter for the outputs of `select_module`.
* Fixes bug in `module.assign()` where the `nparams` is not updated correctly.
* Add `output_alias()` to `hip::sync_stream` and `select_module`.
* Removes `use_local_alloc`.
* Changes `test_select_module_add` verify test to use multiple outputs.
* Fixes the trace compile/schedule bug.

### Questions
* Do we have a standard way of doing what I have in `lowering::add_select_module_op()`? `add_extend_op()` and `add_generic_op` assume there's a gpu version of the operator.
* Should I move `get_input_parameter_names()` and `get_output_parameter_names()` to `module`?


src/include/migraphx/op/select_module.hpp
```diff
@@ -43,25 +43,49 @@ struct select_module
...
     }
 
+    std::vector<std::string> get_input_parameter_names(module_ref mod) const
+    {
+        auto param_names = mod->get_parameter_names();
+        std::vector<std::string> ret;
+        std::copy_if(param_names.cbegin(),
+                     param_names.cend(),
+                     std::back_inserter(ret),
+                     [](auto pn) { return not contains(pn, "#output_"); });
```
umangyadav, line 59:
```suggestion
                     [](auto pn) { return not starts_with(pn, "#output_"); });
```

umangyadav:
> Do we have a standard way of doing what I have in lowering::add_select_module_op()? add_extend_op() and add_generic_op assume there's a gpu version of the operator.

You can use `add_extend_op("select_module", "select_module")` 

> Should I move get_input_parameter_names() and get_output_parameter_names() to module? 

I say yes to that.

src/include/migraphx/op/select_module.hpp
pfultz2: Output parameters start with the module name, so this wouldn't work.

src/include/migraphx/op/select_module.hpp
```diff
@@ -72,21 +96,48 @@ struct select_module
...
+        std::unordered_map<std::string, argument> p_map;
 
-        // add input parameters
-        auto param_names = module_to_run->get_parameter_names();
-        assert(param_names.size() <= args.size());
-        std::transform(param_names.begin(),
-                       param_names.end(),
+        // add input parameters to parameter_map
+        auto in_param_names = get_input_parameter_names(module_to_run);
+        assert(in_param_names.size() <= args.size());
```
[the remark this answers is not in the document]
pfultz2: No because the args include the output buffers for the GPU backend.

src/include/migraphx/op/select_module.hpp
```diff
@@ -72,21 +96,48 @@ struct select_module
...
+        auto out_param_names    = get_output_parameter_names(module_to_run);
+        auto output_sub_objects = args.back().get_sub_objects();
+        std::transform(out_param_names.begin(),
+                       out_param_names.end(),
+                       output_sub_objects.begin(),
+                       std::inserter(p_map, p_map.end()),
+                       [&](auto&& name, auto&& a) {
+                           auto ps = module_to_run->get_parameter_shape(name);
+                           decltype(std::make_pair(name, a)) ret;
+                           if(a.get_shape() != ps)
```
umangyadav, line 122: I thought `compute-shape` would have a tuple of shapes. so it would match exactly. 

Why wouldn't it match?

pfultz2:
> Should I move get_input_parameter_names() and get_output_parameter_names() to module?

No since this is specific to the backends that use `replace_allocate`.

src/include/migraphx/op/select_module.hpp
```diff
@@ -72,21 +96,48 @@ struct select_module
...
+                           {
+                               assert(ps.bytes() == a.get_shape().bytes());
+                               ret = std::make_pair(name, a.reshape(ps));
```
pfultz2, line 125: Just do `return std::make_pair(name, a.reshape(ps))` instead of assigning to another variable.

src/include/migraphx/op/select_module.hpp
pfultz2: I believe this is for the output. The output will return a dynamic shape, but after running the module we will have a static shape.

src/include/migraphx/op/select_module.hpp
```diff
@@ -72,21 +96,48 @@ struct select_module
...
+        std::transform(out_param_names.begin(),
+                       out_param_names.end(),
+                       output_sub_objects.begin(),
+                       std::inserter(p_map, p_map.end()),
+                       [&](auto&& name, auto&& a) {
+                           auto ps = module_to_run->get_parameter_shape(name);
+                           decltype(std::make_pair(name, a)) ret;
+                           if(a.get_shape() != ps)
+                           {
+                               assert(ps.bytes() == a.get_shape().bytes());
```
pfultz2, line 124: Shouldn't this be `s.bytes() <= a.get_shape().bytes()`?

src/include/migraphx/op/select_module.hpp
CharlieL7: The shape might not match because of purely shape changes after a gpu instruction in the submodule. For example, when I do a reduce into a squeeze the output parameter the reduce has shape like [4, 1, 2] but the output shape is [4, 2] after the reduce.

src/include/migraphx/op/select_module.hpp
CharlieL7: Should be the same number of bytes since the shapes should only not be the same if there's something like a squeeze occurring.

src/targets/gpu/lowering.cpp
```diff
@@ -361,29 +361,16 @@ struct miopen_apply
...
 
     /**
-     * Turns on use_local_alloc in the select_module submodules.
-     * Changes the submodule returns to a hip::sync_stream.
+     * Adds dynamic allocation for submodule output parameter.
      */
     void add_select_module_op()
     {
         apply_map.emplace("select_module", [=](instruction_ref ins) {
+            auto s                              = ins->get_shape();
```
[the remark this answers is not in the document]
CharlieL7: I'll leave it as is for now since it's the only time this is done.

src/include/migraphx/op/select_module.hpp
umangyadav: Do we have test coverage for this case ?

CharlieL7: Yes, it's the previous `test_select_module_reduce` verify test.

umangyadav (approved):

PR #1622, opened 2023-03-14
Fix dual literal gather

TedThemistokleous:
Fixes issue with dual literal gather inputs by skipping adding extra converts for the gather operator and its literal inputs

### DIFF
--- a/src/eliminate_data_type.cpp
+++ b/src/eliminate_data_type.cpp
@@ -40,6 +40,7 @@ void eliminate_data_type::apply(module& m) const
                                                            "roialign",
                                                            "scatternd_add",
                                                            "scatternd_mul",
+                                                           "gather",
                                                            "scatternd_none"};
     for(auto ins : iterator_for(m))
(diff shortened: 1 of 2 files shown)


pfultz2:
Why do we need to skip gather?

TedThemistokleous:
> Why do we need to skip gather?

This was causing us to create a bunch of converts to float that weren't being converted back correctly when the gather op gets folded into the resulting literal.

TedThemistokleous:
Here's the log that this is supposed to fix when running with debug build and backtrace.

[debug_gather_literals.log](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/10984143/debug_gather_literals.log)

TedThemistokleous:
Tested and verified the testcase thats failing used here, and closing this ticket out as PR #1626 solves the issue without skipping over things like in this PR. Pushed test case from this PR into #1626  to maintain coverage

PR #1625, opened 2023-03-15
Refactor dynamic_dimension to have multiple optimals

CharlieL7:
* Makes the optimals into a `std::set<std::size_t>`
* Changes `shape` object functions to handle the `opts` change
* Changes to `convolution`, `flatten`, `pooling`, and `convolution` in that they no longer calculate the output optimal dimensions. Instead returns empty `opts`. Will need to change this in the future if we want to support dynamic shapes fully.
* Many changes to tests and shape calls with respect to the new optimals.


src/include/migraphx/op/flatten.hpp
```diff
@@ -59,27 +60,22 @@ struct flatten
...
             shape::dynamic_dimension x = {
                 std::accumulate(
                     min_lens.begin(), min_lens.begin() + axis, std::size_t{1}, std::multiplies<>{}),
                 std::accumulate(
                     max_lens.begin(), max_lens.begin() + axis, std::size_t{1}, std::multiplies<>{}),
-                std::accumulate(opt_lens.begin(),
-                                opt_lens.begin() + axis,
-                                std::size_t{1},
-                                std::multiplies<>{})};
+                {}};
```
umangyadav, line 72: we can put placeholder logic similar to convolution. that is keep the compute logic but make it over `set` of optimal values.  Set would  be empty for now.

src/include/migraphx/op/pooling.hpp
```diff
@@ -134,19 +126,19 @@ struct pooling
...
                 for(size_t i = 0; i < kdims; ++i)
                 {
-                    output_dyn_dims.push_back(shape::dynamic_dimension{1, 1, 1});
+                    output_dyn_dims.push_back(shape::dynamic_dimension{1, 1});
                 }
                 return {input.type(), output_dyn_dims};
             }
             else
             {
+                // does not compute for optimals
```
umangyadav, line 135: Can't we keep placeholder logic ?

src/include/migraphx/op/slice.hpp
```diff
@@ -113,14 +113,13 @@ struct slice
         // For dynamic shape, the adjusted old_lens become the new max values,
         // while updating the old mins and opts if possible.
         std::vector<std::size_t> new_mins;
-        std::vector<std::size_t> new_opts;
         std::vector<std::size_t> old_lens;
         std::vector<std::size_t> old_strides;
+        // Doesn't handle optimals
```
umangyadav, line 118: Keep  placeholder logic ?

src/include/migraphx/shape.hpp
```diff
@@ -246,14 +248,17 @@ struct shape
     /// Return true if this shape or any of the sub_shapes are dynamic
     bool any_of_dynamic() const;
 
+    /// check that the shape is valid (for dynamic shapes)
+    bool valid() const;
```
umangyadav, line 252:
```suggestion
    bool is_valid() const;
```

src/shape.cpp
```diff
@@ -494,6 +504,33 @@ bool shape::any_of_dynamic() const
...
 }
 
+bool shape::valid() const
+{
+    if(not this->dynamic())
+    {
+        return true;
+    }
+    // check min < max and opts within [min, max]
+    bool is_valid = std::none_of(this->dyn_dims().cbegin(), this->dyn_dims().cend(), [&](auto dd) {
```
umangyadav, line 514: just return `std::none_of`

src/shape.cpp
```diff
@@ -532,19 +568,23 @@ shape::dynamic_dimension& shape::dynamic_dimension::operator-=(const std::size_t
...
-    {
-        assert(this->opt >= x);
-        this->opt -= x;
-    }
+    std::set<std::size_t> new_opts;
+    std::transform(this->opts.begin(),
+                   this->opts.end(),
+                   std::inserter(new_opts, new_opts.begin()),
+                   [&x](const auto& opt) {
+                       assert(opt >= x);
```
umangyadav, line 576: if `opt < x` then, perhaps return `0` means no optimal value?

src/shape.cpp
```diff
@@ -553,7 +593,7 @@ bool operator!=(const shape::dynamic_dimension& x, const shape::dynamic_dimensio
 }
 std::ostream& operator<<(std::ostream& os, const shape::dynamic_dimension& x)
 {
-    os << "[" << x.min << ", " << x.max << ", " << x.opt << "]";
+    os << "[ " << x.min << ", " << x.max << ", [" << migraphx::to_string_range(x.opts) << "] ]";
```
umangyadav, line 596:
```suggestion
    os << "[ " << x.min << ", " << x.max << ", {" << migraphx::to_string_range(x.opts) << "}]";
```

src/include/migraphx/shape.hpp
```diff
@@ -132,11 +134,11 @@ struct shape
 
     shape(type_t t, std::vector<dynamic_dimension> dims);
 
-    // Construct a dynamic shape from three sets of lengths (of the same rank)
+    // Construct a dynamic shape from vectors of mins, maxes, and opts
```
umangyadav, line 137: Add a docstring why `opts_list` has a set of optimal values for each dynamic-dimension.

src/shape.cpp
bpickrel: Inside none_of lambda you can reduce to
`return ((dd.min > dd.max) or 
(std::any_of(dd.opts.cbegin(), dd.opts.cend(), [&](auto o) {
               return (o < dd.min or o > dd.max);
           })));`

src/targets/gpu/lowering.cpp
```diff
@@ -361,29 +361,16 @@ struct miopen_apply
...
-                    {
-                        auto sync_out = smod->insert_instruction(
-                            last_ins, make_op("hip::sync_stream"), out_ins);
-                        smod->replace_return({sync_out});
-                    }
-                }
-            }
-            return ins;
+            inputs.push_back(output);
+            return mod->replace_instruction(ins, ins->get_operator(), inputs, ins->module_inputs());
```
bpickrel, line 373: Is this section part of the optimums change?

bpickrel: Do we have a plan for how the optimums are to be used and, therefore, how a good optimum would be chosen for various ops' outputs?   I see a lot of code for enabling this vector but no idea how it's going to be used.

CharlieL7: The opts will have to be supplied by the user. They should propagate through the model if possible. The idea being that we will optimize the code to run as fast as possible for the given optimal sizes while other sizes in the range can be slower.

src/include/migraphx/shape.hpp
CharlieL7: Removed shape.valid()

src/shape.cpp
CharlieL7: Opt should be within the range of [min, max]. So the previous assert on `min` would fail first probably.

src/include/migraphx/op/pooling.hpp
CharlieL7: Looks to be the same problem that convolution has.

bpickrel (approved):
Reminder to any future readers that this PR puts the structures in place but doesn't include an implementation.  There aren't any tests that instantiate a shape with multiple-opt dimensions.

src/include/migraphx/shape.hpp
```diff
@@ -87,12 +89,12 @@ struct shape
     {
         std::size_t min = 0;
         std::size_t max = 0;
-        std::size_t opt = 0;
+        std::set<std::size_t> opts{};
```
pfultz2, line 92: Can we call this `optimals`?

src/shape.cpp
```diff
@@ -663,10 +676,15 @@ void migraphx_from_value(const value& v, shape& s)
             auto v_dd = v.at("dynamic_dimensions");
             std::vector<shape::dynamic_dimension> dyn_dims(v.at("dynamic_dimensions").size());
             std::transform(v_dd.begin(), v_dd.end(), dyn_dims.begin(), [](migraphx::value x) {
-                auto x_min = x.at("min").template to<size_t>();
-                auto x_max = x.at("max").template to<size_t>();
-                auto x_opt = x.at("opt").template to<size_t>();
-                return shape::dynamic_dimension{x_min, x_max, x_opt};
+                auto x_min  = x.at("min").template to<size_t>();
+                auto x_max  = x.at("max").template to<size_t>();
+                auto v_opts = x.at("opts");
```
pfultz2, line 681: I think you can write `auto set_x_opts = from_value<std::set<std::size_t>>(x.at("opts"));`

src/include/migraphx/shape.hpp
CharlieL7: I can make the change in a later PR. Making the change here will likely delay things. Added a TODO note.

src/shape.cpp
```diff
@@ -676,15 +677,12 @@ void migraphx_from_value(const value& v, shape& s)
...
-                               v_opts.end(),
-                               std::inserter(set_x_opts, set_x_opts.begin()),
-                               [](auto v_opt) { return v_opt.template to<size_t>(); });
-                return shape::dynamic_dimension{x_min, x_max, set_x_opts};
+                auto x_min      = x.at("min").template to<size_t>();
+                auto x_max      = x.at("max").template to<size_t>();
+                auto v_optimals = x.at("optimals");
+                std::set<size_t> set_x_optimals =
+                    from_value<std::set<std::size_t>>(x.at("optimals"));
+                return shape::dynamic_dimension{x_min, x_max, set_x_optimals};
```
pfultz2, line 685: Actually `dynamic_dimension` has a reflect method so this could just be `return from_value<shape::dynamic_dimension>(x)`.

pfultz2 (approved):

PR #1631, opened 2023-03-17
[MLIR] add dot offloads with manual tuning support

manupak:
This commit adds dot + pointwise fusion support
along with manual tuning using rocMLIR.

# Usage

## Non-tuning (heuristic mode) default flow

```
MIGRAPHX_ENABLE_MLIR=1 ./bin/migraphx-driver perf --onnx ../models/model_torch_export_fp32.onnx --fill1 input_ids --input-dim @input_ids 1 384
```

This would show warnings that the current compilation is not using tuning db as follows : 

```
Compiling ... 
Reading: ../models/model_torch_export_fp32.onnx
WARNING: MLIR tuning db not found. Please set MIGRAPHX_MLIR_TUNING_DB for optimal performance.
fails to set param ongfx90a:sramecc+:xnack-	-transA false -transB false -g 1 -m 384 -n 768 -k 768 -t f32
WARNING: MLIR tuning db not found. Please set MIGRAPHX_MLIR_TUNING_DB for optimal performance.
....
```

### Obtain the tuning configuration for the model

```
MIGRAPHX_ENABLE_MLIR=1 MIGRAPHX_MLIR_TUNING_CFG=bert.cfg ./bin/migraphx-driver compile --onnx ../models/model_torch_export_fp32.onnx --fill1 input_ids --input-dim @input_ids 1 384
```

This will create at most two files:
1) bert.cfg.gemm
2) bert.cfg.conv (if fused convs were present in the model -- no for bert case)

### Tune using helper python scripts of rocMLIR

1) checkout https://github.com/ROCmSoftwarePlatform/rocMLIR
2) build using the target : ninja check-rocmlir-build-only ci-performance-scripts
3) tune using the configs from the previous step

```
cd build
./bin/tuningRunner.py --op=gemm --configs_file=/var/mkarunar/work/AMDMIGraphX/build/bert.cfg.gemm --output=bt.tsv --verify-mode=none
# if convs were present
./bin/tuningRunner.py --op=conv --configs_file=/var/mkarunar/work/AMDMIGraphX/build/bert.cfg.conv --output=bt.tsv --verify-mode=none
```

Note : even if the .cfg has duplicated lines, the tuning will not perform duplicated tuning and it will append .tsv files. Therefore, one could do this for several models and obtain a single database file.

### Use tuning db with MIGraphX

```
MIGRAPHX_ENABLE_MLIR=1 MIGRAPHX_MLIR_TUNING_DB=/var/mkarunar/work/rocMLIR/build/bt.tsv ./bin/migraphx-driver perf --onnx ../models/model_torch_export_fp32.onnx --fill1 input_ids --input-dim @input_ids 1 384
```

Ideally the above command should not give any warning as we expect all the problem configs to be found in the tuned database.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/823 (more like not needed now if we can make this work)

### DIFF
--- a/src/targets/gpu/fuse_mlir.cpp
+++ b/src/targets/gpu/fuse_mlir.cpp
@@ -39,8 +39,10 @@ namespace gpu {
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_MLIR);
 
 #ifdef MIGRAPHX_MLIR
+
 struct mlir_conv
 {
+    std::string name() const { return "gpu::mlir_conv"; }
     operation op = make_op("convolution");
 
     template <class Self, class F>
@@ -49,7 +51,6 @@ struct mlir_conv
         return pack(f(self.op, "op"));
     }
 
-    std::string name() const { return "gpu::mlir_conv"; }
     shape compute_shape(std::vector<shape> inputs, const std::vector<module_ref>& mods) const
     {
         check_shapes{inputs, *this}.packed_or_broadcasted();
@@ -63,43 +64,47 @@ struct mlir_conv
 };
 MIGRAPHX_REGISTER_OP(mlir_conv);
 
-namespace {
-
-MIGRAPHX_PRED_MATCHER(is_mlir_conv, instruction_ref ins)
+struct mlir_dot
 {
-    if(ins->name() != "convolution")
-        return false;
-    value v    = ins->get_operator().to_value();
-    auto group = v.at("group").to<int>();
-    if(group != 1)
-        return false;
-    // Avoid MLIR assertion: Index < Length && "Invalid index!"
-    if(ins->get_shape().lens().size() != 4)
-        return false;
-    return true;
-}
+    std::string name() const { return "gpu::mlir_dot"; }
+    operation op = make_op("dot");
 
-struct find_conv_pointwise
-{
-    // Find a convolution followed by a pointwise operation.
-    auto matcher() const
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
     {
-        auto convolution =
-            match::skip(match::name("contiguous"))(is_mlir_conv().bind("convolution"));
-        return match::name("pointwise")(match::any_of[match::inputs()](convolution.bind("x")));
+        return pack(f(self.op, "op"));
     }
 
+    shape compute_shape(std::vector<shape> inputs, const std::vector<module_ref>& mods) const
+    {
+        check_shapes{inputs, *this}.packed_or_broadcasted();
+        if(mods.size() != 1)
+            MIGRAPHX_THROW("should have one submodule.");
+        if(inputs.size() < 2)
+            MIGRAPHX_THROW("should have at least two inputs.");
+        auto n = inputs.size();
+        return op.compute_shape({inputs[n - 2], inputs[n - 1]});
+    }
+};
+MIGRAPHX_REGISTER_OP(mlir_dot);
+
+namespace {
+
+template <typename mlir_anchor_op>
+struct find_gemm_based_op_base
+{
     void apply(module_pass_manager& mpm, const match::matcher_result& r) const
     {
-        auto ins      = r.result;
-        auto conv_ins = r.instructions["convolution"];
-        auto x_ins    = r.instructions["x"]; // input after contiguous
-        auto* pm      = ins->module_inputs().front();
-        auto names    = pm->get_parameter_names();
+        auto ins           = r.result;
+        auto gemm_based_op = r.instructions["gemm_based_op"];
+        auto x_ins         = r.instructions["x"]; // input after contiguous
+        auto* pm           = ins->module_inputs().front();
+        auto names         = pm->get_parameter_names();
         // Whitelist pointwise operators
         if(std::any_of(pm->begin(), pm->end(), [](const auto& i) {
-               return not contains({"@literal", "@param", "@return", "convolution", "add", "relu"},
-                                   i.name());
+               return not contains(
+                   {"@literal", "@param", "@return", "convolution", "dot", "add", "relu"},
+                   i.name());
            }))
             return;
         // Only fuse with fp32/fp16
@@ -113,10 +118,10 @@ struct find_conv_pointwise
         mm->set_bypass();
         std::unordered_map<instruction_ref, instruction_ref> param_map;
         auto x    = mm->add_parameter("x" + std::to_string(names.size()),
-                                   conv_ins->inputs().at(0)->get_shape());
+                                   gemm_based_op->inputs().at(0)->get_shape());
         auto w    = mm->add_parameter("x" + std::to_string(names.size() + 1),
-                                   conv_ins->inputs().at(1)->get_shape());
-        auto conv = mm->add_instruction(conv_ins->get_operator(), {x, w});
+                                   gemm_based_op->inputs().at(1)->get_shape());
+        auto conv = mm->add_instruction(gemm_based_op->get_operator(), {x, w});
         std::transform(names.begin(),
                        names.end(),
                        ins->inputs().begin(),
@@ -133,12 +138,57 @@ struct find_conv_pointwise
         std::copy_if(ins->inputs().begin(),
                      ins->inputs().end(),
                      std::back_inserter(inputs),
-                     [&](auto input) { return input != conv_ins; });
-        inputs.insert(inputs.end(), conv_ins->inputs().begin(), conv_ins->inputs().end());
+                     [&](auto input) { return input != gemm_based_op; });
+        inputs.insert(inputs.end(), gemm_based_op->inputs().begin(), gemm_based_op->inputs().end());
         mpm.get_module().replace_instruction(
-            ins, mlir_conv{conv_ins->get_operator()}, inputs, {mm});
+            ins, mlir_anchor_op{gemm_based_op->get_operator()}, inputs, {mm});
     }
 };
+
+MIGRAPHX_PRED_MATCHER(is_mlir_conv, instruction_ref ins)
+{
+    if(ins->name() != "convolution")
+        return false;
+    value v    = ins->get_operator().to_value();
+    auto group = v.at("group").to<int>();
+    if(group != 1)
+        return false;
+    // Avoid MLIR assertion: Index < Length && "Invalid index!"
+    if(ins->get_shape().lens().size() != 4)
+        return false;
+    return true;
+}
+
+struct find_conv_pointwise : public find_gemm_based_op_base<mlir_conv>
+{
+    // Find a convolution followed by a pointwise operation.
+    auto matcher() const
+    {
+        auto convolution =
+            match::skip(match::name("contiguous"))(is_mlir_conv().bind("gemm_based_op"));
+        return match::name("pointwise")(match::any_of[match::inputs()](convolution.bind("x")));
+    }
+};
+
+MIGRAPHX_PRED_MATCHER(is_mlir_dot, instruction_ref ins)
+{
+    if(ins->name() != "dot")
+    {
+        return false;
+    }
+    return true;
+}
+
+struct find_dot_pointwise : public find_gemm_based_op_base<mlir_dot>
+{
+    // Find a dot followed by a pointwise operation.
+    auto matcher() const
+    {
+        auto dot = match::skip(match::name("contiguous"))(is_mlir_dot().bind("gemm_based_op"));
+        return match::name("pointwise")(match::any_of[match::inputs()](dot.bind("x")));
+    }
+};
+
 } // namespace
 
 #endif
@@ -150,6 +200,7 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
     if(mlir_enabled)
     {
         match::find_matches(mpm, find_conv_pointwise{});
+        match::find_matches(mpm, find_dot_pointwise{});
     }
     else
     {
--- a/src/targets/gpu/fuse_mlir.cpp
+++ b/src/targets/gpu/fuse_mlir.cpp
@@ -40,9 +40,9 @@ MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_MLIR);
 
 #ifdef MIGRAPHX_MLIR
 
-struct mlir_conv
+struct mlir_op
 {
-    std::string name() const { return "gpu::mlir_conv"; }
+    std::string name() const { return "gpu::mlir_op"; }
     operation op = make_op("convolution");
 
     template <class Self, class F>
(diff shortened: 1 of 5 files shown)


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -63,43 +64,47 @@ struct mlir_conv
...
+        check_shapes{inputs, *this}.packed_or_broadcasted();
+        if(mods.size() != 1)
+            MIGRAPHX_THROW("should have one submodule.");
+        if(inputs.size() < 2)
+            MIGRAPHX_THROW("should have at least two inputs.");
+        auto n = inputs.size();
+        return op.compute_shape({inputs[n - 2], inputs[n - 1]});
+    }
+};
+MIGRAPHX_REGISTER_OP(mlir_dot);
```
pfultz2, line 89: These could be combined into the same operator:

```cpp
struct mlir_op
{
    operation op;

    {
        return pack(f(self.op, "op"));
    }

    std::string name() const { return "gpu::mlir_op"; }
    shape compute_shape(std::vector<shape> inputs, const std::vector<module_ref>& mods) const
    {
        check_shapes{inputs, *this}.packed_or_broadcasted();
        if(mods.size() != 1)
            MIGRAPHX_THROW("should have one submodule.");
        if(inputs.size() < 2)
            MIGRAPHX_THROW("should have at least two inputs.");
        auto n = inputs.size();
        return op.compute_shape({inputs[n - 2], inputs[n - 1]});
    }
};
MIGRAPHX_REGISTER_OP(mlir_op);
```

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -133,12 +138,57 @@ struct find_conv_pointwise
...
 };
+
+MIGRAPHX_PRED_MATCHER(is_mlir_dot, instruction_ref ins)
+{
+    if(ins->name() != "dot")
+    {
+        return false;
+    }
+    return true;
+}
```
pfultz2, line 180: This matcher is the same as `match::name("dot")`.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -133,12 +138,57 @@ struct find_conv_pointwise
...
+
+struct find_dot_pointwise : public find_gemm_based_op_base<mlir_dot>
+{
+    // Find a dot followed by a pointwise operation.
+    auto matcher() const
+    {
+        auto dot = match::skip(match::name("contiguous"))(is_mlir_dot().bind("gemm_based_op"));
+        return match::name("pointwise")(match::any_of[match::inputs()](dot.bind("x")));
+    }
```
pfultz2, line 189: These two matchers can be combined using `match::any_of(match::name("dot"), is_mlir_conv())`:

```cpp
auto matcher() const
{
  auto dot_or_conv = match::skip(match::name("contiguous"))(
      match::any_of(match::name("dot"), is_mlir_conv()).bind("gemm_based_op"));
  return match::name("pointwise")(
      match::any_of[match::inputs()](dot_or_conv.bind("x")));
}
```

src/targets/gpu/fuse_mlir.cpp
manupak: Yes could be, if thats preferred.

I'd use mlip_binary_op or something because just mlir_op seems too generic for that specific compute_shape implementation.

src/targets/gpu/mlir.cpp
```diff
@@ -498,11 +503,24 @@ struct mlir_program
         return ins->get_shape();
     }
 
+    void set_sym_name(const module& m)
```
pfultz2, line 506: I think this should just return the symbol name:

```cpp
static std::string get_symbol_name(const module& m)
    {
        {
            if(ins->name() == "convolution" or ins->name() == "dot")
            {
                return "mlir_" + ins->name();
            }
        }
        return "main";
    }
```

src/targets/gpu/mlir.cpp
```diff
@@ -498,11 +503,24 @@ struct mlir_program
...
+    {
+        for(auto ins : iterator_for(m))
+        {
+            if(ins->name() == "convolution" || ins->name() == "dot")
```
pfultz2, line 510: Use `or` instead of `||`.

src/targets/gpu/mlir.cpp
pfultz2: And then do `sym_nam = get_symbol_name(m)` in `parse` method.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
...
+    // used with rocMLIR tuning scripts.
+    void dump_tuning_cfg(const char* probConfig)
+    {
+        std::string tuning_cfg_path = string_value_of(MIGRAPHX_MLIR_TUNING_CFG{});
+        if(!tuning_cfg_path.empty())
+        {
+            std::stringstream key(probConfig);
+            std::string prob;
+            int token_count = 0;
+            for(std::string s; getline(key, s, '\t');)
```
pfultz2, line 640: Use `split_string(key, '\t')`.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
...
+                    prob = s;
+                }
+                if(token_count > 1)
+                {
+                    break;
+                }
+                token_count++;
+            }
+
+            if(prob.substr(0, 4) == std::string("conv"))
```
pfultz2, line 655: Use `starts_with(prob, "conv")`.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
...
+        return true;
+    }
+
     mlir_context ctx;
     MlirLocation location;
     mlir_module mmodule;
     problem_params pp;
     std::deque<std::string> strings{};
     std::string target_arch;
+    MlirRockTuningTable tuning_table;
```
pfultz2, line 687: This should just be stored in the `get_module_tuned`. I dont see why it needs to be a member variable.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
...
+            {
+                std::string arch, prob, perf, key;
+                std::getline(tuning_db_tsv, arch);
+                while(std::getline(tuning_db_tsv, arch, '\t'))
+                {
+                    std::getline(tuning_db_tsv, prob, '\t');
+                    std::getline(tuning_db_tsv, perf);
+                    key             = arch + "\t" + prob;
+                    char* key_cstr  = strdup(key.c_str());
+                    char* perf_cstr = strdup(perf.c_str());
```
pfultz2, line 619: Dont use `strdup`, this leaks memory.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
...
+        {
+            std::ifstream tuning_db_tsv(tuning_db_path);
+            if(tuning_db_tsv)
+            {
+                std::string arch, prob, perf, key;
+                std::getline(tuning_db_tsv, arch);
+                while(std::getline(tuning_db_tsv, arch, '\t'))
+                {
+                    std::getline(tuning_db_tsv, prob, '\t');
+                    std::getline(tuning_db_tsv, perf);
```
pfultz2, line 616: Use `split_string` here too. Its really hard to follow what is happening here.

src/targets/gpu/fuse_mlir.cpp
pfultz2: > I'd use mlir_binary_op or something because just mlir_op seems too generic for that specific compute_shape implementation.

We can make it work with any number of inputs by checking the number of parameters in the submodule, but we can tackle that later when we do different parameters.

pfultz2:
Also, make sure `make tidy` is clean on this PR.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
...
+            {
+                // first token is arch
+                // second one is the problem cfg
+                if(token_count == 1)
+                {
+                    prob = s;
+                }
+                if(token_count > 1)
+                {
+                    break;
```
pfultz2, line 650: Actually, this probably should be `std::find_if` instead of `split_string` since you are only reading the first element, but either way would be fine though.

src/targets/gpu/mlir.cpp
pfultz2: Actually this should be a static variable so we only read the db file once and not everytime we do a mlir compile.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
 
     std::string get_tune_params(bool xdlops) { return get_mlir_perf_for_conv(pp, xdlops); }
 
+    void tuning_table_create()
+    {
+        tuning_table               = mlirRockTuningTableCreate();
```
pfultz2, line 604: Isnt there a `mlirRockTuningTableDestroy` that needs to be called?

src/targets/gpu/fuse_mlir.cpp
manupak: ok. done.

src/targets/gpu/fuse_mlir.cpp
manupak: done

src/targets/gpu/mlir.cpp
manupak: done

src/targets/gpu/mlir.cpp
manupak: done

src/targets/gpu/mlir.cpp
manupak: done

src/targets/gpu/mlir.cpp
manupak: oh dear! its not even needed. Apologies Im also picking part of this code.... removed.

src/targets/gpu/mlir.cpp
```diff
@@ -580,12 +599,93 @@ struct mlir_program
...
+                if(token_count > 1)
+                {
+                    break;
+                }
+                token_count++;
+            }
+
+            if(prob.substr(0, 4) == std::string("conv"))
+            {
+                tuning_cfg_path = tuning_cfg_path + ".conv";
```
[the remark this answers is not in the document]
manupak: done.

src/targets/gpu/mlir.cpp
manupak: done.

src/targets/gpu/mlir.cpp
manupak: Yes there is .. and it should be called -- now done.

manupak:
Thanks @pfultz2 for detailed review. much appreciated.. (Im also picking part of this code up).

I've addressed all comments except the tuning table lifetime concern. From what I can see, mlir compiler is a static set of functions to be called through out compiling the whole graph.

> Actually this should be a static variable so we only read the db file once and not everytime we do a mlir compile.

Whilst I agree with you we should only be creating the tuning table once per compilation, I dont see an immediate way to to such an init and destroy -- also personally not a fan of global static here but then I dont see non-invasive way (i.e. outside the build ifdef guard) to inject mlir things to compile flow.  

Any suggestions ?

(I've added the destroy tuning table functionality for now -- so it doesn't live beyond a single mlir_program)

src/targets/gpu/mlir.cpp
```diff
@@ -641,6 +697,34 @@ void adjust_param_shapes(module& m, const std::vector<instruction_ref>& inputs)
     }
 }
 
+static void update_tuning_table(MlirRockTuningTable& tuning_table)
```
pfultz2, line 700: This function should return the `tuning_table` and be named `create_tuning_table`.

```cpp
static mlir_tuning_table create_tuning_table()
{
...
}
```

pfultz2:
> Whilst I agree with you we should only be creating the tuning table once per compilation, I dont see an immediate way to do such an init and destroy -- also personally not a fan of a static var here but then I dont see non-invasive way (i.e. outside the build ifdef guard) to inject mlir things to compile flow. If you insist, I could add that.

Just add it as a static variable to `get_module_tuned`:

```cpp
bool get_module_tuned() const
{
    static auto tuning_table = create_tuning_table();
    if(!mlirRockTuningSetFromTable(tuning_table.get(), mmodule.get()))
    {
        const char* prob_config = mlirRockTuningGetKey(tuning_table, mmodule.get());
        std::stringstream key(prob_config);
        std::cerr << "fails to set param on" << prob_config << std::endl;
        dump_tuning_cfg(prob_config);
        return false;
    }
    return true;
}
```

src/targets/gpu/mlir.cpp
```diff
@@ -654,10 +738,13 @@ code_object_op compile_mlir(const context&, module m, const std::vector<instruct
...
     mp.find_target();
     mp.parse(m);
+    MlirRockTuningTable tuning_table = mlirRockTuningTableCreate();
+    update_tuning_table(tuning_table);
     auto mod_op = mlirModuleGetOperation(mp.mmodule.get());
     if(trace)
         std::cout << mlir_print(&mlirOperationPrint, mod_op) << std::endl;
-    auto co   = mp.compile();
+    auto co = mp.compile(tuning_table);
+    mlirRockTuningTableDestroy(tuning_table);
```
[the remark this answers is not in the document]
manupak: done. (now I know about MIGRAPHX_MANAGE_MLIR_HANDLE)

src/targets/gpu/mlir.cpp
```diff
@@ -654,10 +738,13 @@ code_object_op compile_mlir(const context&, module m, const std::vector<instruct
     mlir_program mp;
     mp.find_target();
     mp.parse(m);
+    MlirRockTuningTable tuning_table = mlirRockTuningTableCreate();
```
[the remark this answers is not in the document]
manupak: done

manupak:
> Just add it as a static variable to `get_module_tuned`:

Done.

This change requires a bump to rocMLIR commit as well (which is done here). 
Will the CI build the docker images ?

pfultz2 (approved):

causten:
> 

Lately we've been clearing the cache manually.  Not sure why it hasn't been doing that automatically

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -79,27 +80,27 @@ MIGRAPHX_PRED_MATCHER(is_mlir_conv, instruction_ref ins)
...
+        auto gemm_based_op = r.instructions["gemm_based_op"];
+        auto x_ins         = r.instructions["x"]; // input after contiguous
+        auto* pm           = ins->module_inputs().front();
+        auto names         = pm->get_parameter_names();
         // Whitelist pointwise operators
         if(std::any_of(pm->begin(), pm->end(), [](const auto& i) {
-               return not contains({"@literal", "@param", "@return", "convolution", "add", "relu"},
-                                   i.name());
+               return not contains(
+                   {"@literal", "@param", "@return", "convolution", "dot", "add", "relu"},
```
sjw36, line 102: Does this restrict our fusion flow to only add/relu elementwise ops?

sjw36 (approved):
Looks good overall. Just 1 question.

src/targets/gpu/fuse_mlir.cpp
manupak: Historically it has been the case -- the actual reasons might predate me :). 
Thus, its not a change in this PR (see the red bits), we can look to expand the whitelist in a future PR -- but I dont think we can remove it altogether (because all elementwise ops of migraphx does not have a lowering in rocMLIR)

manupak:
@causten I did clear the caches w.r.t this branch (though it wasn't that much roughly 300ish mbs).
Although, if you look at GH actions they just fail in post-run docker caching step -- IIUC, the checks have passed.

What are the next steps here ?

Issue #1627, opened 2023-03-17
Add tweak and patch version numbers 


umangyadav:
Need to add them to namespace for release and packaging and need `BUILD_DEV` flag to avoid having it for faster ccache performance.

PR #1633, opened 2023-03-17
Use version number as part of internal namespace symbol

umangyadav:
This would help prevent dynamically loading the target library that is not compiled with the same version of MIGraphX core lib. 

This change would slow down compilation if using `ccache` for the first time after bumping version. 

MIGraphX hasn't really been using the patch version so far.  With each ROCm release, we bump the version by 1. But From going 5.4.2 to 5.4.3 we don't change it.  So not sure if `patch` is actually useful or not, unless we decide to bump it each time with rocm patch release.

We also have the possibility of adding a `tweak` version .but not sure if that is any useful. 

Fixes https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1627


TedThemistokleous:
> This would help prevent dynamically loading the target library that is not compiled with the same version of MIGraphX core lib.
> 
> This change would slow down compilation if using `ccache` for the first time after bumping version.
> 
Thats reasonable though if we're using CI/externally since we won't have to track SHA-1s 

> MIGraphX hasn't really been using the patch version so far. With each ROCm release, we bump the version by 1. But From going 5.4.2 to 5.4.3 we don't change it. So not sure if `patch` is actually useful or not, unless we decide to bump it each time with rocm patch release.
> 
This would be useful since we can sanity check builds should someone have issues instead of always point to a release version.

TedThemistokleous (approved):

src/include/migraphx/config.hpp
```diff
@@ -23,11 +23,19 @@
...
-#define MIGRAPHX_INLINE_NS version_1
+
+#define MIGRAPHX_CONCAT_1(x, y) x##y
+#define MIGRAPHX_CONCAT(x, y) MIGRAPHX_CONCAT_1(x, y)
+
+#define MIGRAPHX_VERSION_M MIGRAPHX_CONCAT(MIGRAPHX_VERSION_MAJOR, MIGRAPHX_VERSION_MINOR)
+
+#define MIGRAPHX_VERSION MIGRAPHX_CONCAT(MIGRAPHX_VERSION_M, MIGRAPHX_VERSION_PATCH)
+
+#define MIGRAPHX_INLINE_NS MIGRAPHX_CONCAT(VERSION_, MIGRAPHX_VERSION)
```
pfultz2, line 38: When `BUILD_DEV` is set we should just set it to `version_1`.

pfultz2:
> This change would slow down compilation if using ccache for the first time after bumping version.

It will also slow down clang tidy as well which uses a cache. There should be a `BUILD_DEV` flag to use the same version in dev/CI so we dont have to worry about this.

> So not sure if patch is actually useful or not, unless we decide to bump it each time with rocm patch release.

We should include it so if we do bump the patch version in the future it will update all the symbols.

> We also have the possibility of adding a tweak version .but not sure if that is any useful.

The tweak should contain the commit hash, which might be useful for non-rocm releases. For sure, it will be useful for printing from the driver.

The tweak version may need to be sanitized as it could contain characters that aren't valid for C identifiers.

src/api/CMakeLists.txt
```diff
@@ -26,7 +26,8 @@ add_library(migraphx_c
     api.cpp
 )
 set_target_properties(migraphx_c PROPERTIES EXPORT_NAME c)
-rocm_set_soversion(migraphx_c 3.0)
+
+rocm_set_soversion(migraphx_c 3.0) # why does it this not get same .so number as other libs ?
```
pfultz2, line 30: This is a stable API interface so the version is only bumped when we break binary compatibility.

The other components do not have binary compatibility across versions so we bump the so version along with the migraphx version. Although, we actually do this wrong though. The SO version for 2.6 should be 20600.0 instead of 2.6, because the major version should be bumped on every version. A major version jump says the two SO files are not compatible, whereas a minor version bump means its backwards compatible. So we need to always bump the major version.

TedThemistokleous: If we do this wrong, should we fix it right now then instead of building ontop of it?

src/include/migraphx/config.hpp
umangyadav: Done.

src/include/migraphx/config.hpp
```diff
@@ -23,12 +23,23 @@
...
 #ifndef MIGRAPHX_GUARD_CONFIG_HPP
 #define MIGRAPHX_GUARD_CONFIG_HPP
-
+#include <migraphx/version.h>
 namespace migraphx {
 
 #if !defined(MIGRAPHX_USE_CLANG_TIDY) && !defined(DOXYGEN)
+
+#define MIGRAPHX_CONCAT_1(x, y) x##y
+#define MIGRAPHX_CONCAT(x, y) MIGRAPHX_CONCAT_1(x, y)
```
pfultz2, line 32: Define this as `MIGRAPHX_VERSION_CONCAT` and `MIGRAPHX_VERSION_PRIMITIVE_CONCAT`.

src/include/migraphx/config.hpp
```diff
@@ -23,12 +23,23 @@
  */
 #ifndef MIGRAPHX_GUARD_CONFIG_HPP
 #define MIGRAPHX_GUARD_CONFIG_HPP
-
+#include <migraphx/version.h>
 namespace migraphx {
```
pfultz2, line 27: The `namespace` can be removed since its not used.

src/include/migraphx/config.hpp
```diff
@@ -23,12 +23,23 @@
  */
 #ifndef MIGRAPHX_GUARD_CONFIG_HPP
 #define MIGRAPHX_GUARD_CONFIG_HPP
-
+#include <migraphx/version.h>
```
pfultz2, line 26: This should only be included when `BUILD_DEV` is not defined.

src/include/migraphx/config.hpp
```diff
@@ -23,12 +23,23 @@
...
 namespace migraphx {
 
 #if !defined(MIGRAPHX_USE_CLANG_TIDY) && !defined(DOXYGEN)
+
+#define MIGRAPHX_CONCAT_1(x, y) x##y
+#define MIGRAPHX_CONCAT(x, y) MIGRAPHX_CONCAT_1(x, y)
+
+#define MIGRAPHX_VERSION_M MIGRAPHX_CONCAT(MIGRAPHX_VERSION_MAJOR, MIGRAPHX_VERSION_MINOR)
+#define MIGRAPHX_VERSION MIGRAPHX_CONCAT(MIGRAPHX_VERSION_M, MIGRAPHX_VERSION_PATCH)
```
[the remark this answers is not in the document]
umangyadav: No, it for some reason requires indirection.

src/CMakeLists.txt
```diff
@@ -291,6 +291,7 @@ endif()
 if(HAVE_HALF_EXPR)
     target_compile_definitions(migraphx PUBLIC -DHAS_HALF_V1)
 endif()
+target_compile_definitions(migraphx PUBLIC -DBUILD_DEV)
```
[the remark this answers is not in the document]
umangyadav: Done

src/include/migraphx/config.hpp
umangyadav: Done.

src/include/migraphx/config.hpp
umangyadav: Done

src/include/migraphx/config.hpp
```diff
@@ -23,12 +23,23 @@
  */
 #ifndef MIGRAPHX_GUARD_CONFIG_HPP
 #define MIGRAPHX_GUARD_CONFIG_HPP
-
+#include <migraphx/version.h>
 namespace migraphx {
 
 #if !defined(MIGRAPHX_USE_CLANG_TIDY) && !defined(DOXYGEN)
+
+#define MIGRAPHX_CONCAT_1(x, y) x##y
```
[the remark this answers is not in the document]
umangyadav: DOne.

src/api/CMakeLists.txt
umangyadav: Other  libraries are following 
`
librocm-core.so.1.0.50402
`
kind of versioning. 

Why do they need `1.0` as prefix ?

umangyadav: Should be versioning like `2.6.50402` in that case ?

umangyadav: >  because the major version should be bumped on every version.

Every version you mean going from rocm-5.x.y to rocm-6.x.y ? Or from going rocm-5.4.2 to rocm-5.5.x ?

src/driver/main.cpp
```diff
@@ -55,6 +55,9 @@ namespace migraphx {
 namespace driver {
 inline namespace MIGRAPHX_INLINE_NS {
 
+#define MIGRAPHX_STRINGIZE_1(...) #__VA_ARGS__
+#define MIGRAPHX_STRINGIZE(...) MIGRAPHX_STRINGIZE_1(__VA_ARGS__)
```
pfultz2, line 59: This is already defined under stringutils.hpp

umangyadav: Removed.

src/include/migraphx/config.hpp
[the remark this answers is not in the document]
umangyadav: Updated. It works.

pfultz2 (approved):

umangyadav:
@performance test scripts needs to be updated to use migraphx build with `-BUILD_DEV`  flag. 
One of the performance tests is  passing `mxr` file which is not compatible.

umangyadav:
> performance test scripts needs to be updated to use migraphx build with `-BUILD_DEV` flag. One of the performance tests is passing `mxr` file which is not compatible.

Or needs better solution `check_context` is using typename for operator name:

Example: DLRM which used MXR file failed on performance tests with the following error

terminate called after throwing an instance of 'migraphx::version_2_6_0::exception'
  what():  /src/AMDMIGraphX/src/include/migraphx/ranges.hpp:112: at: Operator not found: check_context::migraphx::version_1::gpu::context

PR #1634, opened 2023-03-20
Enable header tests for FPGA and CPU backend

umangyadav:
This would create executables with individual header files. 

CPU header tests would get tested inside Jenkins as part of `NOGPU` stage.  
FPGA tests would get run through GH actions. 

Ideally, it would be better to add matrix for each backend here in GH actions. https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/7a7040aa668362cb9db35c30b73bc816301a466d/.github/workflows/ci.yaml#L186


TedThemistokleous (approved):

pfultz2 (approved):

umangyadav:
@causten  this can go in

PR #1635, opened 2023-03-20
Fix default target in driver

umangyadav:
Recent changes #1608  removed `migraphx_all_target` lib from `driver` and that led to missing compile time definitions.

Missing compile definitions led to change of default target in `driver`.

This PR fixes that. 

can't use lazy loading in `driver`. it requires GPU header files for perf measurements.


umangyadav:
can't use lazy loading in `driver`. it requires GPU header files for perf measurements.

umangyadav:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1608 

Not sure how performance tests passed on this one, it would have used `CPU` backend.

TedThemistokleous (approved):

pfultz2 (approved):

PR #1636, opened 2023-03-21
Always build ref target when building MIGraphX

umangyadav:
https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents

"Libraries following INTERFACE are appended to the link interface and are not used for linking <target>."

`migraphx_ref` would not be used linking `migraphx` but it would be part of link interface for `migraphx` and therefore would get built if any other tests/targets links to `migraphx`.


src/CMakeLists.txt
```diff
@@ -271,6 +271,7 @@ add_subdirectory(tf)
 add_subdirectory(py)
 add_subdirectory(targets/ref)
 target_link_libraries(migraphx_all_targets INTERFACE migraphx_ref)
+target_link_libraries(migraphx INTERFACE migraphx_ref)
```
pfultz2, line 274: We shouldn't do this as all of libraries will link with ref target even when its not used(including other targets, the frontend parser(such as onnx), the C and python API). We should just link the tests with `migraphx_ref` directly instead of doing this.

umangyadav: I have to link `ref` in all necessary tests then, i don't have a way to test lazy loading in our unit-test suites. 

> We shouldn't do this as all of the libraries will link with ref target even when it is not used 
Yes, but is it bad in any way? It is a dynamic shared object. It would be in the link interface and would get loaded during runtime, but wouldn't get used.   we link rocBLAS, miopen etc. but they don't get used always either.

pfultz2: > I have to link ref in all necessary tests then, i don't have a way to test lazy loading in our unit-test suites. I'll have to write special examples tests.

Doing `target_link_libraries(migraphx INTERFACE migraphx_ref)` will make the tests link in `ref` anyways, and there is no way to disable it. The CMake target interface should only contain whats needed to include the headers. 

> Yes, but is it bad in any way?

Yes if you want to test the lazy loading of the ref target, you dont want to link in the ref target. 

> It is a dynamic shared object.

Most of the time, but not always. Users could build with `-DBUILD_SHARED_LIBS=Off`.

umangyadav: Linked to `ref `  and necessary targets in all tests.

TedThemistokleous (approved):

pfultz2 (approved):

PR #1637, opened 2023-03-21
Add Additional flags to accuracy_checker.py

TedThemistokleous:
Useful to get more insight into Onnxruntime. Allows us to reuse the accuracy checker code while also allowing us to capture Execution Provider output with the --ort_run  and --ort_logging flags

Also added the --target flag as well to allow us to force using either a specific target for the accuracy checking. Originally this was defaulting to the GPU. This now allows us to use ref,  fpga, etc to quickly change targets.

Currently using this to quickly sanity check Resnext operators and the network runs


tools/accuracy/accuracy_checker.py
```diff
@@ -65,6 +65,21 @@ def parse_args():
                         help='specify input parameter dimension \
                                 with the following format --input_dim input_name:dim0,dim1,dim2...'
                         )
+    parser.add_argument('--target',
+                        type=str,
+                        default="gpu",
```
kahmed10, line 70:
```suggestion
                        default='gpu',
```
Just for consistency

tools/accuracy/accuracy_checker.py
```diff
@@ -159,7 +174,8 @@ def main():
     if args.verbose:
         print(model)
 
-    model.compile(migraphx.get_target('gpu'))
+    if args.ort_run is False:
```
umangyadav, line 177: how do you run this with ORT ? does ORT generates a compiled model that you pass to the accuracy checker?

umangyadav: The `ort_run` flag enables to run of a model through ORT with random data with optional logging.  
From what i understand then, it is not doing any accuracy checking. `accuracy_checker` script may not be the best place in that case.  It is fine for now, but later if you want to compare optimizations between different EPs and MIGraphX EPs, I would rather do it in a separate script later instead of injecting it here.

tools/accuracy/accuracy_checker.py
```diff
@@ -239,14 +264,15 @@ def load_tf_graph(model_name):
...
 
-    is_correct = check_correctness(pred_fw, pred_migx, args.tolerance,
-                                   args.tolerance, args.verbose)
-    verbose_string = ' Rerun with --verbose for detailed information.' \
-            if not args.verbose else ''
-    if is_correct:
-        print('PASSED: MIGraphX meets tolerance')
-    else:
-        print('FAILED: MIGraphX is not within tolerance.' + verbose_string)
+    if args.ort_run is False:
```
umangyadav, line 267:
```suggestion
    if not args.ort_run:
```

tools/accuracy/accuracy_checker.py
umangyadav, line 177:
```suggestion
    if not args.ort_run:
```

TedThemistokleous: I agree with some of what you mean here, but the intent here is down this would fit in the existing workflow as we can now invoke different EP as well to test accuracy relative to MIGraphX as well. Fallback performance vs standalone vs a pure MIGraphX run through our API can now be benchmarked if we were to add additional flags on top of these changes

I think the only contentious point is the ability to sanity check whether a model will even run correctly on ORT standalone in this program vs splitting.  I think that would be useful to have this here vs split it  as this tool would become a swiss army knife tool of comparison between both API's workflows that MIGraphX interfaces if a customer doesn't want to use our driver solution. Again, this is a bigger discussion in the future.

umangyadav (approved):

kahmed10 (approved):

tools/accuracy/accuracy_checker.py
```diff
@@ -65,6 +65,21 @@ def parse_args():
                         help='specify input parameter dimension \
                                 with the following format --input_dim input_name:dim0,dim1,dim2...'
                         )
+    parser.add_argument('--target',
+                        type=str,
+                        default='gpu',
+                        help='target to compile and run MIGraphX on')
+
+    parser.add_argument('--ort_run',
```
pfultz2, line 73: Uses dashes not underscores: `--ort-run`.

tools/accuracy/accuracy_checker.py
```diff
@@ -65,6 +65,21 @@ def parse_args():
...
+                        action='store_true',
+                        default=False,
+                        help='only perform an onnxruntime run')
+
+    parser.add_argument('--ort_logging',
```
pfultz2, line 78: Same here.

tools/accuracy/accuracy_checker.py
```diff
@@ -65,6 +65,21 @@ def parse_args():
                         help='specify input parameter dimension \
                                 with the following format --input_dim input_name:dim0,dim1,dim2...'
```
kahmed10, line 66:
```suggestion
                                with the following format --input-dim input_name:dim0,dim1,dim2...'
```
Bugfix in description not matching parser argument

kahmed10 (changes requested):

tools/accuracy/accuracy_checker.py
TedThemistokleous: changed

tools/accuracy/accuracy_checker.py
TedThemistokleous: done

tools/accuracy/accuracy_checker.py
TedThemistokleous: Done

tools/accuracy/accuracy_checker.py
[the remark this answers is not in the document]
TedThemistokleous: Done

kahmed10 (approved):

PR #1651, opened 2023-03-27
Threaded ref nms

TedThemistokleous:
Update NMS for resnext50. These set of changes shave down NMS from 44s per run to about 2s for the ref implementation. 

Current GPU runs still running into an issue but currently reflected in the tests added in ref_ops_test.


src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -266,31 +283,38 @@ struct nonmaxsuppression
...
                 // select next top scorer box and remove any boxes from boxes_heap that exceeds IOU
                 // threshold with the selected box
-                const auto next_top_score = boxes_heap.top();
-                boxes_heap.pop();
-                selected_boxes_inside_class.push_back(next_top_score);
+                const auto next_top_score = boxes_heap.front();
+                auto next_box             = batch_box(batch_boxes_start, next_top_score.second);
+                auto next_box_idx         = next_top_score.second;
+                // Poor man's "pop" for vector
+                boxes_heap.erase(boxes_heap.begin());
```
pfultz2, line 296: The reason why its ugly to write `pop_front` is because its really slow. I think you can just skip the first element in the `copy_if` instead `std::copy_if(std::execution::par, boxes_heap.begin()+1, boxes_heap.end(), ...)`

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -186,11 +187,19 @@ struct nonmaxsuppression
         return result;
     }
 
-    inline bool suppress_by_iou(box b1, box b2, double iou_threshold) const
+    inline bool suppress_by_iou(box& b1, box& b2, double iou_threshold) const
```
pfultz2, line 190: Pass by value.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -219,30 +226,42 @@ struct nonmaxsuppression
...
     template <class T>
-    std::priority_queue<std::pair<double, int64_t>>
+    std::vector<std::pair<double, int64_t>>
     filter_boxes_by_score(T scores_start, std::size_t num_boxes, double score_threshold) const
     {
-        std::priority_queue<std::pair<double, int64_t>> boxes_heap;
+        std::vector<std::pair<double, int64_t>> boxes_heap;
         auto insert_to_boxes_heap =
-            make_function_output_iterator([&](const auto& x) { boxes_heap.push(x); });
+            make_function_output_iterator([&](const auto& x) { boxes_heap.push_back(x); });
```
pfultz2, line 234: `std::back_inserter` can just be used instead.

src/include/migraphx/shape_for_each.hpp
```diff
@@ -54,6 +55,40 @@ void shape_for_each(const migraphx::shape& s, F f)
...
+        // keep track and spawn child thread for each call
+        threads.push_back(std::thread(call(), indices));
+    }
+
+    // wait till all children finished running
+    for(auto& th : threads)
+    {
+        th.join();
+    }
+}
```
pfultz2, line 91: Just remove this, we already have `par_for`.

Dockerfile
```diff
@@ -50,6 +50,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     hipify-clang \
     half \
     libssl-dev \
+    libtbb-dev \
```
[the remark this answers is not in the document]
umangyadav: how is TBB used in this change set ? 

Is it requirement for `std::exeuction::par` ? I don't see it being used otherwise.

test/ref_ops_test.cpp
```diff
@@ -5138,6 +5138,107 @@ TEST_CASE(nms_test)
...
+    migraphx::shape scores_s{migraphx::shape::float_type, {n}};
+    std::vector<float> scores_vec(n);
+
+    size_t sample_size = 100000;
+    float seed         = 0.0f;
+    std::mt19937 gen(seed);
+    std::uniform_real_distribution<> dis(0.0, 1.0);
+    std::vector<float> rand_samples(sample_size);
+    std::generate(boxes_vec.begin(), boxes_vec.end(), [&]() { return dis(gen); });
+    std::generate(scores_vec.begin(), scores_vec.end(), [&]() { return dis(gen); });
```
umangyadav, line 5159: we already have something like `generate_literal` as utility .

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -219,30 +226,42 @@ struct nonmaxsuppression
...
+                scores_start + num_boxes,
+                insert_to_boxes_heap,
+                [&](auto sc) {
+                    box_idx++;
+                    return sc >= score_threshold;
+                },
+                [&](auto sc) { return std::make_pair(sc, box_idx - 1); });
+        }
+        else
+        { // score is irrelevant, just push into boxes_heap and make a score-index pair
```
umangyadav, line 250: not sure if score_threshold negative is possible or not, but even in that case we want to only select boxes that have scores higher than the (negative) threshold.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: Isn't it faster to pass by ref?

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: got it.

src/include/migraphx/shape_for_each.hpp
TedThemistokleous: Will do

Dockerfile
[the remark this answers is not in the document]
TedThemistokleous: @umangyadav  yes just for std::execution

@pfultz2 - will do.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: Score can't be negative and is bounded between 0-1 since its statistical confidence the box is in a certain class. In this case it pre filters the boxes based on the confidence before we perform the IOU filtering.

TedThemistokleous: There can be a case here to probably leverage execution::par for the transform for the unfiltered (score = 0) case and probably copy_if for the score > 0 case to do the copy in parallel...

Dockerfile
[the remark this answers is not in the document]
TedThemistokleous: Done

src/include/migraphx/shape_for_each.hpp
TedThemistokleous: Done- reverted the commit entirely.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: Done

Dockerfile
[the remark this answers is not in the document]
umangyadav: Docker file would run` install_prereqs.sh` so no need to keep it here.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: Makes sense. Done

Dockerfile
pfultz2, line 53: This can be removed since its in the install_prereqs.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -140,6 +141,7 @@ struct nonmaxsuppression
     {
         std::array<double, 2> x;
         std::array<double, 2> y;
+        double stored_area;
```
pfultz2, line 144: Area is pretty trivial to compute so I dont think we need to cache this. This could also lead to errors with reading uninitialized value as well.

Dockerfile
TedThemistokleous: Done

test/ref_ops_test.cpp
umangyadav: > So just use generate_literal then? 

Yes

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: Done

CMakeLists.txt
```diff
@@ -285,5 +285,5 @@ rocm_create_package(
     MAINTAINER "AMDMIGraphX Maintainer <migraphx-lib.support@amd.com>"
     LDCONFIG
     PTH
-    DEPENDS miopen-hip rocblas hip-rocclr hip-base half
+    DEPENDS miopen-hip rocblas hip-rocclr hip-base half libtbb-dev
```
pfultz2, line 288: It should just be the runtime package which I believe is called `libtbb2`.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -253,7 +253,9 @@ struct nonmaxsuppression
                                return std::make_pair(sc, box_idx - 1);
                            });
         }
-        std::sort(std::execution::par, boxes_heap.begin(), boxes_heap.end());
+        std::sort(std::execution::par, boxes_heap.begin(), boxes_heap.end(), [&](auto a, auto b) {
```
pfultz2, line 256: Use `std::greater<>{}` instead.

CMakeLists.txt
TedThemistokleous: Done

test/ref_ops_test.cpp
```diff
@@ -5138,6 +5139,68 @@ TEST_CASE(nms_test)
     EXPECT(migraphx::verify_range(result, gold));
 }
 
+/*
```
umangyadav, line 5142: why is it commented out ?

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -219,30 +225,44 @@ struct nonmaxsuppression
...
+                std::back_inserter(boxes_heap),
+                [&](auto sc) {
+                    box_idx++;
+                    return sc >= score_threshold;
+                },
+                [&](auto sc) { return std::make_pair(sc, box_idx - 1); });
+        }
+        else
+        { // score is irrelevant, just push into boxes_heap and make a score-index pair
+            std::transform(scores_start,
```
umangyadav, line 248: why do we not use `execution` policy here as well ?

TedThemistokleous:
Related to #1283 - Speeds up nms

src/include/migraphx/op/nonmaxsuppression.hpp
pfultz2: std::pair uses lexicographical comparison, which means it would only compare the second value if the first value is equal.

If two scores are equal the sorting order is ambiguous(unless you use `stable_sort`), so comparing the second value is not a problem, but it does make the sorting more consistent without using a slower sort algo. This also matches the previous behavior with `priority_queue`.

TedThemistokleous: Yeah I saw that about pair comparison, but I guess that makes sense if we wanted to keep it consistent with the previous priority_queue  implementation. I'll update

TedThemistokleous: done

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: The use of box_idx++ here thats then used for box_idx - 1 having shared state between runs as the idx, is tied to a localtion in the box list of 4 points it needs to grab later when we batch_box

pfultz2: Also we are using `std::back_inserter` which probably wont be parallelized.

pfultz2:
So we need to check this on sles and centos to make sure there isnt a problem with using executors. We might need to have wrappers around them.

CMakeLists.txt
```diff
@@ -285,5 +285,5 @@ rocm_create_package(
     MAINTAINER "AMDMIGraphX Maintainer <migraphx-lib.support@amd.com>"
     LDCONFIG
     PTH
-    DEPENDS miopen-hip rocblas hip-rocclr hip-base half
+    DEPENDS miopen-hip rocblas hip-rocclr hip-base half libtbb2 libtbb-dev
```
pfultz2, line 288: Remove `libtbb-dev` package. We should only depend on the runtime package.

TedThemistokleous: I can't do a compile locally if I try libtbb-dev  evne if I have libtbb2 locally. It throws an error when when I remove that

```
Could not find a package configuration file provided by "TBB" with any of
  the following names:

    TBBConfig.cmake
    tbb-config.cmake

  Add the installation prefix of "TBB" to CMAKE_PREFIX_PATH or set "TBB_DIR"
  to a directory containing one of the above files.  If "TBB" provides a
  separate development package or SDK, be sure it has been installed.

```

pfultz2: Yes, thats for building migraphx. These dependencies are for running migraphx after it has been built.

TedThemistokleous: Done

TedThemistokleous:
> So we need to check this on sles and centos to make sure there isnt a problem with using executors. We might need to have wrappers around them.

Do we have any specific verisions we're pinned too, or is it valid here to just test centos:latest as well as sles:latest?

TedThemistokleous:
> > So we need to check this on sles and centos to make sure there isnt a problem with using executors. We might need to have wrappers around them.
> 
> Do we have any specific verisions we're pinned too, or is it valid here to just test centos:latest as well as sles:latest?

@pfultz2 looking at the current available builds after getting an initial centOS container going, it looks like its limited to ROCm 5.2.5 for nowis that going to be an issue for you? 

Going to see how SLES is later this evening. Had some fun fighting with Python 3 libs/dependancies

TedThemistokleous:
Reworked this to detect which OS we're using due to spotty support with tbb/oneTBB. We can ease in support for Other OS's if required and gives us control via CMakeLists.txt

test/ref_ops_test.cpp
TedThemistokleous: I think it was breaking earlier. Let me fix this.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -184,13 +185,20 @@ struct nonmaxsuppression
...
     }
 
     inline bool suppress_by_iou(box b1, box b2, double iou_threshold) const
     {
-        b1.sort();
-        b2.sort();
+        const double area1 = b1.area();
+        const double area2 = b2.area();
+
+        if(area1 <= .0f or area2 <= .0f)
```
lakhinderwalia, line 198: Area can never be negative, given the current implementation of area(), so I am not sure what is this clause `<=` going to do here.  Perhaps, for an FP value, it is best to compare against an epsilon value, if the intent is to detect zeros.

CMakeLists.txt
```diff
@@ -32,6 +32,30 @@ if (NOT CMAKE_CONFIGURATION_TYPES)
...
+        message(STATUS "Ubuntu OS detected!")
+        add_compile_definitions(DEBIAN_DISTRO)
+        set(USE_TBB On)
+    elseif(DIST STREQUAL "CentOS")
+        message(STATUS "CENTOS detected!")
+    elseif(DIST STREQUAL "Red Hat")
+        message(STATUS "Red Hat detected!")
+    elseif(DIST STREQUAL "SUSE")
+        message(STATUS "SLES detected!")
+    endif()
```
lakhinderwalia, line 55: Is it useful to flag the` else `condition? Say, for debugging purposes?

src/include/migraphx/op/nonmaxsuppression.hpp
lakhinderwalia: Please get rid of this if/else clause. Let it all be handled in a uniform way, particularly when the score is between 0 and 1 (inclusive).

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -220,30 +226,42 @@ struct nonmaxsuppression
...
+                scores_start + num_boxes,
+                std::back_inserter(boxes_heap),
+                [&](auto sc) {
+                    box_idx++;
+                    return sc >= score_threshold;
+                },
+                [&](auto sc) { return std::make_pair(sc, box_idx - 1); });
+        }
+        else
+        { // score is irrelevant, just push into boxes_heap and make a score-index pair
```
lakhinderwalia, line 248: Please get rid of this if/else clause. Let it all be handled in a uniform way, particularly when the score is between 0 and 1 (inclusive).

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -184,13 +185,20 @@ struct nonmaxsuppression
             result.y = {static_cast<double>(start[0]), static_cast<double>(start[2])};
         }
 
+        result.sort();
```
lakhinderwalia, line 188: Can half_width and half_height be negative.? This sort() should move into the else clause above. Thanks.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -38,6 +38,7 @@
 #include <migraphx/check_shapes.hpp>
 #include <migraphx/output_iterator.hpp>
 #include <migraphx/argument.hpp>
+#include <migraphx/par.hpp>
```
lakhinderwalia, line 41: Totally optional. Please cut-n-paste the documentation from the onnx spec for NMS. Example: please see Unique Operator. Thanks.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -220,30 +226,42 @@ struct nonmaxsuppression
...
+        { // score is irrelevant, just push into boxes_heap and make a score-index pair
+            std::transform(scores_start,
+                           scores_start + num_boxes,
+                           std::back_inserter(boxes_heap),
+                           [&](auto sc) {
+                               box_idx++;
+                               return std::make_pair(sc, box_idx - 1);
+                           });
+        }
+        par_sort(boxes_heap.begin(), boxes_heap.end(), std::greater<std::pair<double, int64_t>>{});
```
lakhinderwalia, line 257: A dumb question. The onnx spec doesn't specify sorting is required.. I must be missing something. Thanks.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: sort() sorts the coordinates of the box result given back since a coordinates for each box is in ascending order.

lakhinderwalia: Agreed. Sorted result is what you automatically get - in the if clause. Thanks.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: It is to check for zeros but in this cause should there be some error that double is less than zero erroneously, this would also return a false value.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: The sorting is done to determine the intersection of union of the boxes between selected boxes. This sort allows us to find the boxes with the greatest overlap in the set of all detected boxes

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -267,31 +283,33 @@ struct nonmaxsuppression
...
-                           batch_box(batch_boxes_start, next_top_score.second),
-                           iou_threshold))
-                    {
-                        remainder_boxes.push(iou_candidate_box);
-                    }
-                    boxes_heap.pop();
-                }
+                selected_indices.push_back(next_box_idx);
+
+                std::vector<std::pair<double, int64_t>> remainder_boxes(boxes_heap.size());
```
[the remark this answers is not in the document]
TedThemistokleous: Artifact from the old way we did things. Now we're doing this as a filter op but the functionality is technically the same. We're doing a repeat filter operation to sift to get the same result without requiring a priority queue

CMakeLists.txt
TedThemistokleous: Didn't think to add it as I wanted to be explicit where this is run. This change arose as originally we were having issues with TBB in other distros outside of ubuntu and this is why this change set was shelved for the time being. I know Paul got SLES working with TBB so its why I've resurrected this PR

lakhinderwalia: Appreciate your resurrecting the PR. Please add an else clause just to indicate it as an error condition, if that is logically correct. Thanks.

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: You need to sort for both cases because you can't have a negative area after creating a batch box from the stored coordinates in the index of boxes. 

Having the sort at the end ensures ordering of start coordinates doesn't matter as they may be different based on how the graph was generated and converted

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: Done

src/include/migraphx/op/nonmaxsuppression.hpp
TedThemistokleous: Add doc spec from Onnx to clear this up.

src/include/migraphx/op/nonmaxsuppression.hpp
lakhinderwalia: Thanks, but that is not a reference document, Ted. I am merely aiming that we should do the bare-minimum, as per the spec, and not do costly overheads, which are not specified. I understand, this has nothing to do with just your set of changes, and sort of changes even the prior logic..

lakhinderwalia: I think we should let sort stay. I have looked at TF and at https://pytorch.org/vision/stable/generated/torchvision.ops.nms.html, and they specify sorted order. _"int64 tensor with the indices of the elements that have been kept by NMS, sorted in decreasing order of scores"_

The onnx spec is not very clear about it.

lakhinderwalia: TF: _Greedily selects a subset of bounding boxes in descending order of score_. https://www.tensorflow.org/api_docs/python/tf/image/non_max_suppression

lakhinderwalia (commented):
Just curious, what is the performance improvement with this set of changes. Thanks.

TedThemistokleous:
> Just curious, what is the performance improvement with this set of changes. Thanks.

40s during the run to between 1-2s using parallel to do the sort

CMakeLists.txt
```diff
@@ -32,6 +32,30 @@ if (NOT CMAKE_CONFIGURATION_TYPES)
...
+endif()
```
pfultz2, line 56: None of this is needed, We already detect a correct tbb version and use it if its available: https://github.com/ROCm/AMDMIGraphX/blob/develop/src/CMakeLists.txt#L291

CMakeLists.txt
```diff
@@ -32,6 +32,30 @@ if (NOT CMAKE_CONFIGURATION_TYPES)
...
+
+set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "")
```
pfultz2, line 58: This can be removed.

Dockerfile
```diff
@@ -51,6 +51,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     hipblas  \
     hipify-clang \
     half \
+    libtbb2 \
+    libtbb-dev \
```
pfultz2, line 55: This is already handled in the install_preqs.sh script.

hip-clang.docker
```diff
@@ -29,6 +29,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     libnuma-dev \
     miopen-hip \
     rocblas \
+    libtbb2 \
+    libtbb-dev \
```
pfultz2, line 33: Same here.

CMakeLists.txt
```diff
@@ -323,5 +347,5 @@ rocm_create_package(
     MAINTAINER "AMDMIGraphX Maintainer <migraphx-lib.support@amd.com>"
     LDCONFIG
     PTH
-    DEPENDS miopen-hip rocblas hip-rocclr hip-base half
+    DEPENDS miopen-hip rocblas hip-rocclr hip-base half libtbb2
```
pfultz2, line 350: This actually needs to be added conditionally. Might skip it for this PR, but open an issue to track this. The issue is that we detect TBB in the `src` directory so we need to propagate this up to this directory. There is probably a way, but it might be more work to figure out so it would probably be best is a seperate PR.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -184,13 +236,20 @@ struct nonmaxsuppression
             result.y = {static_cast<double>(start[0]), static_cast<double>(start[2])};
         }
 
+        result.sort();
+
```
lakhinderwalia, line 240: Line 220: this could take a const ref..?

pfultz2: Its just an iterator so its cheap to copy.

CMakeLists.txt
TedThemistokleous: Understood. Created  #2565

hip-clang.docker
TedThemistokleous: Done

Dockerfile
TedThemistokleous: Removed

CMakeLists.txt
TedThemistokleous: Done

CMakeLists.txt
TedThemistokleous: Removed

pfultz2 (approved):

src/include/migraphx/op/nonmaxsuppression.hpp
lakhinderwalia: (const reference won't hurt a templated function.)

Also, compute_nms can take const references to avoid copies. Thanks.

pfultz2: > (const reference won't hurt a templated function.)

It can be a pessizimization since it introduces aliasing. 

> Also, compute_nms can take const references to avoid copies.

Those are tensor views so they dont need to be passed as reference either. We also usually never pass them as reference to the visitors.

lakhinderwalia: https://www.oreilly.com/library/view/c-coding-standards/0321113586/ch10.html will hopefully convince you, and not be so pessimistic about const references :-)

src/include/migraphx/op/nonmaxsuppression.hpp
lakhinderwalia: This reference should help point out the pitfalls of FP comparisons:
https://www.boost.org/doc/libs/1_78_0/libs/math/doc/html/math_toolkit/float_comparison.html
A FP check for zero could be written as:   
`if (area1 < std::numeric_limits<float>epsilon()) {  ... }`

lakhinderwalia:
Does resnet50, mentioned in the PR above, use NMS? Thanks.

TedThemistokleous:
> Does resnet50, mentioned in the PR above, use NMS? Thanks.

Should be. Just rebuilding locally and will get you a perf output to confirm.

lakhinderwalia:
> > Does resnet50, mentioned in the PR above, use NMS? Thanks.
> 
> Should be. Just rebuilding locally and will get you a perf output to confirm.

Thanks. You had mentioned resnext50. I misread it as resnet50. Sorry about that.

TedThemistokleous:
> > > Does resnet50, mentioned in the PR above, use NMS? Thanks.
> > 
> > 
> > Should be. Just rebuilding locally and will get you a perf output to confirm.
> 
> Thanks. You had mentioned resnext50. I misread it as resnet50. Sorry about that.

All good

here's the output if I rebuild and run resnext50 off this branch

```
Summary:
gpu::topk: 6846.38ms / 5 = 1369.28ms, 91%
nonmaxsuppression: 518.818ms / 1 = 518.818ms, 7%
gpu::code_object::reduce_max_kernel: 45.8907ms / 1 = 45.8907ms, 1%
gpu::nonzero: 41.5577ms / 5 = 8.31155ms, 1%
hip::copy_from_gpu: 28.6588ms / 4 = 7.1647ms, 1%
hip::copy_to_gpu: 17.1335ms / 1 = 17.1335ms, 1%
gpu::code_object::mlir_convolution_add_relu: 10.6448ms / 45 = 0.236552ms, 1%
gpu::code_object::gather_kernel: 5.09506ms / 51 = 0.0999031ms, 1%
gpu::code_object::mlir_convolution_add: 3.50833ms / 13 = 0.269871ms, 1%
gpu::code_object::convert_kernel: 2.84764ms / 12 = 0.237304ms, 1%
gpu::code_object::concat_kernel: 2.39282ms / 19 = 0.125938ms, 1%
gpu::code_object::mlir_slice_convolution_add_relu: 2.30368ms / 10 = 0.230368ms, 1%
gpu::code_object::mlir_convolution_add_add_relu: 2.02869ms / 12 = 0.169058ms, 1%
gpu::convolution: 1.70367ms / 16 = 0.10648ms, 1%
gpu::code_object::mul_add_mul_add_kernel: 1.26544ms / 10 = 0.126544ms, 1%
gpu::code_object::sub_kernel: 1.20046ms / 20 = 0.0600228ms, 1%
gpu::code_object::max_min_kernel: 1.01927ms / 10 = 0.101927ms, 1%
gpu::code_object::min_exp_mul_mul_kernel: 1.00145ms / 10 = 0.100145ms, 1%
gpu::code_object::mlir_reshape_convolution: 0.798219ms / 2 = 0.399109ms, 1%
gpu::code_object::add_kernel: 0.690196ms / 11 = 0.0627451ms, 1%
gpu::code_object::concat_add_kernel: 0.656245ms / 1 = 0.656245ms, 1%
gpu::code_object::add_relu_kernel: 0.560613ms / 16 = 0.0350383ms, 1%
gpu::code_object::gathernd_kernel: 0.456609ms / 5 = 0.0913218ms, 1%
gpu::code_object::contiguous_kernel: 0.393391ms / 10 = 0.0393391ms, 1%
gpu::code_object::concat_mul_mul_sub_kernel: 0.385204ms / 1 = 0.385204ms, 1%
gpu::code_object::mlir_convolution: 0.324767ms / 3 = 0.108256ms, 1%
load: 0.268283ms / 324 = 0.000828034ms, 1%
gpu::code_object::less_convert_convert_logical_xor_mod_equal_convert_convert_not_logical_and_mul_add_where_kernel: 0.255928ms / 5 = 0.0511857ms, 1%
gpu::code_object::greater_convert_kernel: 0.204977ms / 5 = 0.0409954ms, 1%
gpu::code_object::sigmoid_kernel: 0.182931ms / 5 = 0.0365862ms, 1%
gpu::code_object::mlir_reshape_convolution_add_relu: 0.17755ms / 1 = 0.17755ms, 1%
hip::hip_copy_literal: 0.152868ms / 149 = 0.00102596ms, 1%
gpu::code_object::mul_kernel: 0.14588ms / 1 = 0.14588ms, 1%
gpu::pooling: 0.0890644ms / 1 = 0.0890644ms, 1%
slice: 0.0829474ms / 36 = 0.0023041ms, 1%
gpu::code_object::add_add_kernel: 0.0689686ms / 2 = 0.0344843ms, 1%
broadcast: 0.0649445ms / 47 = 0.0013818ms, 1%
gpu::code_object::mul_add_kernel: 0.0601526ms / 1 = 0.0601526ms, 1%
multibroadcast: 0.0592436ms / 55 = 0.00107716ms, 1%
reshape_lazy: 0.0584336ms / 60 = 0.000973893ms, 1%
step: 0.0448278ms / 33 = 0.00135842ms, 1%
unsqueeze: 0.0374466ms / 42 = 0.000891586ms, 1%
gpu::code_object::relu_kernel: 0.0252852ms / 1 = 0.0252852ms, 1%
get_tuple_elem: 0.019226ms / 10 = 0.0019226ms, 1%
transpose: 0.0182303ms / 15 = 0.00121535ms, 1%
squeeze: 0.0142783ms / 8 = 0.00178479ms, 1%
check_context::migraphx::gpu::context: 0.00687026ms / 1 = 0.00687026ms, 1%
hip::sync_stream: 0.005278ms / 1 = 0.005278ms, 1%
@param: 0.00506326ms / 4 = 0.00126581ms, 1%
flatten: 0.00381802ms / 5 = 0.000763604ms, 1%
hip::hip_allocate_memory: 0.00301426ms / 1 = 0.00301426ms, 1%

Batch size: 1
Rate: 0.132936 inferences/sec
Total time: 7522.41ms
Total instructions time: 7539.77ms
Overhead time: 0.230419ms, -17.3588ms
Overhead: 0%, -0%
[ MIGraphX Version: 2.9.0. ] Complete: migraphx-driver perf resnext50_32x4d_fpn.onnx
```

lakhinderwalia (approved):
Approving for now. The fp handling of zeros should be taken care of later.

src/include/migraphx/op/nonmaxsuppression.hpp
pfultz2: Sure, but pass-by-reference is not appropriate here since you are already passing a pointer type. It much faster to pass a pointer than to pass an address to another pointer.

TedThemistokleous:
@causten  CI is all green on this, not sure why github isn't reporting it correctly still.

http://micimaster.amd.com/blue/organizations/jenkins/AMDMIGraphX/detail/PR-1651/10/pipeline/

PR #1652, opened 2023-03-28
Driver dynamic batch update

CharlieL7:
* Added `default_dyn_dim` and `dyn_param_dims` options to the driver to allow for setting dynamic shapes.
* Created associated parsers for the dynamic dimension strings.
* Changed behavior of `batch` option to handle running a specific batch size when running with a dynamic batch model.
* `shape.to_static(size_t)` function to allow for easier conversion.

### DIFF
--- a/src/driver/argument_parser.hpp
+++ b/src/driver/argument_parser.hpp
@@ -151,11 +151,20 @@ struct value_parser
         T result;
         std::stringstream ss;
         ss.str(x);
-        ss >> result;
+        // handle whitespace in string
+        if constexpr ( std::is_same<T, std::string>::value )
+        {
+            result = ss.str();
+        }
+        else
+        {
+            ss >> result;
+        }
         if(ss.fail())
             throw std::runtime_error("Failed to parse '" + x + "' as " + type_name<T>::apply());
         return result;
     }
+    
 
     template <MIGRAPHX_REQUIRES(std::is_enum<T>{} and not is_multi_value<T>{})>
     static T apply(const std::string& x)
--- a/src/driver/argument_parser.hpp
+++ b/src/driver/argument_parser.hpp
@@ -152,7 +152,7 @@ struct value_parser
         std::stringstream ss;
         ss.str(x);
         // handle whitespace in string
-        if constexpr ( std::is_same<T, std::string>::value )
+        if constexpr(std::is_same<T, std::string>::value)
         {
             result = ss.str();
         }
@@ -164,7 +164,6 @@ struct value_parser
             throw std::runtime_error("Failed to parse '" + x + "' as " + type_name<T>::apply());
         return result;
     }
-    
 
     template <MIGRAPHX_REQUIRES(std::is_enum<T>{} and not is_multi_value<T>{})>
     static T apply(const std::string& x)
--- a/src/driver/main.cpp
+++ b/src/driver/main.cpp
@@ -67,7 +67,9 @@ struct loader
     bool brief                  = false;
     std::string output_type;
     std::string output;
+    std::string default_dyn_dim;
     std::vector<std::string> param_dims;
+    std::vector<std::string> dyn_param_dims;
     std::vector<std::string> output_names;
 
     void parse(argument_parser& ap)
@@ -82,7 +84,7 @@ struct loader
         ap(file_type, {"--tf"}, ap.help("Load as tensorflow"), ap.set_value("tf"));
         ap(file_type, {"--migraphx"}, ap.help("Load as MIGraphX"), ap.set_value("migraphx"));
         ap(file_type, {"--migraphx-json"}, ap.help("Load as MIGraphX JSON"), ap.set_value("json"));
-        ap(batch, {"--batch"}, ap.help("Set batch size for model"));
+        ap(batch, {"--batch"}, ap.help("For a static model, set batch size. For a dynamic batch model, sets the batch size at runtime."));
         ap(is_nhwc, {"--nhwc"}, ap.help("Treat tensorflow format as nhwc"), ap.set_value(true));
         ap(skip_unknown_operators,
            {"--skip-unknown-operators"},
@@ -95,7 +97,14 @@ struct loader
            ap.help("Dim of a parameter (format: \"@name d1 d2 dn\")"),
            ap.append(),
            ap.nargs(2));
-
+        ap(dyn_param_dims,
+           {"--dyn-input-dim"},
+           ap.help("Dynamic dimensions of a parameter (format: \"@name {min1, max1, [opt1_1, opt1_2, ...]} {min2, max2}, etc.\")"),
+           ap.append(),
+           ap.nargs(2));
+        ap(default_dyn_dim,
+            {"--default-dyn-dim"},
+            ap.help("Default dynamic dimension (format: \"{min, max, [opts]}\")."));
         ap(output_names,
            {"--output-names"},
            ap.help("Names of node output (format: \"name_1 name_2 name_n\")"),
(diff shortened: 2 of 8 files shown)


src/targets/gpu/hip.cpp
```diff
@@ -187,10 +187,28 @@ argument register_on_gpu(const argument& arg)
...
+//{
+//    auto p = write_to_gpu(arg.data(), arg.get_shape().bytes(), host);
+//    return {arg.get_shape(), p};
+//}
+
 argument to_gpu(const argument& arg, bool host)
 {
-    auto p = write_to_gpu(arg.data(), arg.get_shape().bytes(), host);
-    return {arg.get_shape(), p};
+    argument result;
```
umangyadav, line 198: This can be a separate fix : 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1465/files

umangyadav: You can add unit-tests with that.

CharlieL7: Added a unit test. The change is required to run the driver. I could separate it, but it would likely lead to longer review time.

src/driver/main.cpp
```diff
@@ -147,6 +160,67 @@ struct loader
...
+            });
+        return ret;
+    }
+
+    static migraphx::shape::dynamic_dimension parse_dyn_dim_str(std::string dd_str)
+    {
+        // expecting string formatted as "{min, max, {opts}}" or "[min, max]"
+        if(dd_str.empty())
+        {
+            return migraphx::shape::dynamic_dimension{1, 1};
```
umangyadav, line 184: Shouldn't it check if default_dyn_dim is available if so, return that ?

src/driver/main.cpp
```diff
@@ -96,7 +101,15 @@ struct loader
            ap.help("Dim of a parameter (format: \"@name d1 d2 dn\")"),
            ap.append(),
            ap.nargs(2));
-
+        ap(dyn_param_dims,
+           {"--dyn-input-dim"},
+           ap.help("Dynamic dimensions of a parameter (format: \"@name\" \"{min1, max1, [opt1_1, "
```
umangyadav, line 106: Looks like `opts ` be passed as {opt1, opt2, ..} with curly braces instead of square brackets.  We should put that here.

src/driver/main.cpp
```diff
@@ -180,11 +255,25 @@ struct loader
...
+                {
+                    options.default_dyn_dim_value = parse_dyn_dim_str(default_dyn_dim);
+                }
                 options.skip_unknown_operators = skip_unknown_operators;
                 options.print_program_on_error = true;
-                options.map_input_dims         = map_input_dims;
-                p                              = parse_onnx(file, options);
+                if(map_dyn_input_dims.empty())
+                {
+                    options.map_input_dims = map_input_dims;
```
umangyadav, line 270: users can only use one of the other of `map_input_dims` & `map_dyn_input_dims`.   Need to error if they supply both.

src/targets/gpu/hip.cpp
```diff
@@ -187,10 +187,28 @@ argument register_on_gpu(const argument& arg)
     return {s, [p, a = std::move(arg_shared)]() mutable { return get_device_ptr(p.get()); }};
 }
 
+// argument to_gpu(const argument& arg, bool host)
```
umangyadav, line 190: this can just be removed.

src/driver/perf.cpp
```diff
@@ -39,19 +39,27 @@ auto get_hash(const T& x)
...
+            auto s = x.second.to_static(batch);
+            arg    = generate_argument(s, get_hash(x.first));
+        }
         if(not offload)
             arg = t.copy_to(arg);
     }
     return m;
 }
 
+// this function is never used?
```
umangyadav, line 62: Yeah, we can remove it.

src/include/migraphx/stringutils.hpp
```diff
@@ -95,6 +95,14 @@ inline std::vector<std::string> split_string(const std::string& s, char delim)
     return elems;
 }
 
+template <class LeftF, class RightF>
+std::string trim(const std::string& s, LeftF lf, RightF rf)
```
umangyadav, line 99: add test inside `test/stringutils.cpp`.

umangyadav:
> "@ data" "{1, 4}" "{3, 3}" "{224, 224}" "{224, 224}"

is this necessary to be inside quotes ?  for the `--input-dims` we don't use quotes.

CharlieL7:
> > "@ data" "{1, 4}" "{3, 3}" "{224, 224}" "{224, 224}"
> 
> is this necessary to be inside quotes ? for the `--input-dims` we don't use quotes.

The quotes are necessary because of the comma and possible whitespace.

src/include/migraphx/stringutils.hpp
CharlieL7: We don't actually test `stringutils.trim()` it seems.

src/driver/main.cpp
CharlieL7: It can be either curly braces or square brackets.

umangyadav: okay. I would find using  `{}` more convenient. is it possible to add that to `help` ?

src/include/migraphx/stringutils.hpp
umangyadav: Can we add some unit-tests then ?

src/driver/argument_parser.hpp
```diff
@@ -151,7 +151,15 @@ struct value_parser
         T result;
         std::stringstream ss;
         ss.str(x);
-        ss >> result;
+        // handle whitespace in string
+        if constexpr(std::is_same<T, std::string>::value)
```
pfultz2, line 155: Use `{}` instead of `::value`.

src/driver/argument_parser.hpp
```diff
@@ -151,7 +151,15 @@ struct value_parser
...
+        {
+            result = ss.str();
```
pfultz2, line 157: There is no point in constructing a stringstream if we are just returning a string. Just do `return x`.

src/driver/argument_parser.hpp
```diff
@@ -151,7 +151,15 @@ struct value_parser
...
+        }
+        else
+        {
+            ss >> result;
```
pfultz2, line 161: The stringstream should be constructed here in the `else`.

src/driver/main.cpp
pfultz2: I would rather use json for this instead of custom parsing: `@name '[{max:1,min:64,optimals:[1,2,4,8]},3,224,224]'`. Then you can convert it from json with something like:

```cpp
auto v = from_json_string(convert_to_json(s));
std::vector<migraphx::shape::dynamic_dimension> dyn_dims;
std::transform(v.begin(), v.end(), std::back_inserter(dyn_dims), [&](auto x) {
    if (x.is_object()) 
        return from_value<migraphx::shape::dynamic_dimension>(x);
    auto d = x.to<std::size_t>();
    return migraphx::shape::dynamic_dimension{d, d};
});
```

pfultz2: This also allows specifying the fixed dimensions with just one integer.

examples/migraphx/migraphx_driver/README.md
```diff
@@ -29,7 +29,7 @@ See below for a comprehensive list of commands and option arguments, as well as
 | --tf                                     | Load file as a tensorflow graph                           |
 | --migraphx                               | Load file as a migraphx graph                             |
 | --migraphx-json                          | Load file as a migraphx JSON graph                        |
-| --batch                                  | Set batch size for the model                              |
+| --batch                                  | For a static model, set batch size. For a dynamic batch model, sets the batch size at runtime.|
```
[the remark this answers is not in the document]
CharlieL7: We could call this option something else? But yeah, it doesn't just set the batch size.

For the second point, it should be fine as the the `default_dyn_dim` is `{1, 1}` that will be simplified to a static shape if all the dimensions are fixed. So the same behavior as before.

src/driver/main.cpp
```diff
@@ -158,13 +207,39 @@ struct loader
         return output_node_names;
     }
 
+    tf_options set_tf_options()
```
pfultz2, line 210: This should probably be `get_tf_options`.

src/driver/main.cpp
```diff
@@ -158,13 +207,39 @@ struct loader
...
+    {
+        auto map_input_dims    = parse_param_dims(param_dims);
+        auto output_node_names = parse_output_names(output_names);
+        return tf_options{is_nhwc, batch, map_input_dims, output_node_names};
+    }
+
+    onnx_options set_onnx_options()
```
pfultz2, line 217: This should probably be `get_onnx_options`.

src/driver/main.cpp
```diff
@@ -158,13 +207,39 @@ struct loader
         return output_node_names;
     }
 
+    tf_options set_tf_options()
+    {
+        auto map_input_dims    = parse_param_dims(param_dims);
+        auto output_node_names = parse_output_names(output_names);
+        return tf_options{is_nhwc, batch, map_input_dims, output_node_names};
```
pfultz2, line 214: Use the field's name to set it like what you do for `onnx_options`.

examples/migraphx/migraphx_driver/README.md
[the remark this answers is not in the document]
umangyadav: We can keep this  issue for another time, resolving for now.

examples/migraphx/migraphx_driver/README.md
```diff
@@ -44,12 +44,16 @@ See below for a comprehensive list of commands and option arguments, as well as
...
+| --input-dim                              | Set static dimensions of a parameter                      |
+| --dyn-input-dim                          | Set dynamic dimensions of a parameter                     |
+| --default-dyn-dim                        | Set default dynamic dimension                             |
 | --gpu                                    | Compile on the gpu                                        |
 | --cpu                                    | Compile on the cpu                                        |
 | --ref                                    | Compile on the reference implementation                   |
 | --enable-offload-copy                    | Enable implicit offload copying                           |
 | --disable-fast-math                      | Disable fast math optimization                            |
 | --exhaustive-tune                        | Enable exhaustive search to find fastest kernel           |
+| --split-single-dyn-dim                   | Enable split_single_dyn_dim compiler pass                 |
```
pfultz2, line 56: The documentation needs to be updated for these new flags. This is just an example and is not the driver documentation. We should probably either autogenerate this or just link to the driver documentation.

umangyadav (approved):
LGTM

examples/migraphx/migraphx_driver/README.md
CharlieL7: I've forgotten how we handle this. `make doc` generates the documentation, but the generated files are not kept in the repo.

PR #1653, opened 2023-03-28
Fix bug when concatting with the vectorization axis

pfultz2:


test/verify/test_concat_axis_2.cpp
```diff
@@ -33,13 +33,12 @@ struct test_concat_axis_2 : verify_program<test_concat_axis_2>
...
-        auto l0 = mm->add_parameter("x", s0);
-        auto l1 = mm->add_parameter("y", s1);
-        auto l2 = mm->add_parameter("z", s2);
-        mm->add_instruction(migraphx::make_op("concat", {{"axis", 2}}), l0, l1, l2);
+        migraphx::shape s{migraphx::shape::int32_type, {3, 2, 1}};
+        auto x0 = mm->add_parameter("x0", s);
+        auto x1 = mm->add_parameter("x1", s);
+        auto x2 = mm->add_parameter("x2", s);
+        auto x3 = mm->add_parameter("x3", s);
+        mm->add_instruction(migraphx::make_op("concat", {{"axis", 2}}), x0, x1, x2, x3);
```
TedThemistokleous, line 41: Do we want another test instead of modifying the old one?

TedThemistokleous (approved):

umangyadav (approved):

TedThemistokleous:
related to #1283 . Required to get proper passes working on resnext50

test/verify/test_concat_axis_2.cpp
pfultz2: I dont think it makes a difference. Its just 4 parameters so it will hit the vectorization(which should not be applied).

TedThemistokleous: Was more curious your reasoning here and in case we were concerned about breaking or need to still cover some previous functionality. All good.

PR #1655, opened 2023-03-29
expose enum datatypes to python api

shivadbhavsar:
Currently we cannot use the Python API to add instructions that require a datatype attribute (ex. convert). This change is to simply expose the shape::type_t values to be used by the python api and is required by torch_migraphx to support torchbench models.


src/py/migraphx_py.cpp
```diff
@@ -261,6 +262,20 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
...
+        .value("float_type", migraphx::shape::type_t::float_type)
+        .value("double_type", migraphx::shape::type_t::double_type)
+        .value("uint8_type", migraphx::shape::type_t::uint8_type)
+        .value("int8_type", migraphx::shape::type_t::int8_type)
+        .value("uint16_type", migraphx::shape::type_t::uint16_type)
+        .value("int16_type", migraphx::shape::type_t::int16_type)
+        .value("int32_type", migraphx::shape::type_t::int32_type)
+        .value("int64_type", migraphx::shape::type_t::int64_type)
+        .value("uint32_type", migraphx::shape::type_t::uint32_type)
+        .value("uint64_type", migraphx::shape::type_t::uint64_type);
```
pfultz2, line 277: This should be done using `MIGRAPHX_SHAPE_VISIT_TYPES`. I think it would be something like:

```cpp
#define MIGRAPHX_PYTHON_GENERATE_SHAPE_ENUM(x, t) \
    .value(#x, migraphx::shape::type_t::t);

py::enum_<migraphx::shape::type_t>(shape_cls, "type_t")
    MIGRAPHX_SHAPE_VISIT_TYPES(MIGRAPHX_PYTHON_GENERATE_SHAPE_ENUM)
```

umangyadav (approved):

pfultz2 (approved):

PR #1658, opened 2023-03-31
Enable hiprtc by default

pfultz2:
It will still be disabled when using `BUILD_DEV=On`.


TedThemistokleous (approved):

umangyadav (approved):

causten:
Maybe we should hold off until you review why all the tests are failing

umangyadav:
Performance tests are failing because, it requires `MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS=1` flag.

src/targets/gpu/CMakeLists.txt
```diff
@@ -33,7 +33,11 @@ if(NOT TARGET MIOpen)
     message(SEND_ERROR "Cant find miopen")
 endif()
 
-set(MIGRAPHX_USE_HIPRTC OFF CACHE BOOL "Use hipRTC APIs")
+if(BUILD_DEV)
+    set(MIGRAPHX_USE_HIPRTC OFF CACHE BOOL "Use hipRTC APIs")
+else()
+    set(MIGRAPHX_USE_HIPRTC ON CACHE BOOL "Use hipRTC APIs")
```
umangyadav, line 39: Can we  use `rocm_find_program_version` on `amdclang++` ? 

If it is less than 16.0.0  we can `set(ENV{MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS} True)`

pfultz2: That only sets it for cmake time and not runtime. Also, there is performance impact with the workarounds so we don't want to enable it by default.

pfultz2:
> Performance tests are failing because, it requires MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS=1 flag.

Performance tests should build with `-DMIGRAPHX_USE_HIPRTC=Off` since the workarounds do have performance impact.

umangyadav:
> > Performance tests are failing because, it requires MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS=1 flag.
> 
> Performance tests should build with `-DMIGRAPHX_USE_HIPRTC=Off` since the workarounds do have performance impact.

They would fail to compile with hipRTC then.

umangyadav:
> > They would fail to compile with hipRTC then.
> 
> Thats why they should disable it.

You mean we should be using `clang++` as compiler for performance tests ?

pfultz2:
> You mean we should be using clang++ as compiler for performance tests ?

Yes at least until they upgrade to a working hiprtc.

umangyadav:
@causten this one is ready to go. Performance tests are working.

PR #1666, opened 2023-04-04
conv dot fusion non standard reshape bugfix

shivadbhavsar:
Updated logic for when contiguous should be added to the slices generated by find_conv_dot_horiz_fusion. Found an edge case in huggingface model where the set of ops:
``` dot -> dot -> add -> reshape```
would be fused to:
``` concat -> dot -> slice -> slice -> add -> reshape ```

However, if the add is a unit op (removed by find_unit_ops), this leads to a reshape on a non standard output. See newly added test case for concrete example.

### CODE REVIEW

File: src/simplify_algebra.cpp [hunk 1]
```diff
@@ -928,21 +928,26 @@ struct find_conv_dot_horiz_fusion
...
-                {
-                    if(output->name() != "reshape")
-                        continue;
-                    auto x = m.insert_instruction(output, make_op("contiguous"), arg);
-                    m.replace_instruction(output, output->get_operator(), x);
-                }
+                auto outputs             = arg->outputs();
+                auto requires_contiguous = std::any_of(outputs.begin(), outputs.end(), [](auto o) {
+                    return o->get_shape().standard();
+                });
```
CharlieL7, line 934: This doesn't look to be a good check since the output might be a standard shape even if the next instruction does not need a standard shape. We should instead check if the next instruction requires a standard shape. We probably need a new matcher that checks the operation `attributes()` for `"require_std_shape"`. I'm not too familiar with the matcher syntax though.

CharlieL7: Actually, I believe you can just get the next instruction, convert it to a value, and then check the `attributes`.

shivadbhavsar: Ya I agree, the idea behind the loose check was that the eliminate_contiguous pass will always take care of the unnecessary contiguous ops added. Let me see if doing it based on that attribute works, that sounds better

shivadbhavsar: Checking for the "require_std_shape" attribute fails for the new test case I added

umangyadav:
> However, if the add is a unit op (removed by find_unit_ops), this leads to a reshape on a non standard output 

We better fix this in `find_unit_ops` in addition to what you have `find_conv_dot_horiz`.  Add, Mul, Div has the side effect of making output contiguous. This bug could appear in other places as well.

pfultz2:
> We better fix this in find_unit_ops in addition to what you have find_conv_dot_horiz. Add, Mul, Div has the side effect of making output contiguous. This bug could appear in other places as well.

Thats why we should make reshape a copy instead of trying to hunt this down everywhere.

causten:
This is killing performance in a few models.  Please examine

shivadbhavsar:
Closing PR, we should take care of reshape issues in a more general way mentioned in #1677

PR #1668, opened 2023-04-05
Blas auto-tuning for GEMMs

bpickrel:
This recreates PR 1585, which was spoiled by a corrupted git merge.

Implements a new feature in the rocBLAS API to tune GEMM solutions automatically. The finalize() method calls a new tune() method which simply iterates through the list of available solutions, timing them all and choosing the fastest.

At the time of creating this PR, these API calls are a Beta-only feature and are not available in regular ROCm releases. It is enabled in the Rocm ver 5.5 API by using preprocessor value ROCBLAS_BETA_FEATURES_API, which is set by CMakeLists.txt if the needed rocBLAS version is installed at compile time.

See https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1553


src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,198 +95,515 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
-template <class R, class... Ts, class... Us>
-R rocblas_invoke(R (*f)(Ts...), Us... xs)
+/**
+ * Returns results of rocblas_status_success, rocblas_status_perf_degraded,
+ * or rocblas_status_invalid_value.  Caller
+ * is expected to check for invalid index.  Any other result causes an exception.
+ *
+ */
+template <class F, class Pack, class... Ts>
+auto rocblas_invoke(F f, Pack p, Ts... xs)
```
bpickrel, line 105: @pfultz2 this PR code no longer checks the status returned by rocblas_invoke(). Do we need to retain the ability to check for degraded performance, etc?

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -163,6 +167,20 @@ struct miopen_convolution
...
         std::size_t workspace_size = 0;
+
+        // TODO:  Using an environment variable to disable convolution tuning while still allowing
+        // other (i.e. GEMM) tuning with --exhaustive-tune, is a workaround for a problem in which
+        // convolution tuning takes multiple hours.
+        const bool convo_tune_enabled = enabled(MIGRAPHX_ENABLE_CONV_TUNING{});
+        if(ctx.get_exhaustive_tune_flag() and not convo_tune_enabled)
+        {
+            std::cerr
+                << "WARNING: MIGraphX convolution tuning not enabled. To run tuning, set both the "
```
bpickrel, line 178: This causes a flood of warning messages. Do we have a tidy way to limit this message to only output once per compilation?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,198 +95,515 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+        {
+            // Define the function to be timed
+            auto run_func = [&]() {
+                run(ctx, input_args, sol);
+                ctx.finish();
+            };
+
+            // Warmup: the first few calls to an op. may not be representative since there is
+            // more time taken initializing caches, etc. so we won't time them.
+            for(int cc = 0; cc < cold_calls; ++cc)
```
bpickrel, line 456: todo:  I reduced cold calls to only 1 in response to a comment from Paul, but I haven't fully satisfied myself that more iterations wouldn't help performance.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,198 +95,515 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+            {
+                run_func();
+            }
+            double host_time = 0.0;
+
+            for(int hc = 0; hc < hot_calls; ++hc)
+            {
+                ctx.finish();
+                host_time += time<microseconds>(run_func);
```
bpickrel, line 465: Comments transferred from the old PR:

[pfultz2](https://github.com/pfultz2) [last month](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1585#discussion_r1127103617)
This should just use the time_op function from the driver. We just need to move it to the gpu lib so it can be reused here.


@[bpickrel](https://github.com/bpickrel) bpickrel [last month](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1585#discussion_r1128807853)
You'll have to specify what you have in mind for that. gemm_impl doesn't use instances of the Migraphx operation, which is what time_op takes as an argument. Instead, we're making the rocblas calls directly. So we can't reuse it directly--what would we substitute here? (Also, does this change belong in the same PR?)

src/targets/gpu/include/migraphx/gpu/convolution.hpp
pfultz2: This should be removed, we want to enable miopen tuning by default with the exhaustive flag. If the user only wants to tune miopen then they can set the find enforce env variables.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
```diff
@@ -163,6 +167,20 @@ struct miopen_convolution
         auto w_desc                = make_tensor(reshape_if_1d(inputs[1]), int8_x4_format);
         auto y_desc                = make_tensor(reshape_if_1d(output_shape));
         std::size_t workspace_size = 0;
+
+        // TODO:  Using an environment variable to disable convolution tuning while still allowing
+        // other (i.e. GEMM) tuning with --exhaustive-tune, is a workaround for a problem in which
```
pfultz2, line 172: It would be better to just add a env variable to enable GEMM tuning even when the exahustive flag is not passed.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,198 +95,515 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+template <class F, class Pack, class... Ts>
+auto rocblas_invoke(F f, Pack p, Ts... xs)
 {
-    if constexpr(sizeof...(Ts) == sizeof...(Us))
-        return f(xs...);
-    else
-        return f(xs..., nullptr, nullptr);
+    return p([=](auto... ws) {
+        auto status = f(ws..., xs...);
+        if(status != rocblas_status_success and status != rocblas_status_invalid_value)
```
pfultz2, line 109: Why are we skipping the invalid value error?

src/targets/gpu/gemm_impl.cpp
pfultz2: Dont worry about using time_op for now.

src/targets/gpu/gemm_impl.cpp
pfultz2: Just remove the loop since this is only called once.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,198 +95,515 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+            auto common_sol_args = create_gemm_ex_args_common(ctx, input_args);
+            rocblas_invoke(&rocblas_gemm_ex_get_solutions,
+                           common_sol_args,
+                           rocblas_gemm_algo_solution_index,
+                           int8_flag,
+                           solution_indices.data(),
+                           &list_size);
+        }
+
+        double bestTime   = std::numeric_limits<double>::max();
```
pfultz2, line 442: Use C++ naming conventions for variables.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,198 +95,515 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+                      float alpha,
+                      float beta,
+                      bool int8_x4_format,
+                      bool compute_fp32,
+                      int32_t solution_idx)
 {
-    gemm_impl(ctx, output_shape, args, alpha, beta, int8_x4_format, compute_fp32);
+#ifdef ROCBLAS_BETA_FEATURES_API
+
+    if(ctx.get_exhaustive_tune_flag() && solution_idx == 0)
```
pfultz2, line 553: Use `and` instead of `&&`.

src/targets/gpu/gemm_impl.cpp
bpickrel: Because the validate() method looks at the result.  My earlier comment that we never look at the return value from this function was in error.

src/targets/gpu/gemm_impl.cpp
bpickrel: done

src/targets/gpu/gemm_impl.cpp
bpickrel: done

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,198 +100,516 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+                      int32_t alpha,
+                      int32_t beta,
+                      bool int8_x4_format,
+                      bool compute_fp32,
+                      int32_t solution_idx)
 {
-    gemm_impl(ctx, output_shape, args, alpha, beta, int8_x4_format, compute_fp32);
+#ifdef ROCBLAS_BETA_FEATURES_API
+
+    if((enabled(MIGRAPHX_ENABLE_GEMM_TUNING{}) or ctx.get_exhaustive_tune_flag()) and
```
bpickrel, line 592: This isn't going to work...working on a correction

bpickrel: Fixed by commit a82ea1.  The `if((enabled...` logic is moved to gemm.hpp and gemm_impl.hpp.

src/targets/gpu/include/migraphx/gpu/convolution.hpp
bpickrel: I've refactored this so that instead of specially DISabling tuning for convolutions, user can specially ENable it for GEMMs by setting MIGRAPHX_ENABLE_GEMM_TUNING.  See  src/targets/gpu/include/migraphx/gpu/gemm.hpp:142

src/targets/gpu/include/migraphx/gpu/convolution.hpp
bpickrel: Done.  I've refactored this so that instead of specially DISabling tuning for convolutions, user can specially ENable it for GEMMs by setting MIGRAPHX_ENABLE_GEMM_TUNING.  See  src/targets/gpu/include/migraphx/gpu/gemm.hpp:142

bpickrel:
@pfultz2 has volunteered to troubleshoot the performance problems.

src/targets/gpu/CMakeLists.txt
```diff
@@ -284,7 +284,7 @@ else()
 endif()
 
 if(HAS_ROCBLAS_BETA_FEATURES_API)
-    target_compile_definitions(migraphx_gpu PUBLIC -DROCBLAS_BETA_FEATURES_API -DROCBLAS_NO_DEPRECATED_WARNINGS)
+    target_compile_definitions(migraphx_gpu PUBLIC -DMIGRAPHX_USE_ROCBLAS_TUNING_API -DROCBLAS_BETA_FEATURES_API -DROCBLAS_NO_DEPRECATED_WARNINGS)
```
bpickrel, line 287: Should DROCBLAS_BETA_FEATURES_API be removed from this list now?  Looks like it's been eliminated?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -123,13 +123,13 @@ auto rocblas_invoke(F f, Pack p, Ts... xs)
 
 static bool is_transposed(const shape& s) { return s.transposed() and s.strides().back() != 1; }
 
-static rocblas_int get_batch_stride(const argument& a)
+static rocblas_int get_batch_stride(const shape& s)
```
bpickrel, line 126: Did you say you were going to make this explicit?

bpickrel: Ah, I see it now.  A different PR.  Which will have to be merged into this one to make it work right.

pfultz2: The explicit constructor is not needed for this PR. I only added so we can catch these mistakes on compile-time in the future.

src/targets/gpu/CMakeLists.txt
pfultz2: No its still needed to use the API from rocblas.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,196 +96,531 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
-        return f(xs...);
-    else
-        return f(xs..., nullptr, nullptr);
+    return p([=](auto... ws) {
+        auto status = f(ws..., xs...);
+        if(status != rocblas_status_success and status != rocblas_status_invalid_value)
+        {
+            if(status == rocblas_status_perf_degraded)
+            {
+                std::cerr << "WARNING: degraded perf. in rocBLAS call" << std::endl;
```
umangyadav, line 114: when do we expect to see perf degraded status  ? if it is in the case when rocblas is not tuned, then can we print message for user to enable tuning ?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,196 +96,531 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
 
-        auto out_lens   = output_shape.lens();
-        rocblas_int m   = out_lens[dim_0];
-        rocblas_int n   = out_lens[dim_1];
-        rocblas_int k   = args[0].get_shape().lens()[dim_1];
-        auto to_pointer = [&](auto&& arg) { return as.from(arg.data()); };
-        if(args[0].get_shape().type() == shape::int8_type and (k % 4) != 0 and int8_x4_format)
+#if ROCBLAS_VERSION_MAJOR < 3
+        int8_flag = int8_x4_format ? rocblas_gemm_flags_pack_int8x4 : rocblas_gemm_flags_none;
+#endif
```
umangyadav, line 205: Would it be better to use this ? https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/0a4254a2af2da7a1d4e380e9adf093e94e4e8bc2/src/targets/gpu/include/migraphx/gpu/rocblas.hpp#L43

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,196 +96,531 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+        auto b_lens = input_shapes[1].lens();
+
+        auto out_lens = output_shape.lens();
+        m             = out_lens[dim_0];
+        n             = out_lens[dim_1];
+        k             = input_shapes[0].lens()[dim_1];
+        if(input_shapes[0].type() == shape::int8_type and (k % 4) != 0 and int8_x4_format)
         {
-            MIGRAPHX_THROW("ROCBLAS_GEMM: k size of int8 type input must be mutlple of 4!");
+            MIGRAPHX_THROW("ROCBLAS_GEMM: k size of int8 type input must be multiple of 4!");
```
umangyadav, line 216:
```suggestion
            MIGRAPHX_THROW("ROCBLAS_GEMM: k size of int8 type input must be multiple of 4");
```
I dont' know why there is an exclamation mark. I read it as factorial.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,196 +96,531 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+            run(ctx, input_args, sol);
+            double host_time = time<milliseconds>([&] {
+                for([[maybe_unused]] int hc : range(hot_calls))
+                    run(ctx, input_args, sol);
+                ctx.finish();
+            });
+
+            // todo:  Measured time dropped from 20 us to about 6.7 us when I raised hot_calls from
+            // 1 to 11. The higher the hot_calls value, the faster per-call time up to at least 25,
+            // and increasing cold_calls makes little or no difference.  Why?
```
umangyadav, line 461: Does it need these comments ?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,196 +96,531 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+                      int32_t alpha,
+                      int32_t beta,
+                      bool int8_x4_format,
+                      bool compute_fp32,
+                      int32_t solution_idx)
+{
+#ifdef MIGRAPHX_USE_ROCBLAS_TUNING_API
+
+    // This code should be called only if either the environment var.
+    // MIGRAPHX_ENABLE_GEMM_TUNING, or option --exhaustive-tune, is set
```
umangyadav, line 603: Are these comments relevant here?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,196 +96,531 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+    {
+        if(strided_batched)
+        {
+            auto common_args = create_strided_batched_args_common(ctx, input_args);
+            rocblas_invoke(&rocblas_gemm_strided_batched_ex,
+                           common_args,
                            rocblas_gemm_algo_standard,
-                           0,
-                           flag);
+                           solution_idx,
```
umangyadav, line 243: Does it need to use `rocblas_gemm_algo_solution_idx` flag ?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -21,15 +21,30 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+
+/**
+ * Contains a templated struct implementation that wraps several rocBLAS API calls
+ * used by the GEMM operator.  These are accessed by methods declared in gemm_impl.hpp
+ *
```
umangyadav, line 28: These comments doesn't seems relevant here.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -81,196 +96,531 @@ shape transpose_batch(const shape& s, unsigned trans_batch)
...
+            });
+
+            // todo:  Measured time dropped from 20 us to about 6.7 us when I raised hot_calls from
+            // 1 to 11. The higher the hot_calls value, the faster per-call time up to at least 25,
+            // and increasing cold_calls makes little or no difference.  Why?
+            host_time /= hot_calls;
+
+            // dev/evaluation only: track time for first solution.
+            if(first_time < 0)
+                first_time = host_time;
```
umangyadav, line 466: Why do we need to track time for the first solution ?

causten:
@pfultz2  can you resolve the conflicts and change the reviewers so we can get this merged

causten:
Format is claiming an error @pfultz2

causten:
@pfultz2 accuracy failure in Jenkins job

src/targets/gpu/gemm_impl.cpp
pfultz2: According to the docs it means: 

> indicates that a slower algorithm was used because of insufficient device memory for the optimal algorithm

It not related to tuning.

src/targets/gpu/gemm_impl.cpp
pfultz2: So we can show how much speedup there was from tuning.

bpickrel: When I originally did this, the measured speed during the test run seemed to vary with the number of "warm-up" solutions I gave it; it did not simply come up to full speed with one solution performing all the memory initialization, and this code was to help look for the optimum.  Since then I think we decided to just assume only one warm-up was needed.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -78,6 +79,8 @@ struct rocblas_gemm
     {
         std::vector<shape> in_shapes(inputs);
         in_shapes.pop_back();
+        // When input shapes are A, B, C the GEMM equation is  C  =  α AB+ β C   where α, β are
```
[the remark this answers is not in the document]
bpickrel: I pasted them in.  I don't recall whether I got the originals from a Web page, or from the alternate character sets in Microsoft Word.

causten:
@pfultz2 unet model failing, agent model not passing accuracy

pfultz2:
Those are the same failures on develop.

pfultz2:
Also, the unet bug is the same on develop, but we dont catch it since we dont check the return of rocblas call. The fix will very likely fix performance because we arent even calling a gemm due to the rocblas error.

umangyadav (approved):

PR #1669, opened 2023-04-05
Fix 2 input broadcast bug for dynamic batch and output parameter ordering

CharlieL7:
* Adds a matcher to `split_single_dyn_dim` to find all `broadcast` or `multibroadcast` with two static shape inputs and replaces the instruction with the one input version.
* Sorts the `get_output_parameters()` list to ensure the correct ordering. (Was getting an error for some models.) 
  * There's a possibility this sorting might not work properly if we have >9 parameters?


src/split_single_dyn_dim.cpp
```diff
@@ -67,6 +69,33 @@ has_one_dyn_dim(const std::unordered_map<std::string, shape>& param_shapes)
...
+    }
+
+    void apply(module& m, const match::matcher_result& mr) const
+    {
+        auto ins          = mr.result;
+        auto out_lens     = ins->get_shape().lens();
+        auto broadcast_op = ins->get_operator();
+        if(broadcast_op.name() == "broadcast")
+        {
+            broadcast_op.from_value({{"out_lens", out_lens}});
```
TedThemistokleous, line 89: Add case for then branch of If

TedThemistokleous (approved):

TedThemistokleous:
> * Adds a matcher to `split_single_dyn_dim` to find all `broadcast` or `multibroadcast` with two static shape inputs and replaces the instruction with the one input version.
> 
>     * Sorts the `get_output_parameters()` list to ensure the correct ordering. (Was getting an error for some models.)
>       
>       * There's a possibility this sorting might not work properly if we have >9 parameters?

Why 9 parameters?  Not sure I understand why that's the case

CharlieL7:
> Why 9 parameters? Not sure I understand why that's the case

The string sort when the output parameter is #output_10 will probably order it as #output_0, #output_1, #output_10, #output2, ...

src/split_single_dyn_dim.cpp
```diff
@@ -120,6 +150,8 @@ void split_single_dyn_dim::apply(module_pass_manager& mpm) const
         }
         mm->replace_return(outputs);
     }
+
+    migraphx::run_passes(*mm, {migraphx::dead_code_elimination{}});
```
pfultz2, line 154: You can do `mpm.run_pass(dead_code_elimination{})`, but why dont we just add this pass to target.cpp?

src/split_single_dyn_dim.cpp
```diff
@@ -67,6 +69,33 @@ has_one_dyn_dim(const std::unordered_map<std::string, shape>& param_shapes)
                                     dds_it->max};
 }
 
+struct find_static_2in_broadcasts
```
pfultz2, line 72: Wrap this in an anonymous namespace.

src/split_single_dyn_dim.cpp
CharlieL7: I don't understand

pfultz2: I think he means add a test case for this line.

src/split_single_dyn_dim.cpp
```diff
@@ -23,11 +23,13 @@
  */
 
 #include <migraphx/split_single_dyn_dim.hpp>
+#include <migraphx/dead_code_elimination.hpp>
```
umangyadav, line 26: This is not used. Remove it.

src/include/migraphx/op/select_module.hpp
```diff
@@ -57,6 +57,7 @@ struct select_module
                      param_names.cend(),
                      std::back_inserter(ret),
                      [](auto pn) { return not contains(pn, "#output_"); });
+        std::sort(ret.begin(), ret.end());
```
umangyadav, line 60: Do we know exact reason why it requires sorting ? 

`get_parameter_names()` returns stable_sorted parameter list already. In other places we are using a map so order should be irrelvanat. 

Only place i see is during `compute()` of `select_module`.  `std::vector<argument> args` has fixed order. 

But `select_module` itself is added by split_single_dyn_dim pass which again uses maps and stable_sorts already. So I don't see where this problem is coming from.

CharlieL7: AFAIK, the parameters from `module::get_parameter_names()` are sorted by the parameter order. Which for a reason unknown to me will sometimes reorder the output parameters so they are not in name order. The  `select_module` op created by `split_single_dyn_dim` expects the output parameters to be in name order (`#output_0, #output_1, ...`) to match the order of arguments to `select_module`. The bug I would see is that a dynamic batch model would fail to run for certain batch sizes because the arguments and parameters were mismatched.

pfultz2 (approved):

TedThemistokleous (approved):

PR #1671, opened 2023-04-06
Updates to README

pfultz2:
This removes the `--cxx` flags from the rbuild commands since it is not necessary. Also added a section about using rbuild to set up an environment  for development.


README.md
```diff
@@ -184,8 +180,20 @@ target_link_libraries(myApp migraphx::c)
...
+Using rbuild, the dependencies for development can be installed with:
+
+```
+rbuild develop
+```
+
+This will install the dependencies for development into the `deps` directory and
+configure `cmake` to use those dependencies in the `build` directory. These
+directories can be changed be changed by passing the `--deps-dir` and
+`--build-dir` flags to `rbuild` command.
```
umangyadav, line 194: can we just put one line of example with `--dep-dir` and `--build-dir` ?

umangyadav (approved):

CharlieL7 (approved):

TedThemistokleous (approved):

Issue #1656, opened 2023-03-30
[RFC] Multi-target compilation in MIGraphX


umangyadav:
### Motivation : 

MIGraphX can support multiple backend targets. It fully supports GPU target currently. It can target the CPU backend with limited functionality and FPGA support is in the BETA mode. Currently, MIGraphX can compile models for each of those individual backend targets. 

With a larger UIF common runtime goal, MIGraphX is a candidate to become a unified frontend for all the AMD (or even Intel/Nvidia) backends that includes FPGA, GPU, and CPUs. The goal is not limited to a single backend but it also should consider heterogeneous systems. That is MIGraphX should be able to compile models for combinations of different targets i.e. CPU + GPU, FPGA + CPU, GPU + FPGA,  FPGA + GPU + CPU, etc. 

We would name this heterogeneous compilation a `multi-target` compilation for the rest of the discussion. 

### Primer on IR structure of MIGraphX : 

Deep learning models from ONNX/TF are parsed into MIGraphX program. Each program consists of one `main` module and optionally several submodules. Modules can be also be nested. `main` is the root of module tree. 

Each module is made of up `instructions`. Each instruction is a "computation operator" or "node" in the computational graph.   Each instruction consumes other instructions as inputs thereby establishing a computational graph.

Modules are a grouping of instructions to aid in the compilation of programs. Modules are useful for defining control flow as well e.g. for the loop, IfThenElse operators. Another example is the pointwise module which helps in producing JIT kernels. 

### Pass manager : 
MIGraphX has two types of passes. 
	1. module pass 
	2. program pass.  

Module pass only runs pass on module v/s program pass applies the pass on the entire program.  

Currently, only the DeadCodeElimination is a program pass. The rest of the passes are module passes but they run on the main module so essentially making them program passes. 

### Multi-target Compilation :

`multi-target` compilation feature implementation can be broken down into the following few pieces :
	1. Graph partitioning 
		a. Target annotation
		b. (Proposed) Merge target regions 
	2. Subgraph compilation 

### Graph-partitioning : 

MIGraphX currently has a placeholder code for this process. 

The way it works right now is that : 

1. Each target would implement the `is_supported` function which would return True/False if it has an implementation for a particular instruction or not. 
2. Each target would also implement `find_supported` which would return `set<instruction>` from entire program for which it has support. 
[...]

PR #1672, opened 2023-04-06
Update pass manager to handle multi-target compilation

umangyadav:
partially solves https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1656
This PR only handles compilation part of multitarget. 

Need to add `TRACE_COMPILE` flag specific to single target. 

The update for the `eval()` is for the next PR. Update for the API is due for the next PR. 

~Need to change CI to have a stage with all targets enabled~
 added stage in Jenkins.


CharlieL7 (commented):
I don't fully understand the design here.
- What does `run_on_target` intended to do?
- Does the run() passed to `run_on_target.compute()` do something different on different targets?

src/program.cpp
```diff
@@ -205,8 +208,70 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
 
 bool program::is_compiled() const { return not this->impl->target_name.empty(); }
 
+void program::compile(const std::vector<std::string>& targets,
```
pfultz2, line 211: This should take a `std::vector<target>` not string names of target.

src/program.cpp
```diff
@@ -205,8 +208,70 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+                      std::unordered_map<std::string, compile_options> compile_opt_map)
```
pfultz2, line 212: This can probably be `std::vector<compile_options>`

src/include/migraphx/literal.hpp
```diff
@@ -50,6 +50,8 @@ struct literal : raw_data<literal>
      */
     explicit literal(shape::type_t shape_type) : m_shape(shape_type, {}) {}
 
+    explicit literal(const shape& s) : m_shape(s) {}
```
pfultz2, line 53: Why is this added? A `literal` should always have a buffer set when the shape is not empty.

src/include/migraphx/module.hpp
```diff
@@ -69,6 +69,8 @@ struct module
 
     std::string name() const;
 
+    std::string get_target() const;
+    void set_target(const std::string& target_name);
```
pfultz2, line 73: This should be removed. Its not used, and doesnt make sense on modules that are for multiple targets.

src/program.cpp
```diff
@@ -205,8 +208,70 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+    trace(*this);
+    trace();
+
+    // Run passes on each root target
+    for(const auto& t : targets)
+    {
+        auto root_target           = migraphx::make_target(t);
+        auto root_modules_range    = roots.equal_range(t);
+        this->impl->context_map[t] = root_target.get_context();
+        for(auto i = root_modules_range.first; i != root_modules_range.second; i++)
```
pfultz2, line 245: Use `for(auto i:range(root_modules_range))`

src/pass_manager.cpp
```diff
@@ -41,6 +41,12 @@ inline namespace MIGRAPHX_INLINE_NS {
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_PASSES);
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TIME_PASSES);
 
+template <class T>
+std::vector<T*> generic_get_modules(T* mm);
+
+template <>
+std::vector<module*> generic_get_modules(module* mm);
```
pfultz2, line 48: Yea I dont think we should put specialization for these template functions. Actually, its much shorter to just push the current module to the front:

```cpp
auto mods = root_mod->get_submodules();
mods.insert(mods.begin(), root_mod);
```

src/pass_manager.cpp
```diff
@@ -116,34 +126,25 @@ struct module_pm : module_pass_manager
         {
             p.apply(*this);
         }
+        trace("Pass: ", p.name());
```
pfultz2, line 129: This shouldn't be added here. #1667 already adds a print out of the pass name.

src/program.cpp
```diff
@@ -205,8 +208,70 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+        {
+            auto* current_mod = i->second;
+            i->second->set_target(t);
+            run_passes(*this,
+                       current_mod,
+                       root_target.get_passes(this->impl->context_map[t], compile_opt_map[t]),
```
pfultz2, line 251: So we should append a pass to mark each of the instructions with the target_id:

```cpp
struct mark_instruction_target
{
    std::size_t target_id;
    void apply(module& m) const
    {
        for(auto& ins:m)
            ins.set_target_id(target_id);
    }
};

auto passes = root_target.get_passes(this->impl->context_map[t], compile_opt_map[t])
passes.push_back(mark_instruction_target{id});
```

test/CMakeLists.txt
```diff
@@ -187,6 +187,25 @@ if(MIGRAPHX_ENABLE_PYTHON)
     add_subdirectory(py)
 endif()
 
+# multitarget test
+if(MIGRAPHX_ENABLE_GPU AND MIGRAPHX_ENABLE_CPU AND MIGRAPHX_ENABLE_FPGA)
```
pfultz2, line 191: Since FPGA is still missing it would be better to have tests without needing this enabled for now.

We could also have unit tests that can run when no target is enabled. We can just create different targets for unit tests, but maybe that is too much work.

test/multi_target/multitarget_test.cpp
```diff
@@ -0,0 +1,447 @@
...
+    run_ref_mod->add_return({run_ref_ins});
+
+    std::vector<migraphx::shape> sub_shapes = {};
+    sub_shapes.push_back(migraphx::shape{migraphx::shape::float_type, {{1, 4}, {4, 4}}});
+    sub_shapes.push_back(migraphx::shape{migraphx::shape::float_type, {{1, 4}, {4, 4}}});
+    migraphx::shape out_attr = migraphx::shape{sub_shapes};
+    auto sm_ins              = mm->add_instruction(
+        migraphx::make_op("select_module", {{"output_dyn_shapes", migraphx::to_value(out_attr)}}),
+        {input},
+        {run_cpu_mod, run_gpu_mod, run_fpga_mod, run_ref_mod});
```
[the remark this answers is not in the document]
umangyadav: Added parameter for each module such that it would work with `select_module`.

src/program.cpp

src/pass_manager.cpp
```diff
@@ -106,6 +112,10 @@ struct module_pm : module_pass_manager
         trace("Pass: ", p.name());
         assert(mod);
         assert(mod->validate() == mod->end());
+        if(mod->bypass())
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
umangyadav: Removed.

src/include/migraphx/module.hpp
umangyadav: Those target annotations are only for the root module. If it contains further nested submodules that are for different targets, then those modules themselves would have a `target` attribute.

src/program.cpp
```diff
@@ -205,8 +208,70 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
 
 bool program::is_compiled() const { return not this->impl->target_name.empty(); }
 
+void program::compile(const std::vector<std::string>& targets,
+                      std::unordered_map<std::string, compile_options> compile_opt_map)
+{
+    // Gather all the target roots
+    std::unordered_multimap<std::string, module_ref> roots;
```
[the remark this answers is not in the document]
umangyadav: I find it better to keep it `<std::string, module_ref>` rather than referencing different targets by index.  Atleast for now `compile()`.  We keep `target` as std::string for each root module so that makes it better.

src/program.cpp
umangyadav: That would require knowing the index of the target each time we reference `compile_options`.

src/pass_manager.cpp
umangyadav: Done

src/include/migraphx/literal.hpp
umangyadav: It was required to save/load `hip::copy_literal` as part of the test inside `test/operators.cpp`.  It wasn't linking any backend libraries so, it was only testing core MIGraphX Ops. As part of this PR, I made it to link `migraphx_all_targets` and it started testing all backend ops as wel. 

During serialization literals only generate `data` field if  it not empty but during de-serialization it expects "data"  field to be present and it would fail for `hip::copy_literal` save/load test since it has empty literal.

src/program.cpp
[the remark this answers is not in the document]
pfultz2: > I find it better to keep it <std::string, module_ref> rather than referencing different targets by index.

But if two targets have the same name then we can't distinguish them.

src/program.cpp
pfultz2: If the `roots` is `std::unordered_multimap<std::size_t, module_ref>` then you should have the index when you iterate the roots.

src/program.cpp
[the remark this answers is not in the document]
umangyadav: Yes, but right now `make_target()` only construct target based on string target_name input. So if targets are different then, their names must be different at the moment.

test/CMakeLists.txt
umangyadav: > We can just create different targets for unit tests, 

Yeah. That would require creating some example passes for that target as well. 

>  when no target is enabled 

I have unit-test with single target and also have unit-test for the `run_on_target()` operation.

src/program.cpp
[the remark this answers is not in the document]
umangyadav: Made it `map<size_t, module_ref>`.

src/program.cpp

src/program.cpp
```diff
@@ -59,6 +61,7 @@ struct program_impl
     std::unordered_map<std::string, module> modules;
     context ctx;
     std::string target_name;
+    std::vector<context> contexts;
```
pfultz2, line 64: So from #1751 I also store the targets(and I remove the target_name), but this requires updates to the serialization.

Also this is missing updates to the `assign` method.

src/include/migraphx/module.hpp
pfultz2: But `get_target` is never used anywhere(except in the unit test), we should add it only if we need it.

src/program.cpp
```diff
@@ -205,8 +219,78 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+    {
+        auto root_target_name                 = root_target.name();
+        auto root_target_idx                  = target_idx(root_target_name);
+        auto root_modules_range               = roots.equal_range(root_target_idx);
+        this->impl->contexts[root_target_idx] = root_target.get_context();
+        for(const auto& [id, current_mod] : range(root_modules_range))
+        {
+            auto passes = root_target.get_passes(this->impl->contexts[root_target_idx],
+                                                 compile_opt_map[target_idx(root_target_name)]);
+            passes.push_back(mark_instruction_target{static_cast<unsigned int>(root_target_idx)});
```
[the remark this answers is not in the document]

src/include/migraphx/op/run_on_target.hpp
```diff
@@ -0,0 +1,98 @@
...
+                                  std::vector<migraphx::module_ref> mods) const
+    {
+        if(mods.size() != 1)
+        {
+            MIGRAPHX_THROW("RUN_ON_TARGET: must have exactly 1 module argument");
+        }
+        auto* mod_input = mods.front();
+        if(inputs.size() != mod_input->get_parameter_shapes().size())
+        {
+            MIGRAPHX_THROW("RUN_ON_TARGET: Number of input paramters mismatches");
```
kahmed10, line 62:
```suggestion
            MIGRAPHX_THROW("RUN_ON_TARGET: Mismatched number of input parameters");
```

src/instruction.cpp
```diff
@@ -406,6 +406,9 @@ void instruction::print(std::ostream& os,
     // skip return instruction shape
     if(ins->name() != "@return")
         os << " -> " << ins->get_shape();
+    // print tid
+
+    os << ", tid=" << ins->target_id;
```
kahmed10, line 411: Can we just expand it so there's no ambiguity?
```suggestion
    os << ", target_id=" << ins->target_id;
```

src/program.cpp
```diff
@@ -205,8 +219,95 @@ target_assignments program::get_target_assignments(const std::vector<target>& ta
...
+        this->impl->contexts.resize(targets.size());
+        compile_opts.resize(targets.size(), migraphx::compile_options{});
+    }
+    // mark all the instruction as ref target first, later change target_id based on root-target
+    run_passes(*this, {mark_instruction_target{ref_target_id}});
+
+    // Run passes on each root target
+    for(const auto& root_target : targets)
+    {
+        auto root_target_name                = root_target.name();
```
umangyadav, line 278: I still need to make use of `name()`.  We need to update target class to add `==` operator so that we can search through `vector<target>`

kahmed10 (approved):

src/program.cpp

src/module.cpp
```diff
@@ -53,8 +53,9 @@ struct module_impl
     std::list<instruction> instructions;
     std::unordered_set<instruction*> instruction_set;
     std::string name;
-    uint32_t nparams = 0;
-    bool bypass      = false;
+    uint32_t nparams   = 0;
+    bool bypass        = false;
+    std::string target = "ref";
```
pfultz2, line 58: I dont think this is used anywhere.

CharlieL7 (approved):

PR #1675, opened 2023-04-06
[mlir] Adding quantizelinear, dequantizelinear and quant_convolution support

jerryyin:
This pull request updated the commit hash that allows the 3 ops to be supported on the individual op level from migraphx:
 - quant_convolution
 - quantizelinear
 - dequantizelinear

This unit test is passing that ensures MLIR's implementation is aligned with reference implementation.

Note: fusion of quant_convolution together with pointwise ops will come in successive PRs.


src/targets/gpu/mlir.cpp
```diff
@@ -483,7 +487,7 @@ struct mlir_program
     static value get_operator_value(const operation& op)
     {
         auto v = op.to_value();
-        if(op.name() == "convolution")
+        if(op.name() == "convolution" || op.name() == "quant_convolution")
```
pfultz2, line 490: Use `or` instead of `||`.

jerryyin: Sounds good, done in 9de43f8.

manupak (approved):
I have no concerns here...  LGTM.

pfultz2 (approved):

turneram (approved):

PR #1680, opened 2023-04-12
Add clean up for CI cache and change keys for caching

umangyadav:
This PR updates current `ci.yaml` file to do the following things 

1. Update cache keys to not use timestamps. Using timestamps in the primary key would always result in cache misses. The cache could be recovered by using restore-keys but that cache would be considered "stale" and GH Actions would do a fresh caching with a timestamp as the primary key. This loop would continue running resulting in bloating of caches.

2. Use`actions/cache@v3` instead of `always-upload-cache`.  Not sure why it was used earlier, But Fork mentions two issues that have already been resolved in main `actions/cache`. Also `actions/cache` seems to be ahead of `always-upload-cache` by several commits. 

3. Use branch name as part of the primary cache key. That seems better than using timestamp. 

4. Changed YAML to trigger `ci.yaml` workflow only on pull requests. Currently, it triggers for any "push" event and runs on any branches that haven't turned into Pull requests.  Not sure if we want to keep "push" trigger. Would it be useful for master or release branches or CK branches? @causten @pfultz2 @turneram 

5. Adds another workflow to clean caches for the closed PRs.

`ccache` caches are taking 400+ MB for each instance. That means we have an upper limit of 20-something PRs before we out of the 10GB limit.  One way to lower ccache sizes is to use a higher compression level. But that could slow down builds.


umangyadav:
Looking to add https://github.com/actions/cache/tree/main/restore explicit restore/save option for `ccache` and `dependencies`.

causten:
Can you add a description of what the change is as part of the PR request

causten:
> 4\. Changed YAML to trigger `ci.yaml` workflow only on pull requests. Currently, it triggers for any "push" event and runs on any branches that haven't turned into Pull requests.  Not sure if we want to keep "push" trigger. Would it be useful for master or release branches or CK branches?

I think it makes sense to only use CI resources on branches that are PRs.

umangyadav:
Need to keep push event for `develop` branch. Will update one this run finishes.

.github/workflows/ci.yaml
```diff
@@ -217,35 +200,29 @@ jobs:
...
-      run: |
-        python -m pip install --upgrade pip
-        pip install https://github.com/RadeonOpenCompute/rbuild/archive/master.tar.gz
-        rbuild prepare -d cget -s gh
-        sudo apt-get install -y lld
-    - name: Prepare timestamp
-      id: cache_timestamp
-      shell: bash 
-      run: echo timestamp="$(date +'%Y-%m-%dT%H:%M:%S')" >> $GITHUB_OUTPUT
+      if: steps.deps_cache.outputs.cache-hit != 'true'
```
pfultz2, line 215: There is no need for this. Rbuild already keeps a hash and will only rebuild if the dependencies have changed.

.github/workflows/ci.yaml
```diff
@@ -319,6 +297,7 @@ jobs:
         # Look to see if there is a cache hit for the corresponding requirements file
         key: 
           ${{ matrix.os }}-cget-4-${{ hashFiles('requirements.txt', 'dev-requirements.txt') }}
+        restore-keys: |
```
pfultz2, line 300: Same here the restore keys should include the full key.

pfultz2:
> Use actions/cache@v3 instead of always-upload-cache. Not sure why it was used earlier

We use always-upload-cache so it will still cache the build even when there is a failure.

.github/workflows/ci.yaml
```diff
@@ -217,35 +200,29 @@ jobs:
...
-      shell: bash 
-      run: echo timestamp="$(date +'%Y-%m-%dT%H:%M:%S')" >> $GITHUB_OUTPUT
+      if: steps.deps_cache.outputs.cache-hit != 'true'
+      run: rbuild prepare -d cget -s gh
 
     - name: Cache files for ccache
       # Ignore the failure of a step and avoid terminating the job.
       continue-on-error: true
-      uses: pat-s/always-upload-cache@v2.1.3
+      uses: actions/cache@v3
```
pfultz2, line 221: Will `actions/cache@v3` upload the cache even when there is a build failure?

pfultz2:
> Need to keep push event for develop branch.

Yes, because caches can't be reused across branches. GH will only uses the caches from the parent branch(which would be develop). So we need to run on develop to update the caches.

pfultz2:
> Using timestamps in the primary key would always result in cache misses. 

This is not true. There are cache hits, I just looked at the log for tidy from #1675:

```
Run pat-s/always-upload-cache@v3.0.11
Received 157460 of 157460 (100.0%), 0.4 MBs/sec
Cache Size: ~0 MB (157460 B)
/usr/bin/tar -z -xf /home/runner/work/_temp/ffe5ab41-5dfc-4fb1-add5-52d2b43d1162/cache.tgz -P -C /home/runner/work/AMDMIGraphX/AMDMIGraphX
Cache restored successfully
Cache restored from key: tidy-cache-2023-04-10T17:16:22
```

And saving the cache shows this:

```
Post job cleanup.
/usr/bin/tar --posix -z -cf cache.tgz --exclude cache.tgz -P -C /home/runner/work/AMDMIGraphX/AMDMIGraphX --files-from manifest.txt
Cache Size: ~0 MB (157218 B)
Cache saved successfully
Cache saved with key: tidy-cache-2023-04-10T18:58:20
```

There is a key without the timestamp which will restore the first key that closely matches. That is why the restore and save key for the cache is different.

The timesteps are used so the cache can be updated. Caches are immmutable so the only way to updated it is to give it a unique key(in this case a timestep).

.github/workflows/ci.yaml
```diff
@@ -17,38 +17,22 @@ jobs:
...
-      shell: bash 
-      run: echo timestamp="$(date +'%Y-%m-%dT%H:%M:%S')" >> $GITHUB_OUTPUT
-
     - name: Cache files for tidy
-      uses: pat-s/always-upload-cache@v3.0.11
+      uses: actions/cache@v3 
       with:
         path: tidy-cache
-        key: tidy-cache-${{ steps.cache_timestamp.outputs.timestamp }}
+        key: tidy-cache-${{ github.ref }}
```
pfultz2, line 34: This should be `github.run_id` otherwise the cache will never update on the branch.

.github/workflows/ci.yaml
pfultz2: From the issue [here](https://github.com/actions/cache/issues/92#issuecomment-1362456629) we need to use action/cache/save. This section describes exactly the same scenario of saving the cache even when there is a test failure:

https://github.com/actions/cache/tree/main/save#always-save-cache

.github/workflows/ci.yaml
```diff
@@ -217,35 +200,29 @@ jobs:
...
       continue-on-error: true
       uses: actions/cache@v3
+      id: deps_cache
       with:
         # This path is specific to Ubuntu
         path: ${{ github.workspace }}/cget
         # Look to see if there is a cache hit for the corresponding requirements file
         key: 
           ${{ matrix.os }}-cget-4-${{ hashFiles('requirements.txt', 'dev-requirements.txt') }}
+        restore-keys: |
```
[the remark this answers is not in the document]
umangyadav: Why is that ? 

It would try to first restore cache by using primary key that is key defined under field `key:` 

If that doesn't exist then, it would try to use `restore-keys`.   I don't see why we need to duplicate keys in both `key:` and `restore-keys:`.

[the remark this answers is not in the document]
pfultz2: OK I see, I didnt know it would restore the `key` first before trying the `restore-keys`.

umangyadav:
> This is not true. There are cache hits, I just looked at the log for tidy from https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1675: 

Yes, it worked before it used restore-key without timestamp. But that would be considered as "stale" cache restore.

pfultz2:
> But that would be considered as "stale" cache restore.

What does that mean?

umangyadav:
> What does that mean?

That means it nearest matching cache. Just for an example, if cache is used for dependencies and if it were to store stale cache then, it is possible that, it didn't restore all the dependencies and would need to still install remaining ones.

umangyadav:
If it re-stores stale cache then, it would be considered as cache miss by GH actions and it would re-generate and store cache again. Since we are using time stamps as part of the key, it would happen all the time.

pfultz2:
> Just for an example, if cache is used for dependencies and if it were to store stale cache then, it is possible that, it didn't restore all the dependencies and would need to still install remaining ones.

Thats why we dont use timesteps for dependencies because we can easily create a hash for dependencies.

> it would be considered as cache miss by GH actions and it would re-generate

What do you mean re-generate?

> and store cache again. Since we are using time stamps as part of the key, it would happen all the time.

Yes thats exactly why we use the timesteps, so the cache can be updated with latest. Say on a branch you push a change that needs to recompile 20 files(the rest is cached) and then later push a second change that only needs to recompile 1 additional file. If we didnt use timesteps then we would need to recompile 21 files instead of 1 file on the second change since the cache wasn't saved from recompiling the 20 files.

umangyadav:
> Yes thats exactly why we use the timesteps, so the cache can be updated with latest everytime.

Yes but that would mean we save cache for each and every run and we only have 10 GB limit. 

> What do you mean re-generate?

I mean it will save cache again with new key.

.github/workflows/ci.yaml
umangyadav: I find it better to keep this than relying on rbuild.

pfultz2:
> Yes but that would mean we save cache for each and every run and we only have 10 GB limit.

Yea thats unfortunate, but there doesn't seem to be any other way to update the cache:

https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache

However, GH should automatically remove the older caches(based on last access) when the 10GB limit is reached:

https://github.com/actions/cache/blob/main/README.md#cache-limits

umangyadav:
> However, GH should automatically remove the older caches(based on last access) when the 10GB limit is reached:

Yeah, but It is not doing it for some reason. I've seen it reach 80 GB and then suddenly problems appear in CI.

umangyadav:
> but there doesn't seem to be any other way to update the cache: 

I should work if we restore and save for each run.

.github/workflows/ci.yaml
umangyadav: That would be similar to using timestamp. It would generate different each time for each run and would miss and save cache each time.

.github/workflows/ci.yaml
pfultz2: Why? This puts more burden on the yaml script when we already have a tool to handle this.

.github/workflows/ci.yaml
```diff
@@ -70,6 +60,14 @@ jobs:
           ..
         make -j2 -k onnx-proto tf-proto tidy
 
+    - name: Save cache files for tidy
+      uses: actions/cache/save@v3 
+      if: always()
```
pfultz2, line 65: So this will run even if the previous steps have failed, which will match the behavior of the previous always-upload-cache we were using:

https://stackoverflow.com/a/58859404/375343

umangyadav:
> It can possibly work if we restore and save for each run.

https://github.com/actions/cache/issues/342

Need to do `restore --> clear --> save`  as mentioned in that issue as a workaround for "ccache" cache.

pfultz2:
> Need to do restore --> clear --> save as mentioned in that issue as a workaround for "ccache" cache.

Haha, I just saw that and was about to post it as well. This comment here looks like a good solution:

https://github.com/actions/cache/issues/342#issuecomment-1399442670

umangyadav:
I keep getting this error during `clear` stage.  Unable to figure out why is that. 


```
could not check for binary extension: HTTP 401: Bad credentials (https://api.github.com/repos/actions/gh-actions-cache/releases/latest)
Try authenticating with:  gh auth login
Error: Process completed with exit code 1.
```

umangyadav:
Need to pin actions-cache to v1.0.1 as GLIBC_2.32 and GLIBC_2.34 are not found on ubuntu-20.04 
https://github.com/actions/gh-actions-cache/issues/44#issuecomment-1334013063

causten (approved):

PR #1683, opened 2023-04-14
[mlir] Adding quant convolution fusion as anchor op

jerryyin:
- Exposed the `mlir_enabled()` call the decide for lowering pipeline's enablement
    - Disabled the rewrite quantization pipeline in mlir compilation
 - Added quant convolution as anchor ops
    - Fixed the return type expectations
 - Added the fall back hip implementation for quantizelinear and dequantizelinear
    - Will need advises to improve the implementation for quantizelinear


src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -38,6 +38,10 @@ namespace op {
 struct quantizelinear
 {
     std::string name() const { return "quantizelinear"; }
+
+    value attributes() const { return {{"pointwise", true}, {"point_op", 
+        "${function:max}(${function:min}(${function:round}(${function:convert}<float>(${0}) / ${1}) + ${function:convert}<float>(${2}), 127.0), -128.0)"}}; }
```
jerryyin, line 43: I cannot think of a way to get rid of the hard coded constants for clip: -128 and 127. I tried `std::numeric_limits<decltype(${2})>::max()` or `...::lowest()` but that require the existence of ${2} in the first place.

manupak:
A general question : 

What happens to quantization ops surrounding dot ops and other non-quant-convolution ops (say avgpool for e.g.) ? 
i.e. if fuse_mlir does not absorb quantization ops will the compilation break when mlir is enabled?

Can we try running a bert model with --int8 flag ?

jerryyin:
> What happens to quantization ops surrounding dot ops and other non-quant-convolution ops ... Can we try running a bert model with --int8 flag?

Good point, it is my immediate next step is to run this with more realistic workloads too. I will probably start with a quantized resnet50 then bert with --int8. Right now the focus of this PR is the sequence of `quant_conv -> dequantizelinear -> activation -> quantizelinear`. Should I need to address issues where use cases become different, I will scope that in the follow-up PRs.

>  if fuse_mlir does not absorb quantization ops will the compilation break when mlir is enabled?

If `fuse_mlir` doesn't absorb quantization, it will fall back as a regular pointwise module, to a fused hip kernel. `point_op` attribute is where a fall back is defined.

manupak:
> If `fuse_mlir` doesn't absorb quantization, it will fall back as a regular pointwise module, to a fused hip kernel

My concern is stemming from the fact we dont run rewrite_quantization pass if mlir is enabled.

Am I right in understanding that we dont need to run rewrite_quantization to get functionally working output ?
(I acknowledge that pointwise mods will remain but unsure whether they are compilable w/o running rewrite_quantization  on them)

I was more curious whether we should add a rewrite_quantization  after fuse_mlir with the gating if mlir_enabled -- essentially enabling mlir will just delay rewrite_quantization  after fuse_mlir rather than not running it completely.

jerryyin:
> Am I right in understanding that we dont need to run rewrite_quantization to get functionally working output ?

Correct, even if I don't run fuse_mlir and rewrite_quantization, test cases will still pass and be able to fuse as pointwise kernels.

> I was more curious whether we should add a rewrite_quantization after fuse_mlir

That was the alternative when I discuss with @pfultz2. I think the decision ended up to be that if hip definition is defined, we don't need to run a second time. Having said that, should there be additional caveat that can't be addressed by the existing approach, we can fall back to this way.

pfultz2:
> That was the alternative when I discuss with @pfultz2. I think the decision ended up to be that if hip definition is defined, we don't need to run a second time. Having said that, should there be additional caveat that can't be addressed by the existing approach, we can fall back to this way.

We can do a better job of optimizing when the quantization is rewritten. For example, we can replace the scales with the reciprocals to get rid of any division.

jerryyin:
@pfultz2 If that's your preference, I can add a rewrite_quantization after the fuse_mlir pass, to capture additional opportunity for optimizations.

src/include/migraphx/op/quantizelinear.hpp
pfultz2: Yea this is getting kind of messy. Thinking about it more, we could just set `pointwise` to true and then run `rewrite_quantization` in `generate_pointwise`(we already run some passes before generating).

jerryyin: Also, I'm assuming you prefer the same for both quantizelinear and dequantizelinear? Reason I'm asking is due to math in quantizelinear is much more straightforward and there's no harm to keep a definition there.

pfultz2: > You mean to add it here? 

Yes.

> Also, I'm assuming you prefer the same for both quantizelinear and dequantizelinear? 

Yes.

src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -37,6 +37,14 @@ namespace op {
 
 struct dequantizelinear
 {
+
+    value attributes() const {
+        // Note: point_op attribute is not used in this op. Instead, in
+        // gpu compilation pipeline, rewrite_quantization will be invoked
+        // from generate_pointwise() to rewrite this op.
+        return {{"pointwise", true}, {"point_op", ""}};
```
pfultz2, line 45: Just remove the `point_op` attribute.

src/cpp_generator.cpp
```diff
@@ -174,6 +174,11 @@ std::string cpp_generator::generate_point_op(const operation& op,
         else if(with_char(::isdigit)(key[0]))
         {
             auto i = std::stoul(key);
+            // For an optional argument where i >= args.size(), treat
+            // the optional argument as a straight zero. This will
+            // cacel out the optional bias, if it exists.
```
pfultz2, line 179: This comment doesnt match what is done in the code.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -68,7 +92,7 @@ namespace {
 
 MIGRAPHX_PRED_MATCHER(is_mlir_conv, instruction_ref ins)
 {
-    if(ins->name() != "convolution")
+    if(ins->name() != "convolution" && ins->name() != "quant_convolution")
```
pfultz2, line 95: Use `and` instead of `&&`.

src/cpp_generator.cpp
jerryyin: Nice catch.

src/targets/gpu/fuse_mlir.cpp
jerryyin: Sorry, has not been very careful with this convention. I'll pay more attention to this.

jerryyin:
Please note that the clang format failure is unrelated with changes from this PR. (Somehow it is complaining about the `attributes()` code block in that file that I can't even find).

With that I believe this PR has full pass in CI and addressed all code review comments. Please let me know if you have any further suggestions. Thanks!

src/include/migraphx/op/dequantizelinear.hpp
pfultz2: If we do keep the `point_op` we should make it a `static_assert`, so any codegen can have a clearer message if it tries to generate this operator:

```cpp
return {{"pointwise", true}, {"point_op", R"(static_assert(false, "dequantizelinear does not have a point_op"))" }};
```

Of course, we already through an error if `point_op` is missing, which I think is cleaner then doing this `static_assert` injection. So I think it would be better to remove the `assert`.

jerryyin: Sounds good, seems like a reasonable decision. Got rid of the assert as well as attribute in 6040c79751f850f33d9cb7c53394dfce38160d97

test/gpu/quantization.cpp
```diff
@@ -110,7 +111,16 @@ TEST_CASE(int8_quantization)
...
 
-        EXPECT(migraphx::verify_range(ref_result, gpu_result));
+        // Note: the tolerance for mlir_enabled result is temporarily bumped
+        // higher because the lowering pipeline between mlir fallback and
+        // regular non-mlir pipeline diverged. MLIR fallback uses the
+        // rewrite_quantization at the very end of the pipeline, whereas
+        // the regular pipeline uses the rewrite_quantization in the much
+        // earlier stage.
+        if(migraphx::gpu::mlir_enabled())
+            EXPECT(migraphx::verify_range(ref_result, gpu_result, 1e5));
```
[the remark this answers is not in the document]
jerryyin: This isn't comparing to MLIR. This is comparing from
 - a. Quantization that happens normally, in very front of the pipeline
 - b. MLIR's fusion didn't capture the pointwise module, so quantization happens in fallback mode, in the very back of the pipeline.

In both cases, the quantizelinear and dequantizelinear was lowered within migraphx implementation. The error increased because the sequence matters. Effectively, in the fallback path, because of the sequence difference, it becomes the comparison between two possible lowering sequence of migraphx.

---

That's also why I'm trying to make migraphx support int32 type natively. Only if it does support int32 type, will it not populate redundant convert ops. And only when it doesn't populate redundant convert ops can MLIR lowering of quantizelinear and dequantizelinear happen at the front the pipeline. Then we can do the error comparison fairly.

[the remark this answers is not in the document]
pfultz2: Ok I see. It does make sense to use a higher tolerance since this is using int8 internally and no fp32. I would rather use the same tolerance for both paths instead of making it conditional. We can lower it again when we get the int32 working.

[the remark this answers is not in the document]
jerryyin: Sounds good, I'd prefer to just keep as-is if is it not a strong preference :-). I have double checked that once in32 type is enabled, this test will be able to pass as-is... So this is really just throw-away code in the next PR either one way or another.

pfultz2 (approved):

src/targets/gpu/compile_gen.cpp
```diff
@@ -171,7 +172,8 @@ std::string make_transformer_args(std::vector<std::string> transformers)
 void generate_pointwise(cpp_generator& gg, const module& pm, const std::string& name)
 {
     module m = pm;
-    run_passes(m, {eliminate_common_subexpression{}, dead_code_elimination{}});
+    run_passes(m,
+               {rewrite_quantization{}, eliminate_common_subexpression{}, dead_code_elimination{}});
```
manupak, line 176: right, we have this as a backup!

PR #1686, opened 2023-04-17
Add trace flag for propagate_constant

pfultz2:
This will show whats being replaced with a constant. This is useful for debugging where a literal comes from.

### DIFF
--- a/src/propagate_constant.cpp
+++ b/src/propagate_constant.cpp
@@ -88,7 +88,7 @@ void propagate_constant::apply(module& m) const
     {
         if(not literals[i].empty())
         {
-            if (enabled(MIGRAPHX_TRACE_PROPAGATE_CONSTANT{}))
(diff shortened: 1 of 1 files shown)


TedThemistokleous (approved):

src/propagate_constant.cpp
```diff
@@ -27,11 +27,14 @@
 #include <migraphx/literal.hpp>
 #include <migraphx/functional.hpp>
 #include <migraphx/par_for.hpp>
+#include <migraphx/env.hpp>
 #include <unordered_set>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_PROPAGATE_CONSTANT)
```
umangyadav, line 36: we may want to document all ENV flags.

Also, this makes me wonder again about better logging interface.

umangyadav (approved):

shivadbhavsar (approved):

PR #1689, opened 2023-04-17
Handle broadcasts across dot and concat

pfultz2:
This improves the constant propagation for bert models. Larger batch size no longer use as large of constants.


umangyadav (commented):
Looks good. Need to fix some tests in CI though. also Performance tests have `nan` and BERT is dropping in perf.

src/simplify_algebra.cpp
```diff
@@ -388,12 +383,72 @@ struct find_inner_broadcast
...
+                auto axis = v.at("axis").to<std::size_t>() - naxes;
+                return m.insert_instruction(
+                    ins, make_op("broadcast", {{"axis", axis}, {"out_lens", lens}}), input);
+            }
+            assert(false);
+            return m.end();
+        };
+        auto a1        = insert_broadcast(a);
+        auto b1        = insert_broadcast(b);
+        auto dot       = m.insert_instruction(ins, make_op("dot"), a1, b1);
```
umangyadav, line 438: `dot` seems to be returning standard-shaped outputs in `compute_shape()`.  This transformation would make it return broadcasted output.

umangyadav: Not sure if this would add extra `contiguous` if consumers expect standard shapes.

causten:
I tested this PR.  I saw a performance gain.  More interestingly it also significantly speed up the compile phase.

src/simplify_algebra.cpp
CharlieL7: I think that would be a fine tradeoff in comparison to doing a larger gemm.

CharlieL7 (commented):
Jenkins looks to be failling on py_backend for shufflenet_cpu.

src/simplify_algebra.cpp
CharlieL7: Should be fine in comparison to doing a larger gemm.

TedThemistokleous (approved):

CharlieL7 (approved):

Issue #1311, opened 2022-07-21
Investigate shape::multi() and non-standard shapes


CharlieL7:
### Definitions
* **buffer indices**: the indices to to actual memory buffer, changes with non-standard shapes with the same lengths
* **dimensional indices**: the dimensional indices of the form [a_1, a_2, ..., a_r], where r is the rank of the tensor
   * should be the same for any shape with the same lengths
* **1D indices**: the flattened indices from [0, number of elements -1]

### Info
* `shape::multi()` converts 1D indices (usually passed from functions like `par_for()`) into dimensional indices.
*  `shape::multi()` is currently limited to standard shapes only
   * but the dimensional indices should be the same for any shape with the same lengths, so it should also work for non-standard shapes
   * could then use the dimensional indices to access the buffer data with the `shape::index()` function

### Todo
* need to look into how such a change might affect all the operators that use `shape::multi()`
* prototype such a change
* clean up some code design
   * `(void)end` and double same assert

pfultz2:
> shape::multi() is currently limited to standard shapes only

This can work for non-standard shapes as well, we just need to use just the lens instead of the strides. The implementation on the GPU for [`multi`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/shape.hpp#L119) actually works for non-standard shapes.

PR #1690, opened 2023-04-17
Update multi() to work with non-std shapes

umangyadav:
Solves #1311

### DIFF
--- a/src/shape.cpp
+++ b/src/shape.cpp
@@ -373,6 +373,7 @@ void shape::multi_copy(std::size_t idx, std::size_t* start, const std::size_t* e
 {
     size_t tidx = idx;
     (void)end;
+    assert(idx < elements());
     assert(lens().size() <= (end - start));
     for(size_t ii = lens().size() - 1; ii > 0; ii--)
     {
--- a/src/shape.cpp
+++ b/src/shape.cpp
@@ -361,29 +361,23 @@ std::size_t shape::index(std::size_t i) const
     }
 }
 
-std::vector<std::size_t> shape::multi(std::size_t i) const
+std::vector<std::size_t> shape::multi(std::size_t idx) const
 {
-    assert(this->standard());
-
+    if(idx >= elements())
+    {
+        MIGRAPHX_THROW("Can not convert element-index: " + std::to_string(idx) +
+                       " to multi-dimensional index for shape with num elements: " +
+                       std::to_string(elements()));
+    }
     std::vector<std::size_t> indices(lens().size());
-    multi_copy(i, indices.data(), indices.data() + lens().size());
-
+    size_t tidx = idx;
+    for(size_t ii = indices.size() - 1; ii > 0; ii--)
+    {
+        indices[ii] = tidx % lens()[ii];
+        tidx        = tidx / lens()[ii];
+    }
+    indices[0] = tidx;
     return indices;
-}
-
-void shape::multi_copy(std::size_t i, std::size_t* start, const std::size_t* end) const
-{
-    assert(this->standard());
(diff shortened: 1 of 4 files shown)


src/targets/cpu/include/migraphx/cpu/pointwise.hpp
```diff
@@ -43,7 +43,8 @@ struct multi_index
     {
```
umangyadav, line 43: This `multi_index` struct is not used anywhere for the CPU. 

Should we just remove that ?

pfultz2: Its used below.

umangyadav: I don't find any reference when I search for its constructor or when i search for "multi_index"

src/shape.cpp
```diff
@@ -361,31 +361,25 @@ std::size_t shape::index(std::size_t i) const
     }
 }
 
-std::vector<std::size_t> shape::multi(std::size_t i) const
+std::vector<std::size_t> shape::multi(std::size_t idx) const
 {
-    assert(this->standard());
-
+    if(idx >= elements())
```
CharlieL7, line 366: Add a throw for if the shape is dynamic.

CharlieL7 (changes requested):

src/shape.cpp
umangyadav: Call to `elements()` would throw in that case. I can update test.

CharlieL7 (approved):

src/targets/cpu/include/migraphx/cpu/pointwise.hpp
```diff
@@ -43,7 +43,8 @@ struct multi_index
...
         assert(n < max_size);
         std::copy(s.lens().begin(), s.lens().end(), dims);
-        s.multi_copy(i, index, index + max_size);
+        const std::vector<size_t> m_idx = s.multi(i);
+        std::copy(m_idx.begin(), m_idx.end(), index);
```
pfultz2, line 47: This should use `multi_copy` so we dont need to construct a temporary vector(which uses heap allocation). The `index` member is an array so we dont need to do heap allocation when constructing.

src/shape.cpp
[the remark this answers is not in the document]
umangyadav: Why is that ? Shouldn't it be illegal during runtime as well ?

src/targets/cpu/include/migraphx/cpu/pointwise.hpp
umangyadav: Added back.

src/shape.cpp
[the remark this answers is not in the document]
pfultz2: We shouldn't do this extra comparison when its not necessary. Calling with the correct index is a precondition of the function.

src/shape.cpp
```diff
@@ -363,21 +363,23 @@ std::size_t shape::index(std::size_t i) const
 
 std::vector<std::size_t> shape::multi(std::size_t idx) const
 {
-    if(idx >= elements())
-    {
-        MIGRAPHX_THROW("Can not convert element-index: " + std::to_string(idx) +
-                       " to multi-dimensional index for shape with num elements: " +
-                       std::to_string(elements()));
-    }
+    assert(idx < elements());
```
pfultz2, line 366: The assert should probably go into `multi_copy` as well.

umangyadav: Done.

pfultz2 (approved):

PR #1696, opened 2023-04-19
Expose instruction shape and operator through python api

shivadbhavsar:
Simply exposing `get_shape` and `get_operator` methods for instruction_ref object in the python API.

Motivation: Having these methods accessible will help clean up torch_migraphx converters and also allow us to not rely on an additional torch shape prop pass to infer input shapes when required.

### DIFF
--- a/test/py/test_instruction.py
+++ b/test/py/test_instruction.py
@@ -1,0 +1,52 @@
+#####################################################################################
+# The MIT License (MIT)
+#
+# Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


test/py/test_instruction.py
```diff
@@ -0,0 +1,52 @@
...
+    mm = p.get_main_module()
+    input_shape = migraphx.shape(lens=[2, 24])
+    i = mm.add_parameter("x", input_shape)
+    i2 = mm.add_instruction(migraphx.op("relu"), [i])
+    out_op = i2.op()
+
+    assert out_op.name() == "relu"
+
+
+test_instruction_shape()
```
umangyadav, line 51: cover this under `if __name_ == __main__:`

umangyadav (approved):

TedThemistokleous (approved):

PR #1705, opened 2023-04-21
[MLIR][5.7] add input fusion support for view ops

manupak:
This commit adds support for slice,transpose,contigous and reshape
fusions into input tensors for a fused mlir kernel.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/864


manupak:
built on top of https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1678

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -44,22 +44,20 @@ struct mlir_op
...
     shape compute_shape(std::vector<shape> inputs, const std::vector<module_ref>& mods) const
     {
-        check_shapes{inputs, *this}.packed_or_broadcasted();
-        if(mods.size() != 1)
-            MIGRAPHX_THROW("should have one submodule.");
-        if(inputs.size() < 2)
-            MIGRAPHX_THROW("should have at least two inputs.");
-        auto n = inputs.size();
-        return op.compute_shape({inputs[n - 2], inputs[n - 1]});
+        // Currently fused mlir fusion uses static shapes.
```
manupak, line 57: How important is it calculate the shape from the scratch here given that we only support static shaped kernels ?

pfultz2: > How important is it calculate the shape from the scratch here given that we only support static shaped kernels ?

The shape can change after `eliminate_contiguous`. And most likely we would want to propagate the layout if it changes.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -44,22 +44,20 @@ struct mlir_op
...
     template <class Self, class F>
     static auto reflect(Self& self, F f)
     {
-        return pack(f(self.op, "op"));
+        return pack(f(self.op, "op"), f(self.output_shape, "output_shape"));
     }
 
     shape compute_shape(std::vector<shape> inputs, const std::vector<module_ref>& mods) const
     {
-        check_shapes{inputs, *this}.packed_or_broadcasted();
```
pfultz2, line 56: This should not be removed as MLIR doesnt support non-packed tensors as inputs.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -44,22 +44,20 @@ struct mlir_op
...
-        if(mods.size() != 1)
-            MIGRAPHX_THROW("should have one submodule.");
-        if(inputs.size() < 2)
-            MIGRAPHX_THROW("should have at least two inputs.");
```
pfultz2, line 60: This should not be removed.

src/targets/gpu/mlir.cpp
```diff
@@ -532,11 +537,26 @@ struct mlir_program
...
             auto name = get_name(ins);
             auto ops  = create_operation_state(name);
             ops.add_attribute_value(get_operator_value(ins->get_operator()));
             if(ins->name() != "@return")
                 ops.add_results({get_shape(ins)});
+            if(ins->name() == "@literal")
+            {
+                auto e = ins->eval();
+                literal r{};
+                e.visit_at([&](auto x) { r = literal{x}; });
```
pfultz2, line 554: Use `ins->get_literal()`.

src/targets/gpu/fuse_mlir.cpp
manupak: ok added shape inference.

src/targets/gpu/fuse_mlir.cpp
manupak: restored.

src/targets/gpu/mlir.cpp
manupak: done.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -58,8 +58,40 @@ struct mlir_op
...
+            if(ins->name() == "@return")
+            {
+                return ins_shapes[ins->inputs().at(0)];
+            }
+            std::vector<shape> input_shapes;
+            for(auto in : ins->inputs())
+            {
+                shape in_shape = ins_shapes[in];
+                input_shapes.push_back(in_shape);
+            }
```
pfultz2, line 91: Use `std::transform` here.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -89,6 +121,56 @@ struct find_mlir_op
...
+            std::vector<operation> op_stream;
+            while(contains({"slice", "transpose", "contiguous", "reshape"}, input->name()))
+            {
+                op_stream.push_back(input->get_operator());
+                input = input->inputs().at(0);
+            }
+            top_inputs.push_back(input);
+            instruction_ref prev_input =
+                mm->add_parameter("y" + std::to_string(input_cnt++), input->get_shape());
+            if(!op_stream.empty())
```
pfultz2, line 160: Remove this condition its redundant.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -89,6 +121,56 @@ struct find_mlir_op
...
+            {
+                for(auto op : reverse_iterator_for(op_stream))
```
pfultz2, line 162: Just use `for(const auto& op:reverse(op_stream))`.

src/targets/gpu/fuse_mlir.cpp
manupak: done.

src/targets/gpu/fuse_mlir.cpp
manupak: done.

src/targets/gpu/fuse_mlir.cpp
manupak: done.

manupak:
a friendly ping! @pfultz2...
I 've rebased over other PR that got merged yesterday.

PR #1712, opened 2023-04-24
Update C/C++ API for dynamic batch

CharlieL7:
* Relies on #1711
* Exposes `dynamic_dimension` as a opaque object with `dynamic_dimensions` and `optimals`
* Exposes ONNX `dyn_input_dims` and `default_dyn_dim` to run with dynamic batch
* Updates api.py to be able to create objects from aggregate initialization (used for `dynamic_dimension`)
* Uses offload copy for now
  * Must create a maximum size buffer with offload copy (max batchsize) every time
  * Will need to update API later to allow for allocating tuple arguments


src/api/migraphx.py
```diff
@@ -45,56 +45,49 @@ def shape_type_wrap(p):
...
-    if p.returns:
-        p.add_param('migraphx_onnx_options *')
-        p.bad_param('${name} == nullptr', 'Null pointer')
-        p.write = ['*${name} = migraphx::to_onnx_options(${result})']
-    else:
-        p.add_param('migraphx_onnx_options *')
-        p.read = '${name} == nullptr ? migraphx::onnx_options{} : migraphx::to_onnx_options(*${name})'
+@api.handle('migraphx_optimals', 'std::set<size_t>')
+def optimals(h):
+    h.constructor('create')
```
pfultz2, line 58: The `create` should take a pointer and size similar to the `instructions` and `modules` classes.

src/api/migraphx.py
```diff
@@ -45,56 +45,49 @@ def shape_type_wrap(p):
...
+    h.method('insert',
```
pfultz2, line 59: This shouldn't be exposed.

src/api/migraphx.py
```diff
@@ -45,56 +45,49 @@ def shape_type_wrap(p):
...
-def auto_handle(*args, **kwargs):
-    def with_handle(f):
-        return api.handle('migraphx_' + f.__name__, 'migraphx::' + f.__name__,
-                          *args, **kwargs)(f)
-
-    return with_handle
+@api.handle('migraphx_dynamic_dimensions',
+            'std::vector<migraphx::shape::dynamic_dimension>')
+def dynamic_dimensions(h):
+    h.constructor('create')
```
pfultz2, line 81: This should take a pointer and size similar to modules and instructions. On the C++ side you will need the variadiac constructor instead of initializer lists.

src/api/migraphx.py
```diff
@@ -45,56 +45,49 @@ def shape_type_wrap(p):
...
+    h.method('size', returns='size_t')
+    h.method('get',
+             api.params(idx='size_t'),
+             fname='at',
+             cpp_name='operator[]',
+             returns='const migraphx::shape::dynamic_dimension&')
+    h.method('add',
```
pfultz2, line 88: This should not be exposed.

src/api/migraphx.py
```diff
@@ -182,6 +189,11 @@ def program_parameters(h):
              api.params(name='const char*',
                         argument='const migraphx::argument&'),
              invoke='${program_parameters}[${name}] = ${argument}')
+    h.method('contains',
```
pfultz2, line 192: I dont see why we need to expose this in the API.

src/api/migraphx.py
```diff
@@ -109,26 +102,40 @@ def shape(h):
...
+                   dims='std::vector<migraphx::shape::dynamic_dimension>'))
     h.method('lengths',
              fname='lens',
              returns='const std::vector<size_t>&',
              const=True)
     h.method('strides', returns='const std::vector<size_t>&', const=True)
+    h.method('dyn_dims',
+             returns='std::vector<migraphx::shape::dynamic_dimension>',
+             const=True)
+    h.method('to_static',
```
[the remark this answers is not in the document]
CharlieL7: It can probably be worked around; was a convenience thing for getting a static shape for running dynamic batch.  I will see what I can do.

src/api/migraphx.py
CharlieL7: I will remove it, used it previously. Not used anymore.

src/api/migraphx.py
CharlieL7: Not clear to me why we can't use an initializer_list for this. Are we limited to the variadic constructor because of something to do with type-safety?

pfultz2: Because we need to convert the C++ classes to C classes. Thats why we do it for instructions and modules classes. You can follow the pattern there where we create a local array for the conversion.

kahmed10 (approved):

src/api/api.cpp
```diff
@@ -187,6 +200,14 @@ std::vector<const char*> get_names(const std::unordered_map<std::string, Value>&
     return result;
 }
 
+template <class T>
+std::set<T> make_set(const T* x, std::size_t n)
+{
+    std::set<T> result;
+    std::transform(x, x + n, std::inserter(result, result.begin()), [&](auto&& y) { return y; });
+    return result;
```
pfultz2, line 208: I think you can write this as `return {x, x+n}`.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -571,10 +571,91 @@ using require_interface =
...
+ * Container to hold optimal dynamic dimension values.
+ */
+struct optimals : MIGRAPHX_HANDLE_BASE(optimals)
+{
+    MIGRAPHX_HANDLE_CONSTRUCTOR(optimals)
+
+    optimals(std::initializer_list<size_t> init_list)
+    {
+        std::vector<size_t> a{init_list.begin(), init_list.end()};
+        this->make_handle(&migraphx_optimals_create, a.data(), a.size());
```
pfultz2, line 584: You should be able to do `this->make_handle(&migraphx_optimals_create, init_list.begin(), init_list.size())`.

src/api/migraphx.py
```diff
@@ -45,56 +45,48 @@ def shape_type_wrap(p):
...
-        p.bad_param('${name} == nullptr', 'Null pointer')
-        p.write = ['*${name} = migraphx::to_onnx_options(${result})']
-    else:
-        p.add_param('migraphx_onnx_options *')
-        p.read = '${name} == nullptr ? migraphx::onnx_options{} : migraphx::to_onnx_options(*${name})'
 
+@api.handle('migraphx_optimals', 'std::set<size_t>')
+def optimals(h):
+    h.constructor('create',
+                  api.params(ptr='size_t*', size='size_t'),
```
pfultz2, line 59: Should be `const size_t*`

src/api/migraphx.py
```diff
@@ -109,27 +101,37 @@ def shape(h):
...
     h.method('elements', returns='size_t', const=True)
     h.method('bytes', returns='size_t', const=True)
+    h.method('ndim', returns='size_t', const=True)
     h.method('equal',
              api.params(x='const migraphx::shape&'),
              invoke='migraphx::equal($@)',
              returns='bool',
              const=True)
     h.method('standard', returns='bool', const=True)
+    h.method('dynamic', returns='bool', const=True)
```
umangyadav, line 126:
```suggestion
    h.method('is_dynamic', returns='bool', const=True)
```

CharlieL7: We've named this method `shape.dynamic()` in the internal code. Should it be different in the API or should we make them all `is_dynamic()`? I made it `dynamic()` because the other boolean methods are one word.

umangyadav: Okay, it make senses to follow pattern then and keep it one word. Thanks for bringing that to attention.

pfultz2 (approved):

PR #1716, opened 2023-04-25
update rocBLAS version check to support 3.0 and above

kahmed10:
Previous logic was bugged if the major version was 3, but the minor was below 38.

### DIFF
--- a/src/targets/gpu/gemm_impl.cpp
+++ b/src/targets/gpu/gemm_impl.cpp
@@ -140,13 +140,8 @@ void gemm_impl(context& ctx,
             compute_type = rocblas_datatype_f32_r;
     }
 
-#if ROCBLAS_VERSION_MAJOR >= 3 || ROCBLAS_VERSION_MAJOR == 2 && ROCBLAS_VERSION_MINOR >= 38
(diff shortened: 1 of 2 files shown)


src/targets/gpu/gemm_impl.cpp
```diff
@@ -140,7 +140,7 @@ void gemm_impl(context& ctx,
             compute_type = rocblas_datatype_f32_r;
     }
 
-#if ROCBLAS_VERSION_MAJOR >= 2 && ROCBLAS_VERSION_MINOR >= 38
+#if ROCBLAS_VERSION_MAJOR >= 3 || ROCBLAS_VERSION_MAJOR == 2 && ROCBLAS_VERSION_MINOR >= 38
```
pfultz2, line 143: These checks can be completely removed. This is a really old version of rocblas.

causten: yes, older then rocm 4.3

kahmed10: removed

TedThemistokleous (approved):

pfultz2 (approved):

Issue #1679, opened 2023-04-12
Multiple ops (test cases) implicitly rely on the int32_type to be unsupported


jerryyin:
Refer to here: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/ec4b79c2c9a41f49c7ee86e95e3943418a15721b/src/targets/gpu/target.cpp#L96-L101

The list of unsupported types doesn't include int32_type. This make `int32_type` to be unsupported type. While it is okay that int32_type is unsupported because there isn't enough of the use cases, the existing test cases, however, shouldn't rely on this side effect to be passing. I believe the right way to handle this is to:
 - Correct all existing tests that rely on implicit int32_type to other type conversion
    - Replace the int32_type to be float_type, or remove the test case al-together
 - Leave only one unit test to check int32_type only, if we care about the coverage of this type
 - To go one further step, we should enable int32_type to enable fusion with quantized convolution output

The list of test cases would fail if I enable the `int32_type` are listed below:
 - test_scatter0
 - test_scatter1
 - test_reduce_op_small<migraphx::op::reduce_max, 2, migraphx::shape::int32_type>
 - test_reduce_op_small<migraphx::op::reduce_min, 2, migraphx::shape::int32_type>

PR #1721, opened 2023-04-26
Enabling native int32 type support

jerryyin:
This is what we have discussed at the meeting @pfultz2. The main issues with int32 type are related with unit tests:
 - For scatter op, it was relying the implicit conversion from int32 -> fp32 because int32 wasn't explicitly supported
 - For reduction op, I identified 2 issues, but has trouble to fix it in the MIGraphX-style code. Therefore all the dirty hacks.
   -  The `lowest{}` initialization wasn't functioning correctly for int32 type. From what I see, the templatized implementation looks correct, but ended up initializing to a mysterious larger integer
   - `v_max_i32` and `v_min_i32` needs to be used such that it can do correct signed comparison.

This address https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1679


src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
```diff
@@ -105,7 +120,23 @@ __device__ auto block_reduce(index idx, Op op, T init, Index n, F f)
...
     __shared__ type buffer[idx.max_nlocal() / lanes_per_thread];
+    // Hack: Note when in int32_t and lowest/highest op, I need to overrid the init value
     type x = init;
+    type y = init;
+    if constexpr(is_same<type, int32_t>{})
+    {
+        if constexpr(is_same<T, lowest>{})
+        {
+            x = INT_MIN;
+            y = INT_MIN;
```
pfultz2, line 131: If this is needed, then I think we might have a bug in `numeric_lowest`.

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
```diff
@@ -105,7 +120,23 @@ __device__ auto block_reduce(index idx, Op op, T init, Index n, F f)
...
+        }
+        else if constexpr(is_same<T, highest>{})
+        {
+            x = INT_MAX;
+            y = INT_MAX;
```
pfultz2, line 136: And this could be a bug in `numeric_max`. Is that actually needed? Or was this added for completeness?

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
jerryyin: I stared at `numeric_lowest` like ten times, and it looks correct. Just don't know how to debug that...

pfultz2: `numeric_lowest` does `-numeric_max<T>() - 1`, I dont think that will overflow, but maybe an implicit conversion is causing an issue.

test/verify/test_scatter0.cpp
```diff
@@ -34,7 +34,11 @@ struct test_scatter0 : verify_program<test_scatter0>
         migraphx::program p;
         auto* mm = p.get_main_module();
         migraphx::shape sd{migraphx::shape::float_type, {3, 3}};
-        migraphx::shape si{migraphx::shape::int32_type, {2, 3}};
+        // Note, scatter op should tolerate indices of type int32_t.
+        // However, the existing implementation of scatter op is
+        // not jit and is based on hip_visit_all, which requires
+        // argument types to be similar with result type.
```
pfultz2, line 40: We can add int32_t to the visitor.

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
jerryyin: FYI, printing out the x as-is in `%d` (without the hack) will give me a mysterious `-2069889000`. Does not occur to be a number that can overflow to.

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
jerryyin: FYI, printing out the x as-is in %d (without the hack) will give me a mysterious `-260046824`.

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
pfultz2: But that is a wrong value, shouldn't it be `-2147483648`?

test/verify/test_scatter0.cpp
jerryyin: It turns out that adding int32_t to the visitor doesn't help with https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/ed6542ee4ee2106fe5c104cdef7057ffdb57f314/src/targets/gpu/device/include/migraphx/gpu/device/visit.hpp#L122

But I did give a shot in just commenting it out, which results in a computing error in the scatter0 unit test. I'm not sure that's worth my time to debugging again. Could you give some context on why there's the type must be same with result requirement in the first place?

pfultz2: Because they are all using the same visitor aka switch statement. If we want different types then we will need to use a nested visitor.

test/gpu/jit.cpp
```diff
@@ -356,4 +356,50 @@ TEST_CASE(compile_math)
...
+        if(contains({migraphx::shape::bool_type,
+                     migraphx::shape::tuple_type,
+                     migraphx::shape::int64_type,
+                     migraphx::shape::uint64_type},
+                    t))
+            continue;
+        auto name = migraphx::shape::cpp_type(t);
+
+        migraphx::shape::visit(t, [&](auto as) {
+            if(!as.is_integral())
```
pfultz2, line 391: We should check floating point types as well.

test/verify/test_scatter0.cpp
jerryyin: Do you mean to add additional nest level in the implementation of `src/targets/gpu/device/scatter.cpp`? I have to admit I'm clueless about what is going on in that implementation or why there's already two level nest of `hip_visit_all` and `gs_launch`.

src/targets/gpu/kernels/include/migraphx/kernels/type_traits.hpp
```diff
@@ -218,19 +218,23 @@ using common_type_t = typename common_type<Ts...>::type;
 
 #define MIGRAPHX_REQUIRES(...) class = enable_if_t<__VA_ARGS__>
 
-constexpr unsigned long int_max(unsigned long n) { return (1u << (n * 8)) - 1; }
+// Note, this cannot be used to get the maximum value of int64_type or
+// uint64_type because it is undefined behavior to left shift 64 bits for
+// these types
+constexpr unsigned long int_max(unsigned long n) { return (1ul << (n * 8)) - 1; }
```
pfultz2, line 224: Maybe we could do `if (n == 8) return -1;` to handle the 64 bit case?

test/verify/test_scatter0.cpp
pfultz2: So there is no need for us to pass `inds` to the first visitor. We can `indices.get_shape()` instead of `s1`.

But the visitors actually look wrong. The updates and data should be the same type, and only the indices are different(thats the way we implement this in the ref implementation).

pfultz2: `gs_launch` will launch the lambda on the device and use a global strided loop and pass the index to the lambda(ie the `auto i` parameter).

The `hip_visit_all` will do the visit like we normally do for the data types to get a `tensor_view` but instead will create a `hip_tensor_view`, which also needs to visit the tensor rank so the dimensions and strides of the shape can be allocated as a static array(but this will make us generate a different kernel for each tensor rank).

test/gpu/jit.cpp
```diff
@@ -356,4 +356,50 @@ TEST_CASE(compile_math)
...
+// NOLINTNEXTLINE
+const std::string assert_template = R"__migraphx__(
+#include <migraphx/kernels/math.hpp>
+#include <migraphx/kernels/types.hpp>
+using namespace migraphx;
+extern "C" {
+__global__ void kernel(void*) 
+{
+    static_assert(numeric_max<${type}>() == std::numeric_limits<${type}>::max(), "");
+    static_assert(numeric_lowest<${type}>() == std::numeric_limits<${type}>::lowest(), "");
```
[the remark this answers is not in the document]
jerryyin: Ha good point... Otherwise why not directly call `numeric_limits<>()`.

> {"max", as.max()}, {"lowest", as.min()}

That was my first attempt yesterday. But turns out it doesn't know how to convert `as.max()` or `as.min()` into a string. So what's the utility that would convert a arbitrary type into a string correctly?

src/targets/gpu/kernels/include/migraphx/kernels/type_traits.hpp
jerryyin: That sounds to be a good idea!

test/gpu/jit.cpp
jerryyin: I was using `==` to do the comparison between two values, and hip compilation would error out saying that I shouldn't do that.

I'm wondering if there's a kernel utility in migraphx that can tell if the two float point values are close enough?

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
```diff
@@ -79,20 +79,21 @@ __device__ void dpp_reduce(T& in, Op op)
...
     }                                                                                              \
     __device__ inline void dpp_reduce(uint32_t& x, op) { MIGRAPHX_DPP_REDUCE_ASM(x, prefix##_u32); }
 
-MIGRAPHX_DPP_REDUCE(op::sum, v_add)
-MIGRAPHX_DPP_REDUCE(op::max, v_max)
-MIGRAPHX_DPP_REDUCE(op::min, v_min)
-MIGRAPHX_DPP_REDUCE(op::product, v_mul)
+// Note: when max and min are in int32_t, signed version of instruction needs to be used.
+MIGRAPHX_DPP_REDUCE(op::sum, v_max, _u)
+MIGRAPHX_DPP_REDUCE(op::product, v_min, _u)
```
jerryyin, line 94: Typo: Should be `v_add` and `v_mul`

test/gpu/jit.cpp
pfultz2: I think it would be better to disable the warning for this test. I think you can do `options.params = "-Wno-float-equal"`.

causten:
fixed with pr #1721

test/verify/test_scatter0.cpp
jerryyin: Hmmm... I made an attempt to change the implementation like, if that's what you suggest:

```cpp
hip_visit_all(result, arg0)([&](auto output, auto data) {
   ...
   auto s1 = arg1.get_shape();
   gs_launch...
       auto out_idx = s1.mult(i);
       ...
       out_idx[axis] = index;
```
 Then the compiler complaints that `out_idx[axis] = index` doesn't work because I'm trying to call operator[] on device. 

I don't quite get why all of the sudden this becomes an issue for the compiler... Does passing it as an argument (the original way of implementation) implicitly copy the vector to the device into a device-compatible type (which has `operator[]` implemented?

test/gpu/jit.cpp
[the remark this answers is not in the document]
jerryyin: Done in dc852557c2ab6cb6c44e4ac140225c1413906a26

src/targets/gpu/kernels/include/migraphx/kernels/reduce.hpp
jerryyin: Done in dc852557c2ab6cb6c44e4ac140225c1413906a26

src/targets/gpu/kernels/include/migraphx/kernels/type_traits.hpp
jerryyin: Done in dc852557c2ab6cb6c44e4ac140225c1413906a26

test/gpu/jit.cpp
jerryyin: Done in dc852557c2ab6cb6c44e4ac140225c1413906a26

test/verify/test_scatter0.cpp
pfultz2: > Then the compiler complaints that out_idx[axis] = index doesn't work because I'm trying to call operator[] on device.

The shape class doesn't run on the device, and uses dynamically allocated memory on the host. Thats why you need to pass it to the visitor to get `hip_shape` class which will run on the device and all the memory is statically allocated, so when the lambda passed to `gs_launch` captures the class by value, its copied to the device.

jerryyin: > Thats why you need to pass it to the visitor to get hip_shape class which will run on the device

Are we back from where we started? Or did I captured your initial suggestion wrong?

> We can indices.get_shape() instead of s1.

From this point my understanding is that we need to get rid of this s1 argument such as to avoid the `type must match` requirement from `hip_visit_all_impl()`. But then following the conversation, getting rid of s1 won't work because it will make it the wrong `shape` type instead of `hip_shape` type. And now we are back to the need to pass the shape again to the first `hip_visit_all()`. I'm still confused about the proper way to fix the original type must match requirement.

If you have a good idea of what needs to be done, would you mind to push the commit so this becomes a collaborated effort? I think I understand why the existing kernel is implemented but to make code changes here still seems to be very tricky and it constantly results in ten pages of compiling error. Or if it is convenient for you, we can have another meeting.

pfultz2: > Are we back from where we started? Or did I captured your initial suggestion wrong?

The suggestion was to use `indices.get_shape()` instead of `s1`. 

> From this point my understanding is that we need to get rid of this s1 argument such as to avoid the type must match requirement from hip_visit_all_impl(). 

This will fix the first `hip_visit_all`, but there might be a problem with the second. It should match the visits that are used in the ref implementation where `result`, `arg0`(the data), and `arg2`(the update tensor) are in the same visit and the `indices` are a seperate visit:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/op/scatter.hpp#L82

test/gpu/jit.cpp
```diff
@@ -356,4 +356,71 @@ TEST_CASE(compile_math)
...
+        std::string max = "";
+        // Note 9223372036854775808 is a constant literal that is outside the range of long long
+        // type For the same reason, 18446744073709551616 needs postfix ULL to be parsed correctly
+        auto min_override_int64  = "(-9223372036854775807LL - 1)";
+        auto max_override_uint64 = "18446744073709551615ULL";
+
+        migraphx::shape::visit(t, [&](auto as) {
+            if(t == migraphx::shape::int64_type)
+            {
+                min = min_override_int64;
```
pfultz2, line 398: Cant you do `std::to_string(as.min()) + "LL"`?

test/gpu/jit.cpp
```diff
@@ -356,4 +356,71 @@ TEST_CASE(compile_math)
...
+                max = std::to_string(as.max());
+            }
+            else if(t == migraphx::shape::uint64_type)
+            {
+                min = std::to_string(as.min());
+                max = max_override_uint64;
```
pfultz2, line 404: Cant you do `std::to_string(as.max()) + "ULL"`?

test/gpu/jit.cpp
```diff
@@ -356,4 +356,71 @@ TEST_CASE(compile_math)
...
+
+        std::string min = "";
+        std::string max = "";
+        // Note 9223372036854775808 is a constant literal that is outside the range of long long
+        // type For the same reason, 18446744073709551616 needs postfix ULL to be parsed correctly
+        auto min_override_int64  = "(-9223372036854775807LL - 1)";
+        auto max_override_uint64 = "18446744073709551615ULL";
+
+        migraphx::shape::visit(t, [&](auto as) {
+            if(t == migraphx::shape::int64_type)
```
pfultz2, line 396: I dont think you need `if` statements here. I believe adding `LL` and `ULL` unconditionally should work for all types.

jerryyin: OKay, yes, it seems reasonable to always add `LL` for minimal and `ULL` for max. Then I'd still need to `if` based on `int64_type` to override with `min_override_int64` because in that scenario without override it will give a `-9223372036854775808` which the compiler is not happy about.

pfultz2: > because in that scenario without override it will give a -9223372036854775808 which the compiler is not happy about.

What error do you get from the compiler?

jerryyin: The error I get is: `error: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Werror,-Wimplicitly-unsigned-literal]`. 

As far as I know `"(-9223372036854775807LL - 1)"` is a reasonable way to override it.

Besides, the need for `LL` and `ULL` really only exist in 64 bit integer types. I thought it is going to be more clear that we take them as special cases. 

Now, because of this compiler integer parsing limitations, does this make `as.min()` and `as.max()` problemetic for 64 bit integer as eventually, those are replaced as-is to string as `${max}` and `${min}`? It looks to be a caveat exposed by this new unit test we've written.

jerryyin: Talked with @pfultz2 and inline the literal in ec8c252. Please feel free to provide any additional feedback.

test/gpu/jit.cpp
```diff
@@ -356,4 +356,69 @@ TEST_CASE(compile_math)
...
+
+        migraphx::shape::visit(t, [&](auto as) {
+            std::string min = "";
+            std::string max = "";
+            // Note 9223372036854775808 is a constant literal that is outside the range of long
+            // long type For the same reason, 18446744073709551616 needs postfix ULL to be parsed
+            // correctly
+            if(t == migraphx::shape::int64_type)
+            {
+                min = "(-9223372036854775807LL - 1)";
```
pfultz2, line 396: Do `min = "(" + std::to_string(as.min() + 1) + "LL - 1)"`

test/gpu/jit.cpp
```diff
@@ -356,4 +356,69 @@ TEST_CASE(compile_math)
...
+                max = std::to_string(as.max());
+            }
+            else if(t == migraphx::shape::uint64_type)
+            {
+                min = std::to_string(as.min());
+                max = "18446744073709551615ULL";
```
pfultz2, line 402: Do `max = std::to_string(as.max()) + "ULL"`

test/gpu/jit.cpp
jerryyin: Sounds reasonable, this avoids the hard-coded literal. Done in ae927f1

pfultz2 (approved):

jerryyin:
@causten Gentle ping, I believe this PR is ready to merge.

TedThemistokleous (approved):

PR #1722, opened 2023-04-27
Add migraphx.create_argument to python API

pfultz2:


shivadbhavsar (commented):
Its already possible create an argument object with user-defined data in the current API using `migraphx.argument(...)`. Whats the use case for having this option, is there a case where np arrays aren't sufficient for doing this?

pfultz2:
> Its already possible create an argument object with user-defined data in the current API using migraphx.argument(...). Whats the use case for having this option, is there a case where np arrays aren't sufficient for doing this?

This is to enable the driver output to create the same scalars when using `--py`(currently it just generates random values), and this output should work with only `import migraphx`, which is why using numpy wouldn't work.

shivadbhavsar (approved):

doc/src/reference/py.rst
```diff
@@ -121,6 +121,15 @@ argument
 
     :rtype: argument
 
+.. py:function:: create_argument(s, values)
+
+    Create an argument of shape s with a set of values.
+
+    :param shape s: Shape of argument to create.
+    :param list values: Values to put in the argument. Must be the same number of elements as the shape.
```
umangyadav, line 129: > Must be the same number of elements as the shape

This is not enforced. Assert inside `fill()` checks for `<=`.

CharlieL7 (approved):
Approved, should probably update the doc wrt Umang's comment.

doc/src/reference/py.rst
pfultz2: Added a check for the size.

src/py/migraphx_py.cpp
```diff
@@ -547,6 +547,13 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
         py::arg("format") = "msgpack");
 
     m.def("get_target", &migraphx::make_target);
+    m.def("create_argument", [](const migraphx::shape& s, const std::vector<float>& values) {
```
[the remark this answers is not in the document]
pfultz2: Switched to use double.

umangyadav (approved):

PR #1723, opened 2023-04-27
Python API update for dynamic batch

CharlieL7:
* Relies on #1711 
* Updates the python API to use dynamic shapes and dynamic batch
* Only works with `offload_copy` set to True for now
  * Need updates to create tuple shapes and handle parameter list with tuple arguments
  * Also will probably hide the `to_gpu` type of methods
* Changed serialization of dyn_dims in a shape to use reflect method


src/py/migraphx_py.cpp
```diff
@@ -95,6 +95,11 @@ void visit_py(T x, F f)
     {
         f(x.template cast<std::string>());
     }
+    else if(py::isinstance<migraphx::shape::dynamic_dimension>(x))
+    {
+        // skip serializing dynamic_dimension objects
+        f("");
```
pfultz2, line 101: This should serailize the object.

src/py/migraphx_py.cpp
```diff
@@ -239,33 +247,54 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
...
-            auto t    = migraphx::shape::parse_type(v.get("type", "float"));
-            auto lens = v.get<std::size_t>("lens", {1});
-            if(v.contains("strides"))
-                return migraphx::shape(t, lens, v.at("strides").to_vector<std::size_t>());
+            auto v = migraphx::to_value(kwargs);
+            auto t = migraphx::shape::parse_type(v.get("type", "float"));
+            if(v.contains("dyn_dims"))
+            {
+                auto dyn_dims =
+                    kwargs["dyn_dims"].cast<std::vector<migraphx::shape::dynamic_dimension>>();
```
pfultz2, line 255: This should do `from_value<std::vector<migraphx::shape::dynamic_dimension>>(v)`.

pfultz2 (approved):

shivadbhavsar (approved):

src/shape.cpp
```diff
@@ -706,9 +706,10 @@ void migraphx_from_value(const value& v, shape& s)
         {
             auto v_dd = v.at("dynamic_dimensions");
             std::vector<shape::dynamic_dimension> dyn_dims(v.at("dynamic_dimensions").size());
-            std::transform(v_dd.begin(), v_dd.end(), dyn_dims.begin(), [](migraphx::value x) {
-                return from_value<shape::dynamic_dimension>(x);
-            });
+            std::transform(
+                v_dd.begin(), v_dd.end(), dyn_dims.begin(), [](const migraphx::value& x) {
+                    return from_value<shape::dynamic_dimension>(x);
+                });
```
pfultz2, line 712: I think this whole thing can be replaced with `from_value(v.at("dynamic_dimensions"), dyn_dims)` or `auto dyn_dims = from_value<std::vector<shape::dynamic_dimension>>(v.at("dynamic_dimensions"))`.

PR #1729, opened 2023-05-02
adjust docker files to support new rocm 5.5

causten:
PR to upgrade MIGraphX CI infrastructure to use ROCM 5.5


umangyadav:
Need to rebuild Performance test docker image.

umangyadav:
> Need to rebuild Performance test docker image.

also would need to tune MIOpen DBs

umangyadav:
Rocm-5.5 requires one more workaround compilation flag `-Wno-unused-parameter` to be added during compilation. But following test_verify tests are failing for some reason. I took `test_slice2` as case study for comparision with rocm-5.4.2. I enabled, `MIGRAPHX_TRACE_HIPRTC`, `MIGRAPHX_GPU_DEBUG`, `MIGRAPHX_GPU_DUMP_SRC` and `MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS`.  Logs for both rocm-5.4.2 and rocm-5.5 comes out to be same for `test_slice2`. 

Not sure what is wrong.  I had to add some logic for `MIGRAPHX_GPU_DUMP_SRC` for hiprtc. 

[  FAILED  ] 10 tests failed
[  FAILED  ] test_add_conv_constant
[  FAILED  ] test_conv_group_add
[  FAILED  ] test_conv_pooling
[  FAILED  ] test_pad
[  FAILED  ] test_pad_transposed
[  FAILED  ] test_reduce_op_large<migraphx::op::reduce_prod, 2, migraphx::shape::float_type>
[  FAILED  ] test_roialign
[  FAILED  ] test_roialign_nondefault
[  FAILED  ] test_roialign_nonstandard
[  FAILED  ] test_slice2

test/gpu/jit.cpp
```diff
@@ -206,8 +206,10 @@ TEST_CASE(compile_warnings)
     EXPECT(not compile("").empty());
     EXPECT(not compile("-Wunused-parameter -Wno-error").empty());
     EXPECT(not compile("-Wno-unused-parameter -Werror").empty());
+#ifndef MIGRAPHX_USE_HIPRTC
     EXPECT(test::throws([&] { compile("-Werror=unused-parameter"); }));
     EXPECT(test::throws([&] { compile("-Wunused-parameter -Werror"); }));
+#endif
```
pfultz2, line 212: We should only disable these tests when the workarounds are enabled.

umangyadav: Done.

test/gpu/jit.cpp
```diff
@@ -206,10 +206,11 @@ TEST_CASE(compile_warnings)
     EXPECT(not compile("").empty());
     EXPECT(not compile("-Wunused-parameter -Wno-error").empty());
     EXPECT(not compile("-Wno-unused-parameter -Werror").empty());
-#ifndef MIGRAPHX_USE_HIPRTC
-    EXPECT(test::throws([&] { compile("-Werror=unused-parameter"); }));
-    EXPECT(test::throws([&] { compile("-Wunused-parameter -Werror"); }));
-#endif
+    if(migraphx::disabled(migraphx::gpu::MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS{}))
```
pfultz2, line 209: It should be `if (not migraphx::enabled(migraphx::gpu::MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS{}))`. Using `disabled` would require setting the env variable to enable the tests(ie `MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS=0`). We want the tests enabled by default without any env variables and only skipped when `MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS=1` is set.

umangyadav: Done.

pfultz2 (approved):

TedThemistokleous (approved):

src/onnx/parse_randomuniform_ops.cpp
```diff
@@ -35,8 +35,7 @@ namespace onnx {
 
 struct parse_randomuniform_ops : op_parser<parse_randomuniform_ops>
 {
-    const std::set<shape::type_t> valid_types = {
-        shape::float_type, shape::half_type, shape::double_type};
+    std::set<shape::type_t> valid_types = {shape::float_type, shape::half_type, shape::double_type};
```
pfultz2, line 38: This should be static const.

umangyadav: Yes, but i have to make it `static const inline` then.

umangyadav: And then, it complains about : warning: declaration requires a global destructor [-Wglobal-constructors]
    static const inline std::set<shape::type_t> valid_types = {shape::float_type, shape::half_type, shape::double_type};

umangyadav:
With rocm-5.5, clang-tidy is updated to v16. It has some more checks which are failing. They need to be disabled most probably. 
For example, it complains about do-while loop and also asking to avoid const refs. I tried adding check to disable do-while in `.clang-tidy` file, but it seems  order matters for it and couldn't get it to have it correct order to make it work. I would say keep tidy fixes for other PR to get ahead with CI upgrade.

umangyadav:
> > With rocm-5.5, clang-tidy is updated to v16. It has some more checks which are failing. They need to be disabled most probably. For example, it complains about do-while loop and also asking to avoid const refs. I tried adding check to disable do-while in `.clang-tidy` file, but it seems order matters for it and couldn't get it to have it correct order to make it work. I would say keep tidy fixes for other PR to get ahead with CI upgrade.
> 
> Found that, i need to add those checks inside `CmakeLists.txt` to make them appear in correct order.
> 
> One issue is still pending about `-clang-diagnostic-disabled-macro-expansion`

I removed definition for `-Dmain` inside CMakeLists.txt. That seemed to have solve issue with clang-tidy macro expansion. 

@pfultz2  can you take another look ?

src/targets/fpga/vitis_ai_adapter.cpp
```diff
@@ -33,7 +33,7 @@ migraphx::shape x_model::get_shape() const { return shape; };
 
 void x_model::set_shape(migraphx::shape s) { shape = s; }
 
-x_model create_xmodel(const migraphx::module_ref mod)
+x_model create_xmodel(migraphx::module_ref mod)
```
pfultz2, line 36: Use `const_module_ref` instead.

test/gpu/jit.cpp
```diff
@@ -206,8 +206,16 @@ TEST_CASE(compile_warnings)
...
+#ifdef MIGRAPHX_USE_HIPRTC
+    if(not migraphx::enabled(migraphx::gpu::MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS{}))
+    {
+        EXPECT(test::throws([&] { compile("-Werror=unused-parameter"); }));
+        EXPECT(test::throws([&] { compile("-Wunused-parameter -Werror"); }));
+    }
+#else
     EXPECT(test::throws([&] { compile("-Werror=unused-parameter"); }));
     EXPECT(test::throws([&] { compile("-Wunused-parameter -Werror"); }));
+#endif
```
pfultz2, line 218: It would be better to just remove the ifdef and the just do:

```cpp
    if(not migraphx::enabled(migraphx::gpu::MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS{}))
    {
        EXPECT(test::throws([&] { compile("-Werror=unused-parameter"); }));
        EXPECT(test::throws([&] { compile("-Wunused-parameter -Werror"); }));
    }
``` 

I dont think anyone will set `MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS` without using hiprtc in the first place.

umangyadav: > I dont think anyone will set MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS without using hiprtc in the first place. 

It could happen if someone is switching between hipclang and hiprtc locally. They could have have `MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS` env set in their env while using hip-clang and that would disable this test.

pfultz2 (approved):

PR #1740, opened 2023-05-05
[MLIR] Improve MLIR symbols

manupak:
Currently, the mlir kernel symbols indicate whether a dot or a convolution. However, it would be good to see the dimensions and fusions in perf traces.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/878


pfultz2:
> However, it would be good to see the dimensions and fusions in perf traces.

We already show the sizes in the perf report.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -190,6 +191,57 @@ struct find_mlir_op
...
+        {
+            auto a_shape = gemm_based_op->inputs().front()->get_shape();
+            auto b_shape = gemm_based_op->inputs().back()->get_shape();
+
+            auto m = a_shape.lens()[a_shape.lens().size() - 2];
+            name += "_m" + std::to_string(m);
+            auto k = a_shape.lens().back();
+            name += "_k" + std::to_string(k);
+            auto n = b_shape.lens().back();
+            name += "_n" + std::to_string(n);
```
pfultz2, line 208: Dont add the sizes to the MLIR symbol name as this makes the summary list each dot or convolution seperately, which the point of the summary is to show the times for the types of operators and not for each instruction, which is what the perf report already prints out.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -190,6 +191,57 @@ struct find_mlir_op
...
+            if(contains({"@literal", "@param", "@return"}, ins->name()))
+            {
+                continue;
+            }
+            name += "_" + ins->name();
+        }
+        static size_t counter = 0;
+        static std::mutex g_mlirc_mutex; // NOLINT
+        const std::lock_guard<std::mutex> lock(g_mlirc_mutex);
+        name += ":" + std::to_string(counter++);
```
pfultz2, line 241: Same here, dont add a counter here as the summary becomes too large.

manupak: No it wont. this is just to make the module names unique and they are stripped when obtaining the sym_names.

src/targets/gpu/fuse_mlir.cpp
pfultz2: > I was more after getting symbol names in rocprof trace more detailed rather than the perf summary, 

The perf report shows the times for each instruction and you can see the sizes. Also, if you want to use rocprof, we do have `driver roctx` which will insert markers with information about which instruction was executed. Although, I think the perf report is much easier to use.

>  Maybe I can hide this under an option ?

Yea you can add an environment variable.

> From a kernel generator PoV, it would be useful to see breakdown of gemm op rather than collating unrelated gemms

Yea thats why we show the times for each instruction.

src/targets/gpu/fuse_mlir.cpp
pfultz2: Why do we need to make the module names unique?

manupak: I've observed collisions otherwise; even the pointwise modules have this :<number> postfix. I thought it was there to avoid collisions ?

src/targets/gpu/fuse_mlir.cpp
manupak: >  we do have driver roctx
 
 I wasn't aware of this. I ll have a go and see if it suffices.

src/targets/gpu/fuse_mlir.cpp
pfultz2: I see this for migraphx module created. There is no need for a lock, this isn't ran in parallel.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -190,6 +191,57 @@ struct find_mlir_op
...
+            MIGRAPHX_THROW("mlir gemm_based_op has to be a dot or a convolution");
+        }
+        for(auto ins : iterator_for(*pointwise_mod))
+        {
+            if(contains({"@literal", "@param", "@return"}, ins->name()))
+            {
+                continue;
+            }
+            name += "_" + ins->name();
+        }
```
pfultz2, line 237: The symbol name computation should be done in `get_symbol_name` in mlir.cpp.

src/targets/gpu/fuse_mlir.cpp
pfultz2: Actually, this shouldn't be needed as you can use the name from the pointwise module so it should always be unique.

krzysz00:
@manupak @pfultz2 Would y'all be open to narrowing the scope of the PR some so that, for example, a MLIR kernel that does a dot + add + relu becomes `mlir_dot_add_relu` by analogy to how pointwise modules work? This'll be helpful for, when I'm looking at the output of MIGraphX runs, knowing what operations got fused into which MLIR kernel.

causten:
Is this still needed   @manupak  ?

manupak:
I thought I closed this... apologies.

PR #1742, opened 2023-05-05
Test and doc update for shape.from_permutation()

CharlieL7:
* Need to finish updating the tests


src/include/migraphx/shape.hpp
```diff
@@ -156,8 +156,33 @@ struct shape
...
+     * The "original" dimensions of the shape can be considered to be the forward permutation of the
+     * given output dimensions.
+     * The output shape's strides are the inverse permutation of the standard strides generated by
+     * the "original" dimensions.
+     *
+     * 2D example:
+     *   parameters:
+     *     l = [2, 3], perm = [1, 0]
+     *   therefore:
+     *     inverse_perm = [1, 0]
```
umangyadav, line 171:
```suggestion
     *     inverse_perm = [0, 1]
```

CharlieL7: No, the inverse permutation is [1, 0]. It needs to swap it back.

src/include/migraphx/shape.hpp
```diff
@@ -156,8 +156,33 @@ struct shape
 
     shape(const std::vector<shape>& subs);
 
+    /**
+     * Creates an output shape with dimensions equal to the input lengths and strides determined
+     * by the permutation argument.
```
pfultz2, line 161: Strides are created so that `find_permutation` will return the permutation that is passed.

src/include/migraphx/shape.hpp
```diff
@@ -156,8 +156,33 @@ struct shape
...
+     * The "original" dimensions of the shape can be considered to be the forward permutation of the
+     * given output dimensions.
+     * The output shape's strides are the inverse permutation of the standard strides generated by
+     * the "original" dimensions.
```
pfultz2, line 165: This is really confusing. The strides are set so that `find_permutation` returns the same permutation. There is no need to mention an inverse permutation, its not really relevant.

src/include/migraphx/shape.hpp
```diff
@@ -156,8 +156,33 @@ struct shape
...
+     *
+     * 2D example:
+     *   parameters:
+     *     l = [2, 3], perm = [1, 0]
+     *   therefore:
+     *     inverse_perm = [1, 0]
+     *     original_shape = {lens = [3, 2], strides = [2, 1]}
+     *     output_shape = {lens = [2, 3], strides = [1, 2]
```
pfultz2, line 173: The lens should be 3, 2, because `from_permutation` doesn't change the order of the lens.

src/include/migraphx/shape.hpp
pfultz2: Just remove this.

test/shape_test.cpp
```diff
@@ -956,4 +956,24 @@ TEST_CASE(test_multi_index)
     EXPECT(migraphx::verify_range(s.multi(34), std::vector<size_t>{1, 1, 4}));
 }
 
+TEST_CASE(from_2d_permutation)
+{
+    std::vector<std::size_t> out_lens = {2, 3};
+    std::vector<int64_t> permutation  = {1, 0};
+    migraphx::shape out_shape =
+        migraphx::shape::from_permutation(migraphx::shape::float_type, out_lens, permutation);
+    EXPECT(migraphx::verify_range(out_shape.lens(), out_lens));
```
pfultz2, line 965: Use equality not `verify_range`.

test/shape_test.cpp
```diff
@@ -956,4 +956,24 @@ TEST_CASE(test_multi_index)
...
+    EXPECT(migraphx::verify_range(out_shape.strides(), std::vector<std::size_t>{1, 2}));
```
pfultz2, line 966: Dont check the strides directly. Instead check `find_permutation(out_shape) == permutation`.

src/include/migraphx/shape.hpp
CharlieL7: This is correct, same as from_2d_permutation testcase.

test/shape_test.cpp
CharlieL7: I'm going to make additional tests for `find_permutation(s)`. It's not clear to me what it's doing right now.

src/include/migraphx/shape.hpp
CharlieL7: I kept the inverse_perm explanation as that's what the code is doing.

src/include/migraphx/permutation.hpp
```diff
@@ -56,12 +56,12 @@ inline std::vector<int64_t> sort_permutation(const Vector& data, Op op)
...
 
 /*!
- * Returns the permutation needed to apply to the shape to undo the current permutation
+ * Returns the inverse permutation that could be applied to reverse the inputted permutation
  */
 std::vector<int64_t> invert_permutation(const std::vector<int64_t>& permutation);
 
 /*!
- * Finds the permutation most likely from a transpose operator that has been applied to the shape.
+ * Finds the permutation that would make the shape standard
```
umangyadav, line 64: not necessarily `standard`. But such that strides are in descending order.

src/include/migraphx/permutation.hpp
```diff
@@ -56,12 +56,12 @@ inline std::vector<int64_t> sort_permutation(const Vector& data, Op op)
...
 
 /*!
- * Returns the permutation needed to apply to the shape to undo the current permutation
+ * Returns the inverse permutation that could be applied to reverse the inputted permutation
  */
 std::vector<int64_t> invert_permutation(const std::vector<int64_t>& permutation);
 
 /*!
- * Finds the permutation most likely from a transpose operator that has been applied to the shape.
+ * Finds the permutation that would put the strides in descending order
```
pfultz2, line 64: This seems backwards. The permutation is from the transpose(or reorder_shape) when the strides were in descending order. The permutation cannot be used to put the strides in descending order, you need `invert_permutation` for that.

Also, it would be better to say "not transposed"(like `s.transposed()`) instead of strides in descending order, which isn't clear.

CharlieL7: The tests I made for `find_permutation` does make it look like it gives a permutation that puts the strides in a "not transposed" order.

src/include/migraphx/permutation.hpp
```diff
@@ -56,12 +56,12 @@ inline std::vector<int64_t> sort_permutation(const Vector& data, Op op)
 }
 
 /*!
- * Returns the permutation needed to apply to the shape to undo the current permutation
+ * Returns the inverse permutation that could be applied to reverse the inputted permutation
```
pfultz2, line 59: It doesn't reverse a permutation. Doing `invert_permutation({0,1,2,3})` is not 3,2,1,0, rather its 0,1,2,3.

src/include/migraphx/shape.hpp
pfultz2: In the test `out_shape.lens() == out_lens` so the order of the dimensions do not change.

CharlieL7: Yeah, the input `l` is [2, 3] the output_shape has lens = [2, 3]. What I mean by "original" shape is the un-transposed shape.

src/include/migraphx/permutation.hpp
pfultz2: This is the permutation needed to put the `lens` in the same order as they are in memory. So if we start with `NCHW`, each letter could be `0,1,2,3` so then `NHWC` would be `0,2,3,1`. So applying a transpose with the permutation would put the shape in the same order as it is in memory. 

For example, with a NHWC tensor like `{1, 3, 16, 16}, {768, 1, 48, 3}`. The dimensions are in NCHW order, but in memory they are in NHWC order. So `find_permutation` will return `0,2,3,1` and applying a transpose(or `reorder_shape`) with this permutation will yield `{1, 16, 16, 3}, {768, 48, 3, 1}`.

src/include/migraphx/shape.hpp
pfultz2: Its an implementation detail(there are other ways this could be implemented) and doesnt explain what the function is doing.

src/include/migraphx/permutation.hpp
CharlieL7: Changing it to "undo" rather than "reverse".

PR #1743, opened 2023-05-05
Optimize file space of github runners

causten:
Remove various file systems not required for what we use Github runners for


umangyadav:
It is passing on this one. I think it problem on github runner. we probably just need to re-run.

PR #1751, opened 2023-05-12
Enable eval to handle multiple contexts

pfultz2:
This is to help enable multi-target execution. We store a vector of targets and contexts. Currently this will only compile a single target, the PR #1672 is needed to enable multiple targets.

This will also serialize the targets and contexts.

When using the `execution_environment` or `prog.get_context()` it will always use the context from the first target assuming this is the "primary" target. Although, its unlikely a user would use `execution_environment` with a multi-target environment.


src/program.cpp
```diff
@@ -390,44 +383,24 @@ std::vector<argument> generic_eval(const module* mod,
...
-#endif
+    auto& contexts = this->impl->contexts;
 
     auto trace_level = value_of(MIGRAPHX_TRACE_EVAL{});
     std::vector<argument> ret;
 
     if(exec_env.async)
     {
-        ctx.wait_for(exec_env.queue);
+        contexts.front().wait_for(exec_env.queue);
```
umangyadav, line 403: Perhaps add assert that async execution is not using multi-targets before just using `front()` context ?

src/include/migraphx/instruction.hpp
```diff
@@ -136,6 +136,8 @@ struct instruction
 
     operation normalized_operator() const;
 
+    unsigned int get_target_id() const;
```
umangyadav, line 139: Make it `uint32` ?

src/program.cpp
```diff
@@ -440,67 +413,64 @@ std::vector<argument> program::eval(parameter_map params, execution_environment
...
+            auto result = f();
+            double t1   = t.record<milliseconds>();
+            ctx.finish();
+            double t2 = t.record<milliseconds>();
+            std::cout << "Time: " << t1 << "ms, " << t2 << "ms" << std::endl;
+            if(trace_level > 1 and ins->name().front() != '@' and ins->name() != "load" and
+               not result.empty())
+            {
+                const target& tgt = this->impl->targets[ins->get_target_id()];
+                auto buffer       = tgt.copy_from(result);
```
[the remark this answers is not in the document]
pfultz2: Yea we probably need an attribute for `offload_copy_from` so we can check for this in tracing. This should be opened as a seperate issue though.

src/program.cpp
```diff
@@ -152,7 +146,7 @@ std::vector<shape> program::get_output_shapes() const
     return mm->get_output_shapes();
 }
 
-context& program::get_context() const { return impl->ctx; }
+context& program::get_context() const { return impl->contexts.front(); }
```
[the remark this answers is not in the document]
pfultz2: I think this is to support the `experimental_get_context` API. We probably should update it for multiple contextes.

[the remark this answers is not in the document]
TedThemistokleous: Couldn't we just add another overload of target_id=0 still though or is it looking for a call without an argument?

``` 
program::get_context(size_t target_id = 0 ) const {
    assert(target_id < impl->context.size());
    return impl->context.front() + target_id}
    }
```

src/program.cpp
```diff
@@ -539,67 +511,64 @@ std::vector<argument> program::eval(parameter_map params, execution_environment
...
 
     return ret;
 }
 
-const int program_file_version = 5;
+void program::finish() const
+{
+    for(const auto& ctx : this->impl->contexts)
+        ctx.finish();
+}
```
TedThemistokleous, line 562: Are we going to assume then, that all context are running in a syncrhonous mode and not one of these will be async?

TedThemistokleous (commented):
Fix issue with CI, some questions in comments

src/program.cpp
```diff
@@ -299,24 +292,24 @@ void program::compile(const std::vector<target>& targets, std::vector<compile_op
                 MIGRAPHX_THROW("Dangling reference in module " + current_mod->name() +
                                " from instruction " + std::to_string(index));
             }
-            current_mod->finalize(this->impl->contexts[root_target_id]);
         }
     }
+    this->finalize();
```
umangyadav, line 297: This would only finalize main module.

src/program.cpp
```diff
@@ -539,67 +511,64 @@ std::vector<argument> program::eval(parameter_map params, execution_environment
...
 value program::to_value() const
 {
     value result;
-    result["version"] = program_file_version;
-    result["target"]  = this->impl->target_name;
-    if(not this->impl->target_name.empty())
-        result["context"] = this->impl->ctx.to_value();
+    result["version"]  = program_file_version;
+    result["targets"]  = migraphx::to_value(this->impl->targets);
+    result["contexts"] = migraphx::to_value(this->impl->contexts);
```
umangyadav, line 571: I think we want to serialize compile options for each target as well.

src/program.cpp
pfultz2: It will finalize all the submodules to main as well:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/module.cpp#L666

src/program.cpp
pfultz2: They might all be async, which is why we are calling `finish`.

src/program.cpp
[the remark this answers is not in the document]
umangyadav: > I think this is to support the experimental_get_context API. We probably should update it for multiple contextes. 

okay, we can add assert for `impl->contexts.size() ==1`. for now.

TedThemistokleous (approved):

TedThemistokleous (changes requested):
Fix CI

src/program.cpp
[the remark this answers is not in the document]
umangyadav: can we add an assert here ?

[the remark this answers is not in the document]
pfultz2: There is already an assert here: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1751/files#diff-fef0ccbb040b2743949266b2829d1a8e63dcb627a6d3af422c41b435e6b03669R164

umangyadav (approved):

TedThemistokleous (approved):

PR #1752, opened 2023-05-15
Update xdlops/rocblas fp32 arch

kahmed10:
I've implemented a more general check for gfx908 and above, so please provide feedback. If this doesn't seem as portable, I can go back to the other implementation and add the new arch name in.


src/targets/gpu/mlir.cpp
```diff
@@ -164,10 +164,10 @@ std::string mlir_print(F f, T x)
     return ss.str();
 }
 
-const std::unordered_set<std::string>& get_xdlops_archs()
+bool has_xdlops(const std::string& target_arch)
 {
-    static std::unordered_set<std::string> supported_archs{"gfx908", "gfx90a"};
-    return supported_archs;
+    const auto device_name = trim(split_string(target_arch, ':').front());
+    return (contains(device_name, "gfx9") and device_name >= "gfx908");
```
umangyadav, line 170: There is no set logic on gfx arch name e.g. it could be `gfx90a` or `gfx1030`. IMO `>=` in such cases is not good check. 

is there a way to retrieve such information by calling any HIP API or by other means ?

pfultz2, line 170: It would be better to use `starts_with(device_name, "gfx9")`.

pfultz2: > There is no set logic on gfx arch name e.g. it could be gfx90a or gfx1030. IMO >= in such cases is not good check.

The comparison only happens for gfx9.

umangyadav: > The comparison only happens for gfx9. 

`target_arch` is set by `get_device_name()` or `gcn.ArchName`.  Would this work for MI300?

kahmed10: Fixed

kahmed10: > > The comparison only happens for gfx9.
> 
> `target_arch` is set by `get_device_name()` or `gcn.ArchName`. Would this work for MI300?

Yes, this change will work for anything gfx908 and above. So currently announced hardware and potential future hardware.

umangyadav (approved):

PR #1755, opened 2023-05-16
Add option to use type erased matchers to reduce symbol names

pfultz2:
This is needed on windows because the large symbol names created from the matchers causes the compiler to hang. This will type erase the matchers to avoid long symbol names, but compilation will be slower.


pfultz2:
There is a slowdown of compilation with this change, which is why its conditionally enabled. Resnet50 goes from 7s to 12s and bert goes from 13s to 26s, so its almost a 2x slowdown.

TedThemistokleous (approved):

umangyadav (approved):

PR #1758, opened 2023-05-18
Update cpp generator to handle inf from  float

TedThemistokleous:
Use std::numeric_limits<type>::min/max() functions plus the appropriate value to encode -inf/inf so that when code is generated we get the following:

```
auto a = float (( numeric_limits<float>::max() + 1);
auto b = float (numeric_limits<float>::min() - 1);
```

vs

```
auto a = float(inf);
auto b = float(-inf);
```

which can't be assigned. Originally tried to use std::numeric_limits<type>::inifnity() but that still causes a conversion to inf/-inf before again assigning it to the creation of float incorrectly

### DIFF
--- a/src/cpp_generator.cpp
+++ b/src/cpp_generator.cpp
@@ -207,19 +207,23 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
     });
     f.set_name(name).set_types(m).set_body(
         m, [&](instruction_ref ins, const auto& names) -> std::string {
-            if(ins->name() == "@literal"){
-                auto string_literal = ins->get_literal().to_string();
-
-                if (string_literal == "inf") {
-                    string_literal = "1.0 / 0.0";
-                }
-
-                if(string_literal == "-inf") {
-                    string_literal = "- 1.0 / 0.0";
-                }
-
-                auto s = shape::cpp_type(ins->get_shape().type()) + "(" +
-                    string_literal + ")";
+            if(ins->name() == "@literal") {
+                std::string string_literal;
+                ins->get_literal().visit([&](auto v) {
+                    assert(v.size() == 1);
+                    auto x = v.front();
+                    if(x > __builtin_huge_valf or x < - __builtin_huge_valf) 
+                    {
+                        string_literal = "__builtin_huge_valf";
+                        if(x < 0)
+                            string_literal = "- __builtin_huge_valf";
+                    }
+                    else if(std::isnan(x))
+                        string_literal = "__builtin_nanf";
+                    else
+                        string_literal = std::to_string(x);
+                });
+                auto s = shape::cpp_type(ins->get_shape().type()) + "(" + string_literal + ")";
 
                 return s;
             }
--- a/src/cpp_generator.cpp
+++ b/src/cpp_generator.cpp
@@ -207,19 +207,20 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
     });
     f.set_name(name).set_types(m).set_body(
         m, [&](instruction_ref ins, const auto& names) -> std::string {
-            if(ins->name() == "@literal") {
+            if(ins->name() == "@literal")
+            {
                 std::string string_literal;
                 ins->get_literal().visit([&](auto v) {
                     assert(v.size() == 1);
                     auto x = v.front();
-                    if(x > __builtin_huge_valf or x < - __builtin_huge_valf) 
+                    if(std::isinf(x))
                     {
-                        string_literal = "__builtin_huge_valf";
+                        string_literal = "__builtin_huge_valf()";
                         if(x < 0)
-                            string_literal = "- __builtin_huge_valf";
+                            string_literal = "-__builtin_huge_valf()";
                     }
                     else if(std::isnan(x))
-                        string_literal = "__builtin_nanf";
+                        string_literal = "__builtin_nanf()";
                     else
                         string_literal = std::to_string(x);
                 });
--- a/src/cpp_generator.cpp
+++ b/src/cpp_generator.cpp
@@ -222,7 +222,7 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
                     else if(std::isnan(x))
                         string_literal = "__builtin_nanf()";
                     else
-                        string_literal = std::to_string(x);
+                        string_literal = ins->get_literal().to_string();
                 });
                 return shape::cpp_type(ins->get_shape().type()) + "(" + string_literal + ")";
             }
--- a/src/cpp_generator.cpp
+++ b/src/cpp_generator.cpp
@@ -31,7 +31,6 @@
 #include <migraphx/iterator_for.hpp>
 #include <map>
 #include <sstream>
-#include <limits>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
@@ -212,11 +211,11 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
                 auto string_literal = ins->get_literal().to_string();
 
                 if (string_literal == "inf") {
-                    string_literal =  std::numeric_limits<decltype(shape::cpp_type(ins->get_shape().type()))>::max() + "+ 1";
+                    string_literal = "1.0 / 0.0";
                 }
 
                 if(string_literal == "-inf") {
-                    string_literal =  std::numeric_limits<decltype(shape::cpp_type(ins->get_shape().type()))>::min() + "- 1";
+                    string_literal = "- 1.0 / 0.0";
                 }
 
                 auto s = shape::cpp_type(ins->get_shape().type()) + "(" +
--- a/src/cpp_generator.cpp
+++ b/src/cpp_generator.cpp
@@ -224,9 +224,7 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
                     else
                         string_literal = std::to_string(x);
                 });
-                auto s = shape::cpp_type(ins->get_shape().type()) + "(" + string_literal + ")";
-
-                return s;
+                return shape::cpp_type(ins->get_shape().type()) + "(" + string_literal + ")";
             }
             auto s = g(ins, names);
             if(impl->fresult)
--- a/src/cpp_generator.cpp
+++ b/src/cpp_generator.cpp
@@ -31,6 +31,7 @@
 #include <migraphx/iterator_for.hpp>
 #include <map>
 #include <sstream>
+#include <limits>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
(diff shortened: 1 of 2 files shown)


src/cpp_generator.cpp
```diff
@@ -207,9 +208,22 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
...
     f.set_name(name).set_types(m).set_body(
         m, [&](instruction_ref ins, const auto& names) -> std::string {
-            if(ins->name() == "@literal")
-                return shape::cpp_type(ins->get_shape().type()) + "(" +
-                       ins->get_literal().to_string() + ")";
+            if(ins->name() == "@literal"){
+                auto string_literal = ins->get_literal().to_string();
+
+                if (string_literal == "inf") {
+                    string_literal =  std::numeric_limits<decltype(shape::cpp_type(ins->get_shape().type()))>::max() + "+ 1";
```
pfultz2, line 215: This is wrong. The type of `decltype(shape::cpp_type(ins->get_shape().type()))` is `shape::type_t` not a floating point. And this probably points to a string in random memory.

You need to use a visitor here:

```cpp
std::string string_literal;
auto cpp_type = shape::cpp_type(ins->get_shape().type());
ins->get_literal().visit([&](auto v) {
    assert(v.size() == 1); 
    auto x = v.front();
    if (std::isinf(x)) {
        string_literal = "std::numeric_limits<" + cpp_type + ">::infinity()";
    } else if (std::isnan(x)) {
        string_literal = "std::numeric_limits<" + cpp_type + ">::quietNaN()";
    } else {
        string_literal = std::to_string(x);
    }
});
```

Although this wont work with hiprtc.

pfultz2:
We should add a test to `test_verify` that uses an infinity literal. We could probably pass it to `isnan` operator so we avoid nans in the output.

src/cpp_generator.cpp
TedThemistokleous: Right. Totally missed the visit part. I think we can get around using std::numeric limits by just performing the floating point division between 1.0/0.0 with proper sign here since that will get evaluated at compile as INF correctly when the new code is generated. This should work for both float and double, since and inf promoted from float to double are treated the same

src/cpp_generator.cpp
```diff
@@ -207,9 +207,22 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
     });
     f.set_name(name).set_types(m).set_body(
         m, [&](instruction_ref ins, const auto& names) -> std::string {
-            if(ins->name() == "@literal")
-                return shape::cpp_type(ins->get_shape().type()) + "(" +
-                       ins->get_literal().to_string() + ")";
+            if(ins->name() == "@literal"){
+                auto string_literal = ins->get_literal().to_string();
+
+                if (string_literal == "inf") {
```
pfultz2, line 213: We shouldn't rely on the string representation which could change, we should just check with `std::isinf`.

src/cpp_generator.cpp
```diff
@@ -207,19 +207,23 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
...
-                }
-
-                auto s = shape::cpp_type(ins->get_shape().type()) + "(" +
-                    string_literal + ")";
+            if(ins->name() == "@literal") {
+                std::string string_literal;
+                ins->get_literal().visit([&](auto v) {
+                    assert(v.size() == 1);
+                    auto x = v.front();
+                    if(x > __builtin_huge_valf or x < - __builtin_huge_valf)
```
pfultz2, line 215: Just do `if(std::isinf(x))` here.

src/cpp_generator.cpp
```diff
@@ -207,19 +207,23 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
...
-                auto s = shape::cpp_type(ins->get_shape().type()) + "(" +
-                    string_literal + ")";
+            if(ins->name() == "@literal") {
+                std::string string_literal;
+                ins->get_literal().visit([&](auto v) {
+                    assert(v.size() == 1);
+                    auto x = v.front();
+                    if(x > __builtin_huge_valf or x < - __builtin_huge_valf) 
+                    {
+                        string_literal = "__builtin_huge_valf";
```
pfultz2, line 217: It should be `__builtin_huge_valf()`.

src/cpp_generator.cpp
```diff
@@ -207,19 +207,23 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
...
+                        if(x < 0)
+                            string_literal = "- __builtin_huge_valf";
+                    }
+                    else if(std::isnan(x))
+                        string_literal = "__builtin_nanf";
```
pfultz2, line 222: It should be `__builtin_nanf()`.

test/verify/test_literal_limits.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm   = p.get_main_module();
+        auto input_s = migraphx::shape(migraphx::shape::float_type, {3, 1});
+        auto infinity_val = std::numeric_limits<float>::infinity();
+        std::vector<float> s_data{infinity_val, -infinity_val, std::numeric_limits<float>::quiet_NaN()};
+        auto input = mm->add_literal(migraphx::literal{input_s, s_data});
+        mm->add_instruction(migraphx::make_op("isnan"), input);
```
pfultz2, line 40: This will all get const folded. Maybe we can multiply the numbers by the input parameter before passing it to `isnan`.

TedThemistokleous: Okay i'll try that.

TedThemistokleous: looks like that did the trick.

```
Run instruction: main:@0 = check_context::migraphx::gpu::context  -> float_type, {}, {}
Time: 0.00601ms, 0.00652ms
Run instruction: main:@1 = hip::hip_copy_literal[id=main:@literal:0] -> float_type, {3, 1}, {1, 1}
Time: 0.00632ms, 0.00679ms
Run instruction: output = @param:output -> bool_type, {3, 1}, {1, 1}
Time: 0.0009ms, 0.00134ms
Run instruction: test_input = @param:test_input -> float_type, {3, 1}, {1, 1}
Time: 0.00023ms, 0.00067ms
Run instruction: main:@4 = gpu::code_object[code_object=13632,symbol_name=mul_isnan_kernel,global=3,local=1024,](main:@1,test_input,output) -> bool_type, {3, 1}, {1, 1}
Time: 0.0245ms, 0.068711ms
Run instruction: main:@0 = check_context::migraphx::gpu::context  -> float_type, {}, {}
Time: 0.00053ms, 0.00104ms
Run instruction: main:@1 = hip::hip_copy_literal[id=main:@literal:0] -> float_type, {3, 1}, {1, 1}
Time: 0.00078ms, 0.00135ms
Run instruction: output = @param:output -> bool_type, {3, 1}, {1, 1}
Time: 0.00041ms, 0.00087ms
Run instruction: test_input = @param:test_input -> float_type, {3, 1}, {1, 1}
Time: 0.00026ms, 0.00079ms
Run instruction: main:@4 = gpu::code_object[code_object=13632,symbol_name=mul_isnan_kernel,global=3,local=1024,](main:@1,test_input,output) -> bool_type, {3, 1}, {1, 1}
Time: 0.00703ms, 0.048141ms

```

TedThemistokleous: Literal looks fine too here:

`Run instruction: main:@0 = @literal{inf, -inf, nan} -> float_type, {3, 1}, {1, 1}
Time: 0.00717ms, 0.00731ms`

src/cpp_generator.cpp
TedThemistokleous: Done

src/cpp_generator.cpp
TedThemistokleous: done

test/verify/test_literal_limits.cpp
TedThemistokleous: Updated and pushed

src/cpp_generator.cpp
TedThemistokleous: Done

src/cpp_generator.cpp
TedThemistokleous: Resolved since we're using builtins__*** now

TedThemistokleous:
> 

This is done as well. Let me know if you want an updated test. I guess maybe i could template this out for double/float then too

TedThemistokleous:
Odd, I seem to be failing now on quantization.cpp

Looks like the tolerance has shrunk from 80 for the default value, to 1e4.

```
[   RUN    ] int8_quantization
0.00099522
```

TedThemistokleous:
Turns out issue wasn't with int8_quant but with using std::to_string(x) instead of inx->get_literal()->to_string() which doesn't truncate the value on conversion. Waiting till CI passes. Will be back later this evening if there's anymore comments

TedThemistokleous:
Rebased this off develop to fire off CI. Had pulled in another branch during debug for int8 changes 'fp16-int8-quantize'

src/cpp_generator.cpp
```diff
@@ -208,8 +208,24 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
...
         m, [&](instruction_ref ins, const auto& names) -> std::string {
             if(ins->name() == "@literal")
-                return shape::cpp_type(ins->get_shape().type()) + "(" +
-                       ins->get_literal().to_string() + ")";
+            {
+                std::string string_literal;
+                ins->get_literal().visit([&](auto v) {
+                    assert(v.size() == 1);
+                    auto x = v.front();
+                    if(std::isinf(x))
```
umangyadav, line 216: Should this be `if constexpr` ?

src/cpp_generator.cpp
```diff
@@ -208,8 +208,24 @@ cpp_generator::function cpp_generator::generate_module(const module& m,
...
+                    {
+                        string_literal = "__builtin_huge_valf()";
```
umangyadav, line 218: would this work for `double` ? `valf` is `float` type.

pfultz2: Yea it should probably be `__builtin_huge_val()`, and it will implicitly convert to the smaller float size.

TedThemistokleous: Done

test/verify/test_literal_limits.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/make_op.hpp>
+#include <limits>
+
```
umangyadav, line 29: Can you add tests for `half`, and `double` as well? I think you can just templatize this. 

For the int32/8, you can create another test or see if you can make use of template for that as well.

TedThemistokleous: Done.

umangyadav (approved):

TedThemistokleous:
> This still has changes from #1757. Looking at the files changed tab still shows changes from #1757.

Odd, my rebase still included your changes since I did a merge. I've backed them out now as Khalique and I originally thought your changes would help fix the accuracy issue when we first saw it.

pfultz2 (approved):

PR #1764, opened 2023-05-19
Use half HIP APIs to compute max and min

umangyadav:
MIGraphX was using float overloads for half types to compute max and mins. HIP has added API for half type max() and min(). We should use that for better perf in softmax.


pfultz2 (approved):

umangyadav:
Performance tests are using rocm-5.4.2 and HIP API for `hmax` and `hmin` doesn't seem to exist there. Need to upgrade  perf test docker to rocm-5.5.

manupak:
This seems like a breaking change for something under 5.x as it will strictly require 5.5 ?
Any suggestion (build or runtime option ?) to make it work with older versions ?

manupak:
Is it possible to guard the usage of halfs (ideally checking the env) ? 
or maybe add a runtime switch to disable that we can use.

Just wanted to check above as it would give some backward compatibility and some slack for us to do the update.

If not yea, we'll have to maintain a branch with the revert and above would be much easier for us if that is possible until we do the upgrade to ROCm-5.5. I ll check what it means to upgrade to ROCm-5.5.

manupak:
Alright.. Im checking with rocm-5.5.

PR #1770, opened 2023-05-23
[Draft] Enable MIGraphX on Windows

apwojcik:
That draft PR illustrates the changes required to build and run MIOpen on Windows. The changes were tested on Linux, however, they are not yet in the CI.

More details on the effort will come soon.


src/CMakeLists.txt
```diff
@@ -22,6 +22,14 @@
 # THE SOFTWARE.
 #####################################################################################
 
+cmake_policy(SET CMP0057 NEW)
+
+include(GenerateExportHeader)
+
+if(NOT HIP_PATH)
+    set(HIP_PATH "$ENV{HIP_PATH}")
```
pfultz2, line 30: `CMAKE_PREFIX_PATH` should be used to tell cmake where hip is. Dont use custom cmake variables for dependencies.

src/CMakeLists.txt
```diff
@@ -226,69 +240,91 @@ register_migraphx_ops(
...
+if(NOT WIN32)
+    find_package(nlohmann_json 3.8.0 REQUIRED)
+endif()
+
 target_link_libraries(migraphx PRIVATE nlohmann_json::nlohmann_json)
 
-find_package(PkgConfig)
-pkg_check_modules(SQLITE3 REQUIRED IMPORTED_TARGET sqlite3)
-target_link_libraries(migraphx PRIVATE PkgConfig::SQLITE3)
+generate_export_header(migraphx
```
pfultz2, line 272: Cant we set `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` instead?

src/CMakeLists.txt
```diff
@@ -226,69 +240,91 @@ register_migraphx_ops(
...
+    target_link_libraries(migraphx PUBLIC msgpackc-cxx)
 endif()
-target_link_libraries(migraphx PRIVATE msgpackc-cxx)
-# Make this available to the tests
-target_link_libraries(migraphx INTERFACE $<BUILD_INTERFACE:msgpackc-cxx>)
-
-add_library(migraphx_all_targets INTERFACE)
 
-set(PACKAGE_DEPENDS)
+add_custom_target(migraphx_all_targets)
```
pfultz2, line 292: Why is this changed? A custom target cant be linked.

src/api/api.cpp
```diff
@@ -442,7 +451,6 @@ struct migraphx_optimals
     std::set<size_t> object;
 };
 
-extern "C" struct migraphx_dynamic_dimension;
```
pfultz2, line 445: These changes will get overwritten when doing `make generate`.

src/targets/cpu/CMakeLists.txt
```diff
@@ -53,46 +54,77 @@ add_library(migraphx_cpu
     target.cpp
     write_literals.cpp
 )
+
+generate_export_header(migraphx_cpu
+    EXPORT_FILE_NAME
+        ${CMAKE_BINARY_DIR}/include/migraphx/cpu/export.h)
```
pfultz2, line 60: Why do we generate another export header instead of reusing the one from migraphx/export.h?

src/targets/gpu/CMakeLists.txt
```diff
@@ -22,43 +22,135 @@
 # THE SOFTWARE.
 # ####################################################################################
 
-list(APPEND CMAKE_PREFIX_PATH /opt/rocm)
```
pfultz2, line 25: This shouldn't be removed.

src/targets/gpu/CMakeLists.txt
```diff
@@ -22,43 +22,135 @@
...
+target_compile_options(migraphx_device
+    PRIVATE
+        -fno-gpu-rdc
+        -Xclang
+        -fallow-half-arguments-and-returns
+        -Wno-shadow-uncaptured-local
+        -Wno-cuda-compat
+        -Wno-unused-command-line-argument
+        -Wno-invalid-command-line-argument
+        -Wno-option-ignored)
```
pfultz2, line 137: Why are these flags repeated?

src/targets/gpu/CMakeLists.txt
```diff
@@ -22,43 +22,135 @@
...
+
+target_compile_features(migraphx_device PRIVATE cxx_std_17)
```
pfultz2, line 139: Dont use `cxx_std_17`. We already set the C++ version at the top level.

src/targets/gpu/CMakeLists.txt
```diff
@@ -125,6 +219,41 @@ add_library(migraphx_gpu
...
+target_compile_options(migraphx_gpu
+    PRIVATE
+        -fno-gpu-rdc
+        -Xclang
+        -fallow-half-arguments-and-returns
+        -Wno-shadow-uncaptured-local
+        -Wno-cuda-compat
+        -Wno-unused-command-line-argument
+        -Wno-invalid-command-line-argument
+        -Wno-option-ignored)
```
pfultz2, line 250: These flags should not be added to `migraphx_gpu`. It doesnt do any gpu code.

src/targets/gpu/CMakeLists.txt
```diff
@@ -181,18 +316,20 @@ if(MIGRAPHX_ENABLE_MLIR)
 endif()
 
 if(MIGRAPHX_USE_HIPRTC)
+    find_package(hiprtc REQUIRED)
     message(STATUS "MIGraphX is using hipRTC")
     target_compile_definitions(migraphx_gpu PRIVATE -DMIGRAPHX_USE_HIPRTC=1)
 else()
+    find_package(hip REQUIRED)
```
pfultz2, line 323: This should not be here. We use hip with hiprtc.

src/targets/gpu/allocation_model.cpp
```diff
@@ -30,18 +30,18 @@ namespace migraphx {
...
 }
 
-operation gpu_allocation_model::preallocate(const shape& s, const std::string& id) const
+operation gpu_allocation_model::preallocate(const shape& s, const std::string& id)
 {
     return make_op("hip::hip_allocate_memory", {{"shape", to_value(s)}, {"id", id}});
 }
 
-std::string gpu_allocation_model::copy() const { return "hip::copy"; }
+std::string gpu_allocation_model::copy() { return "hip::copy"; }
```
pfultz2, line 44: Why is `const` being removed?

src/targets/gpu/analyze_streams.cpp
```diff
@@ -38,17 +38,17 @@ struct hip_stream_model
     std::unordered_map<migraphx::instruction_ref, std::size_t> ins2stream{};
     std::size_t get_nstream() const { return max_stream + 1; }
     std::size_t get_stream(migraphx::instruction_ref ins) const { return ins2stream.at(ins); }
-    std::size_t get_event_id(migraphx::instruction_ref ins) const
+    static std::size_t get_event_id(migraphx::instruction_ref ins)
```
pfultz2, line 41: This should not be static, so it matches the call in the interface.

src/targets/gpu/compile_hip.cpp
```diff
@@ -304,17 +304,17 @@ compile_hip_src(const std::vector<src_file>& srcs, std::string params, const std
     if(has_compiler_launcher())
         compiler.launcher = MIGRAPHX_STRINGIZE(MIGRAPHX_HIP_COMPILER_LAUNCHER);
 #endif
-    if(enabled(MIGRAPHX_GPU_DUMP_SRC{}))
+    if(enabled(MIGRAPHX_GPU_DUMP_SRC))
```
pfultz2, line 307: I dont really understand how this would compile.

src/targets/gpu/driver/main.cpp
```diff
@@ -21,6 +21,9 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+
+#define MIGRAPHX_STATIC_DEFINE
```
pfultz2, line 25: What is this?

src/include/migraphx/env.hpp
```diff
@@ -33,18 +33,20 @@ inline namespace MIGRAPHX_INLINE_NS {
...
 // Declare a cached environment variable
 #define MIGRAPHX_DECLARE_ENV_VAR(x)               \
-    struct x                                      \
+    struct env_variable_##x                       \
     {                                             \
         static const char* value() { return #x; } \
-    }; // NOLINT
+    };                                            \
+                                                  \
+    inline constexpr auto x = env_variable_##x{};
```
pfultz2, line 41: Why is this needed for windows?

cmake/Embed.cmake
```diff
@@ -45,80 +51,87 @@ function(generate_embed_source EMBED_NAME)
...
-        ")
+        { \"migraphx/kernels/${BASE_NAME}\", { ${START_SYMBOL}, ${SIZE_SYMBOL} }},")
     endforeach()
 
     file(WRITE "${PARSE_HEADER}" "
-#include <unordered_map>
-#include <string>
-#include <utility>
-const std::unordered_map<std::string, std::pair<const char*,const char*>>& ${EMBED_NAME}();
+#include \"migraphx/compile_src.hpp\"
```
pfultz2, line 67: Dont include migraphx headers. Embed library shouldn't depend on any migraphx library.

cmake/Embed.cmake
```diff
@@ -45,80 +51,87 @@ function(generate_embed_source EMBED_NAME)
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace gpu {
```
pfultz2, line 71: Dont use migraphx namespace. This should just got into the global namespace.

CMakeLists.txt
```diff
@@ -21,41 +21,37 @@
...
 # THE SOFTWARE.
 #####################################################################################
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
 
-if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
-    message(FATAL_ERROR "The binary and source directroy cannot be the same")
-endif()
+set(CMAKE_DISABLE_SOURCE_CHANGES ON)
+set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
```
pfultz2, line 27: These are undocumented variables. We should just check this properly so it will work with future versions of cmake.

src/api/CMakeLists.txt
```diff
@@ -22,17 +22,27 @@
...
+add_library(migraphx_c api.cpp)
 set_target_properties(migraphx_c PROPERTIES EXPORT_NAME c)
 
-# migraphx_c is stable API interface library. SO version of this should be 
-# bumped when binary compatibility is broken. 
-rocm_set_soversion(migraphx_c 3.0)   
+if(NOT WIN32)
+    # migraphx_c is stable API interface library. SO version of this should be 
+    # bumped when binary compatibility is broken. 
+    rocm_set_soversion(migraphx_c 3.0)
```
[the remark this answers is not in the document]
apwojcik: There's no SO version on Windows. Instead, the version information is provided in a resource file embedded in the EXE or DLL file. The linker embeds the version in the PE header, too. So even though the function `rocm_set_soversion()` checks for `WIN32`, we explicitly excluded it to separate Linux and Windows code visually.

src/targets/cpu/CMakeLists.txt
apwojcik: Those are two different APIs. In the context of the target `migraphx_cpu`, the definition of `MIGRAPHX_EXPORT` is `__declspec(dllimport)`. That means we import symbols from the dynamic library `migraphx.dll` to link `migraphx_cpu.dll`. However, the required definition of `MIGRAPHX_CPU_EXPORT` is `__declspec(dllexport)`. That is because we want to export symbols of the dynamic library `migraphx_cpu.dll`. So we cannot use `migraphx/export.h` because it is set to the importing symbols in the context of `migraphx_cpu`.

src/targets/gpu/CMakeLists.txt
apwojcik: We must repeat it since it looks like that information is not propagated on Windows correctly by the CMake (or tools).

src/targets/gpu/CMakeLists.txt
apwojcik: It is not compiling on Windows when `find_package(hip REQUIRED)` is missing. We will recheck it with a more recent HIP SDK.

src/targets/gpu/allocation_model.cpp
apwojcik: Tool for static analysis suggested turning it to `static` or `constexpr` for better optimization.

src/targets/gpu/analyze_streams.cpp
apwojcik: It matches the calling interface. Tool for static analysis suggested turning it to `static` or `constexpr` for better optimization.

src/targets/gpu/driver/main.cpp
apwojcik: The driver uses symbols statically defined (linked) and not being imported or exported (dynamic linking). That is how you tell the export.h header file you want static defines. I will remove it from here and move it to the CMake instead of the target property. That would be much cleaner.

src/include/migraphx/env.hpp
apwojcik: However, I will turn it back if you prefer the old way.

apwojcik: Actually, considering the same optimization options, both versions resolve to the same machine code. The compiler is doing the right job anyway: https://godbolt.org/z/Yr7P1P3KP

Let me know if you want to revert the change or keep it. I will remove it from this PR since it is not required.

src/targets/gpu/CMakeLists.txt
pfultz2: The flags for c++17 should be added with `add_compile_options` at the top-level cmake, like we do here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/CMakeLists.txt#L93

Also, we dont want to use `target_compile_features` since the version isnt stable.

src/targets/gpu/analyze_streams.cpp
pfultz2: BTW, its a member function in the interface, see here: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/stream_model.hpp#L72.

src/targets/gpu/compile_hip.cpp
pfultz2: I see, we probably want this in a seperate PR.

src/targets/cpu/CMakeLists.txt
pfultz2: Yea I think it would be better to use `WINDOWS_EXPORT_ALL_SYMBOLS` property on all libraries except the API libraries instead. For out-of-tree backends, we need all symbols available to them not the just the ones we use. All symbols are versioned so there shouldn't be an issue with different versions of migraphx used together.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -1487,13 +1488,16 @@ quantize_int8(const program& prog, const target& ptarget, const quantize_int8_op
...
-    virtual shape compute_shape(shapes inputs) const                            = 0;
-    virtual std::vector<size_t> output_alias(shapes) const { return {}; }
+    experimental_custom_op_base()                                                             = default;
+    experimental_custom_op_base(experimental_custom_op_base const&)                           = default;
+    virtual ~experimental_custom_op_base()                                                    = default;
+
+    [[nodiscard]] virtual std::string name() const                                            = 0;
+    [[nodiscard]] virtual argument compute(context ctx, shape output, arguments inputs) const = 0;
+    [[nodiscard]] virtual shape compute_shape(shapes inputs) const                            = 0;
+    [[nodiscard]] virtual std::vector<size_t> output_alias(shapes) const { return {}; }
```
pfultz2, line 1498: `[[nodiscard]]` requires C++17. This header should work with at least C++11, so it should be removed.

src/driver/main.cpp
```diff
@@ -52,8 +53,7 @@
 
 #include <fstream>
 
-namespace migraphx {
-namespace driver {
+namespace migraphx::driver {
```
pfultz2, line 56: This should be moved to a seperate PR.

src/include/migraphx/context.hpp
```diff
@@ -373,7 +373,7 @@ struct context
     private_detail_te_handle_base_type& private_detail_te_get_handle()
     {
         assert(private_detail_te_handle_mem_var != nullptr);
-        if(not private_detail_te_handle_mem_var.unique())
+        if(private_detail_te_handle_mem_var.use_count() != 1)
```
pfultz2, line 376: These changes will be overwritten when calling `make generate`.

src/include/migraphx/half.hpp
```diff
@@ -25,7 +25,7 @@
 #ifndef MIGRAPHX_GUARD_RTGLIB_HALF_HPP
 #define MIGRAPHX_GUARD_RTGLIB_HALF_HPP
 
-#include <half/half.hpp>
+#include <half.hpp>
```
pfultz2, line 28: This should be `half/half.hpp`. Thats a requirement for rocm packages.

src/include/migraphx/match/layernorm.hpp
```diff
@@ -50,35 +50,35 @@ struct layernorm_matcher
...
-    auto layernorm_onnx() const
+    [[nodiscard]] auto layernorm_onnx() const
     {
         auto div_sqrt  = f("div")(arg(0)(x_minus_mean()), arg(1)(sqrt_add_eps("sqrt")));
         auto mul_rsqrt = f("mul")(either_arg(0, 1)(x_minus_mean(), sqrt_add_eps("rsqrt")));
         return any(any_of(div_sqrt, mul_rsqrt));
     }
 
-    auto matcher() const { return layernorm_onnx(); }
+    [[nodiscard]] auto matcher() const { return layernorm_onnx(); }
```
pfultz2, line 81: These `[nodiscard]]` changes should go into a seperate PR.

src/driver/CMakeLists.txt
```diff
@@ -40,11 +40,11 @@ add_custom_command(
...
-target_link_libraries(driver migraphx_all_targets migraphx_onnx migraphx_tf)
+if(CLANG_TIDY_EXE)
+    rocm_clang_tidy_check(driver)
+endif()
 
-rocm_install_targets(
-  TARGETS driver
-)
+target_link_libraries(driver PRIVATE migraphx migraphx_onnx migraphx_tf)
+rocm_install_targets(TARGETS driver)
```
pfultz2, line 50: This is missing a newline at the end. Also formatting changes to cmake should go into a seperate PR.

src/driver/main.cpp
```diff
@@ -38,6 +38,7 @@
 #include <migraphx/json.hpp>
 #include <migraphx/version.h>
 
+#include <migraphx/instruction.hpp>
```
pfultz2, line 41: Why do we need an extra include for instruction here?

src/include/migraphx/compile_src.hpp
```diff
@@ -37,21 +37,25 @@ inline namespace MIGRAPHX_INLINE_NS {
 struct src_file
 {
     fs::path path;
-    std::pair<const char*, const char*> content;
-    std::size_t len() const { return content.second - content.first; }
+    std::string_view content;
+
+    src_file() = delete;
+    src_file(fs::path path_, std::string_view content_)
```
pfultz2, line 43: These changes should go into a seperate PR.

src/targets/cpu/CMakeLists.txt
apwojcik: `WINDOWS_EXPORT_ALL_SYMBOLS` has a downside - cannot cause import global data symbols automatically. For those, we still need to use `__declspec(dllimport)`. Anyway, we may end up using the `generate_export_header()` function.

src/targets/gpu/CMakeLists.txt
apwojcik: I see. The line you are referring to must have been removed while merging the branch `develop` to the branch `uif2/develop`. I cannot see it on the `uif2/develop` branch.

src/targets/gpu/CMakeLists.txt
apwojcik: Noted

src/CMakeLists.txt
```diff
@@ -226,69 +240,91 @@ register_migraphx_ops(
...
-pkg_check_modules(SQLITE3 REQUIRED IMPORTED_TARGET sqlite3)
-target_link_libraries(migraphx PRIVATE PkgConfig::SQLITE3)
+generate_export_header(migraphx
+    EXPORT_FILE_NAME
+        ${CMAKE_BINARY_DIR}/include/migraphx/export.h)
+
+target_include_directories(migraphx PRIVATE ${CMAKE_BINARY_DIR}/include)
+
+if(WIN32)
+    target_link_libraries(migraphx PRIVATE sqlite3::sqlite3 msgpackc-cxx)
```
pfultz2, line 279: This `if` branch should be removed, since they should have the same target name.

src/dynamic_loader_win32.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <Windows.h>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+struct dynamic_loader_impl
+{
+    dynamic_loader_impl() = default;
+    dynamic_loader_impl(const fs::path& p, std::shared_ptr<tmp_dir> t = nullptr)
+        : handle(LoadLibrary(p.string().c_str())), temp(std::move(t))
```
pfultz2, line 40: `manage_deleter<decltype(&FreeLibrary), &FreeLibrary>{}` should be passed to the shared_ptr's deleter.

src/dynamic_loader_win32.cpp
```diff
@@ -0,0 +1,92 @@
...
+            MIGRAPHX_THROW("Error loading DLL: " + p.string() +
+                           " (" + std::to_string(GetLastError()) + ")");
+        }
+    }
+
+    ~dynamic_loader_impl()
+    {
+        if (handle != nullptr)
+        {
+            FreeLibrary(handle);
```
pfultz2, line 52: Do not call `FreeLibrary` in the destructor. This could break when `dynamic_loader_impl` or  class `handle` is copied. This should be passed to the `shared_ptr`s deleter.

src/dynamic_loader_win32.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+    static std::shared_ptr<dynamic_loader_impl> from_buffer(const char* image, std::size_t size)
+    {
+        auto t = std::make_shared<tmp_dir>("dloader");
+        auto f = t->path / "tmp.dll";
+        write_buffer(f.string(), image, size);
+        return std::make_shared<dynamic_loader_impl>(f, t);
+    }
+
+    HMODULE handle = nullptr;
```
pfultz2, line 63: This should be `shared_ptr<HMODULE>`.

src/targets/cpu/include/migraphx/cpu/dnnl.hpp
```diff
@@ -300,91 +305,109 @@ struct dnnl_op : auto_register_op<Derived>
...
+
 #ifndef NDEBUG
-        auto prim_attr = get_primitive_attr(md);
+        execute = std::bind(&dnnl_op::execute_impl_, this, output_shape, inputs, md, prim,
+                            arg_lookup, std::placeholders::_1, std::placeholders::_2);
+#else
+        execute = std::bind(&dnnl_op::execute_impl_, this, md, prim, arg_lookup,
+                            std::placeholders::_1, std::placeholders::_2);
 #endif
-        execute = [=](context&, const std::vector<argument>& args) {
```
pfultz2, line 311: Assigning a lambda to `std::function` is not UB. This should use the lambda.

src/targets/cpu/lowering.cpp
```diff
@@ -117,11 +108,11 @@ struct cpu_im2col
             const std::size_t col_height = (height - kernel_h + 2 * pad_h) / stride_h + 1;
             const std::size_t col_width  = (width - kernel_w + 2 * pad_w) / stride_w + 1;
             // account for padding for the starting position of the input pixels
-            long iinput = kdiv2_h - long(pad_h);
+            std::size_t iinput = kdiv2_h - long(pad_h);
```
pfultz2, line 111: This should be a signed variable.

src/targets/gpu/CMakeLists.txt
```diff
@@ -125,6 +219,41 @@ add_library(migraphx_gpu
...
+            -fms-compatibility)
+
+    target_sources(migraphx_gpu
+        PRIVATE
+            kernels/migraphx_kernels.cpp
+            kernels/migraphx_kernels.hpp
+            kernels/resource.h
+            kernels/resource.rc)
+
+    target_include_directories(migraphx_gpu PRIVATE kernels)
```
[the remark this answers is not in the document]
pfultz2: You are adding the resource files using `target_sources`. So instead of `target_sources(migraphx_gpu PRIVATE ...)` you can write `target_sources(${EMBED_NAME} INTERFACE ...)`. And when we do `target_link_libraries(migraphx_gpu PRIVATE kernels)` cmake will add the sources to `migraphx_gpu` as if we wrote `target_sources(migraphx_gpu PRIVATE ...)`.

src/CMakeLists.txt
```diff
@@ -226,69 +240,91 @@ register_migraphx_ops(
...
 )
 
-
-check_cxx_linker_flag(-lstdc++fs HAS_LIB_STD_FILESYSTEM)
-if(HAS_LIB_STD_FILESYSTEM)
-target_link_libraries(migraphx PRIVATE -lstdc++fs)
+if(NOT WIN32)
+    check_cxx_linker_flag(-lstdc++fs HAS_LIB_STD_FILESYSTEM)
+    if(HAS_LIB_STD_FILESYSTEM)
+        target_link_libraries(migraphx PRIVATE -lstdc++fs)
```
pfultz2, line 258: We dont need to wrap this in `WIN32` since we already check if the flag is valid.

src/CMakeLists.txt
```diff
@@ -226,69 +240,91 @@ register_migraphx_ops(
...
+    endif()
+    target_link_libraries(migraphx PRIVATE -ldl)
```
pfultz2, line 260: We can use `${CMAKE_DL_LIBS}` instead and avoid another `if(WIN32)`.

src/CMakeLists.txt
pfultz2: So doing something like this:

```cmake
function(export_dll LIBRARY DEFINE)
    if(WIN32)
        target_compile_definitions(${LIBRARY} PRIVATE "-D${DEFINE}=__declspec(dllexport)")
        target_compile_definitions(${LIBRARY} INTERFACE "-D${DEFINE}=__declspec(dllimport)")
    else()
        target_compile_definitions(${LIBRARY} PUBLIC "-D${DEFINE}=")
    endif()
endfunction()
```

This should reduce down the `if(WIN32)` that we would have to do. And there is no need for includes or adding defines as its already handled by the usage requirements.

src/include/migraphx/run_loop.hpp
```diff
@@ -24,14 +24,16 @@
 #ifndef MIGRAPHX_GUARD_RTGLIB_RUN_LOOP_HPP
 #define MIGRAPHX_GUARD_RTGLIB_RUN_LOOP_HPP
 
+#include <array>
```
umangyadav, line 27: Is the `<array>` being used ?

src/shape.cpp
```diff
@@ -547,7 +547,7 @@ std::vector<std::size_t> shape::max_lens() const
     return this->dynamic() ? impl->max_lens() : this->lens();
 }
 
-std::vector<std::set<std::size_t>> shape::opt_lens() const { return impl->opt_lens(); }
+const std::vector<std::set<std::size_t>> shape::opt_lens() const { return impl->opt_lens(); }
```
umangyadav, line 550: This shouldn't return const.

src/targets/cpu/include/migraphx/cpu/lowering.hpp
```diff
@@ -24,20 +24,17 @@
 #ifndef MIGRAPHX_GUARD_RTGLIB_CPU_LOWERING_HPP
 #define MIGRAPHX_GUARD_RTGLIB_CPU_LOWERING_HPP
 
-#include <string>
 #include <migraphx/config.hpp>
+#include <migraphx/program.hpp>
```
umangyadav, line 28: Is it being used ?

src/targets/cpu/include/migraphx/cpu/pointwise.hpp
```diff
@@ -24,6 +24,8 @@
 #ifndef MIGRAPHX_GUARD_AMDMIGRAPHX_CPU_POINTWISE_HPP
 #define MIGRAPHX_GUARD_AMDMIGRAPHX_CPU_POINTWISE_HPP
 
+#include <array>
```
umangyadav, line 27: Is this being used ?

src/targets/gpu/driver/CMakeLists.txt
```diff
@@ -22,10 +22,22 @@
...
-rocm_clang_tidy_check(gpu-driver)
-target_include_directories(gpu-driver PRIVATE include)
-target_link_libraries(gpu-driver PRIVATE migraphx_gpu)
+add_executable(migraphx_gpu_driver
+        action.cpp
+        include/migraphx/gpu/driver/action.hpp
+        compile_op.cpp
+        main.cpp
+        parser.cpp
+        include/migraphx/gpu/driver/parser.hpp
```
umangyadav, line 31: instead of header files here can it make use of `target_include_directories` ?

test/promote_literals_test.cpp
```diff
@@ -26,6 +26,7 @@
 #include <migraphx/program.hpp>
 #include <migraphx/dead_code_elimination.hpp>
 #include <migraphx/eliminate_common_subexpression.hpp>
+#include <migraphx/instruction.hpp>
```
umangyadav, line 29: Is this being used ?

src/targets/gpu/kernels/resource.h
```diff
@@ -0,0 +1,64 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#ifndef MIGRAPHX_GUARD_MIGRAPHX_KERNELS_RESOURCE_H
+#define MIGRAPHX_GUARD_MIGRAPHX_KERNELS_RESOURCE_H
+
+#define TEXTFILE                    256
+
+#define IDR_ALGORITHM_HPP           101
```
umangyadav, line 30: What these numbers represent i.e. 101, 102, ... ?

test/api/test_c_op_construct.c
```diff
@@ -30,7 +30,7 @@ void expect_equal(const char* x, const char* y)
         abort();
 }
 
-int main()
+int main(void)
```
umangyadav, line 33: Void necessary ?

src/targets/gpu/sync_device.cpp
```diff
@@ -31,7 +31,7 @@ namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
-void sync_device::apply(module& m) const
+void sync_device::apply(module& m)
```
umangyadav, line 34: This should remain `const`.

src/targets/gpu/target.cpp
```diff
@@ -87,7 +87,7 @@ pass enable_pass(bool enabled, pass p)
     return id_pass{};
 }
 
-std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_options& options) const
+std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_options& options)
```
umangyadav, line 90: Methods should remain const

src/CMakeLists.txt
pfultz2: If we are still using `generate_export_header` then we should still use `export_dll` function(or we can use a different name):

```cmake
function(export_dll LIBRARY DIRECTORY)
    set(FILENAME ${CMAKE_CURRENT_BINARY_DIR}/export/include/${DIRECTORY}/export.h)
    generate_export_header(${LIBRARY} EXPORT_FILE_NAME ${FILENAME})
    target_include_directories(${LIBRARY} PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/export/include>)
    rocm_install(FILES ${FILENAME} DESTINATION include/${DIRECTORY}/)
endfunction()
```

This will setup the header with correct include path and installation.

src/CMakeLists.txt
```diff
@@ -226,69 +240,91 @@ register_migraphx_ops(
...
 target_link_libraries(migraphx PRIVATE nlohmann_json::nlohmann_json)
 
-find_package(PkgConfig)
-pkg_check_modules(SQLITE3 REQUIRED IMPORTED_TARGET sqlite3)
-target_link_libraries(migraphx PRIVATE PkgConfig::SQLITE3)
+generate_export_header(migraphx
+    EXPORT_FILE_NAME
+        ${CMAKE_BINARY_DIR}/include/migraphx/export.h)
+
+target_include_directories(migraphx PRIVATE ${CMAKE_BINARY_DIR}/include)
```
pfultz2, line 276: So this needs to be `PUBLIC` since this is included in the headers that downstream consumers will use(such as the test). This is a reason why it would be best to have a `export_dll` function to encapsulate all these fine details.

src/targets/gpu/kernels/migraphx_kernels.cpp
```diff
@@ -0,0 +1,87 @@
...
+        { "migraphx/kernels/roialign.hpp", resource::read(IDR_ROIALIGN_HPP) },
+        { "migraphx/kernels/scatternd.hpp", resource::read(IDR_SCATTERND_HPP) },
+        { "migraphx/kernels/shape.hpp", resource::read(IDR_SHAPE_HPP) },
+        { "migraphx/kernels/softmax.hpp", resource::read(IDR_SOFTMAX_HPP) },
+        { "migraphx/kernels/tensor_view.hpp", resource::read(IDR_TENSOR_VIEW_HPP) },
+        { "migraphx/kernels/type_traits.hpp", resource::read(IDR_TYPE_TRAITS_HPP) },
+        { "migraphx/kernels/types.hpp", resource::read(IDR_TYPES_HPP) },
+        { "migraphx/kernels/vec.hpp", resource::read(IDR_VEC_HPP) },
+        { "migraphx/kernels/vectorize.hpp", resource::read(IDR_VECTORIZE_HPP) }
+    };
```
pfultz2, line 81: These files should not be added to the repo. They should just be generated by cmake in the build directory.

src/targets/gpu/CMakeLists.txt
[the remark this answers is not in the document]
pfultz2: For now I added a simple fallback for the embed module in #1804. That should help get things to compile for windows and then we can look into a better design for this at a later time.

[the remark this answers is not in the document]
apwojcik: Again, that is not how resource embedding works on Windows. The mechanism is __native__ to Windows, and toolchains have a dedicated resource compiler. The compiler produces a `*.res` binary file from the `*.rc` text file. Therefore, making it a separate static library or linking it explicitly to an executable or dynamic library is unnecessary. The linker automatically merges the `*.res` file to output binary. The only thing that needs to be done is adding a `*.rc` text file to the target's source files. Additionally, the WIN32 has a dedicated API to operate on resource file while embedded in a binary.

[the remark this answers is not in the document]
apwojcik: Also, please do not push #1804 because that change will not help and will complicate porting even more (it is not portable).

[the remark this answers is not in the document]
apwojcik: What you do in Embed.cmake is native to Linux (and only to Linux/UNIX). None other linkers provide such a functionality. Even llvm-lld (GNU ld emulation) on Windows cannot do that.

src/include/migraphx/context.hpp
apwojcik: The problem is `unique()` is deprecated on Windows and fails to compile due to the C++17 standard used. So what can I do to make it persistent?

src/include/migraphx/half.hpp
apwojcik: Noted

src/include/migraphx/matcher.hpp
```diff
@@ -124,14 +126,14 @@ template <class M>
 auto bind_match(M m, std::string name)
 {
     return make_function_matcher(
-        [=, name = std::move(name)](matcher_context& ctx,
-                                    instruction_ref ins) -> optional<instruction_ref> {
+        [=, name_ = std::move(name)](matcher_context& ctx,
```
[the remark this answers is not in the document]
apwojcik: The compilation on Windows is failing because of the capture `name` is shadowing the function argument `name`. I need to provide different name for the captured variable.

src/driver/main.cpp
apwojcik: The compilation on Windows is failing because of incomplete type `instruction`.

src/include/migraphx/compile_src.hpp
apwojcik: We need those changes to read from embedded resource file.

src/include/migraphx/compile_src.hpp
```diff
@@ -37,21 +37,25 @@ inline namespace MIGRAPHX_INLINE_NS {
...
 
 struct src_compiler
 {
     std::string compiler                      = "c++";
-    std::string flags                         = "";
-    std::string output                        = "";
-    std::string launcher                      = "";
+    std::string flags;
+    std::string output;
+    std::string launcher;
```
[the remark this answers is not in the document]
apwojcik: I will add it back. However, according to the specification, it is unnecessary - the default construction is the empty string (explicit and implicit).

src/targets/cpu/include/migraphx/cpu/dnnl.hpp
apwojcik: That is not compiling on Windows. The fatal error message says something like that "cannot assign temporary object (lambda) to class data member".

src/CMakeLists.txt
apwojcik: That whole part is Linux-specific and does not concern Windows.

CMakeLists.txt
```diff
@@ -100,165 +100,169 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 include(EnableCompilerWarnings)
-include(ROCMClangTidy)
-if(CMAKE_CXX_COMPILER MATCHES ".*clang\\+\\+")
+#include(ROCMClangTidy)
```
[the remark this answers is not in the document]
apwojcik: Yes, that is a mistake. I forgot to uncomment it. I will fix it on the `uif2-initial` branch. Thanks for letting me know.

CMakeLists.txt
```diff
@@ -89,8 +88,9 @@ set(MIGRAPHX_ENABLE_CPU Off CACHE BOOL "")
 # Disable fpga backend by default
 set(MIGRAPHX_ENABLE_FPGA Off CACHE BOOL "")
 
-set(CMAKE_CXX_STANDARD_DEFAULT "")
-add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-std=c++17>)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 if(${CMAKE_VERSION} VERSION_LESS "3.12.0")
```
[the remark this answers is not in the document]
apwojcik: Good catch. I missed that. Thanks!

src/compile_src.cpp
```diff
@@ -21,12 +21,14 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+
+#include <cassert>
```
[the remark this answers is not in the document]
apwojcik: It is required on Windows. There are differences in header files that are included implicitly on Windows and Linux. On Linux, some header files are included by other header files, and the same header files on Windows might not have included the ones Linux does included.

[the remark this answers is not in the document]
apwojcik: Here compilation fails because of unknown symbol `assert`.

src/include/migraphx/eliminate_data_type.hpp
```diff
@@ -42,8 +42,8 @@ struct eliminate_data_type
 {
     std::set<shape::type_t> types;
     shape::type_t target_type;
-    std::string name() const { return "eliminate_data_type"; }
-    void apply(module& m) const;
+    static std::string name() { return "eliminate_data_type"; }
```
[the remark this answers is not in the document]
apwojcik: Yes, I already have revert it back on `uif2-initial` branch.

src/include/migraphx/op/add.hpp
```diff
@@ -35,12 +35,12 @@ struct add : binary<add>
 {
     value attributes() const
     {
-        auto a           = base_attributes();
-        a["commutative"] = true;
+        value a          = base_attributes();
+        a["commutative"] = value(true);
```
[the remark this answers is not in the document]
apwojcik: Correct, it does not work. I have no clue why, to be honest. I will look into it later.

src/include/migraphx/op/nonmaxsuppression.hpp
```diff
@@ -24,6 +24,7 @@
 #ifndef MIGRAPHX_GUARD_OPERATORS_NONMAXSUPPRESSION_HPP
 #define MIGRAPHX_GUARD_OPERATORS_NONMAXSUPPRESSION_HPP
 
+#include <array>
```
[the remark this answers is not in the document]
apwojcik: Yes, it is used implicitly in the header file (at least on Windows).

src/include/migraphx/op/roialign.hpp
```diff
@@ -33,6 +33,7 @@
 #include <migraphx/dfor.hpp>
 #include <migraphx/ranges.hpp>
 #include <migraphx/shape_for_each.hpp>
+#include <array>
```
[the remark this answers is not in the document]
apwojcik: Yes, the same as above.

src/include/migraphx/run_loop.hpp
apwojcik: Yes, the same as above.

src/py/CMakeLists.txt
```diff
@@ -22,16 +22,14 @@
 # THE SOFTWARE.
 #####################################################################################
 
-option(MIGRAPHX_ENABLE_PYTHON "Enable python bindings" ON)
-if(MIGRAPHX_ENABLE_PYTHON)
```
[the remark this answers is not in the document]
apwojcik: We moved the option to the upper `CMakeList.txt`. Since we are not building Python we need that option, too.

src/targets/cpu/include/migraphx/cpu/pointwise.hpp
apwojcik: Yes, implicitly in the header file (at least on Windows). Without it the compilation fails - fatal error.

test/api/test_c_op_construct.c
apwojcik: Yes, that is C file (not C++). C standard does not have prototype `int main()`. The compilation is failing here.

src/targets/gpu/kernels/resource.h
apwojcik: Windows resources identifiers. Ordinal numbers. That is the standard resource header file, and must be done in such a way.

test/promote_literals_test.cpp
apwojcik: Yes, compiler on Windows complains on incomplete type `instruction`.

src/targets/gpu/driver/CMakeLists.txt
apwojcik: As of the specification, the CMake expects all files comprising an executable or library to be added to the source files list (even the ones not being compiled). In IDEs natively working with CMake, we use that information to view a complete project tree, symbol search, and more. For instance, JetBrains CLion, when a file is not added to the target's source files list, only `#included`, the tool is not considering that file in symbol searches (however, that can be overridden). Adding the `*.hpp` file to the list significantly speeds up project file operations (in contrast to parsing 1600 source files to build the dependency tree every couple of seconds when the content changes).

apwojcik: However, the Windows will compile without adding *.hpp to the list.

src/targets/gpu/kernels/migraphx_kernels.cpp
apwojcik: It is a temporary solution. We have already added a Jira ticket to work on that. We need something to start working on enabling tests, etc.

src/CMakeLists.txt
apwojcik: Fixed on `uif2-initial` branch - not yet pushed to repo.

src/CMakeLists.txt
apwojcik: Fixed on `uif2-initial` branch - not yet pushed to repo.

src/CMakeLists.txt
apwojcik: I explained below @umangyadav's comment the reason why we changed that target. However, we misunderstood the meaning of the target. We thought it was aggregating target, but making it such had not disturbed the linking, because the target doubled the linking information.

src/api/CMakeLists.txt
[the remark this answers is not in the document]
apwojcik: Fixed on `uif2-initial` branch - not yet pushed to repo.

src/api/api.cpp
```diff
@@ -38,26 +43,29 @@
...
 migraphx_status try_(F f, bool output = true) // NOLINT
 {
+#ifdef BUILD_DEV
     if(disable_exception_catch)
     {
         f();
     }
     else
     {
+#endif
```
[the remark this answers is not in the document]
apwojcik: Fixed on `uif2-initial` branch - not yet pushed to repo.

src/api/api.cpp
apwojcik: How can I make them permanent?

src/targets/gpu/CMakeLists.txt
apwojcik: The `CMAKE_PREFIX_PATH` is set in upper CMake already. It is redundant information.

src/CMakeLists.txt
```diff
@@ -226,69 +240,91 @@ register_migraphx_ops(
 )
 register_op(migraphx HEADER migraphx/op/rnn_variable_seq_lens.hpp OPERATORS op::rnn_var_sl_shift_output op::rnn_var_sl_shift_sequence)
 register_op(migraphx HEADER migraphx/builtin.hpp OPERATORS builtin::literal builtin::param builtin::returns)
-rocm_clang_tidy_check(migraphx)
+
+if(CLANG_TIDY_EXE)
```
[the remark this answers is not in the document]
apwojcik: I added a Jira ticket. However, we need it here until `rocm-cmake` is updated. Otherwise, we are failing the configuration of a build.

src/targets/gpu/kernels/migraphx_kernels.cpp
apwojcik: BTW why this file needs to be generated? On Windows, resource file format supports including external files (binary and text). We do not need to generate anything.

pfultz2: > It is a temporary solution. We have already added a Jira ticket to work on that. We need something to start working on enabling tests, etc.

Yea hopefully #1804 can help get it working for windows and then we can work on trying to get it to work with the resource files later on.

> why this file needs to be generated? 

We shouldn't need to update source files when we add new file to embed, that should be handled automatically with the `add_embed_library` function.

test/api/test_c_op_construct.c
pfultz2: Yea this is fixing a clang warning. I am not sure why we dont catch it in our CI though.

test/verify/CMakeLists.txt
```diff
@@ -27,9 +27,12 @@ file(GLOB VERIFY_TESTS ${CONFIGURE_DEPENDS} *.cpp)
 add_executable(test_verify ${VERIFY_TESTS})
 add_dependencies(tests test_verify)
 add_dependencies(check test_verify)
-target_link_libraries(test_verify migraphx migraphx_all_targets)
-target_include_directories(test_verify PUBLIC ../include)
-rocm_clang_tidy_check(test_verify)
+target_link_libraries(test_verify PRIVATE migraphx)
```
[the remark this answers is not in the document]
pfultz2: Setting the policy to NEW should fix this warning.

src/targets/cpu/include/migraphx/cpu/dnnl.hpp
pfultz2: Is this an error on clang?

src/include/migraphx/matcher.hpp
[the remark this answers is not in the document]
pfultz2: Why does shadowing cause the compilation to fail? I assume it would just be a warning.

src/include/migraphx/context.hpp
pfultz2: Sure, it looks like you put the change into te.py which is the correct place for it. If not thats where the change would need to be.

src/targets/gpu/driver/main.cpp
pfultz2: Actually, these define should not be necessary since we arent compiling static and shared in the same build tree. The export header function from takes care of setting the correct macro automatically.

src/targets/gpu/CMakeLists.txt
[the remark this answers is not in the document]
pfultz2: > We will not use GNU for building on Windows because it introduces a non-native POSIX layer, and translations to and from WIN32 might impact the performance.

Yes of course, I wouldn't expect you to use GNU linker on windows.

In #1804 it will only use the GNU LD on linux. For other OSs, it will generate C arrays in source files so it will work portably for all OSs.  Later, we can work on building a windows-specific implementation for this in the future if we think that will be better than the generic portable version.

src/api/api.cpp
pfultz2: You need to modify tool/api.py that is the script generating this.

src/targets/cpu/include/migraphx/cpu/dnnl.hpp
pfultz2: > The fatal error message says something like that "cannot assign temporary object (lambda) to class data member".

Since its complaining about a temporary, maybe you can assign it to a local variable first before assigning it to the member variable.

CMakeLists.txt
```diff
@@ -68,12 +64,15 @@ set(CMAKE_EXTRA_INCLUDE_FILES)
 include(ROCMSetupVersion)
 
 option(BUILD_DEV "Build for development purpose only" OFF)
+option(BUILD_SHARED_LIBS "Build as a shared library" ON)
+
+if(BUILD_TESTING AND NOT BUILD_DEV)
+    set(BUILD_DEV ON CACHE BOOL "" FORCE)
```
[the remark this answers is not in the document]
pfultz2: I dont think we should force `BUILD_DEV` to `ON` when tests are enabled. We want to be able to run our tests on the actual released version. I think these two options should be orthogonal.

test/CMakeLists.txt
```diff
@@ -24,36 +24,37 @@
 
 cmake_policy(SET CMP0057 NEW)
 
-include(CTest)
+if(MIGRAPHX_ENABLE_GPU)
+    find_package(hip REQUIRED)
+    find_package(miopen REQUIRED)
+endif()
```
pfultz2, line 30: This should not be repeated here. These should go under the gpu directory.

test/CMakeLists.txt
```diff
@@ -100,74 +101,91 @@ endfunction()
...
-    rocm_clang_tidy_check(test_${BASE_NAME})
+    if(CLANG_TIDY_EXE)
+        rocm_clang_tidy_check(test_${BASE_NAME})
+    endif()
 endforeach()
 
 if(MIGRAPHX_ENABLE_GPU)
-    # gpu tests
-    file(GLOB GPU_TESTS ${CONFIGURE_DEPENDS} gpu/*.cpp)
+    set(GPU_TESTS
```
pfultz2, line 124: The tests should still use globbing.

test/CMakeLists.txt
```diff
@@ -100,74 +101,91 @@ endfunction()
...
-        rocm_clang_tidy_check(test_gpu_${BASE_NAME})
-        set_tests_properties(test_gpu_${BASE_NAME} PROPERTIES
-            COST 10
-            RESOURCE_LOCK gpu
-        )
+        if(CLANG_TIDY_EXE)
+            rocm_clang_tidy_check(test_gpu_${BASE_NAME})
+        endif()
+        set_tests_properties(test_gpu_${BASE_NAME} PROPERTIES COST 10 RESOURCE_LOCK gpu)
+        target_link_libraries(test_gpu_${BASE_NAME} PRIVATE hip::device MIOpen migraphx_gpu)
```
pfultz2, line 144: The gpu tests should only link `migraphx_gpu`.  If `MIOpen` is needed to link then the `migraphx_gpu` target was set up wrong. 

Also, `hip::device` is only for gpu-device code which are not in the tests(it should only be linked with `migraphx_device` target as that it the only library which has GPU code).

test/CMakeLists.txt
```diff
@@ -100,74 +101,91 @@ endfunction()
...
+        if(WIN32)
+            target_include_directories(test_gpu_${BASE_NAME} PRIVATE ${CMAKE_SOURCE_DIR}/src/targets/gpu/kernels)
+        else()
+            target_link_libraries(test_gpu_${BASE_NAME} PRIVATE migraphx_kernels)
```
pfultz2, line 148: These should not be linked here. If they are needed then you need to make sure `migraphx_gpu` is setup correctly.

test/CMakeLists.txt
```diff
@@ -190,6 +210,10 @@ if(MIGRAPHX_ENABLE_PYTHON)
     add_subdirectory(py)
 endif()
 
+if(MIGRAPHX_ENABLE_CPU)
+    find_package(OpenMP REQUIRED COMPONENTS CXX)
```
pfultz2, line 214: This should not be here. The `find_package` call should go into targets/cpu.

test/CMakeLists.txt
```diff
@@ -201,6 +225,20 @@ function(test_header NAME HEADER)
...
+    endif()
+    if(MIGRAPHX_ENABLE_CPU)
+        target_compile_options(${NAME} PRIVATE -Wno-source-uses-openmp -Wno-ignored-attributes)
+        target_include_directories(${NAME} PRIVATE ${CMAKE_SOURCE_DIR}/src/targets/cpu/include)
+        target_link_libraries(${NAME} PRIVATE OpenMP::OpenMP_CXX)
+        if(MIGRAPHX_ENABLE_ZENDNN)
+        else()
+            target_link_libraries(${NAME} PRIVATE dnnl)
+        endif()
+    endif()
```
pfultz2, line 241: None of this should be added here.

test/CMakeLists.txt
```diff
@@ -211,7 +249,6 @@ function(test_headers PREFIX)
         string(MAKE_C_IDENTIFIER ${HEADER_REL} TEST_NAME)
         get_filename_component(BASE_NAME ${HEADER} NAME_WE)
         test_header(header_${TEST_NAME} ${PREFIX}/${BASE_NAME}.hpp)
-        target_link_libraries(header_${TEST_NAME} migraphx_all_targets)
```
pfultz2, line 214: This should not be removed.

test/verify/CMakeLists.txt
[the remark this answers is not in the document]
pfultz2: Removing `migraphx_all_targets` may pass CI, but it wont run tests on any of the backends. 

@umangyadav Make sure to check that the verify tests are running for the backends. I think running with `MIGRAPHX_TRACE_TEST=1` should show what targets is used.

src/dynamic_loader_win32.cpp
apwojcik: https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-freelibrary

apwojcik:
New PR #1828 was created from the branch `uif2-initial` instead.

src/CMakeLists.txt
apwojcik: Currently Windows installer use HIP_PATH and ROCM_PATH to locate HIP/ROCm installation. We will keep it.

PR #1771, opened 2023-05-23
Backout fp16 max/min HIP API change

umangyadav:
#1764  broke perf/accuracy CI tests because it used rocm-5.4.2 inside the docker.  
rocm-5.4.2 doesn't have `hmax` and `hmin` APIs. it only exists in rocm-5.5+.  Backing out those changes temporarily untill we have rocm-5.5 in perf/accuracy CI tests.


causten (approved):

manupak:
bump. (just to be in the loop to get notified when this PR is merged)

pfultz2:
I wonder if we can check the version of hip. And conditionally enable these overloads.

umangyadav:
> I wonder if we can check the version of hip. And conditionally enable these overloads.

It is possible. But for the hipRTC, it needs to use different API to retrieve version.  and I tried retreiving hiprtc version for both 5.5 and 5.4.2 but they both have same major and minor versions.

pfultz2 (approved):

PR #1782, opened 2023-05-25
Update install prereqs python fix

TedThemistokleous:
-Updated install_prereqs when using 22.04
 -Added new Dockerfile for Ubuntu 22.04 used for testing
 -Kept removed items commented in new dockerfile for visibility in this PR, can remove later.
 
 Tested and ran with Navi10 and MI250 servers and make check passes

### DIFF
--- a/Dockerfile_Ubuntu_2204
+++ b/Dockerfile_Ubuntu_2204
@@ -1,0 +1,127 @@
+FROM ubuntu:22.04
+
+ARG PREFIX=/usr/local
+
+# Support multiarch
+RUN dpkg --add-architecture i386
+
+# Install rocm key
+RUN apt-get update && apt-get install -y gnupg2 --no-install-recommends curl && \
+    curl -fsSL http://repo.radeon.com/rocm/rocm.gpg.key | gpg --dearmor -o /etc/apt/trusted.gpg.d/rocm-keyring.gpg
+
+# Add rocm repository
+RUN sh -c "echo 'deb [arch=amd64 signed-by=/etc/apt/trusted.gpg.d/rocm-keyring.gpg] http://repo.radeon.com/rocm/apt/5.5 jammy main' > /etc/apt/sources.list.d/rocm.list"
+
+# From docs.amd.com for installing rocm. Needed to install properly
+RUN sh -c "echo 'Package: *\nPin: release o=repo.radeon.com\nPin-priority: 600' > /etc/apt/preferences.d/rocm-pin-600"
+
+
+# Install dependencies
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated \
+    apt-utils \
+    build-essential \
+    #clang-format-10 \
+    cmake \
+    curl \
+    doxygen \
+    #g++-7 \
+    gdb \
+    git \
+    lcov \
+    locales \
+    pkg-config \
+    python3 \
+    python3-dev \
+    python3-pip \
+    software-properties-common \
+    wget \
+    rocm-device-libs \
+    hip-base \
+    libnuma-dev \
+    miopen-hip \
+    rocblas \
+    hipfft \
+    rocthrust \
+    rocrand \
+    hipsparse \
+    rccl \
+    rccl-dev \
+    rocm-smi-lib \
+    rocm-dev \
+    roctracer-dev \
+    hipcub  \
+    hipblas  \
+    hipify-clang \
+    half \
+    libssl-dev \
+    zlib1g-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+
+# add this for roctracer dependancies
+RUN pip3 install CppHeaderParser
+
+# Workaround broken rocm packages
+RUN ln -s /opt/rocm-* /opt/rocm
+RUN echo "/opt/rocm/lib" > /etc/ld.so.conf.d/rocm.conf
+RUN echo "/opt/rocm/llvm/lib" > /etc/ld.so.conf.d/rocm-llvm.conf
+RUN ldconfig
+
+RUN locale-gen en_US.UTF-8
+RUN update-locale LANG=en_US.UTF-8
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+# Install dependencies
+ADD dev-requirements.txt /dev-requirements.txt
+ADD requirements.txt /requirements.txt
+ADD rbuild.ini /rbuild.ini
+
+COPY ./tools/install_prereqs.sh /
+RUN /install_prereqs.sh /usr/local / && rm /install_prereqs.sh
+RUN test -f /usr/local/hash || exit 1
+
+# Install yapf
+RUN pip3 install yapf==0.28.0
+
+# Install doc requirements
+ADD doc/requirements.txt /doc-requirements.txt
+RUN pip3 install -r /doc-requirements.txt
+
+# Download real models to run onnx unit tests
+ENV ONNX_HOME=/.onnx
+COPY ./tools/download_models.sh /
+RUN /download_models.sh && rm /download_models.sh
+
+# Install latest ccache version
+RUN cget -p $PREFIX install facebook/zstd@v1.4.5 -X subdir -DCMAKE_DIR=build/cmake
+RUN cget -p $PREFIX install ccache@v4.1 -DENABLE_TESTING=OFF
+RUN cget -p /opt/cmake install kitware/cmake@v3.24.3
+
+COPY ./test/onnx/.onnxrt-commit /
+
+ARG ONNXRUNTIME_REPO=https://github.com/Microsoft/onnxruntime
+ARG ONNXRUNTIME_BRANCH=main
+ARG ONNXRUNTIME_COMMIT
+
+RUN git clone --single-branch --branch ${ONNXRUNTIME_BRANCH} --recursive ${ONNXRUNTIME_REPO} onnxruntime && \
+    cd onnxruntime && \
+    if [ -z "$ONNXRUNTIME_COMMIT" ] ; then git checkout $(cat /.onnxrt-commit) ; else git checkout ${ONNXRUNTIME_COMMIT} ; fi && \
+    /bin/sh /onnxruntime/dockerfiles/scripts/install_common_deps.sh
+
+
+ADD tools/build_and_test_onnxrt.sh /onnxruntime/build_and_test_onnxrt.sh
+
+RUN cget -p /usr/local install ROCmSoftwarePlatform/rocMLIR@a997d5f51314b45d7a4c04f1599966dcf53f9b4d -DBUILD_MIXR_TARGET=On -DLLVM_ENABLE_ZSTD=Off -DLLVM_ENABLE_THREADS=Off
+
+ENV MIOPEN_FIND_DB_PATH=/tmp/miopen/find-db
+ENV MIOPEN_USER_DB_PATH=/tmp/miopen/user-db
+ENV LD_LIBRARY_PATH=$PREFIX/lib
+
+# Setup ubsan environment to printstacktrace
+ENV UBSAN_OPTIONS=print_stacktrace=1
+ENV ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1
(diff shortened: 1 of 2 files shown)


Dockerfile_Ubuntu_2204
```diff
@@ -0,0 +1,127 @@
...
+RUN cget -p /usr/local install ROCmSoftwarePlatform/rocMLIR@a997d5f51314b45d7a4c04f1599966dcf53f9b4d -DBUILD_MIXR_TARGET=On -DLLVM_ENABLE_ZSTD=Off -DLLVM_ENABLE_THREADS=Off
+
+ENV MIOPEN_FIND_DB_PATH=/tmp/miopen/find-db
+ENV MIOPEN_USER_DB_PATH=/tmp/miopen/user-db
+ENV LD_LIBRARY_PATH=$PREFIX/lib
+
+# Setup ubsan environment to printstacktrace
+ENV UBSAN_OPTIONS=print_stacktrace=1
+ENV ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1
+RUN ln -s /opt/rocm/llvm/bin/llvm-symbolizer /usr/bin/llvm-symbolizer
```
pfultz2, line 126: How do we keep this dockerfile in sync with the other dockerfile? This looks like a duplicate of Dockerfile. Also docker files should be named Dockerfile or *.docker.

tools/install_prereqs.sh
```diff
@@ -56,8 +56,15 @@ echo "Dependencies are installed at $PREFIX"
 # Install deps with rbuild
 rbuild prepare -d $PREFIX -s develop
 
-# install onnx package for unit tests
-pip3 install onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
+if ! python3 -c 'import sys; assert sys.version_info >= (3, 10)' > /dev/null; then
+  python3.8 -m pip install protobuf==3.20.3 onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
```
pfultz2, line 60: You just checked for version 3.10 so you shouldn't be using 3.8.

tools/install_prereqs.sh
```diff
@@ -56,8 +56,15 @@ echo "Dependencies are installed at $PREFIX"
...
+  export CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+  python3.10 -m pip install onnx==1.10.2
+  python3.10 -m pip install protobuf==3.20.3 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
```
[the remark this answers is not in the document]
TedThemistokleous: Shouldnt it work since we're invoking it through python3.10?

tools/install_prereqs.sh
```diff
@@ -56,8 +56,15 @@ echo "Dependencies are installed at $PREFIX"
...
+else
+  # install onnx package for unit tests
+  pip3 install onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
 
-# pin version of protobuf in Python for onnx runtime unit tests
-pip3 install protobuf==3.20.0
+  # pin version of protobuf in Python for onnx runtime unit tests
+  pip3 install protobuf==3.20.0
```
pfultz2, line 69: Doing `CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON" pip3 install protobuf==3.20.3` doesnt work on older python?

tools/install_prereqs.sh
```diff
@@ -56,8 +56,15 @@ echo "Dependencies are installed at $PREFIX"
...
+fi
```
pfultz2, line 70: If you need to install different protobuf versions, you should do:

```
pip3 install onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
if ! python3 -c 'import sys; assert sys.version_info >= (3, 10)' > /dev/null; then
    CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON" pip3 install protobuf==3.20.3
else
    pip3 install protobuf==3.20.0
fi
```

But I wonder if we can just use the same version of protobuf.

tools/install_prereqs.sh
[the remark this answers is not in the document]
pfultz2: You just checked the version was 3.10, so it should be using 3.10 not 3.8.

tools/install_prereqs.sh
TedThemistokleous: Probably can its just a small version diff 3.20.* , i'll retry to default it all to 3.20.0

Dockerfile_Ubuntu_2204
```diff
@@ -0,0 +1,127 @@
+FROM ubuntu:22.04
```
umangyadav, line 1: I think it is better to move all dockerfiles inside `Dockerfiles/` folder.

tools/install_prereqs.sh
TedThemistokleous: Not sure, I need to recheck that

tools/install_prereqs.sh
TedThemistokleous: Removed

tools/install_prereqs.sh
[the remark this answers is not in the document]
TedThemistokleous: Removed

tools/install_prereqs.sh
```diff
@@ -56,8 +56,14 @@ echo "Dependencies are installed at $PREFIX"
...
-# install onnx package for unit tests
-pip3 install onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
+if python3 -c 'import sys; assert sys.version_info >= (3, 10)' > /dev/null; then
+  export CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+  python3.10 -m pip onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
+else
+  # install onnx package for unit tests
+  pip3 install onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
+
+  # pin version of protobuf in Python for onnx runtime unit tests
```
umangyadav, line 66: This comment looks misplaced.

tools/install_prereqs.sh
```diff
@@ -56,8 +56,14 @@ echo "Dependencies are installed at $PREFIX"
 # Install deps with rbuild
 rbuild prepare -d $PREFIX -s develop
 
-# install onnx package for unit tests
-pip3 install onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
+if python3 -c 'import sys; assert sys.version_info >= (3, 10)' > /dev/null; then
+  export CMAKE_ARGS="-DONNX_USE_PROTOBUF_SHARED_LIBS=ON"
+  python3.10 -m pip onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
```
umangyadav, line 61:
```suggestion
  python3 -m pip onnx==1.10.2 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==23.0
```
It should fall back to `python3.10` automatically if `if` condition using `python3` was run and true. 

But what was the issue that required a separate install for python3.10?  

Is that multiple pythons versions are in 22.04 ?

tools/install_prereqs.sh
TedThemistokleous: I can remove it

Dockerfile_Ubuntu_2204
TedThemistokleous: I dont disagree but we have CI tied to that too. Do we want to do this part of this PR or a down the road thing?

tools/install_prereqs.sh
TedThemistokleous: Yeah there is, the default is 3.10 which needs these installs and flags to have make check working correctly

tools/install_prereqs.sh
umangyadav: Keep it but move it to correct line.

Dockerfile_Ubuntu_2204
umangyadav: > Do we want to do this part of this PR or a down the road thing? 

I think it should be easy to modify CI. It should just be about changing paths for `docker build -f Dockerfile.docker`, if so we can do it in this PR.

pfultz2: > I think it is better to move all dockerfiles inside Dockerfiles/ folder.

We should have one Dockerfile a the top-level, and other docker files should be moved to `tools/docker`, and docker files(except the top-level Dockerfile) should be named *.docker.

Dockerfile_Ubuntu_2204
pfultz2: > I suppose we can migrate the specifics to the install_prereqs.sh script for 20.04 , and then pass in which target dist we want somehow for our FROM?

The problem is not just the differences but the large amount that is copied. I dont think we can move everything into install_preqs.sh(that also can destroy any docker layer caching). 

Instead, we should either upgrade the current Dockerfile to use ubuntu 22.04(which seems like a seperate PR), or just drop this extra dockerfile thats not tested(and I would prefer not to increase CI jobs to test it though).

tools/install_prereqs.sh
TedThemistokleous: Moved

Dockerfile_Ubuntu_2204
TedThemistokleous: Done

umangyadav (approved):

Dockerfile_Ubuntu_2204
TedThemistokleous: I mean I would agree with updating the currently docker to 22.04 but then preserving the older dockerfile in Dockerfiles/Docker_<version>.dockerfile so if we need to revert back/testing would be a good historical record of what's worked / used by us

Dockerfiles/Dockerfile_Ubuntu_2204.dockerfile
```diff
@@ -0,0 +1,127 @@
...
+
+ENV MIOPEN_FIND_DB_PATH=/tmp/miopen/find-db
+ENV MIOPEN_USER_DB_PATH=/tmp/miopen/user-db
+ENV LD_LIBRARY_PATH=$PREFIX/lib
+
+# Setup ubsan environment to printstacktrace
+ENV UBSAN_OPTIONS=print_stacktrace=1
+ENV ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1
+RUN ln -s /opt/rocm/llvm/bin/llvm-symbolizer /usr/bin/llvm-symbolizer
+
```
pfultz2, line 127: Dockerfiles should go under `tool/docker`. Files and directories should all be lowercased(except the top-level Dockerfile). And no reason to repeat docker twice, you can just name it `ubuntu-2204.docker`.

Dockerfile_Ubuntu_2204
pfultz2: > I mean I would agree with updating the currently docker to 22.04 but then preserving the older dockerfile in Dockerfiles/Docker_.dockerfile so if we need to revert back/testing would be a good historical record of what's worked / used by us

We can get the old docker from git. No need to duplicate it.

Dockerfiles/Dockerfile_Ubuntu_2204.dockerfile
TedThemistokleous: done

tools/install_prereqs.sh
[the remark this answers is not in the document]
TedThemistokleous: Done

Dockerfile_Ubuntu_2204
TedThemistokleous: So should I just remove the 22.04 docker, or is the plan to migrate it in another PR at some target date?

causten (changes requested):

CharlieL7 (approved):
LGTM, testing ubuntu2204 dockerfile on my Navi21 system just in case

CharlieL7:
> LGTM, testing ubuntu2204 dockerfile on my Navi21 system just in case

Looks to work

PR #1787, opened 2023-05-30
Fix shape typo in API test

CharlieL7:
* Fixing a typo in these test I made.

### DIFF
--- a/test/api/test_gpu.cpp
+++ b/test/api/test_gpu.cpp
@@ -167,7 +167,7 @@ TEST_CASE(dynamic_batch_load_and_run_offload)
 
     CHECK(shapes_before.size() == outputs.size());
     CHECK(bool{outputs.front().get_shape() ==
-               migraphx::shape(migraphx_shape_float_type, {2, 1, 3, 3})});
+               migraphx::shape(migraphx_shape_float_type, {2, 2, 3, 3})});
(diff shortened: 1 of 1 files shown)


umangyadav (approved):

test/api/test_gpu.cpp
```diff
@@ -167,7 +167,7 @@ TEST_CASE(dynamic_batch_load_and_run_offload)
 
     CHECK(shapes_before.size() == outputs.size());
     CHECK(bool{outputs.front().get_shape() ==
-               migraphx::shape(migraphx_shape_float_type, {2, 1, 3, 3})});
+               migraphx::shape(migraphx_shape_float_type, {2, 2, 3, 3})});
```
umangyadav, line 170: Why wasn't this failing on the CI ?

CharlieL7: It looks like using `CHECK` rather than `EXCEPT` doesn't trigger a test failure. I should probably make an issue about that?

umangyadav: > I should probably make an issue about that? 

Yes. and also change this test to use EXPECT for now.

pfultz2 (approved):

TedThemistokleous (approved):

PR #1793, opened 2023-05-30
Load python files in the driver

pfultz2:
The `--py` output can be loaded back in the driver. This will embed the migraphx interperter so we can execute the python directly. There is a migraphx_py library which will dynamically load the version of the library for python version is available on the system.


src/driver/main.cpp
```diff
@@ -253,7 +254,9 @@ struct loader
                 else if(ends_with(file, ".pb"))
                     file_type = "tf";
                 else if(ends_with(file, ".json"))
-                    file_type = "json";
+                    file_type = "py";
```
umangyadav, line 257:
```suggestion
                    file_type = "json";
```

src/dynamic_loader.cpp
```diff
@@ -29,6 +29,8 @@
 #include <utility>
 #include <dlfcn.h>
 
+#include <iostream>
```
umangyadav, line 32: is it used ?

cmake/PythonModules.cmake
```diff
@@ -38,12 +38,22 @@ macro(find_python version)
...
+        endif()
         separate_arguments(_python_includes UNIX_COMMAND "${_python_include_args}")
+        separate_arguments(_python_ldflags UNIX_COMMAND "${_python_ldflags_args}")
         string(REPLACE "-I" "" _python_includes "${_python_includes}")
         add_library(python${version}::headers INTERFACE IMPORTED GLOBAL)
         set_target_properties(python${version}::headers PROPERTIES
             INTERFACE_INCLUDE_DIRECTORIES "${_python_includes}"
         )
+        add_library(python${version}::runtime INTERFACE IMPORTED GLOBAL)
+        set_target_properties(python${version}::runtime PROPERTIES
```
umangyadav, line 53: Why does it require `python::runtime` ?  All the header files are only including `pybind/*.hpp`.

src/dynamic_loader.cpp
```diff
@@ -48,7 +48,7 @@ struct dynamic_loader_impl
 #pragma GCC diagnostic ignored "-Wignored-attributes"
 #endif
     dynamic_loader_impl(const fs::path& p, std::shared_ptr<tmp_dir> t = nullptr)
-        : handle(dlopen(p.string().c_str(), RTLD_LAZY),
+        : handle(dlopen(p.string().c_str(), RTLD_GLOBAL | RTLD_NOW),
```
umangyadav, line 51: Can it work with `RTLD_LAZY` ?  Not sure but having it `RTLD_NOW` would force it to resolve all the symbols.

src/py/CMakeLists.txt
```diff
@@ -23,14 +23,24 @@
 #####################################################################################
 
 option(MIGRAPHX_ENABLE_PYTHON "Enable python bindings" ON)
+add_library(migraphx_py py_loader.cpp)
```
umangyadav, line 26: can this be named differently? It already has `migraphx_py_${PYTHON_VERSION}` later for `py.cpp`

src/py/CMakeLists.txt
```diff
@@ -23,14 +23,24 @@
...
+target_include_directories(migraphx_py PRIVATE include)
+target_link_libraries(migraphx_py PUBLIC migraphx)
+rocm_install_targets(TARGETS migraphx_py INCLUDE include)
```
umangyadav, line 29: Why does it need to include `include` headers ?

src/driver/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ add_custom_command(
 set_directory_properties(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/driver)
 rocm_clang_tidy_check(driver)
 
-target_link_libraries(driver migraphx_all_targets migraphx_onnx migraphx_tf)
+target_link_libraries(driver migraphx_all_targets migraphx_onnx migraphx_tf migraphx_py)
```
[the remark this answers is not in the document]
pfultz2: `migraphx_py` library is the entry point to use the API but its just a shim layer which will forward to `migraphx_py_${PYTHON_VERSION}` depending on which python version is installed. That is why its called `py_loader.cpp` since it will just load the function in `py.cpp` using the correct python version.

src/dynamic_loader.cpp
pfultz2: It doesnt work. It seems like `RTLD_LAZY` would not be as useful since we need to load everything right away. We can add some kind of lazy flag in the future if need to load them lazily.

Jenkinsfile
```diff
@@ -28,7 +28,7 @@ def rocmtestnode(Map conf) {
             rm -rf build
             mkdir build
             cd build
-            cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_DEV=On ${flags} ..
+            cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_DEV=On -DCMAKE_EXECUTE_PROCESS_COMMAND_ECHO=On ${flags} ..
```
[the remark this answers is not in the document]
pfultz2: It is for debugging, but it seems good to have in the CI always.

src/py/CMakeLists.txt
pfultz2: No, its named accordingly. `migraphx_py` is the library you would link, which will load the corresponding `migraphx_py_${PYTHON_VERSION}` depending which python version is installed on the system.

cmake/PythonModules.cmake
umangyadav: Looks like we can link to `pybind11::embed` for that https://pybind11.readthedocs.io/en/stable/advanced/embedding.html#getting-started

src/py/py.cpp
```diff
@@ -0,0 +1,69 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#include <migraphx/config.hpp>
+#include <migraphx/program.hpp>
+#include <migraphx/dynamic_loader.hpp>
+#include <migraphx/file_buffer.hpp>
+#include <pybind11/embed.h>
+#include <pybind11/eval.h>
```
umangyadav, line 29: https://github.com/pybind/pybind11/blob/d0232b119f5f0cf2bf6cecacaf099c301c811ca8/include/pybind11/embed.h#L13

i think only need to include `embed.h`.

src/driver/main.cpp
```diff
@@ -241,21 +242,28 @@ struct loader
...
+    static std::string get_file_type(const std::string& file)
+    {
+        if(ends_with(file, ".onnx"))
+            return "onnx";
+        else if(ends_with(file, ".pb"))
+            return "tf";
+        else if(ends_with(file, ".json"))
+            return "json";
+        else if(ends_with(file, ".py"))
+            return "py";
```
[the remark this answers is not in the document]
umangyadav: for the `mxr` file we have `--migraphx` flag to indicate input type. 

Can we perhaps use `--py-out` for dumping python and `--py-in` for input ?

src/dynamic_loader.cpp
umangyadav: > It doesnt work. 

Curions, why it wouldn't work. Lazy would load on need basis. If it requires something it would load it. `RTLD_GLOBAL` would load everything.  I am trying to figure out where it would break for `LAZY`.

cmake/PythonModules.cmake
pfultz2: Yea that doesnt work when we are building for multiple versions of python.

src/driver/main.cpp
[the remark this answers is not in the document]
pfultz2: Yea, I dont think we would have a reason to load python without a python extensions. So I would prefer keeping the flags as they are. `--py-out` and `--py-in` are kind of ugly to write.

src/dynamic_loader.cpp
pfultz2: > RTLD_GLOBAL would load everything.

No, `RTLD_GLOBAL` makes the symbols available for subsequently loaded libraries to use. 

>  I am trying to figure out where it would break for LAZY.

With how python loads the libraries it causes undefined references from missing dependencies. The only way to it to work is to use `RTLD_GLOBAL | RTLD_NOW`. Of course, you can try to see if you can find another way to make it work with `RTLD_LAZY`.

src/driver/main.cpp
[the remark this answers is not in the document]
umangyadav: Okay. Would need to add `--py` flag here saying it prints out program using python API
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/docs/driver/read.rst

src/py/py.cpp
umangyadav: Does it work without this `<embed.h>` ?

src/driver/main.cpp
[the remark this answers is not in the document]
pfultz2: Doc added.

src/py/py.cpp
pfultz2: I removed the eval.h header.

pfultz2:
@umangyadav @TedThemistokleous Anymore feddback on this PR?

umangyadav (approved):

TedThemistokleous (approved):

TedThemistokleous:
> @umangyadav @TedThemistokleous Anymore feddback on this PR?

Nope. LGTM Was just waiting for CI

Issue #1695, opened 2023-04-18
Driver offload_copy and loading mxr file


CharlieL7:
* Currently you need to pass `--enable-offload-copy` to the driver when running a model that compiled with offload_copy.
  * If you don't it will throw an error about GPU arguments or memory objects
* We should either make the error point towards the actual problem (not passing --enable-offload-copy) or have the mxr file store the compile option.
  * Note that there's probably not a good way to tell if a compiled model used the flag or not without storing the flag.

PR #1802, opened 2023-06-02
Print message from driver if offload copy is set for compiled program

umangyadav:
MXR files can be produced with/without offload copy. 

By default offload-copy is false for the driver. 

When MXR file compiled with offload-copy is run with driver  without `--enable-offload-copy` flag, it would generate parameters and Copy them to GPU before it starts executing the model. 

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/49b341d3c1e5accb3b13e245615d2f03f663d4fa/src/driver/perf.cpp#L55

During the copy, it first registers the CPU memory.  https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/49b341d3c1e5accb3b13e245615d2f03f663d4fa/src/targets/gpu/include/migraphx/gpu/hip.hpp#L121

But later since the model would have another duplicate `hip::copy_to_gpu`, at that time registration would fail. 

By doing an automatic deduction of `offload_copy` on the compiled model, this issue can be avoided.

### DIFF
--- a/src/driver/main.cpp
+++ b/src/driver/main.cpp
@@ -459,8 +459,8 @@ struct compiler
         {
             if(ct.target_name == "gpu" && is_offload_copy_set(p))
             {
-                std::cout << "Compiled program likely has offload_copy set, Try passing "
-                             "`--enable-offload-copy` if program fails.";
+                std::cout << "Compiled MIGraphX program likely has offload_copy set, Try passing "
+                             "`--enable-offload-copy` if program run fails.";
             }
             return p;
         }
--- a/src/driver/main.cpp
+++ b/src/driver/main.cpp
@@ -460,14 +460,14 @@ struct compiler
             if(ct.target_name == "gpu" && is_offload_copy_set(p))
             {
                 std::cout << "Compiled MIGraphX program likely has offload_copy set, Try passing "
-                             "`--enable-offload-copy` if program run fails.";
+                             "`--enable-offload-copy` if program run fails.\n";
             }
             else if(ct.target_name == "gpu")
             {
                 std::cout
                     << "Compiled MIGraphX program likely compiled without offload_copy set, Try "
                        "removing "
-                       "`--enable-offload-copy` flag if passed to driver, if program run fails.";
+                       "`--enable-offload-copy` flag if passed to driver, if program run fails.\n";
             }
             return p;
         }
--- a/src/driver/main.cpp
+++ b/src/driver/main.cpp
@@ -455,8 +455,10 @@ struct compiler
     {
         auto p = l.load();
         // Dont compile if its already been compiled
+
         if(p.is_compiled())
         {
+
             if(ct.target_name == "gpu" && is_offload_copy_set(p))
             {
                 std::cout << "Compiled MIGraphX program likely has offload_copy set, Try passing "
--- a/src/driver/main.cpp
+++ b/src/driver/main.cpp
@@ -456,7 +456,11 @@ struct compiler
         auto p = l.load();
         // Dont compile if its already been compiled
         if(p.is_compiled())
+        {
+            if(ct.target_name == "gpu")
+                co.offload_copy = is_offload_copy_set(p);
             return p;
+        }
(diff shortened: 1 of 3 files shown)


CharlieL7:
This looks like it's related to #1695. How would this fair with something like what we do with `NMS`? Where we do the calculation on the host device?

umangyadav:
> This looks like it's related to #1695. How would this fair with something like what we do with `NMS`? Where we do the calculation on the host device?

Yes looks like this solves #1695.  

It would work with NMS as well as long as it is compiled for GPU. 

Case 1: For NMS without offload : 

param_on_gpu --> copy_from_gpu --> NMS -->copy_to_gpu-->return_back_to_gpu

This PR logic checks for `copy_to_gpu` on the params and `copy_from_gpu` for the return args. Therefore it would classify case 1 as without_offload. 

Case 2 : NMS with offload_copy : 

param_on_cpu  --> copy_to_gpu --> copy_from_gpu --> NMS --> copy_to_gpu --> copy_from_gpu --> return. 

PR logic would classify it as compiled with Offload correctly.

This is intended as more of a HINT than a correct classification. I can see that it wouldn't work with multi-target compilation.

umangyadav:
Case 1 : NMS compiled without offload_copy.

```
root@e485da2797f4:~/repo/AMDMIGraphX/build(driver_offload)# ./bin/driver compile ../test/onnx/nms_test.onnx --binary --output nms.mxr
Compiling ...
Reading: ../test/onnx/nms_test.onnx
root@e485da2797f4:~/repo/AMDMIGraphX/build(driver_offload)# ./bin/driver run nms.mxr
Compiling ...
Reading: nms.mxr
Compiled MIGraphX program likely compiled without offload_copy set, Try removing `--enable-offload-copy` flag if passed to driver, if program run fails.
Allocating params ...
module: "main"
...
...
...
root@e485da2797f4:~/repo/AMDMIGraphX/build(driver_offload)# ./bin/driver run nms.mxr --enable-offload-copy
Compiling ...
Reading: nms.mxr
Compiled MIGraphX program likely compiled without offload_copy set, Try removing `--enable-offload-copy` flag if passed to driver, if program run fails.
Allocating params ...
Memory access fault by GPU node-2 (Agent handle: 0x285c200) on address 0x28b1000. Reason: Page not present or supervisor privilege.
Aborted (core dumped)

```

Case 2: NMS compiled with Offload-copy : 
```
root@e485da2797f4:~/repo/AMDMIGraphX/build(driver_offload)# ./bin/driver compile ../test/onnx/nms_test.onnx --enable-offload-copy --binary --output nms.mxr
Compiling ...
Reading: ../test/onnx/nms_test.onnx
root@e485da2797f4:~/repo/AMDMIGraphX/build(driver_offload)# ./bin/driver run nms.mxr 
Compiling ...
Reading: nms.mxr
Compiled MIGraphX program likely has offload_copy set, Try passing `--enable-offload-copy` if program run fails.
Allocating params ...

```

CharlieL7 (approved):

TedThemistokleous (approved):

src/driver/main.cpp
```diff
@@ -455,8 +455,24 @@ struct compiler
...
+
         if(p.is_compiled())
+        {
+
+            if(ct.target_name == "gpu" && is_offload_copy_set(p))
+            {
+                std::cout << "Compiled MIGraphX program likely has offload_copy set, Try passing "
+                             "`--enable-offload-copy` if program run fails.\n";
+            }
+            else if(ct.target_name == "gpu")
```
kahmed10, line 467:
```suggestion
            if(ct.target_name == "gpu" and is_offload_copy_set(p) and not co.offload_copy)
            {
                std::cout << "Compiled MIGraphX program likely has offload_copy set, Try passing "
                             "`--enable-offload-copy` if program run fails.\n";
            }
            else if(ct.target_name == "gpu" and co.offload_copy)
```
I think we should check what the offload copy flag is set to before printing any messages.

PR #1299, opened 2022-07-11
Add pass to rewrite gelu as fast gelu

turneram:
Rewrite_gelu pass replaces the gelu formula of x * (1/2) * (1 + erf(x/sqrt(2))) with the sigmoid approximation of x * Sigmoid(x * 1.702)

PR #1810, opened 2023-06-05
Conditionally enable GeLU approximation 

umangyadav:
Sigmoid approximation for GeLU was introduced in #1299  for  Fp16.  The sigmoid approximation is known to get better perf but lower accuracy. https://arxiv.org/pdf/1606.08415.pdf
image

This PR adds a compile-time switch for enabling GeLU approximation

### DIFF
--- a/src/targets/gpu/target.cpp
+++ b/src/targets/gpu/target.cpp
@@ -74,6 +74,8 @@ namespace gpu {
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_SCHEDULE_PASS)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_REDUCE_FUSION)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_NHWC)
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_FAST_GELU)
+
 struct id_pass
 {
     std::string name() const { return "id"; }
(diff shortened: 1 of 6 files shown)


src/targets/gpu/target.cpp
```diff
@@ -121,7 +121,7 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         inline_module{},
         rewrite_pooling{},
         dead_code_elimination{},
-        rewrite_gelu{},
+        enable_pass(options.fast_math, rewrite_gelu{}),
```
turneram, line 124: It might be better to add its own flag so it can be independent from the other fast math optimizations. We could just add `MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_FAST_GELU)` 
at line 77 and make this `enable_pass(not enabled(MIGRAPHX_DISABLE_FAST_GELU{}), rewrite_gelu{}),` Then it would also be a run time option instead of a compile time option.

umangyadav: Yes, I thought of ENV flag as well. 

But then we have many Env flags around eg. fast_softmax, convert_instancenorm and `disable_fast_gelu` etc. 

Also i found that we are not even using the `fast_math` compile option anywhere.  We may want to unify later all env flag under `fast_math` probably.

umangyadav: Made it a env flag for now.  we can discuss more on this when paul's back.

turneram (approved):

shivadbhavsar (approved):

kahmed10 (approved):

src/targets/gpu/target.cpp
pfultz2: This should just use the `fast_math` flag. The reason we dont use that for fast softmax is because it can easily lead to nans rather than less accurate results.

PR #1814, opened 2023-06-06
Revert "Handle broadcasts across dot and concat (#1689) (#1731)"

TedThemistokleous:
This reverts commit a46f378eb038cbe3055a09541685f509109ac937.

Needs to occur as new kernels appear to require additional tuning with this change. Target for 5.7 instead of 5.6


umangyadav:
This revert would degrade performance.  

> new kernels appear to require additional tuning with this change 

Is that additional tuning not supported by rocblas ?

umangyadav (approved):

Issue #1766, opened 2023-05-19
padding computation in Pooling op.


bpickrel:
The Pooling operator does not give correct results for pooling windows that include padding.  The computation that is performed ignores the values that are in the padding area.  For example, for average pooling with a 2x2 window, we'd expect that when the window is placed in a corner of the original data such that it contains a 1 and three 0's, the pooled value should be 0.25.  Instead, it returns 1.  This has presumably escaped attention until now because all of the existing unit tests either don't have a "padding" attribute or set the padding size to 0.

The following code snippet and result illustrate the problem for max pooling.  For max pooling, the pooled value in every position should be 0:

```
TEST_CASE(maxpool_dyn_test)
{
    migraphx::program p;
    auto* mm = p.get_main_module();
    auto s   = migraphx::shape{migraphx::shape::float_type, {{1, 4}, {1, 3}, {2, 4}, {2, 4}}};
    auto x   = mm->add_parameter("X", s);
    mm->add_instruction(migraphx::make_op("pooling",
                                          {{"mode", migraphx::op::pooling_mode::max},
                                           {"lengths", {2, 3}},
                                           {"padding", {1, 2}},
                                           {"stride", {1, 1}}}),
                        x);
    p.compile(migraphx::make_target("ref"));

    std::vector<float> data{-1, -2, -3, -4};
    /**
     * 0 0  0  0 0 0 
     * 0 0 -1 -2 0 0
     * 0 0 -3 -4 0 0      0-padding should look like this 
     * 0 0  0  0 0 0 

     
    */
    // migraphx::shape input_fixed_shape{migraphx::shape::float_type, {1, 3, 2, 2}};
    migraphx::shape input_fixed_shape{migraphx::shape::float_type, {1, 1, 2, 2}};
    migraphx::parameter_map params;
    params["X"] = migraphx::argument(input_fixed_shape, data.data());
    auto result = p.eval(params).back();
    std::vector<float> results_vector(12);
    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
printf("**** result: ");
for(auto aa : results_vector) printf(" %f,", aa);printf("\n");
    std::vector<float> gold{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
    EXPECT(migraphx::verify_range(results_vector, gold));
}
```

[...]

PR #1823, opened 2023-06-08
Pooling op. calculation changes

bpickrel:
Changes to the way Pooling operation calculates pooling when there's padding.  Old code would clip off any padding values before computing; for instance if an Average pooling window contained  0 1 2 where the 0 is padding, the result  was 1.5 instead of 1.0.    See [Issue 1766](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1766)

_Edit 6/29/2023:  After more discussion and clarification of requirements, the original pooling behavior is not wrong.  The Onnx spec allows either of the two rules to be used, with a Boolean attribute `count_include_pad` to select which to use.  The default is false (original behavior).  Migraphx' Onnx parsing supports this attribute by inserting a `pad` operation into the program before the pooling op. but the pooling operation does not support it directly and cannot because MIOpen doesn't support it.  The Onnx parsing workaround has the shortcoming that it can't be applied at runtime in the case that auto-padding is necessary with a dynamic input shape, and therefore this attribute is not allowed in combination with a dynamic input shape when reading Onnx files.  Adding dynamic input + count_include_padding support would require either writing an Issue with MIOpen team to add count_include_pad support, or finding a workaround in MIGraphX code that would potentially impact performance.  Both are out of scope for now.  We do not know of a requirement for this feature in any of the ML models we're endeavoring to support._

_Keeping this PR open while not changing the original calculation rule, because it adds tests for nonzero padding sizes that didn't exist before, and contains fixes for logic and indexing problems that these tests turned up._

This logic is based on the behavior of Onnx Runtime where the Onnx documentation didn't specify behavior
exhaustively; for instance when to clip the pooling kernel if both ceil_mode and padding are specified.

Also corrected some erroneous out-of-bounds checks.

Added and modified tests to include nonzero padding and the ceil_mode flag (which affects output size)


src/include/migraphx/op/pooling.hpp
```diff
@@ -112,7 +136,7 @@ struct pooling
         const shape& input = inputs.at(0);
         auto padding_size  = padding.size();
         size_t kdims       = input.ndim() - 2;
-        if(input.ndim() != padding_size / 2 + 2 and input.ndim() != padding_size + 2)
+        if(input.ndim() * 2 != padding_size + 4 and input.ndim() != padding_size + 2)
```
TedThemistokleous, line 139: Are we not concerned of overflowing ndim() since we're now shifting it up a bit in size? 

Also, i suppose if ndim == 2  and padding_size == 0, this would trigger? Is that the intent here?

bpickrel: --Are we not concerned of overflowing ndim() since we're now shifting it up a bit in size?

No, it represents the dimensions of a tensor and is a 64-bit size_t; I'd call it very safe to assume the input is smaller than that.

-- Also, i suppose if ndim == 2 and padding_size == 0, this would trigger? Is that the intent here?
In that case, input.ndim() * 2 equals padding_size + 4 which would not trigger the exception.  But it's wrong because the op. assumes 1 dimension for batch; 1 dimension for channels, and 1 or more channels for dimensions.  I should add such a check.

bpickrel:
A new issue has come up in CI testing:  When I modified the test `verify/test_avg_pooling_1d` to have nonzero padding size, a failure occurred which @pfultz2 tells me comes from OneDNN.  I replicated this locally by setting build variable `MIGRAPHX_ENABLE_CPU:BOOL=On`  Some combinations of padding and kernel (`lengths`) sizes work and some don't.  A look at [OneDNN documentation for convolution](https://oneapi-src.github.io/oneDNN/dev_guide_convolution.html#doxid-dev-guide-convolution) shows a rule that no padding dimension can be greater than 1/2 the corresponding kernel size, but trial and error shows that behavior doesn't always follow the rule--some tests with larger padding were accepted; some with less padding were rejected.  At this point I don't know how to predict behavior so I modified the test to have a padding size that passes.  I also wrote code to check for this condition in `pooling.hpp` but commented it out since it seems more a burden than a help.  

I also found another OneDNN rule in the documentation that we don't follow:  No kernel dimension can be larger than 14.

Question:  what restrictions are we willing to impose on MIGraphX to be compatible with OneDNN limitations?  Adding @kahmed10

test/verify/test_avg_pooling_1d.cpp
```diff
@@ -31,11 +31,12 @@ struct test_avg_pooling_1d : verify_program<test_avg_pooling_1d>
 {
     migraphx::program create_program() const
     {
+        // pooling test with nonzero padding
         migraphx::program p;
         auto* mm = p.get_main_module();
         auto input =
-            mm->add_parameter("x", migraphx::shape{migraphx::shape::float_type, {1, 3, 5}});
-        auto op = migraphx::op::pooling{migraphx::op::pooling_mode::average, {0}, {1}, {3}};
+            mm->add_parameter("x", migraphx::shape{migraphx::shape::float_type, {1, 3, 7}});
```
pfultz2, line 38: Dont change this test. Just add a new test(like `test_avg_pooling_1d_pad`) to set nonzero padding.

bpickrel: Done.

bpickrel:
Wrote separate issue for OneDNN failure. 

[Issue #1913](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1913)

src/include/migraphx/op/pooling.hpp
```diff
@@ -42,16 +42,48 @@ namespace op {
...
-    pooling_mode mode                = {pooling_mode::average};
+    pooling_mode mode = {pooling_mode::average};
+
+    // The amount each dimension is padded, both before and after.  Can be
+    // either the same size as lengths, or twice as long.  When padding
+    // is double the length of lengths, the first n values are read as
+    // pre-padding for each dimension and the last n values are post-padding.
+    // (The latter is necessary if the total padding desired is an odd number)
+    //
+    //
```
CharlieL7, line 53:
```suggestion
    // Padding along each input dimension
    // Can be ndim or 2*ndim values
    // ndim values means symmetrical padding along each dimension
    // 2*ndim values sets left and right padding for each dimension
```

src/include/migraphx/op/pooling.hpp
```diff
@@ -42,16 +42,48 @@ namespace op {
...
     std::vector<std::size_t> padding = {0, 0};
-    std::vector<std::size_t> stride  = {1, 1};
+
+    // Size of stride to take from one placement of the pooling kernel to the next.
+    // This is distinct from the strides used by the shape class.  If set the same as lengths, the
+    // kernel tiles over the input with no gaps or overlaps. Must be the same size as lengths.
```
CharlieL7, line 58:
```suggestion
    // Striding for the pooling kernel
```
Too much information. Developer can look at any outside pooling description to understand this.

src/include/migraphx/op/pooling.hpp
```diff
@@ -42,16 +42,48 @@ namespace op {
...
+    std::vector<std::size_t> stride = {1, 1};
+
+    // dimensions of the pooling kernel or window.  Must have size
+    // 2 smaller than the input tensor rank, since the first
+    // two dimensions are considered
+    // non-spatial: batch size and channel (NCHW layout)
```
CharlieL7, line 64:
```suggestion
    // dimensions of the pooling kernel
```

CharlieL7 (approved):
Minor comments

TedThemistokleous (approved):

PR #1825, opened 2023-06-09
Fix compile warnings for shadowing variable names

umangyadav:


TedThemistokleous (approved):

turneram (approved):

pfultz2:
Where are these compiler warnings coming from? We should probably disable them since these are false positives. These variables are shadowed on purpose to avoid using a moved from value. Renaming them makes the code more error prone as you can accidentally use the moved from value.

umangyadav:
> Where are these compiler warnings coming from? 

They were showing up inside rocm-5.6 or mainline dockers. 

> These variables are shadowed on purpose to avoid using a moved from value. Renaming them makes the code more error prone as you can accidentally use the moved from value. 

Makes sense. but at least for now the functions changed are small enough to notice any errors.

PR #1827, opened 2023-06-09
Enable hipRTC

causten:
switch from hipClang to hipRTC which allows rocm package installs to be located in different directories


umangyadav (approved):

umangyadav (changes requested):
Can we block this untill we figure out cause of CK failure with hipRTC ?

PR #1828, opened 2023-06-09
[Draft] enable MIGraphX on Windows

apwojcik:
The PR enables MIGraphX on Windows. It replaces the previous draft PR #1770.

Features enabled on Windows:

- migraphx
- migraphx C API
- ONNX bridge
- Tensorflow bridge
- Backends: Reference, CPU (oneDNN, ZenDNN work in progress), GPU, FPGA
- Drivers: GPU, HIPRTC
- Kernels cache
- tests

Windows CI is a work in progress.


pfultz2 (changes requested):
This is smaller than the previous PR, but its still quite large. I would like to see this split up into smaller PRs for develop:

- Export headers/macros for all migraphx libraries except libmigraphx_c(since it needs special treatment)
- Export headers/macros for libmigraphx_c
- We can also add CI job that sets the default visibility to hidden so we can ensure these export macros stay up to date
- Embed.cmake updates and changes on the C++ side to use string_view
- `migraphx_all_targets` updates
- Disabling `migraphx_test_private_disable_exception_catch` and optional `BUILD_TESTING`
- Workarounds or fixes needed in C++ to compile on windows

PR #1836, opened 2023-06-13
2+ input multibroadcast and 2+ input dynamic shape insert_common_op

CharlieL7:
* Allows `insert_common_op()` to accept 2+ dynamic shape inputs
* Changes `multibroadcast` to accept an arbitrary number of arguments
  * `multibroadcast` still only broadcasts the first input to the common shape of the set of input shapes
* There doesn't seem to be a place where we test the `common.cpp` functions separately


src/common.cpp
```diff
@@ -77,31 +82,38 @@ std::vector<shape::dynamic_dimension> compute_broadcasted_dyn_dims(shape s0, sha
...
-                               migraphx::to_string_range(s1.dyn_dims()) + "} mismatch!");
-            }
-        });
+    std::transform(s0.dyn_dims().cbegin(),
+                   s0.dyn_dims().cend(),
+                   s1.dyn_dims().cbegin() + offset,
+                   out_dims.begin() + offset,
+                   [&](auto a, auto b) {
+                       if(a == b or b == 1)
+                       {
```
TedThemistokleous, line 91: Looks like your prev errors went away once you combined conditions

TedThemistokleous (approved):

src/common.cpp
```diff
@@ -32,19 +32,24 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
-// Example:
+// Right-handed shape broadcasting rules like in Numpy.
+// example 1:
 // s0 = (3,2,4,5) and s1 = (2,1,1)
 //
 // In this case we need to broadcast (:,1,1) portion of
 // s1 plus broadcast the 1st dimension of s1
```
bpickrel, line 40:
```suggestion
// s1 plus broadcast the 1st dimension of s0
```

test/op_shape_test.cpp
```diff
@@ -1641,6 +1641,115 @@ TEST_CASE(multibroadcast_2in_static_static_error0)
...
+
+TEST_CASE(multibroadcast_3in_dyn_static)
+{
+    std::vector<migraphx::shape::dynamic_dimension> a{{1, 4}, {2, 4, {2}}, {2, 4}};
+    migraphx::shape a_shape{migraphx::shape::float_type, a};
+    std::vector<migraphx::shape::dynamic_dimension> b{{2, 4, {2}}, {2, 4}};
+    migraphx::shape b_shape{migraphx::shape::float_type, b};
+    migraphx::shape c_shape{migraphx::shape::float_type, {5, 1, 1, 1}};
+    expect_shape(
+        migraphx::shape{migraphx::shape::float_type, {{5, 5}, {1, 4}, {2, 4, {2}}, {2, 4}}},
```
bpickrel, line 1706: If all subcases create the same result shape, why not assign it to a variable and reuse it?

test/op_shape_test.cpp
```diff
@@ -1641,6 +1641,115 @@ TEST_CASE(multibroadcast_2in_static_static_error0)
...
+        b_shape);
+}
+
+TEST_CASE(multibroadcast_3in_dyn_dyn)
+{
+    std::vector<migraphx::shape::dynamic_dimension> a{{1, 4}, {2, 4, {2}}, {2, 4}};
+    migraphx::shape a_shape{migraphx::shape::float_type, a};
+    std::vector<migraphx::shape::dynamic_dimension> b{{2, 4, {2}}, {2, 4}};
+    migraphx::shape b_shape{migraphx::shape::float_type, b};
+    std::vector<migraphx::shape::dynamic_dimension> c{{1, 5, {1, 5}}, {1, 1}, {2, 4, {2}}, {2, 4}};
```
bpickrel, line 1731: Same comment as above: If all cases create the same result shape, why not assign it to a variable and reuse it?

src/include/migraphx/common.hpp
```diff
@@ -41,6 +41,9 @@ std::vector<shape::dynamic_dimension> compute_broadcasted_dyn_dims(shape s0, sha
 
 shape common_shape(const std::vector<shape>& shapes);
 
+std::vector<std::size_t> compute_common_lens(const std::vector<shape>& shapes);
+std::vector<shape::dynamic_dimension> compute_common_dyn_dims(const std::vector<shape>& shapes);
```
umangyadav, line 45: Add docstring using `\brief @param @return` . It is pretty obvious what is doing but better to get into habit of adding docstring.

src/include/migraphx/op/multibroadcast.hpp
```diff
@@ -102,19 +102,19 @@ struct multibroadcast
...
                 {
                     return {t, output_dyn_dims};
                 }
-                return {t, compute_broadcasted_dyn_dims(s0, s1)};
+                return {t, compute_common_dyn_dims(inputs)};
             }
             else
             {
-                auto bcast_lens    = compute_broadcasted_lens(s0.lens(), s1.lens());
+                auto bcast_lens    = compute_common_lens(inputs);
```
umangyadav, line 117: Could we do similar check like dynamic shape here that if `out_lens` is set then, no need to recalculate common_lens again ?

src/common.cpp
```diff
@@ -154,34 +166,29 @@ insert_common_args(module& m, instruction_ref ins, std::vector<instruction_ref>
...
-            inputs[1] = m.insert_instruction(
-                ins,
-                make_op("multibroadcast", {{"out_dyn_dims", to_value(c_dyn_dims)}}),
-                inputs[1],
-                inputs[0]);
+                ins, make_op("multibroadcast", {{"out_dyn_dims", to_value(c_dyn_dims)}}), inputs);
         }
+        std::transform(inputs.begin() + 1, inputs.end(), inputs.begin() + 1, [&](auto input) {
+            // uses previous multibroadcast to avoid recalculating the common shape from the
+            // full set of input shapes at runtime
```
umangyadav, line 181: `out_dyn_dims` is already set then why would it even recalculate common shape?

src/include/migraphx/op/multibroadcast.hpp
CharlieL7: No, `out_lens` will never be set for 2+ input version of multibroadcast. It is only used for the 1 input version. If 2+ input multibroadcast uses only static shapes it should be converted to a one input version. We might have to make that a compiler pass in the future.

src/common.cpp
CharlieL7: This is for runtime when the shape is given. By using the previous `multibroadcast` we're only comparing two shapes rather than the whole set.

src/common.cpp
```diff
@@ -32,19 +32,24 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
-// Example:
+// Right-handed shape broadcasting rules like in Numpy.
```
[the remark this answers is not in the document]
CharlieL7: Like doing long addition, move all the shapes to be right justified and compare columns.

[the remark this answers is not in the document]
umangyadav: Okay. Makes sense now, but we probably need better term instead of right-handed shape broadcasting. 
Or you can add what you just described additionally i.e. " move all the shapes to be right justified and compare columns"

src/include/migraphx/op/multibroadcast.hpp
umangyadav: It would be better to put that into comments somewhere that `out_lens` is never set for static shape 2+ inputs. 
But `dyn_out_lens` is set for 2+ input case because during runtime, `dyn_out_lens` wouldn't be used and 2+ inputs would have been converted into static shapes.

src/common.cpp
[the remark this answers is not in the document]
bpickrel: suggested wording:  "// Right-handed shape broadcasting rules as in Numpy.  Last dimension must match last dimension, or be a broadcastable/fixed 1; next-to-last must match, etc."

src/include/migraphx/common.hpp
bpickrel: I think VSCode will do that for you if you start typing `/**`.  I'm not sure what extension does that for me.

bpickrel (approved):
Approved as soon as merge is fixed.  All my remaining comments are cosmetic.

src/common.cpp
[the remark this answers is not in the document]
CharlieL7: Yes, `multibroadcast` has a commutative shape function.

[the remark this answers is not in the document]
CharlieL7: Updated the comments to be clearer. Using the same explanation from numpy.

umangyadav (approved):

umangyadav:
@CharlieL7  I am not sure why but performance tests are failing. Can you take a look into it ?

pfultz2:
Whats the reason why we need multiple inputs for broadcast?

CharlieL7:
> Whats the reason why we need multiple inputs for broadcast?

Easier to broadcast ternary or other higher number of dynamic shape inputs this way. The alternative I had was to do many 2 input broadcasts.

Issue #1822, opened 2023-06-08
investigate Benchmarks failures TIMM


causten:
Collect the error logs

TedThemistokleous:
error related to changes with respect to reshape right now

Current changes to solve these issues is found at : trace_simplify_alg_matches

TedThemistokleous:
related to this effort is #1677  after with speaking with @pfultz2 and doing further debug. It looks like we need to add reshape_copy type of operator to handle some of the edge cases when we hit a reshape operator during the inner_broadcast matcher.

@shivadbhavsar  I heard you started something like this earlier. How far did you get? Is there anything we can reuse? 

from my understanding it looks like we have to do the following

- [ ] reshape becomes "lazy" reshape op between simple inner_broadcasts
- [ ] reshape_copy to handle more complex shapes that we can't perform without modifying memory on compile
- [ ] update unit tests for reshape
- [ ] add unit tests for reshape_copy

@pfultz2  let me know if I've missed anything.

TedThemistokleous:
Current branch of work is in modify_reshapes_pre_1751. needs to be cleaned up but having issue with populating dims() for new reshape_lazy op during lowering. 

@pfultz2

TedThemistokleous:
This is just in review still no more work needed

PR #1838, opened 2023-06-14
Add trace for SIMPLIFY_ALGEBRA matches

TedThemistokleous:
Adding trace for a specific set of a matchers for simplify algebra here

If we use the default MIGRAPHX_TRACE_MATCHES this significantly slows down debugging efforts when debugging modules and associated matches

Added overloaded call to the find_matches() set of functions to handle this. Arisen from debugging issues with reshape/simplify_algebra.

Most likely can be applied for other modules that perform matches as well by adding the trace argument to find_matches

related #1822


src/include/migraphx/matcher.hpp
```diff
@@ -428,6 +428,62 @@ void find_matches(Mod& mod, Ms&&... ms)
     }
 }
 
+/// Find matches for an instruction in the module for per section of matchers
+template <class Mod, class... Ms>
+void find_matches(int trace_mod, Mod& mod, instruction_ref ins, Ms&&... ms)
```
umangyadav, line 433: You can keep just one overload and make line 427 do `find_matches(0, mod, ins, ms...);`

TedThemistokleous: Done

src/include/migraphx/matcher.hpp
```diff
@@ -372,9 +372,9 @@ match::matcher_result find_match(module& modl, M&& m)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_MATCHES)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_VALIDATE_MATCHES)
 
-/// Find matches for an instruction in the module
+/// Find matches for an instruction in the module for per section of matchers
 template <class Mod, class... Ms>
-void find_matches(Mod& mod, instruction_ref ins, Ms&&... ms)
+void find_matches(size_t trace_mod, Mod& mod, instruction_ref ins, Ms&&... ms)
```
umangyadav, line 377:
```suggestion
void find_matches(size_t trace_pass, Mod& mod, instruction_ref ins, Ms&&... ms)
```
IMO, trace_pass would be better name. `MIGRAPHX_TRACE_MATCHES` is also already matching inside the module.

TedThemistokleous: Done

umangyadav (approved):

shivadbhavsar (approved):

PR #1840, opened 2023-06-15
Fix convert operation for NaNs

umangyadav:
Ref implementation of the `convert` op in MIGraphX was using clamping along with type-casting.  It doesn't work for NaNs.

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1758 Added a verify test with `double` type literal NaNs. 

MIGraphX converts double-typed literals to float. This convert operation gets const-folded at compile time using Ref Implementation and doesn't result in NaNs. That resulted in unit-test failures and was reported by QA.

For some reason, this error is only re-producible on certain machines. But I can verify using the following small test that `type-casting+clamping` doesn't work for NaNs. 

```cpp
#include <limits>
#include <iostream>
#include <cmath>


int main() {
    double d = std::numeric_limits<double>::quiet_NaN();
    float converted_f = std::min(std::numeric_limits<float>::max(), std::max(std::numeric_limits<float>::min(), float(d)));
    std::cout << "is nan: " << std::isnan(d) << "\n";
    std::cout << "is converted value a nan: " << std::isnan(converted_f) << "\n";
    std::cout << "is casted value a nan:  " << std::isnan(float(d)) << "\n";
    return 0;
}
```

Running that program print the following: 

```
is nan: 1
is converted value a nan: 0
is casted value a nan:  1
```


test/ref_ops_test.cpp
```diff
@@ -1849,6 +1849,40 @@ TEST_CASE(cosh_dyn_test)
     EXPECT(migraphx::verify_range(results_vector, gold));
 }
 
+TEST_CASE(convert_nan_upcast_test)
```
TedThemistokleous, line 1852: Add a test case for half as well and even a double->float->half to ensure things stay preserved.

TedThemistokleous (commented):
Add another two  test cases going between the three conversions up and down

TedThemistokleous:
@umangyadav what machines was this failing on? What about the other direction of clamping using:

```
float converted_f_upper_clamp = std::max(std::numeric_limits<float>::min(), std::min(std::numeric_limits<float>::max(), float(d)));
```

test/ref_ops_test.cpp
umangyadav: I don't see the need to add a Double convert test (a) it is very unlikely to have triple converts inside a model. (b) Double converts are just two convert operations in tandem. I've individual convert tests, if they work, then double converts would work for sure. 

For the GPU target, we run simplify_algebra/reshapes. Which would convert such double converts to single convert. It doesn't happen for "ref" though.  

I added extra two tests.

umangyadav:
> @umangyadav what machines was this failing on? What about the other direction of clamping using:
> 
> ```
> float converted_f_upper_clamp = std::max(std::numeric_limits<float>::min(), std::min(std::numeric_limits<float>::max(), float(d)));
> ```

It doesn't work either.

TedThemistokleous (approved):

kahmed10 (approved):

PR #1842, opened 2023-06-16
Update CK commit hash and add gfx940 to supported archs

turneram:


umangyadav:
Do we need to add workaround for the assertion failure in CK_GEMM ?

turneram:
> Do we need to add workaround for the assertion failure in CK_GEMM ?

We should address that in another PR. I don't currently have a fix that's acceptable for develop branch.

kahmed10 (approved):

PR #1816, opened 2023-06-06
[mlir] Adding mlir quant_dot operator support

jerryyin:
This pull request adds the mlir quant_dot operator support. It:
 - Adds the quant_dot lowering in rocMLIR
 - Make `quant_dot` to be an anchor op
 - Bumped the commit hash to pull in quant_dot op from migraphx dialect

PR #1848, opened 2023-06-19
Update list of pointwise operators supported by MLIR

krzysz00:
- Bump MLIR commit to include latest supported pointwise ops.
- Expand the MLIR whitelist
- Ensure that operations such as tanh() that don't have integer implementations (at least in MLIR) aren't used within MLIR modules.
- Add additional tests.

Depends on #1816


pfultz2 (approved):

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -191,39 +191,81 @@ struct find_mlir_op
...
+        // Preliminary type check.
+        if(not contains(allowed_types, result_type))
+        {
+            return false;
+        }
+        const std::initializer_list<std::string> any_type_ops = {"@literal", "@param", "@return"};
+        const std::initializer_list<std::string> no_bool_ops  = {"convolution",
+                                                                "quant_convolution",
+                                                                "dot",
+                                                                "quant_dot",
```
jerryyin, line 216: This comment is non-blocking, but I believe the granularity of those can be improved. In a corner case, for example, we don't support int32 type operands for all those ops (yet it is valid migraphx IR).

BTW, should fp16 type conv computation go with the `convolution` op or `quant_convolution` op in migraphx and mlir respectively?

jerryyin (approved):

test/gpu/fuse_mlir.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+    migraphx::run_passes(p, {migraphx::gpu::fuse_mlir{}, migraphx::dead_code_elimination{}});
+}
+
+// cppcheck-suppress definePrefix
+// NOLINTNEXTLINE
+#define VERIFY_MLIR(pass, ref)            \
+    do                                    \
+    {                                     \
+        if(migraphx::gpu::mlir_enabled()) \
```
pfultz2, line 45: Yea I dont think you can check this here, unless we want to move the `mlir_op` outside of the ifdef.

krzysz00: Isn't `mlir_enabled` present even when MLIR is off and then it just statically returns false?

pfultz2: Yes, but the problem is that there is no `mlir_op` so constructing the test fails when `mlir_enabled` returns false. You can either move the condition into the test case at the beginning or move the `mlir_op` out of the `ifdef`.

This is why its failing on the CI.

krzysz00: Ah.

How do y'all generally handle this sort of thing? Are other conditionally-enableable ops defined outside off the relevant ifdef?

I might end up moving the enable_mlir to main

pfultz2: > I might end up moving the enable_mlir to main

Thats seems like a simpler solution.

pfultz2 (approved):

src/targets/gpu/fuse_mlir.cpp
krzysz00: I think fp16 is regular convolution, not the quantized one, unless we want to have fp16 returning fp32 values as an operation.

And true, it's probably not out of the question to be even more strict here and really whitelist the supported conv and dot types - should I do that in this PR?

jerryyin: > I think fp16 is regular convolution, not the quantized one, unless we want to have fp16 returning fp32 values as an operation.

Agreed. Sounds like fp16 is a different model compared to fp32 entirely.

> to be even more strict here and really whitelist the supported conv and dot types - should I do that in this PR?

I think you can do that independently in another PR. I'm not concerned because I haven't really seen any model that would operate on those corner cases. As far as I understand, we and migraphx's support range overlap nicely. Probably worthwhile a ticket to be more defensive here as a reminder.

PR #1855, opened 2023-06-19
Use fast_math flag instead of ENV flag for GELU

umangyadav:


pfultz2 (approved):

umangyadav:
We also have `MIGRAPHX_DISABLE_FP16_INSTANCENORM_CONVERT`  which could be related to fast_math. But it is used during onnx-parsing.  But no way to make use of `fast_math` flag.

kahmed10 (approved):

shivadbhavsar (approved):

PR #1857, opened 2023-06-20
Fix assertion error during verify and make DCE work with tuples

### LINKED ISSUE #1856: Reduce verify fails with Debug builds for FP16 models
It fails on assertion during `replace_instruction`. 

Steps to reproduce `./driver verify --onnx resnet50.onnx --fp16 --reduce` .

umangyadav:
Fixes #1856 and additionally makes DCE work for Tuple Types

`verify` was failing with `reduce` in debug builds here : 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a0fa3742f92239fb60c58d4b73f458efb08ef0ed/src/module.cpp#L349

Assertion was failing because of the following pattern:
`ins --> convert(original_type) --> module_end`. 

And then, it was replacing `ins` with `convert(original_type)` here. https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a0fa3742f92239fb60c58d4b73f458efb08ef0ed/src/quantize_fp16.cpp#L60

`convert(original_type)` would have no outputs while `ins` would have one output and therefore it would fail assertion here : 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a0fa3742f92239fb60c58d4b73f458efb08ef0ed/src/module.cpp#L349

### DIFF
--- a/src/dead_code_elimination.cpp
+++ b/src/dead_code_elimination.cpp
@@ -50,7 +50,9 @@ void dead_code_elimination::apply(module& m) const
             break;
         // Skip instruction with empty shape as output unless its [dynamic, builtin, undefined,
         // identity, allocate]
-        if((not i->get_shape().dynamic() and i->get_shape().elements() == 0) and
+        if((not i->get_shape().dynamic() and
+            (i->get_shape().elements() == 0 and
+             i->get_shape().type() != migraphx::shape::tuple_type)) and
            not(i->name().front() == '@') and not contains({"identity", "allocate"}, i->name()) and
            not i->is_undefined())
             continue;
--- a/src/dead_code_elimination.cpp
+++ b/src/dead_code_elimination.cpp
@@ -49,7 +49,7 @@ void dead_code_elimination::apply(module& m) const
         if(i == last)
             break;
         // Skip instruction with empty shape as output unless its [dynamic, builtin, undefined,
-        // identity, allocate]
+        // identity, allocate or tuple_type]
         if((not i->get_shape().dynamic() and
             (i->get_shape().elements() == 0 and
              i->get_shape().type() != migraphx::shape::tuple_type)) and
--- a/test/dead_code_elimination_test.cpp
+++ b/test/dead_code_elimination_test.cpp
@@ -273,59 +273,17 @@ TEST_CASE(param_not_eliminated)
     EXPECT(p == create_program());
 }
 
-TEST_CASE(eliminate_tuple_ins)
+TEST_CASE(tuple_test)
 {
-    auto create_program = [](bool add_dead_ins) {
-        migraphx::program p;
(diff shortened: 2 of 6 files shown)


test/dead_code_elimination_test.cpp
```diff
@@ -274,4 +273,59 @@ TEST_CASE(param_not_eliminated)
...
+        auto r1  = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 1}}), ret);
+        auto r16 = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 2}}), ret);
+        mm->add_return({r0, r1, r16});
+
+        return p;
+    };
+    auto p = create_program(true);
+    run_pass(p);
+    EXPECT(p == create_program(false));
+}
```
pfultz2, line 329: It seems it would be simpler to add a `tuple_op` under basic_ops.hpp header to create a tuple to test with:

```cpp
struct tuple_op
{
    std::string name() const { return "tuple_op"; }
    migraphx::shape compute_shape(std::vector<migraphx::shape> inputs) const
    {
        return {inputs};
    }
};

TEST_CASE(tuple_test)
{
    migraphx::program p;
    auto one = mm->add_literal(1);
    auto two = mm->add_literal(2);
    mm->add_instruction(tuple_op{}, one, two);
    mm->add_return({one, two});
    auto count = std::distance(mm->begin(), mm->end());
    run_pass(p);
    EXPECT(std::distance(mm->begin(), mm->end()) == (count - 1));
}
```

umangyadav: Done.

test/include/basic_ops.hpp
```diff
@@ -186,6 +186,18 @@ struct nop
     migraphx::shape compute_shape(const std::vector<migraphx::shape>&) const { return {}; }
 };
 
+struct tuple_op
+{
+    std::string name() const { return "tuple_op"; }
+    migraphx::shape compute_shape(std::vector<migraphx::shape> inputs) const { return {inputs}; }
```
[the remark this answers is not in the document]
umangyadav: Done.

src/dead_code_elimination.cpp
```diff
@@ -50,7 +50,9 @@ void dead_code_elimination::apply(module& m) const
             break;
         // Skip instruction with empty shape as output unless its [dynamic, builtin, undefined,
         // identity, allocate]
```
kahmed10, line 52: I think this comment should be updated as well

umangyadav: Done.

CharlieL7 (approved):

TedThemistokleous (approved):

kahmed10 (approved):

PR #1858, opened 2023-06-20
Add a MIGRAPHX_TRACE_MATCHES_FOR env variable to filter the matcher trace

pfultz2:
You can use `MIGRAPHX_TRACE_MATCHES_FOR` to only show matches from source files that match the string passed to it. So to only get matches from simplify_algebra you can do `MIGRAPHX_TRACE_MATCHES_FOR=simplify_algebra`. This will work across all passes instead of being just for one pass. 

It uses `source_location` to get the location. Since `find_matches` is a variadic template function the `source_location` cant be added to the end of the function without needing to make all template parameters explicit. So instead I use a templated class with deduction guides to deduce the parameters.


TedThemistokleous:
I see the idea here, but I don't see how this would work with cases like wanting to only trace matches from two modules  (reshapes, simplify_algebra) but not the others

src/include/migraphx/matcher.hpp
```diff
@@ -420,23 +420,19 @@ void find_matches(size_t trace_pass, Mod& mod, instruction_ref ins, Ms&&... ms)
...
-/// Find matches in a pass
 template <class Mod, class... Ms>
-void find_matches(size_t trace_pass, Mod& mod, Ms&&... ms)
-{
-    for(auto ins : iterator_for(get_module(mod)))
-    {
-        find_matches(trace_pass, mod, ins, ms...);
-    }
-}
+find_matches(Mod& mod, Ms&&... ms) -> find_matches<Mod, Ms...>;
```
umangyadav, line 435: What is this doing ?

src/include/migraphx/matcher.hpp
```diff
@@ -370,31 +371,30 @@ match::matcher_result find_match(module& modl, M&& m)
...
+    const bool trace_for    = not trace_filter.empty() and
+                           (contains(std::string{location.file_name()}, trace_filter) or
+                            contains(std::string{location.function_name()}, trace_filter));
+    bool match = false;
     each_args(
         [&](auto&& m) {
             if(match)
                 return;
-            if(trace > 1 or trace_pass > 1)
+            if(trace > 1)
```
umangyadav, line 392:
```suggestion
            if(trace > 1 or trace_for)
```

src/include/migraphx/matcher.hpp
pfultz2: This is a [deduction guide](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction#User-defined_deduction_guides).

src/include/migraphx/matcher.hpp
```diff
@@ -370,31 +371,30 @@ match::matcher_result find_match(module& modl, M&& m)
...
-        bool validate = enabled(MIGRAPHX_VALIDATE_MATCHES{});
-    bool match        = false;
+void find_matches_for(source_location location, Mod& mod, instruction_ref ins, Ms&&... ms)
+{
+    const int trace         = value_of(MIGRAPHX_TRACE_MATCHES{});
+    const bool validate     = enabled(MIGRAPHX_VALIDATE_MATCHES{});
+    const auto trace_filter = string_value_of(MIGRAPHX_TRACE_MATCHES_FOR{});
+    const bool trace_for    = not trace_filter.empty() and
+                           (contains(std::string{location.file_name()}, trace_filter) or
+                            contains(std::string{location.function_name()}, trace_filter));
```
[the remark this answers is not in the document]
pfultz2: This is the function name where `find_matches` is called, which is usually `$PASS_NAME::apply`.

umangyadav (approved):

TedThemistokleous (approved):

TedThemistokleous:
> > I see the idea here, but I don't see how this would work with cases like wanting to only trace matches from two modules (reshapes, simplify_algebra) but not the others
> 
> We could look at adding syntax to handle multiple cases or possibly some regex(will need to use another library) in the future.

I think this is a bit more involved but I see the point here. This makes things more verbose than adding an extra env var.

PR #1859, opened 2023-06-20
Improvement to ck integration

pfultz2:
- Add a CI job to test CK
- Add `MIGRAPHX_TUNE_CK` env variable to only do tuning for CK
- Continue tuning even when there is invalid configs
- Fix a bug with parallel compilation not using all available threads
- Add additional test for gemms using half types
- Removed int32 as supported type since it doesnt pass our test suite


src/targets/gpu/compile_ops.cpp
```diff
@@ -185,7 +203,10 @@ void par_compile(std::size_t n, F f)
 {
     if(n == 0)
         return;
-    par_for(n, n / value_of(MIGRAPHX_GPU_COMPILE_PARALLEL{}, n), f);
+    auto d = value_of(MIGRAPHX_GPU_COMPILE_PARALLEL{});
```
umangyadav, line 206:
```suggestion
    auto d = value_of(MIGRAPHX_GPU_COMPILE_PARALLEL{}, n);
```

src/targets/gpu/jit/ck_gemm.cpp
```diff
@@ -436,8 +435,10 @@ struct ck_gemm_compiler : compiler<ck_gemm_compiler>
     }
 
     optional<tuning_config>
-    get_tuning_config(context& ctx, instruction_ref ins, const operation& op) const
+    get_tuning_config(context& ctx, instruction_ref ins, const operation& op, bool exhaustive) const
     {
+        if(not exhaustive and not enabled(MIGRAPHX_TUNE_CK{}))
```
umangyadav, line 440:
```suggestion
        if(not exhaustive or not enabled(MIGRAPHX_TUNE_CK{}))
```

src/targets/gpu/compile_ops.cpp
pfultz2: That wont work, because the `n` will be cached when `MIGRAPHX_GPU_COMPILE_PARALLEL` is not set.

pfultz2:
@turneram Any feedback?

umangyadav (approved):

turneram (approved):
If we want to leave int32 enabled, then adding the requirement that m, n, and k are all divisible by 4 appears to maintain correctness. We could also investigate further to get more precise criteria and then add it back in another PR.

PR #1864, opened 2023-06-22
Print max,min,mean and stddev values for TRACE_EVAL = 2

umangyadav:
Helps to find out overflows, underflow when converting data types  between operators

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -365,7 +365,7 @@ void print_statistics(std::ostream& os, const argument& a)
         [&](auto t) {
             os << "Min value: " << *std::min_element(t.begin(), t.end()) << ", ";
             os << "Max value: " << *std::max_element(t.begin(), t.end()) << ", ";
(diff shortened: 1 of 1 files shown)


pfultz2 (approved):

src/program.cpp
```diff
@@ -365,8 +365,8 @@ void print_statistics(std::ostream& os, const argument& a)
         [&](auto t) {
             os << "Min value: " << *std::min_element(t.begin(), t.end()) << ", ";
             os << "Max value: " << *std::max_element(t.begin(), t.end()) << ", ";
-            auto num_elements = t.size();
-            auto mean         = std::reduce(t.begin(), t.end()) / num_elements;
+            auto num_elements = static_cast<double>(t.size());
```
pfultz2, line 368: Just do `double num_elements = t.size()` instead of using a `static_cast`.

shivadbhavsar (approved):

PR #1866, opened 2023-06-22
Add MIGraphX version inside MXR file

umangyadav:
need to detect operators name changes e.g. deconvolution to convolution_backwards.


src/program.cpp
```diff
@@ -605,13 +606,26 @@ std::vector<argument> program::eval(parameter_map params, execution_environment
...
+*/
+const int program_file_version = 6;
 
 value program::to_value() const
 {
     value result;
-    result["version"] = program_file_version;
-    result["target"]  = this->impl->target_name;
+    result["version"]      = program_file_version;
+    result["migx_version"] = get_migraphx_version();
```
pfultz2, line 627: I think it would be better for it be `migraphx_version`.

pfultz2 (approved):

CharlieL7 (approved):

PR #1867, opened 2023-06-22
[mlir] Improve context handling to potentially solve threading bugs

krzysz00:
Update `mlir_program` to only create one dialect registry, and to call registerRocMLIRPasses() (which is needed and may not be thread-safe) exactly once. This might resolve issues with parallel compiles (or it won't, testing that is a future PR).

In addition, use a single thread pool across all contexts. This is recommended practice upstream for libraries that perform a lot of compile jobs, and saves on the overhead of creating and destroying a lot of threads (which MLIR uses to, for example, process functions in parallel).

Finally, stop setting the unregistered dialects flag, as it is not needed.

### DIFF
--- a/src/targets/gpu/mlir.cpp
+++ b/src/targets/gpu/mlir.cpp
@@ -122,6 +122,9 @@ struct mlir_handle
 #define MIGRAPHX_MANAGE_MLIR_HANDLE(T, F) migraphx::gpu::mlir_handle<T, decltype(&F), &F> // NOLINT
 
 using mlir_context           = MIGRAPHX_MANAGE_MLIR_HANDLE(MlirContext, mlirContextDestroy);
+using mlir_thread_pool = MIGRAPHX_MANAGE_MLIR_HANDLE(MlirLlvmThreadPool, mlirLlvmThreadPoolDestroy);
+using mlir_dialect_registry  = MIGRAPHX_MANAGE_MLIR_HANDLE(MlirDialectRegistry,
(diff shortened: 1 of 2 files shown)


jerryyin (approved):
LGTM

src/targets/gpu/mlir.cpp
```diff
@@ -173,16 +176,39 @@ bool has_xdlops(const std::string& target_arch)
...
+        return the_registry;
+    }
+
+    static mlir_thread_pool& get_thread_pool()
+    {
+        static std::once_flag init_guard;
+        static mlir_thread_pool the_pool;
+        // To save on overhead, we create one LLVM thread pool and reuse it
+        // across all MLIR contexts as recommended by MLIR upstream.
+        std::call_once(init_guard, [&]() { the_pool = mlirLlvmThreadPoolCreate(); });
```
pfultz2, line 210: This can just be `static mlir_thread_pool the_pool = mlirLlvmThreadPoolCreate()`.

krzysz00: Can it be? Does C++ impose restrictions on concurrently initializing static variables?

pfultz2: Yes its guarantee in C++11 or later to be thread-safe: https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

pfultz2 (approved):

PR #1875, opened 2023-06-23
Update passes to use offload_copy based on root module

umangyadav:
Needed to run multi-targeted program where "main" isn't the only root module. There could be many root modules other than main.


CharlieL7 (commented):
I understand the change to use `get_root_module()` and including a root module pointer. However you mentioned that there might be multiple root modules with multi-target compilation? I do not understand what benefit that gives us compared to having one root module and sub-modules that go to specific targets.

umangyadav:
> one root module and sub-modules that go to specific targets.
One root module is the "main" module.  This main root module would run on the "Ref" target initially and it would call different target-specific root modules at the top level. 
But those root modules can further have nesting.  So it is possible that GPU module would call into CPU module. In this case GPU root module would have CPU root module nested inside it.  Execution flow would GPU-->CPU-->GPU in this case.

CharlieL7 (approved):
Approved.
Question for my understanding:
If we have offload_copy on and a program that goes GPU->CPU->GPU with the changing root modules, this means that each module would have its own offload copies right? I could see MIGX optimizing the second GPU offloads if we can reuse those from the first GPU section.
On the other hand, if offload_copy is off we would have output parameters for both the GPU sections?

umangyadav:
> each module would have its own offload copies right? 

Yes each root module would have its own parameters and offload copies. 

> I could see MIGX optimizing the second GPU offloads if we can reuse those from the first GPU section. 

Currently, I don't think MIGX is optimizing it but we can try to do it optimization later. Right now idea is that each target root module is compiled as separate self-containing programs. So each target root module would have its own parameters and return values. 

> On the other hand, if offload_copy is off we would have output parameters for both the GPU sections? 

Yes if offload_copy is off that it would have output parameters, but it can't work without offload-copies for now for multi-targets. We don't have a mechanism to put parameters into GPU and copy back the final result from the host. Also for the multi-targets where execution is switching between GPU/CPU/FPGA, it may not make sense to not have offload-copy, compiler would have to inject necessary copies to switch between targets.

kahmed10 (approved):

PR #1893, opened 2023-06-27
add 'make generate' script for Windows

apwojcik:

### DIFF
--- a/tools/generate.cmd
+++ b/tools/generate.cmd
@@ -1,0 +1,43 @@
+::
+:: The MIT License (MIT)
+::
+:: Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
+::
+:: Permission is hereby granted, free of charge, to any person obtaining a copy
+:: of this software and associated documentation files (the "Software"), to deal
+:: in the Software without restriction, including without limitation the rights
+:: to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+:: copies of the Software, and to permit persons to whom the Software is
+:: furnished to do so, subject to the following conditions:
+::
+:: The above copyright notice and this permission notice shall be included in
+:: all copies or substantial portions of the Software.
+::
+:: THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+:: IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


pfultz2:
This should be either a python or cmake script so it uses the same script for both windows and linux.

tools/generate.cmd
```diff
@@ -0,0 +1,43 @@
+::
+:: The MIT License (MIT)
+::
+:: Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
causten, line 4: 2023

IncinX:
> This should be either a python or cmake script so it uses the same script for both windows and linux.

I agree, creating a common Windows and Linux solution is preferred. Our initial set of UIF2 changes were explicitly instructed to not break Linux and creating a unifying script for this would impact it. Since this is not high priority, would you mind if we create a work item for it and address it after we upstream changes since this change works for Windows and has low impact for Linux?

tperry-amd:
> This should be either a python or cmake script so it uses the same script for both windows and linux.

I agree, a unified script for Linux and Windows is preferred. Our mandate for these initial changes was to minimally impact Linux. Are you okay if we generate a work item to address this after all the Windows changes have made it upstream? The reasoning is that this change is blocking Windows and the work for a new unified script has a testing requirement for both Windows and Linux that will take some time and is lower priority compared to enabling the larger Windows effort.

pfultz2:
> The reasoning is that this change is blocking Windows 

`make generate` isnt needed to build migraphx either on windows or linux. Its only needed when a developer changes one of the autogenerated files, which can be done on linux at PR time. So I dont see how this could be a blocker, especially if all the changes has been upstreamed already.

tperry-amd:
> > The reasoning is that this change is blocking Windows
> 
> `make generate` isnt needed to build migraphx either on windows or linux. Its only needed when a developer changes one of the autogenerated files, which can be done on linux at PR time. So I dont see how this could be a blocker, especially if all the changes has been upstreamed already.

Fair point. It is still additional work that affects Linux more directly. Our prerogative for the initial changes was to minimally impact Linux and follow your patterns. The pattern here was to solve this problem with a shell script and we solved it with a shell script for Windows. We can change the pattern and create a unified script that works on both OS but it could potentially affect Linux which increases the testing work.

I think we should consider a change in prerogative. A modest increase in Linux risk is okay if we create a unified solution. We'll pull the change, move it to another branch, and hold onto the branch until we have time to do the unified solution.

PR #1894, opened 2023-06-27
cmake: make cppcheck optional if the tool is not found

apwojcik:


pfultz2 (changes requested):
This should be done in rocm-cmake. Its fine to merge it into the UIF branch, but we wont merge it into develop.

apwojcik:
merged to rocm-cmake

PR #1899, opened 2023-06-28
Enable threading in MLIR

krzysz00:
This commit removes the build options to disable threading and removes the mutex in compile_mlir.

The commit being tested is a draft PR on rocMLIR that'll get merged if this passes.


Dockerfile
```diff
@@ -113,7 +113,7 @@ RUN git clone --single-branch --branch ${ONNXRUNTIME_BRANCH} --recursive ${ONNXR
 
 ADD tools/build_and_test_onnxrt.sh /onnxruntime/build_and_test_onnxrt.sh
 
-RUN cget -p /usr/local install ROCmSoftwarePlatform/rocMLIR@8d25af3b3721c159bb41cc6388e9453b1018c126 -DBUILD_MIXR_TARGET=On -DLLVM_ENABLE_ZSTD=Off -DLLVM_ENABLE_THREADS=Off
+RUN cget -p /usr/local install ROCmSoftwarePlatform/rocMLIR@ 87a47adb1c8c3df8ca355290a53b325af31d1ecc -DBUILD_MIXR_TARGET=On
```
pfultz2, line 116: The space after `@` needs to be removed. That is what is causing the failures in CI right now.

krzysz00: Thanks!

(I'll admit I didn't see your comment until after I'd found the issue from the CI logs, but I do appreciate the note on what I'd done wrong)

Jenkinsfile
```diff
@@ -130,7 +130,10 @@ rocmtest clang_debug: rocmnode('vega') { cmake_build ->
     stage('MLIR Debug') {
         withEnv(['MIGRAPHX_ENABLE_MLIR=1']) {
             def sanitizers = "undefined"
-            def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers}"
+            // Note: the -fno-sanitize= is copied from upstream LLVM_UBSAN_FLAGS.
+            // This is, as best as I can tell, requirede to build LLVM with
+            // both UBSAN and threads.
+            def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize-recover=${sanitizers} -fno-sanitize=vptr,function"
```
krzysz00, line 136: Having gone and looked, this could be it, because LLVM explicitly doesn't use C++'s RTTI and instead does its own thing, which could be making UBSAN confused.

pfultz2: I dont think there is a function sanitizer. The CI errors with `cc: error: unrecognized argument to ‘-fno-sanitize=’ option: ‘function’`.

krzysz00: @pfultz2 What compiler are y'all using for this build? Because upstream LLVM has that exact option

umangyadav:
@krzysz00 I've applied `skip bot checks` label to this PR which would disable accuracy and performance runs for this one. 
Please remove that label when this PR is ready so that it would run accuracy&perf tests.

krzysz00:
Huh, I didn't know that was an option!

(I think this PR is getting somewhere, but, at the very least, I'll need to revert all the extra debug options I put in)

src/targets/gpu/mlir.cpp
```diff
@@ -389,14 +389,20 @@ struct mlir_program
         mlir_operation_state& add_attributes(const std::vector<named_attribute_t>& named_attrs)
         {
             auto attributes = prog->name_attributes(named_attrs);
-            mlirOperationStateAddAttributes(&op_state, attributes.size(), attributes.data());
+            if(!attributes.empty())
```
pfultz2, line 392: Use `not` instead of `!`.

pfultz2: Yea thats how we usually write it in migraphx.

umangyadav (approved):

PR #1902, opened 2023-06-29
add support for rocm 5.6 in CI

causten:
Changes to support ROCm 5.6 for CI


pfultz2:
Lets disable the `misc-use-anonymous-namespace` tidy check. I dont really see why its useful to use an anonymous namespace over `static` function. They are essentially the same.

umangyadav:
> Lets disable the `misc-use-anonymous-namespace` tidy check. I dont really see why its useful to use an anonymous namespace over `static` function. They are essentially the same.

Done. 
Also fixed warnings for the custom ops.

umangyadav (approved):

PR #1914, opened 2023-06-30
Update to Cppcheck 2.11

pfultz2:


src/include/migraphx/op/if_op.hpp
```diff
@@ -71,7 +71,7 @@ struct if_op
         std::unordered_map<std::string, argument> params;
 
         std::set<std::string> pnames;
-        for(const auto& smod : mods)
+        for(const const_module_ref& smod : mods)
```
umangyadav, line 74: Why does it have two `const` ?

src/api/include/migraphx/migraphx.h
```diff
@@ -205,7 +205,7 @@ migraphx_status migraphx_dynamic_dimensions_assign_to(migraphx_dynamic_dimension
 
 migraphx_status
 migraphx_dynamic_dimensions_create(migraphx_dynamic_dimensions_t* dynamic_dimensions,
-                                   const_migraphx_dynamic_dimension_t* ptr,
+                                   const const_migraphx_dynamic_dimension_t* ptr,
```
umangyadav, line 208: `const const_migraphx*` ? Why is that ?

src/module.cpp
```diff
@@ -870,12 +870,11 @@ module::print_py(std::ostream& os,
...
                 os << mname << ".add_literal(";
-                bool use_abs = false;
-                ins->get_literal().visit([&](auto v) {
-                    use_abs = std::none_of(v.begin(), v.end(), [](auto x) { return x < 0; });
-                });
+                const bool use_abs = false;
                 // Disable abs for now
-                use_abs = false;
+                // ins->get_literal().visit([&](auto v) {
+                //     use_abs = std::none_of(v.begin(), v.end(), [](auto x) { return x < 0; });
```
umangyadav, line 876: Why is it commented out ?

src/targets/cpu/target.cpp
```diff
@@ -61,7 +61,7 @@ namespace cpu {
 
 std::string target::name() const { return "cpu"; }
 
-// cppcheck-suppress constParameter
+// cppcheck-suppress constParameterReference
```
umangyadav, line 64: `get_passes` wouldn't change `gctx` AFAIK. Why suppress const-ness ?

src/targets/gpu/jit/ck_gemm.cpp
```diff
@@ -299,6 +299,7 @@ struct ck_gemm_compiler : compiler<ck_gemm_compiler>
         const auto& b_shape = inputs[1];
         const auto& c_shape = inputs.back();
 
+        // cppcheck-suppress unreadVariable
```
umangyadav, line 302: Is it false positive ?

src/targets/gpu/jit/ck_gemm.cpp
```diff
@@ -299,6 +299,7 @@ struct ck_gemm_compiler : compiler<ck_gemm_compiler>
...
         auto rank = a_shape.lens().size();
```
CharlieL7, line 303:
```suggestion
        auto rank = a_shape.ndim();
```

src/include/migraphx/op/if_op.hpp
pfultz2: One is const for the pointer and the other is for the data the pointer points to. I guess it doesnt really make sense to use a const ref here since `const_module_ref` is just a pointer.

src/api/include/migraphx/migraphx.h
pfultz2: First const is for the handle, and the second const is for what the handle points to.

This is using an extra pointer here because it takes an array of `const_migraphx_dynamic_dimension_t` which is passed by pointer. Each `const_migraphx_dynamic_dimension_t` is a pointer as well that points to data that we cant/dont modify. So we are taking a pointer to an array of pointers. Then the additional `const` is because we dont modify the pointers in the array.

src/module.cpp
pfultz2: It needs to be implemented. Cppcheck was complaining about dead code here, so I just commented it out for now.

src/targets/cpu/target.cpp
pfultz2: It gets casted to a non-const `context` which is passed to passes that take the `context` as non-const as well. Cppcheck just doesnt understand the `any_cast`.

src/targets/gpu/jit/ck_gemm.cpp
pfultz2: Yes

src/api/include/migraphx/migraphx.h
umangyadav: Was cppcheck complaning about this pointer not being `const` ? because we also have some other `create()` methods which would also require `const` pointer to `const` object in that case.

umangyadav: If i expand this, it would become `const const migraphx_dynamic_dimension*` . Should it be instead `const _migraphx_dynamic_dimension_t* const ptr` instead?

src/module.cpp
umangyadav: I don't understand how is it deadcode ? It is going over literal and trying to set `use_abs` accordingly and would always execute.

src/api/include/migraphx/migraphx.h
pfultz2: > Should it be instead const _migraphx_dynamic_dimension_t* const ptr instead?

The variable itself is not check for const and this might produce warnings from the compiler when making function parameters const.

src/module.cpp
pfultz2: `use_abs` is assigned `false` and is never assigned `true`. We need to add an API to python for this.

CharlieL7 (approved):

umangyadav (approved):

PR #1920, opened 2023-07-06
Remove use of int8x4 format for the rocblas

### LINKED ISSUE #1916: Remove deprecated `rocblas_query_int8_layout_flag`
rocBLAS has deprecated int8_layout_flag query calls. Need to find workaround or other solution

umangyadav:
Fixes #1916


src/targets/gpu/rocblas.cpp
```diff
@@ -55,9 +55,16 @@ bool get_compute_fp32_flag()
 
 bool get_int8_x4_format(context& ctx)
 {
+    // int8x4 packed format is only available starting from rocblas-v2.38 and it is deprecated in
+    // v3.0 and will be removed in v4.0
+#if ROCBLAS_VERSION_MAJOR >= 2 && ROCBLAS_VERSION_MAJOR < 3 && ROCBLAS_VERSION_MINOR >= 38
```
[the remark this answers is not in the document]
umangyadav: Thanks.

kahmed10 (approved):

src/targets/gpu/rocblas.cpp
```diff
@@ -55,9 +55,16 @@ bool get_compute_fp32_flag()
 
 bool get_int8_x4_format(context& ctx)
 {
+    // int8x4 packed format is only available starting from rocblas-v2.38 and it is deprecated in
+    // v3.0 and will be removed in v4.0
+#if ROCBLAS_VERSION_MAJOR == 2 && ROCBLAS_VERSION_MINOR >= 38
```
pfultz2, line 60: Can you just check `ROCBLAS_VERSION_MAJOR == 2`? Is it necessary to check the minor version? Also the logic seems backwards, shouldn't it be `ROCBLAS_VERSION_MINOR <= 38`?

pfultz2: Ok I see but we removed those checks in #1716 as its a really old version of rocblas. I think it would be better to check if `ROCBLAS_VERSION_MAJOR >= 3`.

umangyadav: current logic would would work for all cases. it would only call to `query_int8x4()` for versions from 2.38 and up to 3.0

pfultz2: > then I also need to take care of versions < 2.38 in addition.

No you dont. Its a really old version and we already removed checks for that old version.

pfultz2 (approved):

PR #1923, opened 2023-07-06
Parse dynamic ONNX `Shape` op and create `dimensions_of` MIGX operator

CharlieL7:
* Previously we only created literals for the ONNX `Shape` operator
* The shape will not be known until model runtime with dynamic shapes. I created the `dimensions_of` MIGX operator to return the dimensions of the input for dynamic shapes.


umangyadav (approved):
LGTM, Fix CI checks.

src/onnx/parse_shape.cpp
```diff
@@ -43,13 +45,48 @@ struct parse_shape : op_parser<parse_shape>
...
+        std::size_t start = 0;
+        std::size_t end = input_shape.ndim();
+        // Normalizing the start and end is handled here because of how the static shape version works
+        auto normalize_ind = [&](std::string attr_name){
+            auto attr = info.attributes.at(attr_name).i();
+            if(is_ind_out_of_bounds(attr))
+            {
+                MIGRAPHX_THROW("PARSE_SHAPE: " + attr_name + " attribute is out of bounds, value: " + std::to_string(attr));
+            }
+            return (attr >= 0) ? attr : input_ndim + attr;
```
pfultz2, line 60: It seems it would be simpler to compute the normalized index first and then check if its out of bounds(which is a simple `attr >= input_ndim`).

src/onnx/parse_shape.cpp
```diff
@@ -43,13 +45,48 @@ struct parse_shape : op_parser<parse_shape>
...
-            return int64_t(i);
-        });
-        return info.add_literal(migraphx::literal{s, vec_shape});
+        auto input_shape = args[0]->get_shape();
+        int input_ndim = input_shape.ndim();
+        auto is_ind_out_of_bounds = [&](auto ind){ return (ind > (input_ndim)) or (ind < (-1 * input_ndim)); };
+        std::size_t start = 0;
+        std::size_t end = input_shape.ndim();
+        // Normalizing the start and end is handled here because of how the static shape version works
+        auto normalize_ind = [&](std::string attr_name){
```
pfultz2, line 54: Instead of passing the string you could pass the index instead.

src/onnx/parse_shape.cpp
```diff
@@ -43,13 +45,48 @@ struct parse_shape : op_parser<parse_shape>
...
-        migraphx::shape s(migraphx::shape::int64_type, {arg_shape.size()});
-        std::transform(arg_shape.begin(), arg_shape.end(), vec_shape.begin(), [](auto i) {
-            return int64_t(i);
-        });
-        return info.add_literal(migraphx::literal{s, vec_shape});
+        auto input_shape = args[0]->get_shape();
+        int input_ndim = input_shape.ndim();
+        auto is_ind_out_of_bounds = [&](auto ind){ return (ind > (input_ndim)) or (ind < (-1 * input_ndim)); };
+        std::size_t start = 0;
+        std::size_t end = input_shape.ndim();
```
pfultz2, line 52: If you are going to use `input_shape.ndim()` directly then you should remove the `input_ndim` variable.

src/onnx/parse_shape.cpp
CharlieL7: That has an issue with something that's out of bounds with negative indices. Ex: there's 4 dimensions and somehow the -8 index is asked for. Would have to check bounds before

src/onnx/parse_shape.cpp
CharlieL7: Renamed to normalize_attr

TedThemistokleous (approved):

pfultz2 (approved):

PR #1930, opened 2023-07-07
Report test failures from CHECK macro

pfultz2:


test/include/test.hpp
```diff
@@ -683,10 +699,10 @@ inline void run(int argc, const char* argv[])
 #define TEST_CAPTURE(...) test::capture{}->*__VA_ARGS__
 
 // NOLINTNEXTLINE
-#define CHECK(...)                                                                                 \
-    test::failed(                                                                                  \
-        test::capture{}->*__VA_ARGS__, #__VA_ARGS__, __PRETTY_FUNCTION__, __FILE__, __LINE__, [] { \
-        })
+#define CHECK(...) \
+    test::failed(  \
+        TEST_CAPTURE(__VA_ARGS__), #__VA_ARGS__, __PRETTY_FUNCTION__, __FILE__, __LINE__, [] {})
```
umangyadav, line 704: I wonder what is the difference between `CHECK` and `EXPECT` ? When should one one but no the other one ?

pfultz2: `CHECK` causes the test case to fail but it will continue execution of the test case. `EXPECT` will stop executing the test case when it fails. This is similar to the [assertion severity levels](https://www.boost.org/doc/libs/1_67_0/libs/test/doc/html/boost_test/testing_tools/tools_assertion_severity_level.html) in Boost.Test, except we use `EXPECT` instead of `REQUIRE` and we dont have a `WARN` assertion.

umangyadav:
@pfultz2 CI is failing many checks.

umangyadav (approved):

CharlieL7 (approved):

PR #1932, opened 2023-07-07
Simplify nested reshapes

pfultz2:
The `find_reshaper` is supposed to do this, but it doesnt work and there were no tests. So I updated for it to work and I added unit tests for it.


src/simplify_reshapes.cpp
```diff
@@ -804,9 +789,9 @@ void simplify_reshapes::apply(module& m) const
         match::find_matches(m,
                             find_where_op{},
                             find_resize{},
-                            find_reshape_cont{},
                             find_nop_reshapes{},
                             find_reshaper{},
+                            find_reshape_cont{},
```
TedThemistokleous, line 794: Is there a reason why we've swapped order in these matches?

TedThemistokleous:
> The `find_reshaper` is supposed to do this, but it doesnt work and there were no tests. So I updated for it to work and I added unit tests for it.

Does this mean we should remove the reshaper pass or is there some other function its providing we need to keep these seperate?

TedThemistokleous (commented):
Code looks good. Curious on perf and some minor questions.

pfultz2:
> Does this mean we should remove the reshaper pass or is there some other function its providing we need to keep these seperate?

You mean the `find_reshaper`? I didnt remove it, I just changed it to work correctly.

TedThemistokleous:
> > Does this mean we should remove the reshaper pass or is there some other function its providing we need to keep these seperate?
> 
> You mean the `find_reshaper`? I didnt remove it, I just changed it to work correctly.

All good. Approved.

src/simplify_reshapes.cpp
```diff
@@ -89,38 +89,23 @@ struct find_reshaper
...
     auto matcher() const
     {
-        return match::name(reshaper_names())(
-            match::any_of[match::outputs()](match::name(reshaper_names())));
+        auto reshaper          = match::name(reshaper_names());
+        auto contiguous        = match::name("contiguous");
+        auto no_output_reshape = match::none_of[match::outputs()](reshaper);
+        auto input_reshape     = match::arg(0)(match::skip(contiguous)(reshaper));
+        auto input             = match::skip(reshaper, contiguous)(match::any().bind("x"));
+        return reshaper(no_output_reshape, input_reshape, input);
```
umangyadav, line 97: This would at max find triply nested reshapes. We can probably do as earlier to walk back the graph inside the `apply()` to find the `start` of the nested reshapes and then do the replace.

pfultz2: > This would at max find triply nested reshapes.

This will find an arbitrary number of nested reshapes, not just three. I updated the test to use 6 nested reshapes. 

> We can probably do as earlier to walk back the graph inside the apply() to find the start of the nested reshapes and then do the replace.

The `match::skip` already does this.

umangyadav (approved):

PR #1939, opened 2023-07-10
Add verify namespace for better symbol resolution

umangyadav:
`verify.hpp` has symbols like `max`, `sum`, `max_diff` etc. which could interfere with some other symbols in other files. 

For example `generate.hpp` and `verify.hpp` can not be included inside a same header file, it would throw warning about `max` being shadowed by another declaration. 

Putting `verify.hpp` inside another namespace would solve it.


TedThemistokleous (approved):

CharlieL7 (approved):
LGTM. Style question: what's out stance on using directives (`using namespace migraphx::verify`) and using declarations (`using migraphx::verify::verify_range`)?

umangyadav:
> LGTM. Style question: what's our stance on using directives (`using namespace migraphx::verify`) and using declarations (`using migraphx::verify::verify_range`)?

I think we can use the `using` directive as long as it doesn't create the same symbol resolution problem. I don't see `using` directive being used except for couple of places inside the tests though.

pfultz2 (approved):

pfultz2:
> LGTM. Style question: what's our stance on using directives (using namespace migraphx::verify) and using declarations (using migraphx::verify::verify_range)?

Not using fully qualified names will be looked up using ADL, which means it could use a different function. For example, if its a custom data type and it has a `verify_range` function in its namespace then the compiler will pick that over the one in `migraphx::verify` namespace. 

In general, `using` declarations should only be used for customization points(ie `std::swap`, `std::begin`, `std::end`, etc).

PR #1940, opened 2023-07-11
Fuse pointwise modules across reshapes

### LINKED ISSUE #1844: Fuse reshape into pointwise kernels (might be MLIR-only?)
I was looking through the compile trace on `distilgpt2_1_fp16_gpu.onnx` and I noticed that there were cases where there was a pointwise op, a `reshape`, and then a long string of pointwise operations, but that the reshape wasn't fused into the pointwise module.

Since, as far as I know, `reshape` is effectively a pointwise operation (it just affects your indexing in subsequent operations), it should be fused in to pointwise modules.

At the very least, I'm pretty sure MLIR fusions can handle this case.

pfultz2:
This should fix #1844.


src/common_dims.cpp
```diff
@@ -0,0 +1,130 @@
...
+            naxes++;
+            start--;
+        }
+        std::vector<std::size_t> axes(naxes);
+        std::iota(axes.begin(), axes.end(), start);
+        return axes;
+    }
+};
+
+static bool commpute_common_dim(std::vector<std::size_t>& cd_dims,
```
umangyadav, line 76:
```suggestion
static bool compute_common_dim(std::vector<std::size_t>& cd_dims,
```

src/common_dims.cpp
```diff
@@ -0,0 +1,130 @@
...
+    });
+    if(x < dim)
+        return start;
+    return it;
+}
+
+template <class Range>
+static auto elements(const Range& r)
+{
+    return std::accumulate(r.begin(), r.end(), std::size_t{1}, std::multiplies<>{});
```
umangyadav, line 26: if begin == end then, it would still return 1 as the number of elements.

src/common_dims.cpp
```diff
@@ -0,0 +1,130 @@
...
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+template <class Iterator>
+static auto compute_end_dim(Iterator start, Iterator last, std::size_t dim)
+{
+    std::size_t x = 1;
+    auto it       = std::find_if(start, last, [&](auto i) {
+        x *= i;
+        return x > dim;
```
umangyadav, line 16: Shouldn't it check for `x == dim` ?

src/common_dims.cpp
```diff
@@ -0,0 +1,130 @@
...
+                                common_dim_state& state2)
+{
+    assert(state1.get() <= state2.get());
+    auto d2    = state2.get();
+    auto dims  = state1.dims_for(d2);
+    auto n     = elements(dims);
+    auto naxes = distance(dims);
+    // If not divisible then we can't compute a common dim
+    if((d2 % n) != 0)
+        return true;
```
umangyadav, line 87: it is better to return `false` and return `true` if it finds a solution.

src/common_dims.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+static bool commpute_common_dim(std::vector<std::size_t>& cd_dims,
+                                common_dim_state& state1,
+                                common_dim_state& state2)
+{
+    assert(state1.get() <= state2.get());
+    auto d2    = state2.get();
+    auto dims  = state1.dims_for(d2);
+    auto n     = elements(dims);
+    auto naxes = distance(dims);
```
umangyadav, line 84: `naxes` could be zero if it has failed to find `compute_end_dim`.

test/common_dims.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+TEST_CASE(common1)
+{
+    auto cd = migraphx::common_dims::compute({2, 32, 2560}, {2, 1280, 8, 8});
+    EXPECT(cd.dims == std::vector<std::size_t>{2, 32, 40, 8, 8});
+    EXPECT(cd.axes_map1 == axes_map{{0}, {1}, {2, 3, 4}});
+    EXPECT(cd.axes_map2 == axes_map{{0}, {1, 2}, {3}, {4}});
+}
+
+TEST_CASE(common2)
```
umangyadav, line 22: Add tests for failure case.

src/include/migraphx/common_dims.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <migraphx/config.hpp>
+#include <cstdint>
+#include <vector>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+struct common_dims
+{
+    static common_dims compute(const std::vector<std::size_t>& dims1,
```
umangyadav, line 13: can we add some docstrings describing what it is doing and how it does it?

src/fuse_pointwise.cpp
```diff
@@ -189,6 +192,50 @@ static bool find_pointwise_modules(module& m)
     }
     return changed;
 }
+namespace {
+struct find_pointwise_reshape_pointwise
+{
+    auto matcher() const
+    {
+        auto reshape = match::name("reshape", "squeeze", "unsqueeze")(match::used_once());
```
umangyadav, line 200: I think we should be able to add `flatten` here as well.

test/fuse_pointwise.cpp
```diff
@@ -361,4 +361,43 @@ TEST_CASE(no_input)
...
+    migraphx::shape s3{migraphx::shape::float_type, {3, 10, 4, 2, 2}};
+    migraphx::program p1;
+    {
+        auto* mm  = p1.get_main_module();
+        auto x    = mm->add_parameter("x", s1);
+        auto y    = mm->add_parameter("y", s1);
+        auto z    = mm->add_parameter("z", s2);
+        auto add1 = mm->add_instruction(migraphx::make_op("add"), x, y);
+        auto reshape =
+            mm->add_instruction(migraphx::make_op("reshape", {{"dims", s2.lens()}}), add1);
```
umangyadav, line 377: Add test with non-standard squeeze+contiguous.

src/common_dims.cpp
pfultz2: No, it needs to be one past the end.

test/fuse_pointwise.cpp
pfultz2: Squeeze doesnt require contiguous.

test/fuse_pointwise.cpp
```diff
@@ -361,4 +361,75 @@ TEST_CASE(no_input)
...
+                return pm->add_instruction(migraphx::make_op("add"), add1, inputs[2]);
+            });
+        auto reshape =
+            mm->add_instruction(migraphx::make_op("reshape", {{"dims", s2.lens()}}), fadd);
+        mm->add_return({reshape});
+    }
+    EXPECT(p1.sort() == p2.sort());
+}
+
+TEST_CASE(add_reshape_add_error)
```
krzysz00, line 403: Why does this fail?

pfultz2: Because we cant compute a common dimensionality that preserves the axes. This can cause us to insert extra contiguous. For example if one of the inputs are broadcasted along the first axis of `s1`(ie a shape of `{6, 35}, {1, 0}`) and we reshape to `{3, 7, 2, 5}` then we will need to insert an extra contiguous as there is no way to set the strides to handle this case because `6` is being mapped to `3, 7` which is not the same number of elements.

In general, models do not do reshapes like this, so it shouldn't be a problem in practice.

krzysz00: Ah, makes sense. So the test, in some sense - at least how I'd describe it, is that you can only fuse when the reshape only
- Takes multiple contiguous dimensions from the input and merges them, or
- Takes one dimension from the input and splits it evenly

src/common_dims.cpp
```diff
@@ -0,0 +1,155 @@
...
+        cd_dims.push_back(state1.rem);
+    state1.next(distance(dims));
+    state2.next();
+    return true;
+}
+
+common_dims common_dims::compute(const std::vector<std::size_t>& dims1,
+                                 const std::vector<std::size_t>& dims2)
+{
+    assert(elements(dims1) == elements(dims2));
```
TedThemistokleous, line 131: Do we want to assert here that either   assert(elements(dims1) != 0) as well?

test/fuse_pointwise.cpp
umangyadav: > inputs are broadcasted along the first axis of s1(ie a shape of {6, 35}, {1, 0}) and we reshape to {3, 7, 2, 5} 

If input shapes are standard then it should be possible to reshape it to {3, 7, 2, 5} I think.  

That's what it is doing here : 
`auto reshape =
            mm->add_instruction(migraphx::make_op("reshape", {{"dims", s2.lens()}}), add1);`

umangyadav: >  and splits it evenly 

it could be uneven split (e.g. 35--> 5 * 7) from what i understand.

test/fuse_pointwise.cpp
umangyadav: Yeah, I mean Matcher would look for `squeeze, unsqueeze, reshape` or `flatten` op and skip match `contiguous` if there is any during the matching. I meant a test case like that. 

`Reshape` requires a standard shape right now.  

I am not sure if any pointwise would generate a non-standard shape but if we can have tests like the following : 
`pointwise ---> (non-standard output) --> contiguous --> "reshaper" op --> pointwise`. 

and see if it gets matched and handled well by the logic. 

In this case, inputs to the first pointwise could be non-standard shaped and the second pointwise should be able to handle non-standard shape as well if contiguous is not present. 

For example : 

`pointwise -->  (non standard output) --> squeeze/unsqueeze --> pointwise. 

Squeeze/unsqueeze can handle non-standard shapes.

src/fuse_pointwise.cpp
```diff
@@ -189,6 +192,53 @@ static bool find_pointwise_modules(module& m)
...
+        auto x_ins       = r.instructions["x"];
+        auto reshape_ins = r.instructions["reshape"];
+
+        auto cd = common_dims::compute(ins->get_shape().lens(), x_ins->get_shape().lens());
+        if(cd.dims.empty())
+            return;
+
+        auto reshape_input = [&](const auto& ins_to_insert) {
+            return [&](auto input) {
+                auto c = m.insert_instruction(ins_to_insert, make_op("contiguous"), input);
```
umangyadav, line 221: Would this be better performance-wise if we have to add `contiguous` for all the inputs? 

`eliminate_contigous` would eliminate some of them but if the inputs are non-standard shaped for the first pointwise module, it could add additional overhead. 

Perhaps we should just limit this fusion to the standard-shaped inputs.

src/fuse_pointwise.cpp
```diff
@@ -189,6 +192,53 @@ static bool find_pointwise_modules(module& m)
...
+            m.insert_instruction(x_ins, x_ins->get_operator(), x_inputs, x_ins->module_inputs());
+
+        auto inputs = ins->inputs();
+        std::transform(inputs.begin(), inputs.end(), inputs.begin(), [&](auto input) {
+            if(input == reshape_ins)
+                return new_x_ins;
+            return reshape_input(ins)(input);
+        });
+        auto pw = m.insert_instruction(ins, ins->get_operator(), inputs, ins->module_inputs());
+        m.replace_instruction(ins, make_op("reshape", {{"dims", ins->get_shape().lens()}}), pw);
```
umangyadav, line 238: Output of `pw` must be standard for the `reshape` to work.  Need to add `contiguous` before reshape if necessary.

umangyadav (changes requested):

src/common_dims.cpp
```diff
@@ -0,0 +1,155 @@
...
+    return it;
+}
+
+template <class Range>
+static auto elements(const Range& r)
+{
+    return std::accumulate(r.begin(), r.end(), std::size_t{1}, std::multiplies<>{});
+}
+
+struct common_dim_state
```
umangyadav, line 52: need to export symbols for windows

src/common_dims.cpp
umangyadav: > begin == end then 

`elements()` would never be called with begin==end. need to assert that.

src/common_dims.cpp
```diff
@@ -0,0 +1,155 @@
...
+    auto n     = elements(dims);
+    auto naxes = distance(dims);
+    if(naxes == 0)
+        return false;
+    // If not divisible then we can't compute a common dim
+    if((d2 % n) != 0)
+        return false;
+    auto rem = d2 / n;
+    state1.add_multi_axes(naxes, cd_dims.size());
+    state2.add_axes(rem == 1 ? naxes : naxes + 1, cd_dims.size());
```
umangyadav, line 115: do we have a test which would have `rem != 1` ?

test/fuse_pointwise.cpp
pfultz2: > Yeah, I mean Matcher would look for squeeze, unsqueeze, reshape or flatten op and skip match contiguous if there is any during the matching. I meant a test case like that.

Test case like what?

> Reshape requires a standard shape right now.

No it doesn't, but it doesnt work for all layouts, which is why we insert a contiguous. eliminate_contiguous can figure out if the contiguous is needed.

test/fuse_pointwise.cpp
pfultz2: > If input shapes are standard then it should be possible to reshape it to {3, 7, 2, 5} I think.

Yes of course, but we wont know if the shapes will be standard or not until after eliminate_contiguous. Doing it prematurely will introduce more contiguous operators.

> it could be uneven split (e.g. 35--> 5 * 7) from what i understand.

Reshape doesnt handle such reshapes with non-standard shape. Furthermore, the calculation of the axes in common_dims would be a lot more complicated.

src/fuse_pointwise.cpp
pfultz2: > Would this be better performance-wise if we have to add contiguous for all the inputs?

Thats why I use the common_dims to prevent the need to add contiguous. 

> Perhaps we should just limit this fusion to the standard-shaped inputs.

We dont know if the inputs will be standard shape until after `eliminate_contiguous`.

src/fuse_pointwise.cpp
pfultz2: > Output of pw must be standard for the reshape to work.

Not always.

>  Need to add contiguous before reshape if necessary.

Contiguous was already inserted on the input reshapes so the output should be standard shape.

src/common_dims.cpp
pfultz2: Yes, `common1` test case: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1940/files#diff-3c971779822603718694fa5365cc5fc99270555c769e1050b9e1338c642dce5eR39

src/common_dims.cpp
```diff
@@ -0,0 +1,155 @@
...
+    auto d2    = state2.get();
+    auto dims  = state1.dims_for(d2);
+    auto n     = elements(dims);
+    auto naxes = distance(dims);
+    if(naxes == 0)
+        return false;
+    // If not divisible then we can't compute a common dim
+    if((d2 % n) != 0)
+        return false;
+    auto rem = d2 / n;
```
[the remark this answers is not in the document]
pfultz2: In test case common1, `d2==1280` and `n==32`, so `d2 % n == 0` and `d2 / n` is equal to 40. I am not sure how you conclude it will always be 1.

test/fuse_pointwise.cpp
umangyadav: > Test case like what?

`pointwise --> (non standard output) --> squeeze/unsqueeze --> pointwise.

src/fuse_pointwise.cpp
pfultz2: By adding the contiguous afterwards can help improve optimizations with eliminate_contiguous. If the last reshape requires a contiguous then its better to do the contiguous on the output instead of on every input.

pfultz2: It doesnt really matter since simplify_reshapes will remove the redundant contiguous.

umangyadav (approved):

src/common_dims.cpp
umangyadav: it is unlikely that this get called with `begin== end` but if you could add an assert for this, it would be good.

krzysz00 (commented):
I don't see any issues here, but I'm also not a qualified reviewer on this

krzysz00:
@pfultz2 If you need me as an approver to get this in, I can do that

PR #1949, opened 2023-07-12
Stamp driver to read the SHA-1 of .onnxrt-commit

TedThemistokleous:
Useful for CI purposes as well as keeping track in our .deb creation which version of Onnxruntime MIGraphX has had CI tested against it.


pfultz2:
I dont think it makes sense to have an onnxrt commit hash in the driver as we don't use or depend on onnxrt. If we want to know what version of onnxrt we tested in our CI then we can get that from migraphxs commit hash since we store our CI scripts in the repo itself.

src/CMakeLists.txt
```diff
@@ -28,6 +28,13 @@ include(ROCMInstallTargets)
 include(ROCMPackageConfigHelpers)
 include(RegisterOp)
 include(CheckCXXLinkerFlag)
+ 
+function(ReadSHA InputFile VAR)
+    file(STRINGS "${InputFile}" String_output)
+    add_compile_definitions(${VAR}="${String_output}")
+endfunction()
+
+ReadSHA("../test/onnx/.onnxrt-commit" MIGRAPHX_ORT_SHA1)
```
umangyadav, line 37:
```suggestion
ReadSHA("${CMAKE_SOURCE_DIR}/test/onnx/.onnxrt-commit" MIGRAPHX_ORT_SHA1)
```

pfultz2:
> This will also happen in the wild if someone just does an apt install migraphx; they won't know what version of onnxruntime we've tested with to use without having to do a lookup in our repo to manually read .onnxrt-commit.

First, I think users will want to use released versions of onnxrt not a development commit. Either way, migraphx should in general work with any commit of onnxrt. Secondly, the CI to test with onnxrt is to test migraphx to make sure we dont introduce an issue downstream, its not to test onnxrt to make sure its a working version. Ideally, there should be CI in onnxrt to test a released version of migraphx on every commit, but until then we will need to rely on QA for that.

TedThemistokleous:
> > This will also happen in the wild if someone just does an apt install migraphx; they won't know what version of onnxruntime we've tested with to use without having to do a lookup in our repo to manually read .onnxrt-commit.
> 
> First, I think users will want to use released versions of onnxrt not a development commit. Either way, migraphx should in general work with any commit of onnxrt. Secondly, the CI to test with onnxrt is to test migraphx to make sure we dont introduce an issue downstream, its not to test onnxrt to make sure its a working version. Ideally, there should be CI in onnxrt to test a released version of migraphx on every commit, but until then we will need to rely on QA for that.
> 
> > Right now we don't have a way to stripe our deb package with the proper information such as SHA-1 since that gets stripped out. If we add the onnxrt-sha to the driver, we can just call migraphx-driver ort-sha
> 
> If our release version is stripping the commit hash from the version, we can still add the commit hash to migraphx back in, it just wont be added as part of the version. Although, it looks like we dont strip the commit hash in rocm-cmake so it should be in the version.h on the release: https://github.com/RadeonOpenCompute/rocm-cmake/blob/develop/share/rocm/cmake/ROCMSetupVersion.cmake#L145.

Wouldn't this be simpler to just integrate at one place though? This lets us determine what version of ORT we've tested against for tests/automation by just looking at the built/installed driver? Having to continually cross reference rocm->migraphx-sha-ort_sha1 seems tedious here.

umangyadav (approved):

causten (approved):

src/CMakeLists.txt
```diff
@@ -28,6 +28,13 @@ include(ROCMInstallTargets)
 include(ROCMPackageConfigHelpers)
 include(RegisterOp)
 include(CheckCXXLinkerFlag)
+ 
+function(ReadSHA InputFile VAR)
+    file(STRINGS "${InputFile}" String_output)
+    add_compile_definitions(${VAR}="${String_output}")
```
pfultz2, line 34: Use target_compile_definitions instead. This should only be added to the driver.

src/CMakeLists.txt
```diff
@@ -28,6 +28,13 @@ include(ROCMInstallTargets)
...
+endfunction()
```
pfultz2, line 35: I don't think this needs to be a separate function, as its only used in one place and is only two lines.

src/driver/main.cpp
```diff
@@ -578,6 +578,12 @@ struct verify : command<verify>
     }
 };
 
+struct ort_sha : command<ort_sha>
+{
+    void parse(const argument_parser&) {}
+    void run() const { std::cout << MIGRAPHX_ORT_SHA1 << std::endl; }
```
pfultz2, line 584: This should be moved to the main function. As ort_sha shouldn't show up in the help.

pfultz2 (changes requested):

src/driver/main.cpp
TedThemistokleous: Done, removed

src/CMakeLists.txt
TedThemistokleous: Moved to below and removed the function wrapping

src/CMakeLists.txt
TedThemistokleous: Done

src/driver/main.cpp
```diff
@@ -789,6 +790,13 @@ int main(int argc, const char* argv[])
 
     auto&& m = get_commands();
     auto cmd = args.front();
+
+    if(std::experimental::string_view(argv[1]) == "ort-sha")
```
umangyadav, line 794: Instead of `experimental` i think we may want do similar to `filesystem.hpp` where we check if CPP version is >17 or not and accordingly use experimental or non-experimental headers

src/CMakeLists.txt
```diff
@@ -304,6 +305,9 @@ if(BUILD_DEV)
     target_compile_definitions(migraphx PUBLIC -DBUILD_DEV)
 endif()
 
+file(STRINGS "${CMAKE_SOURCE_DIR}/test/onnx/.onnxrt-commit" String_output)
+target_compile_definitions(migraphx PUBLIC MIGRAPHX_ORT_SHA1="${String_output}")
```
pfultz2, line 309: This should be added to the driver not to migraphx.

src/driver/main.cpp
pfultz2: Yea dont use experimental, just use `cmd`.

src/CMakeLists.txt
pfultz2, line 309: This should be added to the driver not to migraphx.

pfultz2 (changes requested):

src/CMakeLists.txt
TedThemistokleous: Done and Done

src/CMakeLists.txt
```diff
@@ -304,6 +305,9 @@ if(BUILD_DEV)
     target_compile_definitions(migraphx PUBLIC -DBUILD_DEV)
 endif()
 
+file(STRINGS "${CMAKE_SOURCE_DIR}/test/onnx/.onnxrt-commit" String_output)
+target_compile_definitions(driver PUBLIC MIGRAPHX_ORT_SHA1="${String_output}")
```
pfultz2, line 309: This should go into driver/CMakeLists.txt.

TedThemistokleous: Done

pfultz2 (approved):

PR #1955, opened 2023-07-13
Improve performance of pointwise/reduction kernels when using NHWC layouts

### LINKED ISSUE #1944: Improve performance of pointwise/reduction kernels when using NHWC layouts

pfultz2:
Fixes #1944


test/verify/test_add_nhwc.cpp
```diff
@@ -0,0 +1,44 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+
+struct test_add_nhwc : verify_program<test_add_nhwc>
```
umangyadav, line 30: add test for reduce as well ?

src/include/migraphx/permutation.hpp
```diff
@@ -66,6 +66,10 @@ MIGRAPHX_EXPORT std::vector<int64_t> invert_permutation(const std::vector<int64_
 MIGRAPHX_EXPORT std::vector<int64_t> find_permutation(const shape& s);
 MIGRAPHX_EXPORT std::vector<int64_t> find_permutation(const std::vector<shape>& shapes);
 
+/// Normalize the shapes so the order of dimensions will be in the order it is
+/// in memory as much as possible.
```
umangyadav, line 70: how does it help with performance? it would still be permutated and strided in memory.

pfultz2: >  it would still be permutated and strided in memory.

No it wont since the dimensions are now in the same order as they are in memory. Say we have a transposed shape like this: `{4, 3, 8, 8}, {192, 1, 24, 3}` which has a permutation of `{0, 2, 3, 1}`(ie NHWC). After `normalize_permutation` it will be `{4, 8, 8, 3}, {192, 24, 3, 1}` which is standard shape. So the order of elements in the shape is the same as the order of elements in memory so there wont be any strided reads or writes.

test/verify/test_add_nhwc.cpp
kahmed10: @pfultz2, I'll approve once this is added

pfultz2: Test added.

src/include/migraphx/permutation.hpp
umangyadav: I think this would also change input's memory layout if it is broadcasted.

kahmed10 (approved):

src/include/migraphx/permutation.hpp
pfultz2: > If let's say, it is a pointwise binary operator and one of the input is transposed but the other one is not.

For mixed layouts, it may not still not be optimal(where usually we would need to copy to LDS to do the transpose), but it will still be better then before.

It takes in the output in account as well. So if the output is also transposed so only one tensor will be used as transposed instead of two transposed tensors. Or if one input is NHWC and another CHWN, we can have at least one input be non-transposed rather then neither.

pfultz2: > I think this would also change input's memory layout if it is broadcasted.

It never changes the way its layout'd in memory. Also `find_permutation` skips using broadcasted shapes since those can't be inferred accurately due to the 0 strides.

umangyadav (approved):

src/include/migraphx/permutation.hpp
umangyadav: >  Also find_permutation skips using broadcasted shapes since those can't be inferred accurately due to the 0 strides. 

find_permutation would skip it but `normalize_permutation` would still transform it.

PR #1961, opened 2023-07-14
[MLIR] Changes for tuning API v2, MLIR grid layout changes.

krzysz00:
This PR constitutes the MIGraphX-side changes needed to not break the build in the presence of
https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1136 , and updates what data is sent in to MLIR during the kernel generation and tuning process.

Specifically, as of
https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1145 (which hasn't landed yet but is likely to) we will want to know the number of CUs on the card we're targeting in order to efficiently tile our matrix multiplications. This commit therefore updates the tsv parser (since we'll be adding a new field) and updates the compilation process to pass num_cu into MLIR.

This commit also removes a legacy "xdlopsv2" attribute that is no longer used as we have a more robust architecture feature lookup in MLIR and no longer require it.

We also get rid of a rather annoying "failed to set param" notice when there's no tuning DB loaded so we don't scare everyone with those.


src/targets/gpu/mlir.cpp
```diff
@@ -636,7 +628,11 @@ struct mlir_program
         return op;
     }
 
-    void find_target() { target_arch = get_device_name(); }
+    void get_gpu_properties()
+    {
+        target_arch = get_device_name();
+        num_cu      = get_cu_count();
```
pfultz2, line 634: This should use `ctx.get_current_device().get_cu_count()`. Ideally we should do the same for device name.

src/targets/gpu/mlir.cpp
```diff
@@ -636,7 +628,11 @@ struct mlir_program
         return op;
     }
 
-    void find_target() { target_arch = get_device_name(); }
+    void get_gpu_properties()
```
pfultz2, line 631: This should be `set_gpu_properties`(as 'get' implies its returning a value). This also needs to take the `context` as a parameter.

krzysz00: ... then why didn't it take a context before?

But yeah, `set_target_properties()` is a better name here, agreed.

pfultz2 (approved):

kahmed10 (approved):

kahmed10:
@krzysz00 can you check the failures on the MLIR debug pass?

src/targets/gpu/mlir.cpp
```diff
@@ -693,12 +687,25 @@ struct mlir_program
...
-        mlir_tuning_table tuning_table{mlirRockTuningTableCreate()};
-        tc.problem = std::string{mlirRockTuningGetKey(tuning_table.get(), mmodule.get())};
+        std::string tuningKey;
+        tuningKey.resize(ROCMLIR_TUNING_KEY_BUFSZ);
+        size_t tuningKeyBytes =
+            mlirRockTuningGetKey(mmodule.get(), tuningKey.data(), tuningKey.length());
+        if(tuningKeyBytes > tuningKey.length())
+            MIGRAPHX_THROW("Tuning table key was " + std::to_string(tuningKeyBytes) +
+                           " bytes and thus too long");
+        tc.problem = std::move(tuningKey);
```
pfultz2, line 708: The string needs to be trimmed based on the size of the key.

src/targets/gpu/mlir.cpp
```diff
@@ -677,12 +674,9 @@ struct mlir_program
 
     void set_tuning(const value& v)
     {
-        auto str = v.to<std::string>();
-        // We need to make a copy of the buffer since mlirRockTuningSetFromStr may modify the string
-        std::vector<char> buffer(str.begin(), str.end());
-        buffer.push_back(0);
-        if(not mlirRockTuningSetFromStr(mmodule.get(), buffer.data()))
-            MIGRAPHX_THROW("Failed setting tuning key: " + str);
+        auto str = v.to<std::string_view>();
```
pfultz2, line 677: `string_view` wont work if `v` is not a string, and it may crash in a non-obvious way. Instead use `if_string` method and throw when its null:

```cpp
auto* str = v.if_string();
if(not str)
    MIGRAPHX_THROW("Only string is valid.");
if(not mlirRockTuningSetFromStr(mmodule.get(), make_mlir_string_ref(*str)))
    ...
```

src/targets/gpu/mlir.cpp
```diff
@@ -693,12 +687,25 @@ struct mlir_program
...
+                mlirRockTuningParamToString(param.get(), perfKey.data(), perfKey.length());
+            if(perfKeyBytes > perfKey.length())
+                MIGRAPHX_THROW("Tuning perf key was " + std::to_string(perfKeyBytes) +
+                               " bytes and thus too long");
+            tc.solutions.push_back(std::move(perfKey));
         }
-        mlir_tuning_table tuning_table{mlirRockTuningTableCreate()};
-        tc.problem = std::string{mlirRockTuningGetKey(tuning_table.get(), mmodule.get())};
+        std::string tuningKey;
+        tuningKey.resize(ROCMLIR_TUNING_KEY_BUFSZ);
```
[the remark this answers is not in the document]
krzysz00: Note: no trailing null in returned size

krzysz00:
Enum for tuning space

src/targets/gpu/mlir.cpp
```diff
@@ -677,36 +674,47 @@ struct mlir_program
...
+        tc.problem = std::string(tuning_key.begin(), tuning_key.begin() + tuning_key_bytes);
         return tc;
     }
 
     std::string get_tune_params(bool xdlops) const { return get_mlir_perf_for_conv(pp, xdlops); }
 
     // This function appends to tuning cfg file that could be
     // used with rocMLIR tuning scripts.
-    void dump_tuning_cfg(const char* prob_config) const
+    void dump_tuning_cfg(std::string& prob_config) const
```
pfultz2, line 717: This should be const.

krzysz00: Done, thanks!

krzysz00:
@pfultz2 Is there anything else?

pfultz2 (approved):

pfultz2:
@krzysz00 Need to fix cppcheck warnings.

PR #1963, opened 2023-07-14
Set default installation path to /opt/rocm

pfultz2:


umangyadav (approved):

CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ else()
     endif()
 endif()
 
+set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "")
```
umangyadav, line 46: Would this remain same on windows ?

umangyadav (changes requested):

CMakeLists.txt
umangyadav: Else we can check if it is compiling for windows and set a different default path

pfultz2: Users could use that path on windows, but I assume they most likely would set it because there is no standard install location.

pfultz2: Previously, we set the path to `/opt/rocm/x86_64-w64-mingw32` on windows, which is mainly for cross-compiling with mingw.

PR #1964, opened 2023-07-14
Cancel in progress performance runs if a new commit has been pushed. …

### LINKED ISSUE #1962: Cancel previous performance test runs upon new commits push
Performance tests seem to be kicked off each time a new commit is pushed to PR. it is causing the backlog. It would be better to cancel previous runs if a new commit is pushed to the PR.

umangyadav:
Fixes #1962 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

Unlike `ci.yaml` which uses "cancel job", Performance tests are handled by the "concurrency" mechanism to cancel any pending jobs.  When a new commit is pushed, it would only cancel "pending" jobs but not the one that is already running. I don't see value in keep running performance tests on older commits when new commit is available.


IgorMirosavljevicHTEC:
Triggering workflow on closed PR is only cleaning up data which was related to that PR. Last valid performance result is moved to pool of results used in performance calculations/historical overview and to backup repo. Thats why it shouldn't be removed.

In case of canceling in progress runs, cleanup should be executed for that specific run only, comment creation skipped etc.

umangyadav:
> Triggering workflow on closed PR is only cleaning up data which was related to that PR. Last valid performance result is moved to pool of results used in performance calculations/historical overview and to backup repo. Thats why it shouldn't be removed.

Added back "closed" type for the perf runs. 

> In case of canceling in progress runs, cleanup should be executed for that specific run only, comment creation skipped etc.

Clean-up would be performed by workflow dispatch on the closing of the PR

https://github.com/ROCmSoftwarePlatform/migraphx-benchmark/blob/bebc05a5541976f4f99984e7c69c4a0fb688d247/.github/workflows/perf-test.yml#L249

https://github.com/ROCmSoftwarePlatform/migraphx-benchmark/blob/bebc05a5541976f4f99984e7c69c4a0fb688d247/.github/workflows/perf-test.yml#L258

Even if an in-progress generated comment on the PR before the cancel, that comment would get edited by a newer run when it finishes. 

> Last valid performance result is moved to pool of results used in performance calculations/historical overview and to backup repo

The latest commit would generate the "last valid" result. So i think it should be fine to cancel the in-progress run.

kahmed10 (approved):

causten (approved):

PR #1965, opened 2023-07-15
Enable tuning for MLIR

pfultz2:


krzysz00 (changes requested):
I'm going to request this hold for #1961 , and, as a high level note, point out that failures (by way of `mlirLogicalResultFailed` in the high-level or backend pipelines, are proper errors. On the other hand, just because we gave you a tuning key doesn't mean it'll compile - you need to run the applicability pipeline to check.

pfultz2:
> I'm going to request this hold for https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1961 ,

That seems orthogonal to what I am doing here.

> On the other hand, just because we gave you a tuning key doesn't mean it'll compile - you need to run the applicability pipeline to check.

But if its not applicable it will fail compilation, so I dont see the problem.

krzysz00:
The point is that if something fails compilation this is either because you've got a bad tuning parameter or because there's an actual bug ... or code that wouldn't work for any tuning parameter (usually in the case of the high-level pipeline).

It's possible that the applicability pipeline is an MIOpen-ism that isn't needed, but it still seems helpful to distinguish between "bad tuning parameter" and "something went wrong"

src/targets/gpu/include/migraphx/gpu/mlir.hpp
```diff
@@ -29,23 +29,26 @@
...
-                                                const std::vector<instruction_ref>& inputs);
+                                                const std::vector<instruction_ref>& inputs,
+                                                const value& solution);
 
 MIGRAPHX_GPU_EXPORT instruction_ref insert_mlir(module& m,
                                                 instruction_ref ins,
                                                 code_object_op co,
                                                 const std::vector<instruction_ref>& inputs);
 
+tuning_config get_tuning_config_mlir(module m, const std::vector<shape>& inputs);
```
umangyadav, line 50:
```suggestion
MIGRAPHX_GPU_EXPORT tuning_config get_tuning_config_mlir(module m, const std::vector<shape>& inputs);
```

umangyadav (approved):

PR #1971, opened 2023-07-18
Update cmake to use latest and not pinned to 3.24.3

TedThemistokleous:
Need this since onnxruntime is rolling a version with current versions 3.26.*

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,7 +97,8 @@ RUN /download_models.sh && rm /download_models.sh
 # Install latest ccache version
 RUN cget -p $PREFIX install facebook/zstd@v1.4.5 -X subdir -DCMAKE_DIR=build/cmake
 RUN cget -p $PREFIX install ccache@v4.1 -DENABLE_TESTING=OFF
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -97,7 +97,8 @@ RUN /download_models.sh && rm /download_models.sh
 # Install latest ccache version
 RUN cget -p $PREFIX install facebook/zstd@v1.4.5 -X subdir -DCMAKE_DIR=build/cmake
 RUN cget -p $PREFIX install ccache@v4.1 -DENABLE_TESTING=OFF
-RUN cget -p /opt/cmake install kitware/cmake@v3.24.3
+
+RUN pip3 install cmake
```
pfultz2, line 101: The version should be pinned, so the builds are reproducible. And it should be installed in a non-standard path(like /opt/cmake) so the migraphx builds will not use this newer cmake version.

TedThemistokleous: Fair okay. Ill pin this to 3.26.2 then

umangyadav (approved):

causten (approved):

Issue #1746, opened 2023-05-08
ResNet50 accuracy


attila-dusnoki-htec:
There is a accuracy issue regarding ResNet50.

MIGraphX version: 2bebf64dc993b809801cf9e2d93c62ff5b54b991

Model link: https://zenodo.org/record/2592612/files/resnet50_v1.onnx

Dataset: https://www.image-net.org/download.php

The expected accuracy should be `76.46%`.

With FP32:
```bash
accuracy=76.456%, good=38228, total=50000
```

With FP16:
```bash
accuracy=75.456%, good=37728, total=50000
```

Difference file:  [resnet_fp16_fp32_diff.log](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/11421457/resnet_fp16_fp32_diff.log)

Left column: fp16 | Right column: fp32
The `*` at the end means it is different from the expected.

Snippet:

```bash
# fp16 incorrect, fp32 correct
ILSVRC2012_val_00038032.JPEG, expected: 750, found 36 *	      |	ILSVRC2012_val_00038032.JPEG, expected: 750, found 750 
# fp16 correct, fp32 incorrect
ILSVRC2012_val_00001982.JPEG, expected: 650, found 650 	      |	ILSVRC2012_val_00001982.JPEG, expected: 650, found 546 *
# fp16 incorrect, fp32 incorrect, but both different
ILSVRC2012_val_00003249.JPEG, expected: 502, found 463 *      |	ILSVRC2012_val_00003249.JPEG, expected: 502, found 104 *

# Mapping
36: 'terrapin'
750: 'quilt, comforter, comfort, puff'

650: 'microphone, mike'
546: 'electric guitar'

463: 'bucket, pail'
104: 'wallaby, brush kangaroo'
502: 'clog, geta, patten, sabot'
```

Note: the images require [preprocessing](https://github.com/mlcommons/inference/blob/master/vision/classification_and_detection/python/dataset.py#L168-L184).

These are only the labels, the scores are not used in the scripts. I realize that those numbers would also be important, so I can update this issue with that info later if needed.

The expected accuracy would be better: `Max 99% of 76.46% (75,6954%)`.

kahmed10:
Can we see the difference in scores? Maybe top5 results. Especially where fp16 is incorrect but fp32 is correct.

gyulaz-htec:
@kahmed10 I've collected the top5 scores and extended the previous diff file:
[...]

PR #1974, opened 2023-07-18
Use `optimize_module` pass for the quantization to fp16

umangyadav:
Fixes https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1746

BatchNorm only has `x` as the runtime input parameter for the following equation. All the other parameters are compile-time constants and related operations can be const-folded before quantizing to fp16 to preserve precision. 

`1/sqrt((var+epsilon)) * gamma` can be const folded. 

Adding `optimize_module()` pass before `quantize_fp16()` would do the simplification and const-folding. 

image

This PR along with #1969 produces FP16 ResNet50 accuracy same as Fp32 model ~76.45%

### DIFF
--- a/src/onnx/parse_batchnorm.cpp
+++ b/src/onnx/parse_batchnorm.cpp
@@ -57,12 +57,11 @@ struct parse_batchnorm : op_parser<parse_batchnorm>
         auto x_rank = x_lens.size();
         if(x_rank == 1 or x_rank == 2)
         {
-            auto rt      = info.add_literal(migraphx::literal{migraphx::shape{x_type}, {0.5}});
             auto eps     = info.add_literal(migraphx::literal{migraphx::shape{x_type}, {epsilon}});
             auto numer   = info.add_broadcastable_binary_op("sub", args[0], args[3]);
             auto var_eps = info.add_broadcastable_binary_op("add", args[4], eps);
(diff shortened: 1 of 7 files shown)


src/onnx/parse_batchnorm.cpp
kahmed10, line 1: If optimize_module is all that's needed, why do the parsers need to be modified?

umangyadav: parser changes the order of computation. 

It calculuates `1/sqrt((var+epsilon)) * gamma` first.

src/pass_manager.cpp
```diff
@@ -178,6 +178,8 @@ void run_passes(module& mod, const std::vector<pass>& passes, tracer trace)
 
 void run_passes(program& prog, const std::vector<pass>& passes, tracer trace)
 {
+    if(enabled(MIGRAPHX_TRACE_PASSES{}))
```
kahmed10, line 181: Is this needed for the PR?

umangyadav: not needed actually. I can remove it.

kahmed10 (approved):

turneram (approved):

shivadbhavsar (approved):

PR #1976, opened 2023-07-19
Make global workitems multiple of local workitems 

### LINKED ISSUE #1644: Remove hipRTC workarounds
MIGraphX had put in place some workarounds because of the limitations of hipRTC in 5.4.x. 

Problems in hipRTC have been fixed in 5.6 mainline.

### LINKED ISSUE #1977: ROIAlign tests are failing on  CI
Failures seen on Vega10 and Gfx90a'
Passing on Vega20 
Flaky test. 
Passes on Gfx90a : http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FAMDMIGraphX/detail/batchnorm_fp16/4/pipeline/125

umangyadav:
HIP requires global work items in multiple of local work items. If it is not it is not guaranteed to generate correct results all the time. 
Fixes #1977 
Fixes #1644 
MIGraphX CI has moved to rocm-5.6 which doesn't require hipRTC workarounds. 

Added `set +x` in github actions to avoid failing when it can't find cache. for example this : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/actions/runs/5599331698/jobs/10242149431

### DIFF
--- a/src/targets/gpu/compile_hip_code_object.cpp
+++ b/src/targets/gpu/compile_hip_code_object.cpp
@@ -135,14 +135,11 @@ compute_global_for(context& ctx, std::size_t n, std::size_t over)
     std::size_t max_global = ctx.get_current_device().get_cu_count() *
                              ctx.get_current_device().get_max_workitems_per_cu();
     return [n, over, max_global](std::size_t local) {
-        std::size_t num_elements = n;
+        // hip require global workitems multiple of local workitems
+        std::size_t num_elements = ((n + local - 1) / local) * local;
         std::size_t groups       = (num_elements + local - 1) / local;
         std::size_t max_blocks   = max_global / local;
         std::size_t nglobal      = std::min(max_blocks * over, groups) * local;
-#ifdef MIGRAPHX_USE_HIPRTC
-        if(enabled(MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS{}))
-            num_elements = ((num_elements + local - 1) / local) * local;
-#endif
         return std::min(nglobal, num_elements);
     };
 }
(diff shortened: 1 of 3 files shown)


kahmed10 (approved):

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -140,8 +140,8 @@ compute_global_for(context& ctx, std::size_t n, std::size_t over)
         std::size_t max_blocks   = max_global / local;
         std::size_t nglobal      = std::min(max_blocks * over, groups) * local;
 #ifdef MIGRAPHX_USE_HIPRTC
-        if(enabled(MIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS{}))
-            num_elements = ((num_elements + local - 1) / local) * local;
+        // hip compiler assumes global workitems is always multiple of local workitems 
+        num_elements = ((num_elements + local - 1) / local) * local;
```
pfultz2, line 144: This can affect performance, we should only do this with the workarounds enabled. This should be reported as a bug to hiprtc.

umangyadav: It was reported as a bug and they came back saying it assumes and requires global workitems multiple of local.

You can check SWDEV-398603.

umangyadav: If i dump assembly with hipclang it has this metadata:
 .uniform_work_group_size: 1

hip runtime would start throwing errors if it is not uniform workgroup size. It does show any errors right now, apparently there another bug SWDEV-403995

umangyadav: https://reviews.llvm.org/D155213

HIP would allow non-uniform WG sizes in future but for now by default it is not allowed.

causten:
Should this go in to 5.6.1 too?

TedThemistokleous (approved):

umangyadav:
> Should this go in to 5.6.1 too?

I wouldn't push for it.  5.6 still uses hipclang as default compiler. we have only seen failures with hiprtc compilation. There could be differences between compilation pipeline between hiprtc and hipclang and that's why things are working with hipclang but not with hiprtc when using non-uniform workgroup size.

PR #1990, opened 2023-07-21
export API symbols from MIGraphX C dynamic library

apwojcik:
Export API symbols from MIGraphX C dynamic library. API symbol exporting affects only Windows. It is transparent on Linux.

### DIFF
--- a/src/api/include/migraphx/migraphx.h
+++ b/src/api/include/migraphx/migraphx.h
@@ -26,6 +26,9 @@
 
 #include <stdlib.h>
 #include <stdbool.h>
+
+#include <migraphx/api/export.h>
+
 // Add new types here
 // clang-format off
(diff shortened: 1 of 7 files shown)


umangyadav (approved):

src/api/include/migraphx/migraphx.h
```diff
@@ -26,6 +26,9 @@
 
 #include <stdlib.h>
 #include <stdbool.h>
+
+#include <migraphx/c/export.h>
```
pfultz2, line 30: I think I would prefer the header to be under `migraphx/api/export.h`.

apwojcik: Done

pfultz2 (approved):

PR #1991, opened 2023-07-21
cmake: use IMPORTED target for the threads library

apwojcik:


umangyadav (approved):

CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ project(migraphx LANGUAGES C CXX)
 include(CTest)
 
 find_package(ROCM REQUIRED)
+find_package(Threads REQUIRED)
```
umangyadav, line 54: why was this moved from src/Cmakelists.txt to Toplevel cmakelists ?

apwojcik: The imported target `Threads::Threads` is used in `src/CMakeLists.txt` and `test/CMakeLists.txt`. In the coming PR, I will submit a change that adds the dependency to `src/targets/ref/CMakeList.txt`. It is enough to find the package at the top level instead of calling `find_package()` three times; even if the `find_package()` returns just after realizing the Threads are found, it takes time more time to configure a build.

pfultz2 (approved):

PR #1992, opened 2023-07-21
cmake: do not copy the driver for backward compatibility on Windows

apwojcik:
The PR from the series to enable MIGraphX on Windows. The deprecated `ADDITIONAL_MAKE_CLEAN_FILES` changed to `ADDITIONAL_CLEAN_FILES`. Reference https://cmake.org/cmake/help/latest/prop_dir/ADDITIONAL_MAKE_CLEAN_FILES.html


umangyadav (approved):

TedThemistokleous (approved):

pfultz2 (approved):

pfultz2:
Although it would probably be still nice to only type `./bin/driver` instead of `./bin/migraphx-driver` on windows, but I guess we can add that back later.

PR #1995, opened 2023-07-21
rename function 'near' to 'within_abs'

apwojcik:
The `Windows.h` header file defines two macros - `near` and `far`. That leads to an unresolvable name clash because the preprocessor precedes source code processing. Both can be safely undefined in a source file after the header is included. We could do that here. However, the `Windows.h` header file is implicitly included from an unknown number of places in this project, and undefining both macros globally is dangerous. We prefer to rename the function instead.


umangyadav (approved):

pfultz2:
Windows headers are very problematic. We should not be including them in any headers or test files. When we need to use something from the windows headers then we should only include it in a localized cpp file, similar to how we dont expose the mlir headers.

apwojcik:
We are not including Windows header files (except a few Windows-specific files, which will be part of other PR). Please remember that even if we do not include them explicitly, other library header files do it for us, especially the standard library. That is the case in this PR.

TedThemistokleous (approved):
Seems like a much more sane solution than dealing with the macros entirely.

pfultz2:
> We are not including Windows header files (except a few Windows-specific files, which will be part of other PR). Please remember that even if we do not include them explicitly, other library header files do it for us, especially the standard library. That is the case in this PR.

The microsoft's c++ standard library doesn't include windows.h in any of its headers. Nor does any of the libraries we include in our headers(ie half).

If you want to rename it, then it should be named something like within_abs not make_near which doesn't make sense. However, we should still check for macros near and far and error out as we should not have windows.h in our headers.

umangyadav:
Onnx runtime kept failing on vega10 machine. I've disabled it inside Jenkins to see if it passes. 
cc: @pfultz2  @TedThemistokleous

PR #1996, opened 2023-07-21
temporarily disable CK on Windows

apwojcik:
The project Composable Kernel still needs to be ported to Windows. We will reenable CK for MIGraphX on Windows as soon as we finish porting.

### DIFF
--- a/src/targets/gpu/CMakeLists.txt
+++ b/src/targets/gpu/CMakeLists.txt
@@ -33,7 +33,10 @@ if(NOT TARGET MIOpen)
     message(SEND_ERROR "Cant find miopen")
 endif()
 
-find_package(composable_kernel 1.0.0 COMPONENTS jit_library REQUIRED) 
+if(NOT WIN32)
+    # TODO: re-enable when CK is ported to Windows
+    find_package(composable_kernel 1.0.0 REQUIRED COMPONENTS jit_library)
+endif()
 
 if(BUILD_DEV)
     set(MIGRAPHX_USE_HIPRTC OFF CACHE BOOL "Use hipRTC APIs")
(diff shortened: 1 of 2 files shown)


apwojcik:
Clang-Tidy is failing
`W: GPG error: http://repo.radeon.com/rocm/apt/5.6 focal InRelease: The following signatures couldn't 
be verified because the public key is not available: NO_PUBKEY 9386B48A1A693C5C`

umangyadav (approved):

src/targets/gpu/CMakeLists.txt
```diff
@@ -131,8 +133,23 @@ add_library(migraphx_gpu
...
+    jit/gather.cpp
+    jit/gathernd.cpp
+    jit/layernorm.cpp
+    jit/mlir.cpp
+    jit/pad.cpp
+    jit/pointwise.cpp
+    jit/reduce.cpp
+    jit/roialign.cpp
+    jit/scatternd.cpp
+    jit/softmax.cpp
```
pfultz2, line 146: Use file globbing and just remove ck files from the list on windows.

apwojcik: https://cmake.org/cmake/help/latest/command/file.html#glob

> **Note**: We do not recommend using GLOB to collect a list of source files from your source tree. If no `CMakeLists.txt` file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. **The CONFIGURE_DEPENDS flag may not work reliably on all generators**, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

Currently, among other reasons, because of `file(GLOB` in many places in CMakeLists.txt (not only MIGraphX), we face significant times to build a single project on Windows in Docker (>= 3h, e.g., for Debug versions). We want to shorten the time to build by reducing globbing as much as possible. I understand the list may change dynamically, and you don't want to keep updating it manually. If that is the case so, here is how we handled that in rocMLIR while building the FAT rockCompiler library: https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1160
If not, we prefer to keep it as it is.

apwojcik: On the other hand, we will make it a separate PR in the future, so we will have the chance to discuss it more. I have changed the PR as you asked.

TedThemistokleous (approved):

PR #1999, opened 2023-07-21
prepare for Windows resources with resource script files

apwojcik:
The C++17 version of `string_view` does not offer `string_view(It first, End last)` constructor (since C++20). Therefore we use `string_view(const char* s, size_type count)` constructor instead.
https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view


cmake/Embed.cmake
```diff
@@ -24,11 +24,8 @@
 find_program(EMBED_LD ld)
 find_program(EMBED_OBJCOPY objcopy)
 
-if(LINUX)
-    option(EMBED_USE_LD "Use ld to embed data files" ON)
-else()
-    option(EMBED_USE_LD "Use ld to embed data files" OFF)
-endif()
+include(CMakeDependentOption)
+cmake_dependent_option(EMBED_USE_LD "Use ld to embed data files" ON "NOT WIN32" OFF)
```
pfultz2, line 28: This is not equivalent to what was there before.

cmake/Embed.cmake
```diff
@@ -97,23 +95,23 @@ function(generate_embed_source EMBED_NAME)
         string(REGEX REPLACE ".[A-Za-z0-9_]+$" "" BASE_NAME ${BASE_NAME})
 
         string(APPEND INIT_KERNELS "
-            { \"${BASE_NAME}\", { ${START_SYMBOL}, ${END_SYMBOL}} },
+            { \"migraphx/kernels/${BASE_NAME}\", { ${START_SYMBOL}, ${LENGTH_SYMBOL} } },
```
pfultz2, line 98: Do not prepend `migraphx/kernels/` to the name. Not all the files we plan to embed in the future will be in the same path.

src/targets/gpu/CMakeLists.txt
```diff
@@ -41,11 +41,20 @@ else()
...
+    # TODO: re-enable when CK is ported to Windows
+    list(REMOVE_ITEM KERNEL_FILES
+        ${CMAKE_CURRENT_SOURCE_DIR}/kernels/include/migraphx/kernels/ck_gemm.hpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/kernels/include/migraphx/kernels/ck.hpp)
+endif()
+
+if(NOT WIN32)
+    include(Embed)
+    add_embed_library(migraphx_kernels ${KERNEL_FILES})
+endif()
```
pfultz2, line 57: Windows should use the same `add_embed_library` function, just with a different implementation that uses the resource files instead.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -153,29 +153,35 @@ std::size_t compute_block_size(std::size_t n, std::size_t max_block_size)
...
+    auto kernels{::migraphx_kernels()};
+    std::vector<src_file> result;
+    std::transform(
+        kernels.begin(),
+        kernels.end(),
+        std::back_inserter(result),
+        [](std::pair<std::string_view, std::string_view> const& elem) { return src_file{elem}; });
+    return result;
+}
+#endif
```
pfultz2, line 168: Windows should be calling the same function. This `ifdef` should be removed.

src/include/migraphx/compile_src.hpp
```diff
@@ -37,8 +37,18 @@ inline namespace MIGRAPHX_INLINE_NS {
 struct src_file
 {
     fs::path path;
-    std::pair<const char*, const char*> content;
-    std::size_t len() const { return content.second - content.first; }
+    std::string_view content;
+
+    src_file() = delete;
```
pfultz2, line 42: Dont delete the default constructor.

cmake/Embed.cmake
apwojcik: Yes, I realized that after I submitted the PR. I will revert this change.

cmake/Embed.cmake
```diff
@@ -97,23 +95,23 @@ function(generate_embed_source EMBED_NAME)
...
     endforeach()
 
     file(WRITE "${PARSE_HEADER}" "
-#include <unordered_map>
-#include <string>
 #include <utility>
-const std::unordered_map<std::string, std::pair<const char*,const char*>>& ${EMBED_NAME}();
+#include <string_view>
+#include <vector>
+std::vector<std::pair<std::string_view, std::string_view>> ${EMBED_NAME}();
```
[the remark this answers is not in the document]
apwojcik: Initially, I thought the same - you are using the map to speed up the search. However, when I found the place you call the function `migraphx_kernels()`, I discovered the result is just repacked from `std::map` to `std::vector`, and later `std::vector` is used all the time. So I replaced `std::map` with `std::vector` for RVO to kick in.

src/targets/gpu/CMakeLists.txt
apwojcik: I have an idea of how to do this. That will be the separate PR. I will remove `if(NOT WIN32)` from here.

cmake/Embed.cmake
apwojcik: Noted, I will revert that.

src/targets/gpu/compile_hip_code_object.cpp
apwojcik: We are not translating on Windows from `pair` to `src_file` - the function `migraphx_kernels` returns `std::vector<src_file>` for better optimizations, i.e., RVO since this data is already in memory and desired format.

src/include/migraphx/compile_src.hpp
apwojcik: My thoughts are the structure `src_file` is implementing the association pattern. Although the default constructor does not harm, it does not help to use the structure optimally. One may first construct an instance of the structure and then assign values to its members to associate the file name with its content. Optimally is to call the constructor with arguments and let the compiler optimize things for us. In the case of containers, deleting the default constructor forces us to use in-place construction, which is why optimal than calling, for instance, e.g., `push_back` from `std::vector` (however, that may not be the best example, but I guess you know what I mean).
I will remove that change if you insist on keeping the default constructor.

cmake/Embed.cmake
pfultz2: > However, how would you plan to distinguish between files that need to be prepended from those that are not required to be prepended?

For kernel files we will probably use the `RELATIVE` parameter like what we do in CK. For other types of files, we will probably make a completely different embed library to use.

cmake/Embed.cmake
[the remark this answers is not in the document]
pfultz2: > Initially, I thought the same - you are using the map to speed up the search. However, when I found the place you call the function migraphx_kernels(), I discovered the result is just repacked from std::map<pair<...>> to std::vector<src_file>, and later std::vector<src_file> is used all the time.

That is the case for the kernel files, but when we add the tuning configs then that will lookup a file by name.

> So I replaced std::map with std::vector for RVO to kick in (on Windows only - on Linux, you still need to translate from pair to src_file).

To optimize it, it would be better to do this transformation once and assign it to a local static variable. We do something similar for CK headers as well where we insert pragmas for the headers.

src/include/migraphx/compile_src.hpp
pfultz2: I would prefer to keep the type regular. This also allows improving readability by using the member variables to construct.

src/targets/gpu/compile_hip_code_object.cpp
pfultz2: The function generated by `add_embed_library` should use the same signature.

cmake/Embed.cmake
```diff
@@ -54,148 +51,97 @@ function(wrap_string)
...
-    if(EMBED_USE_LD)
-        target_sources(${EMBED_NAME} INTERFACE ${OUTPUT_FILES})
-    else()
-        target_sources(${INTERNAL_EMBED_LIB} PRIVATE ${OUTPUT_FILES})
-    endif()
-    target_sources(${EMBED_NAME} INTERFACE $<TARGET_OBJECTS:${INTERNAL_EMBED_LIB}>)
-    target_include_directories(${EMBED_NAME} INTERFACE "${EMBED_DIR}/include")
+    message(STATUS "Generating embedding library '${EMBED_NAME}'")
+    generate_embed_source(${EMBED_NAME} ${SRC_FILE} ${HEADER_FILE} "${PARSE_BASE_DIRECTORY}" SYMBOLS ${SYMBOLS} FILES ${PARSE_UNPARSED_ARGUMENTS})
+    add_library(${EMBED_NAME} STATIC ${SRC_FILE} ${HEADER_FILE} ${OUTPUT_FILES})
```
pfultz2, line 143: When not using gnu LD, it should be using an object library. Otherwise, static builds of migraphx wont work.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -153,29 +153,33 @@ std::size_t compute_block_size(std::size_t n, std::size_t max_block_size)
     return std::min(std::max(min_block_size, block_size), max_block_size);
 }
 
+std::vector<src_file> migraphx_kernels()
```
pfultz2, line 156: Dont overload the function name.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -153,29 +153,33 @@ std::size_t compute_block_size(std::size_t n, std::size_t max_block_size)
...
+{
+    auto kernels{::migraphx_kernels()};
```
pfultz2, line 158: Ideally this should be saved to a static variable so it will only be computed once.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -153,29 +153,33 @@ std::size_t compute_block_size(std::size_t n, std::size_t max_block_size)
...
+    std::vector<src_file> result;
+    std::transform(
+        kernels.begin(),
+        kernels.end(),
+        std::back_inserter(result),
+        [](std::pair<std::string_view, std::string_view> const& elem) { return src_file{elem}; });
```
pfultz2, line 164: Use west const.

cmake/Embed.cmake
```diff
@@ -140,10 +140,13 @@ function(add_embed_library EMBED_NAME)
     endforeach()
     message(STATUS "Generating embedding library '${EMBED_NAME}'")
     generate_embed_source(${EMBED_NAME} ${SRC_FILE} ${HEADER_FILE} "${PARSE_BASE_DIRECTORY}" SYMBOLS ${SYMBOLS} FILES ${PARSE_UNPARSED_ARGUMENTS})
-    add_library(object_${EMBED_NAME} OBJECT ${SRC_FILE} ${HEADER_FILE})
+    add_library(object_${EMBED_NAME} OBJECT ${SRC_FILE} ${HEADER_FILE} $<$<NOT:$<BOOL:${EMBED_USE_LD}>>:${OUTPUT_FILES}>)
```
pfultz2, line 143: Use an `if` statement. `EMBED_USE_LD` is a configure-time variable so there is no need to use generator expressions. Plus, the `if` is much more readable.

cmake/Embed.cmake
```diff
@@ -54,148 +51,102 @@ function(wrap_string)
...
 
 function(add_embed_library EMBED_NAME)
-    set(options)
-    set(oneValueArgs RELATIVE)
-    set(multiValueArgs)
-    cmake_parse_arguments(PARSE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
-
-    file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/embed)
-    file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/embed/${EMBED_NAME})
+    cmake_parse_arguments(PARSE "" "BASE_DIRECTORY" "" ${ARGN})
```
pfultz2, line 130: Dont remove the variables. This is not very readable.

cmake/Embed.cmake
```diff
@@ -54,148 +51,102 @@ function(wrap_string)
...
+    set_target_properties(object_${EMBED_NAME} PROPERTIES POSITION_INDEPENDENT_CODE ON)
     add_library(${EMBED_NAME} INTERFACE)
-    if(EMBED_USE_LD)
-        target_sources(${EMBED_NAME} INTERFACE ${OUTPUT_FILES})
-    else()
-        target_sources(${INTERNAL_EMBED_LIB} PRIVATE ${OUTPUT_FILES})
-    endif()
-    target_sources(${EMBED_NAME} INTERFACE $<TARGET_OBJECTS:${INTERNAL_EMBED_LIB}>)
-    target_include_directories(${EMBED_NAME} INTERFACE "${EMBED_DIR}/include")
+    target_link_libraries(${EMBED_NAME} INTERFACE $<BUILD_INTERFACE:
```
pfultz2, line 149: Dont use the `BUILD_INTERFACE`.

cmake/Embed.cmake
```diff
@@ -54,148 +51,102 @@ function(wrap_string)
...
+            $<TARGET_OBJECTS:object_${EMBED_NAME}>;$<$<BOOL:${EMBED_USE_LD}>:${OUTPUT_FILES}>>)
+    target_include_directories(${EMBED_NAME} INTERFACE $<BUILD_INTERFACE:${EMBED_DIR}/include>)
```
pfultz2, line 151: Dont use the `BUILD_INTERFACE`.

cmake/Embed.cmake
```diff
@@ -54,148 +51,102 @@ function(wrap_string)
...
     add_library(${EMBED_NAME} INTERFACE)
-    if(EMBED_USE_LD)
-        target_sources(${EMBED_NAME} INTERFACE ${OUTPUT_FILES})
-    else()
-        target_sources(${INTERNAL_EMBED_LIB} PRIVATE ${OUTPUT_FILES})
-    endif()
-    target_sources(${EMBED_NAME} INTERFACE $<TARGET_OBJECTS:${INTERNAL_EMBED_LIB}>)
-    target_include_directories(${EMBED_NAME} INTERFACE "${EMBED_DIR}/include")
+    target_link_libraries(${EMBED_NAME} INTERFACE $<BUILD_INTERFACE:
+            $<TARGET_OBJECTS:object_${EMBED_NAME}>;$<$<BOOL:${EMBED_USE_LD}>:${OUTPUT_FILES}>>)
```
pfultz2, line 150: Dont use generator expression here either. This is unreadable.

cmake/Embed.cmake
```diff
@@ -54,148 +51,102 @@ function(wrap_string)
...
-    
-    set(INTERNAL_EMBED_LIB embed_lib_${EMBED_NAME})
-    add_library(${INTERNAL_EMBED_LIB} OBJECT "${SRC_FILE}")
-    target_include_directories(${INTERNAL_EMBED_LIB} PRIVATE "${EMBED_DIR}/include")
-    target_compile_options(${INTERNAL_EMBED_LIB} PRIVATE -Wno-reserved-identifier -Wno-extern-initializer -Wno-missing-variable-declarations)
-    set_target_properties(${INTERNAL_EMBED_LIB} PROPERTIES POSITION_INDEPENDENT_CODE On)
-    
+    message(STATUS "Generating embedding library '${EMBED_NAME}'")
+    generate_embed_source(${EMBED_NAME} ${SRC_FILE} ${HEADER_FILE} "${PARSE_BASE_DIRECTORY}" SYMBOLS ${SYMBOLS} FILES ${PARSE_UNPARSED_ARGUMENTS})
+    add_library(object_${EMBED_NAME} OBJECT ${SRC_FILE} ${HEADER_FILE} $<$<NOT:$<BOOL:${EMBED_USE_LD}>>:${OUTPUT_FILES}>)
```
pfultz2, line 143: This should be prefixed with `embed_lib` to make sure the name is unique.

pfultz2:
There seems to be a lot of unnecessary refactoring of code still in this PR. I think it would be better to strip it down to only the minimal changes needed to change it from `std::pair` to `string_view`. This will help make sure we dont break anything and will reduce churn if we merge any update to the embed.cmake file.

cmake/Embed.cmake
apwojcik: We need generator expressions because of MSBuild, which is a multi-config generator.

cmake/Embed.cmake
apwojcik: The name `object_*` is unique as much as `embed_lib_*` was, but if you insist, I will make it again `embed_lib_`.

cmake/Embed.cmake
apwojcik: I was trying to solve the issue of being unable to build the MIGraphX static library and added `BUILD_INTERFACE` as an experiment.

cmake/Embed.cmake
apwojcik: This has nothing to do with readability. We use MSBuild, the multi-config generator that works with generator expressions, not `if` statements and variables.

cmake/Embed.cmake
apwojcik: An experiment to solve the issue of being unable to build the static library.

cmake/Embed.cmake
apwojcik: The example code from the CMake book you refer to uses variables not to increase readability but to illustrate that arguments to `cmake_parse_arguments()` are lists. The list in CMake is `set(mylist a b c)`, which evaluates to `"a;b;c"`. Those are legitimate transcriptions of lists in CMake.
My opinion is the opposite - adding unnecessary variables makes code cluttered and less readable.

cmake/Embed.cmake
pfultz2: It is off by default for linux as well. So its not the same.

src/targets/gpu/compile_hip_code_object.cpp
pfultz2: But name it something else to avoid ambiguity. Why is this even needed for windows?

cmake/Embed.cmake
pfultz2: But you aren't using any of the `$<CONFIG>` expressions so its the same value across all configs so it can be set using `if` statements.

apwojcik:
I only left the changes required for `string_view`. I will submit a separate PR for `Embed.cmake` refactoring - `EMBED_USE_LD` does not currently work on the `develop` branch, and I fixed it while refactoring.

pfultz2 (approved):

PR #2000, opened 2023-07-22
Lw/fix half shape

lakhinderwalia:
Shape (type) fix for test cases: Instruction should operate on shape types, other than just (defaulted to) Float.


umangyadav:
You would need to handle "tuple" types and add unit tests inside `op_shape_test.cpp`.

lakhinderwalia:
> You would need to handle "tuple" types and add unit tests inside `op_shape_test.cpp`.

Do we always know the size of Tuple at the Compile time? A split operator can split a tensor in N slices, N is not a compile time constant. https://www.tensorflow.org/api_docs/python/tf/split

umangyadav:
> Do we always know the size of Tuple at the Compile time? 

The way it is setup right now, It is always known compile time. 

> A split operator can split a tensor in N slices, N is not a compile time constant. 

MIGraphx doesn't support a dynamic-shaped "split" operator at the moment.

lakhinderwalia:
Comedy of errors. You were referring to shape::tuple_type. I was referring to std::tuple. Sorted that out.. :-)

src/module.cpp
```diff
@@ -460,7 +460,8 @@ instruction_ref module::add_parameter(std::string name, shape s)
 
 instruction_ref module::add_return(std::vector<instruction_ref> args)
 {
-    impl->push_back({builtin::returns{}, {}, std::move(args)});
+    shape instr_shape = args.empty() ? shape{} : args[0]->get_shape();
```
pfultz2, line 463: This logic should be in the `compute_shape` method. And then it can just do `shape instr_shape = compute_shape(builtin::returns{}, args)`. This is similar to how `insert_instructions` work.

src/instruction.cpp
```diff
@@ -124,7 +122,7 @@ bool instruction::valid() const
     }
     else if(op.name() == "@return")
     {
-        computed = {};
+        computed = get_shape();
```
pfultz2, line 125: I think this whole `if` branch can be removed.

src/module.cpp
lakhinderwalia: @pfultz2 thanks. Makes sense. Should this logic be added to src/include/migraphx/op/reduce.hpp? (There isn't a file yet like that).

pfultz2: > Should this logic be added to src/include/migraphx/op/reduce.hpp?

What logic are you referring to?

lakhinderwalia: The compute_shape logic for builtin::returns.

src/include/migraphx/builtin.hpp
```diff
@@ -90,7 +90,12 @@ struct param
 struct returns
 {
     std::string name() const { return "@return"; }
-    shape compute_shape(const std::vector<shape>&) const { return {}; }
+    shape compute_shape(const std::vector<shape>& arg) const
+    {
+        if(arg.size() > 1)
+            return shape(arg);
+        return arg[0];
```
umangyadav, line 97: if `arg` is  empty it would seg fault.

test/ref_ops_test.cpp
```diff
@@ -8607,6 +8607,25 @@ TEST_CASE(topk_test)
     }
 }
 
+TEST_CASE(output_shape_is_tuple_type_test)
```
umangyadav, line 8610: No need to add test here. `ref_ops_test.cpp` is meant for testing "ref" implementation and making sure if it is producing correct results or not. Return is a builtin operator and doesn't have `compute()` defined.

test/shape_test.cpp
```diff
@@ -737,6 +738,22 @@ TEST_CASE(tuple)
     EXPECT(test::throws([&] { s.visit_type([](auto) {}); }));
 }
 
+TEST_CASE(return_shape_tuple)
+{
+    using migraphx::shape;
```
umangyadav, line 743: replace `shape::` with `migraphx::shape` in callsites.  it is okay to use `using` but alteast here `shape::` is only used in couple of places.

test/shape_test.cpp
```diff
@@ -737,6 +738,22 @@ TEST_CASE(tuple)
...
+    auto op = migraphx::make_op("@return");
+    std::vector<shape> s{{shape::float_type}, {shape::float_type}};
```
umangyadav, line 745: You can also pass in You can also pass in `lens` and `strides` here while constructing the shape. It is optional though. It would just use default if they are not passed. But in practice it is better to test with lens and strides.

src/include/migraphx/builtin.hpp
```diff
@@ -90,7 +90,12 @@ struct param
 struct returns
 {
     std::string name() const { return "@return"; }
-    shape compute_shape(const std::vector<shape>&) const { return {}; }
+    shape compute_shape(const std::vector<shape>& arg) const
+    {
+        if(arg.size() > 1)
+            return shape(arg);
```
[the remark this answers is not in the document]
lakhinderwalia: (Yes, agreed. That was the original change I had made.. until I added another case..)

test/shape_test.cpp
lakhinderwalia: `    std::vector<shape> s{{shape::float_type}, {shape::float_type}};`

This line looks ugly, with three times the shape scope expansion.. Overflows to 2nd line. 
` std::vector<migraphx::shape> s{{migraphx::shape::float_type}, {migraphx::shape::float_type}};`

That is why `using` is very useful here for readability.

src/include/migraphx/builtin.hpp
lakhinderwalia: Agreed. Thanks. This check was inadvertently removed in code reorg. Good catch.

test/shape_test.cpp
lakhinderwalia: I think while comparing the shape types, we are just doing a default constructor here, and checking again with defaults.. so will take the optional (easy) way out.. and leave it as it is..; Thanks.

src/instruction.cpp
lakhinderwalia: Done.

test/shape_test.cpp
```diff
@@ -737,6 +738,22 @@ TEST_CASE(tuple)
...
+    EXPECT(op.compute_shape(s).type() == shape::tuple_type);
```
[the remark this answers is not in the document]
lakhinderwalia: OK, this doesn't work. This shouldn't work :-) 
(And I failed to verify it, it seems ..)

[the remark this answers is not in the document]
umangyadav: why doesn't it work ? what's the error ?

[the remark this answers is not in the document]
lakhinderwalia: Well, the assertion fails. 
`[2023-07-27T21:13:31.115Z] test_op_shape_test: /home/jenkins/workspace/bs_AMDMIGraphX_lw_fix_half_shape/src/shape.cpp:51: migraphx::shape_impl::shape_impl(shape::type_t): Assertion "t != shape::tuple_type" failed.
[2023-07-27T21:13:31.115Z] CMake Error at gdb/test_test_op_shape_test/run.cmake:12 (message):`

Comparing a default shape of tuple_type: its default construction would give out something: But it won't be comparable to hand-crafted shape (tuple_type) vector of size 2 or 3 or..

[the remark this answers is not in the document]
umangyadav: `EXPECT(op.compute_shape(s) == migraphx::shape{s})` This can work i think.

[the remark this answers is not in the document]
lakhinderwalia: Yes, but I am not sure if it gives out a clear message that it is a tuple_type. I think testing @return beyond a point is an exercise in tautology. Thanks.

test/op_shape_test.cpp
```diff
@@ -2597,6 +2597,22 @@ TEST_CASE(reshape_non_fixed_not_matching_error)
...
 
+TEST_CASE(return_shape_tuple)
+{
+    using migraphx::shape;
+    auto op = migraphx::make_op("@return");
+    std::vector<shape> s{{shape::float_type}, {shape::float_type}};
+    EXPECT(op.compute_shape(s).type() == shape::tuple_type);
+}
+
+TEST_CASE(return_shape_half)
```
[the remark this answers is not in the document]
lakhinderwalia: Sure.

test/shape_test.cpp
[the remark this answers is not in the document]
umangyadav: test_name has clear message that is  meant for testing tuple type. 
IMO, we should be testing for equality of shapes instead of just checking for equality of types which is just one of the member of shape class. 

also return type of "@return" operator should guarantee that it preserves the lens and strides not just type. and we should be testing for that ideally.

umangyadav (approved):

src/instruction.cpp
```diff
@@ -64,10 +64,8 @@ void instruction::replace(const shape& r)
         result = r;
         for(auto&& ins : output)
         {
-            if(ins->name() == "@return")
-                continue;
-
-            assert(ins->name().front() != '@');
+            if(ins->name() != "@return")
+                assert(ins->name().front() != '@');
```
pfultz2, line 68: Instead write this like `assert(ins->name() == "@return" or ins->name().front() != '@')`.

pfultz2 (approved):

src/instruction.cpp
lakhinderwalia: Changed. Thanks.

PR #2001, opened 2023-07-24
[MLIR] Capture diagnostics, handle compile failure

krzysz00:
Add mlir_logger, which registers a MLIR diagnostic handler that captures any information generated by a MLIR compile and saves it to a string.

This will be useful during tuning, where some such errors may be the result of an inapplicable tuning configuration and should be suppressed.

However, if the high-level pipeline rejects an input, this is almost certainly an error in the MLIR construction and an exception.


src/targets/gpu/mlir.cpp
```diff
@@ -167,6 +168,51 @@ std::string mlir_print(F f, T x)
...
+{
+    return reinterpret_cast<mlir_logger*>(logger)->handle(diag);
+}
+
+mlir_logger::mlir_logger(mlir_context& _ctx) : ss(diagnostics), ctx(_ctx)
+{
+    id = mlirContextAttachDiagnosticHandler(ctx.get(), mlir_diagnostic_print_cb, this, nullptr);
+}
+
+mlir_logger::~mlir_logger() { mlirContextDetachDiagnosticHandler(ctx.get(), id); }
```
pfultz2, line 195: Make this inline in the class.

src/targets/gpu/mlir.cpp
```diff
@@ -167,6 +168,51 @@ std::string mlir_print(F f, T x)
...
+
+MlirLogicalResult mlir_logger::handle(MlirDiagnostic diag)
```
pfultz2, line 197: Make this inline in the class.

src/targets/gpu/mlir.cpp
```diff
@@ -167,6 +168,51 @@ std::string mlir_print(F f, T x)
     return ss.str();
 }
 
+struct mlir_logger
+{
+    std::string diagnostics;
+    std::stringstream ss;
+
+    mlir_context& ctx;
```
pfultz2, line 176: Use a pointer instead of reference here.

src/targets/gpu/mlir.cpp
```diff
@@ -735,6 +791,7 @@ struct mlir_program
     mlir_context ctx;
     MlirLocation location;
     mlir_module mmodule;
+    mlir_logger logger;
```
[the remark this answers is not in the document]
krzysz00: It's used to do all the setup

src/targets/gpu/mlir.cpp
```diff
@@ -167,6 +168,51 @@ std::string mlir_print(F f, T x)
     return ss.str();
 }
 
+struct mlir_logger
+{
+    std::string diagnostics;
```
pfultz2, line 173: I dont see this being used anywhere except to construct stringstream.

krzysz00: We pull it out in the `THROW` I remember

src/targets/gpu/mlir.cpp
```diff
@@ -622,12 +669,21 @@ struct mlir_program
         mlir_pass_manager pm_back{mlirPassManagerCreate(ctx.get())};
         // 1st pipeline to call
         mlirMIGraphXAddHighLevelPipeline(pm_front.get());
-        mlirPassManagerRunOnOp(pm_front.get(), mlirModuleGetOperation(mmodule.get()));
+        if(mlirLogicalResultIsFailure(
+               mlirPassManagerRunOnOp(pm_front.get(), mlirModuleGetOperation(mmodule.get()))))
+        {
+            std::string error = "Invalid MLIR created: " + logger.diagnostics;
```
pfultz2, line 675: `logger.disgnostics` is always an empty string. Did you mean to write something like `logger.ss.str()` instead?

src/targets/gpu/mlir.cpp
```diff
@@ -167,6 +168,51 @@ std::string mlir_print(F f, T x)
...
+struct mlir_logger
+{
+    std::string diagnostics;
+    std::stringstream ss;
+
+    mlir_context& ctx;
+    MlirDiagnosticHandlerID id;
+
+    mlir_logger(mlir_context& ctx);
+    ~mlir_logger();
```
pfultz2, line 180: Add a default constructor, move constructor and move assignment operators and delete the copy constructor and copy assignment.

src/targets/gpu/mlir.cpp
pfultz2: But `diagnostics` is never assigned a value. Its always an empty string.

src/targets/gpu/mlir.cpp
```diff
@@ -622,12 +669,21 @@ struct mlir_program
...
+        }
 
         // 2nd pipeline to call
         get_module_tuned();
         mlirMIGraphXAddBackendPipeline(pm_back.get(), target_arch.c_str());
-        mlirPassManagerRunOnOp(pm_back.get(), mlirModuleGetOperation(mmodule.get()));
+        if(mlirLogicalResultIsFailure(
+               mlirPassManagerRunOnOp(pm_back.get(), mlirModuleGetOperation(mmodule.get()))))
+        {
+            std::cerr << "MLIR backend compilation failed" << std::endl;
```
pfultz2, line 685: Shouldn't we print an error from here as well?

src/targets/gpu/mlir.cpp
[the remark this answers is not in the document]
pfultz2: I see now. Its registered in the constructor.

src/targets/gpu/mlir.cpp
```diff
@@ -622,12 +669,21 @@ struct mlir_program
         mlir_pass_manager pm_back{mlirPassManagerCreate(ctx.get())};
         // 1st pipeline to call
         mlirMIGraphXAddHighLevelPipeline(pm_front.get());
-        mlirPassManagerRunOnOp(pm_front.get(), mlirModuleGetOperation(mmodule.get()));
+        if(mlirLogicalResultIsFailure(
```
pfultz2, line 672: We should clear the log before running a pass.

src/targets/gpu/mlir.cpp
```diff
@@ -622,12 +669,21 @@ struct mlir_program
         mlir_pass_manager pm_back{mlirPassManagerCreate(ctx.get())};
         // 1st pipeline to call
         mlirMIGraphXAddHighLevelPipeline(pm_front.get());
-        mlirPassManagerRunOnOp(pm_front.get(), mlirModuleGetOperation(mmodule.get()));
```
pfultz2, line 625: Since these are refactored into seperate functions in #1965, I would prefer that PR is merged first before this one.

src/targets/gpu/mlir.cpp
krzysz00: I didn't realize that stringstream doesn't work like the LLVM version where it takes the argument as a backing store.

src/targets/gpu/mlir.cpp
krzysz00: @pfultz2 Do youn know what the clang-tidy warning just above here is complaining about?

src/targets/gpu/mlir.cpp
```diff
@@ -178,13 +179,85 @@ std::string mlir_print(F f, T x)
...
+    mlir_logger& operator=(const mlir_logger& other) = delete;
+
+    mlir_logger(mlir_logger&& other) noexcept
+        : ss(std::move(other.ss)), ctx(other.ctx), id(other.id)
+    {
+        other.ctx = nullptr;
+        other.id  = std::nullopt;
+    }
+
+    mlir_logger& operator=(mlir_logger&& other) noexcept
```
pfultz2, line 212: This should take the logger by value and swap the elements:

```cpp
mlir_logger& operator=(mlir_logger other) noexcept
{
    std::swap(ss, other.ss);
    std::swap(ctx, other.ctx);
    std::swap(is, other.id);
    return *this;
}
```

src/targets/gpu/mlir.cpp
```diff
@@ -616,14 +689,25 @@ struct mlir_program
     {
         mlir_pass_manager pm_front{mlirPassManagerCreate(ctx.get())};
         mlirMIGraphXAddHighLevelPipeline(pm_front.get());
-        mlirPassManagerRunOnOp(pm_front.get(), mlirModuleGetOperation(mmodule.get()));
+        logger.ss = std::stringstream{};
+        if(mlirLogicalResultIsFailure(
+               mlirPassManagerRunOnOp(pm_front.get(), mlirModuleGetOperation(mmodule.get()))))
+        {
+            std::string error = "Invalid MLIR created: " + logger.ss.str();
```
pfultz2, line 696: Add a `str` method to the logger as well to get the string.

src/targets/gpu/mlir.cpp
causten: 1965 has been merged

pfultz2 (approved):

PR #2003, opened 2023-07-24
Multinomial parse

bpickrel:
Adds new operation rand_uniform and refactors multinomial op. to accept dynamic shape inputs.  These changes will allow multinomial distributions to be created from random numbers at runtime.

The ONNX Multinomial instruction is implemented as a combination of rand_uniform and multinomial operations, since Migraphx' Multinomial doesn't generate a sequence of random numbers but requires them to be passed as an argument.

Use of random seed in rand_uniform is complicated, with a lot of command permutations.  This change allows four ways of seeding the random sequence:

1.   Specify a seed when compiling the model.  This will always result in the same pseudo-random sequence every time the model is run, although it may not be apparent if inputs are of dynamic size.
2.   Specify a seed at runtime.
3.   Instruct the operation to auto-generate seeds with a runtime input.
4.   Command auto-seeding at compile time.  This is the Onnx specified behavior when a seed isn't given.

Options `#2` and `#3` don't apply to Onnx, since it doesn't recognize separate compile and runtime steps for a model.  The advantage of these choices is that the same compiled model can either use a specified seed and thus be repeatable, or auto-generate a seed if none is given.

In order to support options `#2` and `#3`, we would need to make the seed optional at runtime, not just at compile time.  This presents a potential conflict in that `output-lens`, the shape of the output, is also an optional argument because we might have to override the compiled shape in the case of dynamic inputs.  A language such as Migraphx with positional arguments can't have an operation with two optional arguments because if only one is provided at runtime, there is no (simple) way to distinguish which one it is.

The solution to this conflict in this implementation is that an output shape argument is optional but a seed argument is required.  If there is at least one value in the seed argument, it is used.  To command auto-seeding when running an inference, input an argument with a 0-size shape.  A shape with no rank (an empty lens() or dyn_dims() vector) is a runtime error, but a shape with rank one containing a length of 0 represents an empty data array and is legal.  For case `#4`, there is a flag `use_auto_seed` which, if set, causes auto-seeding no matter what seed value is passed.  This latter supports the Onnx requirement.

In Onnx parsing, then, `use_auto_seed` is set if no seed argument is present.


src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+        }
+        else
+        {
+            return s.with_lens(s.lens()).with_type(dtype);
+        }
+    }
+
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    {
+        (void)args; // suppress compiler warning
```
pfultz2, line 99: Remove this line as args are used.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+        auto local_seed(seed);
+        if(args.size() > 1)
+        {
+            if(args.at(1).get_shape().element_space() > 0)
+            {
+                visit_all(args[1])([&](auto data) { local_seed = data[0]; });
+            }
+            else // obtain a seed from the system clock:
+                local_seed = std::chrono::system_clock::now().time_since_epoch().count();
```
pfultz2, line 110: Remove this as the seed should always be set from `arg[0]`.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+    {
+        (void)args; // suppress compiler warning
+        argument result{dyn_out.computed_shape};
+
+        auto local_seed(seed);
+        if(args.size() > 1)
+        {
+            if(args.at(1).get_shape().element_space() > 0)
+            {
+                visit_all(args[1])([&](auto data) { local_seed = data[0]; });
```
pfultz2, line 107: The seed should be first argument, and the second argument will be the dimensions. Also you can write `args[0].at<uint32_t>(0)`.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include <random>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct rand_uniform
+{
+    uint32_t sample_size = {20};
+    uint32_t seed        = {3};
```
pfultz2, line 52: Remove this as the seed should be passed as an argument not a attribute.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+    float range_min      = 0.0f;
+    float range_max      = 1.0f;
+
+    // From Onnx RandomUniform:
+    // dtype : int (default is 1)
+    // The data type for the elements of the output tensor. If not specified, default is
+    // TensorProto::FLOAT. high : float (default is 1.0) Upper boundary of the output values. low :
+    // float (default is 0.0) Lower boundary of the output values. seed : float (Optional) Seed to
+    // the random generator, if not specified we will auto generate one. shape : list of ints
+    // (required) The shape of the output tensor.
```
pfultz2, line 62: I dont think we should add the onnx documentation here as this operator will be different from the onnx spec.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+// #include <migraphx/reflect.hpp>
+#include <random>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct rand_uniform
+{
+    uint32_t sample_size = {20};
```
pfultz2, line 51: This should be removed and `std::vector<size_t> output_lens = {}` should be used instead for the static case.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+    static auto reflect(Self& self, F f)
+    {
+        return pack(
+            f(self.dtype, "dtype"), f(self.sample_size, "sample_size"), f(self.seed, "seed"));
+    }
+
+    value attributes() const { return {{"sample_size", sample_size}, {"seed", seed}}; }
+
+    std::string name() const { return "rand_uniform"; }
+    shape normalize_compute_shape(std::vector<shape> inputs) const
```
pfultz2, line 80: Just use `compute_shape` since you are not using any normalization.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+    {
+        check_shapes{inputs, *this, true}.has(1, 2);
+        if(inputs.size() > 1 and inputs.at(1).element_space() > 0 and
```
pfultz2, line 83: This is backwards as the output dimensions should be the second parameter since those are optional.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+           inputs.at(1).type() != shape::type_t::uint32_type)
+            MIGRAPHX_THROW("RAND_UNIFORM:  Input 2 (seed) must have type unsigned int");
+        auto s = inputs.front();
+        if(s.dynamic())
```
pfultz2, line 87: This should be `if(inputs.size() > 1)`

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+        {
+            return s.with_type(dtype);
```
pfultz2, line 89: This is wrong, you should be returning a dynamic shape based on the rank of `s`.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+        }
+        else
+        {
+            return s.with_lens(s.lens()).with_type(dtype);
+        }
+    }
+
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
```
pfultz2, line 97: Dont use `dyn_output` here.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,128 @@
...
+    {
+        (void)args; // suppress compiler warning
+        argument result{dyn_out.computed_shape};
```
pfultz2, line 100: This is wrong. It should be:

```cpp
{
    // Dynamic shape
    if(args.size() > 1)
    {
        output_shape = shape{dtype, args[1].get<std::size_t>().to_vector()};
    }
    ...
}
```

src/common.cpp
```diff
@@ -209,6 +209,8 @@ instruction_ref insert_common_op(module& m,
                                  const operation& op,
                                  std::vector<instruction_ref> inputs)
 {
+    if(op.name() == "clip")
+        return inputs[0];
```
pfultz2, line 213: This should be removed. This will probably break quantization simplifications.

bpickrel:
Merged the random_uniform and random_seed changes which are now in Develop branch, and updated tests to match.

src/onnx/parse_multinomial.cpp
```diff
@@ -46,38 +46,79 @@ struct parse_multinomial : op_parser<parse_multinomial>
...
 
-        // Pre-compute random distribution
-        std::mt19937 gen(std::chrono::high_resolution_clock::now().time_since_epoch().count());
+        instruction_ref seed_input;
         if(contains(info.attributes, "seed"))
-            gen.seed(info.attributes.at("seed").f());
+        {
+            uint32_t seed = info.attributes.at("seed").i();
+            migraphx::shape s{migraphx::shape::uint32_type, {1}};
+            std::vector<uint32_t> data = {seed};
```
CharlieL7, line 77: ONNX spec says the `seed` attribute is a float. This will convert to a `uint32_t`. I'm not sure what's the intended behavior by the ONNX  spec. There will be a problem with the current `random_seed` op if we request it to create a floating point value since `std::random_device` only does integers. We probably want to check the accuracy of `multinomial` against ORT or something in a `verify_onnx.cpp` test.

src/onnx/parse_multinomial.cpp
```diff
@@ -46,38 +46,79 @@ struct parse_multinomial : op_parser<parse_multinomial>
...
+                    migraphx::make_op("random_uniform"), seed_input, rand_dummy);
+            }
+        }
 
-        std::uniform_real_distribution<> dis(0.0, 1.0);
-        size_t batch_size = args[0]->get_shape().lens().front();
-        migraphx::shape dist_shape{migraphx::shape::float_type, {batch_size, sample_size}};
+        // TODO:  Confirm that this case with no inputs should not be supported.  The input defines
+        // the number of categories, which can't be defaulted.  You can't get to this line with no
+        // inputs because there will be an exception when accessing args[0] above.
```
CharlieL7: Yes, `multinomial` without any inputs is not in the ONNX spec.

src/onnx/parse_multinomial.cpp
```diff
@@ -46,38 +46,79 @@ struct parse_multinomial : op_parser<parse_multinomial>
             dtype = info.attributes.at("dtype").i();
         shape::type_t output_type = get_type(dtype);
 
+        size_t batch_size = 1;
+        if(contains(info.attributes, "batch_size"))
+            batch_size = info.attributes.at("batch_size").i();
```
CharlieL7, line 54: What is this attribute for? It's not in the ONNX spec

src/include/migraphx/op/multinomial.hpp
```diff
@@ -21,11 +21,42 @@
...
+ * Multinomial or categorical distribution.  Performs a sampling and returns a count of
+ *         each category, or bucket.  This does not require the standard multinomial
+ *         distribution but instead takes a probability distribution as an input.
+ *
+ *          In the large number limit, the fractional counts approach the multinomial distribution.
+ *
+ *      Inputs:   args[0] - a vector of probabilities for each category.  Values are running
+ *                          totals as provided by op prefix_scan_sum.
+ *                          Values are log normalized (i.e. start with any set of numbers > 0, then
+ *                          val[i] = log(val[i]) / sum (log(val[0]) + log(val[1])+ ...) )
```
CharlieL7, line 35: Didn't the `parse_multinomial` take the element-wise exponent of the probabilities? The probabilities would be then no longer be `log(x)`.

src/include/migraphx/op/multinomial.hpp
```diff
@@ -47,31 +78,49 @@ struct multinomial
...
-    argument compute(const shape& output_shape, std::vector<argument> args) const
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
     {
-        argument result{output_shape};
-        size_t batch_size  = output_shape.lens().front();
-        size_t class_size  = args[0].get_shape().lens().back();
-        size_t sample_size = output_shape.lens().back();
+        argument result{dyn_out.computed_shape};
+        size_t batch_size  = dyn_out.computed_shape.lens().front();
+        size_t class_size  = args[0].get_shape().max_lens().back();
```
CharlieL7, line 109: `lens()` should work because the argument has a static shape at runtime.

src/include/migraphx/op/multinomial.hpp
```diff
@@ -21,11 +21,42 @@
...
+ *          In the large number limit, the fractional counts approach the multinomial distribution.
+ *
+ *      Inputs:   args[0] - a vector of probabilities for each category.  Values are running
+ *                          totals as provided by op prefix_scan_sum.
+ *                          Values are log normalized (i.e. start with any set of numbers > 0, then
+ *                          val[i] = log(val[i]) / sum (log(val[0]) + log(val[1])+ ...) )
+ *                          This input has Rank 2.  Dimension 0 is batch #.  The size of dimension
+ *                          1 is the number of categories.
+ *                args[1] - a vector of random numbers.  Can be 1-, 2- or more dimensions but all
+ *                          except the last dim must be 1.
```
CharlieL7, line 39: I don' understand this comment, the second argument can have dimensions `{batch_size, sample_size}`.

test/ref/multinomial.cpp
```diff
@@ -78,5 +170,6 @@ TEST_CASE(multinomial_test)
     std::transform(res_dist.begin(), res_dist.end(), res_norm.begin(), [&](auto n) {
         return static_cast<double>(n) / res_dist_sum;
     });
+    // The given test tolerance is about 10x the typical error
```
CharlieL7, line 173: Wouldn't this test tolerance be `1e6 / 80` of the default tolerance? The default multiplier we have on the epsilon is 80.

src/onnx/parse_multinomial.cpp
```diff
@@ -46,38 +46,79 @@ struct parse_multinomial : op_parser<parse_multinomial>
...
-        auto mb_maxes = info.add_instruction(
-            migraphx::make_op("multibroadcast", {{"out_lens", args[0]->get_shape().lens()}}),
-            maxes);
-        auto cdf = info.add_instruction(migraphx::make_op("sub"), args[0], mb_maxes);
+        auto cdf = info.add_common_op("sub", args[0], maxes);
         // Take the element-wise exponent to get probabilities in the range (0, 1]
         cdf = info.add_instruction(migraphx::make_op("exp"), cdf);
         // Compute the cumulative density function
         cdf = info.add_instruction(
             migraphx::make_op("prefix_scan_sum", {{"axis", 1}, {"exclusive", false}}), cdf);
```
CharlieL7, line 70: I have a doubt that this is working correctly. For example if you do the `cdf` calculation using the example probabilities in the ref_ops test: `[15, 25, 15, 25, 20]`. The resulting `cdf` is `[0.6, 1.6, 2.2, 3.2, 4.0]`. Looks like there's a missing division to me.

CharlieL7 (changes requested):
There's CI errors that look like some out-of-sync problem with the CI vs. the code.

src/onnx/parse_multinomial.cpp
bpickrel: That's right, removing.  batch_size appears only implicitly.

src/onnx/parse_multinomial.cpp
bpickrel: Subtracting from the logs is effectively division. Discussion continued in Teams chat.

src/onnx/parse_multinomial.cpp
bpickrel: I think Paul kept asking me to make changes to the seed type for the op and in one version the op only supported one type, but I don't think there's any obstacle to changing it to a float here now.  I just want to make sure we agree on what we're trying to accomplish before I make the change.

bpickrel: I would have to change some tests and Onnx test files to match the change.

test/ref/multinomial.cpp
bpickrel: This is with 100,000 samples, so that's not necessarily such a large tolerance.

src/include/migraphx/op/multinomial.hpp
bpickrel: I need to add a test with a dynamic batch size > 1 to clarify this.  I'll get back to you.

bpickrel: I'm finding problems with the existing logic that takes batch size from arg[0] and another dimension from arg[1].  The only way it would make sense to include a batch dimension in arg[0]--which is the probability distribution--is if we wanted to do multiple multinomials with different probability distributions.  I don't think there's any requirement for such a thing.

src/include/migraphx/op/multinomial.hpp
bpickrel: changed

src/include/migraphx/op/multinomial.hpp
bpickrel: I refactored the logic and updated the comment.  It's much simpler now.  First argument is a vector, size=number of categories.  Second argument can be any shape, and the output shape is that of the second argument.  There's no special handling for batch size.

Just one bug, under investigation:  it crashes if the second argument has less than 3 dimensions.  Extra size-1 dimensions can be added with Unsqueeze without affecting results.

test/ref/multinomial.cpp
bpickrel: This test has been replaced with verify_range_with_tolerance()

src/include/migraphx/op/multinomial.hpp
bpickrel: On review, that's what the Onnx spec calls for after all:  you can have multiple CDFs within the same batch.  So, for instance with batch size 2, 5 categories, and 7000 data samples for each distribution, input 1 would be 2x5 and input 2 would be 2x7000.  Input 2 could actually have more dimensions as long as the last dimension is 7000 to deduce the sample_size.  For instance 1x1x2x7000 but I don't think there's any need for that so I changed the comment.

src/include/migraphx/op/multinomial.hpp
bpickrel: This comment has been corrected.  Inputs to parse_multinomial are logs of probabilities and it adds an exp instruction to convert to non-logs.  The MigraphX Multinomial op. doesn't use the same inputs.  It takes (a cumulative sum of) straight values as inputs, but they do not have to be normalized.  Multiplying or dividing them by any constant leaves the relative probabilities unchanged.  The normalization factor is the last value of the cumulative sum and is applied at runtime.

src/onnx/parse_multinomial.cpp
bpickrel: Changed the type of seed in parse_multinomial to float.  Should be transparent to the MIGraphX ops.

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -65,11 +65,14 @@ struct random_uniform
         return inputs.at(1);
     }
 
-    argument compute(const shape&, std::vector<argument> args) const
-    {
-        // Output goes into the passed buffer, not the shape output.
-        auto result = args[1];
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    // const shape&, std::vector<argument> args) const
```
CharlieL7, line 69: cleanup comment

src/onnx/parse_multinomial.cpp
```diff
@@ -49,35 +52,95 @@ struct parse_multinomial : op_parser<parse_multinomial>
...
+            std::vector<int64_t> data1{1, 0};
+            auto l1        = info.add_literal(s, data1);
+            auto batch_arg = info.add_instruction(migraphx::make_op("mul"), dim_of, l1);
+            std::vector<int64_t> data2(2, 0);
+            // make an argument of (0, sample_size)
+            data2[1]         = sample_size;
+            auto l2          = info.add_literal(s, data2);
+            auto alloc_shape = info.add_instruction(migraphx::make_op("add"), batch_arg, l2);
+            // alloc_shape should contain the input-based shape dimensions as its values at runtime,
+            // and its own shape is (2, 0)
```
CharlieL7, line 114: It looks like it's shape is dims = {2}

src/onnx/parse_multinomial.cpp
```diff
@@ -49,35 +52,95 @@ struct parse_multinomial : op_parser<parse_multinomial>
...
+                migraphx::shape(s0.type(), {s0.dyn_dims().front(), {sample_size, sample_size}});
 
-        std::uniform_real_distribution<> dis(0.0, 1.0);
-        size_t batch_size = args[0]->get_shape().lens().front();
-        migraphx::shape dist_shape{migraphx::shape::float_type, {batch_size, sample_size}};
+            // Allocate on-device storage for the random values
+            // TODO:  this creates data type half_type if "buf_type" is not specified
+            auto alloc = info.add_instruction(
+                migraphx::make_op("allocate",
+                                  {{"shape", to_value(compile_shape)}, {"buf_type", s0.type()}}),
```
CharlieL7, line 126: You don't need to set `buf_type` if the `shape` attribute is set since `allocate` will use the type from `shape`. I can create a small update to change buf_type to default to nothing and add a check for it if the `shape` attribute is not set.

src/include/migraphx/op/multinomial.hpp
```diff
@@ -47,36 +88,52 @@ struct multinomial
...
     {
-        check_shapes{inputs, *this}.has(2).only_dims(2);
-        size_t sample_size = inputs.back().lens().back();
+        check_shapes{inputs, *this, true}.has(2).only_dims(2);
 
-        if(not contains({shape::int32_type, shape::int64_type}, dtype))
-            MIGRAPHX_THROW(
-                "Multinomial: Invalid output type. Valid types are int32_type and int64_type.");
+        if(inputs.back().ndim() < 1)
+            MIGRAPHX_THROW("Multinomial: Second input shape (sample) has no dimensions");
```
CharlieL7, line 94: Do we have test coverage for this?

CharlieL7 (commented):
Looks good overall; needs some small fixes

src/onnx/parse_multinomial.cpp
bpickrel: Do you want to commit that to this PR, or I do it?  It won't pass tests without the fix.

src/include/migraphx/op/multinomial.hpp
bpickrel: Added.

src/onnx/parse_multinomial.cpp
bpickrel: Fixed

src/include/migraphx/op/random_uniform.hpp
bpickrel: Done

src/onnx/parse_multinomial.cpp
CharlieL7: Working on it and an enhancement

causten:
Needs #2362  to be merged first

src/onnx/parse_multinomial.cpp
bpickrel: This is now merged.

CharlieL7 (approved):

PR #2011, opened 2023-07-28
Change check_shapes to templated class

CharlieL7:
* Should work for iterators or pointers now
* Removed asserts because they do not work for iterators


CharlieL7:
I noticed a possible problem if it deduces the class template type as a shape. Going to make an update

src/include/migraphx/check_shapes.hpp
```diff
@@ -34,29 +34,30 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+template <class T>
```
pfultz2, line 37: It would be better to name this `Iterator` instead of just `T`.

src/include/migraphx/check_shapes.hpp
```diff
@@ -351,17 +340,13 @@ struct check_shapes
     {
         if(begin == end)
             return false;
-        assert(begin != nullptr);
-        assert(end != nullptr);
         return std::any_of(begin, end, p);
     }
 
     const shape* get(long i) const
```
pfultz2, line 346: This needs to be update to use the iterator type.

CharlieL7: No compatible check for both std::iterator and a pointer

src/include/migraphx/check_shapes.hpp
```diff
@@ -34,9 +34,23 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+// Check that deduced type is incrementable, dereferencable, and comparable
+// Should limit valid types to iterators or pointers
+template<typename, typename = void>
+struct is_pointer_or_iterator{};
+ 
+template<typename T>
```
pfultz2: Use class to be consistent with the other template parameters.

src/include/migraphx/check_shapes.hpp
```diff
@@ -34,9 +34,23 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
+// Check that deduced type is incrementable, dereferencable, and comparable
+// Should limit valid types to iterators or pointers
+template<typename, typename = void>
+struct is_pointer_or_iterator{};
```
pfultz2: This should just be named `is_iterator` since a pointer is an iterator.

src/include/migraphx/check_shapes.hpp
```diff
@@ -34,9 +34,23 @@
...
+    T,
+    std::void_t<decltype(++std::declval<T&>()),
+                decltype(*std::declval<T&>()),
+                decltype(std::declval<T&>() == std::declval<T&>())
+    >> : std::true_type {};
+
 template <class T>
 struct check_shapes
 {
+    static_assert(is_pointer_or_iterator<T>::value, "CHECK_SHAPES: Must be a pointer or iterator");
```
pfultz2: Use `{}` instead of `::value`.

src/include/migraphx/check_shapes.hpp
pfultz2: What check are you talking about?

pfultz2: Also, I am not sure how this would compile, because `begin+i` would not be a pointer to shape if `begin` is an iterator.

CharlieL7: I think we need to differentiate between `std::iterator` and the concept of an iterator. `begin + i` will only work for a random_access_iterator. For that case it will return another `std::iterator` object.

src/targets/cpu/gemm.cpp
```diff
@@ -43,7 +43,10 @@ struct dnnl_gemm : dnnl_extend_op<dnnl_gemm, dnnl::matmul, op::dot>
                 MIGRAPHX_DNNL_PREFIX(ARG_BIAS)};
     }
 
-    void required(const check_shapes& cs) const { cs.not_broadcasted(); }
+    void required(const check_shapes<std::vector<shape>::const_iterator>& cs) const
```
pfultz2, line 46: It would be better to template this:

```cpp
template<class T>
void required(const check_shapes<T>& cs) const
```

src/targets/cpu/include/migraphx/cpu/dnnl.hpp
```diff
@@ -400,7 +400,10 @@ struct dnnl_extend_op : dnnl_op<Derived, Primitive>
     }
 
     // dnnl has some issues with non-packed inputs
-    void required(const check_shapes& cs) const { cs.packed_or_broadcasted(); }
+    void required(const check_shapes<std::vector<shape>::const_iterator>& cs) const
```
pfultz2, line 403: Same here.

src/targets/cpu/gemm.cpp
CharlieL7: Would have to make a header file to do that, are we OK with that?

pfultz2: No need to make a header, this is not used directly outside of the cpp file.

umangyadav (approved):

src/include/migraphx/check_shapes.hpp
```diff
@@ -34,29 +34,45 @@
...
+                   std::void_t<decltype(++std::declval<T&>()),
+                               decltype(*std::declval<T&>()),
+                               decltype(std::declval<T&>() == std::declval<T&>())>> : std::true_type
+{
+};
+
+template <class Iterator>
 struct check_shapes
 {
-    const shape* begin;
```
umangyadav, line 39: Is there a guarantee that Iterators would be `const iterators` ?

pfultz2 (approved):

PR #2018, opened 2023-07-31
Add cmake logic to look for local gfx arch

### LINKED ISSUE #1906: Create cmake script for target identification
stop using hip-config.cmake and capture the GPU_TARGET option from what is on the system

umangyadav:
Fixes #1906  

Makes use of `rocm_agent_enumerator`.  It is not available on windows. also looks like there is a way for users to actually override what `rocm_agent_enumerator` returns by setting `ROCM_TARGET_LIST` or by modifying `targets.lst` file. 

modified a bit https://github.com/RadeonOpenCompute/rocm-cmake/blob/f0717bc1a0182b2ddc9194e3144972ab58a88e99/share/rocm/cmake/ROCMLocalTargets.cmake#L16C57-L16C57


examples/migraphx/custom_op_hip_kernel/CMakeLists.txt
```diff
@@ -20,18 +20,62 @@
...
+                    list(APPEND result "${agent}")
+                endif()
+            endforeach()
+
+            if(result)
+                set(${VARIABLE} "${result}" PARENT_SCOPE)
+            endif()
+        endif()
+    endif()
+endfunction()
```
pfultz2, line 55: Workarounds like this should not go into examples. Users will copy and use this after it no longer is needed.

src/targets/gpu/CMakeLists.txt
```diff
@@ -237,6 +238,19 @@ else()
...
+if(NOT WIN32)
+    include(MigxLocalTarget)
+    rocm_local_targets_migx(MIGXGPU_TARGET)
+    message(STATUS "Compiling for gfx arch: ${MIGXGPU_TARGET}")
+
+    foreach(agent IN LISTS MIGXGPU_TARGET)
+        list(APPEND offload_arch_flags "--offload-arch=${agent}")
+    endforeach()
+
+    target_compile_options(migraphx_gpu PRIVATE ${offload_arch_flags})
```
pfultz2, line 250: `migraphx_gpu` does not have device code.

src/targets/gpu/CMakeLists.txt
```diff
@@ -237,6 +238,19 @@ else()
     endif()
 endif()
 
+if(NOT WIN32)
+    include(MigxLocalTarget)
+    rocm_local_targets_migx(MIGXGPU_TARGET)
+    message(STATUS "Compiling for gfx arch: ${MIGXGPU_TARGET}")
+
+    foreach(agent IN LISTS MIGXGPU_TARGET)
+        list(APPEND offload_arch_flags "--offload-arch=${agent}")
```
pfultz2, line 247: Wont this conflict with `GPU_TARGETS`?

pfultz2 (changes requested):

src/targets/gpu/CMakeLists.txt
umangyadav: what's the difference between AMDGPU_TARGETS or GPU_TARGETS ?

src/targets/gpu/CMakeLists.txt
umangyadav: removed.

examples/migraphx/custom_op_hip_kernel/CMakeLists.txt
umangyadav: Okay, for this example then we can just pass `-DAMDGPU_TARGET` manually then ?

cmake/MigxLocalTarget.cmake
```diff
@@ -0,0 +1,58 @@
...
+            foreach(agent IN LISTS _rocm_agents)
+                if(NOT agent STREQUAL "gfx000")
+                    list(APPEND result "${agent}")
+                endif()
+            endforeach()
+
+            if(result)
+                set(${VARIABLE} "${result}" PARENT_SCOPE)
+            endif()
+        endif()
```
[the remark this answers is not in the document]
umangyadav: this PR is temporary local fix in migraphx. it may not work generally in all cases. e.g. i mentioned it wouldn't work on windows etc.

pfultz2:
Does setting `GPU_TARGETS` to `native` before `find_package(hip)` also fix this issue? That seems simpler.

```cmake
set(GPU_TARGETS "native" CACHE STRING "GPU targets to compile for")
find_package(hip)
```

examples/migraphx/custom_op_hip_kernel/CMakeLists.txt
pfultz2: > Okay, for this example then we can just pass -DAMDGPU_TARGET manually then ?

You would pass `-DGPU_TARGETS=...`.

cmake/MigxLocalTarget.cmake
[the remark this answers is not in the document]
pfultz2: > this PR is temporary local fix in migraphx.

Then it should go into the 5.7 release branch.

examples/migraphx/custom_op_hip_kernel/CMakeLists.txt
umangyadav: Okay. Modified README.md to reflect that.

cmake/MigxLocalTarget.cmake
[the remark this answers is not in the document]
umangyadav: > Then it should go into the 5.7 release branch. 

When we move CI to 5.7 then this would be necessary to pull into develop. also not sure exactly when the required changes would land inside the mainline after 5.7 release. So this is necessary for the master branch as well.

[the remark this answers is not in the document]
pfultz2: > When we move CI to 5.7 then this would be necessary to pull into develop.

We should just update jenkins file to add `-DGPU_TARGETS` cmake flag.

> also not sure exactly when the required changes would land inside the mainline after 5.7 release.

This is a bug in hip, so its an issue they need to get fixed as soon as possible. I just dont understand why we are adding workarounds for bugs that hip needs to fix. 

> So this is necessary for the master branch as well.

So its not a temporary fix. Then I think I would prefer to only enable this when the env var `MIGRAPHX_ENABLE_WORKAROUNDS_FOR_BROKEN_HIP_CMAKE` is set. If `GPU_TARGETS` is set to an empty string after calling `find_package(hip)` then we could print a message for the user to set this variable to build migraphx for their gpu.

src/targets/gpu/CMakeLists.txt
pfultz2: `AMDGPU_TARGETS` is used by many math libraries, but its deprecated and undocumented. The documented variable is `GPU_TARGETS`.

cmake/MigxLocalTarget.cmake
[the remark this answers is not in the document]
umangyadav: > Then it should go into the 5.7 release branch.

changed base branch go rocm-5.7

src/targets/gpu/CMakeLists.txt
```diff
@@ -21,8 +21,18 @@
...
 # ####################################################################################
-
 list(APPEND CMAKE_PREFIX_PATH /opt/rocm)
+
+if(NOT WIN32)
+    include(MigxLocalTarget)
+    rocm_local_targets_migx(MIGXGPU_TARGETS)
+    if(MIGXGPU_TARGETS)
+        set(GPU_TARGETS ${MIGXGPU_TARGETS} CACHE STRING "GPU targets to compile for")
+        set(AMDGPU_TARGETS ${MIGXGPU_TARGETS} CACHE STRING "AMDGPU targets to compile for")
```
pfultz2, line 31: Dont set this variable. Its not necessary since you already setting `GPU_TARGETS`.

umangyadav: Done.

pfultz2 (approved):

Issue #1809, opened 2023-06-05
bug in driver argument parser


kahmed10:
Although this isn't a driver read command, the following will hang:
```
./bin/driver read --model resent50 --enable-offload-copy
```
The reason it's hanging is because the argument parser is trying to spellcheck `--enable-offload-copy`. That uses the levenshtein distance algorithm to find the closest similar flag name, but looks like it's getting stuck in infinite recursion.

I was able to narrow down the flag where it's getting stuck, and it's comparing `--migraphx-json` with `--enable-offload-copy`.

lakhinderwalia:
The first problem is, in runaction() the argument is actually taken as a file name -- the file is not found in the system, of course. Then the exception that is thrown is effectively being ignored. It shouldn't be. So let us first fix that, and see what more is needed.. 

BTW, after ignoring the exception, the spellcheck part is invoked in run_action()..

lakhinderwalia:
_(After a small code change to intercept the runtime error ..)_

The command `./bin/driver read --model resent50 --enable-offload-copy` has now 3 scenarios to consider:

-  If exactly the above command is executed:
```
./bin/driver read --model resent50 --enable-offload-copy
error: Path does not exists: --enable-offload-copy
terminate called after throwing an instance of 'migraphx::version_1::exception'
  what():  /home/lakhindr/Latest/AMDMIGraphX/src/driver/main.cpp:301: load: Unknown model: resent50
Aborted (core dumped)
```

-  If the `resent50` is typed as `resnet50`, the command first prints an error: 
`the path does not exist: --enable-offload-copy`. 
And then goes on to dump a fairly long list of lines for presumably some resnet50 model.. 

- Even if the model option isn't given:
```
./bin/driver read  --enable-offload-copy
error: Path does not exists: --enable-offload-copy
Reading:
terminate called after throwing an instance of 'std::length_error'
  what():  cannot create std::vector larger than max_size()
Aborted (core dumped)
```

pfultz2:
I do see three issues that do need to be fixed:

* Add validation for `--model` flag so we can catch an invalid model name during parse time instead of runtime
* Update `levenshtein_distance` algorithm on larger strings to use an O(n^2) algorithm
* If the argument starts with `-` print an error about non-existing flag when `levenshtein_distance` can't find a flag thats close instead of printing the error message from the exception(which is probably confusing).

lakhinderwalia:
[...]

PR #2020, opened 2023-08-01
Bug in driver argument parser. Fix Levenshtein_distance.

lakhinderwalia:


src/driver/argument_parser.hpp
```diff
@@ -590,6 +590,12 @@ struct argument_parser
                 v(*this, inputs);
             return arg.action(*this, inputs);
         }
+        catch(const std::runtime_error& e)
+        {
+            msg = e.what();
+            std::cerr << "error: " << msg << std::endl;
+            return false;
+        }
```
pfultz2, line 598: This is wrong as this will bypass all the spellcheking. For example, running `./bin/driver read -h2` should output:

```
error: Found argument '-h2' which wasn't expected, or isn't valid in this context
       Did you mean -h?

USAGE:
    ./bin/driver read -h
```

However, with this change it would skip the spellchecking and then continue to try to run the command(because you are returning `false`), So it would have this output:

```
error: Path does not exists: -h2
Reading:
terminate called after throwing an instance of 'std::length_error'
  what():  cannot create std::vector larger than max_size()
Aborted (core dumped)
```

Also, the `runtime_error` exception is already caught by `catch(const std::exception& e)` below, so we aren't ignoring the exception.

lakhinderwalia: @pfultz2 thanks. I don't understand the flow here, so please do explain what you find wrong. But in my mind, if a file isn't found, and an exception is raised it is getting ignored in the (prior) exception catchers.. 
Do we still want a spellcheck when a runtime exception has been raised..?

src/driver/argument_parser.hpp
```diff
@@ -590,6 +590,12 @@ struct argument_parser
...
         catch(const std::exception& e)
         {
             msg = e.what();
```
lakhinderwalia, line 601: This clause is simply storing the exception type. Effectively ignoring it..

src/driver/argument_parser.hpp
pfultz2: > But in my mind, if a file isn't found, and an exception is raised it is getting ignored in the (prior) exception catchers..

I am not sure I follow, there is only one catch handler. What flags are you passing to the driver? And what is the expected output you would like?

>  Do we still want a spellcheck when a runtime exception has been raised..?

Yes. Otherwise we can't spellcheck incorrect arguments, like I showed in the example above.

src/driver/argument_parser.hpp
pfultz2: It doesn't ignore. We print the message from the exception below:

```cpp
        try
        {
            ...
        }
        catch(const std::exception& e)
        {
            msg = e.what();
        }
        catch(...)
        {
            msg = "unknown exception";
        }
        std::cout << color::fg_red << color::bold << "error: " << color::reset;
        auto sc = spellcheck(inputs);
        if(sc.distance < 5)
        {
            ...
        }
        else
        {
            const auto& flag_name = flag.empty() ? arg.metavar : flag;
            std::cout << "Invalid input to '" << color::fg_yellow;
            std::cout << arg.usage(flag_name);
            std::cout << color::reset << "'" << std::endl;
            std::cout << "       " << msg << std::endl; // <- Print the error message from the exception
            std::cout << std::endl;
            print_usage({arg.usage()});
        }
```

We then return `true` which means to not continue executing the command.

src/driver/argument_parser.hpp
lakhinderwalia: @pfultz2 PTAL at https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1809 it has all the details of the exact command that I typed. I can also paste it here..

"And what is the expected output you would like?". No, I am not expecting anything. I just don't want to break anything. 

This code that I changed is just to show the 3 error cases. I don't mean it to be final, but the exception handling has to change a little bit..? I need clarity on what is to be done here.. Thanks.

pfultz2: >  don't mean it to be final, but the exception handling has to change a little bit..?

The exception handling shouldn't need to be changed.

lakhinderwalia:
_With the latest check-in_

- the Levenshtein distance calculation has been fixed with another (non-recursive) API.
- error cases are handled as below:

**_error case_**
```
bin/driver  read  --model resent50
error: Invalid input to '--model [resnet50|inceptionv3|alexnet]'
       Invalid argument: resent50. Valid arguments are {alexnet, inceptionv3, resnet50}

USAGE:
    bin/driver read --model [resnet50|inceptionv3|alexnet]


For more information try '--help'
```
_**error case**_
```
bin/driver  read  --model resnet50 --enable-offload-copy
error: Invalid input to '<input file> [std::string]'
       Path does not exist: --enable-offload-copy

USAGE:
    bin/driver read <input file>


For more information try '--help'
```
_**error case**_

```
 bin/driver  read  --model resent50 --enable-offload-copy
error: Invalid input to '<input file> [std::string]'
       Path does not exist: --enable-offload-copy

USAGE:
    bin/driver read <input file>


For more information try '--help'
```

pfultz2:
LGTM, just need to fix the tidy errors.

src/include/migraphx/algorithm.hpp
```diff
@@ -90,6 +90,27 @@ levenshtein_distance(Iterator1 first1, Iterator1 last1, Iterator2 first2, Iterat
     return std::ptrdiff_t{1} + std::min({x1, x2, x3});
 }
 
+inline size_t levenshtein_distance(const std::string& s1, const std::string& s2)
```
kahmed10, line 93: I don't see the previous implementation being used anywhere. Maybe we should remove it.

umangyadav (approved):

umangyadav:
> bin/driver  read  --model resnet50 --enable-offload-copy
error: Invalid input to '<input file> [std::string]'
       Path does not exist: --enable-offload-copy

USAGE:
    bin/driver read <input file>


For more information try '--help' 

IMO, For this case it should just ignore `--enable-offload-copy` and print a message about ignoring, that flag is only useful for compiling not reading the model.

lakhinderwalia:
> IMO, For this case it should just ignore `--enable-offload-copy` and print a message about ignoring, that flag is only useful for compiling not reading the model.

You can pass anything instead of this flag, actually. Say `xxxxyyyz` or `--zssyyy`, and it will give the same results; That flag isn't recognized in any manner here, as I understand it.

src/include/migraphx/algorithm.hpp
lakhinderwalia: OK. Agreed. May I move this new algorithm to the `src/driver` subdirectory as well? Thanks.

kahmed10: It was originally designed to be here because of possible reuse in other situations. Can't think of any right now, but probably not worth moving just yet.

umangyadav (approved):

kahmed10 (approved):
Can we rename the pull request, otherwise LGTM

src/include/migraphx/algorithm.hpp
```diff
@@ -95,20 +95,24 @@ inline size_t levenshtein_distance(const std::string& s1, const std::string& s2)
     const size_t l1 = s1.length();
     const size_t l2 = s2.length();
 
-    std::vector<std::vector<size_t>> d(l1 + 1, std::vector<size_t>(l2 + 1));
+    std::vector<size_t> d0(l2 + 1), d1(l2 + 1);
```
umangyadav: Should one of them be `l1` here ?

lakhinderwalia: No, this is correct. It is a modified algorithm. These are two identical sized vectors pertaining to just one of those strings: `s2`. Then we iterate through them in an inner loop (which is covered inside the outer loop by an index counter `1 .. l1`). Thanks.

src/include/migraphx/algorithm.hpp
```diff
@@ -95,20 +95,24 @@ inline size_t levenshtein_distance(const std::string& s1, const std::string& s2)
...
 
-    for(size_t i = 1; i <= l1; i++)
-        d[i][0] = i;
     for(size_t j = 1; j <= l2; j++)
-        d[0][j] = j;
+        d0[j] = j;
```
pfultz2, line 102: Use `std::iota`.

lakhinderwalia: We are explicitly using the indices here inside the loop. Those index values are very important and therefore doing a duplicate index calculation in iota controlled loops isn't most efficient here.. -- unless I have missed something. We need `i, i-1, j, j-1.`  in these two (nested) for loops. Thanks.

src/include/migraphx/algorithm.hpp
```diff
@@ -95,20 +95,24 @@ inline size_t levenshtein_distance(const std::string& s1, const std::string& s2)
...
 
     for(size_t i = 1; i <= l1; i++)
+    {
+        d1[0] = i;
```
pfultz2, line 106: I dont think you need a seperate vector for this. You could just store the previous distance in a variable and read that:

```cpp
for (size_t i = 1; i <= l1; i++)
{
    size_t previous = 0;
    for (size_t j = 1; j <= l2; j++)
    {
        size_t cost_insert_or_delete = std::min(d0[j - 1] + 1, d0[j] + 1);
        size_t cost_substitute       = previous + (s1[i - 1] == s2[j - 1] ? 0 : 1);
        previous                     = d0[j];
        d0[j]                        = std::min(cost_substitute, cost_insert_or_delete);
    }
}
```

It seems simpler than swapping the vector at each loop.

lakhinderwalia: > It seems simpler than swapping the vector at each loop.

@pfultz2, I am not so sure if a vector swap (which is effectively just an exchange of its internal buffer pointers, and is performed outside the loop) is any less simpler than some operation inside the inner loop?  That algorithm original looks more understandable, and is intuitively what the flow-chart would be. 

I grant you it takes two vectors instead of one. 

But, if you strongly dislike it, I can change it, else leave it as it is for algorithmic simplicity/clarity.. I don't think performance is an issue here -- well it is in the original algorithm.. N**3 vs N**2.

src/include/migraphx/algorithm.hpp
pfultz2: You are setting the vector to `0,1,2,3,4,...` in this loop, so doing `std::iota(d1.begin(), d1.end(), 0)` is equivalent. We [prefer standard algorithms over raw loops](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/wiki/Developer's-Guide#raw-loops) when its available.

lakhinderwalia: (Totally agreed with your sentiment.., but.. )

I wouldn't (logically speaking) set an element of `vector<size_t>` to a `value type`: `vector<size_t>::iterator`, line # 106:
`d1[0] = i;`
The index `i` is also the cost of levenshtein_distance (boundary condition).
Thanks.

src/include/migraphx/algorithm.hpp
pfultz2: > is performed outside the loop)

Its not performed in the inner loop but it is still performed in the outer loop. 

> That algorithm original looks more understandable, and is intuitively what the flow-chart would be.

It took quite bit of work to realize that you are just using the previous distance. Its not very obvious. Maybe that matches the pseudo algorithm on wikipedia, but using the `previous` variable makes what the algorithm is actually doing in the code much clearer. Even rosetta code uses only one array(although the implementation is different):

https://rosettacode.org/wiki/Levenshtein_distance#Generic_ISO_C++_version

> But, if you strongly dislike it, I can change it,

I would prefer to use a single vector unless my analysis of the loop is incorrect(which is quite possible as well).

lakhinderwalia:
The single vector approach has been checked in. This replaces the previous N-sq 2-vector approach with another N-sq single-vector approach, so we are all good -- because it replaces the original N-cube or worse(?) approach. Some extra minor additional enhancements have also been added: to use the minimal of the two strings for buffer. Or to bypass some extra computations when current characters in two strings match.

_Trivia for the hyper-critical who would compare the single-vector approach vs two-vector approach:_ Single-vector approach comes with a slightly more complex (and definitely non-intuitive) handling of computation with-in the inner for-loop, compared to the 2-vector approach it replaces. Also, performance wise, and this is just the CPU cycles we are talking of: the previous approach of using a vector swap (in 2-vector case) outside that inner loop was arguably more efficient if one were to compare the extra cost of _repeatedly_ swapping variables in that inner loop, simply because _one_ vector swap is a simple _move_ operation in lieu of all that swapping N times - It is not an element-by-element swap. So, to have a very long string of N characters does N extra jugglery steps in the new approach, whereas the previous alternative was to avoid it all with just one swap (a move in reality) step outside the loop. 

_Conclusion: the good thing is, both these approaches were order N-square, so in most testing they should come put pretty much the same is my guess._

pfultz2 (approved):

PR #2037, opened 2023-08-03
Enable MLIR by default in migraphx builds

pfultz2:


rbuild.ini
```diff
@@ -19,6 +19,7 @@ define =
 [develop]
 cxx = ${rocm_path}/llvm/bin/clang++
 cc = ${rocm_path}/llvm/bin/clang
+ignore = ROCmSoftwarePlatform/rocMLIR
```
umangyadav, line 22: Why does this one ignore it ? I use `rbuild develop` to install dependencies.

pfultz2: Yea this is the wrong place. It should be under the `gh` session.

umangyadav:
what's the purpose behind having it enabled by default ? 
I see that we already have Jenkins job that builds MIGraphX with MLIR.

Dockerfile
```diff
@@ -77,6 +77,9 @@ ADD dev-requirements.txt /dev-requirements.txt
 ADD requirements.txt /requirements.txt
 ADD rbuild.ini /rbuild.ini
 
+# Temporarily install a new cmake until switching to ubuntu 22.04
+RUN pip3 install cmake==3.22.1
```
causten, line 81: ORT needs 3.26.4 would it make sense to bump this to 3.26.4 too?

pfultz2: no, we should make sure our stuff builds with the cmake in ubuntu 22.04.

causten (approved):
builds and runs as expected.

pfultz2:
The tidy error requires a fix in rocm-cmake: https://github.com/RadeonOpenCompute/rocm-cmake/pull/140

umangyadav (approved):

PR #2038, opened 2023-08-04
Change auto contiguous to always insert contiguous

kahmed10:


src/auto_contiguous.cpp
```diff
@@ -33,43 +33,60 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+        });
+
+        if(new_args != args)
         {
-            auto c = m.insert_instruction(std::next(ins), make_op("contiguous"), ins);
-            m.replace_instruction(ins, c);
+            if(not ins->module_inputs().empty())
+                m.replace_instruction(ins, ins->get_operator(), new_args, ins->module_inputs());
+            else
+                m.replace_instruction(ins, ins->get_operator(), new_args);
```
pfultz2, line 82: You shouldn't replace the input arguments as that can cause multiple contiguous operators to be inserted when its used in multiple places. This should just do:

```cpp
if(contains({"layout", "contigous"}, ins->name()))
    continue;
// for last instruction that is NOT a return
if(ins->outputs().empty() and ins != last)
    continue;
shape s = ins->get_shape();
if(s.dynamic())
    continue;
if(s.type() == shape::tuple_type)
    continue;
if(s.standard() and ins->name() == "@literal")
    continue;
auto c = m.insert_instruction(std::next(ins), make_op("contiguous"), ins);
m.replace_instruction(ins, c);
```

src/auto_contiguous.cpp
```diff
@@ -33,43 +33,60 @@ inline namespace MIGRAPHX_INLINE_NS {
...
-        {
-            auto args     = ins->inputs();
-            auto new_args = args;
-            std::transform(args.begin(), args.end(), new_args.begin(), [&](auto in) {
-                if(in->name() == "contiguous")
-                {
-                    return in;
-                }
-                return m.insert_instruction(ins, make_op("contiguous"), in);
-            });
```
pfultz2, line 50: We should still have this until reshape copy is implemented.

src/auto_contiguous.cpp
```diff
@@ -59,17 +60,24 @@ void auto_contiguous::apply(module& m) const
     auto last = std::prev(m.end());
     for(auto ins : iterator_for(m))
     {
-        if(ins->name() == "layout")
+        if(contains({"layout", "contiguous", "@return", "@param", "@outline"}, ins->name()))
```
pfultz2, line 63: We shouldn't be skipping `@param`, is there a test that fails for that?

I dont think `@outline` is used anywhere.

test/auto_contiguous_test.cpp
```diff
@@ -174,8 +176,11 @@ TEST_CASE(standard_reshape)
         auto data = m2.add_parameter("2x2", {migraphx::shape::float_type, {2, 3, 4, 5}});
         auto add  = m2.add_instruction(migraphx::make_op("add"), data, data);
         auto ca   = m2.add_instruction(migraphx::make_op("contiguous"), add);
-        auto r    = m2.add_instruction(migraphx::make_op("reshape", {{"dims", {2, 1, 12, 5}}}), ca);
-        m2.add_return({r});
+        // extra contiguous coming from reshape logic which has "requires_std_shape" attribute
+        auto cb = m2.add_instruction(migraphx::make_op("contiguous"), ca);
```
pfultz2, line 180: We shouldnt have two contiguous. Maybe we should check that the outputs are all `contiguous` operators before inserting it in.

src/auto_contiguous.cpp
```diff
@@ -59,17 +60,24 @@ void auto_contiguous::apply(module& m) const
...
             continue;
         shape s = ins->get_shape();
-        if(not s.dynamic() and not s.standard() and s.elements() != 0)
+        if(s.dynamic())
+            continue;
+        if(s.type() == shape::tuple_type)
+            continue;
+        if(s.standard() and ins->name() == "@literal")
+            continue;
+        if(s.scalar() and not contains(ins->name(), "broadcast"))
```
pfultz2, line 75: I would think this would be `if(s.scalar() and s.standard())`.

TedThemistokleous:
Why close this out? Guessing rebase and push forward or not needed anymore?

PR #2039, opened 2023-08-04
Variable input slice

CharlieL7:
* Depends on #2011 
* Allows slice to work with variable `starts`, `ends`, and `axes` input
* Outputs a dynamic shape even with a static shape data input when the `starts` and `ends` are variable
* Still does aliasing of the buffer during compute()


src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,141 @@ struct slice
...
+                                                   norm_inputs.at("input_starts"),
+                                                   norm_inputs.at("input_ends"),
+                                                   norm_inputs.at("input_axes")),
+                                         input_shape.strides()};
+                });
+            return {output_shape, [=] { return input.data() + offset; }};
+        }
+        default: {
+            // Should never get here; covering in case some code change occurs
+            MIGRAPHX_THROW("SLICE: invalid number of inputs");
```
TedThemistokleous, line 323: Good idea on this also with the comment.

TedThemistokleous (approved):
Fix CI otherwise looks good.

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,141 @@ struct slice
...
+                               index = 0;
+                           }
+                           else if(index > dim)
+                           {
+                               index = dim;
+                           }
+                           return index;
+                       });
+        return indices;
+    }
```
pfultz2, line 243: A function like this should go into normalize_attributes.hpp and it should reuse the`tune_attribute` function we have. I would envision a function like this:

```cpp
auto normalize_indices(const std::vector<int64_t> &indices,
                       const std::vector<int64_t> &axes,
                       const shape &input_shape,
                       const std::vector<op::normalize_attribute> &attributes);
``` 

This way we can use the same attributes that we use for the static case.

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,141 @@ struct slice
...
+    {
+        auto input       = args[0];
+        auto input_shape = input.get_shape();
+        std::size_t offset;
+        switch(args.size())
         {
-            return shape{t, new_mins, new_lens, {}};
+        case 1: {
+            offset = compute_offset(input_shape);
+            return {normalize_compute_shape({input_shape}), [=] { return input.data() + offset; }};
```
pfultz2, line 288: Why doesnt this use the `output_shape` that got passed in?

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,141 @@ struct slice
...
+                auto norm_inputs = normalize_inputs(input_shape, input_starts, input_ends);
+                offset = compute_offset(input_shape, norm_inputs.at("input_starts"), this->axes);
+                output_shape = {input_shape.type(),
+                                lens_calc(input_shape.lens(),
+                                          norm_inputs.at("input_starts"),
+                                          norm_inputs.at("input_ends"),
+                                          this->axes),
+                                input_shape.strides()};
+            });
+            return {output_shape, [=] { return input.data() + offset; }};
```
pfultz2, line 302: I think it would be better to remove these returns outside of the switch statement. `output_shape` and `offset` are already defined in the outerscope.

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,141 @@ struct slice
...
+        return {{"input_starts", normalize_indices(input_shape, input_starts, input_axes)},
+                {"input_ends", normalize_indices(input_shape, input_ends, input_axes)},
+                {"input_axes", input_axes}};
+    }
+
+    argument compute(const shape&, std::vector<argument> args) const
+    {
+        auto input       = args[0];
+        auto input_shape = input.get_shape();
+        std::size_t offset;
```
pfultz2, line 283: This should be initialized to zero.

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,141 @@ struct slice
...
+                           auto dim = input_shape.lens().at(axis);
+                           if(index < 0)
+                           {
+                               index += dim;
+                           }
+                           if(index < 0)
+                           {
+                               index = 0;
+                           }
+                           else if(index > dim)
```
bpickrel, line 236: Didn't it used to be legal to give an index larger than dim?  Why the change?

src/include/migraphx/op/slice.hpp
CharlieL7: That's possible but very messy to reuse tune_attribute. It  would also have to be a template because the types of `indices` and `in_axes` are not necessarily the same.

src/include/migraphx/op/slice.hpp
CharlieL7: It can if we add a conditional on if the shape is dynamic to recompute

src/include/migraphx/op/slice.hpp
CharlieL7: What do you mean here? Output shape is not defined in the outer scope

src/include/migraphx/op/slice.hpp
CharlieL7: I don't think we want to do this with how operation.hpp and normalize_attribute.hpp is set up. @pfultz2 What do you think?

src/include/migraphx/op/slice.hpp
CharlieL7: I kept the returns because there would have been `break` statements otherwise.

src/include/migraphx/op/slice.hpp
pfultz2: > That's possible but very messy to reuse tune_attribute.

It shouldn't be messy. I think an implementation like this would work:

```cpp
std::vector<int64_t>
normalize_indices(const std::vector<int64_t> &indices,
                  const std::vector<int64_t> &axes, const shape &input_shape,
                  const std::vector<op::normalize_attribute> &attributes)
{
  return tune_attribute(indices, axes, attributes, input_shape,
                        [] { return ""; });
}
```

In the future we could consider changing `tune_attribute` to take `std::vector<op::normalize_attribute>` instead of `value`, but that doesn't need to be addressed in this API. 

To improve the error message we could add a another `prefix` parameter:

```cpp
std::vector<int64_t>
normalize_indices(const std::vector<int64_t> &indices,
                  const std::vector<int64_t> &axes, const shape &input_shape,
                  const std::vector<op::normalize_attribute> &attributes,
                  const std:string& prefix = "")
{
  return tune_attribute(indices, axes, attributes, input_shape,
                        [&] { return prefix; });
}
```


> It would also have to be a template because the types of indices and in_axes are not necessarily the same.

No need to template it, we can just convert them to `std::vector<int64_t>` to call the function.

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,149 @@ struct slice
...
         {
-            auto axis            = axes[i];
-            size_t sliced_length = ends[i] - starts[i];
-            // A Numpy indexing convention: a slice size larger than the actual dimension
-            // is legal and the "ends" value is clipped to the axis size
-            new_lens[axis] = std::min(new_lens[axis], sliced_length);
-            if(input_shape.dynamic())
+        case 1: {
+            offset = compute_offset(input_shape);
+            if(output_shape.dynamic())
```
pfultz2, line 288: Why dont you use `dyn_output`?

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,149 @@ struct slice
...
+        return {{"input_starts", normalize_indices(input_shape, input_starts, input_axes)},
+                {"input_ends", normalize_indices(input_shape, input_ends, input_axes)},
+                {"input_axes", input_axes}};
+    }
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
+    {
+        auto input       = args[0];
+        auto input_shape = input.get_shape();
+        std::size_t offset = 0;
```
pfultz2, line 283: Actually the scope of this variable can be moved into the case statements.

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,149 @@ struct slice
...
+        switch(args.size())
         {
-            auto axis            = axes[i];
-            size_t sliced_length = ends[i] - starts[i];
-            // A Numpy indexing convention: a slice size larger than the actual dimension
-            // is legal and the "ends" value is clipped to the axis size
-            new_lens[axis] = std::min(new_lens[axis], sliced_length);
-            if(input_shape.dynamic())
+        case 1: {
+            offset = compute_offset(input_shape);
```
pfultz2, line 287: `std::size_t offset = compute_offset(input_shape);`

src/include/migraphx/op/slice.hpp
```diff
@@ -90,80 +190,149 @@ struct slice
...
             }
         }
-        if(input_shape.dynamic())
-        {
-            return shape{t, new_mins, new_lens, {}};
+        case 3: {
+            shape calc_shape;
+            visit_all(args[1], args[2])([&](auto input_starts, auto input_ends) {
+                auto norm_inputs = normalize_inputs(input_shape, input_starts, input_ends);
+                offset = compute_offset(input_shape, norm_inputs.at("input_starts"), this->axes);
```
pfultz2, line 302: `std::size_t offset = compute_offset(input_shape, norm_inputs.at("input_starts"), this->axes);`

src/onnx/parse_slice.cpp
```diff
@@ -41,9 +41,41 @@ struct parse_slice : op_parser<parse_slice>
                           onnx_parser::node_info info,
                           std::vector<instruction_ref> args) const
     {
+        auto op_tuple = construct_slice_op(parser, info, args);
         op::slice op;
+        std::vector<instruction_ref> op_args;
+        std::vector<int64_t> steps;
+        std::vector<int64_t> raxes;
```
pfultz2, line 48: Make this a struct.

src/onnx/parse_slice.cpp
```diff
@@ -69,8 +111,19 @@ struct parse_slice : op_parser<parse_slice>
...
+                if(not steps.empty())
+                {
+                    MIGRAPHX_THROW(
+                        "PARSE_SLICE: steps and variable starts and ends is not supported");
+                }
+            }
+            else
+            {
+                end_arg.visit([&](auto s) { op.ends.assign(s.begin(), s.end()); });
+            }
```
pfultz2, line 126: When you make the return a struct you can add a method to do this pattern:

```cpp
struct slice_desc
{
    std::vector<instruction_ref> op_args;
    std::vector<int64_t> steps;
    std::vector<int64_t> raxes;

    std::vector<int64_t> insert(instruction_ref arg)
    {
        std::vector<int64_t> result;
        migraphx::argument arg_value = arg->eval();
        if(arg_value.empty())
        {
            op_args.insert(op_args.begin(), arg);
        }
        else
        {
            arg_value.visit([&](auto s) { result.assign(s.begin(), s.end()); });
        }
        return result;
    }
};
```

Then you can just write the above as `op.ends = sd.insert(args.at(2))`. This should get rid of the extra branching so it will pass tidy.

Also, I think you could check for the `steps` afterwards:

```cpp
if (not steps.empty())
{
    if (op.starts.empty() or op.ends.empty())
        MIGRAPHX_THROW("PARSE_SLICE: steps and variable starts and ends is not supported");
    if (op.axes.empty())
        MIGRAPHX_THROW("PARSE_SLICE: steps and variable axes is not supported");
}
```

src/include/migraphx/op/slice.hpp
pfultz2: Makes sense.

src/include/migraphx/op/slice.hpp
CharlieL7: I tried it out, having issues with converting the `tensor_view` into a `std::vector<int64_t>`:
```
    /**
     * Three input version of the normalize_inputs.
     * This one also checks that the input_axes are valid.
     */
    template <class Inds>
    std::unordered_map<std::string, Inds>
    normalize_inputs(shape input_shape, Inds input_starts, Inds input_ends, Inds input_axes) const
    {
        auto attrs = this->attributes().at("normalize_axes");
        return {{"input_starts", normalize_input(input_starts.to_vector(),
                    this->axes, input_shape, attrs.at("starts"), "Slice variable input_starts")},
                {"input_ends", normalize_input(input_ends.to_vector(),
                    this->axes, input_shape, attrs.at("ends"), "Slice variable input ends")},
                {"input_axes", normalize_input(input_axes.to_vector(),
                    {}, input_shape, attrs.at("axes"), "Slice variable input_axes")}};
    }
```

CharlieL7: Think I can get around it by using `tensor_view.template to_vector<int64_t>`. Similar thing needs to be done for the converting the attributes value.

bpickrel:
Did you use `python3 tools/format.py ` to format the code?  Format fails look like a mismatch with Paul's new script.

CharlieL7:
> Did you use `python3 tools/format.py -i ` to format the code? Format fails look like a mismatch with Paul's new script (?)

Been having weird formatting issues

src/api/migraphx.py
```diff
@@ -46,6 +46,7 @@ def shape_type_wrap(p):
 
 
 def auto_handle(*args, **kwargs):
+
```
pfultz2, line 49: This extra line should be removed.

pfultz2:
There are lot of formatting changes to unrelated files to this PR. Those files should go in a different PR.

src/include/migraphx/op/slice.hpp
```diff
@@ -70,6 +85,90 @@ struct slice
 
     std::string name() const { return "slice"; }
 
+    /**
+     * Computes the slice output shape dimensions for given starts, ends,and axes.
+     * Templated to also handle tensor views.
+     * Possibily different type between in_starts, in_ends and in_axes if in_axes is this object's
```
bpickrel, line 91:
```suggestion
     * Possibly different types for in_starts, in_ends and in_axes if in_axes is this object's
```

bpickrel (approved):
Approved pending fixes to test and tidy fails

test/onnx/gen_onnx.py
```diff
@@ -32,7 +32,9 @@
 
 
 def onnx_test(external_data=False):
+
```
pfultz2, line 35: Yapf is complaining about these extra lines on 35 and 37.

PR #2048, opened 2023-08-06
Add builds for sles in github actions

pfultz2:


pfultz2:
So we need #2063 merged first so we can skip building mlir in sles.

TedThemistokleous (approved):

PR #2050, opened 2023-08-07
Check that migraphx_gpu is not using device compilation in migraphx_gpu

pfultz2:


umangyadav (approved):

src/targets/gpu/no_device.cpp
```diff
@@ -0,0 +1,5 @@
+
```
TedThemistokleous, line 1: Add license file header here.

TedThemistokleous (changes requested):

src/targets/gpu/no_device.cpp
```diff
@@ -0,0 +1,28 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
TedThemistokleous, line 4: make it 2023

TedThemistokleous (approved):

PR #2051, opened 2023-08-07
added rand_uniform operation closes #1958

### LINKED ISSUE #1958: Create rand_uniform operator to support Multinomial op
For the Multinomial op to be able to generate a distribution at runtime, a prerequisite is a `RandomSequence` operation that can generate random numbers on any target device.  The algorithm needs to be the same across devices to make results repeatable for verification testing.  (Currently, multinomial distributions are generated during Onnx parsing and the results loaded as `literal` instructions, but this won't work for dynamic input shapes.)

Multinomial and any other op. using the `RandomSequence` op would need to be able to accept a seed as an argument, to create a repeatable deterministic sequence for testing, and also a way to start with a random (clock-based) seed.  This needs to be able to be done at runtime.  

i.e. foo(shape, seed)
Consider using rocRand for GPU but make sure it uses a random algorithm we can implement everywhere

bpickrel:
New op that populates a shape with random numbers with a uniform distribution.  The `rand_uniform` op. can implement the Onnx RandomUniform instruction, and can also create the random number sequence necessary to implement Multinomial.  (At this time, our Onnx Multinomial parsing generates a random sequence of numbers when parsing as a workaround, so that the resulting program uses the same "random" set every time.)

Arguments:  `shape, seed`.  Shape is required; can be static or dynamic.  Seed is still optional in this version.  If it's not given at inference time, use the value in the creation attribute `seed`.  _Update:  deleted  A boolean `use_auto_seed` causes any given seed to be ignored._

Not implemented yet:  No Onnx parsing of RandomUniform.  Only shape type allowed is float.  Min, max range values are not supported and the range is always (0,1).  Do we want to add any of these features now?


bpickrel:
Significant change in the latest commit:  instead of optionally letting the rand_uniform op. auto-seed at runtime, there is a new op random_seed which generates the auto-seed (also at runtime).  This then can be passed as an argument. This means that a program compiled with a random_seed argument to rand_uniform will never give a repeatable result.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <migraphx/reflect.hpp>
+#include <random>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct rand_uniform
+{
+    uint32_t seed = {0};
```
pfultz2, line 56: There should be no seed attribute. The seed should always come from the input argument.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.dtype, "dtype"), f(self.seed, "seed"));
+    }
+
+    std::string name() const { return "rand_uniform"; }
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this, true}.has(1, 2);
```
pfultz2, line 74: `has(2)`.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+        if(inputs.size() > 1 and inputs.at(1).type() != shape::type_t::uint32_type)
```
pfultz2, line 76: The seed should be `uint64_type`. It also should be the first argument and the output buffer should be the second argument.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+            MIGRAPHX_THROW("RAND_UNIFORM:  Input 2 (seed) must have type unsigned int");
+        auto s = inputs.front();
```
pfultz2, line 78: This should just return the shape of the output buffer.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+        }
+        else
+        {
+            return s.with_lens(s.lens()).with_type(dtype);
+        }
+    }
+
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    {
+        argument result{dyn_out.computed_shape};
```
pfultz2, line 91: You need to use the output buffer passed in.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+        // or the default.
+
+        std::mt19937 gen(local_seed);
+        std::uniform_real_distribution<> dis(range_min, range_max);
+        result.visit([&](auto output) {
+            std::generate(output.begin(), output.end(), [&]() { return dis(gen); });
+        });
+        return result;
+    }
+};
```
pfultz2, line 108: This also needs an `output_alias` method added.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <migraphx/argument.hpp>
+#include <migraphx/par_for.hpp>
+#include <migraphx/reflect.hpp>
+#include <random>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct rand_uniform
```
pfultz2, line 54: I think I would prefer the name `random_uniform`.

src/include/migraphx/op/random_seed.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ *    Generates a random seed for the use of random number generators.  Generating the seed
+ * at runtime guarantees there will be a different random sequence on every execution.
+ * This operation has no inputs or attributes, and outputs an unsigned integer tensor with
+ * a single value.
+ */
+struct random_seed
+{
+    shape::type_t dtype = shape::type_t::uint32_type;
```
pfultz2, line 46: Remove this. The datatype should always be `uint64_t`.

src/include/migraphx/op/random_seed.hpp
```diff
@@ -0,0 +1,79 @@
...
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.dtype, "dtype"));
+    }
+
+    std::string name() const { return "random_seed"; }
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        (void)inputs;
```
pfultz2, line 57: This should check that `inputs` is empty.

src/include/migraphx/op/random_seed.hpp
```diff
@@ -0,0 +1,79 @@
...
+        return migraphx::shape(dtype, {1});
```
pfultz2, line 58: `return shape{shape::uint64_type}`

src/include/migraphx/op/random_seed.hpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
+    {
+        (void)args;
+        argument result(output_shape);
+
+        result.visit([&](auto output) {
+            std::generate(output.begin(), output.end(), [&]() {
```
pfultz2, line 67: There is only one element and this should call `std::random_device` to get a random seed. So this should just be `output.front() = std::random_device{}()`.

src/include/migraphx/op/rand_uniform.hpp
bpickrel: What's the reason behind this?  The STL functions allow a variety of seed types and there seems to be no meaningful possibility of harmful repetition.

src/include/migraphx/op/rand_uniform.hpp
bpickrel: Should be different in what way from what it's doing now?  The only additional processing enforces the data type.

pfultz2: It should just do `return s`, the data type should come from the output buffer not an attribute.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    uint32_t seed = {0};
+
+    // todo:  not currently settable
+    float range_min = 0.0f;
+    float range_max = 1.0f;
+
+    // todo:  integer data type(s) not yet supported
+    shape::type_t dtype = shape::type_t::float_type;
```
pfultz2, line 64: This should be removed. The data type should come from the output buffer not an attribute.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,117 @@
...
+        {
+            return s.with_lens(s.lens()).with_type(dtype);
+        }
+    }
+
+    argument compute(const shape& output, std::vector<argument> args) const
+    {
+        // Output goes into the passed buffer, not the shape output
+        (void)output;
+        argument result{args[1].get_shape()};
```
pfultz2, line 98: Need to use the output buffer as the result: `auto result = args[1]`.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,117 @@
...
+        else
+        {
+            return s.with_lens(s.lens()).with_type(dtype);
+        }
+    }
+
+    argument compute(const shape& output, std::vector<argument> args) const
+    {
+        // Output goes into the passed buffer, not the shape output
+        (void)output;
```
pfultz2, line 97: Remove parameter name instead of adding void statements.

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,117 @@
...
+        argument result{args[1].get_shape()};
+
+        uint32_t local_seed = args[0].at<uint32_t>(0);
+
+        std::mt19937 gen(local_seed);
+        std::uniform_real_distribution<> dis(range_min, range_max);
```
pfultz2, line 103: You need to use `uniform_int_distribution` for when its an integer. Ideally this should probably be moved into the visitor:

```cpp
            using type = typename decltype(output)::type;
            if constexpr(std::is_integral<type>{})
            {
                std::uniform_real_distribution<type> dis;
                std::generate(output.begin(), output.end(), dis);
            }
            else
            {
                std::uniform_int_distribution<type> dis;
                std::generate(output.begin(), output.end(), dis);
            }
        });
```

src/include/migraphx/op/rand_uniform.hpp
```diff
@@ -0,0 +1,117 @@
...
+        result.visit([&](auto output_shape) {
```
pfultz2, line 104: This should be named `output` and not `output_shape` because its a tensor_view and not a shape.

src/include/migraphx/op/rand_uniform.hpp
bpickrel: This PR has been full of comments for some time stating that we need to decide whether this operation needs to support integer data types at this time.  I take it from this comment that you think it should.  This will require me to add range arguments such as min and max, since discrete distributions don't exist without a range and the default range of (0, 1) can't be used with integers.  I'll also have to add tests for the new arguments.  I don't know if we have a use case at hand that requires the availability of discrete random distributions.  Work on the Multinomial op and dynamic batch input testing will be blocked for an additional day or two while I carry out this work.

pfultz2: > This PR has been full of comments for some time stating that we need to decide whether this operation needs to support integer data types at this time. I take it from this comment that you think it should.

The code you wrote runs on integers except it will compute incorrect results. Its simple enough to switch to `uniform_int_distribution` for integers. Plus, its good for reference operators to have coverage for all input types. 

> This will require me to add range arguments such as min and max, since discrete distributions don't exist without a range and the default range of (0, 1) can't be used with integers.

I agree, but I dont think this needs to be implemented in this PR. You can just use the default constructor for the distribution classes which will pick (0, 1) for floats and (0, max) for integers. 

> I don't know if we have a use case at hand that requires the availability of discrete random distributions.

If the model is quantized it could very likely use integral distributions. 

> Work on the Multinomial op and dynamic batch input testing will be blocked for an additional day or two while I carry out this work.

Sure, we still need to merge in Charlie's allocation PR anyways, and there are other comments in the PR that needs to be addressed.

src/include/migraphx/op/rand_uniform.hpp
pfultz2: > The STL functions allow a variety of seed types

Thats a good point. We should probably just check that the type is an integer then rather then just uint32.

src/include/migraphx/op/random_seed.hpp
pfultz2: Yea, since we do want different seed type then this should return `shape{dtype}`. This will make the shape a scalar.

src/include/migraphx/op/rand_uniform.hpp
bpickrel: That line with decltype doesn't compile.  Is there another way to get the data type of a tensor_view?  The following doesn't give a compilation error but gives a wrong result:


```
             using type =  decltype((output[0]));

            if constexpr(std::is_integral<type>{})
...
```

The error message in this case indicates the decltype is a signed char &

src/include/migraphx/op/rand_uniform.hpp
bpickrel: The with_lens() call is to ensure the output is a standard shape; but do you think it matters?

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+struct random_uniform
+{
+    // The random_uniform operation does not contain a random number generator seed
+    // as a member, and expects it to be passed as a runtime input.
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack();
```
bpickrel, line 62: If there are no attributes, should a reflect() method be defined or not?

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -0,0 +1,122 @@
...
+        std::mt19937 gen(local_seed);
+
+        result.visit([&](auto output) {
+            using type = typename decltype(output)::value_type;
+            if constexpr(std::is_integral<type>{})
+            {
+                // default range for all integer types is (0, INT_MAX) which depends
+                // on the integral type.  To clamp
+                // to a different range, apply min or max ops. to the output of this.
+                std::uniform_int_distribution<type> dis;
```
bpickrel, line 102: We agreed in an offline conversation that we don't need `max_range` and `min_range` attributes and can instead default to INT_MAX which is different for different integer types.  @pfultz2 you said that if a different range is desired we could clamp values; I think you meant by inserting a Max operator in the graph.  But clamping some of the values in a uniform distribution of numbers would create a distribution that's no longer uniform.  Do we have a way to set ranges other than `(0, INT_MAX)` that will still result in a uniform distribution?

pfultz2: We can add additional (optional) arguments to pass in to the operator in the future if clip doesnt work, but we dont need to address this in this PR.

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -0,0 +1,122 @@
...
+     *   Input 2:  output shape
+     */
+    std::string name() const { return "random_uniform"; }
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this, true}.has(2);
+
+        if(inputs.front().type() != shape::type_t::uint64_type)
+            MIGRAPHX_THROW("RANDOM_UNIFORM:  Input 1 (seed) must have type long unsigned int");
+        auto s = inputs.at(1);
```
pfultz2, line 76: This should do `return inputs.at(1)`. We shouldn't change the shape that gets passed in by the user.

src/include/migraphx/op/random_uniform.hpp
pfultz2: > If there are no attributes, should a reflect() method be defined or not?

Yea `reflect()` can be removed here.

test/ref_ops_test.cpp
```diff
@@ -6458,6 +6458,168 @@ TEST_CASE(quantizelinear)
...
+
+    p.compile(migraphx::make_target("ref"));
+    auto result = p.eval({}).back();
+    std::vector<uint32_t> result_vec1(1);
+    result.visit([&](auto output) { result_vec1.assign(output.begin(), output.end()); });
+    std::vector<uint64_t> result_vec2(1);
+
+    result = p.eval({}).back();
+    result.visit([&](auto output) { result_vec2.assign(output.begin(), output.end()); });
+    EXPECT(result_vec1[0] != result_vec2[0]);
```
CharlieL7, line 6620: It looks like this is doing an implicit conversion from `uint32_t` to `uint64_t` and then comparing that the results are not the same between the two random number generations. Technically sequential numbers from a random number generation can be the same number. This might make the test fail randomly. I think all we can test is that we do get an integer value out of this function.

src/include/migraphx/op/random_seed.hpp
```diff
@@ -0,0 +1,74 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#ifndef MIGRAPHX_GUARD_OPERATORS_RANDOM_SEED_HPP
+#define MIGRAPHX_GUARD_OPERATORS_RANDOM_SEED_HPP
+
+#include <migraphx/check_shapes.hpp>
+#include <migraphx/argument.hpp>
+#include <migraphx/par_for.hpp>
```
CharlieL7, line 30: Doesn't look like we use this

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -0,0 +1,117 @@
...
+ *
+ *      Output:   Same shape.
+ *
+ */
+#ifndef MIGRAPHX_GUARD_OPERATORS_RANDOM_UNIFORM_HPP
+#define MIGRAPHX_GUARD_OPERATORS_RANDOM_UNIFORM_HPP
+
+#include <migraphx/check_shapes.hpp>
+#include <migraphx/argument.hpp>
+#include <migraphx/par_for.hpp>
```
CharlieL7, line 46: unused?

test/ref_ops_test.cpp
```diff
@@ -6458,6 +6458,168 @@ TEST_CASE(quantizelinear)
...
+
+    // Runtime randomization seed
+    migraphx::shape seed_shape{migraphx::shape::uint64_type, {1}};
+    std::vector<uint64_t> seed_data{seed};
+    auto seed_input = mm->add_literal(migraphx::literal(seed_shape, seed_data));
+
+    mm->add_instruction(migraphx::make_op("random_uniform"), seed_input, input);
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::parameter_map params0;
```
CharlieL7, line 6483: parameter_map is not editted, can call as `p.eval({}).back()` instead

test/ref_ops_test.cpp
```diff
@@ -6458,6 +6458,168 @@ TEST_CASE(quantizelinear)
...
+    auto result = p.eval(params0).back();
+    std::vector<float> result_vec(sample_size);
+    result.visit([&](auto output) { result_vec.assign(output.begin(), output.end()); });
+
+    // Compare result with the STL's mt19937 generator
+    std::mt19937 gen(seed);
+    std::uniform_real_distribution<> dis(0.0, 1.0);
+    std::vector<float> rand_samples(sample_size);
+    std::generate(rand_samples.begin(), rand_samples.end(), [&]() { return dis(gen); });
+    EXPECT(migraphx::verify::verify_range(result_vec, rand_samples, 100000));
```
CharlieL7, line 6493: Why are these tolerance values so high? Aren't these using the same random number generator and seed?

CharlieL7 (changes requested):

src/include/migraphx/op/random_uniform.hpp
bpickrel: ok, changed.  I'm also removing the type check in line 74 entirely because in fact any type can be used as a random seed.

test/ref_ops_test.cpp
bpickrel: I think the change in types is a mistake and I'm fixing it, but the real test here is that two identical calls to the random seed generator give different values.  We shouldn't drop the test for fear of a vanishingly unlikely false positive.  When testing random data, we have to be robust enough to deal with a certain amount of randomness.

test/ref_ops_test.cpp
bpickrel: The actual error should be exactly 0, but I'm leery of ever comparing floating-point values without an error epsilon.  The tolerance given is not really so high since verify_range() multiplies it by an epsilon, but I'm going to reduce it to 100 anyway.

test/ref_ops_test.cpp
```diff
@@ -6458,6 +6458,168 @@ TEST_CASE(quantizelinear)
...
+    std::vector<float> result_vec(sample_size);
+    result.visit([&](auto output) { result_vec.assign(output.begin(), output.end()); });
+
+    // Compare result with the STL's mt19937 generator.  Expected verify_range error ~0.4
+    std::mt19937 gen(control_seed);
+    std::uniform_real_distribution<> dis(0.0, 1.0);
+    std::vector<float> rand_samples(sample_size);
+    std::generate(rand_samples.begin(), rand_samples.end(), [&]() { return dis(gen); });
+    EXPECT(migraphx::verify::verify_range(result_vec, rand_samples, 1.e8));
+}
```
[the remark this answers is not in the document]
bpickrel: Your second sentence is correct.  This is a rather special testing situation since `random_seed` is, to my knowledge, the only operation whose only use is as a prerequisite for a specific other operation.  After giving it some more thought, I don't think comparing the result against a second random distribution is such a bad idea.  It should give us an error if the code breaks.

src/include/migraphx/op/random_uniform.hpp
bpickrel: fixed

test/ref_ops_test.cpp
CharlieL7: I don't think having the chance for a false negative is good to have in our unit tests. @umangyadav what is your opinion on this?

CharlieL7: Talked with Umang, keep it as is and add a note that this is not checking for randomness (would need to do a statistical analysis). And that this could possibly produce a false negative.

bpickrel: Comment added

test/ref_ops_test.cpp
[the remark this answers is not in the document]
bpickrel: After more discussion, I removed the comparison against a second random distribution.  The test is just that it runs both ops without failing.

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Random Uniform distribution operator.  Given a shape, populate it with random
+ * values.  Calls to random_uniform using the same randomization seed will
+ * always generate the same pseudo-random sequence.  Seed can
+ * be given as a runtime argument containing a single value, or a compile-time
+ * attribute.
+ *
+ *      Inputs:   (1) randomization seed (any type is allowed)
+ *                (2) the shape of the set to be populated.
```
CharlieL7, line 33: The input would not be a `shape` object but an `argument`. 

```suggestion
 *                (2) output buffer argument to be populated.
```

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -0,0 +1,108 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+/**
+ * random_uniform populates the passed shape with random numbers, in a uniform
+ * distribution.  Range for floating-point data types is (0, 1);
+ * for integer types it is [0, <max value for the type>]
+ *
+ *   Input 1:  seed
+ *   Input 2:  output shape
```
CharlieL7, line 58: same clarification

test/ref_ops_test.cpp
```diff
@@ -6458,6 +6458,161 @@ TEST_CASE(quantizelinear)
...
+    auto result = p.eval(params0).back();
+    std::vector<uint16_t> result_vec(sample_size);
+    result.visit([&](auto output) { result_vec.assign(output.begin(), output.end()); });
+
+    // Compare result with the STL's mt19937 generator
+    std::mt19937 gen(seed);
+    std::uniform_int_distribution<uint16_t> dis;
+    std::vector<uint16_t> rand_samples(sample_size);
+    std::generate(rand_samples.begin(), rand_samples.end(), [&]() { return dis(gen); });
+    EXPECT(migraphx::verify::verify_range(result_vec, rand_samples, 100000));
```
CharlieL7, line 6529: Why not keep the default epsilon multiplier of 80?

test/ref_ops_test.cpp
```diff
@@ -6458,6 +6458,161 @@ TEST_CASE(quantizelinear)
...
+
+    std::vector<float> result_vec(sample_size);
+    result.visit([&](auto output) { result_vec.assign(output.begin(), output.end()); });
+
+    // Compare result with the STL's mt19937 generator
+    std::mt19937 gen(seed);
+    std::uniform_real_distribution<> dis(0.0, 1.0);
+    std::vector<float> rand_samples(sample_size);
+    std::generate(rand_samples.begin(), rand_samples.end(), [&]() { return dis(gen); });
+    EXPECT(migraphx::verify::verify_range(result_vec, rand_samples, 100000));
```
CharlieL7, line 6568: same question

CharlieL7 (approved):
Minor comments

pfultz2 (approved):

PR #2054, opened 2023-08-08
Remove deprecated gcnArch

umangyadav:
`gcnArch` was of type `int` and some of newer arch names are mix of char and int e.g. `gfx90a`. Therefore `gcnArch` was deprecated in rocm-3.7 and will be removed in future release. 

i don't think anyone would use MIGraphX with rocm-3.7 or older.

### DIFF
--- a/src/targets/gpu/include/migraphx/gpu/context.hpp
+++ b/src/targets/gpu/include/migraphx/gpu/context.hpp
@@ -49,7 +49,6 @@ struct hip_device
     hip_device()
     {
         device_props.gcnArchName[0]      = '\0';
-        device_props.gcnArch             = 0;
         device_props.multiProcessorCount = 0;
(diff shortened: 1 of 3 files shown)


umangyadav:
closing https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2053 solves it.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -49,7 +49,6 @@ struct hip_device
     hip_device()
     {
         device_props.gcnArchName[0]      = '\0';
-        device_props.gcnArch             = 0;
```
pfultz2, line 52: We should just default initialize `hipDeviceProp_t` instead.

PR #2058, opened 2023-08-10
Use array of binary for literals size >4GB to get around `msgpack` limit

### LINKED ISSUE #2057: Handle binary blobs larger than 4GB in msgpack
https://github.com/msgpack/msgpack-c/issues/766

msgpack uses `uint32_t` as the size for the binary blobs which means it has upper limit on ~ 4GB as the size of the binary blobs. 

Many large model such DLRM has weights in range of 10GB (as noted in https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2031). When such models are converted to MXR and run with MIGraphX, MIGraphX seg faults because of the heap buffer overflow. 

Need to handle greater than 4GB blobs in msgpack by representing them as an array of binary. May need to some same with other data structures as well if necessary.

umangyadav:
Fixes #2057 

Tested on DLRM model and it works.

For binary blobs larger than 4GB, this could slower speed of deserialization.

Another problem is that `literal` doesn't have a constructor that can accept `std::move()`. Here I need to create a temporary vector of huge size and pass it to a literal constructor which in turn allocates huge buffer size and just copies data to the internal buffer from the temporary vector. 

having a `value::binary_array` type would solve this in a better way.


TedThemistokleous (approved):

TedThemistokleous (changes requested):
fix ci, otherwise i get what you're doing here.

src/serialize.cpp
```diff
@@ -75,9 +75,11 @@ void migraphx_from_value(const value& v, literal& l)
     else
     {
         auto v_data = v.at("data");
-        assert(v_data.is_array());
+        if(not v_data.is_array())
+        {
+            MIGRAPHX_THROW("Literal is larger than 4GB but it is not stored as binary array");
```
pfultz2, line 80: If you want to add exceptions for this, this should go into msgpack. Other formats(such as json) do not have these limitations.

src/serialize.cpp
```diff
@@ -37,15 +38,64 @@ void raw_data_to_value(value& v, const RawData& rd)
...
+        {
+            size_t array_size = 1 + ((binary_size - 1) / partition_length);
+            std::vector<migraphx::value> v_array(array_size);
+            for(size_t i = 0; i < array_size; ++i)
+            {
+                size_t chunk_size =
+                    (i == (array_size - 1)) ? (binary_size % partition_length) : partition_length;
+                v_array[i] =
+                    migraphx::value::binary{(rd.data() + (i * partition_length)), chunk_size};
+            }
```
pfultz2, line 54: This should be done in msgpack.cpp. This is a limitation of msgpack not of the value class.

umangyadav: Inside msgpack while unpacking, i do not have a way to determine if array<binary> is for the literal or is it really just array<binary> . If it is for the literal, i want to bundle array<binary> as single value::binary.

src/serialize.cpp
umangyadav: Would anyone use JSON for such large data ? It would be too slow.

src/serialize.cpp
pfultz2: We should do this chunking for all arrays and objects.

pfultz2: Here's my attempt to do chunking for msgpack only: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2059

umangyadav:
closing this, superseded by #2068

PR #2061, opened 2023-08-11
Windows: Fix instruction_ref iterator issues in debug builds

m-gallus:
Currently in MIGraphX the `instruction_ref` being an alias to ` std::list<instruction>::iterator` is in many places used to reference instructions coming from a given module. However, there are many places in the code where these `instruction_ref`s originate from different `std::list` containers, but they are compared with `opeator==` or `operator!=` to one another. According to C++11 standard this is illegal. Be default Windows' STL implementation of iterators has asserts in place to verify if the iterators compared come from the same container, or if an `.end()` of a list is being dereferenced. This causes various tests to fail such as:
•	test_eliminate_common_subexpression_test 
•	test_eval_test 
•	test_fuse_reduce 
•	test_inline_module_test 
•	test_matcher 
•	test_memory_coloring_test 
•	test_module_test 
•	test_replace_allocate 
•	test_onnx_test 

The fix proposed makes `iterator_ref` a class inheriting from `std::list<instruction>::iterator` and overloads certain comparison operators, as well as conversion operator form `std::list<instruction>::iterator` to this type.

Before arriving at this approach, several other approaches were undertaken, namely:
- Attempt to fix the issue at its root cause, however the practice of unsafe iterator comparisons was so prevalent in the codebase it would require an overhaul of part of the framework's architecture.
- Overloading of equals comparison function `operator==` - failed due to that operator already being overloaded for ` std::list<instruction>::iterator` 
- Wrapping `std::list<instruction>::iterator` in a class - there were cases where `instruction_ref` was fed to STL iterator functions such as `std::prev` and etc.

This is by no means an ultimate solution to the problem which is to restructure how `instruction_ref`s are used in the code, however it unblocks many unit tests and allows windows developers to concentrate their efforts on the remaining issues.


TedThemistokleous (changes requested):
Looks good. Fix errors in CI

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,40 @@ namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 struct instruction;
+#if defined(_WIN32) && !defined(NDEBUG)
+class instruction_ref : public std::list<instruction>::iterator
```
pfultz2, line 36: Use `struct` since everything is pulbic.

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,40 @@ namespace migraphx {
...
+{
+public:
+    instruction_ref() = default;
+    instruction_ref(const std::list<instruction>::iterator& other)
```
pfultz2, line 40: It seems it would be simpler to do `using std::list<instruction>::iterator::std::list<instruction>::iterator` here.

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,40 @@ namespace migraphx {
...
+    {
+        _Ptr = other._Ptr;
+        _Adopt(other._Getcont());
+    }
+
+    bool operator==(const instruction_ref& other) const
```
pfultz2, line 46: You should write:

```cpp
friend bool operator==(const std::list<instruction>::iterator& x, const std::list<instruction>::iterator& y)
{
    return this->_Unwrapped()._Ptr == other._Unwrapped()._Ptr;
}
```

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,40 @@ namespace migraphx {
...
+    {
+        return this->_Unwrapped()._Ptr == other._Unwrapped()._Ptr;
+    }
+
+    bool operator==(const std::list<instruction>::const_iterator& other) const
+    {
+        return this->_Unwrapped()._Ptr == other._Unwrapped()._Ptr;
+    }
+
+    bool operator!=(const instruction_ref& other) const
```
pfultz2, line 61: This also should be declared as `friend bool operator!=(const std::list<instruction>::iterator& x, const std::list<instruction>::iterator& y)`.

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,47 @@ namespace migraphx {
...
+
+    instruction_ref() = default;
+    instruction_ref(const instruction_iter& other)
+    {
+        _Ptr = other._Ptr;
+        _Adopt(other._Getcont());
+    }
+
+    friend bool operator==(const instruction_ref& x, const instruction_ref& y);
+    friend bool operator==(const instruction_ref& x, const instruction_iter& y);
```
pfultz2, line 49: I dont think this overload is needed.

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,47 @@ namespace migraphx {
...
+    friend bool operator==(const instruction_ref& x, const instruction_const_iter& y);
```
pfultz2, line 50: Is this overload needed as well?

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,47 @@ namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 struct instruction;
+#if defined(_WIN32) && !defined(NDEBUG)
+struct instruction_ref : public std::list<instruction>::iterator
```
pfultz2, line 36: `public` can be removed.

src/include/migraphx/instruction_ref.hpp
m-gallus: Yes, a comparison between an instance of `instruction_ref` and an .end( ) of a list is performed somewhere in the code, hence the operator overload was necessary.

src/include/migraphx/instruction_ref.hpp
```diff
@@ -32,7 +32,44 @@ namespace migraphx {
...
+struct instruction_ref : std::list<instruction>::iterator
+{
+    using instruction_iter       = std::list<instruction>::iterator;
+    using instruction_const_iter = std::list<instruction>::const_iterator;
+
+    instruction_ref() = default;
+    instruction_ref(const instruction_iter& other)
+    {
+        _Ptr = other._Ptr;
+        _Adopt(other._Getcont());
```
pfultz2, line 45: Cant we use the base copy constructor for this?

src/include/migraphx/instruction_ref.hpp
pfultz2: Yea, thats why it should be `friend bool operator==(const std::list<instruction>::iterator& x, const std::list<instruction>::iterator& y)` so it will work with both `std::list<instruction>::iterator` and `instruction_ref`.

src/include/migraphx/instruction_ref.hpp
m-gallus: Good idea, done.

src/include/migraphx/instruction_ref.hpp
pfultz2: We arent overriding whats in the standard library.  A friend function is not visible to the outer scope namespace, it can only be called through ADL, which means you need to be using an `instruction_ref` argument.

m-gallus: Even though we don't overload what is in STL, there is an ambiguity when it comes to choosing the comparison operator, as an already fitting one is found in the STL:
```
C:/develop/AMDMIGraphX/src/module.cpp:512:12: error: use of overloaded operator '!=' is ambiguous (with operand types 'std::_List_iterator<std::_List_val<std::_List_simple_types<migraphx::instruction>>>' and 'instruction_ref')
    if(ins != this->end())
       ~~~ ^  ~~~~~~~~~~~
C:/develop/AMDMIGraphX/src/include\migraphx/instruction_ref.hpp:63:17: note: candidate function
    friend bool operator!=(const instruction_iter& x, const instruction_iter& y)
                ^
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.35.32215\include\list:202:21: note: candidate function
    _NODISCARD bool operator!=(const _List_const_iterator& _Right) const noexcept {
                    ^
C:/develop/AMDMIGraphX/src/include\migraphx/instruction_ref.hpp:58:17: note: candidate function
    friend bool operator!=(const instruction_ref& x, const instruction_ref& y)
                ^
```

umangyadav (approved):

src/include/migraphx/instruction_ref.hpp
pfultz2: That is because you have two overloads that match at line 58 and line 63. There should only be one `operator==` defined:

```cpp
friend bool operator==(const std::list<instruction>::iterator& x, const std::list<instruction>::iterator& y)
{
    ...
}
```

If that doesnt resolve the ambiguity then we could try a template:

```cpp
template<class T, class U, MIGRAPHX_REQUIRES(std::is_same<T, instruction_ref>{} or std::is_same<U, instruction_ref>{})>
friend bool operator==(const T& x, const U& y)
{
    ...
}
```

Otherwise you are going to need 5 overloads, so it will be simpler if you can just define one.

m-gallus: It's done. Trying the first option with `std::list<instruction>::iterator`s was causing the function body to never be invoked, however the second approach with conditional enablement did the job.

pfultz2 (approved):

TedThemistokleous (approved):

PR #2063, opened 2023-08-11
Disable mlir in default builds for now

pfultz2:


pfultz2:
So I disabled it by default for dev builds since MLIR fails on other OSs like sles. However, I still enable it when running tidy, so we shouldn't need new tidy fixes when we re-enable it.

.github/workflows/ci.yaml
```diff
@@ -55,6 +55,7 @@ jobs:
           -DMIGRAPHX_ENABLE_GPU=On \
           -DMIGRAPHX_ENABLE_CPU=On \
           -DMIGRAPHX_ENABLE_FPGA=On \
+          -DMIGRAPHX_ENABLE_MLIR=On \
```
TedThemistokleous, line 58: Should this be off if we're disabling mlir?

TedThemistokleous (approved):

.github/workflows/ci.yaml
pfultz2: No, I want tidy to be enabled so we can still catch tidy issues.

TedThemistokleous: all good. I saw your previous message after I wrote that. Should be fine

PR #2068, opened 2023-08-13
Add chunking of binaries when writing with msgpack

pfultz2:
The will help workaround the limitations of msgpack, but this only applies to binary data. For arrays and maps it will now throw an error when trying to save such a large size.


TedThemistokleous (approved):

src/msgpack.cpp
```diff
@@ -25,6 +25,33 @@
 #include <migraphx/serialize.hpp>
 #include <msgpack.hpp>
 
+namespace migraphx {
```
umangyadav, line 28: Add unit-test with more than 4GB ? It would be slow but good to have

src/msgpack.cpp
```diff
@@ -102,8 +144,12 @@ MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS)
         {
             const auto* data = reinterpret_cast<const char*>(x.data());
             auto size        = x.size();
-            o.pack_bin(size);
-            o.pack_bin_body(data, size);
+            o.pack_array(migraphx::msgpack_chunk_size(x));
+            migraphx::msgpack_chunk_for_each(
+                data, data + size, [&](const char* start, const char* last) {
+                    o.pack_bin(last - start);
+                    o.pack_bin_body(data, last - start);
```
umangyadav, line 151:
```suggestion
                    o.pack_bin_body(data+start, last - start);
```

Shoud it also update the base pointer ?

src/msgpack.cpp
```diff
@@ -63,16 +90,31 @@ MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS)
                 break;
             }
             case msgpack::type::BIN: {
+                // For backwards compatibility
```
umangyadav, line 93: This is not helpful, all previously generated MXRs would fail inside `from_value()` because of file version mismatch.

test/msgpack.cpp
```diff
@@ -25,13 +25,37 @@
...
+    msgpack::pack(os, src);
+}
+void write_msgpack(std::ostream& os, const std::vector<char>& src)
+{
+    const auto limit = std::numeric_limits<uint32_t>::max() - 1;
+    std::vector<std::vector<char>> chunks;
+    if(src.size() > limit)
+    {
+        // Only test two chunks
+        assert(std::distance(src.begin() + limit, src.end()) < limit);
```
umangyadav, line 43: This part is not used in any test. `test_msgpack_binary` is only using one  chunk

src/msgpack.cpp
```diff
@@ -25,6 +25,33 @@
 #include <migraphx/serialize.hpp>
 #include <msgpack.hpp>
 
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+// Leave an extra byte for error checking
```
umangyadav, line 31: Where and how this extra byte is being used?

src/msgpack.cpp
```diff
@@ -25,6 +25,33 @@
...
+constexpr std::size_t msgpack_size_limit = std::numeric_limits<uint32_t>::max() - 1;
+
+template <class Range>
+std::size_t msgpack_chunk_size(const Range& r)
+{
+    return 1 + r.size() / msgpack_size_limit;
```
umangyadav, line 37: This should do ceil division. Should it be `1 + (r.size() -1)  / msgpack_size_limit` ?

src/msgpack.cpp
pfultz2: Yes it would fail in `from_value`, but with an error about the incompatible version, which I think is a much more informative error message.

src/msgpack.cpp
pfultz2: I assumed it would be clamped, I guess it will overflow so we cant check for overflow.

pfultz2: Still it might be good to avoid fully saturated values just in case there is some issue.

umangyadav (approved):
Verified that it works with DLRM fp16 model. Fp32 model doesn't fit into memory.

PR #2071, opened 2023-08-14
Fix bug in eliminate_contiguous for operators that have module inputs

pfultz2:


test/eliminate_contiguous_test.cpp
```diff
@@ -196,15 +196,47 @@ TEST_CASE(contiguous_pointwise)
...
-    EXPECT(std::distance(mm->begin(), mm->end()) == (count - 1));
+    EXPECT(std::distance(mm->begin(), mm->end()) == (count - 2));
     EXPECT(std::none_of(
         mm->begin(), mm->end(), [](auto&& ins) { return ins.name() == "contiguous"; }));
 }
 
+TEST_CASE(contiguous_nhwc_pointwise)
+{
+    auto s = migraphx::shape::from_permutation(
+        migraphx::version_1::shape::float_type, {2, 3, 8, 8}, {0, 2, 3, 1});
```
umangyadav, line 212: Looks like vscode auto-gen. Fix tidy.

umangyadav (approved):
Surprised, this didn't show up earlier as a bug so far.

test/eliminate_contiguous_test.cpp
pfultz2: Oops. Actually, I use sublime not vscode. Either way its from the clangd autocompletions.

kahmed10 (approved):

causten:
@pfultz2 onnx-taau-downsample dropped significantly, can you review

pfultz2:
> onnx-taau-downsample dropped significantly, can you review

So this doesn't seem to run with miopen on my local machine. After using #2102(which fixes this issue), the compiled programs is the same as the program on develop. Most likely a mixed layout is being used and miopen is calling a different kernel, which most likely produces incorrect results.

Issue #1677, opened 2023-04-10
Add reshape_copy operator


umangyadav:
Currently reshape operator requires standard shape inputs only.  Reshape operator currently aliases input. 

Add reshape operator that copies instead of aliases. 

This would resolve many of the bugs encountered in simplify_algebra, for example this one: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1666

TedThemistokleous:
In review. No more work required

PR #2072, opened 2023-08-14
Unique Operator.

lakhinderwalia:
Skeletal code.. not ready yet.


test/ref_ops_test.cpp
```diff
@@ -8765,26 +8765,102 @@ TEST_CASE(transpose_dyn_test)
...
-    auto ret   = mm->add_instruction(migraphx::make_op("unique"), {{"sorted", {0}}}, input);
-    mm->add_return({ret});
-    p.compile(migraphx::make_target("ref"));
-    auto result = p.eval({}).back();
-    std::vector<int64_t> result_vector;
-    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
-    std::vector<int64_t> gold = {0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 1, 1, 0, 0,
-                                 1, 1, 0, 0, 0, 0, 0, 1, 0, 2, 0, 2, 0, 2, 0, 0, 0, 0};
-    EXPECT(migraphx::verify::verify_range(result_vector, gold));
+    auto create_program = [](int64_t axis, int64_t sorted, size_t test_sz) {
```
umangyadav, line 8770: Rebase with respect to latest develop branch. 
There has been some changes in ref ops test structure.

lakhinderwalia:
Unique Operator: the dynamic data case works, with zeros paddeding: The reference test checks for the expected (zero-padded) results. Static data case: const-folding test cases -- no zero padding is required.

src/include/migraphx/op/scatter.hpp
```diff
@@ -66,7 +66,7 @@ struct scatter : op_name<Derived>
 
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(3).standard();
+        // check_shapes{inputs, *this}.has(3).standard();
```
lakhinderwalia, line 69: @pfultz2 I have disabled a couple of tests in scatternd5 and scatternd_dyn5, just to get past the shape tests that were effectively tripped by disabling the check here.

lakhinderwalia:
I have a question for the reviewers:  How should one demarcate the zero-filled results of an operator like unique where the dynamic output size/shape isn't known in advance -- its shape is a subset of the input shape. We zero-fill our results, yes, but zero is part of a valid output, so it is a bit unclear how do we demarcate. 

We do have one way out of this here, specifically for the unique operator. One of the resulting output -- it is always non-zero: the count. So, I can use that as an OOB signal. 

@pfultz2, @umangyadav, @TedThemistokleous

CharlieL7:
> I have a question for the reviewers: How should one demarcate the zero-filled results of an operator like unique where the dynamic output size/shape isn't known in advance -- its shape is a subset of the input shape. 

Does the operator require us to demarcate the zero-filled results? If not, it should just have a padded output. Keep in mind that we probably won't be using this operator with zero padding later, but rather dynamic shapes.

lakhinderwalia:
> > I have a question for the reviewers: How should one demarcate the zero-filled results of an operator like unique where the dynamic output size/shape isn't known in advance -- its shape is a subset of the input shape.
> 
> @lakhinderwalia Does the operator require us to demarcate the zero-filled results? If not, it should just have a padded output. Keep in mind that we probably won't be using this operator with zero padding later, but rather dynamic shapes.

Zero-filled results is our current (I think intermediate) approach; But zero-filled (demarcation) isn't blessed by the operator (specification). Thanks.

pfultz2:
> Zero-filled results is our current (I think intermediate) approach; But zero-filled (demarcation) isn't blessed by the operator (specification). Thanks.

Yea you need to return a dynamic shape in `compute_shape`, and then in `compute` you will need to do a reshape on the outputs arguments to trim off the extra zeros.

test/ref/unique.cpp
```diff
@@ -0,0 +1,338 @@
...
+        EXPECT(idx == gold_y_idx);
+
+        std::vector<int64_t> gold_x_rev = {0};
+        EXPECT(x_rev == gold_x_rev);
+
+        std::vector<int64_t> gold_ct = {1};
+        EXPECT(ct == gold_ct);
+    }
+
+    // case 2  sorted. all unique input..
```
umangyadav, line 123: can we create different test cases for all those cases ?

src/include/migraphx/tune_axis.hpp
```diff
@@ -24,21 +24,21 @@
...
 
-inline int tune_axis(const int n_dim, const int axis, const std::string& op_name = "OPERATOR")
+inline int tune_axis(int n_dim, int axis, const std::string& op_name = "OPERATOR")
 {
-    if(axis >= n_dim or std::abs(axis) > n_dim)
-    {
+    if(axis < 0)
+        axis += n_dim;
+
+    if(axis < 0 or axis >= n_dim)
```
umangyadav, line 38: Why this logic was changed ? I think previous way of just checking for `abs(axis)` is cleaner IMO

lakhinderwalia: It is logically confusing to do a comparison of `axis` twice in the same conditional -- once as its absolute version, the other as its original version. What is this code really doing by its intent..? And it is thus simpler to handle the negative axis first, and then do the right boundary checks. Thanks.

test/onnx/gen_onnx.py
```diff
@@ -8994,6 +8994,80 @@ def undefined_test():
...
+    y_ind = helper.make_tensor_value_info('indices', TensorProto.INT64, [4])
+    x_ind = helper.make_tensor_value_info('inverse_indices', TensorProto.INT64,
+                                          [6])
+    count = helper.make_tensor_value_info('counts', TensorProto.INT64, [4])
+
+    node = onnx.helper.make_node(
+        'Unique',
+        inputs=['X'],
+        outputs=['Y', 'indices', 'inverse_indices', 'counts'],
+        axis=0,
```
umangyadav, line 9010: add onnx test with axis not specified

test/ref/unique.cpp
lakhinderwalia: You mean separate them out within this file? Thanks.

umangyadav: yes separate `TEST_CASE` for each case.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+        size_t dim_x       = sh_x.ndim();
+        size_t max_uniq_ct = sh_x.elements();
+
+        int64_t t_axis = axis;
+        if(t_axis != axis_none)
+        {
+            t_axis = migraphx::tune_axis(dim_x, t_axis, name());
+            if(t_axis != 0)
+                MIGRAPHX_THROW("Unique: Only supports axis = 0 or None");
+            max_uniq_ct /= sh_x.strides()[t_axis];
```
umangyadav, line 233: better to do just pick `max_uniq_ct = sh_x.lens[t_axis];`

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+        // min = 1 unique element; max = full dimension along the axis
+        // The three outputted Indices are just 1-D:
+        std::vector<shape::dynamic_dimension> d_out{{1, max_uniq_ct}};
+        shape sh_idx{shape::int64_type, {d_out}};
+
+        // Unique elements themselves aren't necessarily 1-D.. fix its output shape:
+        for(size_t idx = 1; t_axis != axis_none && t_axis + idx < dim_x; idx++)
+            d_out.push_back({lens_x[t_axis + idx], lens_x[t_axis + idx]});
+        shape sh_y = {sh_x.type(), {d_out}};
```
umangyadav, line 244:
```suggestion
        shape sh_y = {sh_x.type(), d_out};
```

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+        return {{sh_y, sh_idx, sh_idx, sh_idx}};
```
umangyadav, line 246:
```suggestion
        return {sh_y, sh_idx, sh_idx, sh_idx};
```

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+        // go through all the elements and find the unique elements..
+        size_t count_x = input_data.size();
+        for(size_t f_idx = 0, x_idx = 0; f_idx < count_x; f_idx += chunk_sz, x_idx++)
+        {
+            y_info entry          = {.y_idx = uniq_val_map.size(), .x_idx = x_idx};
+            auto [itr, added_new] = uniq_val_map.insert({f_idx, entry});
+            itr->second.ct++;
+            x_rev_indices.push_back(itr->second.y_idx);
+        }
```
pfultz2, line 127: I think it would be much simpler to write this as:

```cpp
for(auto x_idx:range(input_data.size() / chunk_sz))
{
    size_t f_idx = x_idx * chunk_sz;
    y_info entry          = {.y_idx = uniq_val_map.size(), .x_idx = x_idx};
    auto [itr, added_new] = uniq_val_map.insert({f_idx, entry});
    itr->second.ct++;
    x_rev_indices.push_back(itr->second.y_idx);
}
```

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+            d_out.push_back({lens_x[t_axis + idx], lens_x[t_axis + idx]});
+        shape sh_y = {sh_x.type(), {d_out}};
+
+        return {{sh_y, sh_idx, sh_idx, sh_idx}};
+    }
+
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    {
+        size_t uniq_ct     = 0;
+        auto sh_x          = args.front().get_shape();
```
umangyadav, line 252: Just my preference, naming it `x_s`. I think we shape vars named like it in a few places.  Ignore this if you don't like it.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+            // for efficiency, loop is bypassed for single elements. ct = 1
+            for(size_t ct = chunk_sz; ct > 1; idx1++, idx2++, ct--)
+            {
+                if(input_data[idx1] < input_data[idx2])
+                    return true;
+                else if(input_data[idx1] > input_data[idx2])
+                    return false;
+            }
+            return input_data[idx1] < input_data[idx2];
+        };
```
pfultz2, line 178: This can use `std::lexicographical_compare` as well, but instead of repeating this, I think you should make a function so it can be reused:

```cpp
template<class T>
auto make_idx_less_fn(const T& input_data, size_t chunk_sz) const
{
    assert(idx1+chunk_sz < input_data.size());
    assert(idx2+chunk_sz < input_data.size());
    return [&input_data, chunk_sz](auto idx1, auto idx2) {
        return std::lexicographical_compare(input_data.begin()+idx1,
                                            input_data.begin()+idx1+chunk_sz,
                                            input_data.begin()+idx2,
                                            input_data.begin()+idx2+chunk_sz);
    };
}
```

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+        size_t uniq_ct     = 0;
+        auto sh_x          = args.front().get_shape();
+        size_t dim_x       = sh_x.ndim();
+        auto lens_x        = sh_x.lens();
+        shape output_shape = dyn_out.computed_shape;
+        auto vec_ss        = output_shape.sub_shapes();
+        auto x_ct          = sh_x.elements();
+        shape sh_y         = {vec_ss[0].type(), {x_ct}};
+        shape sh_idx       = {vec_ss[1].type(), {x_ct}};
+        shape sh_x_idx     = {vec_ss[1].type(), {x_ct}};
```
umangyadav, line 260: you can reuse same var.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+            y_count[itr->second]++;
+            x_rev_indices.push_back(itr->second);
+        }
+
+        return std::tuple{y_indices, x_rev_indices, y_count};
+    }
+
+    // Default: none. Range: [-rank, rank-1]
+    static constexpr int64_t axis_none = std::numeric_limits<int64_t>::max();
+    int64_t axis                       = axis_none;
```
pfultz2, line 205: I think it would be better to write it with `optional<int64_t> axis = nullopt` instead of using magic values.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+    // Default: 1= sorted. 0 = unsorted.
+    int64_t sorted = 1;
```
pfultz2, line 208: This should be `bool`.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.axis, "axis"), f(self.sorted, "sorted"));
+    }
+
+    std::string name() const { return "unique"; }
+
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this}.has(1).standard();
```
pfultz2, line 220: Remove `.standard()`, we should never require this for ref operators.

src/include/migraphx/op/unique.hpp
umangyadav: never mind, i see it is used differently later.

src/include/migraphx/op/unique.hpp
lakhinderwalia: BTW, this is an int: part of the onnx specification..

src/include/migraphx/op/unique.hpp
lakhinderwalia: These diverge later. And it is important to state (and keep) the two shapes different.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+        {
+            t_axis   = tune_axis(dim_x, t_axis, name());
+            chunk_sz = sh_x.strides()[t_axis];
+        }
+
+        visit_all(args.front(), res_y)([&](auto x, auto y_flat) {
+            using o_type = typename decltype(x)::value_type;
+            std::vector<o_type> x_in(x_ct);
+            auto itr = x_in.begin();
+            shape_for_each(sh_x, [&](auto, auto idx) { *itr++ = x[idx]; });
```
umangyadav, line 286: you can use `to_vector` instead

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        // min = 1 unique element; max = full dimension along the axis
+        // The three outputted Indices are just 1-D:
+        std::vector<shape::dynamic_dimension> d_out{{1, max_uniq_ct}};
+        shape sh_idx{shape::int64_type, {d_out}};
+
+        // Unique elements themselves aren't necessarily 1-D.. fix its output shape:
+        for(size_t idx = 1; t_axis != axis_none && t_axis + idx < dim_x; idx++)
+            d_out.push_back({lens_x[t_axis + idx], lens_x[t_axis + idx]});
```
pfultz2, line 243: Use `transform` for this raw loop:

```cpp
if (t_axis != axis_none) // or t_axis.has_value() if you use optional
{
    transform(range(1, dim_x-t_axis), std::back_inserter(d_out), [&](auto idx) -> shape::dynamic_dimension {
        return {lens_x[t_axis + idx], lens_x[t_axis + idx]};
    });
}
```

src/include/migraphx/op/unique.hpp
pfultz2, line 286: Just do `std::vector<o_type> x_in(x.begin(), x.end())` instead of this.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+            // construct y from x[indices] in flattened form
+            // later we reshape y to the final shape..
+            for(size_t y_idx = 0, uniq_idx = 0; uniq_idx < uniq_ct; uniq_idx++)
+                for(size_t u_idx = y_indices[uniq_idx] * chunk_sz, ct = 0; ct < chunk_sz; ct++)
+                    y_flat[y_idx++] = x_in[u_idx++];
+
+            out_y_idx     = std::move(y_indices);
+            out_x_rev_idx = std::move(x_rev_indices);
+            out_y_ct      = std::move(y_count);
```
pfultz2, line 302: Since these are const references then `std::move` will just do a copy. I think it might better to assign these values directly using `std::tie`:

```cpp
std::tie(out_y_idx, out_x_rev_idx, out_y_ct) =
    sorted ? sorted_uniq_indices(x_in, chunk_sz)
           : unsorted_uniq_indices(x_in, chunk_sz);
```

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+            // for efficiency, loop is bypassed for single elements. ct = 1
+            for(size_t ct = chunk_sz; ct > 1; idx1++, idx2++, ct--)
+            {
+                if(input_data[idx1] < input_data[idx2])
+                    return true;
+                else if(input_data[idx1] > input_data[idx2])
+                    return false;
+            }
+            return input_data[idx1] < input_data[idx2];
+        };
```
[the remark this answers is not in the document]
lakhinderwalia: Lexicographical compare isn't good for FP equality comparison(s) - it gets used at the end of that API. Hence this..

src/include/migraphx/op/unique.hpp
lakhinderwalia: We are supplying a vector there.. so isn't the intent more clear this way..? Thanks.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+        // If axis is not none, for >1D tensors, we have to consider
+        // then, the uniqueness of chunks of sub-tensors: a subsequence of built-ins..
+        // For a built-in type, chunk_sz is of course = 1
+        size_t chunk_sz = 1;
+        int64_t t_axis  = axis;
+        if(axis != axis_none)
+        {
+            t_axis   = tune_axis(dim_x, t_axis, name());
+            chunk_sz = sh_x.strides()[t_axis];
```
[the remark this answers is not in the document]
lakhinderwalia: This is cumulatively adding the chunk size (of all the indices on the right of the t_axis). Hence the strides().

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+                sorted ? sorted_uniq_indices(x_in, chunk_sz)
+                       : unsorted_uniq_indices(x_in, chunk_sz);
+
+            uniq_ct = y_indices.size();
+
+            // construct y from x[indices] in flattened form
+            // later we reshape y to the final shape..
+            for(size_t y_idx = 0, uniq_idx = 0; uniq_idx < uniq_ct; uniq_idx++)
+                for(size_t u_idx = y_indices[uniq_idx] * chunk_sz, ct = 0; ct < chunk_sz; ct++)
+                    y_flat[y_idx++] = x_in[u_idx++];
```
pfultz2, line 298: Use `std::copy_n` here instead of a raw loop:

```cpp
auto y_flat_iter = y_flat.begin();
for(auto uniq_idx:range(uniq_ct))
{
    auto start = y_indices[uniq_idx] * chunk_sz;
    y_flat_iter = std::copy_n(x_in.begin()+start, chunk_sz, y_flat_iter);
}
```

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+            out_y_idx     = std::move(y_indices);
+            out_x_rev_idx = std::move(x_rev_indices);
+            out_y_ct      = std::move(y_count);
+
+            std::vector<size_t> y_lens = {uniq_ct};
+            for(size_t idx = 1; t_axis != axis_none && t_axis + idx < dim_x; idx++)
+                y_lens.push_back(lens_x[t_axis + idx]);
```
pfultz2, line 306: This can use `transform` here like above.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Sure. Old debate for us :-) A transform, and a range, and a lambda later. We still don't catch if dim_x < t_axis (due do some programming error..) and a 2 liner has become 4+ liner.. with no added clarity..  or safety..

src/include/migraphx/op/unique.hpp
lakhinderwalia: And how does this tie with reflect(). The optional<> stuff.. Thanks.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+        });
+
+        visit_all(res_y_idx, res_x_rev_idx, res_y_ct_idx)(
+            [&](auto y_indices, auto x_rev_indices, auto y_count) {
+                for(size_t i = 0; i < uniq_ct; i++)
+                {
+                    y_indices[i]     = out_y_idx[i];
+                    x_rev_indices[i] = out_x_rev_idx[i];
+                    y_count[i]       = out_y_ct[i];
+                }
```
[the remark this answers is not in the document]
lakhinderwalia: Three loops for one loop. :-)

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+                // len(y uniq_values) <= len(input x)
+                // Fill x reverse-indices beyond the len(uniq_values), if any..
+                for(size_t i = uniq_ct; i < out_x_rev_idx.size(); i++)
+                    x_rev_indices[i] = out_x_rev_idx[i];
```
pfultz2, line 324: Use `std::copy(out_x_rev_idx.begin()+uniq_ct, out_x_rev_idx.end(), x_rev_indices.begin()+uniq_ct, x_rev_indices.end())` here.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
pfultz2: What does FP stand for?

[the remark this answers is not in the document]
lakhinderwalia: Floating Point ..

src/include/migraphx/op/unique.hpp
pfultz2: We handle serializing the `optional` here: 

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/serialize.hpp#L92
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/serialize.hpp#L210

And I believe @CharlieL7 updated the printing to make it work with `optional`, so this shouldn't be a problem.

src/include/migraphx/op/unique.hpp
pfultz2: We dont need to match the onnx spec in our operators. Its used as a bool so lets use the correct type. If onnx spec adds other states then we can switch to an enum in the future.

src/include/migraphx/op/unique.hpp
pfultz2: > Sure. Old debate for us :-)

But this is not a range-based for loop. This is a raw loop as you consider it, and we should prefer algorithms over raw loops. 

> We still don't catch if dim_x < t_axis (due do some programming error..)

We can add an `assert` for this case, which also makes the preconditions clearer.

> and a 2 liner has become 4+ liner

Tersness != mor clarity. However, in this case, its same lines of code between both versions.

> with no added clarity.. or safety..

There is better clarity for those that are familiar with STL algorithms.

src/include/migraphx/op/unique.hpp
pfultz2: But the init list makes it look more like `d_out` is an element of the vector instead of the vector itself, which is why its clearer to write it without the `{}`.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
pfultz2: If its not standard shape then the stride may not be equal to the accumulated dimensions. You should probably use `std::accumulate` to compute this instead or maybe `sh_x.elements() / sh_x.lens()[t_axis]`.

src/include/migraphx/op/unique.hpp
lakhinderwalia: You both are right. Sorry. I did change it to a vector somewhere down as the code evolved.. so no need for the extra brace anymore.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
lakhinderwalia: You are right -- (I had read the last return clause too fast, which is just the iterator compare and not a value compare). Will definitely change this one. Good idea. Thanks.

src/include/migraphx/op/unique.hpp
lakhinderwalia: That is a good catch, the const part & the ignored move(), thanks, and I agree -- I wish the compiler had warned. 
Will remove the references actually, as the intent is to automatically move stuff from calls to sorted_uniq_indices() and unsorted_unique_indices().

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,340 @@
...
+        auto lens_x        = sh_x.lens();
+        size_t dim_x       = sh_x.ndim();
+        size_t max_uniq_ct = sh_x.elements();
+
+        int64_t t_axis = axis;
+        if(t_axis != axis_none)
+        {
+            t_axis = migraphx::tune_axis(dim_x, t_axis, name());
+            if(t_axis != 0)
+                MIGRAPHX_THROW("Unique: Only supports axis = 0 or None");
```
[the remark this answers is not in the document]
lakhinderwalia: BTW, there is no specific output (and hence its shape) that can be expected of the Unique Operator -- without being specific about the exact input values X.

src/include/migraphx/op/unique.hpp
lakhinderwalia: No, this is not the same thing.. I need strides() here. Not Lens(). Thanks.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
lakhinderwalia: This operator is standard shape. Hence strides() is proper to use -- rather than to effectively repeat its recalculation via accumulate etc. Lens() is not useful here. Thanks.

src/include/migraphx/op/unique.hpp
CharlieL7: I don't think `optional` works with `normalize_attributes` however.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
bpickrel: You'd expect a dynamic shape back, if I'm reading it right.  With a rank (number of dimensions) and dynamic dimension ranges that are known at compile time.  Then at runtime, in the compute() function, specific inputs and specific output values become available.  `op_shape_tests.cpp` specifically contains tests of compile-time shapes only.  You'll see that none of those tests run a program or create any inputs.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
pfultz2: > This operator is standard shape.

Ref operators should not require standard shape, which means this operator needs to work with inputs that are non standard, which means you can not rely on the strides here.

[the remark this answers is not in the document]
lakhinderwalia: Is there any operator that could input to this Unique Operator with a non-standard shape()? Thanks.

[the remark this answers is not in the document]
pfultz2: Many operators will produce non-standard shape like slice,step,transpose,broadcast,etc, but many more will propagate the permutation in memory as well. 

We dont want to require standard shape for ref operators because when we do graph optimizations it can lead to the shape changing indirectly downstream to a non-standard shape which will lead to an error. We had many of these issues previously when reshape required a standard shape(like #1496, #1579, #2136) which is why we reimplemented reshape to handle nonstandard inputs(#1677, #2099). 

Furthermore, ref operators are evaluated during constant propagation which may have standard shape inputs, so they should be written robustly to handle all memory layouts. 

For backends like gpu, and cpu, they can require standard shape, and we ensure it will be a standard shape if an operator needs it. After lowering we dont do high-level graph rewrites(like algebraic simplifications) that would change the layout in memory.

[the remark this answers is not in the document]
lakhinderwalia: OK. Thanks. I will accumulate those results from the lens, and not use strides().

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
lakhinderwalia: Done.

test/onnx/gen_onnx.py
lakhinderwalia: Done.

test/ref/unique.cpp
lakhinderwalia: Done.

src/include/migraphx/op/unique.hpp
pfultz2: > Will remove the references actually, as the intent is to automatically move stuff from calls

Why do you need to assign it to one variable and move it to another? Why not just assign into the variable directly with `std::tie`?

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,320 @@
...
+
+            out_y_idx     = std::move(y_indices);
+            out_x_rev_idx = std::move(x_rev_indices);
+            out_y_ct      = std::move(y_count);
+
+            std::vector<size_t> y_lens = {uniq_ct};
+
+            // if axis is specified:
+            // the output shape has the n-1 dimensions of x
+            for(size_t idx = 1; axis && idx < dim_x; idx++)
```
pfultz2, line 294: The check for `axis` can be moved out of the loop.

lakhinderwalia: This isn't a performance loop, left intentionally for just simplicity of expression. If it is objectionable, I will move it out.. please let me know. Thanks.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
lakhinderwalia: Strides or accumulate have been replaced by a simpler logic, since we allow only the axis=0, so not relevant any more.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Done.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Done.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,320 @@
...
+                y_lens.push_back(lens_x[idx]);
```
pfultz2, line 295: This can be written using `std::copy`: `std::copy(lens_x.begin()+1, lens_x.begin()+dims_x, std::back_inserter(y_lens))`.

pfultz2, line 295: Actually, thinking about it more, could this be written using an assignment from `x_lens` directly? Like this:

```cpp
std::vector<size_t> y_lens;
if(axis)
{
    y_lens = x_lens;
    y_lens.front() = uniq_ct;
}
else
{
    y_lens = {uniq_ct};
}
```

This assumes that `dim_x == x_lens.size()`.

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,320 @@
...
+        {
+            int64_t t_axis = migraphx::tune_axis(dim_x, *axis, name());
+            if(t_axis != 0)
+                MIGRAPHX_THROW("Unique: Only supports axis = 0 or None");
+
+            // only axis = 0 is supported:
+            max_uniq_ct /= lens_x[0];
+            d_out.push_back({1, max_uniq_ct});
+            for(size_t idx = 1; idx < dim_x; idx++)
+                d_out.push_back({lens_x[idx], lens_x[idx]});
```
pfultz2, line 229: I think this would be simpler to use `to_dynamic` here:

```cpp
d_out = sh_x.to_dynamic().dyn_dims();
d_out.front().min = 1;
d_out.front().max = max_uniq_ct;
```

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,320 @@
...
+        auto idx_less_fn = [&](auto idx1, auto idx2) {
+            return std::lexicographical_compare(input_data.begin() + idx1,
+                                                input_data.begin() + idx1 + chunk_sz,
+                                                input_data.begin() + idx2,
+                                                input_data.begin() + idx2 + chunk_sz);
+        };
+        std::map<size_t, size_t, decltype(idx_less_fn)> uniq_val_map(idx_less_fn);
+
+        std::tuple<std::vector<std::size_t>, std::vector<std::size_t>, std::vector<std::size_t>> rv;
+        auto& [y_indices, x_rev_indices, y_count] = rv;
```
pfultz2, line 174: This is really confusing, Just declare each vector and do `std::move` and `std::make_tuple` at the end.

lakhinderwalia: Returning the` rv `does an RVO. Thus avoid doing a move, or make_tuple.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Logically speaking, one shouldn't change a shape once defined. That is the reason it was done this way..

src/include/migraphx/op/unique.hpp
lakhinderwalia: Good idea. Thanks.

src/include/migraphx/op/unique.hpp
bpickrel: I shouldn't worry about that too much.  Shapes are disposable, lightweight objects, and the compute_shape() method is definitely the most appropriate place to do any wrangling of shapes it takes to create the desired output.  OTOH, either code version looks fine to me.

src/include/migraphx/op/unique.hpp
lakhinderwalia: std::tie will do an assignment operation to the existing variable. A move() is therefore preferred here.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Thanks. We should change shapes  within its implementation in shape.cpp/hpp.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Upon testing, I think tie will in this case just move() as well, i.e. implicitly. So you are right. I can use tie() here and avoid the extra lines of code.. Thanks.

src/include/migraphx/op/unique.hpp
pfultz2: > Logically speaking, one shouldn't change a shape once defined. That is the reason it was done this way..

This is not mutating the shape object(which is not even possible since all the methods to `shape` have to be `const`). Instead, this is just modifying `dout`, which is just a vector of dynamic dimensions.

src/include/migraphx/op/unique.hpp
pfultz2: I think you mean NRVO not RVO. Doing `std::make_tuple` will do RVO. I dont think the extra moves will be a major bottleneck here(and might even be optimzed out) so I would prefer clarity here.

lakhinderwalia: Yes, indeed it is NRVO. I will add a comment here why it is done.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Well firstly, a to_dynamic() is used, that is why I didn't want to mess with the results of it. Finally, I am not sure what is simpler here to use the suggested 3 lines for one line anyways.. -- and perhaps repeated for two sets of logic each.. In any case, are you proposing a value add of some kind here, or can it be left as it is otherwise. Thanks.

pfultz2: > Well firstly, a to_dynamic() is used, that is why I didn't want to mess with the results of it.

Why? I am not sure I follow this. 

> Finally, I am not sure what is simpler here to use the suggested 3 lines for one line anyways..

To be clear, the suggestion is to replace the above 3 lines not the last line. And you could probably make it 2 lines(if you are concerned about the number of lines) if you write it like:

```cpp
d_out = sh_x.to_dynamic().dyn_dims();
d_out.front() = shape::dynamic_dimension{1, max_uniq_ct};
```

So this suggestion is even shorter than what you have written. 

> and perhaps repeated for two sets of logic each..

Are you referring to the non-axis branch? I am not suggesting a change to the non-axis branch, so there is no repeat of the logic. 

>  In any case, are you proposing a value add of some kind here

This is to make it clear you are using the same dimensions from `x` for everything except the first dimension. Which also matches the pattern I suggested [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2072#discussion_r1400960751) except its using dynamic dimensions instead of integers.

src/include/migraphx/op/unique.hpp
lakhinderwalia: This brace is necessary here for compilation. No constructor is found for shape otherwise.. Thanks.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Messy use of to_dynamic() and to its dyn_dims(), and then an effort to modify it.  Instead of using the api to do dynamic shapes. Net value add  is negative, IMO.  However, it is done.

src/include/migraphx/op/unique.hpp
pfultz2: I dont see this change in the latest fixes you pushed.

src/include/migraphx/op/unique.hpp
pfultz2: Thanks, this will be much easier to tweak to support a different axis in the future as well(we just replace the `0` with `*axis`).

src/include/migraphx/op/unique.hpp
```diff
@@ -0,0 +1,321 @@
...
+                y_indices.push_back(x_rev_indices.size());
+            }
+            y_count[itr->second]++;
+            x_rev_indices.push_back(itr->second);
+        }
+
+        return rv;
+    }
+
+    // Axis. Default: none. Range: [-rank, rank-1]
```
pfultz2, line 195: So there is no normalization done for this, which could break optimization passes which will assume it is normalized. I dont think you need to enable normalization in this PR(since there are no graph transformation with `unique`), but we should have an issue open to track this.

Alternatively, we could also use `bool use_axis_0;` since we only support axis 0, and then in the future when we add support for different axes we can change this to `std::optional<int64_t> axis`.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
pfultz2: I dont see any test added to op_shape_tests.cpp. Are there still changes you haven't pushed yet?

src/include/migraphx/op/unique.hpp
lakhinderwalia: Sorry, I am referring to the use of lexicographic compare, and that is already added.

src/include/migraphx/op/unique.hpp
lakhinderwalia: It already is  std::optional<int64_t> axis.

src/include/migraphx/op/unique.hpp
pfultz2: Can you add this function though so you dont need to duplicate it twice?

src/include/migraphx/op/unique.hpp
pfultz2: Yes, but no normalization is done on it to remove the negative axis so we need to open an issue to fix that or bypass it completely by changing it to a simple `bool`.

lakhinderwalia: [Line 221. ](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/8cdf4aad321700d006312f2ad8852e27a130635d/src/include/migraphx/op/unique.hpp#L221)

pfultz2: That only normalizes it internally for `compute_shape`, the `axis` attribute is never updated(and `compute_shape` is not the appropriate place to update it), so any optimization passes that reads the `axis` attribute will read a potential negative axis. This can lead to bugs like in #2465.

Instead, we have a normalization process that will normalize operators' fields so such consumers(like optimization passes) dont need to tune the axis. This is done in `normalize_attributes` function(which is used by the `normalize_ops` pass and `normalized_operator()` method on instructions).  This can be enabled on an operator by providing `attributes()` with which fields you want normalized and how you want it normalized. You can look at something like `concat` to see how it is setup. Probably you can use something like `value::array{normalize_attribute::include_min}` as it is done in concat.

However, I am not asking to do that in this PR, since there is no optimization pass that currently looks for `unique`, but there could be in the future so we should open an issue to address it later.

lakhinderwalia: We just validate it in once in compute_shape, and from then onwards use axis=0, the only valid choice if axis were specified. Thanks.

pfultz2: > We just validate it in once in compute_shape, and from then onwards use axis=0,

Yes in `compute_shape`, but if an optimization pass(which are not part of this class) needs to read it in the future it can read either `-rank` or `0`. The optimization passes should not be tuning the axis, rather it should all be normalized to `0`.

> the only valid choice if axis were specified.

If thats the case, then it should only be the only value that can be specified, which is why I was suggesting to use `bool use_axis_0`, which makes invalid choices unrepresentable.

src/include/migraphx/op/unique.hpp
lakhinderwalia: Done.

src/include/migraphx/op/unique.hpp
lakhinderwalia: OK. We should address it in another PR for the optimization pass. Thanks.

src/include/migraphx/op/unique.hpp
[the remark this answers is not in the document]
lakhinderwalia: Sorry, I missed that. These tests are now added..

umangyadav (approved):

bpickrel (approved):
Approved.

src/include/migraphx/op/unique.hpp
pfultz2: This is still not resolved. Can you write it similiar to what I've shown? Using multiple statements in the `for` loop makes it harder to decipher whats happening. Using a range-based for loop makes it much simpler.

pfultz2 (changes requested):

src/include/migraphx/op/unique.hpp
lakhinderwalia: @pfultz2  Thanks for your suggestion. Both -- the current style and your suggested style -- are efficient ways to implement this small for-loop. 

Code gets complicated with number of lines in a loop, and its readability becomes an issue then. These 4 lines of for-loop (or 5 in your case) are really way too small for anybody to get confused, hence my suggestion is to just leave it as it is. Thanks.

pfultz2: Its much simpler to have single statements in the for loop. I will add a cppcheck rule for this in the future.

lakhinderwalia: Thanks. I suggest we instead check all loops for their size -- and more importantly for their complexity, when there are confusing branching and loop-size changes within that loop: the sentiment expressed by Sean Parent, that we both would readily agree to, I am sure.

pfultz2: > I suggest we instead check all loops for their size

We do have size checks for functions with [`readability-function-size`](https://clang.llvm.org/extra/clang-tidy/checks/readability/function-size.html) and would like to enable [`readability-function-cognitive-complexity`](https://clang.llvm.org/extra/clang-tidy/checks/readability/function-cognitive-complexity.html) but there is some code that doesnt pass this check.

However, these are just checking functions, it would nice to have these for loops as well. And even more so we want to check for multiple statements so develops dont hack it to work by putting more statements in the `for` to avoid complexity in the body.

PR #2079, opened 2023-08-16
Dynamic allocate

CharlieL7:
* Makes a version of `allocate` that takes in dimensions and allocates a buffer
* Going to create a simplify_dynamic_ops compiler pass that will use the `use_shape_attr` flag
* The ONNX op ConstantOfShape needs the buffer to be filled with a specific value, so going to make another PR for that and a fill operator


src/include/migraphx/op/allocate.hpp
```diff
@@ -36,20 +36,53 @@ namespace op {
...
     {
-        return {output_shape};
+        if(args.empty())
+        {
+            return {output_shape};
+        }
+        else
+        {
+            std::vector<std::size_t> output_dims(output_shape.ndim());
+            ;
```
umangyadav, line 82:
```suggestion
```

src/include/migraphx/op/allocate.hpp
```diff
@@ -36,20 +36,53 @@ namespace op {
...
-        migraphx::check_shapes{inputs, *this, true}.has(0);
-        return s;
+        migraphx::check_shapes{inputs, *this, true}.has(0, 1);
+        if(inputs.empty() or use_shape_attr)
+        {
+            return s;
+        }
+        else
+        {
+            const auto& out_dims = inputs.at(0);
```
umangyadav, line 64:
```suggestion
            const auto& out_shape = inputs.at(0);
```
It is of type shape not dims.

test/op_shape_test.cpp
```diff
@@ -81,6 +81,29 @@ void throws_shape(const migraphx::shape&, Ts...)
...
+    migraphx::shape input{migraphx::shape::int64_type, {2}};
+    auto max_val = std::numeric_limits<std::size_t>::max();
+    std::vector<migraphx::shape::dynamic_dimension> dyn_dims(
+        2, migraphx::shape::dynamic_dimension{0, max_val});
+    expect_shape(migraphx::shape{migraphx::shape::float_type, dyn_dims},
+                 migraphx::make_op("allocate", {{"buf_type", migraphx::shape::float_type}}),
+                 input);
+}
+
+TEST_CASE(allocate_dyn_use_shape_attr)
```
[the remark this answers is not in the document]
CharlieL7: > What about a nonempty input containing a 0? The last time I heard about it, we hadn't fully defined behavior for a shape with 0 length dimensions or no dimensions at all (empty lens vector).

Wouldn't make sense for this operation, no buffer allocation with a zero dimension. Unless we want to change how scalars work in MIGX.

umangyadav (approved):

src/include/migraphx/op/allocate.hpp
```diff
@@ -36,20 +36,52 @@ namespace op {
 struct allocate
 {
     shape s{};
+    // for dynamic allocate to set the buffer type
+    shape::type_t buf_type = shape::half_type;
+    // flag for dynamic allocate to use shape attr rather than [0, max] ranges in compute_shape()
+    // useful for optimizations later on
+    bool use_shape_attr = false;
```
pfultz2, line 43: This should store the dynamic dymension instead of using a boolean. It should be `std::vector<dynamic_dimensions>` and then if its not empy then use those dynamic dimensions for the shape.

CharlieL7: What difference does that make versus using the shape attribute?

test/ref/allocate.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <test.hpp>
+
+TEST_CASE(allocate_dyn)
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape s{migraphx::shape::int64_type, {4}};
+    auto out_dims = mm->add_parameter("out_dims", s);
+    mm->add_instruction(migraphx::make_op("allocate", {{"buf_type", migraphx::shape::float_type}}),
```
bpickrel, line 39: What is the behavior of a program where the author forgets to use allocate?  Is it possible to write a program that attempts to use a shape without allocating it on the GPU, or could a compiler pass make this sort of mistake?  Knowing what to look for would be helpful in catching future problems.

pfultz2 (approved):

PR #2082, opened 2023-08-17
Print warning about miopen_fusion while generating mxr

umangyadav:
MIOpen fusions are not serialized with tuned solutions. Print warnings for such  cases. 

```
root@463cb469625a:~/repo/AMDMIGraphX/build(fusion_warning)# ./bin/driver compile --model resnet50 --binary --output resnet50.mxr
Compiling ...
[WARNING]: Program has miopen_fusion instructions for which tuned solutions are not stored inside serialized MIGraphX program. Consider serializing with MIGRAPHX_DISABLE_MIOPEN_FUSION=1 flag set.
```


src/driver/main.cpp
```diff
@@ -475,13 +475,15 @@ struct compiler
             {
                 if(is_offload_copy_set(p) and not co.offload_copy)
                 {
-                    std::cout << "MIGraphX program was likely compiled with offload_copy set, Try "
-                                 "passing "
-                                 "`--enable-offload-copy` if program run fails.\n";
+                    std::cout
```
kahmed10, line 478: Should we use `std::cerr` here instead? Like the MLIR warnings

umangyadav: it is supposed to be a warning not an error.

src/verify_args.cpp
```diff
@@ -78,16 +78,6 @@ bool verify_args(const std::string& name,
             if(verify::range_zero(target))
                 std::cout << "Target data is all zeros" << std::endl;
 
-            // auto mxdiff = max_diff(ref, target);
```
kahmed10, line 81: is this needed?

kahmed10 (approved):

PR #2087, opened 2023-08-18
`fill` ref operator

CharlieL7:
* Implements a `fill` operator that sets the values in an output buffer to a given value
* Will be used when parsing ONNX `ConstantOfShape`
* Can also be used when a buffer needs to be filled with a value that is determined at runtime


umangyadav (approved):

CharlieL7:
There's a problem with checking if the `default_value` is a scalar. Scalar in MIGX is if all the strides = 0. I think this ONNX op rather wants there to be only one element. Hitting into a problem with how `ConstantOfShape` is parsed.

src/include/migraphx/op/fill.hpp
```diff
@@ -0,0 +1,71 @@
...
+struct fill
+{
+    std::string name() const { return "fill"; }
+
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this, true}.has(2).same_type();
+        // check that the shape of default_value is static
+        check_shapes{
+            inputs.begin(), inputs.begin() + 1, std::string("FILL (literal/value input)"), false};
```
pfultz2, line 50: Why are you passing a custom name?

CharlieL7: For a different prefix to the error message to make it more clear what's wrong.

src/include/migraphx/op/fill.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * has a static shape this operator can be replaced with a literal.
+ */
+struct fill
+{
+    std::string name() const { return "fill"; }
+
+    shape compute_shape(std::vector<shape> inputs) const
+    {
+        check_shapes{inputs, *this, true}.has(2).same_type();
+        if(inputs.at(0).dynamic() or inputs.at(0).elements() != 1)
```
pfultz2, line 48: Why would a static shape allow `elements != 1`? The way you implemented `compute` it can only be one element.

src/include/migraphx/op/fill.hpp
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return inputs.back();
+    }
+
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    {
+        visit_all(args[0], args[1])([&](auto value, auto output) {
+            par_for(dyn_out.computed_shape.elements(), [&](auto i) { output[i] = value.front(); });
+        });
+        return args[1];
```
pfultz2, line 60: You need to add an `output_alias` method,

src/include/migraphx/op/fill.hpp
CharlieL7: I'm not sure I understand your question. I changed it from using `check_shapes` to check if the first input was static because tidy didn't like it. I'm checking to make sure the input is static and one element.

TedThemistokleous (approved):

PR #2102, opened 2023-08-21
Ensure same layout is used for miopen convolution

pfultz2:


src/include/migraphx/check_shapes.hpp
```diff
@@ -228,6 +228,13 @@ struct check_shapes
         return *this;
     }
 
+    const check_shapes& same_layout() const
```
umangyadav, line 231: add docstring similar to other functions

kahmed10 (approved):

src/include/migraphx/check_shapes.hpp
umangyadav: @pfultz2  can you please make these changes ?

pfultz2: I already did.

umangyadav (approved):

PR #2103, opened 2023-08-21
Add environment variable to enable mlir exhaustive tuning configs

pfultz2:


src/targets/gpu/mlir.cpp
```diff
@@ -69,6 +69,7 @@ inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_MLIR);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_TUNE_EXHAUSTIVE);
```
umangyadav, line 72: should we tie this to `exhaustive_tune` compile option as well ?

krzysz00: I don't think so. The default of exhaustive tuning should be our "full", with full-on exhaustive being a debug feature

bpickrel:
I think assigning this to me was a typo.

krzysz00 (approved):
LGTM

PR #2104, opened 2023-08-21
Update time op to more accurately get device time

pfultz2:
Attaching events to the launch kernel has a lot of overhead. Instead run the op in the loop, and then add a hip event at the end of the loop.


src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -323,12 +323,12 @@ struct context
         return std::make_pair(nullptr, nullptr);
     }
 
-    float get_elapsed_ms() const
+    static float get_elapsed_ms(hipEvent_t start, hipEvent_t stop)
```
umangyadav, line 326: Don't we want to keep non-static method if we need in future ?

src/targets/gpu/compile_ops.cpp
```diff
@@ -185,8 +185,7 @@ struct compile_plan
             results.begin(), results.end(), std::back_inserter(times), [&](const auto& cr) {
                 if(not cr.has_value())
                     return std::numeric_limits<double>::max();
-                return time_op(*ctx, cr->replace.code_object, to_shapes(cr->ins->inputs()), 20)
-                    .first;
```
umangyadav, line 189: So earlier it was taking `host_time` for comparsion ? was that incorrect ?

src/targets/gpu/include/migraphx/gpu/context.hpp
pfultz2: I dont think we can overload a static and non-static function. If we do need to the method in the future then we can think about hot to name this better. Most likely we wont need it in the future as launching with the events attached has too much overhead and probably wont be fixed soon.

src/targets/gpu/compile_ops.cpp
umangyadav: Was it not accurate because `enable_perf_measurements()` was recording start and stop events back to back?

src/targets/gpu/time_op.cpp
```diff
@@ -51,21 +50,19 @@ time_op(context& ictx, operation op, const std::vector<shape>& inputs, int n)
     auto output           = op.compute_shape(inputs);
     op.finalize(ctx, output, inputs);
     auto args = generate_arguments(inputs);
-    auto run  = [&] {
-        op.compute(ctx, output, args);
-        ctx.finish();
-    };
-    gctx.enable_perf_measurement();
```
umangyadav, line 58: `enable_perf_measurement` is not being used anywhere now.

src/targets/gpu/compile_ops.cpp
pfultz2: > Was it not accurate because enable_perf_measurements() was recording start and stop events back to back?

No, it didnt measure the host_time with `enable_perf_measurements()` set. 

The problem is each host_time run includes the launch overhead, which might vary. Instead we run the kernel without syncing until the end to minimize the launch overhead due to a larger queue depth.

src/targets/gpu/time_op.cpp
```diff
@@ -51,21 +50,19 @@ time_op(context& ictx, operation op, const std::vector<shape>& inputs, int n)
...
     {
         (void)i;
-        host_time += time<milliseconds>(run);
-        device_time += gctx.get_elapsed_ms();
+        run();
     }
-    return std::make_pair(host_time / n, device_time / n);
+    gctx.get_stream().record(stop.get());
+    gctx.finish();
+    return context::get_elapsed_ms(start.get(), stop.get()) / n;
```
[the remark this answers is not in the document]
pfultz2: I am not sure why it wouldn't work, but either case we wouldn't be using mult-stream execution to time operators.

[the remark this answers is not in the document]
umangyadav: it wouldn't work with multi-stream i think because `hipEventElapsedTime` records time between start and stop events. both those events are put on the default stream.  If any of the op was using some other streams then it wouldn't measure correctly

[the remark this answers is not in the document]
pfultz2: The only op that should change the default stream would be `gpu::set_stream`, which it doesnt make sense for us to do timing for.

causten:
this fails when I tested it with mlir

root@ce81174f8f3c:/workspace/AMDMIGraphX# MIGRAPHX_ENABLE_MLIR=1 migraphx-driver perf  /models/mlperf/resnet50_v1.onnx
Compiling ...
Reading: /models/mlperf/resnet50_v1.onnx
Benchmarking gpu::mlir_op: 21 configs
terminate called after throwing an instance of 'migraphx::version_2_7_0::exception'
  what():  /workspace/AMDMIGraphX/src/targets/gpu/include/migraphx/gpu/context.hpp:333: get_elapsed_ms: Failed hipEventElapsedTime: invalid resource handle
Aborted (core dumped)

pfultz2:
> this fails when I tested it with mlir

Fixed.

umangyadav (approved):

PR #2105, opened 2023-08-22
Add support for the dilations attribute to Pooling ops

mirza-halilcevic:
- Add parsing support for the `dilations` attribute
- Modify reference implementation
- Resolves migraphx-benchmark/AMDMIGraphX#70


umangyadav:
For such changes, it would require tests inside 
`ref_ops_test.cpp` which tests whether reference implementation inside MIGraphX is correct or not. 
`op_shape_test.cpp`.  to test for the `compute_shape()` part 
`onnx_test.cpp` to test whether MIGraphX is parsing operator correctly or not. 

and verify tests. which would run the operator/program on GPU and compare it with reference implementation.

src/include/migraphx/op/pooling.hpp
```diff
@@ -98,12 +100,15 @@ struct pooling
...
         {
             MIGRAPHX_THROW("POOLING: inconsistent attribute sizes");
         }
-        if(std::any_of(lengths.begin(), lengths.end(), [&](auto i) { return (i == 0); }) or
-           std::any_of(stride.begin(), stride.end(), [&](auto i) { return (i == 0); }))
+
+        const auto is_zero = [](auto el) { return el == 0; };
+        if(std::any_of(lengths.begin(), lengths.end(), is_zero) or
+           std::any_of(stride.begin(), stride.end(), is_zero) or
+           std::any_of(dilations.begin(), dilations.end(), is_zero))
```
umangyadav, line 111: change throw message

pfultz2:
> Looks like MIOpen doesn't support dilations so we won't have GPU implementation.

@umangyadav Is there a way we can insert some other operators to get the same dilation result on the GPU, similar to how we do for non-symmetrical padding?

attila-dusnoki-htec:
cc @pfultz2 
I looked at "step" op, but still did not managed to figure out what you meant. :disappointed: 
Instead I ended up collecting the axes, and creating the input with gather.
It works, but has a serious limitation: the axes is a vector and can become too large, which results in bad alloc.

pfultz2:
> I looked at "step" op, but still did not managed to figure out what you meant.

You would insert a step operator with the same value as the dilation, and using the axes for the image. Like `make_op("step", {{"axes", {2, 3}}, {"steps", {dilation_for_h, dilation_for_w}}}).

src/include/migraphx/op/pooling.hpp
```diff
@@ -366,6 +377,16 @@ struct pooling
 
             // for each element in the window...
             shape_for_each(win_shape, [&](const auto& idx_w) {
+                // Skip elements that belong to the dilated area
+                for(size_t axis = 0; axis < idx_w.size(); ++axis)
+                {
+                    if(idx_w[axis] % dilations[axis])
+                    {
+                        pool_size -= 1;
```
pfultz2, line 385: `pool_size` should be adjusted before calling `shape_for_each`.

src/include/migraphx/op/pooling.hpp
```diff
@@ -390,7 +411,15 @@ struct pooling
                     // this is a padding element.  Padding locations
                     // don't contribute to average or max pooling total but can play in
                     // lpnorm pooling.
-                    output_val = op(output_val, 0);
```
pfultz2, line 393: This should just be `output_val = op(output_val, op.template init<Type>());`. Using `0` is a bug.

src/include/migraphx/rewrite_pooling.hpp
```diff
@@ -39,6 +40,10 @@ struct MIGRAPHX_EXPORT rewrite_pooling
 {
     std::string name() const { return "rewrite_pooling"; }
     void apply(module& m) const;
+
+    private:
+    void replace_with_reduce(module& m, instruction_ref ins) const;
+    void replace_dilations_with_gather_pooling(module& m, instruction_ref ins) const;
```
pfultz2, line 46: These should be static functions defined in rewrite_pooling.cpp file.

src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -225,6 +225,15 @@ inline pooling_descriptor make_pooling(const migraphx::op::pooling& op)
...
         MIGRAPHX_THROW(ss.str());
     }
+    if(not std::all_of(
+           op.dilations.cbegin(), op.dilations.cend(), [](std::size_t d) { return d == 1; }))
+    {
+        std::stringstream ss("Unsupported dilations for pooling: [");
+        for(const auto& d : op.dilations)
+            ss << d << ",";
+        ss << "]";
+        MIGRAPHX_THROW(ss.str());
```
pfultz2, line 235: Just do `MIGRAPHX_THROW("Unsupported dilations for pooling: [" + to_string_range(op.dilations) + "]")` instead.

attila-dusnoki-htec:
Addressed requested changes, except this:
> `pool_size` should be adjusted before calling `shape_for_each`.

I tried to calculate it beforehand, but did not managed to resolve the dyn pad cases properly :disappointed:  (`avgpool_dyn_auto_pad_test`, `avgpool_dyn_pad_ceil_test`)
The current approach is the most straighforward. But i can try again if that is a blocker.

test/rewrite_pooling_test.cpp
```diff
@@ -80,6 +81,483 @@ TEST_CASE(rewrite_pooling_test)
                  migraphx::make_op("reduce_max", {{"axes", {2, 3, 4}}}));
 }
 
+TEST_CASE(rewrite_pooling_dialtions_test)
```
CharlieL7, line 84:
```suggestion
TEST_CASE(rewrite_pooling_dilations_test)
```

test/rewrite_pooling_test.cpp
```diff
@@ -80,6 +81,483 @@ TEST_CASE(rewrite_pooling_test)
...
+        -0.4136, 0.5085,  0.4136,  -0.6439, -1.1953, -0.406,  -0.0195, 0.1869,  -0.8664, 1.1364,
+        0.5041,  0.0647,  0.1941,  -1.0819, -0.4629, -0.5107, 0.3612,  -0.3583};
+    auto l0 = mm->add_literal(migraphx::literal{s, data});
+    mm->add_instruction(op, l0);
+    opt_pooling(*mm);
+    p.compile(migraphx::make_target("ref"));
+    auto result = p.eval({}).back();
+    std::vector<float> results_vector;
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<float> gold{0.7812, 1.0449, 2.7666, 2.6859};
```
CharlieL7, line 519: How are you getting these gold values? Want to make sure that we know these are correct.

pfultz2 (approved):

PR #2110, opened 2023-08-23
Added support for standalone mlir-conv

ravil-mobile:


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -314,6 +315,58 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
 #endif
 }
 
+#ifdef MIGRAPHX_MLIR
+
+namespace {
+MIGRAPHX_PRED_MATCHER(is_supported_arch, instruction_ref)
+{
+    // TODO(ravil): debug
+    static std::unordered_set<std::string> supported_consumer_archs{
```
krzysz00, line 324: This list isn't right.

Also, I'm pretty sure we're only meant to offload on `gfx110x`

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -314,6 +315,58 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
...
+}
+
+struct find_mlir_standalone_convolution_op
+{
+    auto matcher() const { return match::name("convolution")(is_supported_arch); }
+
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
+        auto conv_based_op = r.result;
+        // Only fuse with fp32/fp16
```
krzysz00, line 341: Will we need this restriction?

krzysz00: Also, comment wrong

src/targets/gpu/target.cpp
```diff
@@ -143,6 +144,7 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
 #endif
         dead_code_elimination{},
         enable_pass(mlir_enabled(), fuse_mlir{&ctx}),
+        enable_pass(mlir_enabled(), standalone_mlir{&ctx}),
```
krzysz00, line 147: Why's this a separate pass instead of being a rewrite within `fuse_mlir()` (which could be renamed to something like `mlir_offload()`)?

What do folks think here?

ravil-mobile: It is a good question. @manupak and I was discussing this today - i.e., whether to add use one pass or two ones. My argument was that adding `standalone conv` to  `fuse_mlir` was ambiguous. @manupak suggested that we could rename `fuse_mlir` as you mentioned. 

By and large, yes. We can rename `fuse_mlir` to `mlir_offload` which sounds more generic.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -314,6 +315,58 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
...
+    }
+};
+} // namespace
+
+#endif // MIGRAPHX_MLIR
+
+void standalone_mlir::apply(module_pass_manager& mpm) const
+{
+#ifdef MIGRAPHX_MLIR
+    match::find_matches(mpm, find_mlir_standalone_convolution_op{});
```
pfultz2, line 364: This should be moved to `fuse_mlir::appy`.

src/targets/gpu/fuse_mlir.cpp
ravil-mobile: >This list isn't right.

You are correct. I just used it to test the concept on MI200 because I am still waiting to get access to a Navi machine.

From what I understood, `gfx11*` or `gfx110*` denote Navi3X GPUs. Maybe we can use a regex to capture current and future Navi cards?

Yi just updated me a few hours ago that gfx1100 for Navi31, gfx1101 for Navi32

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -314,6 +315,58 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
 #endif
 }
 
+#ifdef MIGRAPHX_MLIR
+
+namespace {
+MIGRAPHX_PRED_MATCHER(is_supported_arch, instruction_ref)
```
pfultz2, line 321: I dont think this needs to be a matcher as it is not reading anything from the graph. It can be a vanilla function and we can call it to conditionally call `match::find_matches(mpm, find_mlir_standalone_convolution_op{})`.

src/targets/gpu/fuse_mlir.cpp
ravil-mobile: Got you. So, the I will remove
https://github.com/ravil-mobile/AMDMIGraphX/blob/63bdf8f04ab416d20023c7ac2acbb26df31172f3/src/targets/gpu/include/migraphx/gpu/standalone_mlir.hpp#L1-L47

src/targets/gpu/fuse_mlir.cpp
ravil-mobile: Thanks! It something that we were discussing with @manupak today and came to the same conclusion. I will rework it

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -314,6 +315,58 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
...
+{
+    // TODO(ravil): debug
+    static std::unordered_set<std::string> supported_consumer_archs{
+        "gfx900", "gfx906", "gfx908", "gfx1030", "gfx940"};
+
+    // static std::unordered_set<std::string> supported_consumer_archs{"gfx1030"};
+    const auto device_name = trim(split_string(get_device_name(), ':').front());
```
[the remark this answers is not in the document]
ravil-mobile: I can see. Let me give a try.

manupak:
I have no major concerns - I ll let maintainers approve it though coz I had some input into the code here.

Only concern is this code is not tested at the minute. 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/6f1c947f686ee7e7265d14c6f416f4b82eec8815/Jenkinsfile#L125-L134

@pfultz2  since we add the flag to force standalone convs; we can test (the integration issues) in cdna using that flag OR do ya'll have a Navi nodes to test this proper ?

src/targets/gpu/CMakeLists.txt
```diff
@@ -109,7 +109,7 @@ add_library(migraphx_gpu
     compiler.cpp
     device_name.cpp
     fuse_ck.cpp
-    fuse_mlir.cpp
+    mlir_offload.cpp
```
pfultz2, line 112: This shouldn't be renamed in this PR. fuse_mlir is consistent with the other fuse passes. So if we rename this like this than we should rename all the other fuse passes. However, such refactoring should go into a separate PR.

ravil-mobile: So, should I rename 1) just file names or 2) file names + the corresponding structures?

ravil-mobile: At this moment, I just changed file names. Let me know if I should to adapt the stuct names as well

src/targets/gpu/include/migraphx/gpu/mlir_offload.hpp
```diff
@@ -35,15 +35,15 @@ namespace gpu {
 
 MIGRAPHX_GPU_EXPORT bool mlir_enabled();
 
-struct MIGRAPHX_GPU_EXPORT fuse_mlir
+struct MIGRAPHX_GPU_EXPORT mlir_offload
```
pfultz2, line 38: The class should be named `fuse_mlir`.

src/targets/gpu/CMakeLists.txt
pfultz2: > I should to adapt the stuct names as well

Yes, the class names should match the filenames.

src/targets/gpu/include/migraphx/gpu/mlir_offload.hpp
ravil-mobile: But then it sounds as if we only provide fused operation. However, this PR aims at additionally providing standalone convs.

https://github.com/ravil-mobile/AMDMIGraphX/blob/81aa21794803cdd73e5c88dfb7bf459914b74d8a/src/targets/gpu/fuse_mlir.cpp#L335-L349

Moreover, we are going to add standalone GEMMs. Additionally, we would like to have an option to switch off fused ops based on the value of an env. variables. In this case, we would like to completely fall back to standalone GEMMs and CONVs taking them from rocMLIR; we think it may be help us during our performance analyses and monitoring.

By and large, the `apply` method is not supposed to be focused on fused operations only; it is supposed to be generic. That is the reason why @krzysz00 and I proposed to rename the struct `mlir_offload`.

pfultz2: > But then it sounds as if we only provide fused operation.

For fuse_ck, we handle standalone gemms, for fuse_pointwise we handle standalone pointwise, and fuse_reduce we handle standalone reduce. So I dont see how it would be different here.

> By and large, the apply method is not supposed to be focused on fused operations only; it is supposed to be generic.

Yea niether do fuse_ck, fuse_pointwise, and fuse_reduce focus on fused operations only. 

> That is the reason why @krzysz00 and I proposed to rename the struct mlir_offload.

Yes, but such renaming should go in a separate PR, because the other fuse passes need to renamed as well.

ravil-mobile: > Yes, but such renaming should go in a separate PR, because the other fuse passes need to renamed as well.

Got your point! Ok, I will rework.

ravil-mobile: done

Issue #75, opened 2018-10-01
Failures reading ONNX files


mvermeulen:
This issue partially to document what I see trying model zoo models and partially to sync on specific versions.  Here is what I see trying each of the five designated models with code as of 10/1:

# resnet50 - using https://github.com/onnx/models/tree/master/resnet50, release 1.3
   Produces list of nodes w/o complaint

# inception_v2 - using https://github.com/onnx/models/tree/master/inception_v2, release 1.3
@481 = @literal{ ... } -> float_type, {64}, {1}
@482 = @literal{ ... } -> float_type, {64}, {1}
@483 = @literal{ ... } -> float_type, {64}, {1}
@484 = @literal{ ... } -> float_type, {64}, {1}
@485 = @literal{ ... } -> float_type, {64, 3, 7, 7}, {147, 49, 7, 1}
data_0 = @param:data_0 -> float_type, {1, 3, 224, 224}, {150528, 50176, 224, 1}
@487 = convolution[padding={3, 3}, stride={2, 2}, dilation={1, 1}](data_0,@485) -> float_type, {1, 64, 112, 112}, {802816, 12544, 112, 1}
@488 = batch_norm_inference(@487,@484,@483,@482,@481) -> float_type, {1, 64, 112, 112}, {802816, 12544, 112, 1}
@489 = unknown:Unsqueeze(@480) -> float_type, {64}, {1}

terminate called after throwing an instance of 'migraph::exception'
  what():  /home/mev/source/MIGraph/src/include/migraph/check_shapes.hpp:66: Dimensions do not match
Aborted (core dumped)

# mobilenet - using https://github.com/onnx/models/tree/master/models/image_classification/mobilenet
   Passes without complaint, but does give one unknown operator:

mev@cafayate:~/source/MIGraph/build$ ./src/onnx/read_onnx /home/mev/dockerx/models/mobilenetv2-1.0/mobilenetv2-1.0.onnx  | grep unknown
@420 = unknown:GlobalAveragePool(@419) -> float_type, {1, 1280, 7, 7}, {62720, 49, 7, 1}

# mnist - using https://github.com/onnx/models/tree/master/mnist, onnx version 1.2
@0 = @literal{-0.044856, 0.00779166, 0.0681008, 0.0299937, -0.12641, 0.140219, -0.0552849, -0.0493838, 0.0843221, -0.0545404} -> float_type, {1, 10}, {10, 1}
[...]

PR #2120, opened 2023-08-24
Add support for axes input to reduce operators

music-dino:
- Add parsing support for axes as a non-constant input
- Update the reference implementation to handle this case
- Resolves migraphx-benchmark/AMDMIGraphX#68


src/include/migraphx/op/reduce_op.hpp
```diff
@@ -80,11 +80,12 @@ template <class Derived>
 struct reduce_op : op_name<Derived>
 {
     std::vector<std::int64_t> axes{};
+    bool noop_with_empty_axes = false;
```
pfultz2, line 83: We want to make reduce_op to always be a noop with empty axes. For this PR, just skip the handling of empty axes for now. In a future PR, we can update the onnx parser to insert extra operators for empy axes that is not a noop.

music-dino: Do you want the empty axes handling to be skipped, or changed even in the case when axes are constant or an attribute?
In the current code, that is handled according to ONNX spec.

umangyadav:
@music-dino  can you fix tidy ? and resolve merge issues ?

src/include/migraphx/op/reduce_op.hpp
pfultz2: This field should be removed. And we should by default be a noop wit empty axes and the parser can insert extra operators to handle the empty case for the ONNX spec.

music-dino: In the constant axes case the parser already handles that, by not inserting an operator at all, and just passing the input back as the output.
In the variable axes case it cannot be known until runtime whether a noop will occur until the runtime value of axes is checked. 
So which operators could the parser insert in the variable axes case to handle it. The If operator?

pfultz2: This field needs to be removed from the operator.

> In the variable axes case it cannot be known until runtime whether a noop will occur until the runtime value of axes is checked.  So which operators could the parser insert in the variable axes case to handle it. The If operator?

We can insert operators such as `equal` to detect this and then insert a `where` operator to pick between the reduce and noop.

music-dino: The Where would only have to be inserted when noop_with_empty_axes is true, and be predicated on the size of the variable axes input. For empty axes the input tensor would need to be returned, while for non-empty axes a reduce would be performed. 
This leads to an issue with Where due to the difference of output dimensions of the two branches, given that one would have the shape of the input, and the other would have whichever shape the variable input axes dictate.

pfultz2: > The Where would only have to be inserted when noop_with_empty_axes is true

`reduce_op` should be a noop with empty axes, by default. So then when `noop_with_empty_axes=false` the `where` would take two different reduces: `is_empty(axes) ? reduce_op(x, computed_axes) : reduce_op(x, axes)`. For both branches the axes will be passed as a variable input(even though the empty case will be taking a literal) so they produce the same output shape.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -173,24 +202,56 @@ struct reduce_op : op_name<Derived>
             static_cast<const Derived&>(*this).output(batch_shape)(val);
     }
 
-    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    argument compute(const shape& computed_shape,
+                     const std::vector<int64_t>& reduce_axes,
+                     argument& data_arg) const
```
CharlieL7, line 207: Can we name this function something else so we're not getting into possible future trouble with how operation.hpp handles the compute function?

src/include/migraphx/op/reduce_op.hpp
CharlieL7: Has this issue been resolved yet?

CharlieL7 (commented):
Not sure what's the status on this PR.

src/include/migraphx/op/reduce_op.hpp
music-dino: Hi @CharlieL7, not yet, I'd been on vacation for most of December so the PR was left unattended.

src/include/migraphx/op/reduce_op.hpp
music-dino: Sure.

src/onnx/parse_reduce_op.cpp
```diff
@@ -31,66 +31,123 @@ namespace migraphx {
...
+            auto all_axes_lit      = info.add_literal(
+                literal{shape{shape::type_t::int64_type, {all_axes.size()}}, all_axes});
+            auto reduce_all_axes = info.add_instruction(reduce_op, args[0], all_axes_lit);
+            auto zero      = info.add_literal(literal{shape{shape::type_t::int64_type}, {0u}});
+            auto axes_size = info.add_instruction(make_op("dimensions_of", {{"end", 1}}), args[1]);
+            auto is_axes_empty = info.add_instruction(make_op("equal"), axes_size, zero);
+            auto is_axes_empty_bc =
+                info.add_instruction(make_op("multibroadcast"), is_axes_empty, reduce_all_axes);
+
+            return info.add_instruction(
```
music-dino, line 99: @pfultz2 I've implemented operator parsing in the fashion you recommended. The `where` here will work during parsing, due to dynamic output shape equivalence between branches. However, when evaluation time comes around, the output shapes will differ which causes `where` to throw an exception(demonstrated by this [test](https://github.com/ROCm/AMDMIGraphX/pull/2120/files#diff-d6b0a0ea0ed42a315a1748fdd544a06facad4a564d07ff588bb9233348db615eR47)).

test/onnx/verify/reducesum_variable_dynamic_axes_test.cpp
```diff
@@ -0,0 +1,48 @@
...
+    parameter_map pm;
+    shape x_shape{shape::float_type, {2, 2, 2}};
+    std::vector<float> x(x_shape.elements());
+    std::iota(x.begin(), x.end(), 0);
+    pm["x"] = argument(x_shape, x.data());
+
+    std::vector<int64_t> axes{1};
+    pm["axes"] = argument(shape{shape::int64_type, {1}}, axes.data());
+
+    auto result = p.eval(pm);
```
music-dino, line 47: @pfultz2 @CharlieL7 

This test demonstrates the Where issue that crops up([here is the code path it tests](https://github.com/ROCm/AMDMIGraphX/pull/2120/files#diff-e2bd4cc9d4b1d5365a82126bf3a7bd6a4b317c30c4a2251cf801b9c3ae72fcd3R87)) we discussed last week.

All three input shapes of Where being the same({1, 2}, {1, 2}, {1, 2}) parsing will succeed.
However, when attempting to evaluate the graph issues pop up.
The shapes being reified, the all axes reduction will have an output shape of {1, 1, 1}, the other one {2, 1, 2}, causing [this check to fail](https://github.com/migraphx-benchmark/AMDMIGraphX/blob/develop/src/include/migraphx/op/where.hpp#L45).

pfultz2: The `compute` method to `where` needs to be updated to remove the `dyn_output` usage so the compute shape can happen lazily. For scalar condition we can pick either branch regardless of the shape difference:

```cpp
    argument compute(shape output_shape, std::vector<argument> args) const
    {
        if(args[0].scalar())
            return args[args[0].at<bool>() ? 1 : 2].copy();
        if(output_shape.dynamic())
            output_shape = compute_shape(to_shapes(args));
        visit_all(result, args[1], args[2])([&](auto output, const auto x, const auto y) {
            args[0].visit([&](const auto condition) {
                par_for(dyn_out.computed_shape.elements(),
                        [&](auto i) { output[i] = condition[i] ? x[i] : y[i]; });
            });
        });
        return result;
    }
```

test/ref/reduce_mean.cpp
```diff
@@ -62,6 +87,31 @@ TEST_CASE(reduce_mean_axis1)
     EXPECT(results_vector == gold);
 }
 
+TEST_CASE(reduce_mean_variable_axis1)
```
lakhinderwalia, line 90: Is a test on `axis = 1` possible with a multidimensional input, with` keep_dims=1`? For example, this is what I am cutting and pasting from Netron.  And/or -- but near identical -- test -- with `axis = -2` instead of 1. Thanks.

```
shape = [3, 2, 2]
axes = np.array([1], dtype=np.int64)
keepdims = 1

node = onnx.helper.make_node(
    "ReduceMean",
    inputs=["data", "axes"],
    outputs=["reduced"],
    keepdims=keepdims,
)

data = np.array(
    [[[5, 1], [20, 2]], [[30, 1], [40, 2]], [[55, 1], [60, 2]]],
    dtype=np.float32,
)
reduced = np.mean(data, axis=tuple(axes), keepdims=keepdims == 1)
# print(reduced)
# [[[12.5, 1.5]]
# [[35., 1.5]]
# [[57.5, 1.5]]]

expect(
    node,
    inputs=[data, axes],
    outputs=[reduced],
    name="test_reduce_mean_keepdims_example",
)

np.random.seed(0)
data = np.random.uniform(-10, 10, shape).astype(np.float32)
reduced = np.mean(data, axis=tuple(axes), keepdims=keepdims == 1)

expect(
    node,
    inputs=[data, axes],
    outputs=[reduced],
    name="test_reduce_mean_keepdims_random",
)
```

test/ref/reduce_mean.cpp
```diff
@@ -109,3 +209,28 @@ TEST_CASE(reduce_mean_int)
...
+    pm["x"] = migraphx::argument(x_shape, x_arg.data());
+    std::vector<int64_t> axes_arg{1, 2};
+    pm["axes"]  = migraphx::argument(axes_shape, axes_arg.data());
+    auto result = p.eval(pm).back();
+    std::vector<int32_t> results_vector;
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+
+    std::vector<int32_t> gold{2, 6, 10};
+    EXPECT(results_vector == gold);
+}
```
lakhinderwalia, line 236: If there is not a test for default axes, with keepdims=1, the following (copied from Netron) would be a good test. Thanks.

```
shape = [3, 2, 2]
axes = np.array([], dtype=np.int64)
keepdims = 1

node = onnx.helper.make_node(
    "ReduceMean",
    inputs=["data", "axes"],
    outputs=["reduced"],
    keepdims=keepdims,
)

data = np.array(
    [[[5, 1], [20, 2]], [[30, 1], [40, 2]], [[55, 1], [60, 2]]],
    dtype=np.float32,
)
reduced = np.mean(data, axis=None, keepdims=keepdims == 1)
# print(reduced)
# [[[18.25]]]

expect(
    node,
    inputs=[data, axes],
    outputs=[reduced],
    name="test_reduce_mean_default_axes_keepdims_example",
)
```

lakhinderwalia (commented):
Thanks for all your commits/responses so far. I have submitted some additional comments. PTAL.

test/ref/reduce_mean.cpp
lakhinderwalia: Please note, the dimension of the output tensor has to be verified as well as the value.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -94,16 +94,57 @@ struct reduce_op : op_name<Derived>
...
+
+    shape compute_static_shape(const std::vector<shape>& inputs) const
+    {
+        const auto& data_shape = inputs[0];
+        if(axes.empty())
+        {
+            std::vector<shape::dynamic_dimension> dims(data_shape.ndim());
+            auto lens = data_shape.lens();
+            std::transform(lens.begin(), lens.end(), dims.begin(), [](auto l) {
+                return shape::dynamic_dimension{1, l};
```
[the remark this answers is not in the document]
music-dino: I've changed the variable name from l to len, to make the code more legible.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -173,24 +201,43 @@ struct reduce_op : op_name<Derived>
...
     }
 
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    {
+        auto&& data_arg = args[0];
+        // cppcheck-suppress knownConditionTrueFalse
+        if(not axes.empty())
+            return reduce(dyn_out.computed_shape, axes, data_arg);
+
+        if(args[1].empty())
```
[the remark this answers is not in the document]
music-dino: The axes member will be empty when the parser was unable to parse an axes value either as an onnx node attribute or as a constant node input. The axes member being empty signifies to the operator that it should expect axes as a second input. This second input then might or might not be empty.

src/onnx/parse_reduce_op.cpp
```diff
@@ -31,66 +31,110 @@ namespace migraphx {
...
+            return std::nullopt;
 
-    if(keep_dims == 1)
-    {
-        return info.add_instruction(make_op(op_name, {{"axes", axes}}), args.front());
+        return parser.parse_value(info.attributes[attribute_name]).at<T>();
     }
-    else
+
+    std::optional<std::vector<int64_t>> parse_constant_axes(std::vector<instruction_ref>& args,
```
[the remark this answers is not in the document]
music-dino: The optional being empty and the contained vector being empty have different meanings. When the optional is empty it signifies that a value couldn't be parse either as an attribute or a constant input, which then means that we should look for axes as a second, variable input. 
When the optional contains a vector, it signifies that we were able to parse a value, and it being empty is a valid option. 
So the optional is needed to differentiate these three possibilities:

1. Couldn't parse value - empty optional
2. Parsed an empty value - optional containing an empty vector
3. Parsed a non empty value - optional containing a non empty vector.

test/ref/reduce_mean.cpp
music-dino: I'm not quite catching your drift here. As for keepdims, that cannot be tested in the reference tests, as it is not a part of the operator itself, but is accomplished by inserting a squeeze operator after the reduction.

lakhinderwalia: I am not sure, in a black box testing it matters if there one or instead two moving parts inside.. It is the results that matter, and so if the squeeze operator is getting inserted, that is fine by me.. 

Maybe I am missing something here? Thanks.

src/onnx/parse_reduce_op.cpp
```diff
@@ -31,66 +31,110 @@ namespace migraphx {
...
+            return reduce;
+        }
+
+        // Handle variable input axes
+        if(keep_dims == 0)
+            MIGRAPHX_THROW("Keepdims not supported with runtime provided axes");
+
+        // Empty axes attribute indicates to the operator to look for axes in the inputs
+        auto reduce_op = make_op(op_name, {{"axes", {}}});
+
```
lakhinderwalia, line 76: Pardon me, I am a little confused here.. if the operator were to look for `axes` in its variable inputs (line #75) as per the comment above it, then how is line #77 below able to deduce that input `axis` is _empty_ -- that is the only case where the `noop_with_empty_axis` should activate, similar to line #55. Thanks.

music-dino: The op does not form any assumptions about input axes at this point. This is delegated to the [compute method](https://github.com/migraphx-benchmark/AMDMIGraphX/blob/5f7b702f88e13dc97ae8be108c11031046b5e1f9/src/include/migraphx/op/reduce_op.hpp#L231). There, if the variable axes are empty it will just return its input, if they're not it will proceed as usual(essentially it always presumes that noop_with_empty_axes is true)

It is worth nothing that following this code path is the only way in which an empty variable axes input can reach the operator. In the subsequent parsing code it can be seen that all branches will provide an axes input which is not empty, thus circumventing the check above for all cases except for the specific one for which it is relevant - noop_with_empty_axes is true, and the variable axes are empty.

test/ref/reduce_mean.cpp
music-dino: I see what you mean. That's for the onnx/verify tests to check. I had already added a couple for the case when axes are a variable dynamic input, but I hadn't for any of the others. An oversight on my part, I'll add a couple more onnx/verify tests.

src/onnx/parse_reduce_op.cpp
lakhinderwalia: Perhaps I should have been clearer.. the compute method you refer to, its outcome is two ways: either the operator has axes as an input or it does not. But the line 77 should have a clause for checking the axes being empty. Right now it applies regardless.. Thanks.

music-dino: As created in line 77 the operator will behave as following:

1. [Check if its axes data member is empty](https://github.com/migraphx-benchmark/AMDMIGraphX/blob/5f7b702f88e13dc97ae8be108c11031046b5e1f9/src/include/migraphx/op/reduce_op.hpp#L228). Seeing that it is, it knows that axes must be in the input.
2. Knowing that axes is an input, [it proceeds to check if this input is empty](https://github.com/migraphx-benchmark/AMDMIGraphX/blob/5f7b702f88e13dc97ae8be108c11031046b5e1f9/src/include/migraphx/op/reduce_op.hpp#L231), and returns the data input if it is. This check has to be in the operator to cover off the case of a variable dynamic axes input.

src/onnx/parse_reduce_op.cpp
```diff
@@ -31,66 +31,110 @@ namespace migraphx {
...
+        if(noop_with_empty_axes != 0)
```
lakhinderwalia, line 77: This clause also needs to check that the axes is indeed empty.

lakhinderwalia (changes requested):

src/onnx/parse_reduce_op.cpp
music-dino: I'm sorry but I still fail to see why it should. Can you demonstrate with a particular situation for which it would be necessary, and how you see the code path for that situation, including the compute method of the operator.

pfultz2: We cant check that the axes is empty because we dont know how many axes there are yet until runtime.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -94,16 +94,75 @@ struct reduce_op : op_name<Derived>
...
+        }
+
+        return {data_shape.type(), dims};
+    }
+
+    // Compute the output shape for cases when the input tensor has a static shape.
+    // Depending on how axes is passed to the operator the output shape can be either dynamic or
+    // static.
+    //
+    // If the axes are passed as an attribute, or a constant input, we can precisely
```
pfultz2, line 143: We cant produce a static shape with a constant input because the input to compute_shape is just the shapes.

src/include/migraphx/op/reduce_op.hpp
```diff
@@ -94,16 +94,75 @@ struct reduce_op : op_name<Derived>
...
+    // determine which axes need to be collapsed even during parsing.
```
pfultz2, line 144: Remove `even during parsing`. There is no parsing here. The onnx parser has already constructed the operator.

pfultz2 (approved):

src/onnx/parse_reduce_op.cpp
lakhinderwalia: (Had a discussion with Paul. The reason this logic works is because, our reduce_op has some default behavior when the axis is empty: to be an identity operator. Hence it is taken care of somewhere else. This code doesn't explain that.)
So please add a comment here to give a hint:
'If axes is empty, the default behavior of reduce_op is to be an identity operator"

lakhinderwalia (approved):
Please update with comments, as I have explained above.. Approving this PR.

src/onnx/parse_reduce_op.cpp
music-dino: I added that default behavior, and that's what I was trying to explain. I've added the suggested comment. Thanks for reviewing.

Issue #2116, opened 2023-08-23
Investigate GPT model compilation with migraphx develop branch


umangyadav:
Model : GPT2-medium fp32 model from UIF

When trying to parse GPT2 medium model with develop branch of MIGraphX, i see following errror:

```
terminate called after throwing an instance of 'migraphx::version_1::exception'
what():  /home/umayadav/repo/AMDMIGraphX/src/include/migraphx/operation.hpp:144: compute_shape_op: At least one input is required for slice

```

with debug builds it is failing on assertion

```
Reading: pt_gpt2_medium_fp32.onnx
driver: /home/umayadav/repo/AMDMIGraphX/src/simplify_algebra.cpp:1452: auto migraphx::find_split_transpose::apply(module &, const match::matcher_result &)::(anonymous class)::operator()(auto) const [i:auto = std::_List_iterator<migraphx::instruction>]: Assertion `i->outputs().size() == 1' failed.
Aborted (core dumped)

```

To reproduce run : 

```
migraphx-driver compile pt_gpt2_medium_fp32.onnx --input-dim @input_ids 64 1024
```
It works though with released version of rocm-5.6 migraphx.

umangyadav:
Fixed with #2126

PR #2126, opened 2023-08-25
Skip "find_split_transpose" if split has more than one outputs

umangyadav:
The current matcher is doing an assert on the number of outputs but instead, it should skip it if the number of outputs is greater than 1. 

This matcher is doing 
```
input --> slice --> transpose 
    | 
     ---->slice --> transpose 
    | 
    ---->slice --> transpose 
```
to 
```
input --> transpose --> slice
                | 
                 ----> slice 
                | 
                ----> slice 

```
This is only useful if slice in original input graph has one output node (transpose in this case). If it has multiple outputs then doing transpose before the slice would change the layout for the slice's outputs (other than transpose) in original graph. 

Partially solves #2116

### DIFF
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -1446,10 +1446,13 @@ struct find_split_transpose
         {
             return;
         }
+        if(std::any_of(split_outputs.begin(), split_outputs.end(), [](auto i) {
+               return i->outputs().size() != 1;
+           }))
+            return;
 
         std::vector<instruction_ref> vec_trans(split_outputs.size());
         std::transform(split_outputs.begin(), split_outputs.end(), vec_trans.begin(), [](auto i) {
-            assert(i->outputs().size() == 1);
             return i->outputs().front();
         });
 
--- a/test/simplify_algebra_test.cpp
+++ b/test/simplify_algebra_test.cpp
@@ -603,8 +603,8 @@ TEST_CASE(simplify_inner_broadcast_scalar)
 
     migraphx::module m2;
     {
-        auto x    = m2.add_parameter("x", {migraphx::shape::int32_type, {1, 384}});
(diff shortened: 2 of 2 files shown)


src/simplify_algebra.cpp
```diff
@@ -1446,10 +1446,13 @@ struct find_split_transpose
         {
             return;
         }
+        if(std::any_of(split_outputs.begin(), split_outputs.end(), [](auto i) {
+               return i->outputs().size() != 1;
+           }))
+            return;
```
pfultz2, line 1452: I think can add `match::used_once()` to the matcher instead.

src/simplify_algebra.cpp
```diff
@@ -1446,10 +1446,13 @@ struct find_split_transpose
...
 
         std::vector<instruction_ref> vec_trans(split_outputs.size());
         std::transform(split_outputs.begin(), split_outputs.end(), vec_trans.begin(), [](auto i) {
-            assert(i->outputs().size() == 1);
```
pfultz2, line 1452: We can keep the assert to ensure the matcher is working correctly.

src/simplify_algebra.cpp
umangyadav: I'll have to do it for all the slices, I dont' think i can do it inside matcher.

kahmed10:
I think we need to add a unit test. I've seen this happen in distilgpt2 as well.

src/simplify_algebra.cpp
pfultz2: I think you could use `match::any_of[match::outputs()]` but that might be more complicated.

pfultz2:
> I think we need to add a unit test. I've seen this happen in distilgpt2 as well.

Yea a unit test would be good.

pfultz2 (approved):

kahmed10 (approved):

PR #2130, opened 2023-08-29
Move operation for memory performance + misc changes for cpu performance

lakhinderwalia:
- Reduce memory footprint by std::move of temporary (potentially very large) containers. 
- Minor cleanup for performance optimization: e.g. of Index() calculation -- which can get repeated millions of times in large tensors/vectors in a single Visit.


src/include/migraphx/shape_for_each.hpp
```diff
@@ -51,7 +51,7 @@ void shape_for_each(const migraphx::shape& s, F f)
                            assert(len > 0 and stride > 0);
                            return (i / stride) % len;
                        });
-        call(indices);
+        f(index_const_ref);
```
umangyadav, line 54: This wouldn't result in correct results. Function `f` needs to be called while calculating new indices as it was done previously.

src/shape.cpp
```diff
@@ -50,13 +50,14 @@ struct shape_impl
     {
         assert(t != shape::tuple_type);
     }
+
     shape_impl(shape::type_t t, std::vector<std::size_t> l)
         : m_type(t), m_lens(std::move(l)), m_standard(true)
     {
         assert(t != shape::tuple_type);
         this->calculate_strides();
-        assert(m_lens.size() == m_strides.size());
```
umangyadav, line 58: this assert should stay  unless it is done somewhere else already

src/include/migraphx/shape_for_each.hpp
lakhinderwalia: It is a const reference. That is what the lambda was doing, while adding an indirection of call.. Thanks.

src/onnx/parse_resize.cpp
```diff
@@ -97,22 +97,19 @@ const auto& get_original_idx_op(const std::string& mode)
 static std::vector<int>
 calc_neighbor_points(const std::vector<std::vector<std::vector<std::size_t>>>& vvv_ind,
                      int i_dim,
-                     const std::vector<std::vector<std::size_t>>& vec_dims,
+                     std::vector<std::vector<std::size_t>>& vec_dims,
```
pfultz2, line 100: I would prefer this to be either by value or rvalue reference, and use `std::move` to pass the vector. This will make it clear at the call site that the vector wont be used anymore.

src/onnx/parse_resize.cpp
```diff
@@ -138,8 +135,8 @@ calc_neighbor_points(const std::vector<std::vector<std::vector<std::size_t>>>& v
                            return dim;
                        });
     }
-
-    return calc_neighbor_points(vvv_ind, i_dim + 1, vec_dims1, in_s);
+    vec_dims = std::move(vec_dims1);
+    return calc_neighbor_points(vvv_ind, i_dim + 1, vec_dims, in_s);
```
pfultz2, line 139: I think it would be simpler to do how you showed yesterday:

```cpp
vec_dims.clear();
return calc_neighbor_points(vvv_ind, i_dim + 1, std::move(vec_dims1), in_s);
```

This way we wont need double moves.

src/shape.cpp
```diff
@@ -151,6 +152,24 @@ struct shape_impl
             m_lens.begin(), m_lens.end(), std::size_t{1}, std::multiplies<std::size_t>());
     }
 
+    std::size_t get_index(size_t i) const
```
pfultz2, line 155: Why is this moved to the impl? The impl class doesnt use this.

src/shape.cpp
```diff
@@ -151,6 +152,24 @@ struct shape_impl
...
+    {
+        std::size_t result = 0;
+        std::size_t s      = 1;
+        std::size_t k      = m_lens.size();
+
+        while(k != 0)
```
pfultz2, line 161: Actually would be cleaner to do `for(auto k:reverse(range(m_lens.size()))`

src/shape.cpp
lakhinderwalia: Because the implementation is heavily involved in all calculations: We are accessing impl, a shared pointer, for every single operand in that calculation of index (which gets repeated millions of times for a large tensor..) 

BTW, just an aside question, why is this impl a shared pointer?  Thanks.

src/shape.cpp
lakhinderwalia: In c++ 17.. ?

pfultz2: Yes we have these functions under the `migraphx/ranges.hpp` header.

src/shape.cpp
pfultz2: > BTW, just an aside question, why is this impl a shared pointer? 

Because its cheaper to ref count then allocate and copy.

lakhinderwalia: The shared_pointer implementation of shape.hpp is overall slower by 7.5% than not using it. 

(I compared the driver read operation on bevdet.onnx; In debug mode, as before. The only difference is the workstation is different this time for the apples to apples comparison.)

pfultz2: > The shared_pointer implementation of shape.hpp is overall slower by 7.5% than not using it.

So what did you use instead of a `shared_ptr`?

pfultz2: > In debug mode

Of course, the debug mode is slow. Its faster to pass `-01` or `-02` to the debug build to get better performance.

lakhinderwalia: (All the testing was done in debug mode, so just wanted to mention the comparison numbers. That also likely uses all the asserts to execute (likely turned off in release mode.)

Instead of a shared_ptr, the structure shape_impl was just included in struct shape.. No pointers. I can check that code in after running all the unit-tests.

src/shape.cpp
lakhinderwalia: Looking at the code, (ranges.hpp): reverse(range()) would create a new range objects (twice)..

src/onnx/parse_resize.cpp
lakhinderwalia: Done.

src/shape.cpp
pfultz2: > I can check that code in after running all the unit-tests.

You could put it into a separate branch and we could try to test it, but I would be suprised that it would be faster in release mode.

src/onnx/parse_resize.cpp
```diff
@@ -138,8 +135,7 @@ calc_neighbor_points(const std::vector<std::vector<std::vector<std::size_t>>>& v
                            return dim;
                        });
     }
-
-    return calc_neighbor_points(vvv_ind, i_dim + 1, vec_dims1, in_s);
+    return calc_neighbor_points(vvv_ind, i_dim + 1, std::move(vec_dims1), in_s);
```
pfultz2, line 138: You should do `vec_dims.clear()` before returning.

src/shape.cpp
pfultz2: Of course, its two function calls.

src/onnx/parse_resize.cpp
lakhinderwalia: duh. yes, the original purpose..; thanks. :-)

src/shape.cpp
lakhinderwalia: Done. 
(arguably.. a complex step for a simple original one..)

src/shape.cpp
lakhinderwalia: I will create a separate branch for it.. thanks.

src/include/migraphx/shape_for_each.hpp
```diff
@@ -51,9 +51,10 @@ void shape_for_each(const migraphx::shape& s, F f)
                            assert(len > 0 and stride > 0);
                            return (i / stride) % len;
                        });
-        call(indices);
+        f(i, index_const_ref);
```
pfultz2, line 54: The `i` should be optional since its only used in a few places, and should be the last parameter. You can create an invoker to call the function like this:

```cpp
template<class F, class Vector>
auto shape_for_each_invoke(rank<2>, F& f, const Vector& v, std::size_t i) -> decltype(f(v, i))
{
    return f(v, i);
}

template<class F, class Vector>
auto shape_for_each_invoke(rank<1>, F& f, const Vector& v, std::size_t) -> decltype(f(v))
{
    return f(v);
}
``` 

And then you can do `shape_for_each_invoke(rank<2>{}, f, index_const_ref, i)` to call the function.

src/include/migraphx/op/gather.hpp
```diff
@@ -125,13 +125,12 @@ struct gather
...
-                        in_index       = (in_index < 0) ? in_index + axis_dim_size : in_index;
-                        data_idx[axis] = in_index;
-                        output[out_comp_shape.index(out_idx.begin(), out_idx.end())] =
-                            data(data_idx.begin(), data_idx.end());
+                    shape_for_each(out_comp_shape, [&](size_t out_idx, const auto& out_idx_v) {
+                        auto data_idx   = out_idx_v;
+                        auto in_index   = indices[data_idx[axis]];
+                        in_index        = (in_index < 0) ? in_index + axis_dim_size : in_index;
+                        data_idx[axis]  = in_index;
+                        output[out_idx] = data(data_idx.begin(), data_idx.end());
```
pfultz2, line 133: It will be faster to do `output(out_idx_v)`, as that will avoid divisions in the index calculations.

pfultz2: Ah wait we dont have an overload for vector directly. So you need to do `output(out_idx_v.begin(), out_idx_v.end())`.

lakhinderwalia: `out_idx` is the (equivalent of) calculation result of `out_comp_shape.index(out_idx.begin(), out_idx.end())` -- that was replaced. I am not clear why do we need to anything else..

src/include/migraphx/shape_for_each.hpp
lakhinderwalia: I first implemented it by making it the second parameter, just because it is optional in a few instances. But then felt it was logical to make it the first one: <an index, followed by its vector representation>. Hence this way.. 

(Personally, I am not a fan of putting a rank<int>, just to invoke a simple function).

lakhinderwalia: BTW, if a template were to be used, we could do it without a rank..;

lakhinderwalia: BTW, `get_original_idx_op()` in parse_resize.cpp does exactly this: its first few parameters in the lambdas are just ignored.. same style here.. Thanks.

pfultz2: > Personally, I am not a fan of putting a rank, just to invoke a simple function

Rank is to avoid ambiguous overloads. Most likely it wont be needed, but if the we were to pass a variadic lambda in the future it would fail. Perhaps the error is acceptable as a variadic lambda does seem a little ambiguous in this context.

So I think the `rank` can be removed from the functions.

src/include/migraphx/op/gather.hpp
pfultz2: Using `out_idx_v` will only do multiplications and additions to get the offset:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/shape.cpp#L349

Using `out_idx` is a more complicated calculation that requires divisions and mods to calculate the offset, so is slower:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/shape.cpp#L370

src/include/migraphx/shape_for_each.hpp
lakhinderwalia: Thanks, Paul. We have a very simple case here where it would be fairly routine stuff for a decent compiler to warn if something was amiss -- e.g. unused parameter. I think we can worry about the future lambda cases later. 

My recommendation is to leave this change as it is, until somebody changes something in future about the variadic lambda(s).

src/include/migraphx/op/gather.hpp
lakhinderwalia: (Thanks for the explanation, and I now understand your concern..)  

However, the default of `this->standard()` case  : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/shape.cpp#L359
is just a pass-through (and would be of course the fastest with `out_idx`)

I can revert this, if you like.. ; Thanks.

src/include/migraphx/shape_for_each.hpp
pfultz2: We should still make the parameter optional(even though we dont use `rank` here), just so this doesn't need to be changed everywhere `shape_for_each` is used and it wont break other developers who may be using `shape_for_each` in their unmerged work or PRs.

src/include/migraphx/op/gather.hpp
pfultz2: Faster with standard and slower for non-standard. I guess the standard shape is more common for ref operators so it might be better to use `out_idx`.

src/include/migraphx/shape_for_each.hpp
pfultz2: Without `rank`:

```cpp
template<class F, class Vector>
auto shape_for_each_invoke(F& f, const Vector& v, std::size_t i) -> decltype(f(v, i))
{
    return f(v, i);
}
template<class F, class Vector>
auto shape_for_each_invoke(F& f, const Vector& v, std::size_t) -> decltype(f(v))
{
    return f(v);
}
```

lakhinderwalia: Would this be okay? I compiled this -- in shape_for_each.hpp: 
```
template <class F, class Vec = std::vector<std::size_t>, class Idx = size_t>
void shape_for_each(const migraphx::shape& s, F f)
{
    std::vector<std::size_t> indices(s.lens().size());
    auto const& index_const_ref = indices;
    shape ss{s.type(), s.lens()};
    size_t max = ss.elements();
    for(std::size_t i = 0; i < max; i++)
    {
        std::transform(ss.strides().begin(),
                       ss.strides().end(),
                       ss.lens().begin(),
                       indices.begin(),
                       [&](std::size_t stride, std::size_t len) {
                           assert(len > 0 and stride > 0);
                           return (i / stride) % len;
                       });
        if constexpr(std::is_invocable_v<F, Vec, Idx>)
            f(index_const_ref, i);
        else
            f(index_const_ref);
    }
}
```

lakhinderwalia: (Above) Code is checked in. But why have some build process failed -- particularly the 20.04 debug build -- it isn't clear to me.

src/include/migraphx/shape_for_each.hpp
```diff
@@ -34,14 +34,14 @@ inline namespace MIGRAPHX_INLINE_NS {
...
  * Iterates the given function over the indices from the shape in order.
  */
-template <class F>
+template <class F, class Vec = std::vector<std::size_t>, class Idx = size_t>
 void shape_for_each(const migraphx::shape& s, F f)
 {
-    // Ensure calls to f use const ref to vector
-    auto call = [&f](const std::vector<std::size_t>& i) { f(i); };
     std::vector<std::size_t> indices(s.lens().size());
+    auto const& index_const_ref = indices;
```
umangyadav, line 41: wouldn't it be same if it was `const auto& index_const_ref` ?

lakhinderwalia: Sure. Thanks. Will change it..

src/onnx/parse_resize.cpp
```diff
@@ -126,8 +123,8 @@ calc_neighbor_points(const std::vector<std::vector<std::vector<std::size_t>>>& v
                        });
     }
 
-    const auto& vv_hi = vv_ind.at(1);
-    for(std::size_t start = 0; start < vec_dims.size(); start += vv_lo.size())
+    const auto& vv_hi = vvv_ind[i_dim][1];
+    for(std::size_t start = 0; start < vec_dims.size(); start += vv_hi.size())
```
umangyadav, line 127:
```suggestion
    for(std::size_t start = 0; start < vec_dims.size(); start += vv_lo.size())
```
Should it be `vv_lo` ?

lakhinderwalia: They are both of same size: @ line 330, 331.

src/include/migraphx/shape_for_each.hpp
lakhinderwalia: Done

umangyadav (approved):

src/onnx/parse_resize.cpp
lakhinderwalia: (Fixed already. Thanks.)

src/include/migraphx/shape_for_each.hpp
pfultz2: Yes that would work, but I would prefer it to be written as `std::is_invocable<F, decltype(index_const_ref), decltype(i)>{}`.

pfultz2: > But why have some build process failed -- particularly the 20.04 debug build -- it isn't clear to me.

Its a network issue, I dont think its an issue with your code.

src/include/migraphx/shape_for_each.hpp
```diff
@@ -51,9 +51,13 @@ void shape_for_each(const migraphx::shape& s, F f)
                            assert(len > 0 and stride > 0);
                            return (i / stride) % len;
                        });
-        call(indices);
+        if constexpr(std::is_invocable_v<F, Vec, Idx>)
```
pfultz2, line 54: Do `std::is_invocable<F, decltype(index_const_ref), decltype(i)>{}` instead. Use `is_invocable` instead of `is_invocable_v` and use `decltype` instead of adding template parameters.

lakhinderwalia: Good idea. Thanks. Done.

pfultz2 (approved):

PR #2141, opened 2023-08-31
Update Constant parsing to support more attributes

music-dino:
- Add parsing support for `value_float`, `value_floats`, `value_int`, `value_ints` attributes
- Disable failing tests
- Resolves migraphx-benchmark/AMDMIGraphX#76

For ONNX version 1.13 and above, certain tests failed due to lack of support for the aforementioned attributes.
After adding support, the same tests fail on account of CastLike being unsupported, so those tests are disabled until it is implemented.


src/onnx/parse_constant.cpp
```diff
@@ -39,16 +39,29 @@ struct parse_constant : op_parser<parse_constant>
                           onnx_parser::node_info info,
                           const std::vector<instruction_ref>& /*args*/) const
     {
-        literal v = parser.parse_value(info.attributes.at("value"));
+        static const std::string attributes[] = {
+            "value", "value_float", "value_floats", "value_int", "value_ints"};
```
CharlieL7, line 43: style: use `std::vector<std::string>` instead

src/onnx/parse_constant.cpp
```diff
@@ -39,16 +39,29 @@ struct parse_constant : op_parser<parse_constant>
...
+
+        auto&& attr = [&] {
+            for(const auto& a : attributes)
+            {
+                if(auto it = info.attributes.find(a); it != info.attributes.end())
+                {
+                    return it->second;
+                }
```
CharlieL7, line 51: style: use `contains(C, a)` instead from `ranges.hpp`

src/onnx/parse_constant.cpp
```diff
@@ -39,16 +39,29 @@ struct parse_constant : op_parser<parse_constant>
...
+            }
+            MIGRAPHX_THROW("Constant does not contain any supported attribute");
+        }();
```
CharlieL7, line 54: Based on the ONNX spec it looks like having more than one value attribute set is also invalid. Probably want to also check that here.

src/onnx/parse_constant.cpp
```diff
@@ -39,16 +39,29 @@ struct parse_constant : op_parser<parse_constant>
...
         // return empty literal
         if(v.get_shape().elements() == 0)
         {
             return info.add_literal(literal{v.get_shape().type()});
         }
 
-        auto dim_size = info.attributes.at("value").t().dims_size();
         // if dim_size is 0, it is a scalar
-        if(dim_size == 0)
+        if(attr.has_t() && attr.t().dims_size() == 0)
```
CharlieL7, line 64:
```suggestion
        if(attr.has_t() and attr.t().dims_size() == 0)
```
style change

src/onnx/parse_constant.cpp
CharlieL7, line 54: style: why are we creating a lambda with capture by reference and then immediately calling it? The lambda doesn't look like it's used again. It looks like you could declare `attr` before and then set it.

CharlieL7, line 54: Should have `onnx_test.cpp` entries that trigger throws for code coverage

CharlieL7 (changes requested):

src/onnx/parse_constant.cpp
music-dino: Fixed

src/onnx/parse_constant.cpp
music-dino: Fixed

src/onnx/parse_constant.cpp
music-dino: Noted, the code uses it now.

src/onnx/parse_constant.cpp
music-dino: Added checks for that case.

music-dino: This lambda has now been removed due to a rewrite.

music-dino: Tests cases for situations when the code is expected to throw have been added.

src/onnx/parse_constant.cpp
```diff
@@ -39,16 +39,41 @@ struct parse_constant : op_parser<parse_constant>
...
+            std::string err = "Constant contains multiple attributes: " + *it;
+            for(auto i = std::next(it); i != attributes.end(); ++i)
+            {
+                if(p(*i))
+                {
+                    err += ", " + *i;
+                }
+            }
+            MIGRAPHX_THROW(err);
+        }
```
pfultz2, line 64: Use `copy_if` to copy all the attribute names to a vector and then if the vector > 1 throw an error about multiple attributes. You can use `join_strings` to convert the vector of strings to a comma-delimited string.

CharlieL7:
It doesn't look like the code changes were pushed up yet?

music-dino:
> It doesn't look like the code changes were pushed up yet?

I fail to see what you mean. The requested changes are present in the diff.

CharlieL7:
> I fail to see what you mean. The requested changes are present in the diff.

Weird, I was unable to see the new changes earlier today. It looks like it was updated now

CharlieL7 (approved):
cppcheck error looks like a false positive from L60 of parse_constant.cpp.

src/onnx/parse_constant.cpp
```diff
@@ -39,16 +40,37 @@ struct parse_constant : op_parser<parse_constant>
...
+            MIGRAPHX_THROW("Constant node does not contain any supported attribute");
+        }
+
+        if(present_attributes.size() > 1)
+        {
+            MIGRAPHX_THROW("Constant contains multiple attributes: " +
+                           join_strings(std::move(present_attributes), ", "));
+        }
+
+        auto&& attr = info.attributes[present_attributes[0]];
```
umangyadav, line 63: you can use `cppcheck-suppress` to avoid false positives.

umangyadav (approved):
Looks good to me

pfultz2 (approved):

causten:
@music-dino  so close, can you fix the cppcheck

music-dino:
> @music-dino so close, can you fix the cppcheck

Should be fixed now.

PR #2142, opened 2023-08-31
[Re-Opened] Added support for standalone mlir-conv (used to be #2110)

ravil-mobile:
The old discussion used to be in PR #2110 which was created from a forked project. This was blocking some pipelines to start. To avoid this, I just created a PR from my local branch of migraphx repo.


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -21,6 +21,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+#include "migraphx/shape.hpp"
```
pfultz2, line 24: Use angle brackets, but I dont think this header is needed as its already included by the other headers.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -301,14 +301,87 @@ struct find_mlir_op
...
+        mm->set_bypass();
+        auto [anchor_op, top_inputs] = fuse_input_ops_and_gemm_based_op(mm, conv_based_op);
+        mm->add_return({anchor_op});
+        mpm.get_module().replace_instruction(
+            conv_based_op, mlir_op{conv_based_op->get_operator()}, top_inputs, {mm});
+    }
+};
+
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_ENABLE_OPS);
+bool is_self_decide() { return env(MIGRAPHX_MLIR_ENABLE_OPS::value()).empty(); }
```
pfultz2, line 330: This should use `string_value_of`, otherwise this will read the env everytime which is slow.

Also, what does `is_self_decided` mean?

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -301,14 +301,87 @@ struct find_mlir_op
...
+
+bool is_requested(std::string_view option)
+{
+    assert(enabled(MIGRAPHX_MLIR_ENABLE_OPS{}));
+    auto string_value = string_value_of(MIGRAPHX_MLIR_ENABLE_OPS{}, "");
+    static const char delim{','};
+    string_value.push_back(delim);
```
pfultz2, line 337: Why is the delimiter added to the end?

src/targets/gpu/fuse_mlir.cpp
ravil-mobile: Hi,

`MIGRAPHX_MLIR_ENABLE_OPS` means that the user specifies concrete operations which rocMLIR has to handle. For example, MIGRAPHX_MLIR_ENABLE_OPS="conv" will only generate convolutions with rocMLIR; fusing will be switched off.

if `MIGRAPHX_MLIR_ENABLE_OPS` is not specified by a user then we decide **ourselves** (named it as **self_decide**) which operations for rocMLIR to enable - i.e., always use fused operations and, if the underlying GPU is from Navi3x, use standalone convolutions

If you want I can rename `MIGRAPHX_MLIR_ENABLE_OPS` to `MIGRAPHX_MLIR_USE_SPECIFIC_OPS`

src/targets/gpu/fuse_mlir.cpp
ravil-mobile: Good question. It is needed to handle a corner case - i.e., when a list of strings separated by `,` has only one value and, thus, there is no `,` at all. 

Maybe it is related to

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/ac96370b92b940fc2102b8b2aacf06ae1576a80a/src/include/migraphx/stringutils.hpp#L89

which just adds a white space to the end. I think it is supposed be `delim`. Probably, in all tests inly whitespaces are used as delimiters.

src/targets/gpu/fuse_mlir.cpp
ravil-mobile: I just removed it. Probably, my vscode included it automatically

src/targets/gpu/fuse_mlir.cpp
ravil-mobile: Updated according the recent changes in `stringutils`. Done

src/targets/gpu/fuse_mlir.cpp
pfultz2: Sure, either name is find. Can you add a comment to source code to explain how this is supposed to work and what the valid values are? I would expected to use `MIGRAPHX_MLIR_ENABLE_OPS=convolution` not `MIGRAPHX_MLIR_ENABLE_OPS=conv` since thats the way we spell the operator, but if its spelled differently for this variable, we should have some comments to document it.

manupak: Lets not introduce a another indirection here..
and stick to migraphx op names ?

pfultz2 (approved):

PR #2143, opened 2023-08-31
Fixed writing the input files for rocMLIR's `tuningRunner.py`

ravil-mobile:
This PR fixes the input tuning config file format for rocMLIR

```bash
MIGRAPHX_ENABLE_MLIR=1 MIGRAPHX_MLIR_TUNING_CFG=unet.cfg ./bin/migraphx-driver compile --onnx /MIGraphXDeps/resnet50-v1-7.onnx
```

Before the PR

```bash
> cat ./unet.cfg.gemm
120
120
120
120
...
```

After (Fixed by this PR)

```bash
> cat ./unet.cfg.gemm
-t f32 -out_datatype f32 -transA false -transB false -g 1 -m 1 -n 1000 -k 2048
```

Closes ROCmSoftwarePlatform/rocMLIR-internal#1103


src/targets/gpu/mlir.cpp
```diff
@@ -22,6 +22,7 @@
  * THE SOFTWARE.
  */
 #include "migraphx/make_op.hpp"
+#include "migraphx/stringutils.hpp"
```
pfultz2, line 25: This should use angle brackets.

krzysz00 (approved):
Approved (have been walked through the problem earlier today)

src/targets/gpu/mlir.cpp
ravil-mobile: done

PR #2145, opened 2023-08-31
Fix ordering inside IF op

umangyadav:
`std::set` stores input parameters using some sorting function and doesn't guarantee order. 
Review with whitespace change disabled. 

Let's say 

ins = if(cond, x, y)[then_mod, else_mod] 

then_mod has param named name "then_mod_param"
else_mod has param named "else_mod_param"
then,  `std::set` was ordering "else_mod_param" first then "then_mod_param" . That was causing `x` to bind with `else_mod_param` and `y` to bind with `then_mod_param`

### DIFF
--- a/src/include/migraphx/op/if_op.hpp
+++ b/src/include/migraphx/op/if_op.hpp
@@ -70,11 +70,11 @@ struct if_op
         module_ref mod = cond ? mods[0] : mods[1];
         std::unordered_map<std::string, argument> params;
 
-        std::set<std::string> pnames;
+        std::vector<std::string> pnames;
         for(const_module_ref smod : mods)
         {
             auto names = smod->get_parameter_names();
-            pnames.insert(names.begin(), names.end());
+            pnames.insert(pnames.end(), names.begin(), names.end());
         }
 
         assert(pnames.size() < args.size());
--- a/test/ref_ops_test.cpp
+++ b/test/ref_ops_test.cpp
@@ -1403,17 +1403,17 @@ TEST_CASE(conv_dyn_weights_shape_test)
                             0.81217463, 0.10789447, 0.43083784, 0.63371852, 0.69742316, 0.09536905};
 
     std::vector<float> c   = {0.98411968,
-                            0.2899219,
-                            0.44638833,
-                            0.30390816,
-                            0.03989896,
-                            0.2445332,
-                            0.32700131,
-                            0.57517075,
-                            0.06956476,
-                            0.93079306,
-                            0.19882314,
-                            0.52940601};
+                              0.2899219,
+                              0.44638833,
+                              0.30390816,
+                              0.03989896,
+                              0.2445332,
+                              0.32700131,
+                              0.57517075,
+                              0.06956476,
+                              0.93079306,
+                              0.19882314,
+                              0.52940601};
     std::vector<float> sol = {1.9939406,
                               2.2703054,
                               1.8896171,
@@ -1437,9 +1437,9 @@ TEST_CASE(conv_dyn_weights_shape_test)
     EXPECT(migraphx::verify::verify_range(results_vector, sol));
 
     c   = {0.98411968, 0.2899219,  0.44638833, 0.30390816, 0.03989896, 0.2445332,  0.32700131,
-         0.57517075, 0.06956476, 0.93079306, 0.19882314, 0.52940601, 0.35624753, 0.35938406,
-         0.9111428,  0.88923574, 0.61040283, 0.2797513,  0.15479768, 0.46534674, 0.16970931,
-         0.49704618, 0.07062198, 0.01678321, 0.53150934, 0.39244495, 0.9963813};
+           0.57517075, 0.06956476, 0.93079306, 0.19882314, 0.52940601, 0.35624753, 0.35938406,
+           0.9111428,  0.88923574, 0.61040283, 0.2797513,  0.15479768, 0.46534674, 0.16970931,
+           0.49704618, 0.07062198, 0.01678321, 0.53150934, 0.39244495, 0.9963813};
     sol = {6.1329393, 4.3199925, 5.448438, 3.8497565};
     migraphx::shape weights_fixed_shape1{migraphx::shape::float_type, {1, 3, 3, 3}};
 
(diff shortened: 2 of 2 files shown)


pfultz2:
The arguments are supposed to be passed in alphabetic order, thats why we used a set.

umangyadav:
> The arguments are supposed to be passed in alphabetic order, thats why we used a set.

it can't be guaranteed always.

pfultz2:
> it can't be guaranteed always.

Why?

umangyadav:
> > it can't be guaranteed always.
> 
> Why?

I am writing an automatic module generator for multi-target compilation and inside it, I can't guarantee it.  

Also just in general, i don't think it is good idea to rely on alphabetic order

src/include/migraphx/op/if_op.hpp
```diff
@@ -70,11 +70,11 @@ struct if_op
         module_ref mod = cond ? mods[0] : mods[1];
         std::unordered_map<std::string, argument> params;
 
-        std::set<std::string> pnames;
+        std::vector<std::string> pnames;
         for(const_module_ref smod : mods)
         {
             auto names = smod->get_parameter_names();
-            pnames.insert(names.begin(), names.end());
+            pnames.insert(pnames.end(), names.begin(), names.end());
```
pfultz2, line 77: This will duplicate the arguments when both submodules have the same parameters. 

In the most likely case, the then module will have parameter 'x', and else module will also have parameter 'x'. So its expected that the user can just pass `if(cond, x)` not `if(cond, x, x)`.

umangyadav: So `pnames` "vector" should be de-duplicated

pfultz2:
> Also just in general, i don't think it is good idea to rely on alphabetic order

Why?

Its better than relying on the order in the module. Especially for an operator like `if`, which has two submodules. The then module could have the parameters in one order and the else module have them in a different order. So which order should they be passed in this case? (For sure we could define an order, but it now becomes more complicated to figure out the order than a simple sort).

Also, the other modules like pointwise, select_module, mlir_op, and fused_reduce use the same alphabetic order as well.

umangyadav:
> `std::set` is already ordered lexicographically. I need to see some of the code mentioned for "automatic module generator for multi-target compilation" to get a better idea of what's the issue.

Multi-target case, i am automatically generating parameter names for sub-modules which are input to the `If`.  I am following some sort of naming convention based on compilation target id of the submodule. For example if the sub-mod is named `target_mod_1_0` that would mean 0th sub module (sub-graph) for target id 1.  Parameters inside those sub modules would have another counter. 

If `If` op binds parameters lexicographically then it can easily screw up the order of arguments to the selected sub-module to execute.

CharlieL7:
> For example if the sub-mod is named `target_mod_1_0` that would mean 0th sub module (sub-graph) for target id 1. Parameters inside those sub modules would have another counter.

There's a naive way to get the lexicographical sorting to work as expected for numbers by using a fixed number of digits, such as `target_mod_001_000`. Not a perfect solution because of possible overflow.

On the other hand, doing it by using the order of the arguments can quickly become quite arbitrary. That actually led to problems for linking `output_parameters`  when I was making `select_module`. If doing the sorting the same way is not possible with the multi-target generator we may want to refactor the `op.compute()` to not rely on the order of the arguments for these types of operators. That would be a pretty major refactor however

umangyadav:
> using a fixed number of digits, such as target_mod_001_000. Not a perfect solution because of possible overflow. 

I am not using a fixed number of digits. I am using `size_t` and converting it to a string. It is almost impossible to overflow.

CharlieL7:
> I am not using a fixed number of digits. I am using `size_t` and converting it to a string. It is almost impossible to overflow.

I meant something like this: `snprintf (str, 34, "%06d", n)`.

umangyadav:
> I'm also still not understanding why the ordering, even if it looks weird, would not work? 

Let's say I have an `if` op with then module named as `target_1_0` and else module named as `target_0_0`. 

Then and else mod do not share the same runtime arguments or same parameter names. 

Let's say `target_mod_1_0` has parameter names like `target_mod_1_0:param_0` and in similar fashion for other sub modules as well. 

For example : 

`target_1_0` module takes x and y as runtime arguments 

and the `target_0_0` module takes `a` and `b` as the runtime arguments. 

I would be constructing `if` op as 
`make_op("if", {cond, x, y, a, b}, {target_mod_1_0, target_mod_0_0}` . 

Let's say the condition is true then current implementation will pass `a` and `b` to the target_mod_1_0 instead of `x` and `y`.

pfultz2:
> Then and else mod do not share the same runtime arguments or same parameter names.

An unlikely use case, I think.

> Let's say the condition is true then current implementation will pass a and b to the target_mod_1_0 instead of x and y.

Yes, `a` and `b` should come first in this scenario due to the name of the parameters. I dont understand why you would expect it to come last.

umangyadav:
> An unlikely use case, I think.

I have a case here where parameter names would be different for each module created (which can possibly be input to if operator and the current sorting mechanism wouldn't work in that case.).
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/f079b678269114e7f1a3e7aac16e36c199492a33/src/generate_root_modules.cpp#L258

umangyadav:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/f079b678269114e7f1a3e7aac16e36c199492a33/test/multi_target/multitarget_test.cpp#L308

I have an example here.

pfultz2:
Why arent you passing the parameter names as `if_mod_param_0`? That way they use the same parameter name for both the if and else.

pfultz2:
Also, for the module generation you can pass a prefix to the `create_test_module` function and then use that prefix instead of trying to use the module name. This will make sure the parameters either share the same name or are lexicological ordered the way you would expect.

umangyadav:
> Also, for the module generation you can pass a prefix to the `create_test_module` function and then use that prefix instead of trying to use the module name. This will make sure the parameters either share the same name or are lexicological ordered the way you would expect.
> 
> ```c++
>     auto then_mod_if =
>         then_mod->add_instruction(migraphx::make_op("if"),
>                                   {then_mod_cond,
>                                    then_mod_param_0,
>                                    then_mod_param_1,
>                                    then_mod_param_2,
>                                    then_mod_ref_ins,
>                                    then_mod_param_1,
>                                    then_mod_param_2},
>                                   {create_test_module(p, 1, "1_"), create_test_module(p, 0, "2_")});
> ```

Yes naming them in that order would solve it. But I find it hacky. We can't expect ONNX models to follow such order.

e.g. https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/20f6ed921be6ce7ecbbdbbbe3e61a8a2782cc9d2/src/onnx/parse_if.cpp#L82

pfultz2:
> We can't expect ONNX models to follow such order.

Onnx spec doesnt allow for additional arguments so its not a problem: https://onnx.ai/onnx/operators/onnx__If.html

umangyadav:
> Onnx spec doesnt allow for additional arguments so its not a problem: https://onnx.ai/onnx/operators/onnx__If.html

I don't find it. It says number of outputs should match. it is not saying anything about number of inputs.

pfultz2:
> > Onnx spec doesnt allow for additional arguments so its not a problem: https://onnx.ai/onnx/operators/onnx__If.html
> 
> I don't find it. It says number of outputs should match. it is not saying anything about number of inputs.

From the doc:

> #### Inputs
> * cond (heterogeneous) - B: Condition for the if

Only `cond` is a valid input.

umangyadav:
> Only `cond` is a valid input.

Okay. Looks like `then` and `else` modules inside onnx directly refers to the outer module's parameters instead of redefining them. 

Inside MIGraphX are we following the convention that the first module in the list module args should the lowest parameter names be in sorted order? (if not we should follow some convention and do some assertions to avoid accuracy problems). 

The module can share parameter names, so it is like following in sorted order. 

first mod args params <= secon mod args param <= .. <= last mod args params?

umangyadav:
Closing this for now.

PR #2146, opened 2023-08-31
Rework split_reshape simplification to handle multiple reshapes from same slice

shivadbhavsar:
This resolves an edge case found in multiple huggingface models where the following happens:
```
slice1 -> cont -> reshape1
       |
        -> cont -> reshape2

slice2 -> cont -> reshape1
       |
        -> cont -> reshape2

slice3 -> cont -> reshape1
       |
        -> cont -> reshape2

```
In some cases the `find_split_reshape` matcher will match with reshape2, but `vec_rsp` will consist of reshape1 dims causing a shape mismatch error. Solution is to include `rsp` when checking all reshapes are the same.

Added unit test reproduces this issue in the current build.


pfultz2:
It seems like this should be written as:

```
reshape1 -> slice1, slice2, slice3
reshape2 -> slice4,slice5,slice6
```

Rather than just stopping.

shivadbhavsar:
This turned into a larger update than intended but here is a summary of changes:
1. The initial bug was in part due to a set of (slice -> cont -> reshape)s where the reshapes did not use the entire input shape. Eg. Input: [4, 6, 30], Slices: [4, 2, 30], [4, 2, 30], Reshapes: [4, 60], [4, 60]. Here, our current logic would try to create a new reshape (from input [4, 6, 60]) of size [4, 120], which is invalid (should be [4, 180]). Allowing this kind of simplification required modifying how the new slice indices are computed.
2. It no longer assumes that the slices have only a single (contiguous -> reshape) attached. We do not modify the existing slice -> contiguous so any nodes using these as inputs should still be valid. If no nodes use these as inputs, dead code elim will take care of it.

src/simplify_algebra.cpp
```diff
@@ -1367,6 +1369,10 @@ struct find_split_reshape
         auto slc_lens     = slc->get_shape().lens();
         auto slc_dim_size = std::accumulate(
             slc_lens.begin() + axis, slc_lens.end(), 1, std::multiplies<std::size_t>());
+        auto input_lens = input->get_shape().lens();
+        auto input_size = std::accumulate(
```
pfultz2, line 1373: Cant you just write `auto input_size = input->get_shape().elements()`?

umangyadav:
Unet accuracy check has failed on compilation. Can you check ?

shivadbhavsar:
> Unet accuracy check has failed on compilation. Can you check ?

This should be fixed in the new run

src/simplify_algebra.cpp
```diff
@@ -1404,6 +1404,11 @@ struct find_split_reshape
         rsp_out_lens[rsp_axis] = 1;
         auto rsp_fixed_size    = std::accumulate(
             rsp_out_lens.begin(), rsp_out_lens.end(), 1, std::multiplies<std::size_t>());
+        // cannot create a valid reshape for simplification
+        if(input_size % rsp_fixed_size != 0)
+        {
+            return;
```
pfultz2, line 1410: Is it possible to add a unit test for this case?

shivadbhavsar: added

TedThemistokleous (approved):

causten:
> But I figured it was coded this way

Is there enough comments in the code to help understand why something was done a certain way?  If you need to make a code change for any other reason feel free to improve the code comments.

src/simplify_algebra.cpp
```diff
@@ -1367,6 +1369,9 @@ struct find_split_reshape
         auto slc_lens     = slc->get_shape().lens();
         auto slc_dim_size = std::accumulate(
             slc_lens.begin() + axis, slc_lens.end(), 1, std::multiplies<std::size_t>());
+        auto input_lens   = input->get_shape().lens();
+        auto input_size   = input->get_shape().elements();
+        auto slc_axis_len = input_lens[axis];
```
umangyadav, line 1374: I think matcher should also check if there is only axis that is being sliced. If multiple axes are sliced then it should return (perhaps we can do better).

src/simplify_algebra.cpp
```diff
@@ -1393,16 +1398,39 @@ struct find_split_reshape
...
+        // calculate new slice start and end indices
+        // indices are scaled using the new reshape axis and the original slice axis
+        // eg. slice with size [1, 2, 4, 30] can be reshaped to [8, 30] and vice-versa
+
+        std::vector<int64_t> new_starts(vec_rsp.size());
+        std::transform(vec_rsp.begin(), vec_rsp.end(), new_starts.begin(), [&](auto is) {
+            auto cont   = is->inputs().front();
+            auto og_slc = cont->inputs().front();
+            return any_cast<op::slice>(og_slc->get_operator()).starts[0] * rsp_axis_len /
+                   slc_axis_len;
```
umangyadav, line 1424: Is the `start` guaranteed to result in integer?

shivadbhavsar: The only other possibility as far as I understand is you can technically have an empty slice op (axes={}, starts={}, ends={}) which would be problematic. But that should never happen because that sort of instruction gets eliminated as dead code. And if it somehow makes it to this matcher, `get_splits` will be empty and we would return right away. 
Other than that, starts is defined as std::vector<int64_t>

src/simplify_algebra.cpp
shivadbhavsar: added

shivadbhavsar:
For more clarity heres an example walking through the code (this simplification wouldnt work with the original implementation):
```
Original Instuctions:
Input: {128, 96}
Slices: 
s0 slice(axes={0}, starts={0}, ends={8})(inp) -> {8, 96}
s1 slice(axes={0}, starts={8}, ends={16})(inp) -> {8, 96}
s2 slice(axes={0}, starts={16}, ends={24})(inp) -> {8, 96}
s3 slice(axes={0}, starts={24}, ends={128})(inp) -> {104, 96}
Reshapes:
reshape(dims={2, 4, 96})(s0)
reshape(dims={2, 4, 96})(s1)
reshape(dims={2, 4, 96})(s2)

Intermediate Variables:
axis = 0
slc_lens = {8, 96}
slc_axis_len = input[axis] = 128
slc_dim_size = 8*96 = 768

rsp_lens = {2, 4, 96}, strides: {96*4*2, 96*4, 96, 1} = {768, 384, 96, 1} (with added stride)
rsp_axis = <index where (rsp_stride == slc_dim_size)> = 0

rsp_out_lens (initial) = {1, 4, 96}
rsp_fixed_size = 1*4*96 = 384
rsp_axis_len = 128 * 96 / 384 = 32 
rsp_out_lens (final) = {32, 4, 96}

new_starts = {0*32/128,  8*32/128, 16*32/128} = {0, 2, 4}
new_ends   = {8*32/128, 16*32/128, 24*32/128} = {2, 4, 6}


New Instructions:
reshape(dims={32, 4, 96})
slice(axes={0}, starts={0}, ends={2}) -> {2, 4, 96}
slice(axes={0}, starts={2}, ends={4}) -> {2, 4, 96}
slice(axes={0}, starts={4}, ends={6}) -> {2, 4, 96}
```

umangyadav:
@shivadbhavsar 
<html>
<body>


onnx-taau-downsample | 1 | 247.81 | 322.80 | -23.23% | 🔴
-- | -- | -- | -- | -- | --



</body>
</html>


Can you check this one ?

PR #2149, opened 2023-09-05
Dont insert reshapes when converting pooling to reduce

pfultz2:
This should improve fusions.


umangyadav:
Performance tests have compilation error

umangyadav:
@pfultz2  accuracy tests are still failing during compile. Can you check ?

pfultz2:
> @pfultz2 accuracy tests are still failing during compile. Can you check ?

@umangyadav I pushed a fix.

src/rewrite_pooling.cpp
```diff
@@ -54,27 +52,18 @@ void rewrite_pooling::apply(module& m) const
...
         if(not std::equal(lens.begin() + 2, lens.end(), op.lengths.begin(), op.lengths.end()))
             continue;
-        std::int64_t n = s.lens()[0];
-        std::int64_t c = s.lens()[1];
-        auto reshape   = m.insert_instruction(
-            ins, make_op("reshape", {{"dims", {n * c, -1}}}), ins->inputs().front());
-        instruction_ref pooling{};
-
+        std::vector<std::int64_t> axes(lens.size() - 2);
+        std::iota(axes.begin(), axes.end(), 2);
```
umangyadav, line 56: Should this be doing any checks for the Layout ?

src/simplify_reshapes.cpp
```diff
@@ -122,6 +122,11 @@ struct find_nop_reshapes
         reshapes.insert("pad");
         reshapes.insert("slice");
         reshapes.insert("transpose");
+        reshapes.insert("reduce_mean");
```
umangyadav, line 125: Would be better to assert that axes are either empty resulting in no-op or 
axes  lengths are 1.

umangyadav:
> This should improve fusions. 

is it improving reduce fusion ? Which of the matcher is helping doing that ?

umangyadav (approved):

src/rewrite_pooling.cpp
umangyadav: @pfultz2  sorry i missed this comment before approving. Would this require check on the layout ?

TedThemistokleous (approved):

Issue #2144, opened 2023-08-31
Do not send any convolutions or dot products whose outputs have non-standard shapes to MLIR


krzysz00:
Currently, the way that MIGraphX lowers to MLIR means that, for example, an NHWC convolution will produce a standard-shape output (NCHW). However, the MIGraphX code will expect the output to be in non-standard (NHWC) shape.

A full fix for this problem (or even a decent workaround to make it feasible) are longer-term tasks.

For now, to prevent correctness failures, no convolution or dot product whose output will be in a non-standard (last stride != 1, etc etc.) shape may be passed to the MLIR flow, and should be instead processed via some sort of fallback mechanism.

This ticket calls for adding that temporary disable to MIGraphX.

krzysz00:
Per discussion in an MLIR meeting, there are some interesting possibilities here:
- We _could_ support NHWC and trans(C) by adding a `migraphx.transpose` on the output - if we don't have any fusions or pointwise operations
  - (Pointwise operations would involve trying to run through and bracket everything in transposes and I'm not even sure that'd work and so on)
- That is, a quick hack to make standalone NHWC work (on Navi only, probably) is possible
- But, initially, an NHWC disable may be the better approach

PR #2150, opened 2023-09-05
Throw an error when mlir's outpus shape is non-standard

pfultz2:
This is currently supported in mlir. This can be removed in the future when mlir is updated.


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -103,7 +103,9 @@ struct mlir_op
             }
             if(ins->name() == "@return")
             {
-                return ins_shapes[ins->inputs().at(0)].with_type(type);
+                auto s = ins_shapes[ins->inputs().at(0)].with_type(type);
```
umangyadav, line 106: can the output be a tuple type ?

pfultz2: Not currently.

umangyadav (approved):

pfultz2:
Fixes: #2144.

pfultz2:
> Will migraphx be able to catch this exception and recover when running nhwc?

Yes because we first insert contiguous operators and then eliminate_contiguous will remove contiguous if there is no exception.

krzysz00:
Wait, does eliminate_contiguous run before or after fuse_mlir? If after ... how does it hit this exception?

pfultz2:
It runs after lowering.

PR #2154, opened 2023-09-06
Fixed stringutils test

tvukovic-amd:
Fixed stringutils test with unbalanced brackets


tvukovic-amd:
This fix is for fixing test case issue where brackets are unbalanced by changing src/include/migraphx/stringutils.hpp. Unit test for case where inputs are correct (which checks if output is correct) and test which throws error when input has unbalanced brackets already exists. 
The test where inputs are correct is [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/665a0cfb6e241aa11244d9be09d7415d46ee2fb5/test/stringutils.cpp#L34C1-L34C1) and test where input has unbalanced brackets and throws an error is [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/665a0cfb6e241aa11244d9be09d7415d46ee2fb5/test/stringutils.cpp#L41).

pfultz2:
>  Unit test for case where inputs are correct (which checks if output is correct) and test which throws error when input has unbalanced brackets already exists.

If we already have a test, then why didnt we catch that now when we run the tests?

tvukovic-amd:
I built MIGraphX on Ubuntu20.04 and used develop branch and this test failed. When I rebased my changes on top of develop branch, test_stringutils also failed, but with my changes test on Windows is working. It is not related to OS, this fix is related to algorithm issue:

The test fails on interpolate_string_unbalanced case throwing exception with message _cannot seek string iterator after end_, after the string iterator exceeds a valid range.

Part of the code needed to be rewritten, since in its current form it uses undefined behaviour as a condition for a while loop
`while(it != input.end())`, so I added case to throw error if we didn't find closed brackets before input end.

pfultz2 (approved):

PR #2155, opened 2023-09-06
Added fix for test_msgpack_float and test_msgpack_object

tvukovic-amd:
fixed result comparison in msgpack test with float and map


test/msgpack.cpp
```diff
@@ -76,7 +76,7 @@ TEST_CASE(test_msgpack_float)
     migraphx::value v = 3.0;
     auto buffer       = migraphx::to_msgpack(v);
     EXPECT(buffer == msgpack_buffer(3.0));
-    EXPECT(migraphx::from_msgpack(buffer) == v);
+    EXPECT(migraphx::from_msgpack(buffer).to<float>() == v.to<float>());
```
pfultz2, line 79: Why does this need to be converted to `float` first?

test/msgpack.cpp
```diff
@@ -109,7 +109,11 @@ TEST_CASE(test_msgpack_object)
     auto buffer       = migraphx::to_msgpack(v);
     EXPECT(buffer == msgpack_buffer(std::map<std::string, double>{
                          {"one", 1.0}, {"three", 3.0}, {"two", 2.0}}));
-    EXPECT(migraphx::from_msgpack(buffer) == v);
+
+    // converted to vector in the following line because in value.cpp value constructor with
+    // unordered map is creating vector<value> with map items as vector elements
```
pfultz2, line 114: Where are we using an unordered map?

tvukovic-amd: Created [buffer](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a0894c2a541a58be81d3527b6e16b8f926c9061d/test/msgpack.cpp#L109) is a map. After its creation there is check if buffer is equal with buffer initialized with map [(here)](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a0894c2a541a58be81d3527b6e16b8f926c9061d/test/msgpack.cpp#L110). 
In value definition inside [value.hpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/665a0cfb6e241aa11244d9be09d7415d46ee2fb5/src/include/migraphx/value.hpp) I haven't seen constructor for map, I found just constructor for unordered map, so because of that I used conversion to vector.

test/msgpack.cpp
tvukovic-amd: Function migraphx::from_msgpack returns value which needs to be converted to other types (except for string and bool case) when we are comparing output values. The tests above convert value to appropriate type [(here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/bded09498e09b6fbfe93e50e4b23d0208b517de5/test/msgpack.cpp#L58) is the example for int).

test/msgpack.cpp
pfultz2: > In value definition inside [value.hpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/665a0cfb6e241aa11244d9be09d7415d46ee2fb5/src/include/migraphx/value.hpp) I haven't seen constructor for map, I found just constructor for unordered map

Its not using the `std::unordered_map` constructor, its using the `std::initializer_list` constructor which will preserve the order of the list.

tvukovic-amd: I saw value constructor for std::initializer_list in [value.cpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/665a0cfb6e241aa11244d9be09d7415d46ee2fb5/src/value.cpp#L124). In this case (where size is not 2 and elements have key) it is calling `set_vector` function, so I think `to_vector()` function can be used for comparing if result values are equal.

pfultz2: `to_vector` will only compare the values and not the keys. The `==` operator for value will compare the keys as well: `migraphx::from_msgpack(buffer) == v`.

test/msgpack.cpp
```diff
@@ -100,7 +100,8 @@ TEST_CASE(test_msgpack_float)
     migraphx::value v = 3.0;
     auto buffer       = migraphx::to_msgpack(v);
     EXPECT(buffer == msgpack_buffer(3.0));
-    EXPECT(migraphx::from_msgpack(buffer) == v);
+    double epsilon = 1e-9;
+    EXPECT(std::abs(migraphx::from_msgpack(buffer).to<float>() - v.to<float>()) < epsilon);
```
pfultz2, line 104: Use `test::within_abs` instead.

test/msgpack.cpp
tvukovic-amd: Yes, but `migraphx::from_msgpack(buffer)` and `v` are different types, so compare in `==` operator is not called. Because of that it should be converted to the same type.

tvukovic-amd: Yes, you are right, we can use `==` operator, but the problem here is that `migraphx::from_msgpack(buffer)` converts any numeric value to `uint64_t` (because of that return value is casted to `int` or `float`, for example in  `EXPECT(migraphx::from_msgpack(buffer).to<int>() == v.to<int>());`). 
If we cast input numeric value which is not uint8_t to uint8_t (for example in `migraphx::value v = {{"a", (uint64_t)1.0}, {"b", "abc"}};` inside `test_msgpack_object_class` or in `migraphx::value v = {{"one", (uint64_t) 1.0}, {"three", (uint64_t) 3.0}, {"two", (uint64_t) 2.0}};` inside `test_msgpack_object`, the test passes without cast `migraphx::from_msgpack(buffer)` and `v` because when comparing their numeric values they have the same type which is `uint8_t`. 

I think the best way here is to change  `migraphx::from_msgpack(buffer)` to cast properly numeric values which somehow it  doesn't cast and to see why is that happening.

tvukovic-amd: If `migraphx::from_msgpack(buffer)` function is fixed, I won't have need to cast output to `float` in the following way which i committed in this PR to (`EXPECT(migraphx::from_msgpack(buffer).to<float>() == v.to<float>());`).

pfultz2: > but the problem here is that migraphx::from_msgpack(buffer) converts any numeric value to uint64_t

Where do you see this happening? `from_msgpack` converts the float32 or float64 to double here: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/msgpack.cpp#L85.

tvukovic-amd: I see it in the debugger during comparing `EXPECT(migraphx::from_msgpack(buffer) == v);`. You can see debugger output in the following screenshot when debugging test_msgpack_float, here x is migraphx::from_msgpack(buffer) output and y is migraphx::value v which is double. If we cast v to be `uint64_t`, the test passes.
image

tperry-amd: My suggestion since we do want to keep comparing the keys and the types are different for the values that we should either change operator== to have a specialization that can properly compare types, call the compare function directly if that will allow us to compare types or create a compare function that can compare the keys and compare with an explicit type conversion (passed as a template parameter).

test/msgpack.cpp
tvukovic-amd: I fixed it in the following [commit](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2155/commits/bf42d614b24dadc434999797e3404179d1b6a4eb).

test/msgpack.cpp
pfultz2: So there is something wrong with either msgpack serialization or the value construction as it should be `float_type` as well.

tvukovic-amd:
I changed code without cast msgpack output to expected type according to following findings on Windows (tests pass on Linux): the thing here is that `migraphx::to_msgpack(v)` and `msgpack_buffer(3.0)` from `msgpack.cpp` file which is used for this test (3.0 is some double value) both call `msgpack::pack(vs, v);` function inside `AMDMIGraphX\build\_deps\msgpack-src\include\msgpack\v1\pack.hpp` on Windows.

That function is calling `msgpack::operator<<(*this, v);` which after couple of steps calls `pack_double` function inside the same `pack.hpp` file with following conditions:

```
template <typename Stream>

inline packer<Stream>& packer<Stream>::pack_double(double d)

{

    if(d == d) { // check for nan 

        // compare d to limits to avoid undefined behaviour

        if(d >= 0 && d <= double(std::numeric_limits<uint64_t>::max()) && d == double(uint64_t(d))) {

            pack_imp_uint64(uint64_t(d));

            return *this;        

        } else if(d < 0 && d >= double(std::numeric_limits<int64_t>::min()) && d == double(int64_t(d))) {

            pack_imp_int64(int64_t(d));

            return *this;

        }

    }



    union { double f; uint64_t i; } mem;

    mem.f = d;

    char buf[9];

    buf[0] = static_cast<char>(0xcbu);



#if defined(TARGET_OS_IPHONE)

    // ok

#elif defined(__arm__) && !(__ARM_EABI__) // arm-oabi

    // https://github.com/msgpack/msgpack-perl/pull/1

    mem.i = (mem.i & 0xFFFFFFFFUL) << 32UL | (mem.i >> 32UL);

#endif

    _msgpack_store64(&buf[1], mem.i);

    append_buffer(buf, 9);

    return *this;

}
```

The thing here is that when we have `double` input which is same as `d == double(int64_t(d))` and is in appropriate range, for example (1.0, 3.0 etc.), first or second conditions pass (inside `d==d` condition) from code above and we get result from pack function which is `uint64_t` or `int64_t` type because here it is converted to it and it is **not** `double` type which is expected to be.

If I change inputs inside test code for example from 3.0 to 3.01 or whatever value which is not same as value after conversion to `int64_t` or `uint64_t` and then to `double`,  all tests pass because conditions from code above are not met and we have returned `double` value which is expected. If we leave 3.0 it will be `uint64` or `int64` and because of that won't pass the test since it doesn't pass type check since their types are different.

umangyadav (approved):

pfultz2 (approved):

PR #2162, opened 2023-09-07
Auto generate root modules given target assignments for multi-target compilation

umangyadav:
~Makes use of #2145  therefore some of the unit-tests fails.~ 

multitarget unit-tests passes. 

Part of #1656


test/multi_target/multitarget_test.cpp
```diff
@@ -337,82 +312,55 @@ TEST_CASE(multitarget_compile_nested_if_then_else)
...
+    tass.insert(tass.begin(), std::make_pair(then_mod_ref_ins, 3));
+    auto then_mod_if =
+        then_mod->add_instruction(migraphx::make_op("if"),
+                                  {then_mod_cond,
+                                   then_mod_param_0,
+                                   then_mod_param_1,
+                                   then_mod_param_2,
+                                   then_mod_ref_ins,
+                                   then_mod_param_1,
+                                   then_mod_param_2},
```
pfultz2, line 337: The `create_test_module` only creates 3 parameters, so this seems wrong to pass 6 parameters.

umangyadav: It creates 3 parameters but for both then and else modules have different parameter names and therefore i need to pass 6 parameters.

causten:
This work is being shelved, but will not delete the branch incase we revisit

PR #2165, opened 2023-09-08
Use parallel STL for parallel execution

pfultz2:
Added a par.hpp header with parallel algorithms wrappers. When its not supported it will use simple parallel implementation or just use the serial version.


tools/install_prereqs.sh
```diff
@@ -52,7 +52,8 @@ else
     python3-dev \
     python3-pip \
     rocblas-dev \
-    rocm-cmake
+    rocm-cmake \
+    libtbb-dev
```
TedThemistokleous, line 56: Are we now not worried about other stuff breaking with libtbb-dev? I think its moved to onetbb right now

TedThemistokleous: My original concern was the SLES docker and future other dockers I saw with CentOS. I think they don't have a libtbb package but something else.

Its fine. Looks like your docker passed, we can cross the CentOS issues later.

CharlieL7:
AFAIK `<executors>` is still in development for the C++ standard library? What benefit does OneTBB offer over the current implementation?

pfultz2:
> AFAIK <executors> is still in development for the C++ standard library?

We arent using executors. We are just using the [execution policies](https://en.cppreference.com/w/cpp/algorithm#Execution_policies) provided by C++17(which has already been finalized for quite a few years). Maybe I need rename the macro to make that clearer.

> What benefit does OneTBB offer over the current implementation?

What exactly are you asking? The current implementation as in this PR or the current implementation on develop?

Also, we arent using TBB directly, we are just calling to standard C++. On windows there is no TBB dependency.

src/include/migraphx/op/unary.hpp
```diff
@@ -84,10 +85,10 @@ struct unary : op_name<Derived>
         argument result{dyn_out.computed_shape};
         result.visit([&](auto output) {
             args[0].visit([&](auto input) {
-                std::transform(input.begin(),
-                               input.end(),
-                               output.begin(),
-                               static_cast<const Derived&>(*this).apply());
+                par_transform(input.begin(),
```
lakhinderwalia, line 88: How can any unary operation (i.e. calling std::transform here) be assumed to be parallelizable? Same question for binary operations using std::transform. Thanks.

src/include/migraphx/par.hpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+template <class InputIt, class UnaryFunction>
+void par_for_each(InputIt first, InputIt last, UnaryFunction f)
+{
+#if MIGRAPHX_HAS_EXECUTORS
+    // Propagate the exception
+    detail::exception_list ex;
+    std::for_each(std::execution::par, first, last, ex.collect(std::move(f)));
+    ex.throw_if_exception();
```
lakhinderwalia, line 105: Why is exception handled here at all, when the std::for_each should behave in a certain way, and this will alter that behavior. Thanks.

src/include/migraphx/par.hpp
```diff
@@ -0,0 +1,133 @@
...
+            catch(...)
+            {
+                this->add_exception();
+            }
+        };
+    }
+    void throw_if_exception() const
+    {
+        if(not exceptions.empty())
+            std::rethrow_exception(exceptions.front());
```
lakhinderwalia, line 68: (Apart from the question of whether to even use it..)
 Would two different type of exceptions (from say two different threads) throw (the same exception) exceptions.front()?

lakhinderwalia: Nit. Please fix copyright dates in all the files modified/added. Thanks.

src/include/migraphx/simple_par_for.hpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+template <class F>
+auto thread_invoke(std::size_t i, std::size_t, F f) -> decltype(f(i))
+{
+    f(i);
+}
+
+template <class F>
+void simple_par_for_impl(std::size_t n, std::size_t threadsize, F f)
```
lakhinderwalia, line 66: For laypersons like me, please document what is n, threadsize, f.

src/include/migraphx/simple_par_for.hpp
```diff
@@ -0,0 +1,119 @@
...
+    ~joinable_thread()
+    {
+        if(this->joinable())
+            this->join();
+    }
+};
+
+template <class F>
+auto thread_invoke(std::size_t i, std::size_t tid, F f) -> decltype(f(i, tid))
+{
```
lakhinderwalia, line 55: Is `i` some kind of an index? For laypersons, could you please rename the variable, or document it.. Thanks.

src/include/migraphx/op/unary.hpp
pfultz2: The `apply` cannot have side effects(unless you do bad things) because it only takes the implicit `this` as a parameter and that is already `const` since `compute` is a `const` method.

src/include/migraphx/par.hpp
pfultz2: > Would two different type of exceptions (from say two different threads) throw (the same exception) exceptions.front()?

Two different threads could throw different exception. We only propagate the first one, which is enough to know that an error ocurred.

The previous behavior with `par_for` also only propagated the first exception, although it aborted the other threads so no other exception could occur. Whereas here there can be multiple exceptions but we only care about the first one.

src/include/migraphx/par.hpp
pfultz2: It breaks some unit tests that tries to detect if an error occurs. I think its better to have an exception thrown which could possibly be recovered than to have `std::terminate` called. We could update the other algorithms the same way, but for now I updated this which fixes an unit tests.

src/include/migraphx/simple_par_for.hpp
pfultz2: `i` is the loop variable or index. Usually this is named `i`.

pfultz2: Also this is just moved from `par_for.hpp` so I am not adding this as new.

src/include/migraphx/simple_par_for.hpp
pfultz2: This is an impl function so its not part of the interface.

src/include/migraphx/par.hpp
lakhinderwalia: The logical way to recover from exceptions would be to handle each of those exceptions -- not just the first one, which isn't happening here.. ; However, as a philosophy, the parallel mode shouldn't be generating exceptions.. so which unit test is failing here specifically, please. Maybe that instance shouldn't be parallelized. Thanks.

src/include/migraphx/op/unary.hpp
lakhinderwalia: OK. Could you please then add a comment here explaining why it is being used: that `apply` can't have side-effects, hence it is parallelizable. (And also to binary.hpp). Thanks.

src/include/migraphx/simple_par_for.hpp
lakhinderwalia: Here it is not a loop variable, but a parameter to thread_invoke, so I would like to know what `i` is, without having to go across functions or files reading code -- and I still don't know what it is here :-)  
Thanks.

src/include/migraphx/par.hpp
pfultz2: > The logical way to recover from exceptions would be to handle each of those exceptions -- not just the first one, which isn't happening here.. ;

We can only rethrow one exception at a time. We could wrap the exception but then we lose the backtrace when we rethrow. 

> However, as a philosophy, the parallel mode shouldn't be generating exceptions.

The only way to report errors from algorithms is to throw an exception. Checking for the errors before hand can be costly as well. Since we are calling many standard library functions which are `noexcept` then we need to recover from the errors for several reasons:

1) Unit tests can check that such errors are check for.
2) In the API functions we need to convert any exception to en error code, terminating the application is not a good option at all.

Furthermore, this is what is suggested by Anthony Williams the author of the  [C++ Concurrency in Action](https://www.manning.com/books/c-plus-plus-concurrency-in-action) book:

https://stackoverflow.com/a/55354021/375343

The C++ standard terminates the program since there is a lot of different ways to handle the multiple exceptions and not a clear best way. For now, I just throw the first exception. This should be good enough for our uses cases(unit test check the exception and the API returns an error code). 

> so which unit test is failing here specifically, please.

The gathernd check for out of bounds error here will fail:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/ref/gathernd.cpp#L389

Which checking for this exception:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/op/gathernd.hpp#L198

Its too costly to move it out of the algorithm(as we would need to run a serial version and parallel version). 

> Maybe that instance shouldn't be parallelized.

Just because we check for out of bounds access and throw an exception doesnt mean we should not make it parallelized. And I see no compelling reason to have parallel algorithms not throw exceptions.

src/include/migraphx/simple_par_for.hpp
pfultz2: > Here it is not a loop variable

Its the loop variable that gets passed to this function:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/par_for.hpp#L71

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/par_for.hpp#L90

The variables are named the same as the variables that are passed to make it easier to follow where the variables go in the function. 


> without having to go across functions or files reading code -- and I still don't know what it is here

Thats because you are trying to read the code bottom-up. If you read it top-down then its easier to see how everything is being called.

src/include/migraphx/par.hpp
lakhinderwalia: Thanks for the detailed explanation :-) But I was referring just to the approach taken here: to parallelize std::transform() into par_for**, and wanted to understand the reason of test case failure. 

Shouldn't bounds checking be done outside of the parallelization logic..? Thanks.

src/include/migraphx/par.hpp
lakhinderwalia: I wasn't clear enough perhaps. My point was, would two threads if they both hit some exceptions, they would rethrow the same exception twice.  Both threads will enter throw_if_exception(), and find exceptions list not empty. Rethrow should be only once. Thanks.

src/include/migraphx/simple_par_for.hpp
lakhinderwalia: Not to belabor too much.. but I was making a simple point here. That I as a reviewer should be able to look at the changes and make sense of what is going on.. without going too far reading the other code. 

Function parameter names should be named for clarity, and not along how (internal) variables are named/used somewhere else -- that is unintuitive for the uninitiated.. 

We have spent here far more energy and lines of explanation on this simple function parameter `i`. A two line comment there in the code (or the renaming of parameter `i`) would help. But I am not going to discuss this any further.. Thanks.

src/include/migraphx/par.hpp
pfultz2: > But I was referring just to the approach taken here: to parallelize std::transform() into par_for**

I dont understand what you are referring to. `par_for` will propagate the exception. 

> Shouldn't bounds checking be done outside of the parallelization logic..?

That is too slow. We would need to practically do everything serially first to chekc the bounds and then run it in parallel which doesnt make sense.

src/include/migraphx/par.hpp
pfultz2: >  Both threads will enter throw_if_exception(), 

No, the `throw_if_exception` happens after the algorithm is called, and is not ran in parallel.

> Rethrow should be only once.

It is only once.

src/include/migraphx/simple_par_for.hpp
pfultz2: > That I as a reviewer should be able to look at the changes 

But I am not making changes here. The changes is renaming the function from `par_of` to `simple_par_for`. 

> without going too far reading the other code.

You need to read the code to figure out whats going on. This is a fairly simple piece of code(ie invoke a function) that it should be easy to follow.

src/include/migraphx/par.hpp
lakhinderwalia: You are right. Thanks.

causten:
```
[ 54%] Building CXX object test/CMakeFiles/header_src_include_migraphx_instruction_ref_hpp.dir/header-main-include-header_src_include_migraphx_instruction_ref_hpp.cpp.o
cd /home/jenkins/workspace/AMDMIGraphX_PR-2340/build/test && ccache /opt/rocm/llvm/bin/clang++ -DBUILD_DEV -DHAS_HALF_V1 -DHAVE_GPU -DMIGRAPHX_HAS_FIND_2_API -DMIGRAPHX_HAS_FIND_MODE_API -DMIGRAPHX_PREALLOCATE_MIOPEN_BUFFERS -DMIGRAPHX_USE_ROCBLAS_TUNING_API -DROCBLAS_BETA_FEATURES_API -DROCBLAS_NO_DEPRECATED_WARNINGS -D__HIP_PLATFORM_AMD__=1 -D__HIP_PLATFORM_HCC__=1 -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/test/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/build/src/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/build/src/onnx/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/build/src/tf/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/build/src/targets/ref/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/targets/ref/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/build/src/targets/gpu/include -I/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/targets/gpu/include -isystem /usr/local/include/half -isystem /opt/rocm-5.7.0/include -isystem /opt/rocm/include -Werror -O3 -DNDEBUG -std=c++17 -Wall -Wextra -Wcomment -Wendif-labels -Wformat -Winit-self -Wreturn-type -Wsequence-point -Wswitch -Wtrigraphs -Wundef -Wuninitialized -Wunreachable-code -Wunused -Wno-sign-compare -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-conversion -Wno-double-promotion -Wno-exit-time-destructors -Wno-extra-semi -Wno-extra-semi-stmt -Wno-float-conversion -Wno-gnu-anonymous-struct -Wno-gnu-zero-variadic-macro-arguments -Wno-missing-prototypes -Wno-nested-anon-types -Wno-option-ignored -Wno-padded -Wno-shorten-64-to-32 -Wno-sign-conversion -Wno-unused-command-line-argument -Wno-weak-vtables -Wno-c99-extensions -MD -MT test/CMakeFiles/header_src_include_migraphx_instruction_ref_hpp.dir/header-main-include-header_src_include_migraphx_instruction_ref_hpp.cpp.o -MF CMakeFiles/header_src_include_migraphx_instruction_ref_hpp.dir/header-main-include-header_src_include_migraphx_instruction_ref_hpp.cpp.o.d -o CMakeFiles/header_src_include_migraphx_instruction_ref_hpp.dir/header-main-include-header_src_include_migraphx_instruction_ref_hpp.cpp.o -c /home/jenkins/workspace/AMDMIGraphX_PR-2340/build/test/header-main-include-header_src_include_migraphx_instruction_ref_hpp.cpp
make  -f test/CMakeFiles/header_src_include_migraphx_iota_iterator_hpp.dir/build.make test/CMakeFiles/header_src_include_migraphx_iota_iterator_hpp.dir/depend
make[2]: Entering directory '/home/jenkins/workspace/AMDMIGraphX_PR-2340/build'
cd /home/jenkins/workspace/AMDMIGraphX_PR-2340/build && /usr/local/lib/python3.8/dist-packages/cmake/data/bin/cmake -E cmake_depends "Unix Makefiles" /home/jenkins/workspace/AMDMIGraphX_PR-2340 /home/jenkins/workspace/AMDMIGraphX_PR-2340/test /home/jenkins/workspace/AMDMIGraphX_PR-2340/build /home/jenkins/workspace/AMDMIGraphX_PR-2340/build/test /home/jenkins/workspace/AMDMIGraphX_PR-2340/build/test/CMakeFiles/header_src_include_migraphx_iota_iterator_hpp.dir/DependInfo.cmake --color=
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/targets/gpu/driver/precompile_op.cpp:50:13: error: no matching member function for call to 'add_instruction'
        mm->add_instruction(preop, args);
        ~~~~^~~~~~~~~~~~~~~
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/include/migraphx/module.hpp:81:21: note: candidate function not viable: 'this' argument has type 'const module', but method is not marked const
    instruction_ref add_instruction(const operation& op, std::vector<instruction_ref> args);
                    ^
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/include/migraphx/module.hpp:76:21: note: candidate template ignored: requirement '(migraphx::and_<false>{})' was not satisfied [with Ts = <std::vector<instruction_ref>>, PrivateRequires75 = 75]
    instruction_ref add_instruction(operation op, Ts... args)
                    ^
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/include/migraphx/module.hpp:83:21: note: candidate function not viable: requires 3 arguments, but 2 were provided
    instruction_ref add_instruction(const operation& op,
                    ^
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/targets/gpu/driver/precompile_op.cpp:48:20: error: 'this' argument to member function 'add_parameter' has type 'const module', but function is not marked const
            return mm->add_parameter("x" + std::to_string(i), input);
                   ^~
/usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/stl_algo.h:4382:14: note: in instantiation of function template specialization 'migraphx::gpu::driver::precompile_op::create_preop_program(const operation &, std::vector<shape>)::(anonymous class)::operator()<migraphx::shape, long>' requested here
        *__result = __binary_op(*__first1, *__first2);
                    ^
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/include/migraphx/ranges.hpp:204:10: note: in instantiation of function template specialization 'std::transform<__gnu_cxx::__normal_iterator<migraphx::shape *, std::vector<migraphx::shape>>, migraphx::basic_iota_iterator<migraphx::id>, std::back_insert_iterator<std::vector<std::_List_iterator<migraphx::instruction>>>, (lambda at /home/jenkins/workspace/AMDMIGraphX_PR-2340/src/targets/gpu/driver/precompile_op.cpp:47:75)>' requested here
    std::transform(r1.begin(), r1.end(), r2.begin(), it, f);
         ^
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/targets/gpu/driver/precompile_op.cpp:47:9: note: in instantiation of function template specialization 'migraphx::transform<std::vector<migraphx::shape> &, migraphx::iterator_range<migraphx::basic_iota_iterator<migraphx::id>>, std::back_insert_iterator<std::vector<std::_List_iterator<migraphx::instruction>>>, (lambda at /home/jenkins/workspace/AMDMIGraphX_PR-2340/src/targets/gpu/driver/precompile_op.cpp:47:75)>' requested here
        transform(inputs, range(inputs.size()), std::back_inserter(args), [&](auto input, auto i) {
        ^
/home/jenkins/workspace/AMDMIGraphX_PR-2340/src/include/migraphx/module.hpp:161:21: note: 'add_parameter' declared here
    instruction_ref add_parameter(std::string name, shape s);
                    ^
```

causten:
@pfultz2 ASAN failures

Issue #1986, opened 2023-07-20
Add `-fno-hip-uniform-block` flag for the compilation when necessary


umangyadav:
HIP requires uniform WG sizes otherwise HIP runtime would through errors or it may result in undefined behaviors. 

Using uniform WG sizes always however may result in performance degradation.  

HIP is going to add `-fno-hip-uniform-block` flag to allow users to use non-uniform WG sizes. 

https://reviews.llvm.org/D155213

it is supposed to become available in either rocm-5.7 or rocm-6.0. When it does, make use of it. 

Related issue : 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1976
SWDEV-406592

PR #2167, opened 2023-09-11
Add flag to accept non-uniform WG sizes

umangyadav:
fixes https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1986
Depends on  #2168 
Relevant issues:
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2307

https://github.com/ROCmSoftwarePlatform/composable_kernel/pull/838


src/targets/gpu/CMakeLists.txt
```diff
@@ -204,6 +204,12 @@ if(MIGRAPHX_ENABLE_MLIR)
     target_link_libraries(migraphx_gpu PRIVATE rocMLIR::rockCompiler)
 endif()
 
+math(EXPR hip_VERSION_FLAT "(${hip_VERSION_MAJOR} * 1000 + ${hip_VERSION_MINOR}) * 100000 + ${hip_VERSION_PATCH}")
+if(${hip_VERSION_FLAT} GREATER_EQUAL 500700000)
+    set(MIGRAPHX_HIP_REQUIRE_UNIFORM_WG ON CACHE BOOL "Whether HIP requires uniform workgroup sizes or not")
+    target_compile_definitions(migraphx_gpu PUBLIC -DMIGRAPHX_HIP_REQUIRE_UNIFORM_WG)
+endif()
```
pfultz2, line 211: This should be checked at runtime, and it should check if the `-fno-offload-uniform-block` flag can be passed to hiprtc by trying to compile a sample program instead of using versions which can be problematic.

pfultz2: I am actually working on adding support to check for flags as runtime to work with disabling newer warnings.

umangyadav: Runtime check is better. For hip-clang i think we can just make use of `check_cxx_compiler_flag` but i wasn't sure how to do it for the hiprtc.  

For the runtime check for the hiprtc, how would that work ? because it is runtime compilation and wouldn't it need to check for each and every compile ? not sure if can be cached somehow

pfultz2: See #2168, I added the function `hip_has_flags` to check for a warning, which works for both hipclang and hiprtc. You could use the same to check for non-uniform blocks as well. 

> wouldn't it need to check for each and every compile ?

You can assign it to a static variable so it only runs the check once.

umangyadav: Okay. Made use of #2168

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -187,6 +194,10 @@ operation compile_hip_code_object(const std::string& content, hip_compile_option
         generate_args_hpp(options.virtual_inputs.empty() ? options.inputs : options.virtual_inputs);
     srcs.push_back(src_file{fs::path{"args.hpp"},
                             std::make_pair(args_hpp.data(), args_hpp.data() + args_hpp.size())});
+    if(options.global % options.local != 0 and hip_accept_non_uniform_wg())
+        options.params += " -fno-offload-uniform-block";
+    else
+        assert(options.global % options.local == 0);
```
pfultz2, line 200: This will throw an assert on older rocm that doesnt support the flag. If the flag is not supported than most likely non-uniform blocks is supported by default. I think we can remove this assert.

umangyadav: if flag is not supported then it would have calculated the global WG multiple of local WG. It would work with older rocms as well. 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1976

umangyadav: >  If the flag is not supported than most likely non-uniform blocks is supported by default 

That isn't the case with 5.7. Neither the Flag nor non-uniform blocks are supported.

src/targets/gpu/kernels/include/migraphx/kernels/index.hpp
```diff
@@ -36,44 +36,33 @@ namespace migraphx {
...
 inline __device__ __attribute__((const)) index_int compute_global_size()
 {
 #ifdef MIGRAPHX_NGLOBAL
     return MIGRAPHX_NGLOBAL;
 #else
-    // This actualy works even when global is not divisible by local size.
-    // This doesnt actually do a multiplicatiosn. Instead it calls a device
-    // function to get the global size, which is why it works.
-    return blockDim.x * gridDim.x;  // NOLINT
+    return compute_max_local_size() * gridDim.x; // NOLINT
```
pfultz2, line 54: This is wrong, this will compute the wrong global size when using non-uniform tiles. It should be `blockDim.x * gridDim.x`(which calls `__ockl_get_global_size` see [here](https://github.com/ROCm-Developer-Tools/clr/blob/5914ac3c6e9b3848023a7fa25e19e560b1c38541/hipamd/include/hip/amd_detail/amd_hip_runtime.h#L324C10-L324C35)) or we could call `__ockl_get_global_size(0)` directly.

Also, this creates an infinite recursion because `compute_max_local_size` will call `compute_global_size` when assertions are enabled.

src/targets/gpu/kernels/include/migraphx/kernels/index.hpp
```diff
@@ -124,11 +113,7 @@ struct index
 #ifdef MIGRAPHX_NLOCAL
     constexpr index_constant<MIGRAPHX_NLOCAL> max_nlocal() const { return {}; }
 #else
-    __device__ index_int max_nlocal() const
-    {
-        MIGRAPHX_ASSERT(blockDim.x > 0);
-        return blockDim.x;
```
pfultz2, line 130: Dont delete this, just call `compute_max_local_size` here.

src/targets/gpu/kernels/include/migraphx/kernels/index.hpp
```diff
@@ -36,44 +36,33 @@ namespace migraphx {
 #if defined(MIGRAPHX_NGLOBAL) && defined(MIGRAPHX_NLOCAL)
 #define MIGRAPHX_NGROUP ((MIGRAPHX_NGLOBAL + MIGRAPHX_NLOCAL - 1) / MIGRAPHX_NLOCAL)
 #endif
+inline __device__ __attribute__((const)) index_int compute_max_local_size()
+{
+#ifdef MIGRAPHX_NLOCAL
+    return MIGRAPHX_NLOCAL;
+#else
+    MIGRAPHX_ASSERT(compute_global_size() % blockDim.x == 0);
+    return blockDim.x;
```
pfultz2, line 45: So Brian Sumner mentioned, we could call `__ockl_get_enqueued_local_size(0)` to get the uniform size instead removing the need for the assert.

src/targets/gpu/kernels/include/migraphx/kernels/index.hpp
umangyadav: That's what it's doing.

src/targets/gpu/kernels/include/migraphx/kernels/index.hpp
umangyadav: > This is wrong, this will compute the wrong global size when using non-uniform tiles. 

For the non-uniform block `blockDim.x` would return dimension of non-uniform block and it wouldn't calculate global size correctly.

src/targets/gpu/kernels/include/migraphx/kernels/index.hpp
```diff
@@ -36,44 +36,33 @@ namespace migraphx {
...
-    const auto nglobal  = compute_global_size();
-    if(group_id == ngroup - 1)
-    {
-        return 1 + (nglobal - 1) % nlocal;
-    }
-    else
-    {
-        return nlocal; // NOLINT
-    }
+    return nlocal; // NOLINT
```
pfultz2, line 65: This is wrong as well. This will return the max size when `MIGRAPHX_NLOCAL` is defined. It should do this instead:

```cpp
inline __device__ __attribute__((const)) index_int compute_local_size()
{
#ifdef MIGRAPHX_HAS_CONST_LOCAL
    return MIGRAPHX_NLOCAL;
#else
    return blockDim.x;
#endif
}
```

And instead of relying on `blockDim.x` which seems to change behavior with non-uniform blocks, we could call `__ockl_get_local_size(0)` directly to get the correct size.

src/targets/gpu/kernels/include/migraphx/kernels/index.hpp
umangyadav: `__ockl_get_enqueued_local_size(0)` gets classified as reserved since it starts with `__`

umangyadav:
> @umangyadav Do we have a verify test that uses the non-uniform blocks?

yes `test_slice2` and `roialign` tests.

src/targets/gpu/compile_hip_code_object.cpp
```diff
@@ -139,20 +139,27 @@ void hip_compile_options::set_launch_params(
...
-        // [TODO]: consider adding "fno-hip-uniform-block" flag when it becomes available.
-        // https://reviews.llvm.org/D155213
-        std::size_t num_elements = ((n + local - 1) / local) * local;
-        std::size_t groups       = (num_elements + local - 1) / local;
-        std::size_t max_blocks   = max_global / local;
-        std::size_t nglobal      = std::min(max_blocks * over, groups) * local;
+        std::size_t num_elements = n;
+        if(not hip_accept_non_uniform_wg())
+        {
+            num_elements = (1 + (n - 1) / local) * local;
```
TedThemistokleous, line 158: why not just do 

local + n - 1, saves a divide and mul op here.

umangyadav: It will not result in same. division is rounding towards zero.

TedThemistokleous: So we're alright with the floor(x) here then when we do the div? Thats why there's the +1 in groups below?

umangyadav: `(1 + (n - 1) / local)` is doing ceil division.

TedThemistokleous (approved):

src/targets/gpu/compile_hip_code_object.cpp
TedThemistokleous: Got it all good.

atamazov:
@umangyadav Is seems like you have some code that checks if compiler supports some option at runtime. FYI this is how we use check_cxx_compiler_flag() to check the same at build time: https://github.com/ROCm/MIOpen/pull/2741

umangyadav:
we want to check at runtime that's why it is done that way. OTherwise we can also use `check_cxx_compiler_flag. `

PR #2168, opened 2023-09-11
Disable unsafe buffer usage warning when its available

pfultz2:


src/targets/gpu/compile_hip.cpp
```diff
@@ -323,6 +351,30 @@ compile_hip_src(const std::vector<src_file>& srcs, std::string params, const std
...
+    compiler.compiler = MIGRAPHX_STRINGIZE(MIGRAPHX_HIP_COMPILER);
+    compiler.flags =
+        join_strings(flags, " ") + " -x hip -c --offload-arch=gfx900 --cuda-device-only";
+
+    std::string src;
+    src_file input;
+    input.path    = "main.cpp";
+    input.content = std::make_pair(src.data(), src.data() + src.size());
+
+    // compiler.compile({input});
```
umangyadav, line 366: remove this line if not needed

PR #2169, opened 2023-09-12
Added support for standalone dot operations with mlir

ravil-mobile:
The PR enables the code generation of standalone dot operations with MLIR.

~~@manupak, could you, please, have a look at this PR. If everything is ok I will ask Paul to make a review.~~

Hi @pfultz2, could you, please, make a review of this PR? It extends PR #2142 by adding the option to generate standalone `dot` operations.

### DIFF
--- a/src/targets/gpu/fuse_mlir.cpp
+++ b/src/targets/gpu/fuse_mlir.cpp
@@ -299,10 +299,8 @@ struct find_mlir_fused_ops
     }
 };
 
-struct find_mlir_standalone_convolution_op
+struct find_mlir_standalone_op
 {
-    auto matcher() const { return match::name("convolution"); }
-
     void apply(module_pass_manager& mpm, const match::matcher_result& r) const
     {
         auto conv_based_op = r.result;
@@ -324,6 +322,16 @@ struct find_mlir_standalone_convolution_op
     }
 };
 
+struct find_mlir_standalone_convolution_op : find_mlir_standalone_op
+{
+    auto matcher() const { return match::name("convolution"); }
+};
+
+struct find_mlir_standalone_dot_op : find_mlir_standalone_op
+{
+    auto matcher() const { return match::name("dot"); }
+};
+
 /**
  * @brief Declares a new MIGraphX environment variable which forces to generate
  * only specific MLIR operations.
@@ -331,7 +339,7 @@ struct find_mlir_standalone_convolution_op
  * The variable, if defined, forces MIGraphX to use only specific operations
  * with MLIR regardless of the underlying GPU architecture. The variable accepts
  * a list of operations separated by comma. The variable recognizes the following
- * operations: "fused", "convolution". If the variable is not defined MIGraphX
+ * operations: "fused", "convolution", "dot". If the variable is not defined MIGraphX
  * will decide by itself which operations to delegate to MLIR. The variable is
  * intended to be primarily used by rocMLIR developers.
  */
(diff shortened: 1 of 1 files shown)


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -325,14 +323,24 @@ struct find_mlir_standalone_convolution_op
     }
 };
 
+struct find_mlir_standalone_convolution_op : find_mlir_standalone_op
+{
+    auto matcher() const { return match::name("convolution"); }
+};
+
+struct find_mlir_standalone_dot_op : find_mlir_standalone_op
```
manupak, line 331: The module name need to be unique;
right now, 
```c++
module_ref mm = mpm.create_module("mlir_" + std::to_string(counter++));
```
will create name collisions for standalone convs and gemms.

So my suggestion would be make the op name a string in find_mlir_standalone_op struct and do something like : 
```c++
match::find_matches(mpm, find_mlir_standalone_op{"dot"});
```
```c++
match::find_matches(mpm, find_mlir_standalone_op{"convolution"});
```

manupak (commented):
There would be module name collisions. I have suggested one way to fix it.

src/targets/gpu/fuse_mlir.cpp
manupak: I know this because I tried this with resnet yesterday :)

ravil-mobile: But, `counter` is a static and it is in the base class.  `counter` will be incremented regardless whether if was called from `find_mlir_standalone_convolution_op` or `find_mlir_standalone_dot_op`.

I also used resnet and there was no name collisions

manupak: oh right you have the static var in the base class where as in I just copied for my experiment. So ignore my comment then.

ravil-mobile:
> There would be module name collisions. I have suggested one way to fix it.

Resolved

manupak (approved):
LGTM!

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -373,6 +381,15 @@ bool is_standalone_convs_enabled(context* ctx)
...
 }
+
+bool is_standalone_dots_enabled()
+{
+    if(is_self_decide())
+    {
+        return false;
+    }
+    return is_requested("dot");
+}
```
pfultz2, line 392: Just make a `is_standalone_op_enabled` that takes the op name as a parameter.

ravil-mobile: @pfultz2. Maybe, instead of `is_standalone_op_enabled`, we can generalize it for all ops and call it, for example, `is_op_enabled`?

What do you think about the following code?

```cpp
bool is_op_enabled(std::string_view op_name, context* ctx)
{
    if(is_self_decide())
    {
        if(op_name == "fused")
        {
            return true;
        }
        else if(op_name == "convolution")
        {
            if(ctx == nullptr)
            {
                return false;
            }
            else
            {
                const auto& device = ctx->get_current_device();
                const std::string navi_family{"gfx110"};
                return starts_with(device.get_gfx_name(), navi_family);
            }
        }
        else
        {
            return false;
        }
    }
    return is_requested(op_name);
}

} // namespace

#endif // MIGRAPHX_MLIR

void fuse_mlir::apply(module_pass_manager& mpm) const
{
#ifdef MIGRAPHX_MLIR
    if(is_op_enabled("fused", this->ctx))
    {
        match::find_matches(mpm, find_mlir_fused_ops{});
    }

    if(is_op_enabled("convolution", this->ctx))
    {
        match::find_matches(mpm, find_mlir_standalone_convolution_op{});
    }

    if(is_op_enabled("dot", this->ctx))
    {
        match::find_matches(mpm, find_mlir_standalone_dot_op{});
    }
#else
    (void)mpm;
#endif
}
```

pfultz2: Yea thats a good idea.

ravil-mobile: Thanks! Done.

pfultz2 (approved):

causten:
@ravil-mobile  don't merge develop in to this PR anymore, I have it in the merge pipeline and additional merges overload CI.

ravil-mobile:
> @ravil-mobile don't merge develop in to this PR anymore, I have it in the merge pipeline and additional merges overload CI.

Ok, I didn't realize that there was a dedicated merge pipeline

PR #2175, opened 2023-09-12
Remove redundant headers from ref ops tests

ahsan-ca:
`onnx.hpp` includes `program.hpp` which is what we actually need. This updates the tests to reflect this change.
Also, removes unnecessary includes from some other files (`loop.cpp`, `dot_op.cpp` and `rnn_ops.cpp`)


umangyadav (approved):

ahsan-ca:
Rebased on top of `develop`

PR #2179, opened 2023-09-13
Update ONNX python backend tests

attila-dusnoki-htec:
~Note: depends on #2121, until marked as Draft~

This PR enables (almost) all tests with onnx.
There were previosly skipped test that are actually implemented already. And there are unsupported operators, which are needed to be disabled.
There are a bunch of tests that covers operators, but fails with the tests because it is partially supported.

Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/131


attila-dusnoki-htec:
cc @causten

umangyadav (approved):

causten (approved):

PR #2184, opened 2023-09-13
Enable MLIR to be built with MIGraphX

causten:
Build in rocMLIR and use it for optimizations when using the MIGRAPHX_ENABLE_MLIR=1 environment variable


pfultz2:
Rbuild.ini file needs to be set to ignore MLIR for the `gh` session. Otherwise we will be building MLIR on the free github actions nodes, which will be very slow.

pfultz2:
See here: 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2037/files#diff-29b948945fc935e2fa622b1869bf7d1e637f2d91a138c1243c96c2c6e945f1c4R9

krzysz00:
While you're here, could you do a commit bump to `16523c21c40acc4836011a8dc3b80e9bb1f4bbc8` to pull in some fixes that might've been causing our tuning issues?

causten:
> Rbuild.ini file needs to be set to ignore MLIR for the `gh` session. Otherwise we will be building MLIR on the free github actions nodes, which will be very slow.

That file was never revered so rbuild.ini still shows it as disabled
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/rbuild.ini#L11

TedThemistokleous (approved):

manupak:
@causten can you bump the MLIR commit to 95b72605ae421dbe177c8e677477459bd0bc18de ?

PR #2188, opened 2023-09-14
qlinearadd operator

lakhinderwalia:


pfultz2:
Do we need a new operator for this? I thought this could be implemented as quantizelinear+add.

lakhinderwalia:
> Do we need a new operator for this? I thought this could be implemented as quantizelinear+add.

Are you suggesting that parse_quantizelinear.cpp should internally invoke quantizelinear + add instead of the new operator..?  That may be one approach.. but I got the impression that a new operator was required... Thanks.

pfultz2:
> Are you suggesting that parse_quantizelinear.cpp should internally invoke quantizelinear + add instead of the new operator..?

Yes but for parse_qlinear_add.cpp(not parse_quantizelinear.cpp).

> but I got the impression that a new operator was required...

The goal is to support the onnx operators but if it can be added using existing operators without too much hassle then we definitely want to do that. See parse_celu.cpp, parse_selu.cpp, parse_softsign.cpp, parse_thresholdrelu.cpp, etc.

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,202 @@
...
+                                    const instruction_ref& x_in,
+                                    const instruction_ref& arg_fscale,
+                                    const instruction_ref& arg_z_pt,
+                                    const onnx_parser::node_info& info) const
+    {
+        auto in_lens = x_in->get_shape().lens();
+
+        // prep 1: broadcast scale. it can come as a scalar or a 1-D tensor.
+        std::vector<float> sc_val;
+        auto ev_arg_fscale = arg_fscale->eval();
```
pfultz2, line 137: There is no reason to require `arg_fscale` to be a literal.

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,202 @@
...
+        return bcast_instr_out;
+    }
+
+    // This method is to prep for quantizelinear or dequantizelinear operation for
+    // either the broadcasting of weight-scale or zero-points of qlinearadd operator
+    // outputs: operator op (inputs x, broadcasted: scale (float) & zero_pt (8-bit))
+    instruction_ref bcast_qdq_instr(const std::string& op_name,
+                                    const instruction_ref& x_in,
+                                    const instruction_ref& arg_fscale,
+                                    const instruction_ref& arg_z_pt,
```
pfultz2, line 130: Pass `instruction_ref` by value.

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,202 @@
...
+                migraphx::make_op("broadcast", {{"axis", 0}, {"out_lens", in_lens}}),
+                info.add_literal(sh_scale, sc_val));
+        else
+            bcast_scale =
+                info.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", in_lens}}),
+                                     info.add_literal(sh_scale, sc_val));
+
+        // prep 2: broadcast zero point. it can come as a scalar or a 1-D tensor.
+        std::vector<int> z_pt_val;
+        auto ev_arg_z_pt = arg_z_pt->eval();
```
pfultz2, line 152: Same here. This is only used to check the size, so it would be better to use `elements()` instead.

lakhinderwalia: Already fixed.

src/onnx/parse_qlinearadd.cpp
lakhinderwalia: We could pull the common functionality (amongst quant operators, e.g. Qlinearconv etc.) bcast_qdq_instr  into a separate file.. Rather than have 4 copies of it..

pfultz2: Yes thats a good idea. You can add it under `include/migraphx/onnx/`.

pfultz2: It also would be better to rename the function to `insert_linear_instructions`.

pfultz2: Another idea, would be to use the same parser class for all these operators. The parser class can list multiple operators:

```cpp
std::vector<op_desc> operators() const 
{ 
    return {{"QLinearAdd", "add"},
            {"QLinearConv", "convolution"}}; 
}
```

The `op_desc.onnx_name` will be the first name(ie `QLinearAdd`) and `op_desc.op_name` will be the second (ie `add`). You can see parse_reduce_op.cpp as an example:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/onnx/parse_reduce_op.cpp#L93

I dont know if that will be easier since it looks this might share more than just one function.

pfultz2:
There should also be a test added to `onnx_test.cpp`.

src/onnx/parse_qlinearadd.cpp
lakhinderwalia: I have looked at parse_pooling.cpp as an example. Not too much commonality amongst these quant operators except the need to quantize/dequantize. They are all of different classes, with qdq as the commonality..

lakhinderwalia:
> There should also be a test added to `onnx_test.cpp`.

Yes, the parser tests will be added later,  I don't want to chase manual graph creation (to verify parsing) until parsing is all nailed down. Thanks.

(There is already a onnx verification test that is in place)

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,193 @@
...
+        if(arg_fscale->get_shape().elements() > 1)
+            bcast_scale = info.add_instruction(
+                migraphx::make_op("broadcast", {{"axis", 0}, {"out_lens", in_lens}}), arg_fscale);
+        else
+            bcast_scale = info.add_instruction(
+                migraphx::make_op("multibroadcast", {{"out_lens", in_lens}}), arg_fscale);
+
+        // prep 2: broadcast zero point. it can come as a scalar or a 1-D tensor.
+        instruction_ref bcast_zero_pt;
+        if(arg_z_pt->get_shape().elements() > 1)
```
umangyadav, line 146: Does this work `get_shape().scalar()`  instead ?

test/onnx/verify_onnx.cpp
```diff
@@ -1245,6 +1245,42 @@ TEST_CASE(nonzero_test)
...
 }
 
+TEST_CASE(qlinearadd_test)
+{
+    // github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.QLinearAdd
+    migraphx::program p = migraphx::parse_onnx("qlinearadd_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape a{migraphx::shape::uint8_type, {64}};
+    std::vector<unsigned char> data_a = {
```
umangyadav, line 1255: `uint8_t`  should be available to use i think.

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+struct parse_qlinearadd : op_parser<parse_qlinearadd>
+{
+    std::vector<op_desc> operators() const { return {{"QLinearAdd"}}; }
+
+    // basic type checking for QLinearAdd Operator
+    void check_inputs(const std::vector<instruction_ref>& args) const
+    {
+        if(args.size() < 7)
+            MIGRAPHX_THROW("QLINEARADD: missing inputs");
```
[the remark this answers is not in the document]
lakhinderwalia: (The operator specifies, it has to be a total of 7 or 8).

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+        const auto& in_a = args[0];
+        const auto& in_b = args[3];
```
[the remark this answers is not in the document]
lakhinderwalia: Yes, that is very confusing. I don't see how the arguments can be optional (except the last one). And that is why the minimum number specified is 7.

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,193 @@
...
+        auto sh_b = in_b->get_shape();
+        if(sh_a != sh_b)
+            MIGRAPHX_THROW("QLINEARADD: mismatched input shapes");
+
+        auto type_a = sh_a.type();
+        auto type_b = sh_b.type();
+        if(type_a != migraphx::shape::int8_type and type_a != migraphx::shape::uint8_type)
+            MIGRAPHX_THROW("QLINEARADD: unsupported input type");
+        if(type_b != migraphx::shape::int8_type and type_b != migraphx::shape::uint8_type)
+            MIGRAPHX_THROW("QLINEARADD: unsupported input type");
```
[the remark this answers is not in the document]
lakhinderwalia: There is no reason why the two inputs can't have different types: one a uint8 other an int8. Hence the code to check both types.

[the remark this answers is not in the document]
umangyadav: ONNX says A and B both are of type T and 
Result C is of the same type as both inputs. (that implies they must be the same). 

You can try generating onnx with both inputs of different types. If it works then we can add a test for that case.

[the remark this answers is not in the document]
lakhinderwalia: You are right. A and B are both the same type -- and also C. Thanks.

test/onnx/verify_onnx.cpp
lakhinderwalia: Sure. (same thing underneath.., but will change it). Thanks.

src/onnx/parse_qlinearadd.cpp
lakhinderwalia: The current style one is lot clearer to me, and therefore sticking to that one. 

scalar() definition: in its include file is referring to its strides of 0  -- and I am not sure of its functional relation to elements(), even though its implementation might mean the equivalence. Thanks.

src/onnx/parse_qlinearadd.cpp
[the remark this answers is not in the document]
lakhinderwalia: The code is rearranged a bit, and hopefully with a clearer error message as well. Thanks.

src/onnx/parse_qlinearadd.cpp
lakhinderwalia: Moved to another file, in the subdir you suggested. Thanks.

lakhinderwalia:
> There should also be a test added to `onnx_test.cpp`.

Done: qlinearadd_test

test/onnx/verify_onnx.cpp
lakhinderwalia: Done.

src/onnx/parse_qlinearadd.cpp
```diff
@@ -0,0 +1,153 @@
...
+                          const onnx_parser::node_info& info,
+                          const std::vector<instruction_ref>& args) const
+    {
+        check_inputs(args);
+
+        // A
+        const auto& in_a         = args[0];
+        const auto& in_scale_a   = args[1];
+        const auto& in_zero_pt_a = args[2];
+        auto dquant_a = bcast_qdq_instr("dequantizelinear", in_a, in_scale_a, in_zero_pt_a, info);
```
kahmed10, line 128:
```suggestion
        auto dquant_a = info.add_common_op("dequantizelinear", {in_a, in_scale_a, in_zero_pt_a}, false);
```
This requires some modifications of `add_common_op` under `onnx_parser.cpp/hpp, common.cpp/hpp`, but this would encourage reuse of existing functions that aim to do the same thing. I've tested by making the convert optional in `add_common_op` by adding a flag to the end of the function (on by default to keep other functional calls intact). Seems like it passes both tests when trying this approach.

@pfultz2 @umangyadav feel free to chime in if this approach is cleaner or not.

lakhinderwalia: Thanks. Please note, this new api uses `Broadcast` and `Multibroadcast`. The other one doesn't use `Broadcast`. I am not sure if we want to make it a kitchen sink there..; 

There are other Quant operators that I am adding which shall use Broadcast, and so that part isn't tested by your new patch.

test/onnx/verify_onnx.cpp
umangyadav: I don't see it changed.

src/onnx/parse_qlinearadd.cpp
pfultz2: Its not equivalent, so its better to use `get_shape().elements()`.

src/onnx/parse_qlinearadd.cpp
pfultz2: I dont think `add_common_op` will work because we compare from the right. In this case we would need to compare from the left, and that only works because it will broadcast on axis 0. For axis 1 this wouldn't work. We could extend the `add_common_op` so an optional axis could be passed to handle 1d tensors. 

Either way we may want to address this another PT.

test/onnx/verify_onnx.cpp
lakhinderwalia: Sorry, it was only partially changed.. Will fix it. Thanks.

lakhinderwalia: Done.

src/onnx/parse_qlinearadd.cpp
```diff
@@ -133,6 +134,21 @@ struct parse_qlinearadd : op_parser<parse_qlinearadd>
...
+        {
+            auto common_lens = compute_broadcasted_lens(sh_a.lens(), sh_b.lens());
+            if(sh_a.lens() != common_lens)
+                dquant_a = info.add_instruction(
+                    migraphx::make_op("multibroadcast", {{"out_lens", common_lens}}), dquant_a);
+
+            if(sh_b.lens() != common_lens)
+                dquant_b = info.add_instruction(
+                    migraphx::make_op("multibroadcast", {{"out_lens", common_lens}}), dquant_b);
+        }
```
umangyadav, line 150: I think here you can use `add_common_op` here

src/onnx/parse_qlinearadd.cpp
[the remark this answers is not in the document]
umangyadav: yeah hot sure how accurate onnx spec is. @CharlieL7  have you seen such case before where optional argument is not the last ?

src/onnx/parse_qlinearadd.cpp
kahmed10: @umangyadav see comment above

lakhinderwalia: Sure. Will do shortly. Anything else, @umangyadav & @kahmed10 ..?   Thanks.

src/onnx/parse_qlinearadd.cpp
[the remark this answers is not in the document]
lakhinderwalia: There is no way to leave optional arguments except at the end of an argument list .. Unless of course named arguments were supported. The minimum limit is 7 for arguments.

src/onnx/parse_qlinearadd.cpp
kahmed10: I think we should address in another PR. Effectively we want to create reusable functions that will automatically handle broadcaasting shapes to be compatible for any operation.

pfultz2: I agree that the common_op updates should be in another PR. We need to discuss how we want to design such an API for these cases.

src/onnx/parse_qlinearadd.cpp
lakhinderwalia: Done.

kahmed10 (approved):

pfultz2 (approved):

pfultz2:
LGTM, but I would like to @umangyadav to finish his review as well.

test/onnx/verify_onnx.cpp
```diff
@@ -1245,6 +1245,79 @@ TEST_CASE(nonzero_test)
...
+                                   76,  74,  72,  70,  68,  66,  64,  62,  60,  58,  56,  54,  52,
+                                   50,  48,  46,  44,  42,  40,  38,  36,  34,  32,  30,  28,  26,
+                                   24,  22,  20,  18,  16,  14,  12,  10,  8,   6,   4,   2};
+
+    migraphx::parameter_map pp;
+    pp["A"]     = migraphx::argument(a, data_a.data());
+    pp["B"]     = migraphx::argument(b, data_b.data());
+    auto result = p.eval(pp).back();
+
+    std::vector<unsigned char> result_vector;
```
umangyadav, line 1273:
```suggestion
    std::vector<uint8_t> result_vector;
```

umangyadav (approved):

test/onnx/verify_onnx.cpp
umangyadav: LGTM just merge this suggestion.

PR #2191, opened 2023-09-15
Enable Virtual Environments in testing

ahsan-ca:
This sets up the virtual environments for running python onnx tests.
ctest's fixtures are used to ensure that the virtual environment is
set up before the tests are run.


ahsan-ca (commented):
@pfultz2 This is a draft version, needs to be cleaned up and improved.
I have asked some questions for clarification/feedback.
Please feel free to leave more comments/feedback that you may have.
Thanks!

tools/install_prereqs.sh
```diff
@@ -51,6 +51,7 @@ else
     openmp-extras \
     python3-dev \
     python3-pip \
+    python3.8-venv \
```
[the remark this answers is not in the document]
causten: sles supports a different level of python (3.7?), Ubuntu 22.04 supports python3.10.  I think you'll need to just install the default python3-venv  and not be explicit to 3.8

test/py/CMakeLists.txt
```diff
@@ -23,6 +23,23 @@
 #####################################################################################
 
 include(PythonModules)
+set(FIX_NAME SETUP_VIRTUAL_ENV)
+set(VENV_DIR ${CMAKE_BINARY_DIR}/test/py/venv)
+
+function(add_py_venv_fixture FIXTURE_NAME VIRTUAL_ENV_DIR REQUIREMENTS_FILE)
+    message("Attempting to create virtual environment: ${VIRTUAL_ENV_DIR}")
+    find_package (Python3 COMPONENTS Interpreter)
```
pfultz2, line 31: This should use the `foreach(PYTHON_VERSION ${PYTHON_VERSIONS})` to add fixtures for all python version and then it needs use `${PYTHON_${PYTHON_VERSION}_EXECUTABLE}` to get the executable. Dont call `find_package(Python3)` as this could get a different python version.

test/py/CMakeLists.txt
```diff
@@ -31,10 +48,15 @@ function(add_py_test NAME SCRIPT)
             "PYTHONMALLOC=debug"
             "MALLOC_CHECK_=3"
         )
-        set(PYTHON_EXECUTABLE ${PYTHON_${PYTHON_VERSION}_EXECUTABLE})
+        set (ENV{VIRTUAL_ENV} ${VENV_DIR})
+        set (Python3_FIND_VIRTUALENV ONLY)
+        unset (Python3_EXECUTABLE)
+        find_package (Python3 COMPONENTS Interpreter Development)
```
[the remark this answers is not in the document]
pfultz2: Dont use `find_package` to find python. Instead do `set(PYTHON_EXECUTABLE ${VENV_DIR}/${PYTHON_VERSION}/bin/python)`.

test/py/CMakeLists.txt
```diff
@@ -23,6 +23,23 @@
 #####################################################################################
 
 include(PythonModules)
+set(FIX_NAME SETUP_VIRTUAL_ENV)
+set(VENV_DIR ${CMAKE_BINARY_DIR}/test/py/venv)
+
+function(add_py_venv_fixture FIXTURE_NAME VIRTUAL_ENV_DIR REQUIREMENTS_FILE)
+    message("Attempting to create virtual environment: ${VIRTUAL_ENV_DIR}")
+    find_package (Python3 COMPONENTS Interpreter)
+    add_test (NAME INITIALIZE_VENV COMMAND "${Python3_EXECUTABLE}" -m venv ${VIRTUAL_ENV_DIR})
```
pfultz2, line 32: This should be `${VIRTUAL_ENV_DIR}/${PYTHON_VERSION}` so there is no conflict with multiple python versions.

pfultz2: Also, `${PYTHON_${PYTHON_VERSION}_EXECUTABLE}` should be used for the executable.

test/py/CMakeLists.txt
```diff
@@ -23,6 +23,23 @@
...
+    set_tests_properties(INITIALIZE_VENV PROPERTIES FIXTURES_SETUP INIT_VENV)
+
+    set (ENV{VIRTUAL_ENV} ${VIRTUAL_ENV_DIR})
+    set (Python3_FIND_VIRTUALENV ONLY)
+    unset (Python3_EXECUTABLE)
+    find_package (Python3 COMPONENTS Interpreter Development)
+    add_test (NAME SETUP_VENV COMMAND "${Python3_EXECUTABLE}" -m pip install -r ${REQUIREMENTS_FILE})
```
pfultz2, line 39: This should use `${VIRTUAL_ENV_DIR}/${PYTHON_VERSION}/bin/python`.

test/py/CMakeLists.txt
```diff
@@ -23,6 +23,23 @@
...
+    set_tests_properties(SETUP_VENV PROPERTIES FIXTURES_REQUIRED INIT_VENV)
+    set_tests_properties(SETUP_VENV PROPERTIES FIXTURES_SETUP ${FIXTURE_NAME})
+endfunction()
 
 function(add_py_test NAME SCRIPT)
```
pfultz2, line 44: This should take `VENV` as input parameter.

test/py/CMakeLists.txt
pfultz2, line 32: The name of the test should be `${FIXTURE_NAME}_INITIALIZE_VENV` to avoid conflicts with different python versions or fixtures.

test/py/CMakeLists.txt
```diff
@@ -23,6 +23,23 @@
...
 
 include(PythonModules)
+set(FIX_NAME SETUP_VIRTUAL_ENV)
+set(VENV_DIR ${CMAKE_BINARY_DIR}/test/py/venv)
+
+function(add_py_venv_fixture FIXTURE_NAME VIRTUAL_ENV_DIR REQUIREMENTS_FILE)
+    message("Attempting to create virtual environment: ${VIRTUAL_ENV_DIR}")
+    find_package (Python3 COMPONENTS Interpreter)
+    add_test (NAME INITIALIZE_VENV COMMAND "${Python3_EXECUTABLE}" -m venv ${VIRTUAL_ENV_DIR})
+    set_tests_properties(INITIALIZE_VENV PROPERTIES FIXTURES_SETUP INIT_VENV)
```
pfultz2, line 33: The fixture setup name should `${FIXTURE_NAME}_INIT_VENV`.

test/py/CMakeLists.txt
```diff
@@ -44,6 +66,9 @@ endfunction()
 add_dependencies(tests migraphx_py)
 add_dependencies(check migraphx_py)
 
+add_py_venv_fixture(
```
pfultz2, line 69: This should be called inside `add_py_test`.

test/py/CMakeLists.txt
```diff
@@ -44,6 +66,9 @@ endfunction()
...
+    ${FIX_NAME} ${VENV_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/REQUIREMENTS.TXT)
```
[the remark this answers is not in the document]
ahsan-ca: Done. I have also added a `requirements-sles.txt` for package requirements on sles.

test/py/CMakeLists.txt
```diff
@@ -23,37 +23,71 @@
...
             "PYTHONPATH=$<TARGET_FILE_DIR:migraphx_pybind_${PYTHON_VERSION}>"
             "PYTHONMALLOC=debug"
             "MALLOC_CHECK_=3"
         )
         set(PYTHON_EXECUTABLE ${PYTHON_${PYTHON_VERSION}_EXECUTABLE})
+        set(VIRTUAL_ENV_DIR ${VENV_DIR}/${PYTHON_VERSION})
+
+        add_test(NAME py_${PYTHON_VERSION}_initialize_env COMMAND ${PYTHON_EXECUTABLE} -m venv ${VIRTUAL_ENV_DIR} --clear)
+        set_tests_properties(py_${PYTHON_VERSION}_initialize_env PROPERTIES FIXTURES_SETUP ${PYTHON_VERSION}_INIT_VENV)
+        set(ENV{VIRTUAL_ENV} ${VIRTUAL_ENV_DIR})
```
ahsan-ca, line 51: Since we are using the path to python in the virtual environment, this might not be needed. Should I remove this?

test/py/CMakeLists.txt
ahsan-ca: Discussed this on Friday with Paul, and he suggested to check if a test already exists before adding tests for init/setup venv to resolve the issue with duplication of test names. I have now made that change and `add_py_venv_fixture` is being called inside `add_py_test`.

TedThemistokleous (approved):
LGTM. Unless @pfultz2 has anything else

Perf errors are from another issue I think we've already resolved on develop.

ahsan-ca:
@pfultz2 I have made changes to the requirement files for version specific python requirement files. Can you please take a look to see if this looks good.
I have not disabled backend tests for 3.6 as they can be run with an older onnx version provided we have the models downloaded (which I believe is how QA does testing).

test/py/base-requirements.txt
```diff
@@ -0,0 +1,25 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
+
+pytest==6.0.1
```
pfultz2, line 25: I dont think we need pytest, but we do need numpy for our tests.

test/py/requirements-python-3.6.txt
```diff
@@ -0,0 +1,28 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
+
+onnx==1.10.2
+protobuf==3.19.6
+numpy==1.19.5
+packaging==21.3
```
pfultz2, line 28: Remove this file, we should just disable the onnx tests when using python 3.6.

test/py/requirements-python-all.txt
```diff
@@ -0,0 +1,28 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
+
+onnx==1.14.1
+protobuf==3.20.2
+numpy==1.21.6
+packaging==23.0
```
pfultz2, line 28: This should be named requirements-onnx.txt.

test/py/CMakeLists.txt
pfultz2: This env variable wont be applied when running the test, so this can be removed.

test/py/CMakeLists.txt
```diff
@@ -23,37 +23,62 @@
...
 if(MIGRAPHX_ENABLE_GPU)
-add_py_test(gpu_offload test_gpu_offload.py WORKING_DIRECTORY ${TEST_ONNX_DIR})
-add_py_test(gpu test_gpu.py WORKING_DIRECTORY ${TEST_ONNX_DIR})
-add_py_test(array test_array.py WORKING_DIRECTORY ${TEST_ONNX_DIR})
-add_py_test(backend onnx_backend_test.py WORKING_DIRECTORY ${TEST_ONNX_DIR})
-add_py_test(gpu_async test_gpu_async.py WORKING_DIRECTORY ${TEST_ONNX_DIR})
+add_py_test(gpu_offload test_gpu_offload.py ${VENV} WORKING_DIRECTORY ${TEST_ONNX_DIR})
+add_py_test(gpu test_gpu.py ${VENV} WORKING_DIRECTORY ${TEST_ONNX_DIR})
+add_py_test(array test_array.py ${VENV} WORKING_DIRECTORY ${TEST_ONNX_DIR})
+add_py_test(backend onnx_backend_test.py ${VENV} WORKING_DIRECTORY ${TEST_ONNX_DIR})
```
pfultz2, line 82: Onnx backend should be using a different virtual environment, and this should be disabled on python 3.6.

test/py/CMakeLists.txt
```diff
@@ -23,37 +23,62 @@
...
         set (ENV_COMMAND ${CMAKE_COMMAND} -E env 
             "PYTHONPATH=$<TARGET_FILE_DIR:migraphx_pybind_${PYTHON_VERSION}>"
             "PYTHONMALLOC=debug"
             "MALLOC_CHECK_=3"
         )
         set(PYTHON_EXECUTABLE ${PYTHON_${PYTHON_VERSION}_EXECUTABLE})
+        if(${PYTHON_VERSION} STREQUAL "3.6")
+            add_py_venv_fixture(${PYTHON_VERSION} ${PYTHON_EXECUTABLE} ${VENV_DIR}/${PYTHON_VERSION} ${REQUIREMENTS_PYTHON_36})
+        else()
+            add_py_venv_fixture(${PYTHON_VERSION} ${PYTHON_EXECUTABLE} ${VENV_DIR}/${PYTHON_VERSION} ${REQUIREMENTS_PYTHON_ALL})
```
pfultz2, line 56: The `add_py_env_fixture` probably needs to be called outside of the function since tests will use different virtual environments.

test/py/CMakeLists.txt
```diff
@@ -23,37 +23,62 @@
...
+
+function(add_py_test NAME SCRIPT VENV_DIR)
     foreach(PYTHON_VERSION ${PYTHON_VERSIONS})
         set (ENV_COMMAND ${CMAKE_COMMAND} -E env 
             "PYTHONPATH=$<TARGET_FILE_DIR:migraphx_pybind_${PYTHON_VERSION}>"
             "PYTHONMALLOC=debug"
             "MALLOC_CHECK_=3"
         )
         set(PYTHON_EXECUTABLE ${PYTHON_${PYTHON_VERSION}_EXECUTABLE})
+        if(${PYTHON_VERSION} STREQUAL "3.6")
```
pfultz2, line 53: Remove this, we shouldn't have python specific versions of the requirements.

test/py/base-requirements.txt
ahsan-ca: I have removed `pytest`, it was needed for onnx.
Tests pass without `numpy`. So I have kept this file empty, I can add `numpy` if you feel it should be here.

test/py/CMakeLists.txt
ahsan-ca: I have disabled the onnx backend tests on python 3.6. I am open to a more elegant solution if you think we can have an alternate way to disable this.

test/py/CMakeLists.txt
ahsan-ca: I have taken the function out of `add_py_test`.

ahsan-ca:
@pfultz2 Thanks for taking a look and providing feedback. I have tried to address the comments. Can you please take a look to see if I have addressed the comments correctly? I am not a 100% sure if the current implementation is how you'd like it.

Side note: The backend test seems to be timing out, I am investigating that but thought to put this up in the meanwhile to get your feedback on the implementation strategy.
Thanks!

ahsan-ca:
> Side note: The backend test seems to be timing out, I am investigating that but thought to put this up in the meanwhile to get your feedback on the implementation approach. Thanks!

The timing out of backend test was system related. A reboot of the system helped fix the issue.

test/py/requirements.txt
```diff
@@ -0,0 +1,24 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
+
```
pfultz2, line 24: Why is this empty? Dont we need numpy installed?

ahsan-ca: The output for pip list:
```
test/py/venv/3.8/bin/pip list
Package       Version
------------- -------
pip           20.0.2 
pkg-resources 0.0.0  
setuptools    44.0.0
```
I also uninstalled `numpy` on the docker container I was testing things on, even though tests run on a virtual environment.
I will also setup a docker container by removing the packages that are installed in `install_prereqs` and do a run with that.

causten: if the tests are passing, then I guess you don't need numpy

pfultz2: We do need numpy for the tests. They dont fail because we silently skip the test if numpy is not found: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/py/test_gpu.py#L27

We should probably remove this try/except clause and then make sure we have the numpy dependencies.

ahsan-ca: > We do need numpy for the tests. They dont fail because we silently skip the test if numpy is not found
Thanks for catching this Paul, I had not realized that this was the case. I have added numpy in the requirements file.

> We should probably remove this try/except clause and then make sure we have the numpy dependencies.
I can remove the try/except clause in a separate PR. I am open to doing it in this one if you think we should make the change here.

ahsan-ca: > I can remove the try/except clause in a separate PR. I am open to doing it in this one if you think we should make the change here.

Here is the PR for removing the try/except clause: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2316
@pfultz2

ahsan-ca:
Looks like we have also have try/except clause in another test `test_numpy.py` as well: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a50cb30242d9b328a05f17e34a225310d1210134/test/py/test_numpy.py#L26

Should we also remove this try except clause or leave it as is? @pfultz2

pfultz2:
That should be removed as well.

ahsan-ca:
> That should be removed as well.

Thanks! Removed through https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2316.

ahsan-ca:
Updated with develop branch to pull fix of cppcheck issue that has been fixed by df5f8c969316b148be4daa854856e9d9bbb28061.

test/py/requirements.txt
pfultz2: So that has been merged, so can the requirements.txt be updated to list numpy?

pfultz2 (approved):
This looks good, but as a followup we should add a flag to disable the virtual environments so we arent installing these packages everytime we are running tests in the CI.

test/py/requirements.txt
ahsan-ca: So that has been merged, so can the requirements.txt be updated to list numpy?

> Yes, that is done.

ahsan-ca:
> add a flag to disable the virtual environments so we arent installing these packages everytime we are running tests in the CI.

@pfultz2 Thanks for approving the PR and for all your suggestions/feedback during the course of it. I have opened an issue to add a flag to disable virtual environments: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2324

PR #2195, opened 2023-09-15
Optimize broadcast+transpose for scalars

kahmed10:
+minor typo changes


src/simplify_reshapes.cpp
```diff
@@ -627,6 +627,28 @@ struct find_transpose_contiguous_reshaper_unary
     }
 };
 
+struct find_broadcast_transpose
+{
+    auto matcher() const
+    {
+        return match::name("transpose")(match::arg(0)(match::name("multibroadcast").bind("bcast_ins")));
```
umangyadav, line 634: multibroadcast instruction should only have once output. so need to add `used_once()`. 
Matcher is replacing that broadcast instruction with other.  Add test for that case as well.

umangyadav (approved):

umangyadav (changes requested):
requires minor changes.

src/simplify_reshapes.cpp
kahmed10: added

src/optimize_module.cpp
```diff
@@ -36,8 +36,12 @@ void optimize_module::apply(module_pass_manager& mpm) const
 {
     for(int i = 0; i < 2; i++)
     {
-        mpm.run_pass(simplify_reshapes{});
-        mpm.run_pass(simplify_algebra{});
+        // loop to further optimize after initial transformations
```
umangyadav, line 39: can't we just increase the outer loop trip count ?

kahmed10: no because `propagate_constant` will otherwise compute the unoptimized shape for the literal.

umangyadav (approved):

src/simplify_reshapes.cpp
```diff
@@ -627,6 +627,30 @@ struct find_transpose_contiguous_reshaper_unary
     }
 };
 
+struct find_broadcast_transpose
+{
+    auto matcher() const
+    {
+        return match::name("transpose")(
+            match::arg(0)(match::name("multibroadcast")(match::used_once().bind("bcast_ins"))));
```
pfultz2, line 635: Dont use `used_once`. `multibroadcast` is an alias operator so its cheap to run, so it ok if is done multiple times.

test/simplify_reshapes_test.cpp
```diff
@@ -67,6 +67,43 @@ migraphx::module make_concat_multibroadcast(const std::vector<size_t>& in_lens,
...
+    auto t1 = mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {1, 0}}}), mb);
+    mm->add_return({t1});
+
+    run_pass(*mm);
+    migraphx::program p2;
+    mm = p2.get_main_module();
+
+    l  = mm->add_parameter("x", {migraphx::shape::float_type, {1}, {0}});
+    mb = mm->add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", {3, 2}}}), l);
+    mm->add_return({mb});
```
pfultz2, line 86: Use anonymous scopes instead of reassigning the variables. You can look at test cases like `transpose_contiguous_unsqueeze_unary` for an example.

test/simplify_reshapes_test.cpp
```diff
@@ -67,6 +67,43 @@ migraphx::module make_concat_multibroadcast(const std::vector<size_t>& in_lens,
     return m;
 }
 
+TEST_CASE(broadcast_transpose_scalar)
+{
+    migraphx::program p;
```
pfultz2, line 72: Dont declare a program. Just declare the module: `module m1;`.

test/simplify_reshapes_test.cpp
```diff
@@ -67,6 +67,43 @@ migraphx::module make_concat_multibroadcast(const std::vector<size_t>& in_lens,
...
+
+    auto l  = mm->add_parameter("x", {migraphx::shape::float_type, {1}, {0}});
+    auto mb = mm->add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", {2, 3}}}), l);
+    auto t1 = mm->add_instruction(migraphx::make_op("transpose", {{"permutation", {1, 0}}}), mb);
+    auto id = mm->add_instruction(migraphx::make_op("identity"), mb);
+    mm->add_return({id, t1});
+
+    migraphx::program p2 = p;
+    run_pass(*mm);
+    EXPECT(p == p2);
```
pfultz2, line 104: This should not be the same program. After dead_code_elimination, there should the original multibroadcast->transpose and another mutlibroadcast.

src/optimize_module.cpp
pfultz2: A better solution would be to update constant propagation to not fold when the inputs are broadcasted across a common axis, but that will be more work than just simply rerunning the passes. So we can save that for another PR.

src/simplify_reshapes.cpp
umangyadav: Oh i misunderstood. I thought i was replacing multibroadcast. It is replacing transpose. So no need to use `used_once()`

pfultz2 (approved):

src/optimize_module.cpp
kahmed10: I've added a basic `optimize_module` test to combine some of the transformations. I don't know the full steps in `simplify_algebra` that led to the broadcast+transpose scalar in the bert model, and it might take me more time to figure out. Maybe I can add that later in a separate PR?

test/optimize_module_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        auto l1  = m2.add_parameter("x", {migraphx::shape::float_type, {1}, {0}});
+        auto l2  = m2.add_parameter("y", {migraphx::shape::float_type, {1}, {0}});
+        auto mul = m2.add_instruction(migraphx::make_op("mul"), l2, l1);
+        auto mb =
+            m2.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", {2, 3, 2}}}), mul);
+        m2.add_return({mb});
+    }
+    EXPECT(m1 == m2);
+}
```
pfultz2, line 63: Will this reproduce the issue? If you make simplify_algebra and simplify_reshapes called once, will this test fail?

kahmed10: It won't reproduce the exact issue just yet. But it will make sure that after the transformation in `simplify_reshapes` is applied, `find_inner_broadcast` will simplify to the intended order of mul+broadcast.

pfultz2: I think doing a test as `dot(x, transpose(a)) * b` where `a` is broadcasted over the batch, `b` is a scalar, and `x` is a parameter input. I think this might reproduce the issue.

PR #2206, opened 2023-09-19
Improvements to verify --reduce

pfultz2:
* Skip verifying literals and parameters
* Print the correct trim value
* Print failure for an exception and continue reduce
* Print if shapes do not match


src/driver/verify.cpp
```diff
@@ -160,8 +179,14 @@ void verify_reduced_program(const program& p,
     const auto* mm = p.get_main_module();
     auto n         = std::distance(mm->begin(), mm->end());
     std::cout << "Verify steps: " << n << std::endl;
-    for(std::size_t i = 0; i < n; i++)
+    for(std::size_t i = 1; i < n; i++)
     {
+        auto last = std::prev(mm->end(), i + 1);
```
umangyadav, line 184: Why is it skipping two instructions to start with ? 
One i assume is "@return"

pfultz2: The `end()` is one past the end, so to get the last instruction(without trimming) would be `std::prev(mm->end(), 1)`, and then we add the trimming which make is `i + 1`.

umangyadav (approved):

PR #2210, opened 2023-09-20
ScatterND unsupported reduction modes, and duplicate index support

music-dino:
- Add min and max reduction modes
- Implement duplicate index handling
- Add corresponding reference, onnx, and verify tests.

- Resolves migraphx-benchmark/AMDMIGraphX#138


src/include/migraphx/op/scatternd_op.hpp
```diff
@@ -121,7 +121,8 @@ struct scatternd_op : op_name<Derived>
                 auto k             = indices_shape.lens().back();
                 auto q             = indices_shape.ndim();
                 auto r             = dyn_out.computed_shape.ndim();
-                par_for(updates_shape.elements(), [&](const auto i) {
+                for(auto i = 0u; i < updates_shape.elements(); ++i)
```
turneram, line 124: What is the reason for removing parallelism here?

music-dino: To produce correct results in case of duplicate indices. A more complex solution could have been implemented, but I don't think the reference implementation warrants it

turneram (approved):

Issue #2156, opened 2023-09-06
When run verification of resnet50 model with fp16 quantization option, the verification fails


stefankoncarevic:
## Description

After quantization fp16 for ResNet50.onnx model, I got the error about mismatch:

./bin/migraphx-driver verify --gpu --onnx /MIGraphXDeps/resnet50-v1-7.onnx --fp16

...
FAILED: /MIGraphXDeps/resnet50-v1-7.onnx
error: 0.00123406
Max diff: 0.0360341
Mismatch at 0: -0.624751 != -0.62793

[fp16-ResNet50-migraphx.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/12537544/fp16-ResNet50-migraphx.txt)

I disabled mlir before that.
This blocks adding our CI tracking fp16 and int8 language onnx (quantization) models.

stefankoncarevic:
@jerryyin

jerryyin:
Were you able to run into this issue when using onnx runtime to verify as well?

In today's meeting discussion, we believe this is the issue in particular related with migraphx's CPU verifier. It'd be desirable if you can get the additional data point on onnx runtime verification.

umangyadav:
> It'd be desirable if you can get the additional data point on onnx runtime verification. 

yes i've seen such failures on resnet50 FP16 verification. @stefankoncarevic  if you can run with real data it should match up with accuracy required.

jerryyin:
@umangyadav Regardless of what data @stefankoncarevic used to run, the driver verify shouldn't fail. If the run pass with real data, then it is still a bug on migraphx  driver because it didn't report the correct result.

The point is that there is patch work needed such that migraphx driver verification can use fp16 threshold even if the model produced a fp32 result.

umangyadav:
Agree that `migraphx-driver` requires some change in logic for threshold. I wanted to say that even though `migraphx-driver` shows failing accuracy, it shouldn't affect accuracy in real use cases and therefore shouldn't be a blocker.

jerryyin:
>  it shouldn't affect accuracy in real use cases and therefore shouldn't be a blocker.

Can you give us that guarantee? That we should ignore the migraphx driver verify result regardless of what it reports?
 - If yes, from today on, we can disable that checker. Please sync with @stefankoncarevic on a reasonable way to not only test resnet50 but also correctness of all other tier-1 models.
[...]

PR #2213, opened 2023-09-20
Add options to set tolerances inside MIGraphX driver 

umangyadav:
MIGraphX verification by default uses normalized RMS error as the basis for the verification. 

This PR adds some logic to allow migraphx to do ["np.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html)" type of elementwise verification using `atol` and `rtol`. 

This elementwise verification is disabled for now since it requires tuning tolerances for various tests and also may require changing QA's or other projects CI test procedures. 

PR also includes changes to consistently pass "gold" or "expected" results as the second argument for "verify_range()" calls.  

default RMS tolerance inside `driver` is set to 0.001 which IMO is high for FP32 compared to what we had earlier.  Need better defaults

Helps resolving but doesn't close it: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2156


examples/migraphx/migraphx_driver/README.md
```diff
@@ -55,7 +55,9 @@ See below for a comprehensive list of commands and option arguments, as well as
 | --exhaustive-tune                        | Enable exhaustive search to find fastest kernel           |
 | --fp16                                   | Quantize for fp16                                         |
 | --int8                                   | Quantize for int8                                         |
-| --tolerance                              | Tolerance for errors                                      |
+| --rms_tol                                | Tolerance for the RMS error                               |
```
lakhinderwalia, line 58: Are these three new options exclusive? Thanks.

umangyadav: Exclusive you mean if they all are different settings ? then yes.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +188,95 @@ double rms_range(const R1& r1, const R2& r2)
...
 
+template <class R>
+double get_rms_tol(const R&, std::size_t tolerance = 80)
+{
+    double threshold = std::numeric_limits<range_value<R>>::epsilon() * tolerance;
+    return threshold;
+}
+
+/*
+C++ doesn't support named arguments, this is just wrapper that helps distinguish between actual
```
lakhinderwalia, line 199: No reason to guard to distinguish between results and expected arguments:
<Ideally, a design should have symmetry and if two arguments (i.e. results vs. expected) are reversed, chalk it up to programming or user error. Yet, the outcome of that would be still logical -- example: in diffing two files, their order produces different results, but still logical. Maybe some difference here would reverse polarity.>

src/driver/verify.hpp
```diff
@@ -37,18 +38,18 @@ void verify_program(const std::string& name,
                     compile_options options     = compile_options{},
                     precision quantize          = precision::fp32,
                     const parameter_map& inputs = {},
-                    double tolerance            = 100);
+                    verify::tolerance tols      = verify::tolerance{});
```
lakhinderwalia, line 41: Just trivia. The value of tolerance is different in three cases (lines 40,45, 51) . Yet they all have the same replacement default: verify::tolerance{}.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +188,95 @@ double rms_range(const R1& r1, const R2& r2)
...
+struct tolerance
+{
+    double rms_tol = 0.001;
+    double atol    = 0.001;
+    double rtol    = 0.001;
+};
+
 template <class R1, class R2>
-bool verify_range(const R1& r1, const R2& r2, double tolerance = 80, double* out_error = nullptr)
+bool allclose(const R1& r1, const R2& r2, tolerance tols)
```
lakhinderwalia, line 229: Please provide a one or two liner documentation for uninitiated like me. I don't know what does allclose() mean. Thanks.

src/include/migraphx/verify.hpp
umangyadav: It can't be symmetric for the elementwise verification 
https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

allclose calculates `rtol *b`.  If it is `rtol*a` it can result in different value.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +188,95 @@ double rms_range(const R1& r1, const R2& r2)
...
+    // disable ewise_verify for now, it requires lot of tests to be fixed
+    // auto ewise_verify = allclose(r1, r2.data(), tols);
+    if(out_rms_error != nullptr)
+        *out_rms_error = rms_error;
+    return rms_error <= tols.rms_tol;
+}
+
+// expected argument should be passed as second, but if it is passed as the first by mistake then
+// flip the order
+template <class R1, class R2>
```
lakhinderwalia, line 271: My comments about symmetric design, above. We shouldn't have to resort to this, as a programmer. If the world would collapse, then yes do this..  :-) 
(But the world would not collapse, if you do it..)

src/driver/verify.hpp
umangyadav: yes. It should be fine since the newer default tolerance value is higher compared to what previous value of 100 would have calcualted. 

Default values need to be worked out in a better way.

src/include/migraphx/verify.hpp
umangyadav: It won't be symmetric if allclose is used. Therefore need to flip the order.

src/include/migraphx/verify.hpp
lakhinderwalia: Design symmetry (of an API) doesn't mean the results are identical if inputs order is changed. As I mentioned above in the example about diffing two files. No body should expect identical results due to symmetry in design -- when input order is changed. Thanks.

src/driver/verify.hpp
lakhinderwalia: I hope there is a way to dump the tolerances via CLI. Thanks.

src/include/migraphx/verify.hpp
lakhinderwalia: (I would add this URL here as a comment/reference.) Thanks.

src/include/migraphx/verify.hpp
pfultz2: > if two arguments (i.e. results vs. expected) are reversed, chalk it up to programming or user error.

We want it to do the right thing. Chalking it up to progammer error just leads to subtle bugs in our code, since the flipped arguments are not diagnosed.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +188,95 @@ double rms_range(const R1& r1, const R2& r2)
...
     auto error       = rms_range(r1, r2);
-    if(out_error != nullptr)
-        *out_error = error;
+    if(out_rms_error != nullptr)
+        *out_rms_error = error;
     return error <= threshold;
 }
 
+template <class R1, class R2>
+bool verify_range_with_tolerance(const R1& r1,
```
pfultz2, line 256: I would prefer this function to be called `verify_range`. We could rename the other overload or keep it the same, since we use a struct for tolerance there should be an implicit conversion.

src/include/migraphx/verify_args.hpp
```diff
@@ -31,11 +31,15 @@
...
-                 const argument& target_arg,
-                 double tolerance = 80);
+MIGRAPHX_EXPORT bool verify_args_with_threshold(const std::string& name,
+                                                const argument& target_arg,
+                                                const argument& ref_arg,
+                                                verify::tolerance);
+
+MIGRAPHX_EXPORT bool verify_args(const std::string& name,
+                                 const argument& target_arg,
+                                 const argument& ref_arg,
```
pfultz2, line 41: The ref data should be taken as `expected<argument>`.

src/include/migraphx/verify.hpp
pfultz2: Ah the other overload doesnt check relative difference.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +188,95 @@ double rms_range(const R1& r1, const R2& r2)
...
+        auto idx = mismatch_idx(r1, r2, [&](auto x, auto y) {
+            return abs_diff(double(x), double(y)) > tols.atol + tols.rtol * std::abs(double(y));
+        });
+        return idx >= range_distance(r1);
+    }
+    return false;
+}
+
+template <class R1, class R2>
+bool verify_range(const R1& r1,
```
pfultz2, line 243: Can we rename this as `verify_rms_range`?

src/driver/main.cpp
```diff
@@ -536,13 +536,15 @@ struct params : command<params>
...
     compiler c;
-    double tolerance     = 80;
+    migraphx::verify::tolerance tols;
     bool per_instruction = false;
     bool reduce          = false;
     void parse(argument_parser& ap)
     {
         c.parse(ap);
-        ap(tolerance, {"--tolerance"}, ap.help("Tolerance for errors"));
+        ap(tols.rms_tol, {"--rms_tol"}, ap.help("Tolerance for the RMS error"));
```
pfultz2, line 545: The flag should be `--rms-rol`.

src/include/migraphx/verify.hpp
lakhinderwalia: Accuracy testing is a science. We are mixing it up with proper API design. 
Anyways, I have belabored it enough :-)

src/include/migraphx/verify.hpp
umangyadav: Yes i can. I renamed other function instead because renaming this would have led to larger changes.

lakhinderwalia (approved):
Approving this PR. With a few observations/enhancements for the future:

1. There should be a way to dump the used tolerance values (defaults or otherwise..)

2. Design should in general be simple. Rather than trying to anticipate programmer errors -- the whole code base of C/C++ and Linux & Windows is prone to such programming errors. Anticipating User errors is necessary when destructive. Example: trying to pump Diesel in a Gasoline car. And here we are trying to anticipate subtle corner case of allclose()! 

3. If rms_tol is exclusive from atol & rtol  (or some variation thereof), then please dump a warning.. Or at least say something about it in the help message.

Thanks.

pfultz2:
> Design should in general be simple. Rather than trying to anticipate programmer errors 

We want to make the API as safe as possible, as it will save hours/days trying to debug subtle issues. Leveraging the type system to aid in safety checks is a simple a straightforward way to achieve this.

>  the whole code base of C/C++ and Linux & Windows is prone to such programming errors

Sure, but I dont think mimicking code bases full of bugs and vulnerabilities is a good choice. 

> Anticipating User errors is necessary when destructive. 

I dont see why safety must be limited "destructiveness"(which is not really well defined). Nor do I see this kind of "destructive"-only safety talked about among C++ experts.

> And here we are trying to anticipate subtle corner case of allclose()!

Of course, we dont want to loose lots of time from such subtle corner cases. Especially, since this is used in verification it can interfere with our ability to find or reduce accuracy issues.

src/driver/verify.hpp
umangyadav: Made some changes. `driver verify --help` prints all options with defaults.

src/include/migraphx/verify.hpp
umangyadav: Added some comments.

src/driver/main.cpp
umangyadav: Done.

src/include/migraphx/verify.hpp
umangyadav: Done.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +188,99 @@ double rms_range(const R1& r1, const R2& r2)
...
+
+/*
+C++ doesn't support named arguments, this is just wrapper that helps distinguish between actual
+results v/s expected results arguments.
+*/
+template <class T>
+struct expected
+{
+    expected() = default;
+    expected(const T& input) : x(&input) {}
```
pfultz2, line 206: This constructor should probably be declared `explicit`.

umangyadav: Done.

src/driver/main.cpp
```diff
@@ -536,13 +536,19 @@ struct params : command<params>
...
     {
         c.parse(ap);
-        ap(tolerance, {"--tolerance"}, ap.help("Tolerance for errors"));
+        ap(tols.rms_tol, {"--rms-tol"}, ap.help("Tolerance for the RMS error (Default: 0.001)"));
+        ap(tols.atol,
+           {"--atol"},
+           ap.help("Tolerance for the elementwise absolute error (Default: 0.001)"));
+        ap(tols.rtol,
+           {"--rtol"},
+           ap.help("Tolerance for the elementwise relative error (Default: 0.001)"));
```
pfultz2, line 551: Same here, it should say `Tolerance for the elementwise relative difference`.

src/include/migraphx/verify.hpp
pfultz2: Still this should just be called `verify_range`.

src/include/migraphx/verify_args.hpp
```diff
@@ -31,11 +31,15 @@
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
-MIGRAPHX_EXPORT
-bool verify_args(const std::string& name,
-                 const argument& ref_arg,
-                 const argument& target_arg,
-                 double tolerance = 80);
+MIGRAPHX_EXPORT bool verify_args_with_threshold(const std::string& name,
```
pfultz2, line 34: This should be named `verify_args` and the other overload can be named `verify_args_with_tolerance`.

src/driver/main.cpp
```diff
@@ -536,13 +536,19 @@ struct params : command<params>
...
     bool per_instruction = false;
     bool reduce          = false;
     void parse(argument_parser& ap)
     {
         c.parse(ap);
-        ap(tolerance, {"--tolerance"}, ap.help("Tolerance for errors"));
+        ap(tols.rms_tol, {"--rms-tol"}, ap.help("Tolerance for the RMS error (Default: 0.001)"));
+        ap(tols.atol,
+           {"--atol"},
+           ap.help("Tolerance for the elementwise absolute error (Default: 0.001)"));
```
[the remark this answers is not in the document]
umangyadav: Done.

src/driver/main.cpp
umangyadav: Done.

src/include/migraphx/verify_args.hpp
umangyadav: Done.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +188,99 @@ double rms_range(const R1& r1, const R2& r2)
...
+                                 const expected<R2>& r2,
+                                 tolerance tols        = tolerance{},
+                                 double* out_rms_error = nullptr)
+{
+    auto rms_error = rms_range(r1, r2.data());
+    // disable ewise_verify for now, it requires lot of tests to be fixed
+    // auto ewise_verify = allclose(r1, r2.data(), tols);
+    if(out_rms_error != nullptr)
+        *out_rms_error = rms_error;
+    return rms_error <= tols.rms_tol;
```
umangyadav, line 270: `allclose()` is disabled right now. 

It needs to enabled then, comment out line 267 and 
return value needs to be modified. 
@CharlieL7   @pfultz2

src/driver/main.cpp
```diff
@@ -536,13 +536,19 @@ struct params : command<params>
...
     compiler c;
-    double tolerance     = 80;
+    migraphx::verify::tolerance tols;
     bool per_instruction = false;
     bool reduce          = false;
     void parse(argument_parser& ap)
     {
         c.parse(ap);
-        ap(tolerance, {"--tolerance"}, ap.help("Tolerance for errors"));
+        ap(tols.rms_tol, {"--rms-tol"}, ap.help("Tolerance for the RMS error (Default: 0.001)"));
```
pfultz2, line 545: In the future, we should look into having `argument_parser` automatically print the default value.

src/include/migraphx/verify.hpp
pfultz2: Can we add an env variable to enable this instead?

lakhinderwalia:
> I dont see why safety must be limited "destructiveness"(which is not really well defined). Nor do I see this kind of "destructive"-only safety talked about among C++ experts.

Paul, perhaps you could refer us examples where C++ libraries have been crafted for guarding against switching the order of parameters of a function. This is not even a guideline for MISRA or AUTOSAR.

Or perhaps provide reference to such an article on the Internet.. 

Thanks.

src/include/migraphx/verify.hpp
umangyadav: Added `MIGRAPHX_VERIFY_ENABLE_ALLCLOSE`

lakhinderwalia:
> * Make interfaces easy to use correctly and hard to use incorrectly.

(Thanks, Paul for the links.) 

We have a specific math issue here on two identical items. Perhaps we should be comparing these two items in both orders, and flagging out the differences if any that exceed the tolerance from either case.

pfultz2:
> Perhaps we should be comparing these two items in both orders, and flagging out the differences if any that exceed the tolerance from either case.

Its best to use the magnitude from the golden data to compute relative difference.

lakhinderwalia:
> Its best to use the magnitude from the golden data to compute relative difference.

The example of numpy.py.html is instructive enough to show why two vectors give a 'false' result, and 'true' result with a very slight variation in the values:
```
np.allclose([1e10,1e-7], [1.00001e10,1e-8])
False

np.allclose([1e10,1e-8], [1.00001e10,1e-9])
True
```
Imagine a test-vector few thousand entries long! 

I doubt a tester is creating a golden-data set with some specific criterion for its relative difference -- that is a huge burden on a tester to understand the statistical nature of a vector of input data and be mindful of the subtlety involved in using numpy.allclose(). 

This comparison utility should always hide such subtle issues, as this is an implementation detail; Just compare the two vector both ways and flag an error from either case -- and which anyways is not an issue, except some subtle cases, as numpy.allclose documentation alludes to. 

Thanks.

src/include/migraphx/verify.hpp
```diff
@@ -187,16 +190,103 @@ double rms_range(const R1& r1, const R2& r2)
...
+tolerance using this formula:  abs(a - b) < atol + rtol(abs(b))
+*/
+template <class R1, class R2>
+bool allclose(const R1& r1, const R2& r2, tolerance tols)
+{
+    std::size_t n = range_distance(r1);
+    if(n == range_distance(r2))
+    {
+        auto idx = mismatch_idx(r1, r2, [&](auto x, auto y) {
+            return abs_diff(double(x), double(y)) > tols.atol + tols.rtol * std::abs(double(y));
```
pfultz2, line 241: The sign needs to be flipped: `abs_diff(double(x), double(y)) < tols.atol + tols.rtol * std::abs(double(y))`. The binary predicate is for the elements that match(or are equal).

pfultz2:
> I doubt a tester is creating a golden-data set with some specific criterion for its relative difference

Yea I dont think so either. This is not something they need to be concerned about, they should only be concerned with the correctness of the golden data. 

> This comparison utility should always hide such subtle issues, as this is an implementation detail;

We do hide these details. The user just needs to tell the function which is the expected data. Then we use that to determine the magnitude for the difference. We dont care about the magnitude of the test data since it could not be correct anyways.

lakhinderwalia:
> Yea an in those cases we would be incorrectly flagging test data that is within the acceptable tolerance.

We should avoid giving false positives -- trivially achievable by just doing two comparisons. One or few extra lines of code! Thanks.

pfultz2:
> We should avoid giving false positives -- trivially achievable by just doing two comparisons

I dont see how it would be a false positive.

src/include/migraphx/verify.hpp
umangyadav: Done

CharlieL7 (approved):

TedThemistokleous (approved):

pfultz2 (approved):

PR #2218, opened 2023-09-20
Improvements to GH Actions in our CI

causten:
Build and Publish publicly accessible images.  This will allow PRs from forks to work.
Improve docker building speed by using layers already built (see `--cache-from`)


.github/workflows/ci.yaml
```diff
@@ -75,11 +76,17 @@ jobs:
 
     - name: Build and publish 
       env:            
-        DOCKERIMAGE: ${{ needs.check_image.outputs.imagetag }}
+        DOCKER_TAG_UBUNTU: ${{ needs.check_image.outputs.imagetag }}
       run: |
+        # The TOKEN and USERID are github secrets, Action failures at this step
+        # can come from a PR from a fork changing a file wichi forces a rebuild
```
umangyadav, line 82:
```suggestion
        # can come from a PR from a fork changing a file which forces a rebuild
```

.github/workflows/ci.yaml
```diff
@@ -75,11 +76,17 @@ jobs:
...
+        # Resolve by making an internal PR of the Forked PR
         echo $DOCKER_TOKEN | docker login -u $DOCKER_USER --password-stdin
-        docker build . --file hip-clang.docker --tag $DOCKERIMAGE;
-        docker push $DOCKERIMAGE;
+
+        docker pull $DOCKER_IMAGE_UBUNTU:latest || true
+        docker build . --file hip-clang.docker --cache-from $DOCKER_IMAGE_UBUNTU:latest --tag $DOCKER_IMAGE_UBUNTU:$DOCKER_TAG_UBUNTU --tag $DOCKER_IMAGE_UBUNTU:latest;
```
umangyadav, line 87: Why two tags ?

.github/workflows/ci.yaml
```diff
@@ -75,11 +76,17 @@ jobs:
...
+        DOCKER_TAG_UBUNTU: ${{ needs.check_image.outputs.imagetag }}
       run: |
+        # The TOKEN and USERID are github secrets, Action failures at this step
+        # can come from a PR from a fork changing a file wichi forces a rebuild
+        # Resolve by making an internal PR of the Forked PR
         echo $DOCKER_TOKEN | docker login -u $DOCKER_USER --password-stdin
-        docker build . --file hip-clang.docker --tag $DOCKERIMAGE;
-        docker push $DOCKERIMAGE;
+
+        docker pull $DOCKER_IMAGE_UBUNTU:latest || true
```
umangyadav, line 86: what `|| true` does here ? i haven't seen such thing before.

.github/workflows/ci.yaml
```diff
@@ -90,18 +97,24 @@ jobs:
     - uses: actions/checkout@v3
     - name: Build and publish SLES
       env:            
-        DOCKERIMAGE_SLES: ${{ needs.check_image.outputs.imagetag_sles }}
+        DOCKER_TAG_SLES: ${{ needs.check_image.outputs.imagetag_sles }}
       run: |
+        # The TOKEN and USERID are github secrets, Action failures at this step
+        # can come from a PR from a fork changing a file wichi forces a rebuild
```
umangyadav, line 103:
```suggestion
        # can come from a PR from a fork changing a file which forces a rebuild
```

.github/workflows/ci.yaml
causten: That forces the return code to be success, there may not be a :latest tag and that would have caused an error.  It's ok if the :latest doesn't exist so just continue

.github/workflows/ci.yaml
causten: It builds the image with a hash and with a known tag named "latest".  That is so later builds can use the --cache-from xxx:latest to use existing layers.  I wouldn't know the name of an older calculated hash, so just creating a well known tag label

.github/workflows/ci.yaml
causten: Fixed

.github/workflows/ci.yaml
umangyadav: Looks like it pushes both tags 
docker push $DOCKER_IMAGE_UBUNTU:$DOCKER_TAG_UBUNTU;
docker push $DOCKER_IMAGE_UBUNTU:latest;

.github/workflows/ci.yaml
umangyadav: I think i get it.  It is helpful for the case when hash changes.

umangyadav (approved):

.github/workflows/ci.yaml
```diff
@@ -75,11 +76,17 @@ jobs:
...
+        DOCKER_TAG_UBUNTU: ${{ needs.check_image.outputs.imagetag }}
       run: |
+        # The TOKEN and USERID are github secrets, Action failures at this step
+        # can come from a PR from a fork changing a file which forces a rebuild
+        # Resolve by making an internal PR of the Forked PR
         echo $DOCKER_TOKEN | docker login -u $DOCKER_USER --password-stdin
-        docker build . --file hip-clang.docker --tag $DOCKERIMAGE;
-        docker push $DOCKERIMAGE;
+
+        docker pull $DOCKER_IMAGE_UBUNTU:latest || true
```
pfultz2, line 86: Is it necessary to pull this image?

TedThemistokleous (approved):

.github/workflows/ci.yaml
causten: From the documentation I've read it needs to be a local cache unless we enable BuildKit.

PR #2224, opened 2023-09-20
Added mutex locks in register_target.cpp and created a multithreading…

### LINKED ISSUE #2208: Make register_target methods thread-safe
Make methods in register_target.cpp thread-safe and create a test for handling of register actions concurrent threads.

The method `migraphx::version_2_6_0::make_target()` was flagged by valgrind as causing a thread safety error.  This in turn caused one of the UIF Inference Server examples, `example_resnet50-migraphx,` to experience a SIGSEGV error.  

This error was first seen when UIF bumped ROCm versions from 5.4.x to 5.6.1.

bpickrel:
… test.


test/targets.cpp
```diff
@@ -50,4 +53,21 @@ TEST_CASE(targets)
     EXPECT(ts.size() >= 1);
 }
 
+TEST_CASE(concurrent_targets)
```
bpickrel, line 56: This test does not run from a clean slate, since the target list has already been modified by the other cases.  How can we reset it?

test/targets.cpp
```diff
@@ -50,4 +53,21 @@ TEST_CASE(targets)
...
+{
+    std::vector<std::thread> threads;
+
+    for(auto i = 0u; i < 10000; i++)
+    {
+        auto thread_body = []() {
+            auto ref_target = migraphx::make_target("gpu");
+            migraphx::register_target(ref_target);
```
bpickrel, line 64: This is a data race even with mutexes in both `make_target()` and `register_target()`.  make_target() accesses the `target_map()` via the destructor `~target_handler()` in register_target.hpp, and register_target() accesses it directly.  How can we ensure that access to both methods together is made atomic or protected by the same lock?

src/register_target.cpp
```diff
@@ -33,42 +34,88 @@ inline namespace MIGRAPHX_INLINE_NS {
...
 }
 
-void register_target(const target& t) { target_map()[t.name()] = t; }
+/**
+ * Insert a target name in the target_map; thread safe.
+ */
+void register_target(const target& t)
+{
+    std::unique_lock<std::mutex> lock(target_mutex());
+    target_map()[t.name()] = t;
```
TedThemistokleous, line 76: You should be unlocking here.

src/register_target.cpp
```diff
@@ -33,42 +34,88 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+
+/**
+ * Search for a target by name in the target_map; thread-safe.
+ */
+migraphx::optional<target> find_target(const std::string& name)
+{
+    // search for match or return none
+    std::unique_lock<std::mutex> lock(target_mutex());
+    const auto it = target_map().find(name);
+
```
TedThemistokleous, line 87: Add unlock before each return

test/targets.cpp
TedThemistokleous: Do something like 

if target state is not empty() -> clear the target_map() 

Maybe its worth creating an internal clear_targets() call that unregisters everything as part of this test as well.

src/register_target.cpp
```diff
@@ -33,42 +34,88 @@ inline namespace MIGRAPHX_INLINE_NS {
...
         store_target_lib(dynamic_loader(target_name));
+        t = find_target(name);
     }
-    const auto it = target_map().find(name);
-    if(it == target_map().end())
-    {
-        MIGRAPHX_THROW("Requested target '" + name + "' is not loaded or not supported");
-    }
-    return it->second;
+    // at this point we should always have a target
```
umangyadav, line 108: Why is it guaranteed to have a target at this point?

We have Three cases. 
1. Target is already registered. 
2. Target is not registered but can be dynamically loaded. 
3. Target requested is not supported by MIGraphX e.g. some target named "TPU". TPU is not registered with MIGraphX and it will not be found or can not be loaded dynamically either.

test/targets.cpp
umangyadav: You can use `unregister_target("ref")`

test/targets.cpp
```diff
@@ -50,4 +51,21 @@ TEST_CASE(targets)
...
 }
 
+TEST_CASE(concurrent_targets)
+{
+    std::vector<std::thread> threads;
+
+    for(auto i = 0u; i < 1000; i++)
+    {
+        auto thread_body = []() {
+            auto ref_target = migraphx::make_target("ref");
```
umangyadav, line 61: target needs to be registered first before it can be made. I think it works right now because it was already registered by other tests.

src/register_target.cpp
bpickrel: With std::unique_lock, the destructor automatically unlocks it when it goes out of scope.  This is supposed not only to be a coding convenience, but gives protection when things are interrupted by an exception. From [cplusplus.com](https://cplusplus.com/reference/mutex/unique_lock/):
"This class guarantees an unlocked status on destruction (even if not called explicitly). Therefore it is especially useful as an object with automatic duration, as it guarantees the mutex object is properly unlocked in case an exception is thrown."

src/register_target.cpp
pfultz2: Agreed, we should check `if (t == nullopt)` a second time and throw.

test/targets.cpp
```diff
@@ -50,4 +51,21 @@ TEST_CASE(targets)
     EXPECT(ts.size() >= 1);
 }
 
+TEST_CASE(concurrent_targets)
+{
+    std::vector<std::thread> threads;
+
+    for(auto i = 0u; i < 1000; i++)
```
pfultz2, line 58: This should use a multiple of the hardware conccurency: `std::thread::hardware_concurrency() * 4` or something like that.

test/targets.cpp
```diff
@@ -50,4 +51,21 @@ TEST_CASE(targets)
...
+    {
+        auto thread_body = []() {
+            auto ref_target = migraphx::make_target("ref");
+            migraphx::register_target(ref_target);
+        };
+
+        threads.emplace_back(thread_body);
+    }
+    for(auto& tt : threads)
+        tt.join();
```
pfultz2, line 68: Use `joinable_thread` instead of manually joining threads.

test/targets.cpp
bpickrel: > target needs to be registered first before it can be made (or alternatively targets can by loaded dynamically). I think it works right now because it was already registered by other tests.

It is registered by auto_register.hpp during program initialization, before the test starts.  I wouldn't have another way to make a target instance without calling `make_target()` first.

Recall that the goal of this test is not to make targets correctly; it's to check thread safety.

bpickrel: > and then call `unregister_target` at the end.

That's not allowed.  `unregister_target` is called automatically on exit, and it throws an exception if it's called twice with the same target name.  As it is, it's still possible in principle for one holder of a target to unregister it while others are still holding it.

Is there any use at all in having `unregister_target`?  It can only be used at program cleanup.

test/targets.cpp
bpickrel: Can you clarify?  That's not a class or function name.

test/targets.cpp
bpickrel: done

test/targets.cpp
bpickrel: Turns out that can't be done.  The target is created during program initialization and unregistered during program termination.  In fact, you can't call `unregister_target` at all for a real target because `unregister_target` will throw an exception at termination.

I think I should just say that how the target map gets cleaned up is not the problem of the mutex locks.

src/register_target.cpp
bpickrel: > Why is it guaranteed to have a target at this point? Let's say Target requested is not supported by MIGraphX e.g. some target named "TPU". TPU is not registered with MIGraphX and it will not be found or can not be loaded dynamically either.

Because the `dynamic_loader` constructor will throw an exception.  I added a line in the test for this.

bpickrel: > Agreed, we should check `if (t == nullopt)` a second time and throw.

That's redundant since `find_target` contains that check.

src/register_target.cpp
```diff
@@ -33,42 +34,88 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+}
 
+/**
+ * Get a target by name.  Load target library and register target if needed.
+ * Thread safe.
+ */
 target make_target(const std::string& name)
 {
-    if(not contains(target_map(), name))
+    //   no lock required here
```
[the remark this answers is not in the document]
bpickrel: The idea is that find_target does contain a lock, so make_target is already protected.

test/targets.cpp
umangyadav: > for a real target because unregister_target will throw an exception at termination. 

Yeah. Missed that. Thanks.

test/targets.cpp
```diff
@@ -50,4 +51,41 @@ TEST_CASE(targets)
...
+
+    auto n_threads = std::thread::hardware_concurrency() * 4;
+
+    for(auto i = 0u; i < n_threads; i++)
+    {
+        auto thread_body = [&target_name]() {
+
+            // remove all existing targets, if any
+            std::vector<std::string> target_list = migraphx::get_targets();
+            for(auto tt : target_list)
```
bpickrel, line 75: This new loop breaks the test now.  Multiple threads registering and unregistering a target with the same name is not safe.  Apparently it's not the target_map that's a problem but the underlying handles.  Trying to figure out how to protect against this, or if it's necessary.

test/targets.cpp
pfultz2: Its a class here: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/par_for.hpp#L36

test/targets.cpp
bpickrel: Followup:  I disabled this extra test but when I tested the existing fixes with amdinfer, the Inference Server, the segmentation violation still occurs.  I'm going to recommend we commit this PR as is but continue to search for the root cause of Issue #2208.

test/targets.cpp
bpickrel: done

test/targets.cpp
```diff
@@ -50,4 +51,43 @@ TEST_CASE(targets)
...
+#else
+    std::string target_name = "ref";
+#endif
+
+    auto n_threads = std::thread::hardware_concurrency() * 4;
+
+    for(auto i = 0u; i < n_threads; i++)
+    {
+        auto thread_body = [&target_name]() {
+            // TODO:  remove all existing targets, if any.
```
bpickrel, line 72: Followup: I disabled this extra test but when I tested the existing fixes with amdinfer, the Inference Server, the segmentation violation still occurs. I'm going to recommend we commit this PR as is but continue to search for the root cause of Issue https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2208.

bpickrel:
I'm recommending that we approve this PR so we can get it into the next ROCm release, even though the underlying problem in the Inference Server still occurs.  I can continue to look for the root cause.  If we can agree that this code is correct as far as it goes then let's merge it.

kahmed10 (approved):

causten:
It's been over a year.  Closing due to inactivity

pfultz2:
These are important changes and closing the PR makes it harder to find it in the future.

TedThemistokleous:
How do we want to handle the changes then @pfultz2 ? Do you want to update and I'll rereview?

TedThemistokleous:
> These are important changes and closing the PR makes it harder to find it in the future.

@kahmed10 @pfultz2  
Looks like there's something related to #2208 we still need to solve for this. Not sure if that's fixed as of late. Guess we should throw this in a sprint and investigate to ensure we don't lose this piece of work.

PR #2225, opened 2023-09-21
qlinearconv operator

lakhinderwalia:


src/onnx/parse_qlinearconv.cpp
```diff
@@ -0,0 +1,277 @@
...
+            auto auto_pad = info.attributes.at("auto_pad").s();
+            if(auto_pad.find("NOTSET") == std::string::npos)
+                MIGRAPHX_THROW("QLINEARCONV: auto_pad option fix..");
+        }
+
+        if(contains(info.attributes, "group"))
+            values["group"] = parser.parse_value(info.attributes.at("group")).template at<int>();
+
+        if(not values.empty())
+            recalc_conv_attributes(values, kdims);
```
umangyadav, line 179: I don't know why this would be necessary. Could remnant of older logic.

src/onnx/parse_qlinearconv.cpp
```diff
@@ -0,0 +1,277 @@
...
+        auto conv_sh   = conv_instr->get_shape();
+        auto conv_lens = conv_sh.lens();
+        auto conv_type = conv_sh.type();
+
+        auto broadcast_bias = info.add_instruction(
+            migraphx::make_op("broadcast", {{"axis", 1}, {"out_lens", conv_lens}}), bias_arg);
+        auto f_bias =
+            info.add_instruction(make_op("convert", {{"target_type", conv_type}}), broadcast_bias);
+
+        return info.add_instruction(migraphx::make_op("add"), conv_instr, f_bias);
```
umangyadav, line 230: This can make use of `add_common_op` which would handle broadcasting and conversion. 

Or better there is `add_bias` inside `onnx_parser.cpp` which can be modified to handle better handle quantized convolutions.

test/onnx/verify_onnx.cpp
```diff
@@ -1245,6 +1245,33 @@ TEST_CASE(nonzero_test)
...
 }
 
+TEST_CASE(qlinearconv_test)
+{
+    // https://xadupre.github.io/draft/onnx/onnx_doc_folder/onnx__QLinearConv.html
+    migraphx::program p = migraphx::parse_onnx("qlinearconv_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape s{migraphx::shape::uint8_type, {1, 1, 7, 7}};
+    std::vector<unsigned char> data = {
```
umangyadav, line 1255:
```suggestion
    std::vector<uint8_t> data = {
```

test/onnx/gen_onnx.py
```diff
@@ -5017,6 +5017,31 @@ def prelu_brcst_test():
     return ([node], [arg0, arg1], [arg_out])
 
 
+@onnx_test()
+def qlinearconv_test():
```
umangyadav, line 5021: Need to add test for parser inside onnx_test.cpp

test/onnx/gen_onnx.py
```diff
@@ -5017,6 +5017,31 @@ def prelu_brcst_test():
...
+    # https://xadupre.github.io/draft/onnx/onnx_doc_folder/onnx__QLinearConv.html
+    x = helper.make_tensor_value_info('X', TensorProto.UINT8, [1, 1, 7, 7])
+    sc_x = helper.make_tensor('1', TensorProto.FLOAT, [], [0.00369204697])
```
umangyadav, line 5024: you can also create test where these are 1-D tensor values instead of scalars or num_elements=1

src/onnx/parse_qlinearconv.cpp
```diff
@@ -0,0 +1,277 @@
...
+            copy(info.attributes.at("pads").ints(), std::back_inserter(padding));
+            check_attr_sizes(kdims, padding.size() / 2, "QLINEARCONV: inconsistent padding");
+            values["padding"] = std::vector<size_t>(padding.begin(), padding.end());
+        }
+
+        if(contains(info.attributes, "auto_pad"))
+        {
+            auto auto_pad = info.attributes.at("auto_pad").s();
+            if(auto_pad.find("NOTSET") == std::string::npos)
+                MIGRAPHX_THROW("QLINEARCONV: auto_pad option fix..");
```
umangyadav, line 172: Add better message like auto padding is not supported.

lakhinderwalia: Thanks. This was a placeholder message for me to add a fix here later. It is done. :-)

test/onnx/gen_onnx.py
lakhinderwalia: I have created many tests now, and one of them is now using a 1x1x1x1 tensor for Weights. Please see if you need something..  Thanks.

test/onnx/gen_onnx.py
lakhinderwalia: Done.

test/onnx/verify_onnx.cpp
lakhinderwalia: Thanks. Done.

src/onnx/parse_qlinearconv.cpp
lakhinderwalia: (We had a parallel discussion in a different thread and won't be using add_common_op). Thanks.

src/onnx/parse_qlinearconv.cpp
lakhinderwalia: Setting one set of attributes can mean checking on other attributes -- which have to change in sync..; (I don't know if it is an older logic.) Thanks.

src/onnx/parse_qlinearconv.cpp
```diff
@@ -0,0 +1,258 @@
...
+w_zero_point : T2
+
+y_scale : tensor(float)
+
+y_zero_point : T3
+
+B (optional) : T4
+
+Outputs
+y : T3
```
kahmed10, line 87:
```suggestion
Version
This version of the operator has been available since version 1 of the 'com.microsoft' operator set.

Attributes
auto_pad : string
channels_last : int
dilations : list of ints
group : int
kernel_shape : list of ints
pads : list of ints
strides : list of ints

Inputs (8 - 9)
x : T1
x_scale : tensor(float)
x_zero_point : T1
w : T2
w_scale : tensor(float)
w_zero_point : T2
y_scale : tensor(float)
y_zero_point : T3
B (optional) : T4

Outputs
y : T3
```
Minor suggestion for formatting consistency.

src/onnx/parse_qlinearconv.cpp
```diff
@@ -0,0 +1,258 @@
...
+        if(inp_arg.size() > 8 and inp_arg[8]->get_shape().type() != shape::int32_type)
+            MIGRAPHX_THROW("QLINEARCONV y bias should be int32");
+    }
+
+    // process all attributes of QLinearConv Operator..
+    value process_attributes(const onnx_parser& parser,
+                             const onnx_parser::node_info& info,
+                             const std::vector<instruction_ref>& args) const
+    {
+        value values;
```
pfultz2, line 142: Initialize this as `value values = value::object{}` so you dont need to check for empty when calling `make_op`.

src/onnx/parse_qlinearconv.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+        // for the dequantized output  y: scale & zero_pt
+        const instruction_ref& in_scale_y   = args[6];
+        const instruction_ref& in_zero_pt_y = args[7];
+
+        auto dquant_x = bcast_qdq_instr("dequantizelinear", in_x, in_scale_x, in_zero_pt_x, info);
+
+        auto dquant_w = bcast_qdq_instr("dequantizelinear", in_w, in_scale_w, in_zero_pt_w, info);
+
+        auto conv_op = values.empty() ? migraphx::make_op("convolution")
```
pfultz2, line 241: Looking at the `process_attributes` function, `values` are always set so `values.empty()` should always be false.

src/onnx/parse_qlinearconv.cpp
lakhinderwalia: Ah. Thanks for reminding me, I actually don't need to check for values being empty any more, because I have changed filled it with default values already -- that logic was recently changed..

test/onnx/gen_onnx.py
umangyadav: It is more for code coverage.  inside `bcast_qdq_instr`, it won't hit 1-D tensor case. I don't see any of the other tests being 1D tensors for scale or z-pt.

src/onnx/parse_qlinearconv.cpp
umangyadav: What was the reason why `add_common_op` can't be used ?. To me looks like this can still use `add_common_op` which. Doesn't make a difference though in terms of logic though.  Therefore minor point. 

FYI there is also `add_bias` inside `onnx_parser.cpp` which is better suited.

lakhinderwalia: add_bias(), as I see it is specific to 3D tensors. There is no such assumption here in this operator. Thanks.

umangyadav: > add_bias(), as I see it is specific to 3D tensor 
where do see that ? i think it just checks for input arg sizes to be 3

lakhinderwalia: Ah, sorry, I misread it.. Thanks.

I am looking inside it carefully. I have a convert operation in place here.. that is missing there..

test/onnx/gen_onnx.py
lakhinderwalia: Thanks. This is actually already covered in the real model resnet50-v1-12-int8.

lakhinderwalia: image

src/onnx/parse_qlinearconv.cpp
lakhinderwalia: (The process_attributes fills in default values, not empty anymore, and consequently the if clause has been removed.. Thanks.)

src/onnx/parse_qlinearconv.cpp
lakhinderwalia: (Removed the check for empty, as 'values' is never empty; Thanks.)

src/onnx/parse_qlinearconv.cpp
lakhinderwalia: Done.

test/onnx/gen_onnx.py
umangyadav: all scale and z-pt have num_elements =1 from what i see.

umangyadav: Is it part of CI ?

lakhinderwalia: No, but let me add a new graph and a test as you suggested..

kahmed10 (approved):

test/onnx/gen_onnx.py
lakhinderwalia: Done.

umangyadav (approved):

PR #2226, opened 2023-09-21
Disable the AttentionTest.AttentionBatch1_Float16 test

causten:
Recently the ORT test AttentionTest.AttentionBatch1_Float16 started failing in CI.  It appears to be completely random because I can rerun the CI job and it passes.  I've tried to get a recreate locally but so far it doesn't happen.  This would turn off the specific failing while investigation of the problem can be done on the side.


TedThemistokleous (approved):

umangyadav (approved):

causten:
it didnt help

PR #2227, opened 2023-09-21
Fix/missing miopen dll

tvukovic-amd:
Fixed `test_gpu_*` issue:` Exit code 0xc0000135` which is caused by missing MIOpen.dll inside AMDMIGraphX/build/bin directory. Added inside src/targets/gpu/CMakeLists.txt MIOpen.dll copy to appropriate directory (in Windows case). This solution fixed MIGraphX test issue which appeared in the following tests:
  - test_gpu_adjust_allocation
  - test_gpu_context_serialize
  - test_gpu_hip
  - test_gpu_literal
  - test_gpu_mlir
  - test_gpu_pack_args
  - test_gpu_pack_int8_args
  - test_gpu_quantization

### DIFF
--- a/src/targets/gpu/CMakeLists.txt
+++ b/src/targets/gpu/CMakeLists.txt
@@ -1,31 +1,31 @@
 # ####################################################################################
-#The MIT License (MIT)
+# The MIT License (MIT)
 #
-#Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
 #
-#Permission is hereby granted, free of charge, to any person obtaining a copy
-#of this software and associated documentation files (the "Software"), to deal
-#in the Software without restriction, including without limitation the rights
-#to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-#copies of the Software, and to permit persons to whom the Software is
-#furnished to do so, subject to the following conditions:
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
 #
-#The above copyright notice and this permission notice shall be included in
-#all copies or substantial portions of the Software.
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
 #
-#THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-#IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-#FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
-#AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-#LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-#OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-#THE SOFTWARE.
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
 # ####################################################################################
 
 list(APPEND CMAKE_PREFIX_PATH /opt/rocm)
 find_package(miopen)
 
-#rocblas
+# rocblas
 find_package(rocblas REQUIRED PATHS /opt/rocm)
 message(STATUS "Build with rocblas")
 
@@ -34,7 +34,7 @@ if(NOT TARGET MIOpen)
 endif()
 
 if(NOT WIN32)
-#TODO : re - enable when CK is ported to Windows
+    # TODO: re-enable when CK is ported to Windows
     find_package(composable_kernel 1.0.0 REQUIRED COMPONENTS jit_library)
 endif()
 
--- a/src/targets/gpu/CMakeLists.txt
+++ b/src/targets/gpu/CMakeLists.txt
@@ -282,3 +282,10 @@ rocm_install_targets(
     INCLUDE
     ${CMAKE_CURRENT_SOURCE_DIR}/include
 )
+
+if(WIN32)
+    add_custom_command(TARGET migraphx_gpu POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy  "${MIOPEN_LOCATION}" $<TARGET_FILE_DIR:migraphx_gpu>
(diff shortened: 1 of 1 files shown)


pfultz2:
Why do we need this for MIOpen and not for rocblas or hip?

Why doesn't setting the `PATH` work to find MIOpen correctly?

Also, MIOpen uses extra files(like perfdb, finddb, kerneldb, etc) which wont be used if we copy it our build directory.

apwojcik:
> Why do we need this for MIOpen and not for rocblas or hip?
> 
> Why doesn't setting the `PATH` work to find MIOpen correctly?
> 
> Also, MIOpen uses extra files(like perfdb, finddb, kerneldb, etc) which wont be used if we copy it our build directory.

UIF is a single team porting MIOpen and MIGraphX (and dependencies) to Windows. Therefore, there is no specific revision (commit) of MIOpen we base development of MIGraphX. We need that for the seamless debugging experience of MIGraphiX and MIOpen from the build directory of MIGraphX with the help of all sorts of debugging tools available.
PATH will not work because there are several builds we want to test against MIGraphX without vast reconfiguration of the whole environment.
It is enough to copy only the MIOpen library to let us run MIGraphX tests. If copying DB files is required, we will do it too. It will work on Windows because the location of DB files is not fixed like it is on Linux.

TedThemistokleous (approved):

TedThemistokleous:
> > Why do we need this for MIOpen and not for rocblas or hip?
> > Why doesn't setting the `PATH` work to find MIOpen correctly?
> > Also, MIOpen uses extra files(like perfdb, finddb, kerneldb, etc) which wont be used if we copy it our build directory.
> 
> UIF is a single team porting MIOpen and MIGraphX (and dependencies) to Windows. Therefore, there is no specific revision (commit) of MIOpen we base development of MIGraphX. We need the change from this PR for the seamless debugging experience of MIGraphiX and MIOpen from the build directory of MIGraphX with the help of all sorts of debugging tools available. PATH will not work because there are several builds of MIOpen we want to test against MIGraphX without vast reconfiguration of the whole environment. It is enough to copy only the MIOpen library to let us run MIGraphX tests. If copying DB files becomes required, we will do it too. It will work on Windows because the location of DB files is not fixed like it is on Linux.

I think this makes sense, since it looks like you've also contained your changes under IF(WIN32) block and shouldn't break existing builds that we use on *Nix right now

pfultz2:
> PATH will not work because there are several builds of MIOpen we want to test against MIGraphX without vast reconfiguration of the whole environment.

I dont understand why you need a vast reconfiguration. You just need to run with the new path:

```
cmake -E env PATH=$MIOPEN_PATH;$PATH cmake --build . --config Debug --target check
```

You could also have a script you can run to copy MIOpen over from the version you want and then you dont need to reconfigure anything or run the build from the command-line. 

Rebuilding migraphx to use a different MIOpen seems to require more reconfiguration. 

This also seems like a hack very specific to your workflow and is not something we want external users to use when we build for windows, especially since MIOpen will not be performant due to missing DB files and kernel caches. 

I dont think this should be merged in.

src/targets/gpu/CMakeLists.txt
```diff
@@ -282,3 +282,10 @@ rocm_install_targets(
     INCLUDE
     ${CMAKE_CURRENT_SOURCE_DIR}/include
 )
+
+if(WIN32)
+    add_custom_command(TARGET migraphx_gpu POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy  "${MIOPEN_LOCATION}" $<TARGET_FILE_DIR:migraphx_gpu>
```
pfultz2, line 288: Where is `MIOPEN_LOCATION` defined? MIOpen's package doesnt define this variable. The only location is the `miopen_DIR` variable that points to the cmake files.

pfultz2 (changes requested):
> PATH will not work because there are several builds of MIOpen we want to test against MIGraphX without vast reconfiguration of the whole environment.

I dont understand why you need a vast reconfiguration. You just need to run with the new path:

```
cmake -E env PATH=$MIOPEN_PATH;$PATH cmake --build . --config Debug --target check
```

You could also have a script you can run to copy MIOpen over from the version you want and then you dont need to reconfigure anything or run the build from the command-line. 

Rebuilding migraphx to use a different MIOpen seems to require more reconfiguration. 

This also seems like a hack very specific to your workflow and is not something we want external users to use when we build for windows, especially since MIOpen will not be performant due to missing DB files and kernel caches. 

I dont think this should be merged in.

apwojcik:
We are almost not using the command line in favor of IDEs for development. However, that requires support in CMakes to facilitate work with IDEs. The CMake team knows that and offers tooling to help developers work with IDEs. In this PR, we used one of the tools provided by the CMake team to make working with Visual Studio much easier.

> You could also have a script you can run to copy MIOpen over from the version you want, and then you don't need to reconfigure anything or run the build from the command-line.

That will not work when we generate Visual Studio project files. It has to be part of CMakeLists.txt to correctly generate post-build steps for the project. Access to MIOpen is required to use the built-in debugger from Visual Studio IDE. So, doing this via script will make it even more complicated. Ultimately, we must add the call to the script as a post-build step. It is much better to do it in CMake natively.

pfultz2:
It is not standard practice to copy dependencies artifcats into the build directory. Plus, messing with MIOpen's installation files can bring lots of problems and tightly couples the migraphx and miopen version. So we dont want to have that in our cmake build scripts. It should be done outside of the CMakeLists.txt.

> We are almost not using the command line in favor of IDEs for development.

This is really a false dichotomy. You should use both to get the best programmer productivity. 

> That will not work when we generate Visual Studio project files.

Why?

> It has to be part of CMakeLists.txt to correctly generate post-build steps for the project.

I dont see why this should be a post-build step. You just need to run the copy before you run the test executable. No need for any rebuilds whatsoever.

> Access to MIOpen is required to use the built-in debugger from Visual Studio IDE.

Sure thats why you have a script to copy it or set the `PATH` variable to find it. 

> So, doing this via script will make it even more complicated.

I dont see how calling a script before launching the debugger is complicated

Although setting the `PATH` in the launch.json seems much simpler to me than making copies.

tperry-amd:
> It is not standard practice to copy dependencies artifcats into the build directory. Plus, messing with MIOpen's installation files can bring lots of problems and tightly couples the migraphx and miopen version. So we dont want to have that in our cmake build scripts. It should be done outside of the CMakeLists.txt.

We are trying to do standard methods to solve this problem. PyTorch provides an [example](https://github.com/pytorch/pytorch/blob/main/docs/cpp/source/installing.rst). Another problem that can crop up from path manipulation is the reason PyTorch made this change in the first place related to Windows finding the wrong DLL's [here](https://github.com/pytorch/pytorch/issues/25457]. PyTorch does use MSVC instead of WIN32 which is a change I think we should consider.

> This is really a false dichotomy. You should use both to get the best programmer productivity.

Agreed, but we should also not neglect IDE's and provide a good experience for them as well.

> I dont see why this should be a post-build step. You just need to run the copy before you run the test executable. No need for any rebuilds whatsoever.

> Sure thats why you have a script to copy it or set the `PATH` variable to find it.

> I dont see how calling a script before launching the debugger is complicated
 
> Although setting the `PATH` in the launch.json seems much simpler to me than making copies.

It's a little hacky when you need to write custom scripts or do manual steps just to get debugging working. Ideally, we have a solution that works great with command line and IDE's that is industry standard for how to solve this problem on Windows. Modifying the launch.json may help this issue and we can definitely get it to work but it is not a great experience for easier developer workflows on Windows similar to what PyTorch provided.

Hopefully I made a convincing argument. If it is still a problem, we will pull the change and create a Python script or something to do the build/copy in a single step.

pfultz2:
> PyTorch provides an [example](https://github.com/pytorch/pytorch/blob/main/docs/cpp/source/installing.rst).

Thats not an example of good practice in cmake: modifying `CMAKE_CXX_FLAGS`, using variables instead of imported targets, etc. I also do not see pytorch or any other rocm component copying external dependencies into the build directory. 

MIOpen defines its installation with `cmake --build . --target install`(or through its generated packages). Writing our own custom install(via post build copies), breaks the encapsulation and is not a supported way of installing miopen. Any changes to miopen's installation(ie new files installed or removed) requires updates to migraphx(which miopen will not do for us since this is an unsupported way of installing miopen). This leads to migraphx and miopen version being tightly coupled, but we want the cmake to work with any future versions of miopen. It also requires extra maintenance to support newer versions of miopen.

> It's a little hacky when you need to write custom scripts or do manual steps just to get debugging working.

The ability to modify configurations like launch.json or write custom scripts is not a sign of a "hacky" approach. It's a testament to the flexibility and extensibility of many modern development environments.

> Ideally, we have a solution that works great with command line and IDE's that is industry standard for how to solve this problem on Windows.

Tightly coupling the versions between miopen and migraphx is not a great solution.

tperry-amd:
Understood, we'll pull the change and stick with developers configuring their environment.

PR #2229, opened 2023-09-21
CK GEMM Int8 Bug Fixes

turneram:
Adds workarounds to avoid passing capture ops and scalar literals from quantization as arguments to ck_gemm.


src/rewrite_quantization.cpp
```diff
@@ -62,9 +64,22 @@ void apply_quantizelinear(module& m, instruction_ref ins)
...
         min_quant = qt.min();
     });
-    auto s        = add_zero_point->get_shape();
-    auto min_arg  = m.add_literal(literal{shape{s.type()}, {min_quant}});
-    auto max_arg  = m.add_literal(literal{shape{s.type()}, {max_quant}});
+    auto s = add_zero_point->get_shape();
+    instruction_ref min_arg;
+    instruction_ref max_arg;
+
+    if(enabled(MIGRAPHX_ENABLE_CK{}))
```
pfultz2, line 71: I think it would be better to have a separate env variable to enable these changes, like `MIGRAPHX_ENABLE_CK_WORKAROUNDS` as this is a pessimization.

turneram: But without it CK will crash. Wouldn't it be better to have them on by default whenever CK is enabled?

test/rewrite_quantization_test.cpp
```diff
@@ -31,10 +31,13 @@
 #include <migraphx/ranges.hpp>
 #include <test.hpp>
 #include <migraphx/make_op.hpp>
+#include<migraphx/env.hpp>
```
umangyadav, line 34:
```suggestion
#include <migraphx/env.hpp>
```

test/rewrite_quantization_test.cpp
```diff
@@ -82,7 +85,7 @@ TEST_CASE(quantizelinear)
     EXPECT(any_of(*p1.get_main_module(), &is_quantizelinear));
     EXPECT(none_of(*p2.get_main_module(), &is_quantizelinear));
     // ensure clip literals created in quantized program are scalar
-    EXPECT(any_of(*p2.get_main_module(), &is_clip_scalar));
+    EXPECT(migraphx::enabled(MIGRAPHX_ENABLE_CK{}) xor any_of(*p2.get_main_module(), &is_clip_scalar));
```
umangyadav, line 88: if CK is enabled, then it should check if `clip` are not not scalar. This would just pass because it is doing `xor`ing

src/rewrite_quantization.cpp
pfultz2: The problem is how do we test if it works? Also, if mlir is enabled along with ck then we dont want to do this pessimization either. I think it would be better to disable CK fusions for int8 when there are scalars. Need to run numbers, but I think MLIR will be much faster then CK for all int8 cases.

umangyadav (approved):

PR #2236, opened 2023-09-22
Move stash for onnxrt build to use All targets vs Hip Clang

TedThemistokleous:
Trying this in hopes of fixing odd ORT build issues we're seeing.


causten:
Add an `env` print to the ONNX Runtime job, that will help identify the specific server just in case the full Jenkins log gets the print order messed up

causten (approved):

pfultz2:
Why do you think this would make a difference? Its the same gpu build.

umangyadav (approved):

umangyadav:
> Why do you think this would make a difference? Its the same gpu build.

yeah i don't think it will be helpful.

TedThemistokleous:
Closing this out as I believe the issue doesn't seem to be related to build used / we're seeing it again.

PR #2242, opened 2023-09-26
Add GroupNorm and LayerNorm onnx parsing

attila-dusnoki-htec:
This PR extends the normalization support with [GroupNorm](https://github.com/onnx/onnx/blob/main/docs/Operators.md#GroupNormalization) and [LayerNorm](https://github.com/onnx/onnx/blob/main/docs/Operators.md#LayerNormalization).
Since they are very similar in implementation and usage, it is grouped in one PR.

I saw that there was [already](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/match/layernorm.hpp) [support](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/cpu/layernorm.cpp) [for](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/layernorm.hpp) [layernorm](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/jit/layernorm.cpp). This extends the parsing of the onnx node, but still uses built in primitives for calculation.
Similar support for groupnorm is not added yet. Is it needed? [dnnl supports it](https://oneapi-src.github.io/oneDNN/group_dnnl_api_group_normalization.html). If so, i would recommend a follow-up PR instead of extending this.
Also [MIOpen just introduced layernorm](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320) as a primitive, but it is in BETA.

Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/94 and https://github.com/migraphx-benchmark/AMDMIGraphX/issues/100


src/onnx/parse_layernorm.cpp
```diff
@@ -0,0 +1,129 @@
...
+            axis = parser.parse_value(info.attributes.at("axis")).at<int64_t>();
+        }
+        float epsilon = 1e-5f;
+        if(contains(info.attributes, "epsilon"))
+        {
+            epsilon = parser.parse_value(info.attributes.at("epsilon")).at<float>();
+        }
+        if(contains(info.attributes, "stash_type"))
+        {
+            std::cerr << "WARNING: LAYERNORM does not support stash_type, it will be ignored.\n";
```
umangyadav, line 54: I don't see why it can't be supported. I find it useful to convert operations to stash_type to avoid accuracy loss.

umangyadav:
> Similar support for groupnorm is not added yet. Is it needed? 
Would be good to add that. But not sure about priority. We already have matcher that pattern matches primitives to LayerNorm. Perhaps it can be tweaked to match for group norm as well.  
cc : @pfultz2  what do you think ?

src/onnx/parse_layernorm.cpp
attila-dusnoki-htec: As far as I understood, that is only used with training, and no tests covers it in the python unit tests.

umangyadav (approved):

src/onnx/parse_layernorm.cpp
```diff
@@ -0,0 +1,129 @@
...
+        {
+            auto x_dims = x_shape.lens();
+            scale_bcast = info.add_instruction(
+                make_op("broadcast", {{"axis", skipped_axes}, {"out_lens", x_dims}}), scale);
+            if(not skip_bias)
+            {
+                bias_bcast = info.add_instruction(
+                    make_op("broadcast", {{"axis", skipped_axes}, {"out_lens", x_dims}}), bias);
+            }
+        }
```
CharlieL7, line 120: Note that this will not work if `x_shape` is dynamic. Probably outside of what we want to get done with this PR however. So should be fine.

CharlieL7 (approved):

PR #2244, opened 2023-09-26
Fix Round operator inaccuracy

gyulaz-htec:
Note: depends on #2179, until it's merged marked as Draft

This PR fixes inaccuracy with Round ONNX operator.
The inaccuracy was caused by ONNX round requires nearest integer rounding for halway (0.5) cases.
`std::round` rounds away from zero, thus giving wrong results with halfway cases.
Replaced `std::round` with `std::nearbyint` which uses the correct rounding by default.

Resolves: https://github.com/migraphx-benchmark/AMDMIGraphX/issues/134


src/include/migraphx/op/round.hpp
```diff
@@ -35,7 +35,7 @@ struct round : unary<round>
 {
     auto apply() const
     {
-        return [](auto x) { return std::round(x); };
+        return [](auto x) { return std::nearbyint(x); };
```
pfultz2, line 38: Rename this operator to `nearbyint`.

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
```diff
@@ -104,7 +104,7 @@ MIGRAPHX_DEVICE_MATH(isnan, ::isnan)
 MIGRAPHX_DEVICE_MATH(log, ::log)
 MIGRAPHX_DEVICE_MATH(pow, ::pow)
 MIGRAPHX_DEVICE_MATH(remainder, ::remainder)
-MIGRAPHX_DEVICE_MATH(round, ::round)
```
pfultz2, line 107: No need to remove the `round` overload.

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
```diff
@@ -104,7 +104,7 @@ MIGRAPHX_DEVICE_MATH(isnan, ::isnan)
...
+MIGRAPHX_DEVICE_MATH(round, ::nearbyint)
```
pfultz2, line 107: This should be `MIGRAPHX_DEVICE_MATH(nearbyint, ::nearbyint)`.

gyulaz-htec:
@pfultz2 I've updated th PR according to your change requests. Also I left the quantizelinear untouched, so those code paths are still using round instead of nearbyint.

umangyadav (approved):

src/include/migraphx/op/nearbyint.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+struct nearbyint : unary<nearbyint>
+{
+    auto apply() const
+    {
+        return [](auto x) { return std::nearbyint(x); };
```
lakhinderwalia, line 38: Requirement for correct behavior: std::fesetround(FE_TONEAREST) needs to be set. 

Else nearbyint() could give different results.

lakhinderwalia (changes requested):
PTAL @ quantizelinear.hpp as well. That needs a change, as per the ONNX spec. A test can also be created for parse_quantizelinear.cpp. Thanks.

src/include/migraphx/operators.hpp
```diff
@@ -84,6 +84,7 @@
 #include <migraphx/op/mod.hpp>
 #include <migraphx/op/mul.hpp>
 #include <migraphx/op/multibroadcast.hpp>
+#include <migraphx/op/nearbyint.hpp>
```
lakhinderwalia, line 87: Do we still need round.hpp? Thanks.

test/ref/round.cpp
```diff
@@ -37,7 +37,7 @@ TEST_CASE(round_test)
     migraphx::shape s{migraphx::shape::float_type, {9}};
     auto l =
         mm->add_literal(migraphx::literal{s, {1.1, 1.5, 1.6, -1.1, -1.5, -1.6, 0.0, 2.0, -2.0}});
-    mm->add_instruction(migraphx::make_op("round"), l);
+    mm->add_instruction(migraphx::make_op("nearbyint"), l);
```
lakhinderwalia, line 40: Sorry for the overlapping question elsewhere. If this test is for the round() operator, why change it by inserting a nearbyint operator, and still leave both operators in existence. One of them should likely go away, it seems, and maybe just the round() operator should be fixed, and still called round. Thanks.

lakhinderwalia: {1.1, 1.5, 1.6, -1.1, -1.5, -1.6, 0.0, 2.0, -2.0}  --> please add some more test values here to test even rounding. 

Suggestion, please add (or something similar) : -1.4,  0.5, 0.6, 2.4, 2.5, 2.6, -2.4, -2.5, -2.6.

test/ref/round.cpp
```diff
@@ -53,7 +53,7 @@ TEST_CASE(round_dyn_test)
     migraphx::shape::dynamic_dimension dd{4, 10};
     migraphx::shape s{migraphx::shape::float_type, {dd}};
     auto input = mm->add_parameter("X", s);
-    mm->add_instruction(migraphx::make_op("round"), input);
+    mm->add_instruction(migraphx::make_op("nearbyint"), input);
```
lakhinderwalia, line 56: (More test values, are suggested above. Thanks.)

src/include/migraphx/op/nearbyint.hpp
gyulaz-htec: The default rounding mode is `FE_TONEAREST` (https://www.gnu.org/software/libc/manual/html_node/Rounding.html), Since no other code paths are using `std::fesetround` I assumed there is no need to explicitly write it there.
I still cann add it if you think this could be a possible source of issues in the future.

src/include/migraphx/operators.hpp
gyulaz-htec: I was confused by these previous review comments [1](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2244#discussion_r1349341859),  [2](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2244#discussion_r1349341998) and I thought I have to provide a new operator for this. Also I was insecure about dropping the previous round operator because it was still used in quantize_linear.

gyulaz-htec:
@lakhinderwalia thanks for the review and the clarification for quantizelinear. So my current plan now is to drop the round.hpp and keep only nearbyint.hpp, and clean-up the previous usage of round (this also involves renaming the test/ref/round.cpp to test/ref/nearbyint.cpp).

src/include/migraphx/operators.hpp
lakhinderwalia: We do have a `round` operator in Onnx. It is just that internally it has to follow the logic of nearbyint(). So please do keep that operator by name. Perhaps it would be far simpler if the name `round` was left as it is, and the new operator name nearbyint() wasn't introduced at all. That is my preference. @pfultz2 & @umangyadav ?

src/include/migraphx/op/nearbyint.hpp
lakhinderwalia: Thanks, but no default is specified in the c++/c specification: https://en.cppreference.com/w/cpp/numeric/fenv/FE_round
(Also, perhaps setting the pragma is a good idea.)

lakhinderwalia:
> @lakhinderwalia thanks for the review and the clarification for quantizelinear. So my current plan now is to drop the round.hpp and keep only nearbyint.hpp, and clean-up the previous usage of round (this also involves renaming the test/ref/round.cpp to test/ref/nearbyint.cpp).

I would recommend keeping round.hpp, as the operator names matches ONNX name. There is no operator called nearbyint.hpp. But let @pfultz2 & @umangyadav also chime in -- I made a similar comment above, but don't want any confusion or suggest any extra work for the PR. This could simply be a bug fix, just replacing std::round() with std::nearbyint() :-) 
All the tests etc. then stay the same (by name) -- just enhance the test values, as suggested above.
Thanks again.

pfultz2:
We don't need to match our operators exactly with the onnx name. I would prefer the operator name match the C++ function since we use the name of the operator to generate the function call for GPU kernel. Furthermore it makes the IR better represent what functions will be called by the backends.

lakhinderwalia:
> We don't need to match our operators exactly with the onnx name.

This is a simple bug to fix to call a different function within <cmath>. That shouldn't drive a name-change.. 

And the intent is not nearestInt -- this name is misleading. It is nearestEvenInt.

gyulaz-htec:
@lakhinderwalia @pfultz2 I've made the following updates:
- Added more test values as requested to test even rounding more.
- Updated the rounding in `quantize_linear`, according to the onnx [spec](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#QuantizeLinear-19) that requires nearest even rounding as well
- Added `std::fesetround` calls and the related pragma as requsted
- Replaced `round.hpp` with `nearbyint.hpp`, since there was no conclusion on what would be the desired way to do it. I can revert the name if it's more prefered.

lakhinderwalia (approved):
Thanks for the changes. 

The overall operator renaming for a simple bug fix is a huge over kill. But it is not worth my time arguing about it as I have made my points earlier, the intent of nearbyint here is not adding any clarity -- it doesn't suggest the mode of rounding, after all:

All rounding operations are nearbyint ;-) They were rounding to the nearby integer even before this change :-)

gyulaz-htec:
> All rounding operations are nearbyint ;-) They were rounding to the nearby integer even before this change :-)

@lakhinderwalia that's not exactly the case. `std::round` rounds away from zero: https://en.cppreference.com/w/cpp/numeric/math/round
and doesn't round to the nearest even in all cases e.g `-2.5` is rounded to `-3.0` instead of the expected`-2.0`.
The fact that the rounding mode of `std:nearbyint` can't be changed with `std::fesetround` drives the entire change to replace it with `std::nerbyint` which respects the setting from `std::fesetround`

test/onnx/verify_onnx.cpp
```diff
@@ -2056,6 +2056,43 @@ TEST_CASE(reversesequence_time_verify_test)
     EXPECT(migraphx::verify::verify_rms_range(result_vector, gold));
 }
 
+TEST_CASE(round_half_test)
+{
+    migraphx::program p = migraphx::parse_onnx("round_half_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape xs{migraphx::shape::half_type, {4, 4}};
+    std::vector<float> tmp = {{-3.51,
```
umangyadav, line 2065:
```suggestion
    std::vector<float> tmp = {-3.51,
```

test/onnx/verify_onnx.cpp
```diff
@@ -2056,6 +2056,43 @@ TEST_CASE(reversesequence_time_verify_test)
...
+                               -1.6,
+                               -1.5,
+                               -0.51,
+                               -0.5,
+                               0.5,
+                               0.6,
+                               2.4,
+                               2.5,
+                               3.5,
+                               4.5}};
```
umangyadav, line 2080:
```suggestion
                               4.5};
```

umangyadav (commented):
LGTM, Fix extra parenthesis

gyulaz-htec:
@umangyadav I've fixed the extra parenthesis

lakhinderwalia:
> > All rounding operations are nearbyint ;-) They were rounding to the nearby integer even before this change :-)
> 
> @lakhinderwalia that's not exactly the case. `std::round` rounds away from zero: 

Please note, just to clarify, I merely pointed out that 'round' operations are nearbyint: they go to the next integer; The pertinent question being handled in this bugfix is, _**which**_ next integer:

It is the renaming exercise that is imo un-necessary, hence I pointed out that renaming to nearbyint doesn't add any clarity. Particularly when it is the rounding mode that decides things. So the above point wasn't an argument about _round_ vs _nearbyint_ :-)

PR #2250, opened 2023-09-26
Add CK GEMM-Softmax-GEMM fusion

turneram:


src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+    std::vector<std::string> result;
+    for(auto& ins : m)
+    {
+        if(starts_with(ins.name(), "@"))
+            continue;
+        if(contains({"multibroadcast", "contiguous"}, ins.name()))
+            continue;
+        if(ins.name() == "pointwise")
+        {
+            return is_mul_module(*ins.module_inputs().front());
```
pfultz2, line 217: Why is this called recursively?

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
     }
 };
 
+static bool is_mul_module(const module& m)
+{
+    std::vector<std::string> result;
+    for(auto& ins : m)
```
pfultz2, line 209: I think this could be implemented with `any_of`.

src/targets/gpu/fuse_ck.cpp
turneram: Meant to simplify that. I adapted it from another function

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
+        auto ins       = r.result;
+        auto gemm2_ins = r.instructions["gemm2"];
+        auto gemm1_ins = r.instructions["gemm1"];
+        auto scale_ins = r.instructions["scale"];
+
+        if(scale_ins->module_inputs().size() != 1 or
+           not is_mul_module(*scale_ins->module_inputs().front()))
```
pfultz2, line 247: Can you make a custom predicate matcher to check for this?

```cpp
MIGRAPHX_PRED_MATCHER(is_pointwise_scale, instruction_ref ins)
{
    if (ins->name() != "pointwise")
        return false;
    if (ins->module_inputs().size() != 1)
        return false;
    return is_mul_module(ins->module_inputs().front());
}
```

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+        mpm.get_module().replace_instruction(
+            ins, ck_gemm_softmax_gemm{gemm2_ins->get_operator(), scale}, inputs);
+    }
+};
+
 } // namespace
 
 void fuse_ck::apply(module_pass_manager& mpm) const
 {
+    match::find_matches(mpm, find_ck_gemm_softmax_gemm{});
```
pfultz2, line 280: I dont think we need an extra `find_matches`, I think this matcher could be added below.

src/targets/gpu/kernels/include/migraphx/kernels/ck_gemm_softmax_gemm.hpp
```diff
@@ -0,0 +1,85 @@
...
+// In CK, the B matrix is ordered as N,K instead of K,N
+template <class Dims>
+constexpr auto ck_transposeb_dims(Dims dims)
+{
+    return unpack(dims, [](auto k, auto n) { return make_const_array(n, k); });
+}
+
+template <class Tensor>
+using ck_transposeb = decltype(make_shape(ck_transposeb_dims(get_shape_c<Tensor>{}.lens),
+                                          ck_transposeb_dims(get_shape_c<Tensor>{}.strides)));
```
pfultz2, line 45: This seems to be copied from ck_gemm.hpp, can we put this in ck.hpp header?

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+}
+
+struct find_ck_gemm_softmax_gemm
+{
+    auto matcher() const
+    {
+        auto gemm1 =
+            match::skip(match::name("contiguous"))(match::name("dot")(is_ck_gemm().bind("gemm1")));
+        auto mul = match::name("pointwise")(match::any_of[match::inputs()](gemm1)).bind("scale");
+        auto softmax = match::name("softmax")(match::any_of[match::inputs()](mul)).bind("softmax");
```
pfultz2, line 234: Use `match::arg(0)` as softmax only takes one argument.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+        return match::name("dot")(is_ck_gemm().bind("gemm2"))(
+            match::any_of[match::inputs()](softmax));
```
pfultz2, line 236: Doesnt the softmax have to match the first argument?

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+    return false;
+}
+
+struct find_ck_gemm_softmax_gemm
+{
+    auto matcher() const
+    {
+        auto gemm1 =
+            match::skip(match::name("contiguous"))(match::name("dot")(is_ck_gemm().bind("gemm1")));
+        auto mul = match::name("pointwise")(match::any_of[match::inputs()](gemm1)).bind("scale");
```
pfultz2, line 233: This should be `match::either_arg(0, 1)(match::constant().bind("scale"), gemm1).bind("mul")`.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+        return match::name("dot")(is_ck_gemm().bind("gemm2"))(
+            match::any_of[match::inputs()](softmax));
+    }
+
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
+        auto ins       = r.result;
+        auto gemm2_ins = r.instructions["gemm2"];
+        auto gemm1_ins = r.instructions["gemm1"];
+        auto scale_ins = r.instructions["scale"];
```
pfultz2, line 244: This should be named `mul_ins` instead and use `scale` for the constant.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+        if(scale_ins->module_inputs().size() != 1 or
+           not is_mul_module(*scale_ins->module_inputs().front()))
+            return;
+        if(not ck_gemm_softmax_gemm::is_ck_supported_type(gemm1_ins->get_shape().type()))
+            return;
+
+        double scale = 1.0;
+        for(auto& in : scale_ins->inputs())
+        {
+            if(in->can_eval())
```
pfultz2, line 255: This should be bound in the matcher.

src/targets/gpu/fuse_ck.cpp
pfultz2: Ok, I see, I dont think we have nested pointwise modules, so I dont think this will be needed.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
 };
 
+static bool is_mul_module(const module& m)
+{
+    std::vector<std::string> result;
+    for(auto& ins : m)
+    {
+        if(starts_with(ins.name(), "@"))
+            continue;
+        if(contains({"multibroadcast", "contiguous"}, ins.name()))
```
pfultz2, line 213: Pointwise modules should never have these functions.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,81 @@ struct find_ck_gemm
...
+            continue;
+        if(ins.name() == "pointwise")
+        {
+            return is_mul_module(*ins.module_inputs().front());
+        }
+        else if(ins.name() == "mul")
+        {
+            return true;
```
pfultz2, line 221: What if there is more operators after mul? It seems like it would be better to check if its returning mul and that the inputs to mul are input parameters. We could write a matcher for it:

```cpp
auto is_mul = match::arg(0)(match::name("mul")(match::all_of[match::inputs()](match::name("@param"))));
return match_instruction(m, std::prev(m.end()), is_mul);
```

src/targets/gpu/fuse_ck.cpp
turneram: We need to match in the order of gemm-softmax-gemm, gemm-pointwise, gemm. When I put the matchers in the same `find_matches` call, I didn't see that order being maintained regardless of the order in which they were passed.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,69 @@ struct find_ck_gemm
...
+        auto ins       = r.result;
+        auto gemm2_ins = r.instructions["gemm2"];
+        auto gemm1_ins = r.instructions["gemm1"];
+        auto scale_lit = r.instructions["scale"];
+
+        if(not ck_gemm_softmax_gemm::is_ck_supported_type(gemm1_ins->get_shape().type()))
+            return;
+
+        double scale = 1.0;
+        scale_lit->get_literal().visit([&](const auto s) {
```
pfultz2, line 247: I think you need to call `eval` to get the literal if its a composition of other functions. `match::is_constant` just means that `can_eval` is true.

src/targets/gpu/fuse_ck.cpp
pfultz2: I was thinking `match::find_matches(mpm, find_ck_gemm_softmax_gemm{}, find_ck_gemm_pointwise{})`, does that work?

src/targets/gpu/kernels/include/migraphx/kernels/ck.hpp
```diff
@@ -154,6 +154,29 @@ struct ck_add
...
+{
+    return unpack(dims, [](auto k, auto n) { return make_const_array(n, k); });
+}
+
+template <class Tensor>
+using ck_transposeb = decltype(make_shape(ck_transposeb_dims(get_shape_c<Tensor>{}.lens),
+                                          ck_transposeb_dims(get_shape_c<Tensor>{}.strides)));
+
+template <class T>
+struct ck_gemm_softmax_gemm_settings
```
pfultz2, line 169: This class should go into the ck_gemm_softmax_gemm.hpp header.

src/targets/gpu/fuse_ck.cpp
turneram: Both `match::find_matches(mpm, find_ck_gemm_softmax_gemm{}, find_ck_gemm_pointwise{});` and `match::find_matches(mpm, find_ck_gemm_pointwise{}, find_ck_gemm_softmax_gemm{});` seem to apply `find_ck_gemm_pointwise` first, which prevents `find_ck_gemm_softmax_gemm` from working.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -65,21 +65,59 @@ struct ck_gemm
...
+    {
+        if(not contains(range(s.strides().rbegin(), s.strides().rbegin() + 3), 1))
+            MIGRAPHX_THROW("Invalid shape for ck_gemm_softmax_gemm");
+    }
+
+    shape compute_shape(std::vector<shape> inputs, const std::vector<module_ref>&) const
+    {
+        check_shapes{inputs, *this}.same_ndims();
+        if(inputs.size() < 3)
+            MIGRAPHX_THROW("Expected 3 inputs but got " + to_string(inputs.size()));
```
umangyadav, line 99:
```suggestion
            MIGRAPHX_THROW("ck::gemm_softmax_gemm: Expected 3 inputs but got " + to_string(inputs.size()));
```

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,69 @@ struct find_ck_gemm
...
+    return is_mul_module(*ins->module_inputs().front());
+}
+
+struct find_ck_gemm_softmax_gemm
+{
+    auto matcher() const
+    {
+        auto gemm1 =
+            match::skip(match::name("contiguous"))(match::name("dot")(is_ck_gemm().bind("gemm1")));
+        auto mul     = match::name("pointwise")(match::either_arg(0, 1)(
```
umangyadav, line 228: can this pointwise module take more than two params ? Or should it be exactly two ? 
Reason i am asking is that `mul` matcher can match if there are more than two inputs to pointwise module. 

```
(match::all_of[match::inputs()](match::name("@param")))
```

src/targets/gpu/jit/ck_gemm_softmax_gemm.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+namespace gpu {
+
+using namespace migraphx::gpu::gen; // NOLINT
+
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_LOG_CK_GEMM);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_CK_TUNING);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_CK_TUNING_VALUE);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_CK_DEBUG);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TUNE_CK);
```
umangyadav, line 53: Would it be better if we had separate header file for all CK ENV vars since they are being defined and used in multiple places ?

src/targets/gpu/jit/ck_gemm_softmax_gemm.cpp
```diff
@@ -0,0 +1,461 @@
...
+}
+
+}
+
+} // namespace migraphx
+
+)__migraphx__";
+
+// NOLINTNEXTLINE
+static const char* const disable_warning_pragma = R"__migraphx__(
```
umangyadav, line 86: This can also be put into separate header file. Many of the other functions can also be put into that

src/targets/gpu/jit/ck_gemm_softmax_gemm.cpp
```diff
@@ -0,0 +1,461 @@
...
+            std::for_each(vinputs.begin() + 2, vinputs.end(), fold_batch_dims);
+            options.virtual_inputs = vinputs;
+        }
+
+        if(v.get("check", false) or enabled(MIGRAPHX_CK_DEBUG{}))
+            options.params += " -DMIGRAPHX_CK_CHECK=1";
+
+        // scale
+        assert(v.contains("scale"));
+        auto scale = v.at("scale").to<float>();
```
umangyadav, line 391: scale was defined as `double` originally. but then used as `float`.  Should be okay most probably but better to use `float` everywhere.

test/verify/ck_gemm_softmax_gemm.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+struct ck_gemm_softmax_gemm : verify_program<ck_gemm_softmax_gemm>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        migraphx::shape m1_shape{migraphx::shape::half_type, {1, 12, 256, 256}};
+        migraphx::shape m2_shape{migraphx::shape::half_type, {1, 12, 256, 256}};
+        auto m2_elements = 1 * 12 * 256 * 256;
```
umangyadav, line 38:
```suggestion
        auto m2_elements = m2_shape.elements();
```

src/targets/gpu/jit/ck_gemm_softmax_gemm.cpp
pfultz2: I think most of these can be removed as it can use the new tuning infrastructure.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,69 @@ struct find_ck_gemm
     }
 };
 
+static auto is_mul_module(module& m)
```
pfultz2, line 206: You need to remove `static` from this function to fix the tidy issue.

src/targets/gpu/fuse_ck.cpp
pfultz2: It should only take two parameters.

src/targets/gpu/fuse_ck.cpp
pfultz2: I see, thats because it matches the last gemm.

src/targets/gpu/jit/ck_gemm_softmax_gemm.cpp
```diff
@@ -0,0 +1,461 @@
...
+                            {p.second.data(), p.second.data() + p.second.size()}};
+        });
+    return srcs;
+}
+
+static const std::vector<src_file>& ck_headers()
+{
+    static const auto& headers = create_ck_headers();
+    return headers;
+}
```
pfultz2, line 128: I think these common functions should be moved to a common header.

src/targets/gpu/fuse_ck.cpp
umangyadav: okay. Better to enforce that using `nargs(2)`

src/targets/gpu/fuse_ck.cpp
```diff
@@ -161,10 +203,69 @@ struct find_ck_gemm
...
+        auto mul     = match::name("pointwise")(match::either_arg(0, 1)(
+            match::is_constant().bind("scale"), gemm1))(is_pointwise_scale());
+        auto softmax = match::name("softmax")(match::arg(0)(mul)).bind("softmax");
+
+        return match::name("dot")(is_ck_gemm().bind("gemm2"))(match::arg(0)(softmax));
+    }
+
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
+        std::cout << "Matched GSG" << std::endl;
```
umangyadav, line 237: Remove `cout`. If you need to, 
You can also set both  `MIGRAPHX_TRACE_MATCHES=1` and `MIGRAPHX_TRACE_MATCHES_FOR="ck_gemm_softmax_gemm` to see matched instructions for just this matcher.

src/targets/gpu/jit/ck_gemm_softmax_gemm.cpp
turneram: I assume keeping `MIGRAPHX_CK_DEBUG` and `MIGRAPHX_TUNE_CK` would be useful. Would `MIGRAPHX_LOG_CK_GEMM` also potentially be?

src/include/migraphx/ck.hpp
```diff
@@ -0,0 +1,81 @@
...
+inline namespace MIGRAPHX_INLINE_NS {
+
+#ifndef _WIN32
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_CK);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_LOG_CK_GEMM);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_CK_DEBUG);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TUNE_CK);
+#endif
+
+struct gemm_softmax_gemm
```
pfultz2, line 41: This should be moved to prefuse_ops.cpp.

src/include/migraphx/ck.hpp
```diff
@@ -0,0 +1,81 @@
...
+    operation op = make_op("dot");
+    float scale  = 1.0;
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.op, "op"), f(self.scale, "scale"));
+    }
+
+    std::string name() const { return "gemm_softmax_gemm"; }
```
pfultz2, line 52: Should be `gpu::gem_softmax_gemm`.

src/include/migraphx/ck.hpp
```diff
@@ -0,0 +1,81 @@
...
+            check_gemm_shape(input);
+        }
+        return op.compute_shape({op.compute_shape({a, b}), b1});
+    }
+
+    static bool is_ck_supported_type(shape::type_t t) { return contains({shape::half_type}, t); }
+};
+
+} // namespace MIGRAPHX_INLINE_NS
+} // namespace migraphx
```
pfultz2, line 79: This header should be under migraphx/gpu as this is just for gpu target.

src/include/migraphx/ck.hpp
pfultz2: Wait you want to inherit that. So this should probably be moved `migraphx/gpu/gemm_softmax_gemm.hpp` header.

umangyadav (approved):

src/targets/gpu/target.cpp
```diff
@@ -76,9 +77,6 @@ namespace gpu {
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_SCHEDULE_PASS)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_REDUCE_FUSION)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_NHWC)
-#ifndef _WIN32
-MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_CK)
```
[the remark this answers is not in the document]
pfultz2: Or make a function, in the `fuse_ck.hpp` header, similiar to `is_mlir_enabled()` function.

src/targets/gpu/fuse_ck.cpp
```diff
@@ -22,10 +22,11 @@
  * THE SOFTWARE.
  */
 #include <migraphx/gpu/fuse_ck.hpp>
+#include <migraphx/gpu/gemm_softmax_gemm.hpp>
 #include <migraphx/matcher.hpp>
 #include <migraphx/pass_manager.hpp>
-#include <migraphx/make_op.hpp>
 #include <migraphx/register_op.hpp>
+#include <migraphx/gpu/ck.hpp>
```
pfultz2, line 29: This header shouldn't be needed here.

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -23,16 +23,17 @@
...
 #include <migraphx/permutation.hpp>
 #include <migraphx/gpu/prefuse_ops.hpp>
+#include <migraphx/gpu/gemm_softmax_gemm.hpp>
 #include <migraphx/match/layernorm.hpp>
-#include <migraphx/check_shapes.hpp>
-#include <migraphx/make_op.hpp>
 #include <migraphx/register_op.hpp>
 #include <migraphx/pass_manager.hpp>
 #include <migraphx/dead_code_elimination.hpp>
+#include <migraphx/gpu/ck.hpp>
```
pfultz2, line 31: Dont include this header here.

pfultz2 (approved):

PR #2254, opened 2023-09-27
Add support for uneven Split operations

mirza-halilcevic:
- Since opset version 18, the Split operator allows splitting into unevenly sized outputs when the `split` input is not present, and a `num_outputs` attribute has been introduced.
- Resolves migraphx-benchmark/AMDMIGraphX#132


umangyadav (approved):

src/onnx/parse_split.cpp
```diff
@@ -68,13 +68,25 @@ struct parse_split : op_parser<parse_split>
...
+                                   " doesn't match actual number of outputs " +
+                                   std::to_string(info.num_outputs) + "!");
+                }
             }
-            auto dl = lens[tuned_axis] / info.num_outputs;
-            vec_splits.resize(info.num_outputs, dl);
+
+            auto dl = lens[tuned_axis] / num_outputs;
+            vec_splits.resize(lens[tuned_axis] % num_outputs, dl + 1);
+            vec_splits.resize(num_outputs, dl);
```
CharlieL7, line 89: It's not clear to me if this is the intended behavior from ONNX. The specification is not clear about how to handle uneven splits and only says the last "chunk" will be smaller. Numpy's `split` function raises an error when it cannot be split evenly.

test/onnx/gen_onnx.py
```diff
@@ -7157,6 +7157,42 @@ def split_test_no_attribute():
...
 
 
+@onnx_test()
+def split_test_uneven():
+    x = helper.make_tensor_value_info('x', TensorProto.FLOAT, [12, 15])
+    y1 = helper.make_tensor_value_info('y1', TensorProto.FLOAT, [3, 15])
+    y2 = helper.make_tensor_value_info('y2', TensorProto.FLOAT, [3, 15])
+    y3 = helper.make_tensor_value_info('y3', TensorProto.FLOAT, [2, 15])
+    y4 = helper.make_tensor_value_info('y4', TensorProto.FLOAT, [2, 15])
+    y5 = helper.make_tensor_value_info('y5', TensorProto.FLOAT, [2, 15])
```
CharlieL7, line 7167: This test highlights the possible issue

CharlieL7 (changes requested):
The ONNX spec is unclear about uneven splits, we need to discuss how we want to handle this

src/onnx/parse_split.cpp
mirza-halilcevic: I agree that the ONNX spec is not clear in what they mean by "chunk". I based this solution on Numpy's [`array_split`](https://numpy.org/doc/stable/reference/generated/numpy.array_split.html).

test/onnx/gen_onnx.py
mirza-halilcevic: In this particular case, if we assumed that "the last chunk" from the ONNX spec means the last output, then the only way to split this into 5 outputs would be `[3, 3, 3, 3, 0]`. I'm not sure if it makes sense to leave the last output empty.

umangyadav: https://github.com/onnx/onnx/pull/4481

https://github.com/onnx/onnx/blob/60b0a34037f2980c519793b1bbc3a6ed79b0dfa3/onnx/defs/tensor/defs.cc#L779

Need to mimic this logic.

umangyadav (approved):

CharlieL7 (approved):
LGTM, fix CI and should be good to go

PR #2264, opened 2023-09-28
Specify full path for libamdhip64.so

ahsan-ca:
test_py_3.6_gpu_async complains about a missing shared library libamdhip64.so
on sles, specifying the full path seems to fix the issue.
This PR makes the change to add the full path to the shared library provided
the shared library exists at that path.

### DIFF
--- a/test/py/test_gpu_async.py
+++ b/test/py/test_gpu_async.py
@@ -26,6 +26,7 @@ import ctypes
 import os
 import glob
 
+
 def test_conv_relu():
     # Full path of the library is needed to fix an issue on sles
     # where the library is not loaded otherwise.
--- a/test/py/test_gpu_async.py
+++ b/test/py/test_gpu_async.py
@@ -24,14 +24,28 @@
 import migraphx
 import ctypes
 import os
+import glob
 
 def test_conv_relu():
+
     library = "libamdhip64.so"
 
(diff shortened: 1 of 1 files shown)


test/py/test_gpu_async.py
```diff
@@ -23,10 +23,17 @@
...
+import os
 
 def test_conv_relu():
-    hip = ctypes.cdll.LoadLibrary("libamdhip64.so")
+    library = "libamdhip64.so"
+
+    # Full path of the library is specified to fix
+    # an issue on sles where the library is not loaded otherwise.
+    if os.path.exists("/opt/rocm/lib/libamdhip64.so"):
+        library = "/opt/rocm/lib/libamdhip64.so"
```
umangyadav, line 34: This could check first if `/opt/rocm` exists or not. 

If not then can check if `/opt/rocm-*` exists or not.

pfultz2 (approved):

test/py/test_gpu_async.py
ahsan-ca: Done.

ahsan-ca:
Modified the PR to check for paths in the order below:

1. `rocm_path` env variable.
2. `/opt/rocm`.
3. `/opt/rocm-*`

umangyadav (approved):

PR #2267, opened 2023-09-30
Update lib target soversion

arvindcheru:
Summary of proposed changes:

- Update lib target so version format to recommended format - (ex: libmigraphx.so.2.7.0 updated to libmigraphx.so.2.7)


pfultz2:
Why are we dropping the patch version?

arvindcheru:
> Why are we dropping the patch version?
Looking latest versions majority of modules are having soversion with major,minor based link names. there are 3 other modules including migraphx were only showing difference by having patch version also. these 4 modules are requested to update using these changes proposed to the recommended soversion format.

pfultz2:
The major, minor and patch should be part of the SO major version:

```cmake
math(EXPR MIGRAPHX_SO_MAJOR_VERSION "(${PROJECT_VERSION_MAJOR} * 1000 * 1000) + (${PROJECT_VERSION_MINOR} * 1000) + ${PROJECT_VERSION_PATCH}")
set(MIGRAPHX_SO_VERSION ${MIGRAPHX_SO_MAJOR_VERSION}.0)
```

umangyadav:
> The major, minor and patch should be part of the SO major version:
> 
> ```cmake
> math(EXPR MIGRAPHX_SO_MAJOR_VERSION "(${PROJECT_VERSION_MAJOR} * 1000 * 1000) + (${PROJECT_VERSION_MINOR} * 1000) + ${PROJECT_VERSION_PATCH}")
> set(MIGRAPHX_SO_VERSION ${MIGRAPHX_SO_MAJOR_VERSION}.0)
> ```
> 
> Or we if we can drop the minor version it could just be:
> 
> ```cmake
> math(EXPR MIGRAPHX_SO_VERSION "(${PROJECT_VERSION_MAJOR} * 1000 * 1000) + (${PROJECT_VERSION_MINOR} * 1000) + ${PROJECT_VERSION_PATCH}")
> ```

I don't see any difference in the EXPRs that you mentioned.

pfultz2:
> I don't see any difference between the two EXPRs that you mentioned.

One has the version as `${VERSION}.0` and the other is just `${VERSION}`.

nunnikri:
Are we good to approve this change? Or need further changes here?

causten:
Merge conflict and you have not addressed Pauls comments

frepaul:
> > I don't see any difference between the two EXPRs that you mentioned.
> 
> One has the version as `${VERSION}.0` and the other is just `${VERSION}`.

The convention for actual library files that we have in ROCm across all the component is to have major.minor.{encoded ROCm version}.
rocm_set_soversion need to pass only major.minor. {encoded ROCm version} is handled by the function and attached as the patch number and ROCm version is passed by the build CI. Hope this clarify why we do not need the patch number passed here.

umangyadav (approved):

PR #2272, opened 2023-10-02
Reset HIP status using `hipGetLastError()` 

umangyadav:
Runtime is making change such that `hipGetLastError()` would return error of the first failed HIP call not the last. 

This PR makes changes inside MIGraphX to comply with that.

### DIFF
--- a/src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
+++ b/src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
@@ -81,6 +81,14 @@ inline auto launch(hipStream_t stream, index_int global, index_int local)
         using f_type = decltype(f);
         dim3 nblocks(global / local);
         dim3 nthreads(local);
+        /*
+        hipGetLastError() returns error for the first failed HIP call that happened previously.
+        MIGraphX calls into various backend libraries and failed HIP calls can also happen there.
+        Calling hipGetLastError() would reset error code to hipSuccess, so that inside MIGraphX
+        failed call to hipLaunchKernelGGL() can be captured.
+        */
+        hipError_t flush_call = hipGetLastError();
+        (void)(flush_call);
         // cppcheck-suppress UseDeviceLaunch
         hipLaunchKernelGGL((launcher<f_type>), nblocks, nthreads, 0, stream, f);
         hipError_t kernel_launch_status = hipGetLastError();


src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
```diff
@@ -81,18 +81,25 @@ inline auto launch(hipStream_t stream, index_int global, index_int local)
...
+        hipError_t flush_call = hipGetLastError();
+        (void)(flush_call);
         // cppcheck-suppress UseDeviceLaunch
         hipLaunchKernelGGL((launcher<f_type>), nblocks, nthreads, 0, stream, f);
         hipError_t kernel_launch_status = hipGetLastError();
         if(kernel_launch_status != hipSuccess)
         {
             std::string message = hipGetErrorString(kernel_launch_status);
-            if(not contains(get_targets(), get_device_name()))
+            if(not contains(get_targets(), get_gfx_name()))
```
pfultz2, line 98: If the target ID is different then this message wouldn't show. For example, if this is for gfx908 with xnack on and it was built with xnack off(which is the default), then we wont see the error message.

src/targets/gpu/hip.cpp
```diff
@@ -47,7 +47,14 @@ MIGRAPHX_REGISTER_OP(hip_copy_literal)
 using hip_ptr      = MIGRAPHX_MANAGE_PTR(void, hipFree);
 using hip_host_ptr = MIGRAPHX_MANAGE_PTR(void, hipHostUnregister);
 
-std::string hip_error(int error) { return hipGetErrorString(static_cast<hipError_t>(error)); }
+std::string hip_error(int error)
+{
+    // reset HIP's internal status to hipSuccess
+    auto last_error = hipGetLastError();
+    assert(last_error == error);
```
pfultz2, line 54: I dont think we should have side effects in the function. It should only convert an error id to a string regardless of where it came from.

umangyadav: I don't see it as a side-effect. 

If `hip_error()` is called two times back to back on two unsuccessful calls, it would generate correct error message. 

Without side effect, it can't.

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
umangyadav: we are not specifying `XNACK` and `SRAMECC` flags for the GPU_TARGETS.

src/targets/gpu/hip.cpp
umangyadav: should the side effect be made explicit  then?

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
pfultz2: > we are not specifying XNACK and SRAMECC flags for the GPU_TARGETS.

But users can(and is necessary) when they want to build for xnack off.

src/targets/gpu/hip.cpp
```diff
@@ -47,14 +47,22 @@ MIGRAPHX_REGISTER_OP(hip_copy_literal)
 using hip_ptr      = MIGRAPHX_MANAGE_PTR(void, hipFree);
 using hip_host_ptr = MIGRAPHX_MANAGE_PTR(void, hipHostUnregister);
 
-std::string hip_error(int error) { return hipGetErrorString(static_cast<hipError_t>(error)); }
```
pfultz2, line 50: This function should not be removed.

src/targets/gpu/hip.cpp
```diff
@@ -73,7 +81,7 @@ void* get_device_ptr(void* hptr)
     void* result = nullptr;
     auto status  = hipHostGetDevicePointer(&result, hptr, 0);
     if(status != hipSuccess)
-        MIGRAPHX_THROW("Failed getting device pointer: " + hip_error(status));
+        MIGRAPHX_THROW("Failed getting device pointer: " + hip_get_and_reset_error());
```
pfultz2, line 84: This should still be `hip_error(status)`. We know what the error code is, so we should just use that instead of relying on the `hipLastError`.

src/targets/gpu/hip.cpp
```diff
@@ -47,14 +47,22 @@ MIGRAPHX_REGISTER_OP(hip_copy_literal)
...
+    return std::string(hipGetErrorString(static_cast<hipError_t>(last_error)));
+}
 
 bool is_device_ptr(const void* ptr)
 {
     hipPointerAttribute_t attr;
     auto status = hipPointerGetAttributes(&attr, ptr);
     if(status != hipSuccess)
+    {
+        (void)(hip_get_and_reset_error());
```
pfultz2, line 63: We should only be calling `hipLastError` for launching the kernel since that is the only call which doesnt return an error.

src/targets/gpu/hip.cpp
pfultz2: By explicit, it should be a call made to `hipGetLastError` outside of the `hip_error` function.

src/targets/gpu/device/include/migraphx/gpu/device/launch.hpp
umangyadav: Removed that change.

pfultz2 (approved):

TedThemistokleous (approved):

PR #2276, opened 2023-10-03
Avoid using Template argument deduction for Vector 

umangyadav:
CTAD may not be supported by older C++ runtimes e.g. (gcc-7). 

Seeing this error with `-Werror` on SLES builds : 

```
../../../src/onnx/parse_resize.cpp:323:47: error: 'vector' may not intend to support class template argument deduction [-Werror,-Wctad-maybe-unsupported]
            auto vvv_ind = std::vector(n_dim, std::vector(2, std::vector<size_t>(out_elements)));
                                              ^
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../include/c++/7/bits/stl_vector.h:216:11: note: add a deduction guide to suppress this warning
    class vector : protected _Vector_base<_Tp, _Alloc>
          ^
../../../src/onnx/parse_resize.cpp:323:28: error: 'vector' may not intend to support class template argument deduction [-Werror,-Wctad-maybe-unsupported]
            auto vvv_ind = std::vector(n_dim, std::vector(2, std::vector<size_t>(out_elements)));
                           ^
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../include/c++/7/bits/stl_vector.h:216:11: note: add a deduction guide to suppress this warning
    class vector : protected _Vector_base<_Tp, _Alloc>
          ^

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -281,6 +281,7 @@ jobs:
           -DBUILD_DEV=On \
           -DCMAKE_CXX_COMPILER_LAUNCHER=/usr/local/bin/ccache \
           -DCMAKE_C_COMPILER_LAUNCHER=/usr/local/bin/ccache \
+          -DCMAKE_CXX_FLAGS="-Werror" \
           -DGPU_TARGETS=gfx908 \
           ..
         make -j$(nproc) tests driver
--- a/src/onnx/parse_resize.cpp
+++ b/src/onnx/parse_resize.cpp
@@ -320,7 +320,8 @@ struct parse_resize : op_parser<parse_resize>
 
             // get the number of dimensions
             std::size_t n_dim = out_lens.size();
-            auto vvv_ind = std::vector(n_dim, std::vector(2, std::vector<size_t>(out_elements)));
+            std::vector<std::vector<std::size_t>> vv_ind(2, std::vector<std::size_t>(out_elements));
+            std::vector<std::vector<std::vector<std::size_t>>> vvv_ind(n_dim, vv_ind);
             std::vector<std::vector<float>> delta(n_dim, std::vector<float>(out_elements));
 
             shape_for_each(out_s, [&](const auto& out_idx_v, size_t out_idx) {


lakhinderwalia (approved):
(Approved, since you are explicitly stating the template type, but there may be more than one such instance of code.. I wonder, and this isn't upto the expected c++ 17, so definitely not optimal.)

However, please check of that version of compiler (gcc 7.x) has an upgrade for SLES.
And also if some specific compiler option could be turned on..

umangyadav:
>  but there may be more than one such instance of code 

I've enabled `-Werror` to catch those in CI. 

> compiler (gcc 7.x) has an upgrade for SLES 

SLES docker had gcc 7.5. I am not sure when it is due for an upgrade in QA or DevOps' docker. But I agree that it should be upgraded, it is old.

lakhinderwalia:
> SLES docker had gcc 7.5. I am not sure when it is due for an upgrade in QA or DevOps' docker. But I agree that it should be upgraded, it is old.

I think gcc 7.x supports CTAD. This warning might be worth ignoring, as an alternative.

pfultz2 (approved):

pfultz2:
>  I am not sure when it is due for an upgrade in QA or DevOps' docker. But I agree that it should be upgraded, it is old.

We need to support the C++ runtime that applications are using which will be the C++ runtime used by the distro, in the case for sles its gcc 7. If we do upgrade the version it needs to use the one provided by the distro that is binary compatible with gcc 7 runtime. I do know rhel provides newer gcc versions, but I dont know about sles.

pfultz2:
> I think gcc 7.x supports CTAD. 

It does, but the C++ runtime doesn't provide the deduction guidelines.

> This warning might be worth ignoring, as an alternative.

We shouldn't ignore this warning, This is useful in diagnosing hard to figure out bugs due to a simple typo:

```cpp
template<class It>
bool is_palindrome(It first, It last) {
    return std::equal(
        first, last,
        // Typo, should be std::make_reverse_iterator
        std::reverse_iterator(last),
        std::reverse_iterator(first)
    );
}
```

In general for container and container-like types(like `std::tuple`) it best to avoid ctad as well since it breaks generic code.

PR #2277, opened 2023-10-03
port class 'process' to Windows

apwojcik:


src/CMakeLists.txt
```diff
@@ -104,6 +103,11 @@ add_library(migraphx
     value.cpp
     verify_args.cpp
 )
+if(WIN32)
+    target_sources(migraphx PRIVATE process_win32.cpp)
```
pfultz2, line 107: Dont create another source file, just add an ifdef to the process.cpp file.

src/process_win32.cpp
```diff
@@ -0,0 +1,213 @@
...
+// cppcheck-suppress definePrefix
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_CMD_EXECUTE)
+
+#define MIGRAPHX_PROCESS_BUFSIZE 4096
```
pfultz2, line 39: Use a const variable.

pfultz2:
Why arent you using `_popen` function, which is the equivalent for windows? I would assume we could just write something like this:

```cpp
#ifdef WIN32
#define MIGRAPHX_PCLOSE _pclose
#define MIGRAPHX_POPEN _popen
#else
#define MIGRAPHX_PCLOSE pclose
#define MIGRAPHX_POPEN popen
#endif

template <class F>
int exec(const std::string& cmd, const char* type, F f)
{
    int ec = 0;
    if(enabled(MIGRAPHX_TRACE_CMD_EXECUTE{}))
        std::cout << cmd << std::endl;
    auto closer = [&](FILE* stream) {
        auto status = MIGRAPHX_PCLOSE(stream);
#ifdef WIN32
        ec = status < 0 : -1 : 0;
#else
        ec = WIFEXITED(status) ? WEXITSTATUS(status) : 0; // NOLINT
#endif
    };
    {
        // TODO: Use execve instead of popen
        std::unique_ptr<FILE, decltype(closer)> pipe(MIGRAPHX_POPEN(cmd.c_str(), type), closer); // NOLINT
        if(not pipe)
            MIGRAPHX_THROW("popen() failed: " + cmd);
        f(pipe.get());
    }
    return ec;
}
```

src/process_win32.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+        LPSTR lpCmdLn{const_cast<LPSTR>(cmd.c_str())};
+
+        BOOL bSuccess = CreateProcess(
+            nullptr, lpCmdLn, nullptr, nullptr, TRUE, 0, nullptr, nullptr, &info, &processInfo);
+
+        if(bSuccess == FALSE)
+            return GetLastError();
+
+        DWORD dwRead, dwWritten;
```
pfultz2, line 119: You need to run clang tidy on this. These variables do not match the naming scheme we use.

src/process_win32.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+process& process::cwd(const fs::path& p)
+{
+    impl->cwd = p;
+    return *this;
+}
+
+void process::exec() { impl->check_exec(impl->get_command()); }
+
+void process::write(std::function<void(process::writer)> pipe_in)
```
pfultz2, line 207: The `pipe_in` variable is not used.

src/process_win32.cpp
```diff
@@ -0,0 +1,213 @@
...
+    public:
+    explicit pipe(bool inherit_handle = true)
+    {
+        SECURITY_ATTRIBUTES attrs;
+        attrs.nLength              = sizeof(SECURITY_ATTRIBUTES);
+        attrs.bInheritHandle       = inherit_handle ? TRUE : FALSE;
+        attrs.lpSecurityDescriptor = nullptr;
+
+        if(CreatePipe(&hRead_, &hWrite_, &attrs, 0) == FALSE)
+            throw GetLastError();
```
pfultz2, line 52: `MIGRAPHX_THROW` should be used.

src/process_win32.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+        if(SetHandleInformation(&hRead_, HANDLE_FLAG_INHERIT, 0) == FALSE)
+            throw GetLastError();
+    }
+
```
pfultz2, line 57: Need to add a move constructor, delete dopy constructor and add an assignment operator.

apwojcik:
The [MSDN documentation](https://msdn.microsoft.com/en-us/library/96ayss4b.aspx) for _popen states:
If used in a Windows program, the _popen function returns an invalid file pointer that causes the program to stop responding indefinitely. _popen works properly in a console application. To create a Windows application that redirects input and output, see [Creating a child process with redirected input and output](https://learn.microsoft.com/en-us/windows/win32/ProcThread/creating-a-child-process-with-redirected-input-and-output) in the Windows SDK.

apwojcik:
More `_popen` creates a pipe and requires translation from POSIX to WIN32 and vice versa. That is a performance penalty we want to avoid, considering the UAI project requirements.

src/CMakeLists.txt
apwojcik: Ok, I will move it to the `process.cpp`. I thought you asked me to do that in the Draft PR, but I might mistaken that with someone else asking me from a different component.

src/process_win32.cpp
apwojcik: I am catching that in the `exec()` function to return the correct status code. That exception is not propagated further up in the call stack.

src/process_win32.cpp
apwojcik: I will. I thought I did. This is Hungarian notation, the WIN32 API notation, and the most seen when someone is writing WIN32 API applications.

pfultz2:
> If used in a Windows program, the _popen function returns an invalid file pointer that causes the program to stop responding indefinitely. _popen works properly in a console application.

I see, it would be better to use [tiny-process-library](https://gitlab.com/eidheim/tiny-process-library) for this. The `popen` has limitations on POSIX side as well(ie no bidirectional pipes). This probably doesnt need to be switched in this PR, though.

pfultz2 (approved):

PR #2278, opened 2023-10-03
port class 'dynamic_loader' to Windows

apwojcik:


src/CMakeLists.txt
```diff
@@ -104,6 +103,11 @@ add_library(migraphx
     value.cpp
     verify_args.cpp
 )
+if(WIN32)
+    target_sources(migraphx PRIVATE dynamic_loader_win32.cpp)
```
pfultz2, line 107: Put this in the same file.

src/dynamic_loader_win32.cpp
```diff
@@ -0,0 +1,92 @@
...
+    dynamic_loader_impl(const fs::path& p, std::shared_ptr<tmp_dir> t = nullptr)
+        : handle(LoadLibrary(p.string().c_str())), temp(std::move(t))
+    {
+        if(handle == nullptr)
+        {
+            MIGRAPHX_THROW("Error loading DLL: " + p.string() + " (" +
+                           std::to_string(GetLastError()) + ")");
+        }
+    }
+
```
pfultz2, line 49: You need to add a move constructor and delete the copy constructor, and add an assignment operator.

src/dynamic_loader_win32.cpp
```diff
@@ -0,0 +1,92 @@
...
+    static std::shared_ptr<dynamic_loader_impl> from_buffer(const char* image, std::size_t size)
+    {
+        auto t = std::make_shared<tmp_dir>("migx-dynload");
+        auto f = t->path / "tmp.dll";
+        write_buffer(f.string(), image, size);
+        return std::make_shared<dynamic_loader_impl>(f, t);
+    }
+
+    HMODULE handle                = nullptr;
+    std::shared_ptr<tmp_dir> temp = nullptr;
```
pfultz2, line 66: Since the class is not copyable, then there is no reason to use `shared_ptr` here.

src/include/migraphx/dynamic_loader.hpp
```diff
@@ -46,6 +47,7 @@ struct MIGRAPHX_EXPORT dynamic_loader
     static fs::path path(void* address);
 
     static optional<dynamic_loader> try_load(const fs::path& p);
+#endif
```
pfultz2, line 50: Why is this disabled on windows?

src/dynamic_loader.cpp
```diff
@@ -81,6 +90,42 @@ fs::path dynamic_loader::path(void* address)
...
+    dynamic_loader_impl(const fs::path& p, tmp_dir t = {})
+        : handle{LoadLibrary(p.string().c_str())}, temp{std::move(t)}
+    {
+        if(handle == nullptr)
+        {
+            MIGRAPHX_THROW("Error loading DLL: " + p.string() + " (" +
+                           std::to_string(GetLastError()) + ")");
+        }
+    }
+
```
pfultz2, line 107: Make sure to delete the copy constructor and add a an assignment operator.

src/include/migraphx/dynamic_loader.hpp
apwojcik: For some reason, on Windows, static analysis is showing the `catch` part of the function is unreachable code.  I made it available on Windows, too.

causten (changes requested):
There is a cpp check failure that needs to be addressed

causten (approved):

PR #2279, opened 2023-10-03
remove disabling execptions from production build

apwojcik:
Also, on Windows, we need to move the function `migraphx_test_private_disable_exception_catch()` from the `migraphx` namespace to the global namespace.


CMakeLists.txt
```diff
@@ -261,11 +261,13 @@ rocm_enable_cppcheck(
         MIGRAPHX_USE_CLANG_TIDY
 )
 
-enable_testing()
-
 include(ROCMCreatePackage)
 include(ROCMTest)
 
+if(BUILD_TESTING)
+    add_compile_definitions($<$<COMPILE_LANGUAGE:C,CXX>:MIGRAPHX_BUILD_TESTING>)
```
pfultz2, line 268: The flags should be added to the target.

umangyadav (approved):
LGTM, address paul's comments though

src/api/api.cpp
```diff
@@ -40,24 +40,29 @@
...
-namespace migraphx {
 
+#ifdef MIGRAPHX_BUILD_TESTING
 static thread_local bool disable_exception_catch = false; // NOLINT
 
 extern "C" MIGRAPHX_C_EXPORT void migraphx_test_private_disable_exception_catch(bool b)
 {
     disable_exception_catch = b;
 }
+#endif
```
pfultz2, line 51: This should be in the `migraphx` namespace.

apwojcik: The purpose of this PR is to move the entire function out of the `migraphx` namespace.
On Windows, putting the function in the `migraphx` namespace pollutes the API of the `migraphx_c` library (which is supposed to be a pure C API library) with C++ mangled symbols. This results in issues while linking `migraphx_c` to C applications (i.e., `migraphx_c` test) because of the missing `ubsan` library. The `ubsan` library is a Clang undefined behavior sanitizer only available for C++ applications and not for C applications.

pfultz2: > On Windows, putting the function in the migraphx namespace pollutes the API of the migraphx_c library (which is supposed to be a pure C API library) with C++ mangled symbols

We use `extern "C"` on the function so the name shouldn't be mangled. 

> This results in issues while linking migraphx_c to C applications (i.e., migraphx_c test) because of the missing ubsan library. The ubsan library is a Clang undefined behavior sanitizer, and it seems, on Windows, it is only available for C++ applications and not C applications.

I am not sure how this is related, or how this would even fix that issue.

pfultz2 (approved):

PR #2280, opened 2023-10-03
make pybind compilation optional

apwojcik:
UAI does not depend on PyBind, and there are no plans to port that part to Windows (at least at the moment).


CMakeLists.txt
```diff
@@ -53,6 +53,9 @@ include(CTest)
 find_package(ROCM REQUIRED)
 find_package(Threads REQUIRED)
 
+include(CMakeDependentOption)
+cmake_dependent_option(MIGRAPHX_ENABLE_PYTHON "Enable python bindings" ON "WIN32" OFF)
```
pfultz2, line 57: Can we write this as:

```cmake
if(WIN32)
option(MIGRAPHX_ENABLE_PYTHON "Enable python bindings" OFF)
else()
endif()
```

Thats the way do other options.

src/py/CMakeLists.txt
```diff
@@ -22,27 +22,25 @@
 # THE SOFTWARE.
 #####################################################################################
 
-option(MIGRAPHX_ENABLE_PYTHON "Enable python bindings" ON)
+add_compile_definitions($<$<COMPILE_LANGUAGE:CXX,C>:MIGRAPHX_ENABLE_PYTHON>)
```
pfultz2, line 25: This should only be added to the driver: `target_compile_definitions(driver MIGRAPHX_ENABLE_PYTHON=1)`

src/py/CMakeLists.txt
```diff
@@ -22,27 +22,25 @@
...
+    rocm_install_targets(TARGETS migraphx_pybind_${PYTHON_VERSION})
+    add_dependencies(migraphx_py migraphx_pybind_${PYTHON_VERSION})
+
+    add_library(migraphx_py_${PYTHON_VERSION} py.cpp)
+    target_include_directories(migraphx_py_${PYTHON_VERSION} PRIVATE include)
+    target_link_libraries(migraphx_py_${PYTHON_VERSION} PUBLIC migraphx)
+    target_link_libraries(migraphx_py_${PYTHON_VERSION} PRIVATE pybind11::pybind11 python${PYTHON_VERSION}::runtime)
+    rocm_install_targets(TARGETS migraphx_py_${PYTHON_VERSION})
+    add_dependencies(migraphx_py migraphx_py_${PYTHON_VERSION})
+endforeach()
```
pfultz2, line 46: Can you preserve the formatting? Its hard to tell what is changed here.

CMakeLists.txt
apwojcik: The library `migraphx_py` is not entirely compiling on Windows. I want to avoid issues reported about that fact by not exposing the option on Windows. That may change in the future as soon as complete support is introduced. However, at the moment, it is not our priority. Please let me know if you still want me to change it.

src/py/CMakeLists.txt
apwojcik: I moved `if(MIGRAPHX_ENABLE_PYTHON)` ... `endif()` from here to `src/CMakeLists.txt`. I wanted the file preamble to be guarded by that check, too (because of configuration issues on Windows). That is also why the indentation changed. Additionally, I moved `option(MIGRAPHX_ENABLE_PYTHON...)` to master `CMakeLists.txt` so `test/CMakeLists.txt` and `test/py/CMakeLists.txt` can see it and interpret it correctly - before the `test/py/CMakeLists.txt` could not see `migraphx_pybind_3.8` target when Python was enabled.

CMakeLists.txt
pfultz2: We would still want to have an option to enable it, otherwise we wont be able to test this without modifying the cmake.

pfultz2 (approved):

PR #2282, opened 2023-10-03
a few c++ fixes to allow compilation on Windows

apwojcik:
Missing header files (mostly `<array>`), a few preprocessor checks, and a portable way for checking attributes.


src/api/include/migraphx/migraphx.hpp
```diff
@@ -66,7 +66,7 @@ template <class PrivateMigraphTypeNameProbe>
 std::string compute_type_name()
 {
     std::string name;
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
```
umangyadav, line 69: why does it requires check for clang. Aren't both exclusive if it is  MSVC, it can't be clang.

src/include/migraphx/matcher.hpp
```diff
@@ -307,6 +308,12 @@ struct matcher_result
         {
         }
 
+        instruction_container& operator=(const std::unordered_map<std::string, instruction_ref>& x)
```
umangyadav, line 311: is this used anywhere ?

apwojcik: I rechecked, and it turned out that change is not needed anymore. I have updated the PR.

src/api/include/migraphx/migraphx.hpp
apwojcik: The Clang compiler on Windows is running in MSVC compatibility mode. It sets `_MSC_VER`, too. So, only checking `_MSC_VER` is not exclusive and is insufficient to differentiate between native MSVC and Clang. See detailed information https://clang.llvm.org/docs/MSVCCompatibility.html

umangyadav (approved):

src/include/migraphx/config.hpp
```diff
@@ -26,6 +26,10 @@
 
 #include <migraphx/export.h>
 
+#if defined(_MSC_VER) && !defined(__clang__)
+#include <ciso646>
```
pfultz2, line 30: This is safe to include on linux so I dont think we need the ifdef.

test/include/test.hpp
```diff
@@ -333,7 +333,7 @@ enum class color
     bg_blue    = 44,
     bg_default = 49
 };
-inline std::ostream& operator<<(std::ostream& os, const color& c)
+inline std::ostream& operator<<(std::ostream& os, [[maybe_unused]] const color& c)
```
pfultz2, line 336: I would prefer a `(void)c` added under `#ifdef _WIN32`, so we can still get diagnostics for unused variable on the linux side.

test/promote_literals_test.cpp
```diff
@@ -26,6 +26,7 @@
 #include <migraphx/program.hpp>
 #include <migraphx/dead_code_elimination.hpp>
 #include <migraphx/eliminate_common_subexpression.hpp>
+#include <migraphx/instruction.hpp>
```
pfultz2, line 29: We dont use the instruction class here either, at least as far as I can tell. If windows has problems with incomplete types then maybe we should add this header in instruction_ref.hpp under `#ifdef _WIN32`

test/verify/run_verify.cpp
```diff
@@ -44,8 +44,8 @@ MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DUMP_TEST)
 
 // An improved async, that doesn't block
 template <class Function>
-std::future<typename std::result_of<Function()>::type> detach_async(Function&& f,
-                                                                    bool parallel = true)
+std::future<typename std::invoke_result_t<Function>> detach_async(Function&& f,
```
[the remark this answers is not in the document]
tperry-amd: Remove typename since it isn't needed anymore

pfultz2 (approved):

PR #2283, opened 2023-10-03
fix call to std::isfinite() on Windows

apwojcik:
The standard library from MSVC does implement `std::infinite()` for floating-point types only. The function is used internally by many other math functions, such as `std::isnan()`. The C++ specification mentions optional overloads for integer types. However, there are no additional overloads for integer types in the standard library from MSFT, and according to the C++ specification, all integer types should be treated as doubles. So, in the presence of overloaded functions, the overload function casts the integer type to double and calls the floating-point version of a function. Reference: https://en.cppreference.com/w/cpp/numeric/math/isfinite

Long story short: without the `static_cast<>`, the MIGraphX library is not compiling on Windows.


umangyadav (approved):

pfultz2 (approved):

TedThemistokleous (approved):

src/program.cpp
```diff
@@ -347,7 +347,7 @@ void program::finalize()
 template <class T>
 std::string classify(T x)
 {
-    switch(std::fpclassify(x))
+    switch(std::fpclassify(static_cast<double>(x)))
```
lakhinderwalia, line 350: This breaks for a subnormal float conversion to double. It won't stay subnormal anymore. Thanks.

PR #2284, opened 2023-10-03
fix compilation of uniform int distribution

apwojcik:
See the comment in the header file and https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution.
The standard library from Microsoft Visual C++ is strict about that - it fails compilation when `unsigned char` and `signed char` are used as deduced type/template parameters.
The Clang compiler on Windows uses the standard library from MSFT Visual C++.


src/include/migraphx/op/random_uniform.hpp
```diff
@@ -75,7 +75,16 @@ struct random_uniform
...
             using type = typename decltype(output)::value_type;
-            if constexpr(std::is_integral<type>{})
+            if constexpr(
+                std::is_integral_v<type>
+#ifdef _MSC_VER
+                // According to the C++ specification, the effect is undefined if the result type
+                // for the generator is not one of short, int, long, long long, unsigned short,
+                // unsigned int, unsigned long, or unsigned long long. See
+                // https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution.
+                && !(std::is_same_v<type, unsigned char> || std::is_same_v<type, signed char>)
```
umangyadav, line 85:
```suggestion
                && !(std::is_same_v<type, uint8_t> || std::is_same_v<type, int8_t>)
```
This should work i think

umangyadav:
This change puts restriction on what data types random uniform operator can support and should better be checked at `compute_shape()`

src/include/migraphx/op/random_uniform.hpp
umangyadav: So it will evaluate to false for int8 and uint8 and would try to generate real random numbers. 

This requires some more changes.

pfultz2, line 85: This will produce wrong results for int8 on windows. When we are using an unsupported type then we need to use a supported one and then clamp it to the supported:

```cpp
template<class T>
using uniform_int_distribution = std::uniform_int_distribution<std::conditional_t<sizeof(T) == 1, int, T>>;
```

And then clamp it when we construct it:

```cpp
uniform_int_distribution<type> dis(0, std::numeric_limits<type>::max());
std::generate(output.begin(), output.end(), [&] { return dis(gen); });
```

apwojcik: That make sense. I will try it and let you know the result.

umangyadav: > Then we need to use a supported one and then clamp it to the supported: 

That would violate how a random uniform should work.  It should produce uniformly distributed numbers across the range. 

I think we can just not support it for int8 for now. ONNX only supports for real values.

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -75,7 +75,16 @@ struct random_uniform
 
         result.visit([&](auto output) {
             using type = typename decltype(output)::value_type;
-            if constexpr(std::is_integral<type>{})
+            if constexpr(
+                std::is_integral_v<type>
```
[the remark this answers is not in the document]
apwojcik: Why do we need to instantiate the static check? That is a compile-time property check of the static-if and not a runtime check. `std::is_integral_v<>` evaluates to `std::integral<>::value`, which is `true` for integral types and `false` for others. Honestly, I have not seen someone using static checks that way.

src/include/migraphx/op/random_uniform.hpp
pfultz2: > That would violate how a random uniform should work. It should produce uniformly distributed numbers across the range.

Yea "clamping" was probably a poor choice of words.

> or otherwise we would need to scale it down to range of int8 and then quantize/round it to nearest int

We dont need to do that, we can pass the range we want to the constructor. This should produce the same results.

Originally we were relying on the default constructor to set the range based on the data type, but since we are using a different generator type then we need to pass the range to the constructor.

src/include/migraphx/op/random_uniform.hpp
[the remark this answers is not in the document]
pfultz2: > Why do we need to instantiate the static check?

You are instantiating the template in both cases.

1. By constructing it we get an `integral_constant` instead of bool which is more powerful in some cases(such as passing this to an intermediate function).
2. Not all type traits provide variable templates, so we need to write it like that
3. Consistency. We shouldnt need to think about whether we will pass it to a function or whether it supports a variable template. We write it the same way everytime.

src/include/migraphx/op/random_uniform.hpp
bpickrel: I strongly recommend not trying to support anything not already supported by STL or other libraries.  There will be a mess of special and edge cases to consider, and any shortcoming will create subtle statistical biases that could turn into problems for our future customers at runtime.  Coding statistical library functions is a specialized field that we do not want to dabble in.

bpickrel (commented):
Can add a check here if the input type is one of the disallowed ones (int8, etc.) and throw an exception.

pfultz2:
> Can add a check here if the input type is one of the disallowed ones (int8, etc.) and throw an exception.

No, our reference implementation should support all types.

src/include/migraphx/op/random_uniform.hpp
bpickrel: A little more reading in StackOverflow indicates that the C++ committee never gave a mathematical reason for disallowing int8 types for `uniform_int_distribution`.  So it could be implemented by instantiating `uniform_int_distribution` with range arguments(0, 255) in the contructor and then assigning to the desired type.

umangyadav (approved):

src/include/migraphx/op/random_uniform.hpp
```diff
@@ -75,7 +75,20 @@ struct random_uniform
 
         result.visit([&](auto output) {
             using type = typename decltype(output)::value_type;
-            if constexpr(std::is_integral<type>{})
+#ifdef _MSC_VER
+            // According to the C++ specification, the effect is undefined if the result type
+            // for the generator is not one of short, int, long, long long, unsigned short,
+            // unsigned int, unsigned long, or unsigned long long. See
+            // https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution.
+            if constexpr(std::is_same_v<type, unsigned char> || std::is_same_v<type, signed char>)
```
pfultz2, line 83: This could be written as `std::is_integral<type>{} and sizeof(type) == 1`, which I think is simpler. 

If not, then it should be `std::is_same<type, unsigned char>{} or std::is_same<type, signed char>{}` to match the style used throughout migraphx.

bpickrel (approved):

PR #2288, opened 2023-10-04
python: add 'make generate' OS agnostic

apwojcik:
The Python version of 'make generate' for Linux and Windows.


tools/CMakeLists.txt
```diff
@@ -22,4 +22,56 @@
...
+            COMMAND ${Python_EXECUTABLE} api.py ${__migraphx_py_native_path} ${__input_native_path} | ${CLANG_FORMAT} -style=file > ${__output_native_path}
+            VERBATIM
+            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+            DEPENDS ${__input_path}
+        )
+    cmake_path(RELATIVE_PATH __output_path BASE_DIRECTORY ${SRC_DIR} OUTPUT_VARIABLE __rel_name)
+    string(MAKE_C_IDENTIFIER ${__rel_name} __cid)
+    add_custom_target(generate_${__cid} DEPENDS ${__output_native_path})
+    add_dependencies(generate generate_${__cid})
+endfunction()
```
pfultz2, line 74: Can this be written using python instead of cmake? Python is much easier to read. I would like a generate.py script to do the same thing as the generate.sh script.

apwojcik: You asked us to do Python or CMake instead of a batch script in PR #1893. We chose CMake because it is more accessible on Windows regarding build configuration and has all the infrastructure for searching for external tools.

tools/generate.py
```diff
@@ -0,0 +1,89 @@
...
+                          shell=True,
+                          check=True,
+                          input=buffer.encode('utf-8'),
+                          cwd=work_dir,
+                          **kwargs).stdout.decode('utf-8')
+
+
+def api_generate(input_path: Path, output_path: Path):
+    with open(output_path, 'w') as f:
+        f.write(clang_format(api.invoke(input_path)))
```
pfultz2, line 46: Just call the `run` function directly: `f.write(clang_format(api.invoke([migraphx_py_path, input_path])))` then there is no need to call `register_functions`.

tools/api.py
```diff
@@ -678,6 +679,10 @@ def add_function(name: str, *args, **kwargs) -> Function:
     return f
 
 
+def register_functions(path: Union[Path, str]) -> None:
```
pfultz2, line 682: This doesnt need to be moved out into a separate function.

tools/te.py
```diff
@@ -431,6 +431,9 @@ def template_eval(template, **kwargs):
     return template
 
 
-f = open(sys.argv[1]).read()
-r = template_eval(f)
-sys.stdout.write(r)
+def invoke(p):
```
pfultz2, line 434: Call this `run` instead of `invoke`.

tools/generate.py
apwojcik: I tried that. It fails with an exception after the second call, saying `"updated() not being called"`. The second call doubles the functions already on the `functions` list, and `@once` attached to `process_functions()` prevents them from being updated. When I removed the decorator/reset the `had_run` flag, I ended up doubling everything: functions, parameters, structures, etc., in the resulting files. So, either I register functions once or drop everything after each call to `api.invoke()`. I decided to register functions once.

apwojcik:
The `cppcheck` is failing not because of this PR but because the other PR merged before. I fixed the check in #2278

PR #2289, opened 2023-10-05
Add support for IsInf ONNX operator

gyulaz-htec:
This PR adds support for IsInf ONNX operator.

Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/99

I had to add parse logic becuase the [operator](https://github.com/onnx/onnx/blob/main/docs/Operators.md#IsInf) can distinguish positive and negative infinity values based on the `detect_negative` and `detect_positive` attributes. Just adding an unary operator wrapping `std::isinf` can't cover these cases.


test/ref/isinf.cpp
```diff
@@ -0,0 +1,91 @@
...
+        mm->add_instruction(migraphx::make_op("isinf"), l1);
+        p.compile(migraphx::make_target("ref"));
+        auto result = p.eval({}).back();
+        std::vector<float> results_vector;
+        result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+        std::vector<float> gold = {0, 0, 1, 1, 0, 0};
+        EXPECT(migraphx::verify::verify_rms_range(results_vector, gold));
+    }
+
+    // half test
```
umangyadav, line 52: can you create a separate test case for this one instead of putting both float and half into one ?

umangyadav (commented):
Looks good. address minor comments

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
```diff
@@ -101,6 +101,7 @@ MIGRAPHX_DEVICE_MATH(erf, ::erf)
 MIGRAPHX_DEVICE_MATH(exp, ::exp)
 MIGRAPHX_DEVICE_MATH(floor, ::floor)
 MIGRAPHX_DEVICE_MATH(isnan, ::isnan)
+MIGRAPHX_DEVICE_MATH(isinf, ::isinf)
```
umangyadav, line 104: For the GPU implementation need to add tests inside `test_verify.cpp`. Can you add a few please ?

test/onnx/gen_onnx.py
```diff
@@ -3828,6 +3828,64 @@ def instance_norm_val_3d_test():
...
+
+    node = onnx.helper.make_node(
+        'IsInf',
+        inputs=['t1'],
+        outputs=['t2'],
+    )
+    return ([node], [t1], [t2])
+
+
+@onnx_test()
```
lakhinderwalia, line 3844: I like your fp16 tests, but a couple of them could test instead for float & double -- you don't have to create additional tests, if you would like to change some of these.. 
(Optional)

test/onnx/verify_onnx.cpp
```diff
@@ -1014,6 +1014,82 @@ TEST_CASE(instance_norm_3d_test)
...
 }
 
+TEST_CASE(isinf_half_test)
+{
+    migraphx::program p = migraphx::parse_onnx("isinf_half_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape s{migraphx::shape::half_type, {2, 3}};
+    migraphx::parameter_map pp;
+    std::vector<float> tmp = {-1.7, NAN, INFINITY, 3.6, -INFINITY, INFINITY};
```
lakhinderwalia, line 1024: Suggestion: Instead of INFINITY, please use a synthesized value. That way we can catch any run time FP results that would bypass FP-units if compiler option like --fast-math is turned on. And also add a comment why we are doing it this way.. Thanks.

example: 
float x = std::numeric_limits<float>::max();
float infinity = x * x; 
(And then use this to fill tmp)

src/onnx/parse_isinf.cpp
```diff
@@ -0,0 +1,99 @@
...
+        if(detect_negative and detect_positive)
+        {
+            return is_inf;
+        }
+
+        auto zero_l = info.add_literal(migraphx::literal{migraphx::shape{x_shape.type()}, {0}});
+        auto mb_zero =
+            info.add_instruction(make_op("multibroadcast", {{"out_lens", x_shape.lens()}}), zero_l);
+
+        if(detect_negative)
```
lakhinderwalia, line 74: The only difference in these if/else is the use of binary op "less" vs "greater". Lot of common code.. Thanks.

lakhinderwalia: Architectural question. Should options `detect_negative`, `detect_positive` be implemented in is_inf.hpp? (Default: true). Thanks.

pfultz2: > Architectural question. Should options `detect_negative`, `detect_positive` be implemented in is_inf.hpp? (Default: true). Thanks.

No we shouldn't as this complicates code gen/analysis and we already have the operators available to implement this after the std::isinf operator is implemented.

test/onnx/verify_onnx.cpp
pfultz2: I think it's better to explicitly use `NAN` and `INFINITY`. The purpose is to check the isinf operator not to check if --fast-math is enabled.

Alternatively `std::numeric_limits` could be used to get these values as well.

src/onnx/parse_isinf.cpp
lakhinderwalia: Reason for the suggestion: If one of: `detect_negative` or `detect_positive` is set, it currently generates additional 'less' or 'greater' operators in the composite graph. This can easily be subsumed in is_inf.hpp, for optimization, to avoid this additional layer.

test/onnx/verify_onnx.cpp
lakhinderwalia: > I think it's better to explicitly use `NAN` and `INFINITY`. The purpose is to check the isinf operator not to check if --fast-math is enabled.


Better how? The idea is not to check if fast-math is enabled. The pain of not finding FP failure early is the issue. The idea is to fail, if FP guarantees can't me made -- this can be only found at run-time by generating infinity. Note: NAN is not an issue.  Thanks.

My suggestion is because we are specifically testing failure related to Infinity. Normally I would also use INFINITY. The proper test is to do it by generating infinity.

src/onnx/parse_isinf.cpp
pfultz2: These operators will all be fused into the same kernel.

test/ref/isinf.cpp
gyulaz-htec: Separated them in the latest version.

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
gyulaz-htec: I've added those test cases

test/onnx/gen_onnx.py
gyulaz-htec: I've changed  some of them to test float and double instead.

test/onnx/verify_onnx.cpp
gyulaz-htec: Since there was no consensus on this discussion, I've added bot `std::numeric_limit` and the proposed synthesized inf values.

gyulaz-htec:
@pfultz2 @umangyadav @lakhinderwalia I've addresed most of you requests, except the [architectural](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2289#pullrequestreview-1693682890) related one, because there are different opinions about that.

test/verify/test_isinf_float.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#include <limits>
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+
+struct test_isinf_float : verify_program<test_isinf_float>
```
umangyadav, line 30: You can templatize this and use it for both half and float since both tests are same.

gyulaz-htec: Replaced them with a templated version.

umangyadav (approved):
Thanks, LGTM.

pfultz2 (approved):

test/onnx/verify_onnx.cpp
lakhinderwalia: Thanks.

lakhinderwalia (approved):
Thanks for doing the changes.

test/onnx/verify_onnx.cpp
```diff
@@ -1014,6 +1014,103 @@ TEST_CASE(instance_norm_3d_test)
...
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape s{migraphx::shape::half_type, {2, 3}};
+    migraphx::parameter_map pp;
+    float nan      = std::numeric_limits<float>::quiet_NaN();
+    float infinity = std::numeric_limits<float>::infinity();
+    float x        = std::numeric_limits<float>::max();
+    // Use a synthesized value to catch any run time FP results
+    // that would bypass FP-units if compiler option like --fast-math is turned on
+    float synth_infinity   = x * x;
```
pfultz2, line 1029: Dont use a synthesized value here since according to @lakhinderwalia this may not produce an infinity value.

pfultz2 (changes requested):

test/onnx/verify_onnx.cpp
lakhinderwalia: @pfultz2: that is the intent. So the unit test should fail. 

This is exactly how those values that are produced in a graph during computation may actually not be INFINITY, and lsinf() would give false results.

pfultz2: The unit test should only fail if we are not calling isinf correctly.

Checking if a synthesized value is infinity, should be done as a separate unit test although I don't see how it would be useful. We don't run these tests when running the model. And if it produced a different value for infinity it may not even effect the model accuracy either.

lakhinderwalia: > The unit test should only fail if we are not calling isinf correctly.

We are in agreement. Hence my suggestions. Thanks.

pfultz2: But if you use a synthesized value then the test can fail because the synthesized value is not infinity,  not because we are calling isinf incorrectly, which is why we shouldn't use a synthesized value for this test.

gyulaz-htec: For the record, those synthesized values are all passing as infinite values in each tests on the CI.

gyulaz-htec:
@pfultz2 @lakhinderwalia I've removed the synthesized values from the tests, even though those values were recognized as infinite values in the tests. For me that seemed like we want to test the cpp implementation rather than the corerctness of the operator. Also the discussions in this PR highlighted that those values not always should produce infinite values.

pfultz2 (approved):

umangyadav (approved):

PR #2297, opened 2023-10-05
QlinearGlobalAveragePool operator

lakhinderwalia:
https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.QLinearGlobalAveragePool


src/onnx/parse_qlinearglavgpool.cpp
```diff
@@ -0,0 +1,176 @@
...
+QLinearGlobalAveragePool consumes an input tensor X and applies
+Average pooling across the values in the same channel. This is
+equivalent to AveragePool with kernel size equal to the spatial
+dimension of input tensor. Input is of type uint8_t or int8_t.
+
+Version
+This version of the operator has been available since version 1 of the 'com.microsoft' operator set.
+
+Attributes
+channels_last : int
```
umangyadav, line 51: need to parse channels_last attribute and take that into consideration

src/onnx/parse_qlinearglavgpool.cpp
```diff
@@ -0,0 +1,176 @@
...
+            MIGRAPHX_THROW("QLINEARGLOBALAVERAGEPOOL: mismatched type: input zero point");
+
+        if(type_x != zero_pt_y->get_shape().type())
+            MIGRAPHX_THROW("QLINEARGLOBALAVERAGEPOOL: mismatched type: output zero point");
+    }
+
+    // This method is to prep for quantizelinear or dequantizelinear operation for
+    // either the broadcasting of weight-scale or zero-points of an operator
+    // outputs: operator op (inputs x, broadcasted: scale (float) & zero_pt (8-bit))
+    instruction_ref bcast_qdq_instr(const std::string& op_name,
```
umangyadav, line 113: If you make separate hpp file for utility methods, move this there

src/onnx/parse_qlinearglavgpool.cpp
umangyadav: Would be good add `onnx_test` for parsing separately to catch such things.

src/onnx/parse_qlinearglavgpool.cpp
lakhinderwalia: (Yes, of course, that is the plan for all these quant operators being coded in parallel ..) Thanks.

src/onnx/parse_qlinearglavgpool.cpp
lakhinderwalia: > need to parse channels_last attribute and take that into consideration

Yes, it will be. This implementation is handling right now the common case. 
I need to verify if our pooling operator can handle channels_last. 

Thanks.

lakhinderwalia: > Would be good to add test inside `onnx_test` for testing parsing separately to catch such things.

I haven't added a parsing test to onnx_test yet. And it will be added later (soon).

lakhinderwalia: - channels_last attribute is taken into account.
- onnx_test has a test: qlinearglobalavgpool_test.

(Thanks)

src/onnx/parse_qlinearglavgpool.cpp
```diff
@@ -0,0 +1,151 @@
...
+Outputs
+Y : T
+Output data tensor from pooling across the input tensor. The output tensor has the same rank as the
+input. with the N and C value keep it value, while the otherdimensions are all 1. Type Constraints
+T : tensor(uint8), tensor(int8)
+Constrain input and output types to signed/unsigned int8 tensors.
+
+*/
+
+struct parse_qlinearglobalaveragepool : op_parser<parse_qlinearglobalaveragepool>
```
pfultz2, line 83: Maybe we should just name this parser as `parse_qlinearpool` so it can be used in the future for other types of pooling.

umangyadav (approved):

src/onnx/parse_qlinearglavgpool.cpp
lakhinderwalia: Paul, (agreed for the future steps..) but, for expediency, I suggest we do it for the next quantized pooling operator. Else, it would mean adding code for operators() etc., I believe, and I will have to verify the code again.. with no additional value add for now. Thanks.

src/onnx/parse_qlinearglavgpool.cpp
```diff
@@ -0,0 +1,151 @@
...
+y_scale : tensor(float)
+Scale of quantized output 'Y'. It must be a scalar.
+
+y_zero_point : T
+Zero point tensor for output 'Y'. It must be a scalar.
+
+Outputs
+Y : T
+Output data tensor from pooling across the input tensor. The output tensor has the same rank as the
+input. with the N and C value keep it value, while the otherdimensions are all 1. Type Constraints
```
kahmed10, line 77:
```suggestion
input. with the N and C value keep it value, while the other dimensions are all 1. Type Constraints
```

kahmed10 (approved):

PR #2298, opened 2023-10-05
Set compile options from accuracy checker

CharlieL7:
Importantly allows you to set disable_fast_math from the accuracy checker.

### DIFF
--- a/tools/accuracy/accuracy_checker.py
+++ b/tools/accuracy/accuracy_checker.py
@@ -82,6 +82,27 @@ def parse_args():
                         default=False,
                         help='Turn on ort VERBOSE logging via session options')
 
+    parser.add_argument(
+        '--disable-offload-copy',
+        dest="offload_copy",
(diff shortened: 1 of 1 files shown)


tools/accuracy/accuracy_checker.py
```diff
@@ -82,6 +82,27 @@ def parse_args():
                         default=False,
                         help='Turn on ort VERBOSE logging via session options')
 
+    parser.add_argument(
+        '--disable-offload-copy',
```
umangyadav, line 86: offload copy by default is false. this should be enable_offload_copy

pfultz2: Its on by default for the python API.

umangyadav: Oh right

umangyadav (approved):

TedThemistokleous (approved):

PR #2299, opened 2023-10-05
[mlir] Add assertion to prevent sending MLIR dynamic shapes

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1105

This code shouldn't be reached, since dynamic shapes ought to be rewritten away way before MLIR lowering ... but it never hurts to be extra careful.


src/targets/gpu/mlir.cpp
```diff
@@ -320,7 +320,8 @@ struct mlir_program
 
     MlirType make_tensor(const shape& s) const
     {
-        assert(s.standard());
+        assert(s.standard() && "MLIR expects all tensors to be in standard shape");
+        assert(not s.dynamic() && "MLIR does not support dynamic shapes");
```
umangyadav, line 324: Use MIGRAPHX_THROW to be more consistent ? Assert would only work for debug  builds

krzysz00: ... yeah, good point, those should be throws

umangyadav (approved):

pfultz2 (approved):

PR #2300, opened 2023-10-05
Run optimize_module for int8 quantization

pfultz2:
This should improve performance and accuracy.


umangyadav (approved):

TedThemistokleous (approved):
Approving this as I believe this will effect changes on the EP side/accuracy once I start testing.

causten (approved):

PR #2301, opened 2023-10-05
Add explanatory documentation to `allocate`

CharlieL7:
Explains the two versions of allocate


umangyadav (approved):

src/include/migraphx/op/allocate.hpp
```diff
@@ -33,6 +33,19 @@ namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace op {
 
+/**
+ * Static allocate:
```
bpickrel, line 37: 1.  You made a reference to GPU instructions during the meeting, which was news to me.  Could you explain that here?
2.  I'd like to see an explanation of use cases for this op.  You've told me in the past that Allocate was the wrong op. when I tried to use it;  how does the model programmer know when to use it?  For starters, when the size to allocate is not known til runtime--what else?
3.   I see this as really having 3 behavior cases, not 2.  They are:
-  1 input and shape attr.
-  No inputs and shape attr.
-  1 input, no shape attr.
4.  I'm still not clear just what happens when there is no input argument.
5.

bpickrel: Also, what is the use case for not giving a buffer type?

CharlieL7: 1. The gpu version of allocate is in gpu/allocation_model.cpp which calls hip::hip_allocate_memory. The allocates are replaced in the replace_allocate pass for gpu. This operator would probably not the appropriate place to describe that process.
2. Either you're using allocate with a static shape attribute to output a static buffer or you're using the one input dynamic version that needs to take in output dimensions that allocates the buffer at runtime. I mentioned that allocate was the wrong op because the usecase you showed didn't use the allocated buffer.
3. There are three ways to use the op, but it essentially between making a buffer at compile time or at runtime. The one input versions are the only different in the compute_shape() function.
4. No input argument means it only uses the shape attribute for both compute_shape() and compute(). That shape attribute must be a static shape. This version of the operator is commonly used for the output parameters for gpu operators in gpu/lowering.
5. Having the output dimensions input and the shape attribute set means you want a specific output shape to compute_shape() at compile time. This is used when you know the dynamic output shape that the compute_shape() should output. The benefit of this is that the compute_shape() function will not output a dynamic shape with dynamic_dimensions of {0, max_int}, but rather what you supply to the attribute.

bpickrel (approved):

PR #2190, opened 2023-09-14
Remove contiguous from reshape parsing

TedThemistokleous:
Related to #2099  

Cleaning up parsing reshape which adds additional contiguous

PR #2304, opened 2023-10-06
2 Input Reshape `ref` implementation

CharlieL7:
* Depends on #2190 
* This implements the two input version of ONNX Reshape where the second input is not constant.
* For dynamic batch this 2 input version will be simplified to the one input version after `simplify_dynamic_ops`
  * I still need to create that matcher 
* Note how the 2 input version takes an output_buffer: `reshape(input_tensor, output_buffer)`
  * This is intended to isolate taking output dimensions as input to the `allocate` operator.


TedThemistokleous (approved):
I think some of your changes will be moot once #2190 gets in but overall this makes sense.

Nice comment block in reshape.hpp. Shows your intent with this change well.

test/onnx/gen_onnx.py
```diff
@@ -5793,6 +5793,24 @@ def reshape_non_standard_test():
     return ([trans, res], [x], [y])
 
 
+@onnx_test()
+def reshape_variable_input_test0():
```
umangyadav, line 5797: Why is it named `variable` I don't see any `None` in dimensions.

src/include/migraphx/op/reshape.hpp
```diff
@@ -215,32 +231,51 @@ struct reshape
...
-        return result;
+            visit_all(result, args[0])([&](auto output, auto input) {
+                std::copy(input.begin(), input.end(), output.begin());
+            });
+            return result;
+        }
+        else
+        {
+            // 2 arg
+            assert(args[0].get_shape().elements() == args[1].get_shape().elements());
```
umangyadav, line 273: It should throw instead of assert. If elements do not match then it is a runtime error.

umangyadav: can you add verify_onnx where it would throw?

test/onnx/gen_onnx.py
CharlieL7: Means the second input is variable rather than constant (as in the other tests).

umangyadav (approved):

src/onnx/parse_spacetodepth.cpp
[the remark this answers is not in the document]
CharlieL7: These are changes from #2190

src/include/migraphx/op/reshape.hpp
```diff
@@ -215,32 +231,56 @@ struct reshape
...
-        check_shapes{inputs, *this, true}.has(1);
+        check_shapes{inputs, *this, true}.has(1, 2);
 
         auto n_neg_dims = std::count(dims.begin(), dims.end(), -1);
         if(n_neg_dims > 1)
             MIGRAPHX_THROW("reshape: Dimensions for reshape can only have one -1 dim");
 
         auto s0 = inputs.front();
-        if(s0.dynamic())
+        if(inputs.size() == 1)
```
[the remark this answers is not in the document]
CharlieL7: I don't understand your question in this context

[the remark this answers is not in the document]
bpickrel: my mistake; I misread the condition

bpickrel (approved):

PR #2306, opened 2023-10-08
Parse slice default steps fix

CharlieL7:
* Depends on #2304
* Fixes an issue that comes up for variable input `slice` with steps set manually in ONNX to default 1's.


src/onnx/parse_slice.cpp
```diff
@@ -144,13 +144,18 @@ struct parse_slice : op_parser<parse_slice>
             sd.op.axes = axes;
         }
 
-        if(not sd.steps.empty())
+        if(std::any_of(sd.steps.begin(), sd.steps.end(), [](auto s) { return s != 1; }))
```
umangyadav, line 147:
```suggestion
        if(std::any_of(sd.steps.begin(), sd.steps.end(), [](auto s) { return std::abs(s) != 1; }))
```

src/onnx/parse_slice.cpp
```diff
@@ -144,13 +144,18 @@ struct parse_slice : op_parser<parse_slice>
...
             if(sd.op.starts.empty() or sd.op.ends.empty())
-                MIGRAPHX_THROW("PARSE_SLICE: steps and variable starts and ends is not supported");
+                MIGRAPHX_THROW(
+                    "PARSE_SLICE: steps and variable starts and/or ends is not supported");
             if(sd.op.axes.empty())
                 MIGRAPHX_THROW("PARSE_SLICE: steps and variable axes is not supported");
         }
+        else
+        {
+            sd.steps.clear();
```
umangyadav, line 157: why steps are being cleared ? Is it because it would have all its values as 1 ?

src/onnx/parse_slice.cpp
CharlieL7: Neither `step` nor `reverse` support dynamic shapes yet. Particularly complicated with `step`.

src/onnx/parse_slice.cpp
CharlieL7: Mainly to get around doing the next new lines for a vector of all 1's.

umangyadav: Why do even need  to clear it ? It if falls into `else` clause it would have all values 1.

umangyadav (approved):

kahmed10 (approved):

PR #2308, opened 2023-10-10
QLinearMatMul

lakhinderwalia:
https://onnx.ai/onnx/operators/onnx__QLinearMatMul.html


src/shape.cpp
```diff
@@ -490,6 +490,26 @@ shape shape::with_lens(const std::vector<std::size_t>& l) const
...
+{
+    if(this->dynamic())
+        MIGRAPHX_THROW("SHAPE: with_lens() called on dynamic shape");
+    auto cur_lens = this->lens();
+    std::vector<std::size_t> nlens(1 + cur_lens.size(), dim);
+    std::copy(cur_lens.begin(), cur_lens.end(), nlens.begin() + 1);
+    return {this->type(), nlens};
+}
+
+shape shape::append_dim(size_t dim) const
```
umangyadav, line 503: Need to add small unit-tests inside shape_test.cpp

src/onnx/parse_qlinearmatmul.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+        if(dim_a == 0 or dim_b == 0)
+            MIGRAPHX_THROW("QLINEARMATMUL: empty input");
+
+        // broadcast supported if either is 1-D -- the other can be a 2-D tensor.
+        if(dim_a == 1)
+        {
+            sh_a   = sh_a.prepend_dim(1);
+            lens_a = sh_a.lens();
+            dim_a  = lens_a.size();
```
umangyadav, line 128: You can use `shape::ndim()`

src/onnx/parse_qlinearmatmul.cpp
```diff
@@ -0,0 +1,199 @@
...
+        const auto& in_b = args[3];
+
+        auto sh_a = in_a->get_shape();
+        auto sh_b = in_b->get_shape();
+
+        auto type_a = sh_a.type();
+        auto type_b = sh_b.type();
+        if(type_a != migraphx::shape::int8_type and type_a != migraphx::shape::uint8_type)
+            MIGRAPHX_THROW("QLINEARMATMUL: unsupported input type");
+        if(type_b != migraphx::shape::int8_type and type_b != migraphx::shape::uint8_type)
```
umangyadav, line 111: Onnx spec for some reason doesn't enforce both types to be equal. MIGraphX quant_dot op requires that.

lakhinderwalia: Yes, they don't have to be the same type actually. (That is why these two lines deal with type_a and type_b separately)

It is more those operators which mandate that their quantized binary operands be the same type T, I would like to understand why -- maybe efficiency..

umangyadav: > I would like to understand why -- maybe efficiency.. 
I think it could be just an error on ONNX spec doc. You can try generating onnx test with different type and see if throws any error. 

i suspect if they are different then it is not possible to implement because the accumulator needs to be int32_t to allow for signed integers but then if one dtype is int8 and the other one is uint8 then it can possibly overflow "faster".

umangyadav: You can try generating small onnx-test with different types and if onnx complains or not.

lakhinderwalia: > if one dtype is int8 and the other one is uint8 then it can possibly overflow "faster"

Int8 or uint8 once dequantized internally shouldn't affect precision of int32 operations..  
This is not a specification error, and we should follow it exactly to the (type) T.  :-)

src/onnx/parse_qlinearmatmul.cpp
lakhinderwalia: As I look at this code, we need lens_a and lens_b, and so it would be more efficient to just leave it as it is, and use size(). However, efficiency is not the main requirement, so if you prefer I can change it to ndim(). Thanks.

src/shape.cpp
lakhinderwalia: Done. Thanks.

src/shape.cpp
```diff
@@ -490,6 +490,26 @@ shape shape::with_lens(const std::vector<std::size_t>& l) const
     return this->with_lens(this->type(), l);
 }
 
+shape shape::prepend_dim(size_t dim) const
+{
+    if(this->dynamic())
+        MIGRAPHX_THROW("SHAPE: with_lens() called on dynamic shape");
```
pfultz2, line 496: `with_lens`? Where is this being used?

src/shape.cpp
```diff
@@ -490,6 +490,26 @@ shape shape::with_lens(const std::vector<std::size_t>& l) const
...
+}
+
+shape shape::append_dim(size_t dim) const
+{
+    if(this->dynamic())
+        MIGRAPHX_THROW("SHAPE: with_lens() called on dynamic shape");
+    auto cur_lens = this->lens();
+    std::vector<std::size_t> nlens(1 + cur_lens.size(), dim);
+    std::copy(cur_lens.begin(), cur_lens.end(), nlens.begin());
+    return {this->type(), nlens};
```
pfultz2, line 510: This completely clears the layout of the original shape. I am not sure if its good to add methods like this.

src/onnx/parse_qlinearmatmul.cpp
pfultz2, line 128: You aren't using the shape, you are just checking the dimensions. It would be better to do:

```cpp
lens_a.insert(lens_a.begin(), 1);
dim_a++;
```

src/onnx/parse_qlinearmatmul.cpp
```diff
@@ -0,0 +1,199 @@
...
+        {
+            sh_a   = sh_a.prepend_dim(1);
+            lens_a = sh_a.lens();
+            dim_a  = lens_a.size();
+        }
+        if(dim_b == 1)
+        {
+            sh_b   = sh_b.append_dim(1);
+            lens_b = sh_b.lens();
+            dim_b  = lens_b.size();
```
pfultz2, line 134: Same here. You aren't using the shape, you are just checking the dimensions. It would be better to do:

```cpp
lens_b.push_back(1);
dim_b++;
```

test/onnx/onnx_test.cpp
```diff
@@ -4904,6 +4904,118 @@ TEST_CASE(qlinearadd_test)
     EXPECT(p.sort() == prog.sort());
 }
 
+TEST_CASE(qlinearmatmul_1D_test)
```
kahmed10, line 4907: Missing 3D parser test. If it looks similar to the 2D test, you may be able to make a generic function that can return the migraphx program adjusted for the inputs' dimensions.

umangyadav (approved):
LGTM, make sure to address other comments. Thanks!

src/shape.cpp
umangyadav: yeah. We can probably add an assert/throw that it is a standard shape or better would be to remove it. I don't foresee this being used otherwise.

src/onnx/parse_qlinearmatmul.cpp
lakhinderwalia: We are using the newly updated lens_a and lens_b later on in lines 138 onwards..  Thanks.

src/shape.cpp
lakhinderwalia: We don't have a choice of clearing out the original shape.. imagine doing a prepend of a dimension with some expectations of its strides and whatnot..

The reason to do this api here is that it shouldn't be done in the actual operator code -- it belongs to shape.* logic -- rather than externalize it.

Thanks.

src/shape.cpp
lakhinderwalia: prepend_dim() is being called from matmul logic here, simply because the standard allows a one-dimensional (N) tensor multiplication -- but only after broadcasting it to two dimensions as 1 x N -- or N x 1 (for append_dim). 
(I am not sure if you are asking something else..)

test/onnx/onnx_test.cpp
lakhinderwalia: It is the 1D test that is a bit special.. hence there are two tests: the other one is for 2D. It works (similarly) for anything beyond 2D. I am not sure what generic function are you referring to here in the unit-test, please? 

The generic function is parse_qlinearmatmul.cpp :-)

src/onnx/parse_qlinearmatmul.cpp
lakhinderwalia: (we use lens_b in lines 138 onwards..) Thanks.

test/onnx/onnx_test.cpp
kahmed10: I see. So the 3D onnx file was created for verify_onnx and wasn't going to add test any additional cases in the onnx parser.

kahmed10 (approved):

src/shape.cpp
pfultz2: But why do you mention `with_lens()` here?

src/shape.cpp
pfultz2: > We don't have a choice of clearing out the original shape..

Which is why it makes this method problematic. This could inadvertently change the layout of the tensor, which can lead to bugs or errors in our code, especially when there are more correct ways to do this. For example, prepending or appending ones will lose the strides, whereas using the equivalent squeeze and unsqueeze will not. 

If you dont care about the strides, then you dont need the shape class at all, you just need a vector of dimensions. And in your case you never look at the strides, which means you dont need the shape class.

src/onnx/parse_qlinearmatmul.cpp
pfultz2: > We are using the newly updated lens_a and lens_b later on in lines 138 onwards.

Yes thats my point, you use `lens_b` and `lens_a`(ie the dimensions or lengths of the shape) but you dont use `sh_b`or `sh_a`(ie the shape class)  at all except to get the initial dimensions.

src/shape.cpp
lakhinderwalia: @pfultz2 , do you have a solution for this? Objections, I understand.

I want a method that prepends the lens, and a method that appends a shape. Do you want me to do it manually in that parse_operator codebase. I will be happy to do it. Thanks.

src/onnx/parse_qlinearmatmul.cpp
lakhinderwalia: OK. Let me use this approach of push_back and insert. Thanks.

src/onnx/parse_qlinearmatmul.cpp
lakhinderwalia: Done.

src/shape.cpp
lakhinderwalia: removed from shape.cpp/hpp.

src/shape.cpp
lakhinderwalia: removed changes..

pfultz2 (approved):

PR #2310, opened 2023-10-10
fp16 change default tolerances for driver verify

CharlieL7:
Did some testing, some limitations with how argument_parser works makes setting the default tolerances when `--fp16` option is given more complicated than it needs to be. Updated the code with what works.


TedThemistokleous (approved):

TedThemistokleous:
Fix the minor formatting stuff but overall looks good.

src/driver/main.cpp
```diff
@@ -536,17 +536,20 @@ struct params : command<params>
 struct verify : command<verify>
 {
     compiler c;
-    migraphx::verify::tolerance tols;
+    // Set to -1. as nonsense initial value
+    double rms_tol = -1.;
+    double atol = -1.;
+    double rtol = -1.;
```
umangyadav, line 542: Either write it as `-1.0` or just `-1`. Seeing a `.` alone looks weird IMHO.

umangyadav:
Need to update `accuracy_checker.py` with default tolerances for fp16 precision to not break CI. As a later PR.

src/driver/main.cpp
```diff
@@ -565,11 +568,42 @@ struct verify : command<verify>
         auto t = c.ct.get_target();
         auto m = c.parameters.generate(p, t, true, c.l.batch);
 
+
+        //TODO remove this and make the driver able to figure out datatype most used in the model
+        // then set the tolerances appropriately. Need to check here because c.to_fp16 only set
```
umangyadav, line 573: I think an easy hack would be to look for dtype of convolution and/or GEMM. and based on that tolerances can be set.

src/driver/main.cpp
```diff
@@ -565,11 +568,42 @@ struct verify : command<verify>
...
+        // after argument_parser.parse() is run. This code is complicated because there's not a good way
+        // to change the default tolerances after reading `--fp16` but before reading `--rms-tol`, `--atol`,
+        // and `--rtol`.
+        migraphx::verify::tolerance tols{};
+        if(c.to_fp16)
+        {
+            tols = migraphx::verify::tolerance{8e-2, 4e-2, 4e-2};
+        }
+        auto dbl_comp = [&](double a, double b){
+            return (std::abs(a - b) <= std::numeric_limits<double>::epsilon());
```
umangyadav, line 583: I am not sure how this works. `a-b` can not ever be represented such that it is less than epsilon

src/driver/main.cpp
```diff
@@ -565,11 +568,42 @@ struct verify : command<verify>
...
+        };
+        if(not dbl_comp(this->rms_tol, -1.))
```
umangyadav, line 585: You can probably use `float_equal` instead. It should work for `double` as well.

src/driver/main.cpp
```diff
@@ -579,22 +579,21 @@ struct verify : command<verify>
...
         {
             tols.atol = this->atol;
         }
-        if(not dbl_comp(this->rtol, -1.))
+        if(not float_equal(this->rtol, -1.0))
         {
             tols.rtol = this->rtol;
         }
 
+        std::cout << "rms_tol: " << tols.rms_tol << std::endl;
```
umangyadav, line 595: is this print intentional? If so you can also print others as well.

umangyadav:
> Need to update `accuracy_checker.py` with default tolerances for fp16 precision to not break CI. As a later PR.

Can you make an issue for this ? if not there is not one already

CharlieL7:
> > Need to update `accuracy_checker.py` with default tolerances for fp16 precision to not break CI. As a later PR.
> 
> Can you make an issue for this ? if not there is not one already

#2318

umangyadav (approved):

src/driver/main.cpp
```diff
@@ -540,17 +540,20 @@ struct params : command<params>
 struct verify : command<verify>
 {
     compiler c;
-    migraphx::verify::tolerance tols;
+    // Set to -1. as nonsense initial value
+    double rms_tol       = -1.0;
+    double atol          = -1.0;
+    double rtol          = -1.0;
```
pfultz2, line 546: Use `optional` instead of magic values.

src/driver/main.cpp
```diff
@@ -569,11 +572,42 @@ struct verify : command<verify>
...
 
+        // TODO remove this and make the driver able to figure out datatype most used in the model
+        //  then set the tolerances appropriately. Need to check here because c.to_fp16 only set
+        //  after argument_parser.parse() is run. This code is complicated because there's not a
+        //  good way to change the default tolerances after reading `--fp16` but before reading
+        //  `--rms-tol`, `--atol`, and `--rtol`.
+        migraphx::verify::tolerance tols{};
+        if(c.to_fp16)
+        {
+            tols = migraphx::verify::tolerance{8e-2, 4e-2, 4e-2};
```
pfultz2, line 583: The tolerance should be set int `verify_program`, `verify_instructions`, and `verify_reduced_program` based on whether the model contains fp16 or not, as this wont set the tolerances correctly for prequantized models. We could probably right a function like `get_tolerances`:

```cpp
verify::tolerance get_tolerances(const program& p, std::optional<double> rms_tol, std::optional<double> atol, std::optional<double> rtol)
{
    verify::tolerance result{};
    bool has_fp16 = any_of(p.get_modules(), [](module_ref m) {
        return any_of(m, [](auto&& ins) {
            return ins.get_shape().type() == shape::half_type;
        });
    });
    if(has_fp16)
    {
        result.rms_tol = 8e-2;
        result.atol = 4e-2;
        result.rtol = 4e-2;
    }
    if(rms_tol)
        result.rms_tol = *rms_tol;
    if(atol)
        result.atol = *atol;
    if(rtol)
        result.rtol = *rtol;
    return result;
}
```

That can be called from these functions to set the tolerance.

pfultz2:
My comments for the review was not addressed. Why was this merged in?

causten:
> My comments for the review was not addressed. Why was this merged in?

I actually didn't see your comments and they came in within minutes of me merging and you submitting the review.  @CharlieL7 can create a new PR and that can get merged once ready.

PR #2313, opened 2023-10-11
Fix MLIR input fusion non-std shapes from squeeze, flatten and unsqueeze

manupak:
Currently, we see MLIR partition candidates recieving non-standard shape due to not fusing in squeeze, flatten and unsqueeze ops. These ops could be canonicalized to reshape without introducing additional ops as long as MLIR backend is concerned.


krzysz00 (approved):
This seems fine to me, but I want to confirm that squeeze, unsqueeze, and flatten can all be replaced by `reshape` correctly like this

Also, can you add a test?

giuseros (approved):
I can confirm that the tests are passing on mi200: http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Ftest/detail/test/30/pipeline/54

manupak:
> This seems fine to me, but I want to confirm that squeeze, unsqueeze, and flatten can all be replaced by `reshape` correctly like this
> 
> Also, can you add a test?

They are special cases of 'reshape' where as in reshape just work all that out by just knowing the output shape.

I ve added tests now.

pfultz2 (approved):
In the future, it would be to make these transformation when we are generating MLIR.

PR #2316, opened 2023-10-11
Remove try-except clause for importing numpy in tests

ahsan-ca:
The tests (test_gpu.py and test_numpy.py) silently exit if numpy is not found. This PR removes the try/except
clause in order to have the test case fail if the numpy requirement is not satisfied.


TedThemistokleous (approved):
Update your CI to not break. but overall looks good. Better we scream if test fail than stay quiet.

ahsan-ca:
Removed `sys` as `pyflakes` test was complaining of it being unused.

PR #2329, opened 2023-10-13
Example PR on how we add SQLites 3 support in Windows

apwojcik:
That PR exemplifies how we have added SQLite3 dependency in CMake.


cmake/SQLite3.cmake
```diff
@@ -0,0 +1,58 @@
...
+            sqlite3
+            GIT_REPOSITORY https://github.com/sqlite/sqlite.git
+            GIT_TAG version-3.40.0
+            GIT_SHALLOW true
+            UPDATE_DISCONNECTED true
+            CONFIGURE_COMMAND ""
+            BUILD_COMMAND ${NMAKE_EXECUTABLE} /f ..\\sqlite3\\Makefile.msc USE_AMALGAMATION=1 NO_TCL=1 TOP=..\\sqlite3 libsqlite3.lib
+            INSTALL_COMMAND "")
+
+    ExternalProject_Get_Property(sqlite3 BINARY_DIR)
```
pfultz2, line 50: Building the dependency should be done in a superproject, not directly in our build scripts.

cmake/SQLite3.cmake
```diff
@@ -0,0 +1,58 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
+
+if(NOT WIN32)
+    find_package(PkgConfig REQUIRED)
```
pfultz2, line 26: We can probably switch to `find_package(SQLite3)`, since we use a cmake build for sqlite.

apwojcik: Unfortunately, we cannot. To use `find_package()`, the CMake has to generate config mode files. The CMake from rocm-recipes does not do that.

cmake/SQLite3.cmake
apwojcik: SQLite3 authors do not provide CMake, only `configure` scripts for Linux/GNU and NMake files for Windows. Using `ExternalProject` is an approved technique to work with non-CMake compatible dependency projects.

umangyadav (approved):

apwojcik:
Will be covered in

PR #2330, opened 2023-10-13
embed.cmake: add support for Windows resource file

apwojcik:
This PR introduces the support of Windows resource files to `Embed.cmake`. It is `ON` by default on Windows, and when it is `OFF` `*.cpp` files will be used. The same applies to Linux - `ON` -> `*.o` (LD) or `OFF` -> `*.cpp` .

This PR fixes building resources on Linux with `ld` and `objcopy` commands.


cmake/Embed.cmake
```diff
@@ -21,17 +21,21 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 #####################################################################################
-find_program(EMBED_LD ld)
-find_program(EMBED_OBJCOPY objcopy)
 
-option(EMBED_USE_LD "Use ld to embed data files" OFF)
```
pfultz2, line 27: We shouldn't remove this option.

cmake/Embed.cmake
```diff
@@ -21,17 +21,21 @@
...
+if(WIN32)
+    option(EMBED_USE_RESOURCES "Use data file embedding to binary" ON)
+else()
+    option(EMBED_USE_RESOURCES "Use data file embedding to binary" OFF)
+endif()
+
+if(NOT WIN32 AND EMBED_USE_RESOURCES)
+    find_program(EMBED_LD ld)
```
pfultz2, line 32: Why are we searching for `ld` when `EMBED_USE_RESOURCES` is enabled on linux? This should be an error.

cmake/Embed.cmake
```diff
@@ -153,50 +176,45 @@ function(embed_file OUTPUT_FILE OUTPUT_SYMBOL FILE)
...
+    endif()
+    set(OUTPUT_SYMBOL ${OUTPUT_SYMBOL} PARENT_SCOPE)
 endfunction()
 
 function(add_embed_library EMBED_NAME)
-    set(options)
-    set(oneValueArgs RELATIVE)
-    set(multiValueArgs)
-    cmake_parse_arguments(PARSE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+    set(oneValueArgs BASE_DIRECTORY)
```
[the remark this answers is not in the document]
apwojcik: I changed `RELATIVE` to `BASE_DIRECTORY` to align with the `cmake_path()` nomenclature.
`cmake_path(RELATIVE_PATH <path-var> BASE_DIRECTORY <input> ...)` to make it cleaner to read.

cmake/Embed.cmake
apwojcik: The technique with LD on Linux is the same as RC on Windows. I changed `EMBED_USE_LD` to `EMBED_USE_RESOURCES` because, in both cases, the literature refers to both techniques as embedding resource files in program binary. Here is the example: https://atwillys.de/content/cc/embedding-resource-files-in-a-c-plus-plus-program-binary-on-linux-unix/?lang=en. I can easily find tonnes more. We only need one option to control resources or `*.cpp` files. It makes no sense to have both and expose `EMBED_USE_LD` on Windows and `EMBOED_USE_RC` on Linux.

cmake/Embed.cmake
apwojcik: The technique with LD on Linux is the same as RC on Windows. I changed `EMBED_USE_LD` to `EMBED_USE_RESOURCES` because, in both cases, the literature refers to both techniques as embedding resource files in program binary. Here is the example: https://atwillys.de/content/cc/embedding-resource-files-in-a-c-plus-plus-program-binary-on-linux-unix/?lang=en. I can easily find tonnes more. We only need one option to control resources or `*.cpp` files. It makes no sense to have both and expose `EMBED_USE_LD` on Windows and `EMBOED_USE_RC` on Linux.

cmake/Embed.cmake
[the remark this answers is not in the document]
pfultz2: It uses `RELATIVE` to match [`file(GLOB)`](https://cmake.org/cmake/help/latest/command/file.html#glob) command. I dont think `cmake_path` makes sense here.

cmake/Embed.cmake
pfultz2: > The technique with LD on Linux is the same as RC on Windows. I changed EMBED_USE_LD to EMBED_USE_RESOURCES because, in both cases, the literature refers to both techniques as embedding resource files in program binary.

The name is the technique used for embedding the resource. We are embedding the resource when we use the C arrays as well, so the `EMBED_USE_RESOURCES` flag would apply there as well, so it doesnt make sense.

> It makes no sense to have both and expose EMBED_USE_LD on Windows and EMBOED_USE_RC on Linux.

I would prefer the two flags because there are other techniques to embed resources as well(such as using `#embed`) so I would prefer to have a flag to specify the technique being used to embed the resources.

apwojcik: I prefer a single option for LD and RC (e.g., `EMBED_USE_BINARY` or `EMBED_USE_NATIVE` or `EMBED_USE_OS_NATIVE`) and separate for `#embed` when we switch to C++23 (e.g. `EMBED_USE_CXX_EMBED` or `EMBED_USE_CXX` or `EMBED_USE_CXX_NATIVE`) or something like that. Two flags for LD and RC would result in different Linux and Windows building instructions. One cannot use LD on Windows and RC on Linux. I prefer to keep them the same for both OSes.

cmake/Embed.cmake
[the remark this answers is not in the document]
apwojcik: I changed `file(GLOB)` to `cmake_path()` for making relative paths. However, I don't care much about the name. If that is the only obstacle blocking this PR and you want the name to be `RELATIVE`, I make it `RELATIVE` again.

cmake/Embed.cmake
pfultz2: We should probably have one `EMBED_USE` variable that can be set to `EMBED_USE=RC` for windows resource files, `EMBED_USE=LD` for gnu linker, and `EMBED_USE=CArrays` for the carrays version.

We can also use `set_property(CACHE EMBED_USE PROPERTY STRINGS "LD;CArrays")` to set the options for linux, so in cmake gui(or tui) you can select the option you want.

apwojcik: Done

pfultz2 (approved):

PR #2331, opened 2023-10-13
Add flag to control virtual environments

ahsan-ca:
This adds a flag to disable the virtual environments to prevent installing packages every time CI runs the tests.


test/py/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(VENV_ONNX ${CMAKE_BINARY_DIR}/test/py/venv-onnx)
 set(REQUIREMENTS ${CMAKE_CURRENT_SOURCE_DIR}/requirements.txt)
 set(REQUIREMENTS_ONNX ${CMAKE_CURRENT_SOURCE_DIR}/requirements-onnx.txt)
 set(PYTHON_VERSION_TO_DISABLE_ONNX 3.6)
+option(ENABLE_VIRTUAL_ENV "Enable python virtual environments" ON)
```
umangyadav, line 31: By default is it On. I thought motivation was to include this flag was to not have it enabled inside CI

ahsan-ca: Thanks for bringing this up Umang, I think its good we make a conscious decision about it.

My initial thoughts were to have it explicitly disabled for CI since that is one scenario, and have it enabled (by default) for all other scenarios unless specified otherwise. If we default to disabling it, all users would have to explicitly enable it to use virtual env.
I am not sure which is a good way to go about. I think it depends on the use cases and how we would like the default to be depending on the use cases? Perhaps @pfultz2 and @causten can weigh in on the preference here.
Thanks!

ahsan-ca: The flag name could also be `DISABLE_VIRTUAL_ENV`, open to feedback on the preference there too.

causten: We have control in Jenkins on how to invoke repeatable processes.  The goal of virtual environments is to take ownership of the environment where `make check` is used by others because we have been battling various inconsistencies.  As an example, we moved to onnx==1.10.1 over a year ago, and this week to saw a JIRA showing onnx==1.8.  Now in mainline we have support for "onnx==1.14.1".  

I'd like the default path of `make check` to invoke the virtual environment.  Jenkins scripts can ensure the non virtual environment is done correctly.

umangyadav: Okay. we can make it `MIGRAPHX_DISABLE_VIRTUAL_ENV` .  But would need to change Jenkins scripts to pass that flag in this PR.

ahsan-ca: Sounds good. Do you know the location of the Jenkins script that needs to be modified?

umangyadav: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/7604ecf5b9cd04f9df4219d6f1ef998951ed7449/Jenkinsfile#L33 Here you would just need to pass that flag.

umangyadav: It would be better to have a stage with virtual env enabled inside Jenkins though to test out Virtual Env as well.

pfultz2: > It would be better to have a stage with virtual env enabled inside Jenkins though to test out Virtual Env as well.

I understand the motivation but I dont think its a good idea, we are just adding extra stages that have to download packages already downloaded.

test/py/CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ function(add_py_test NAME SCRIPT FIXTURE_NAME VENV_DIR)
             "PYTHONMALLOC=debug"
             "MALLOC_CHECK_=3"
         )
-        if(ENABLE_VIRTUAL_ENV)
+        if(NOT MIGRAPHX_DISABLE_VIRTUAL_ENV)
```
umangyadav, line 65: Suggestion : Better to flip if and else blocks

umangyadav (approved):

test/py/CMakeLists.txt
umangyadav: > I understand the motivation but I dont think its a good idea, we are just adding extra stages that have to download packages already downloaded. 

Okay in that case we would need to rely on QA for virtualenv testing.  most of us would be using dockers for dev and won't be using virtualenv.

test/py/CMakeLists.txt
ahsan-ca: Thanks for the suggestion. I have made the change.

pfultz2 (approved):

PR #2274, opened 2023-10-03
Enable MLIR by default for more cases

pfultz2:
This will enable MLIR by default for these cases:

* Any convolution fusion
* Any int8 gemm fusion
* All Navi3 standalone convolutions
* With a flag(ie `MIGRAPHX_ENABLE_MLIR`) to enable MLIR for floating-point gemm fusions

Except:

* 3x3 winnograd convolutions fusions (except on Navi)
* K > 2048 on gemm (as CK)

Also there is `MIGRAPHX_DISABLE_MLIR` to disable MLIR completely.

PR #2334, opened 2023-10-13
Change to check for fp16 and --fp16

CharlieL7:


src/driver/main.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * The MIT License (MIT)
+ii The MIT License (MIT)
```
umangyadav, line 2:
```suggestion
* The MIT License (MIT)
```

src/driver/main.cpp
```diff
@@ -537,25 +539,57 @@ struct params : command<params>
     }
 };
 
+/**
+ * Gives tolerances based on user input (`rms_tol`, `atol`, `rtol` parameters) and defaults.
+ * Sets to fp16 tolerances if `quantize` input is fp16 or any fp16 instruction in found in the model.
+ */
+verify::tolerance get_tolerances(const program& p, precision quantize, std::optional<double> rms_tol, std::optional<double> atol, std::optional<double> rtol)
```
pfultz2, line 546: Move this function to driver/verify.cpp.

src/driver/main.cpp
```diff
@@ -41,6 +41,8 @@
 #include <migraphx/json.hpp>
 #include <migraphx/version.h>
 
+#include <migraphx/instruction.hpp>
+#include <migraphx/module_ref.hpp>
```
pfultz2, line 45: Remove these includes.

src/driver/main.cpp
```diff
@@ -537,25 +539,57 @@ struct params : command<params>
...
 
+/**
+ * Gives tolerances based on user input (`rms_tol`, `atol`, `rtol` parameters) and defaults.
+ * Sets to fp16 tolerances if `quantize` input is fp16 or any fp16 instruction in found in the model.
+ */
+verify::tolerance get_tolerances(const program& p, precision quantize, std::optional<double> rms_tol, std::optional<double> atol, std::optional<double> rtol)
+{
+    bool has_fp16 = any_of(p.get_modules(), [](auto&& m) {
+            return any_of(*m, [](auto&& ins) {
+                    return (ins.get_shape().type() == shape::half_type);
```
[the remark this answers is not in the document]
umangyadav: It's fine i guess to just check for any instruction with fp16. At least I can see the scenario of me using a trimmed onnx model without convolution or gemm.

umangyadav (approved):

causten:
@CharlieL7 can you check if the tolerances are valid, I see two models failing that were not in the past

CharlieL7:
> @CharlieL7 can you check if the tolerances are valid, I see two models failing that were not in the past

@causten Those tests are using the accuracy_checker script, should be unrelated to the changes in this PR. Issue originates from #2274.

src/driver/argument_parser.hpp
```diff
@@ -187,6 +187,13 @@ struct value_parser
     }
 };
 
+// version for std::optional object
+template <class T>
+struct value_parser<std::optional<T>>
+{
+    static T apply(const std::string& x) { return value_parser<T>::apply(x); }
```
pfultz2, line 194: Actually this should return `std::optional<T>`: `static std::optional<T> apply(const std::string& x) { return value_parser<T>::apply(x); }`.

pfultz2 (approved):

PR #2336, opened 2023-10-16
Add scales attribute parse in upsample for older opset versions

attila-dusnoki-htec:
Upsample [version-7](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Upsample-7) has scales as attribute, [version-9](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Upsample-9) has scales as input.

Resolves: https://github.com/migraphx-benchmark/AMDMIGraphX/issues/142


src/onnx/parse_resize.cpp
```diff
@@ -265,23 +265,38 @@ struct parse_resize : op_parser<parse_resize>
...
+        if(all_of(out_lens.cbegin(), out_lens.cend(), [](auto o) { return o == 0; }))
+        {
+            std::transform(
+                in_lens.begin(),
+                in_lens.end(),
+                vec_scale.begin(),
+                out_lens.begin(),
+                [&](auto idx, auto scale) { return static_cast<std::size_t>(idx * scale); });
+        }
+
```
umangyadav, line 299: can you remove `contiguous` on line 306? It shouldn't be necessary anymore. 

See this one : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2190 

That would  be unrelated to this PR but it is small enough change. You may need to change some tests though.

src/onnx/parse_resize.cpp
```diff
@@ -265,23 +265,38 @@ struct parse_resize : op_parser<parse_resize>
...
-                                       return static_cast<std::size_t>(idx * scale);
-                                   });
                 }
             }
         }
 
+        // scales still missing, must be an attribute
+        if(vec_scale.empty())
+        {
+            if(contains(info.attributes, "scales"))
```
umangyadav, line 275: I think it would be better to first check if attribute exists or not. If it exist then, we can skip For loop at Line 219. 
"Resize" operator doesn't have `scales` attribute or input.  Therefore for the resize, it would always use `for` loop.

umangyadav (approved):

causten:
@attila-dusnoki-htec  fix the tidy error and this can merge

attila-dusnoki-htec:
> @attila-dusnoki-htec fix the tidy error and this can merge

Should be fixed now. It was also rebased to develop.

PR #2338, opened 2023-10-16
Update script for build_and_test_onnxrt.sh

TedThemistokleous:
Related to #2335 

Was breaking CI if we use the latest Onnxruntime since this script was updated last week from migraphx_test_launcher.sh vs pai_test_launcher.sh


causten:
shouldn't this be part of the https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2335 PR ?

umangyadav (dismissed):

umangyadav:
> shouldn't this be part of the #2335 PR ?

Yes, it is better to bundle them together.

TedThemistokleous:
Bundling changes to #2335

PR #2343, opened 2023-10-18
Add layout attribute support for LSTM operator

nives-vukovic:
Since ONNX opset version 14, layout attribute has been introduced to LSTM operator which allows two predefined layouts for input and output shapes.

Add corresponding reference, onnx, and verify tests.

Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/102


shivadbhavsar:
Referencing the onnx ops doc: https://github.com/onnx/onnx/blob/main/docs/Operators.md#LSTM
It seems like the layout attribute specifies one of two input permutations. The best way to support this as Paul suggested would be to add a condition in the lstm parser and add an extra transpose op to the input and then apply the lstm op as it currently exists. 

Is there some issue with doing it this way that Im not seeing?

src/onnx/parse_lstm.cpp
```diff
@@ -202,6 +202,12 @@ struct parse_lstm : op_parser<parse_lstm>
             input_forget = parser.parse_value(info.attributes.at("input_forget")).at<int>();
         }
 
+        int layout = 0;
+        if(contains(info.attributes, "layout"))
+        {
+            layout = parser.parse_value(info.attributes.at("layout")).at<int>();
+        }
+
```
shivadbhavsar, line 210: Referencing the onnx ops doc: https://github.com/onnx/onnx/blob/main/docs/Operators.md#LSTM
It seems like the layout attribute specifies one of two input permutations. The best way to support this as Paul suggested would be to add a condition here and add an extra transpose op to the input and then apply the lstm op as it currently exists. 

Is there some issue with doing it this way that Im not seeing?

nives-vukovic: @shivadbhavsar @pfultz2 The main reason I chose this approach is performance. There are three inputs which axes would have to be permuted and after the lstm operation, axes of three outputs would have to be permuted back to the chosen layout. Propagation of the layout attribute to the lstm operation ensures the same performance regardless of the layout, the changes are not really in the algorithm, but only which axes are used depending on the layout. This approach gives correct results, would you prefer it to stay this way or to use the simpler solution with an overhead?

The same attribute has been also introduced to RNN and GRU. Whatever the approach is chosen will be used for these operators as well.

shivadbhavsar: That reasoning makes sense to me. What I dont fully understand is according to the docs I linked above, the outputs seem to always be the same shape regardless of the layout:
```
Y = [seq_length, num_directions, batch_size, hidden_size]
Y_h, Y_c = [num_directions, batch_size, hidden_size]
```
So if layout=1, wont the onnx graph have to transpose those outputs anyways to fit the specs for layout=1 when fed into the next lstm op? Or are the docs missing specifying output shapes for layout=1?

nives-vukovic: They just constructed the documentation poorly. The shape of outputs in case of layout=1 is only visible in description of the layout attribute:

Straight from the page:
If 1, the following shapes are expected: X.shape = [batch_size, seq_length, input_size], Y.shape = **[batch_size, seq_length, num_directions, hidden_size]**, initial_h.shape = Y_h.shape = initial_c.shape = Y_c.shape = **[batch_size, num_directions, hidden_size]**.

shivadbhavsar: Ya I caught that after replying.
So in terms of overhead I think it would be big concern if we are adding these transposes for each input and output for the _n_ LSTM ops that will occur in onnx graphs. But in general if you have a transpose for the output of lstm_1, and the same transpose for input to lstm_2, those will be removed during simplification passes. So the total overhead will be transpose ops for the inital inputs and the final outputs only as far as I understand.
Im not sure we will see a meaningful difference in real models from doing this. @pfultz2 any thoughts on this?

pfultz2: I don't think the extra transposes  should cause a perf issue. First, transpose is an aliased operator so there is no extra kernel or copy. Secondly, all the pointwise and gemms(that's what lstm decomposes to) are optimized for transposed layouts so there shouldn't be a perf issue here either.

nives-vukovic: @pfultz2 @shivadbhavsar The implementation has been changed in line with the discussion. Tests have also been modified to perform transpose of available inputs and requested outputs.

On our local setup everything passes, except for a verify test 'test_lstm_bidirct_hs_layout' on gpu. After analysis it seems that 'eliminate_contiguous' causes some issues for this specific case when sequence_lens input is present and direction is not forward (everything passes when eliminate_contiguous is commented out). There is an active discussion regarding eliminate_contiguous on another PR as well, but we can wait to see if the issue is present here as well.

src/include/migraphx/op/lstm.hpp
```diff
@@ -49,6 +49,7 @@ struct lstm
     rnn_direction direction = rnn_direction::forward;
     float clip              = 0.0f;
     int input_forget        = 0;
+    int layout              = 0;
```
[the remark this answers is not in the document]
nives-vukovic: The implementation has been changed in line with the discussion.

src/onnx/parse_lstm.cpp
shivadbhavsar: The changes look good to me. I've started a CI run so lets see if that test case results in the same issue in the CI environment.

nives-vukovic: @shivadbhavsar I see that the performance tests failed due to the branch not being on your repository, I am not sure in which group verify tests are checked and if they pass (I know they are included in make check). Please let me know if any action is required from my side.

shivadbhavsar: Looks like all the test cases passed in all the builds so everything looks fine there. For the performance tests I believe the only way to get those to run would be to copy over the branch to this repo.

shivadbhavsar (approved):

src/onnx/parse_lstm.cpp
nives-vukovic: @shivadbhavsar [log.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/13317974/log.txt)
Here is the log file with the error that we get on our side for the test_lstm_bidirct_hs_layout verify test, MIGRAPHX_TRACE_ELIMINATE_CONTIGUOUS is enabled. Just in case you encountered something similar.
It reports: 
FAILED: gpu
RMS Error: 0.202458
Max diff: 0.426603
Mismatch at 5: -0.193233 != 0.0861849

nives-vukovic: @shivadbhavsar @causten After analysis and as can be seen in the last comment, we deduced that the cause of the error is not connected directly to this commit or changes done here. Therefore I changed this PR to use a different bidirect verify test and will create a separate issue with detailed description of the problem.

pfultz2 (approved):

PR #2344, opened 2023-10-18
Added fix for test_targets on Windows

tvukovic-amd:
The issue on Windows is that we couldn't get reference target without force to load it. 

Since migraphx_ref.dll is not supposed to be linked during AMDMIGraphX build on Windows, we need to force targets test to load it by making ref target (similar as for GCC) using `auto ref_target = migraphx::make_target("ref");`. On Linux if we are using clang, we don't need to force it.

`migraphx::make_target("ref")` function runs dynamic loader with migraphx_ref.dll after which we can get target using ` migraphx::get_targets()`.


test/targets.cpp
```diff
@@ -43,7 +43,7 @@ TEST_CASE(targets)
 {
     // GCC doesn't load libmigraphx_ref unless necesssary even though it is linked to the test.
     // Force it to load by making ref target
-#if defined(__GNUC__) && !defined(__clang__)
+#if defined(__GNUC__) && !defined(__clang__) || defined(_MSC_VER)
```
umangyadav, line 46: We have moved all Ref tests under `test/ref` therefore we should be able to remove `migraphx_ref` from here. 

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/5139b9308a7b8deda201e972cc24ffa6c78c9036/test/CMakeLists.txt#L28 

And after that, we can remove defines for clang, gcc or msc_ver and call to `make_target("ref")`. In that case, it would dynamically load `migraphx_ref` for all cases including clang, gcc and windows.

tvukovic-amd: Thank you for quick answer, I ran `test_ref` on Windows and it is working.

tvukovic-amd: Alright, I will commit new changes then. Thank you.

umangyadav (approved):

pfultz2 (approved):

PR #2345, opened 2023-10-18
[6.1] Add support for dot-(mul)-softmax-dot offloads to MLIR

manupak:
This PR adds support to be able to offload dot-(mul)-softmax-dot + pointwise offloads.
The current kernel needs more perf improvements to be useful, hence the focus is to get
the integration pipe cleaned.

I only noticed prefuse::gemm-softmax-gemm approach used with CK until later.
Having looked at that, I think this approach has the merit that the IR transformation
could be scoped strictly within fuse_mlir pass. However, open to change if that is better.


Jenkinsfile
```diff
@@ -140,7 +140,9 @@ rocmtest clang_debug: rocmnode('mi100+') { cmake_build ->
             def debug_flags_cxx = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize=vptr,function -fno-sanitize-recover=${sanitizers}"
             def debug_flags = "-g -O2 -fsanitize=${sanitizers} -fno-sanitize=vptr -fno-sanitize-recover=${sanitizers}"
             def gpu_targets = getgputargets()
-            cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags_cxx}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}' -DGPU_TARGETS='${gpu_targets}'")
+            // Since the purpose of this run verify all things MLIR supports,
+            // enabling all possible types of offloads
+            cmake_build(flags: "-DCMAKE_BUILD_TYPE=debug -DMIGRAPHX_ENABLE_PYTHON=Off -DMIGRAPHX_ENABLE_MLIR=On -DMIGRAPHX_MLIR_USE_SPECIFIC_OPS=fused,attention,convolution,dot -DCMAKE_CXX_FLAGS_DEBUG='${debug_flags_cxx}' -DCMAKE_C_FLAGS_DEBUG='${debug_flags}' -DGPU_TARGETS='${gpu_targets}'")
```
pfultz2, line 145: `MIGRAPHX_MLIR_USE_SPECIFIC_OPS` is an environment variable not a cmake variable.

manupak: ah right! changed now.

causten:
I'm moving the MLIR pointer to [Enable MIGraphX to offload gemm-(mul)-softmax-gemm ops. (https://github.com/ROCmSoftwarePlatform/rocMLIR/commit/d5781ff90b77362e32fb49989b4485d009446876)

manupak:
@pfultz2 this is more or less ready (only blocked by bumping the rocMLIR commit pointer which is happening in a seperate PR).
Would you be able to take look in the meantime ?

manupak:
@pfultz2 @causten I would be out next week and would really appreciate to see feedback on this PR once Im back to keep the ball rolling here. Im sure @pfultz2 will have comments :). So it would better to address them early on.

pfultz2:
This should reuse the `prefuse::gemm-softmax-gemm` operator here.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -342,37 +342,209 @@ struct find_mlir_fused_ops
...
+
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
+        auto ins = r.result;
+        auto* pm = ins->module_inputs().front();
+        // Whitelist pointwise operators.
+        if(std::any_of(pm->begin(), pm->end(), [&](const auto& i) {
+               return not is_pointwise_op_supported_by_mlir(i);
+           }))
+            return;
```
pfultz2, line 354: This should be moved to the matcher:

```cpp
MIGRAPHX_PRED_MATCHER(mlir_pointwise, instruction_ref ins)
{
    if (ins->name() != "pointwise")
        return false;
    auto* pm = ins->module_inputs().front();
    return std::all_of(pm->begin(), pm->end(), [&](const auto& i) {
        return is_pointwise_op_supported_by_mlir(i);
    });
}
```

And then use `mlir_pointwise()` instead of `match::name("pointwise")`.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -342,37 +342,209 @@ struct find_mlir_fused_ops
...
+
+struct find_mlir_standalone_attention_op
+{
+    mlir_mode mode = mlir_mode::none;
+    void rewrite(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
         static size_t counter = 0;
-        module_ref mm =
-            mpm.create_module("mlir_" + conv_based_op->name() + std::to_string(counter++));
+        module_ref mm         = mpm.create_module("mlir_" + std::to_string(counter++));
```
pfultz2, line 400: This should be `mlir_attention_` to avoid conflicts.

manupak:
> This should reuse the `prefuse::gemm-softmax-gemm` operator here.

It is more like prefuse::(ck-)gemm-softmax-gemm, hence being the reason I avoided it initially (however open to change for right reasons)

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/225873aab3ebcff5be14631c77d9a922ef05cd73/src/targets/gpu/prefuse_ops.cpp#L131-L176

So the current implementation is hardcoded to find a "ck_gemm" and there is a post-match filtering to cater for what appears as a what CK supports. So do you meant to say that I should add two flavors ?
1) prefuse::ck-gemm-softmax-gemm
2) prefuse::mlir-gemm-softmax-gemm

manupak:
@pfultz2 can you clarify what you want here ? (its blocking addressing other comments because if I am to use that matchers would be entirely different).
cc : @causten

pfultz2:
> can you clarify what you want here ?

I think we can rewrite the fusion to work for all gemms not just ck_gemms. I think we only need the ck limitation when only ck is enabled. I assume mlir can handle all sizes(whereas ck doesnt).

manupak:
> > can you clarify what you want here ?
> 
> I think we can rewrite the fusion to work for all gemms not just ck_gemms. I think we only need the ck limitation when only ck is enabled. I assume mlir can handle all sizes(whereas ck doesnt).

Yes, we should be able to handle any size. 
I am yet to convince myself that this is a better design though.

To elaborate your ask as per my understanding,
1) Make the prefuse_ops pass to be backend sensitive.
2) It will rewrite patterns based on whether CK or MLIR is enabled
3) It will need to be predicated on MIGRAPHX_MLIR_SPECIFIC_OPS (to figure out whether to inject the prefuse op or not)
   Therefore decoding of MIGRAPHX_MLIR_SPECIFIC_OPS logic needs to be moved to a common place to be used between two 
   passes

Now, the current version enjoys the benefit its all done within a single pass : fuse_mlir and not having bleed out mlir partitioning utils (i.e. decoding of MIGRAPHX_MLIR_SPECIFIC_OPS )

Apologies for being very particular about this, however, I d rather get the confirmation before I do the major rework you request.

manupak:
I have done asked the change. PTAL

manupak:
@pfultz2 @causten Im waiting on comments to proceed here...

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -216,93 +356,7 @@ struct find_mlir_fused_ops
...
-        {
-            return std::all_of(i.inputs().begin(), i.inputs().end(), [](const auto& arg) {
-                return contains({type_t::float_type, type_t::half_type}, arg->get_shape().type());
-            });
-        }
-        return false;
-    }
-
-    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    void rewrite(module_pass_manager& mpm, const match::matcher_result& r) const
```
pfultz2, line 359: I dont see why we need a seperate method, this can just go into the `apply` method.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -342,59 +385,138 @@ struct find_mlir_fused_ops
...
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const { rewrite(mpm, r); }
 };
 
 template <auto Matcher>
 struct find_mlir_standalone_op
 {
     mlir_mode mode = mlir_mode::none;
     auto matcher() const { return Matcher(mode); }
+
+    void rewrite(module_pass_manager& mpm, instruction_ref top_ins) const
```
pfultz2, line 398: This should go into the `apply` method.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -342,59 +385,138 @@ struct find_mlir_fused_ops
...
- * with MLIR regardless of the underlying GPU architecture. The variable accepts
- * a list of operations separated by comma. The variable recognizes the following
- * operations: "fused", "convolution", "dot". If the variable is not defined MIGraphX
- * will decide by itself which operations to delegate to MLIR. The variable is
- * intended to be primarily used by rocMLIR developers.
- */
-MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_USE_SPECIFIC_OPS);
+struct find_mlir_standalone_attention_op
+{
+    mlir_mode mode = mlir_mode::none;
```
pfultz2, line 430: This is not even used so it can be removed.

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -31,6 +31,9 @@
 #ifdef MIGRAPHX_USE_COMPOSABLEKERNEL
 #include <migraphx/gpu/ck.hpp>
 #endif
+#ifdef MIGRAPHX_MLIR
```
pfultz2, line 34: We dont need an ifdef for MLIR, we assume its always built with MLIR.

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -124,34 +127,24 @@ struct find_add_layernorm
...
-}
-
+template <auto CheckDot>
 struct find_gemm_softmax_gemm
 {
     auto matcher() const
     {
         auto gemm1 =
-            match::skip(match::name("contiguous"))(match::name("dot")(is_ck_gemm().bind("gemm1")));
+            match::skip(match::name("contiguous"))(match::name("dot")(CheckDot().bind("gemm1")));
```
pfultz2, line 142: This should just do `match::any_of(is_ck_gemm(), is_mlir_gemm())`.

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -179,6 +172,34 @@ struct find_gemm_softmax_gemm
     }
 };
 
+#ifdef MIGRAPHX_USE_COMPOSABLEKERNEL
```
pfultz2, line 175: Dont use `ifdef` for defining the matcher.

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -179,6 +172,34 @@ struct find_gemm_softmax_gemm
...
+}
+
+#endif
+
+#ifdef MIGRAPHX_MLIR
+
+auto is_mlir_gemm()
+{
+    return match::make_basic_pred_matcher([=](instruction_ref ins) {
+        if(ins->name() != "dot")
```
pfultz2, line 195: Check if mlir is enabled here.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -342,59 +385,138 @@ struct find_mlir_fused_ops
...
+    }
+
+    bool check(const match::matcher_result& r) const
+    {
+        auto trailing_pm_ins = r.instructions["trailing_pm"]; // input after contiguous
+        auto* trailing_pm    = trailing_pm_ins->module_inputs().front();
+        // Whitelist pointwise operators.
+        return not(std::any_of(trailing_pm->begin(), trailing_pm->end(), [&](const auto& i) {
+            return not is_pointwise_op_supported_by_mlir(i);
+        }));
```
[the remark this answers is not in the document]
manupak: I;d like to consider that is not a part of the code change here. 
See : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/cdc4f91b84e9975f80ba072a644486b6876c23a4/src/targets/gpu/fuse_mlir.cpp#L305-L316

Whilst I agree its a good stylistic change to be done overall, I m just being consistent with what we had for gemms and convs already. Hence that could be changed in follow up PR to change everywhere ?

src/targets/gpu/fuse_mlir.cpp
manupak: Im going to push_back and say this out of scope and I dont see the reason for this to be done (here and now!) in this PR as this was the pattern used fuse_mlir pass from a long time. 

In principle I dont disagree with you but I feel we can push it out as a cleanup to change such checks through out in fuse_mlir pass not just for attention.

src/targets/gpu/prefuse_ops.cpp
manupak: huh... how so ? 
is_ck_gemm is different is_mlir_gemm and ck would not want to offload what mlir wants to offload (vice versa).

manupak (commented):
All comments are addressed now!
Since this PR has been sitting here for more than a month now, I would appreciate a quick turnaround in reviews. @pfultz2 @causten

src/targets/gpu/prefuse_ops.cpp
manupak: nvm, this works because you asked me later to move the enablements checks inside gemm checks.

src/targets/gpu/fuse_mlir.cpp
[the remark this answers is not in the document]
manupak: nvm I did the change now.

pfultz2 (approved):

manupak:
@causten is there a special trick to run the Jenkins CI ? 
http://micimaster.amd.com/blue/organizations/jenkins/AMDMIGraphX/detail/PR-2345/5/pipeline/189
(it gets permission denied errors...)

manupak:
Does anyone know how can I retrigger jenkins CI on this that actually "runs" with permissions ?

PR #2346, opened 2023-10-18
Make argument constructor explicit

pfultz2:
This implicit conversion was causing the perf issus in #1668, as the function was using `argument` but passing `shape`. So instead we should make the constructor explicit to avoid possible errors like this in the future.


CharlieL7 (approved):
looks fine. What was the perf issue?

TedThemistokleous (approved):

pfultz2:
> looks good. 
> 
> There is also `argument(shape, std::nullptr)`. i wonder if it needs to be marked as explicit or not. 

You only need explicit on unary constructors. There is no implicit conversion for binary constructors.

umangyadav (approved):

umangyadav:
>  What was the perf issue? 

Curious on this ? why implicit conversion caused perf issue ?

bpickrel:
> looks fine. What was the perf issue?

Of the test cases, most ran incrementally faster with this tuning but one or two regressed to taking 10-20x as long as before.

src/include/migraphx/argument.hpp
```diff
@@ -46,7 +46,7 @@ struct MIGRAPHX_EXPORT argument : raw_data<argument>
 {
     argument() = default;
 
-    argument(const shape& s);
+    explicit argument(const shape& s);
```
bpickrel, line 49:
```suggestion
    // explicit to protect against passing argument to a function that only needs a shape 
    explicit argument(const shape& s);
```

bpickrel (approved):

bpickrel:
> > looks fine. What was the perf issue?
> 
> Of the test cases, most ran incrementally faster with this tuning but one or two regressed to taking 10-20x as long as before.

Paul said that when it implicitly converted function arguments from the shape type to the "argument" type, that caused it to make unnecessary memory allocations.  I don't know the details of how much memory, or how many times it happened per inference, or why it only happened with a few models.

PR #2351, opened 2023-10-20
Optimize docker image building in Jenkinsfile

NISHIY-EKSDEE:
### Notes
- Restarting from a stage is impossible, since we have parallel stages, not sequential ones. Alternatively, we can make a RUN_STAGE_NAME parameter for each stage so that unnecessary stages can be turned off


Jenkinsfile
```diff
@@ -1,3 +1,7 @@
+import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
+
+DOCKER_IMAGE = 'rocm/migraphx-ci-ubuntu'
```
NISHIY-EKSDEE, line 3: This DOCKER_IMAGE value was taken from the Github Actions file. Should we change the repository since these pipelines use different Dockerfiles?

causten: The docker is the build environment, not the build of migraphx itself.  Essentially the prereqs.  If the commit hash (not :latest) is the same, then the docker would be valid even from that repo.  The challenge would come from the case where the PR is brand new so the github side didnt create it yet.  Then each job would be building and perhaps attempting to publish at the same time.

pfultz2: That docker image is built from hip-clang.docker. Its not the same as what used on jenkins which is based on the Dockerfile.

TedThemistokleous (approved):
Lets get this in since its an overall benefit to save CI time for now. We'll scale all the PRs across the board (so about a 30-40x) and we can work on blocking out stages if they've been re run before after this is in.

PR #2352, opened 2023-10-20
Update unit as  inference/sec

pramenku:
As a user, I am trying to enable the unit self-explanatory.  

Usually, we see unit format as  <some value>  <what the value for>/<time>. Example: X images/sec,  X frames/sec or etc where images or frames helping user to understand , what the value for.

Below Rate unit, until, I read doc or check with owner, I am not sure what it means.  Just for a layman, if we keep like below, they might not get it. Rate: X/sec

But, if we keep, X inference/sec, it will help clearly to know what this value for.  The Rate is  how many inferences of the given inputs can be processed per second so X inference/sec will help user for the same.

For advance user, it /sec might be ok but I am trying to cover all.

### DIFF
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -936,7 +936,7 @@ void program::perf_report(std::ostream& os,
     os << std::endl;
 
     os << "Batch size: " << batch << std::endl;
-    os << "Rate: " << rate * batch << "/sec" << std::endl;
+    os << "Rate: " << rate * batch << "inference/sec" << std::endl;
     os << "Total time: " << total_time << "ms" << std::endl;
     os << "Total instructions time: " << total_instruction_time << "ms" << std::endl;
     os << "Overhead time: " << overhead_time << "ms"
--- a/src/program.cpp
+++ b/src/program.cpp
@@ -936,7 +936,7 @@ void program::perf_report(std::ostream& os,
     os << std::endl;
 
     os << "Batch size: " << batch << std::endl;
-    os << "Rate: " << rate * batch << "inference/sec" << std::endl;
+    os << "Rate: " << rate * batch << "inferences/sec" << std::endl;
     os << "Total time: " << total_time << "ms" << std::endl;
     os << "Total instructions time: " << total_instruction_time << "ms" << std::endl;
     os << "Overhead time: " << overhead_time << "ms"


src/program.cpp
```diff
@@ -936,7 +936,7 @@ void program::perf_report(std::ostream& os,
     os << std::endl;
 
     os << "Batch size: " << batch << std::endl;
-    os << "Rate: " << rate * batch << "/sec" << std::endl;
+    os << "Rate: " << rate * batch << "inference/sec" << std::endl;
```
umangyadav, line 939:
```suggestion
    os << "Rate: " << rate * batch << "inferences/sec" << std::endl;
```

umangyadav: Minor comment

pramenku:
Hi @causten @umangyadav 
Please feel free to merge as I don't have the permission.

After that need to put the same in master and later 6.0 release branch

pfultz2 (approved):

kahmed10:
This formatting looks wrong. Before it was /sec, which is fine to not have a space between the number. But now we need to modify it as such:
```
    os << "Rate: " << rate * batch << " inferences/sec" << std::endl;
```

causten:
> This formatting looks wrong. Before it was /sec, which is fine to not have a space between the number. But now we need to modify it as such:
> 
> ```
>     os << "Rate: " << rate * batch << " inferences/sec" << std::endl;
> ```

I agree...

```
Batch size: 16
Rate: 1337.84inferences/sec
Total time: 11.9596ms
Total instructions time: 12.9021ms
Overhead time: 0.0595123ms, -0.942473ms
Overhead: 0%, -8%
```

Honestly it would parse a lot easier if it looked like...

```
Batch size: 16
Rate (inf/sec): 1337.84
Total time (ms): 11.9596
Total instructions time (ms): 12.9021
Overhead time (ms): 0.0595123, -0.942473
Overhead: 0%, -8%
```

umangyadav:
> Honestly it would parse a lot easier if it looked like... 
Batch size: 16
Rate (inf/sec): 1337.84
Total time (ms): 11.9596
Total instructions time (ms): 12.9021
Overhead time (ms): 0.0595123, -0.942473
Overhead: 0%, -8% 

I like that better. also i'll need to change my script to reflect this change now. I'll do that now. 

https://github.com/ROCmSoftwarePlatform/YModel-Scripts/blob/90bc6e1dc1b600bc062cd3d55c0791ec6387c334/run_ymodel.sh#L147

pramenku:
Rate (inf/sec): 1337.84
>> This also looks fine. Can we get this merged in develop, master and release 6.0 branch.

After that we need to also modify the our automation code.

cc: @umangyadav @causten

PR #2354, opened 2023-10-20
Add benchmark tracing

pfultz2:


umangyadav (approved):

src/targets/gpu/compile_ops.cpp
```diff
@@ -179,14 +180,29 @@ struct compile_plan
...
-                return time_op(*ctx, cr->replace.code_object, to_shapes(cr->ins->inputs()), 20);
-            });
+        std::transform(results.begin(),
+                       results.end(),
+                       config->solutions.begin(),
+                       std::back_inserter(times),
+                       [&](const auto& cr, const auto& solution) {
+                           if(enabled(MIGRAPHX_TRACE_BENCHMARKING{}))
+                               std::cout << "Benchmarking solution: " << solution << std::endl;
+                           if(not cr.has_value())
```
umangyadav, line 194: Perhaps keep this check  earlier and only print solutions that has binary ? 
Because I suspect with CK there would be a lot of solutions with no binary.

pfultz2:
This adds tracing to the benchmarking.

PR #2355, opened 2023-10-20
Set the SO version correctly

pfultz2:
Different migraphx versions are not binary compatible, so we need to make sure a different major version is used for every version of migraphx.


CMakeLists.txt
```diff
@@ -76,7 +76,8 @@ include(ROCMSetupVersion)
 option(BUILD_DEV "Build for development purpose only" OFF)
 
 rocm_setup_version(VERSION 2.8.0)
-set(MIGRAPHX_SO_VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR})
+math(EXPR MIGRAPHX_SO_MAJOR_VERSION "(${PROJECT_VERSION_MAJOR} * 1000 * 1000) + (${PROJECT_VERSION_MINOR} * 1000) + ${PROJECT_VERSION_PATCH}")
+set(MIGRAPHX_SO_VERSION ${MIGRAPHX_SO_MAJOR_VERSION}.0)
```
umangyadav, line 80: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2267#issuecomment-1765397986 

Isn't the `rocm_set_soversion` already doing this ?

pfultz2: No its not. For 2.7 it will set the SO version to `2.7`, which the minor version should be part of the major version(and no minor version), so it should be `2007000.0`. 

Thats because between 2.6 and 2.7 we want different major versions for the SO. So we want the versions to be `2007000` and `2006000`, not `2.7` and `2.6`(which incorrectly implies they are compatible).

umangyadav: @frepaul can you approve this? Would it result in any incompatibilities?

umangyadav (approved):
approving as this change doesn't affect the API library. For the API library, it would follow whatever `rocm_set_soversion` scheme as defined here. 
(https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/d1abf06fa61a2833c7b66feae4a02696f8684db7/src/api/CMakeLists.txt#L33)

PR #2361, opened 2023-10-24
Limit max loop iterations

gyulaz-htec:
This PR fixes: https://github.com/migraphx-benchmark/AMDMIGraphX/issues/144

SSD-MobileNetV1 has loop iterations with `9223372036854775807` (`uint64_max`), which causes an overflow in later shape_computations.
The intent of the PR is to increase the `max_loop_iterations` to `65535` (`uint16_max`) and limit maximum trip-count with that value.

### DIFF
--- a/src/onnx/include/migraphx/onnx/onnx_parser.hpp
+++ b/src/onnx/include/migraphx/onnx/onnx_parser.hpp
@@ -97,9 +97,9 @@ struct onnx_parser
     shape::dynamic_dimension default_dyn_dim_value = {1, 1};
     std::unordered_map<std::string, std::vector<std::size_t>> map_input_dims;
     std::unordered_map<std::string, std::vector<shape::dynamic_dimension>> map_dyn_input_dims;
-    bool use_dyn_output         = false;
-    bool skip_unknown_operators = false;
-    int64_t max_loop_iterations = 10;
+    bool use_dyn_output          = false;
+    bool skip_unknown_operators  = false;
+    int64_t max_loop_iterations  = 10;
     int64_t limit_max_iterations = std::numeric_limits<uint16_t>::max();
     int64_t opset_version        = 13;
 
--- a/src/onnx/include/migraphx/onnx/onnx_parser.hpp
+++ b/src/onnx/include/migraphx/onnx/onnx_parser.hpp
@@ -100,7 +100,8 @@ struct onnx_parser
     bool use_dyn_output         = false;
     bool skip_unknown_operators = false;
     int64_t max_loop_iterations = 10;
-    int64_t opset_version       = 13;
+    int64_t limit_max_iterations = std::numeric_limits<uint16_t>::max();
+    int64_t opset_version        = 13;
 
     std::unordered_map<std::string, op_func> ops;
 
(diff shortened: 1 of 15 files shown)


src/onnx/parse_loop.cpp
```diff
@@ -54,7 +54,7 @@ struct parse_loop : op_parser<parse_loop>
             auto arg_iters = args.at(0)->eval();
             if(not arg_iters.empty())
             {
-                max_iterations = arg_iters.at<int64_t>();
+                max_iterations = std::min(arg_iters.at<int64_t>(), max_iterations);
```
umangyadav, line 57: I didn't completely understand this issue. For the SSD model, isn't there a loop termination condition, loop trip count won't matter in that case if it is very high?

pfultz2: > I didn't completely understand this issue. For the SSD model, isn't there a loop termination condition, loop trip count won't matter in that case if it is very high? 

Since we don't use dynamic shapes for the loop it will become a tensor of the max size.

src/onnx/include/migraphx/onnx/onnx_parser.hpp
```diff
@@ -99,7 +99,7 @@ struct onnx_parser
     std::unordered_map<std::string, std::vector<shape::dynamic_dimension>> map_dyn_input_dims;
     bool use_dyn_output         = false;
     bool skip_unknown_operators = false;
-    int64_t max_loop_iterations = 10;
+    int64_t max_loop_iterations = std::numeric_limits<uint16_t>::max();
```
pfultz2, line 102: Since it's already 10, I don't think we should increase it.

src/onnx/parse_loop.cpp
pfultz2, line 57: So the model itself is saying the max iterations should be 2^63?

The purpose of `parser.max_iterations` is to provide a default value when one isn't given in the model, it's not meant to clip the max iterations in the model. We probably need a separate flag for this, like `parser.limit_max_iterations`.

gyulaz-htec: Yes, the model has this big number for the loop count.
I will change this and add  `parser.limit_max_iterations` then.

gyulaz-htec:
@pfultz2 I've added `parser.max_iterations`. Do we want to expose it to the python and migraphx API? If yes I can add that as weel.

umangyadav:
> @pfultz2 I've added `parser.max_iterations`. Do we want to expose it to the python and migraphx API? If yes I can add that as weel.

Yes we would need that.

src/onnx/parse_loop.cpp
```diff
@@ -54,7 +54,9 @@ struct parse_loop : op_parser<parse_loop>
             auto arg_iters = args.at(0)->eval();
             if(not arg_iters.empty())
             {
-                max_iterations = arg_iters.at<int64_t>();
+                // cap max_iter because loop uses static shapes with max_iter size and huge numbers
+                // here can cause overflow
+                max_iterations = std::min(arg_iters.at<int64_t>(), parser.limit_max_iterations);
```
umangyadav, line 59: I find it would be better to print message when it is clipping loop iteration to maximum instead of silently clipping it.

gyulaz-htec:
@umangyadav I've addressed your comments and exposed the `limit_max_iterations` setting to the high level c++ and python APIs.

umangyadav (approved):

PR #2362, opened 2023-10-24
fix tests and error cases; allow std::optional for streamutils

CharlieL7:
@pfultz2 please check that the change I made to `streamutils` makes sense to allow `optional` to be used for operator attributes. This should make handling optional attributes better.

Fixes a bug I found in allocate and adds testcases.


src/include/migraphx/op/allocate.hpp
```diff
@@ -62,26 +62,38 @@ struct allocate
 
     shape compute_shape(const std::vector<shape>& inputs) const
     {
-        if(s != shape())
+        if(s)
```
umangyadav, line 65:
```suggestion
        if(s.has_value())
```
Minor Suggestion : For better readability

src/include/migraphx/op/allocate.hpp
```diff
@@ -62,26 +62,38 @@ struct allocate
...
             else
             {
+                if(s->dynamic())
+                {
+                    MIGRAPHX_THROW("ALLOCATE: dynamic shape attribute and no input");
+                }
                 migraphx::check_shapes{inputs, *this, false}.has(0);
             }
-            return s;
+            return *s;
```
umangyadav, line 83:
```suggestion
            return s.value();
```

src/include/migraphx/op/allocate.hpp
```diff
@@ -62,26 +62,38 @@ struct allocate
...
+            {
+                MIGRAPHX_THROW("ALLOCATE: shape and buf_type attributes both not set");
+            }
             migraphx::check_shapes{inputs, *this, false}.has(1).only_dims(1);
             const auto& out_dims = inputs.at(0);
             std::size_t max_val = std::numeric_limits<std::size_t>::max();
             std::vector<shape::dynamic_dimension> dyn_dims(out_dims.lens().at(0),
                                                            shape::dynamic_dimension{0, max_val});
-            return {buf_type, dyn_dims};
+            return {*buf_type, dyn_dims};
```
umangyadav, line 96:
```suggestion
            return {buf_type.value(), dyn_dims};
```

src/include/migraphx/op/allocate.hpp
```diff
@@ -62,26 +62,38 @@ struct allocate
...
+                    MIGRAPHX_THROW("ALLOCATE: dynamic shape attribute and no input");
+                }
                 migraphx::check_shapes{inputs, *this, false}.has(0);
             }
-            return s;
+            return *s;
         }
         else
         {
+            if(not buf_type)
```
umangyadav, line 87:
```suggestion
            if(not buf_type.has_value())
```

src/include/migraphx/op/allocate.hpp
```diff
@@ -94,7 +106,11 @@ struct allocate
         {
             std::vector<std::size_t> output_dims(output_shape.ndim());
             args.at(0).visit([&](auto a) { output_dims.assign(a.begin(), a.end()); });
-            return argument{shape{buf_type, output_dims}};
+            if(s)
```
umangyadav, line 109:
```suggestion
            if(s.has_value())
```

src/include/migraphx/op/allocate.hpp
```diff
@@ -62,26 +62,38 @@ struct allocate
...
+            }
             if(inputs.size() == 1)
             {
                 migraphx::check_shapes{inputs, *this, false}.only_dims(1);
             }
             else
             {
+                if(s->dynamic())
+                {
+                    MIGRAPHX_THROW("ALLOCATE: dynamic shape attribute and no input");
```
umangyadav, line 79: This is inside `else` so technically there could be inputs >=1

src/include/migraphx/streamutils.hpp
```diff
@@ -68,6 +69,13 @@ auto stream_write_value_impl(rank<1>, std::ostream& os, const T& x) -> decltype(
     os << x;
 }
 
+template <class T>
+auto stream_write_value_impl(rank<1>, std::ostream& os, const std::optional<T>& x)
```
pfultz2, line 73: Use `optional<T>` instead of `std::optional<T>`.

src/include/migraphx/op/allocate.hpp
```diff
@@ -48,9 +48,9 @@ namespace op {
  */
 struct allocate
 {
-    shape s{};
+    std::optional<shape> s;
     // for dynamic allocate to set the buffer type
-    shape::type_t buf_type = shape::half_type;
+    std::optional<shape::type_t> buf_type;
```
pfultz2, line 53: These seems mutually exclusive. Maybe we should use `std::variant` instead , but we will need to update the serialization to support variant.

src/include/migraphx/op/allocate.hpp
bpickrel, line 65: This op. only allows 0 or 1 inputs, correct?  Check that here.

src/include/migraphx/op/allocate.hpp
```diff
@@ -48,9 +48,9 @@ namespace op {
  */
```
bpickrel, line 48: Suggested modification:

 - * Either `this.s` or `this.buf_type` must be set to calculate the dynamic output shape at compute
 - * time. 

 * For dynamic shapes, either `this.s` or `this.buf_type` must be set to calculate the dynamic output shape at compute
 * time, but not both.

src/include/migraphx/op/allocate.hpp
bpickrel, line 1: Update copyright date

src/include/migraphx/op/allocate.hpp
CharlieL7: Already checked later on with the check_shapes calls

test/ref/allocate.cpp
```diff
@@ -47,3 +47,21 @@ TEST_CASE(allocate_dyn)
     migraphx::shape sresult{migraphx::shape::float_type, {2, 3, 4, 4}};
     result.visit([&](auto output) { EXPECT(output.get_shape() == sresult); });
 }
+
+TEST_CASE(allocate_dyn1)
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape s{migraphx::shape::int64_type, {4}};
+    migraphx::shape out_shape{migraphx::shape::float_type, {2, 3, 4, 4}};
```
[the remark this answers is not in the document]
CharlieL7: The `allocate` operator with 1 input outputs a dynamic shape

src/include/migraphx/op/allocate.hpp
CharlieL7: check_shapes call later limits it to 0 inputs

pfultz2 (approved):

umangyadav (approved):

PR #2368, opened 2023-10-26
Add missing space in perf report

pfultz2:


src/program.cpp
```diff
@@ -936,7 +936,7 @@ void program::perf_report(std::ostream& os,
     os << std::endl;
 
     os << "Batch size: " << batch << std::endl;
-    os << "Rate: " << rate * batch << "inferences/sec" << std::endl;
+    os << "Rate: " << rate * batch << " inferences/sec" << std::endl;
```
umangyadav, line 939: I liked what chris had suggested. `Rate (inf/sec):`

pramenku: I too agree to go with below what @causten  has suggested. Can we modify this PR and get it merged in develop, master and 6.0 release branch

cc: @umangyadav @pfultz2 

Rate (inf/sec): 1337.84
Total time (ms): 11.9596
Total instructions time (ms): 12.9021
Overhead time (ms): 0.0595123, -0.942473

kahmed10 (approved):

src/program.cpp
pfultz2: We should have the units after the numbers. The scripts already parsing it as such and if we want to make scripting even easier that we should just output json instead because the textual output is not stable.

umangyadav (approved):

src/program.cpp
pramenku: Hi @pfultz2 


1.Couldn't get why if unit in bracket as above is not easier to read. 
If you are talking about standard format then agree, usually we represent any quantity in below format which we have always seen in Physics. :)

Physical quantity(measurement)= magnitude × unit

2. Textual output is not stable. I couldn't get this point.
Can you please elaborate for the awareness

Thanks

pramenku: @causten @pfultz2 Can we get the change merged today in develop and master branch.
Please put in 6.0 branch too.

PR #2369, opened 2023-10-26
Use an older version of numpy for compatibility with Python3.6

ahsan-ca:
numpy==1.21.6 is not compatible with Python 3.6. This downgrades the numpy version
to 1.19.5 for compatibility with Python 3.6.


shivadbhavsar:
> @shivadbhavsar is there a minimum numpy version we need for the torch-migraphx to function correctly?

Is numpy here only used for testing? If so, this will have no effect on torch migraphx. If its needed for functionality past just testing then it might be an issue. Torch migraphx looks for numpy > 1.20

ahsan-ca:
> > @shivadbhavsar is there a minimum numpy version we need for the torch-migraphx to function correctly?
> 
> Is numpy here only used for testing? If so, this will have no effect on torch migraphx. If its needed for functionality past just testing then it might be an issue. Torch migraphx looks for numpy > 1.20

Yes, numpy here is used for testing(in the Python virtual environment) only.

shivadbhavsar (approved):

TedThemistokleous (approved):

umangyadav (approved):

ahsan-ca:
All CI checks pass except the following two:
1. `continuous-integration/jenkins/pr-head` saying `This commit cannot be built`.
2. `Jenkins - clang_release`

@causten when you get a chance, can you please share the error for the Jenkins failures? I am not able to access Jenkins to see the logs. Thanks!

PR #1839, opened 2023-06-15
ASAN build excluding additional files

arvindcheru:
Summary of proposed changes:

- Changes for ASAN Packaging [SWDEV-398093]
- Will enable only required files for ASAN Packaging. (library and License file)

PR #2370, opened 2023-10-26
Cherry Pick ASAN build excluding additional bin files -1839

arvindcheru:
Summary of proposed changes:

- Changes for ASAN Packaging
- Will enable only required files for ASAN Packaging. (library and License file)
- Cherry Pick Changes from https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1839

### DIFF
--- a/src/targets/gpu/hiprtc/CMakeLists.txt
+++ b/src/targets/gpu/hiprtc/CMakeLists.txt
@@ -28,6 +28,10 @@ add_executable(migraphx-hiprtc-driver
 rocm_clang_tidy_check(migraphx-hiprtc-driver)
 target_link_libraries(migraphx-hiprtc-driver PRIVATE migraphx_gpu)
 add_dependencies(migraphx_all_targets migraphx-hiprtc-driver)
-rocm_install_targets(
-    TARGETS migraphx-hiprtc-driver
-)
+
+# For ASAN Enabled Build no need to package binaries
+if( NOT ENABLE_ASAN_PACKAGING )
+  rocm_install_targets(
+      TARGETS migraphx-hiprtc-driver
+  )
+endif()
--- a/src/driver/CMakeLists.txt
+++ b/src/driver/CMakeLists.txt
@@ -55,6 +55,9 @@ if(MIGRAPHX_ENABLE_PYTHON)
     target_compile_definitions(driver PRIVATE MIGRAPHX_ENABLE_PYTHON)
 endif()
 
-rocm_install_targets(
-  TARGETS driver
-)
+# For ASAN Enabled Build no need to package binaries
+if( NOT ENABLE_ASAN_PACKAGING )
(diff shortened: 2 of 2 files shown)


nunnikri (approved):

src/targets/gpu/hiprtc/CMakeLists.txt
```diff
@@ -28,6 +28,10 @@ add_executable(migraphx-hiprtc-driver
...
 target_link_libraries(migraphx-hiprtc-driver PRIVATE migraphx_gpu)
 add_dependencies(migraphx_all_targets migraphx-hiprtc-driver)
-rocm_install_targets(
-    TARGETS migraphx-hiprtc-driver
-)
+
+# For ASAN Enabled Build no need to package binaries
+if( NOT ENABLE_ASAN_PACKAGING )
+  rocm_install_targets(
+      TARGETS migraphx-hiprtc-driver
```
pfultz2, line 35: I think this will make it really slow for ASAN builds since we will compile serially instead of in parallel.

pfultz2:
This should only go into the release branch, it should not go into develoop.

src/targets/gpu/hiprtc/CMakeLists.txt
nunnikri: In our ci the builds are done parallelly. Normal builds are done in a job. And asan builds are done later in a separate job.

causten:
I have back ported this PR to rocm/release-rel-6.0 with a different PR.  We will use 2370 to resolve this for how to resolve this long term.

pfultz2:
> We will use 2370 to resolve this for how to resolve this long term.

This needs to be resolved in rocm-cmake, though.

src/targets/gpu/hiprtc/CMakeLists.txt
pfultz2: I mean when the users are compiling at runtime, this will be very slow. Its already much slower from ASAN itself, but now it will be massively slower(ie 48x).

Why cant we have this executable as part of the ASAN package?

Also, arent users interested in running everything with ASAN? If we remove this for ASAN, then there wont be ASAN coverage for this code, which is run in our release packages.

causten:
> > We will use 2370 to resolve this for how to resolve this long term.
> 
> This needs to be resolved in rocm-cmake, though.

Open a rocm-cmake issue and then this can close

pramenku:
@pfultz2  @causten 
Hope, next steps is done . We need help to get the fix asap for P1S1 issue for 6.1 mainline, SWDEV-430807
Can you please help if plan to keep in rocm-cmake any change.

cc: @nunnikri @arvindcheru

causten:
cmake changes https://github.com/RadeonOpenCompute/rocm-cmake/pull/157

PR #2373, opened 2023-10-26
Remove int8x4 format completely

umangyadav:


src/program.cpp
```diff
@@ -624,7 +624,7 @@ std::string get_migraphx_version()
 program file version is for the data structure or format of the MXR file. Version should be bumped
 if any changes occur to the format of the MXR file.
 */
-const int program_file_version = 7;
+const int program_file_version = 8;
```
pfultz2, line 627: This version shouldn't need to be bumped. We are not changing the format as far as I can see(only changing operators which does not require a version bump).

umangyadav: Reverted. But we would need to bump MIGraphX version now since 6.0 has branched.

TedThemistokleous (approved):

PR #2377, opened 2023-10-27
Echo driver command

kahmed10:
This is a quality of life improvement that echos the command used in the MIGraphX driver. It will echo in the beginning and at the end similar to trtexec.


src/driver/main.cpp
```diff
@@ -811,7 +811,11 @@ struct main_command
 using namespace migraphx::driver; // NOLINT
 int main(int argc, const char* argv[])
 {
-    std::vector<std::string> args(argv + 1, argv + argc);
+    std::vector<std::string> cmd_args(argv, argv + argc);
+    std::string cmd_string = migraphx::to_string_range(cmd_args, " ");
+    std::cout << "Running " << cmd_string << std::endl;
```
umangyadav, line 816:
```suggestion
    std::cout << "Running: " << cmd_string << std::endl;
```

src/driver/main.cpp
```diff
@@ -820,22 +824,23 @@ int main(int argc, const char* argv[])
...
     {
-        m.at(cmd)(argv[0], {args.begin() + 1, args.end()});
+        m.at(cmd_migx)(argv[0], {args.begin() + 1, args.end()});
     }
     else
     {
         run_command<main_command>(argv[0], args);
     }
 
+    std::cout << "command [ " << cmd_string << " ] was successful." << std::endl;
```
umangyadav, line 844:
```suggestion
    std::cout << "command: [ " << cmd_string << " ] was successful." << std::endl;
```

src/driver/main.cpp
umangyadav: Or it could be `[Running]: migraphx-driver perf resnet50.onnx` 

Add prepend `migraphx-driver` to it.

src/driver/main.cpp
umangyadav: `was successful` part does it get print even if it runs `main_command` ? I think we can omit printing `was successful`.

src/driver/main.cpp
causten: I think the [ ] would be saved for a future PR to be the version of the code

src/driver/main.cpp
```diff
@@ -811,7 +811,11 @@ struct main_command
 using namespace migraphx::driver; // NOLINT
 int main(int argc, const char* argv[])
 {
-    std::vector<std::string> args(argv + 1, argv + argc);
```
pfultz2, line 814: Keep this variable the same name.

src/driver/main.cpp
```diff
@@ -811,7 +811,11 @@ struct main_command
...
+    std::vector<std::string> cmd_args(argv, argv + argc);
+    std::string cmd_string = migraphx::to_string_range(cmd_args, " ");
```
pfultz2, line 815: You can do `std::string cmd_string = std::string(argv[0]) + " " + migraphx::to_string_range(args, " ");` here to reuse the `args` vector.

src/driver/main.cpp
```diff
@@ -820,22 +824,23 @@ int main(int argc, const char* argv[])
     }
 
     auto&& m = get_commands();
-    auto cmd = args.front();
+    auto cmd_migx = args.front();
```
pfultz2, line 827: Keep this variable name as `cmd`.

src/driver/main.cpp
kahmed10: I've fixed the logic for this. It won't print for `main_command` anymore.

src/driver/main.cpp
kahmed10: @umangyadav it will point to whatever the first command line argument the user invocated the driver from. Just to make it easier to debug.
@causten I added the version number in brackets now.

src/driver/main.cpp
kahmed10: reverted

kahmed10:
> [ MIGraphX Version: 2.9.0."b6b341466" ] Success: /AMDMIGraphX/build/bin/driver perf /models/ORT/bert_large_uncased_1.onnx --fill1 input_ids --input-dim @input_ids 1 384
> 
> example of what I see. Looks good. The quotes around the commit hash is kinda weird, I don't think you need it, but if you like it, I can live with it :)

This was caused by a change to support Windows that was merged in recently. May need to investigate later if there's a way to avoid this on linux.

PR #2378, opened 2023-10-27
amended License_stamper.py and updated license years

NBolade-224:
If Check Stamp exists then Searches for matching year. If the matching year isn't found then it uses regular expression to replace the outdated year with the current year. 

Makes the assumption that all years are in the following format '2015-\d+ Advanced Micro Devices' I.E. '2015-2022 Advanced Micro Devices'.


causten:
No.  Copyright changes to files occur when code changes occur.  If a file was updated in 2023 then the file can rightfully have the copyright date changed.  I will not be merging a 1000+ PR simply to update the copyright date.

kahmed10:
> No. Copyright changes to files occur when code changes occur. If a file was updated in 2023 then the file can rightfully have the copyright date changed. I will not be merging a 1000+ PR simply to update the copyright date.

We could still probably automate the checking based on git history of each file. That should reduce the amount of changes.

tools/check_stamped.py
```diff
@@ -66,7 +73,17 @@ def needStampCheck(filename: str) -> bool:
                 if hasKeySequence(
                         save,
                         "Advanced Micro Devices, Inc. All rights reserved"):
-                    if debug: print("....Already Stamped: Skipping  file ")
+                    yearOfLastCommit = getYearOfLatestCommit(filename)
```
kahmed10, line 76: You may need to double check that this number isn't getting stuck at 2023. Currently the licensing pass is failing.

PR #2382, opened 2023-10-28
remove support for miopenInt8x4 on Windows

apwojcik:
MIGraphX on Windows is using a reasonably recent MIOpen. The latest MIOpen discontinued support for the `miopenInt8x4` data type. This PR turns off using miopenInt8x4 on Windows and leaves it untouched on Linux.

### DIFF
--- a/src/targets/gpu/include/migraphx/gpu/miopen.hpp
+++ b/src/targets/gpu/include/migraphx/gpu/miopen.hpp
@@ -143,6 +143,7 @@ inline tensor_descriptor make_tensor(const migraphx::shape& os, bool pack = fals
         d = miopenInt32;
     else if(s.type() == shape::int8_type)
     {
+#ifndef _WIN32
         if(pack)
         {
             // update the lens and corresponding strides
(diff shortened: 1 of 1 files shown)


src/targets/gpu/include/migraphx/gpu/miopen.hpp
```diff
@@ -154,6 +155,12 @@ inline tensor_descriptor make_tensor(const migraphx::shape& os, bool pack = fals
         {
             d = miopenInt8;
         }
+#else
+        // MIGraphX on Windows is based on the most recent MIOpen
+        // The support for miopenInt8x4 has been discontinued.
+        (void)pack;
```
pfultz2, line 161: We should probably throw an error if pack is set.

pfultz2:
We are going to remove int8x4 completely in #2373, so I dont think this PR will be needed.

apwojcik:
@pfultz2  @umangyadav, when #2373 will be merged? This is a blocker for Windows since we cannot use older MIOpen.

umangyadav:
> @pfultz2 @umangyadav, when #2373 will be merged? This is a blocker for Windows since we cannot use older MIOpen.

It's been merged now. Can you give it a try ? If it works close this one.

apwojcik:
#2373 fixed the issue.

PR #2385, opened 2023-10-28
suppress warnings while compiling tests on Windows

apwojcik:
The PR suppresses some CRT-related compiler warnings.


test/api/CMakeLists.txt
```diff
@@ -22,14 +22,18 @@
 # THE SOFTWARE.
 #####################################################################################
 function(add_api_test TEST_NAME TEST_SRC TEST_DIR)
+    cmake_parse_arguments(PARSE "" "" "DEPENDS" ${ARGN})
     set(NAME test_api_${TEST_NAME})
     add_executable(${NAME} EXCLUDE_FROM_ALL ${TEST_SRC})
     rocm_clang_tidy_check(${NAME})
-    target_link_libraries(${NAME} migraphx_c migraphx migraphx_all_targets)
```
pfultz2, line 28: Why cant we link in `migraphx` and `migraphx_all_targets` to all the API tests?

apwojcik: Because of the robustness principle - here we are testing C API used from C and CXX applications. If we are linking with everything, we are testing everything and not C API in isolation.

pfultz2: These tests are testing with C++ not C. The `add_c_api_test` tests the pure C API.

pfultz2 (approved):

PR #2386, opened 2023-10-30
Fix gpu::contiguous elimination bug

gyulaz-htec:
When testing `gpu::contiguous` instruction replacability recursive calls to `try_compute_shape` can pick up `gpu::contiguous` instructions as an instruction [output](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/eliminate_contiguous.cpp#L75-L85).
These instructions always passes the `try_compute_shape` check [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/eliminate_contiguous.cpp#L55-L58) because compute_shape of `gpu::contiguous` always returns a standard shape [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/include/migraphx/gpu/contiguous.hpp#L44-L46), thus propagating false `true` values upwards.
For these corner cases we must demand the standard shape of the inputs as well, to avoid ellimination of needed `gpu::contiguous` instructions.

This fixes: https://github.com/migraphx-benchmark/AMDMIGraphX/issues/141#issuecomment-1768010476

cc @TedThemistokleous @causten

### DIFF
--- a/src/eliminate_contiguous.cpp
+++ b/src/eliminate_contiguous.cpp
@@ -41,6 +41,14 @@ static bool try_compute_shape(instruction_ref ins,
                               const std::vector<shape>& inputs,
                               const std::vector<module_ref>& mods)
 {
+    // compute_shape of gpu::contiguous always returns a standard shape
+    // this can propagate false `true` values for non standard shapes in recursive try_compute_shape
+    // calls
+    if(ins->name() == "gpu::contiguous" &&
+       std::any_of(inputs.begin(), inputs.end(), [&](auto& shape) { return not shape.standard(); }))
+    {
+        return false;
+    }
     try
     {
         shape new_shape = ins->get_operator().compute_shape(inputs, mods);


pfultz2:
> thus propagating false true values upwards.

How is it a false `true`? Can you add a unit test for this?

> For these corner cases we must demand the standard shape of the inputs as well, to avoid ellimination of needed gpu::contiguous instructions.

This will make it so we dont remove any contiguous unless they are redundant. We should remove contiguous when the inputs are not standard shape as well if the outputs can handle the non-standard shape.

src/eliminate_contiguous.cpp
```diff
@@ -41,6 +41,14 @@ static bool try_compute_shape(instruction_ref ins,
                               const std::vector<shape>& inputs,
                               const std::vector<module_ref>& mods)
 {
+    // compute_shape of gpu::contiguous always returns a standard shape
+    // this can propagate false `true` values for non standard shapes in recursive try_compute_shape
+    // calls
+    if(ins->name() == "gpu::contiguous" &&
```
pfultz2, line 47: This is checking the wrong name. This should be using `op_name` that is a member variable of the `eliminate_contiguous` class.

gyulaz-htec:
> How is it a false `true`? Can you add a unit test to demonstrate this?

I will try to create one.

> This will make it so we dont remove any contiguous unless they are redundant. We should remove contiguous when the inputs are not standard shape as well if the outputs can handle the non-standard shape.

Is there any API which can tell that an output can handle not standard shape, apart from the API in `eliminate_contiguous`?

src/eliminate_contiguous.cpp
gyulaz-htec: The main reason I was using `ins->name()` instead of `op_name` here, because recursive `try_compute_shape` calls can process `gpu::contiguous` and that's where the issue comes from in my opinion. But I can aggree that this doesn't apply to the idea what you mentioned above:
> We should remove contiguous when the inputs are not standard shape as well if the outputs can handle the non-standard shape.
I will have another iteration on this. First I will start with a unit test, maybe that can highlight the issue better.

pfultz2:
> I will try to create one.

The unit test for this is here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/eliminate_contiguous_test.cpp

It uses `contiguous` instead of `gpu::contiguous` for the test. 

> Is there any API which can tell that an output can handle not standard shape, apart from the API in eliminate_contiguous?

Yea `try_compute_shape` is the function to check this.

gyulaz-htec:
I will close this PR as further investigation highlighted that the issue is not with `eliminate_contiguous`. There is a new issue about the `reshape_lazy` fail where the discussion can be continued: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2422

PR #2390, opened 2023-10-30
Support per-axis quantization

shivadbhavsar:
Reworked the simplify_qdq pass to support:
1. Per-axis quantization (ie. allow 1D scales and zero points)
2. Allow broadcast and transpose ops between dq and quant_op

Leaving PR in draft mode for now as test cases need some work.


src/include/migraphx/matcher.hpp
```diff
@@ -844,13 +857,20 @@ auto skip_broadcasts_converts(Ms... ms)
...
+{
+    return skip(name("broadcast", "multibroadcast", "contiguous", "transpose"))(ms...);
+}
+
 template <class T>
 inline auto has_value(T x, float tolerance = 1e-6)
 {
     return skip_broadcasts_converts(make_basic_pred_matcher([=](instruction_ref ins) {
-        if(ins->name() != "@literal")
+        if(not ins->can_eval())
```
shivadbhavsar, line 870: Using can_eval check rather than literal made more sense to me (and is required by the qdq matcher), but if there is a good reason it was coded this way please let me know. 
Doing it this way seems to be breaking just 1 test case in simplify_algebra (simplify_div_const) since that program now gets further simplified than expected by the test case (this extra simplification is valid though).

shivadbhavsar: Reverting this change since we dont want to propagate constants before its dedicated pass.

src/include/migraphx/matcher.hpp
```diff
@@ -844,6 +857,12 @@ auto skip_broadcasts_converts(Ms... ms)
     return skip(name("broadcast", "multibroadcast", "contiguous", "convert"))(ms...);
 }
 
+template <class... Ms>
+auto skip_broadcasts_transposes(Ms... ms)
```
TedThemistokleous, line 861: add contiguous to name as well . Rename to
`skip_broadcasts_transposes_contiguous`

src/simplify_qdq.cpp
```diff
@@ -45,77 +45,145 @@ std::unordered_set<std::string> get_quantizable_op_names()
...
+        if(dq1->inputs().front()->get_shape().type() != migraphx::shape::int8_type or
+           dq2->inputs().front()->get_shape().type() != migraphx::shape::int8_type)
             return;
 
-        double scale;
-        visit_all(scale1->get_literal(), scale2->get_literal())(
-            [&](const auto s1, const auto s2) { scale = s1.front() * s2.front(); });
+        // Only symmetric quantization supported (ie. non-zero zero_points not allowed)
+        if(not(is_valid_zero_point(zp1) and is_valid_zero_point(zp2)))
+            return;
```
TedThemistokleous, line 135: Have some assert or error here that we throw incase we're using a calibration dataset that doesn't have proper quantization points

src/simplify_qdq.cpp
```diff
@@ -45,77 +45,145 @@ std::unordered_set<std::string> get_quantizable_op_names()
...
 
+        // Only support scalar and 1D scales
+        if(scale1->get_shape().lens().size() != 1 or scale2->get_shape().lens().size() != 1)
```
TedThemistokleous, line 138: Same as above, be verbose if something fails

src/simplify_qdq.cpp
shivadbhavsar: I dont think we want to error out here. Valid quantized graphs can be using affine transform for quantization (in which case zero_points will ne non-zero) but since migraphx does not support that we just dont quantize that op. Maybe we can give a warning. What's the proper way of giving warnings in our compilation passes?

TedThemistokleous:
Heads up @shivadbhavsar, the onnx pass this week will be invalid due to something upstream thats broken but we haven't get integrated to MIGraphX. I may force push a new pass so the rest of the CI wont break since its fixed on upstream Mainline.

src/simplify_qdq.cpp
TedThemistokleous: Maybe its worth just throwing this as part of documentation on quant for now too. I think this comment is good. Thinking about this some more, maybe its worth down the road adding to this to give a list of ops with an affine transformation to give us a case for adding the support down the road.

src/simplify_qdq.cpp
TedThemistokleous: same as be comment above now. Not an issue for this PR

src/simplify_qdq.cpp
```diff
@@ -45,77 +45,145 @@ std::unordered_set<std::string> get_quantizable_op_names()
...
+            out_scale = m.insert_instruction(qop, migraphx::make_op("mul"), s1_bcast, s2_bcast);
         }
         else if(qop->name() == "dot")
         {
-            dq = m.insert_instruction(qop, migraphx::make_op("quant_dot"), qop_args);
+            dq            = m.insert_instruction(qop, migraphx::make_op("quant_dot"), qop_args);
+            auto out_lens = dq->get_shape().lens();
+
+            // For (..., M, N) x (..., N, K) dot, only support cases where quantization axis is M
+            // for input1 and K for input 2
```
TedThemistokleous, line 173: More question, this for all cases of dot or just us due to the above constraints (zero point, symmetry).

TedThemistokleous (commented):
Just rename the one matcher to to skip_broadcasts_transposes_contiguous just to be clear what we're matching by name. Otherwise what you've done makes sense.

TedThemistokleous (approved):

src/simplify_qdq.cpp
shivadbhavsar: This is true for all cases. You cannot write a matrix multiplication xA * yB = zC (where x, y and z are the scales) such that z is a product of x,y if the quantization axis is the N dimension.

src/include/migraphx/matcher.hpp
shivadbhavsar: Done

pfultz2 (approved):

PR #2394, opened 2023-11-01
Handle all slice input variations

CharlieL7:
* This PR makes it so we can handle several of the different variations of inputs to the `slice` operator
* From an ONNX point of view Slice can be called with 3-5 inputs:
  * `Slice(data, starts, ends)`
  * `Slice(data, starts, ends, axes)`
  * `Slice(data, starts, ends, axes, steps)`
* This PR makes it possible to handle any of the inputs, `starts`, `ends`, `axes`, being interchangeably constant or variable.
* This means the MIGX `slice` operator can be called any of these ways (where the variables after the semicolon refer to the operator's attributes:
  * `slice(input); axes, starts, ends set`
  * `slice(input, starts); axes, ends set`
  * `slice(input, ends); starts, axes set`
  * `slice(input, axes); starts, ends set`
  * `slice(input, starts, ends); axes set`
  * `slice(input, starts, axes); ends set`
  * `slice(input, ends, axes); starts set`
  * `slice(input, start, ends, axes); none set`
* I have also changed how `normalize_attributes()` works wrt. the `use_lens` option for dynamic shapes.
  * If it encounters a non-fixed dynamic_dimension `tune_attribute()` will now skip trying to normalize attribute
  * This means that normalization must be done again on the operator attributes if we try to slice over non-fixed dynamic_dimensions
  * I have not updated the 1 input version of `slice` to work over non-fixed dynamic_dimensions. That can be done in a later PR.
  * We could also make the code more efficient by skipping the attribute re-normalization seen in `slice.compute()` for 2+ inputs if it doesn't need it. (Probably going to add quite a bit of complexity to the code)


TedThemistokleous (approved):

causten:
@CharlieL7  Fix Tidy and this can merge

src/include/migraphx/op/slice.hpp
```diff
@@ -36,8 +36,34 @@ namespace migraphx {
...
+ */
+namespace attr_cases {
+inline constexpr std::array<bool, 3> all_set     = {true, true, true};
+inline constexpr std::array<bool, 3> ends_axes   = {false, true, true};
+inline constexpr std::array<bool, 3> starts_axes = {true, false, true};
+inline constexpr std::array<bool, 3> starts_ends = {true, true, false};
+inline constexpr std::array<bool, 3> axes_only   = {false, false, true};
+inline constexpr std::array<bool, 3> ends_only   = {false, true, false};
+inline constexpr std::array<bool, 3> starts_only = {true, false, false};
+inline constexpr std::array<bool, 3> none_set    = {false, false, false};
```
pfultz2, line 50: Move these veriables to the `slice` class(and make them `static`).

pfultz2:
There should be two approvals before merging.

PR #2402, opened 2023-11-01
enable compilation on Windows

apwojcik:
This PR enables the compilation of MIGraphX on Windows. After the merge, Windows MIGraphX can be built from the branch `develop`.


src/CMakeLists.txt
```diff
@@ -28,7 +28,11 @@ include(ROCMInstallTargets)
 include(ROCMPackageConfigHelpers)
 include(RegisterOp)
 include(CheckCXXLinkerFlag)
- 
+
+if(WIN32)
+    # Due to compilation crashing, we need to use type-erased matchers on Windows.
+    add_compile_definitions($<$<COMPILE_LANGUAGE:C,CXX>:MIGRAPHX_USE_TYPE_ERASED_MATCHERS=1>)
```
pfultz2, line 34: I dont think the generator expression is needed here since it applies to both C and C++. Also this could be applied to the migraphx target since this is the component that uses the define: `target_compile_definitions(migraphx INTERFACE MIGRAPHX_USE_TYPE_ERASED_MATCHERS=1)`.

src/onnx/CMakeLists.txt
```diff
@@ -37,7 +41,11 @@ set_target_properties(migraphx_onnx PROPERTIES EXPORT_NAME onnx)
 migraphx_generate_export_header(migraphx_onnx)
 rocm_set_soversion(migraphx_onnx ${MIGRAPHX_SO_VERSION})
 rocm_clang_tidy_check(migraphx_onnx)
-target_link_libraries(migraphx_onnx PRIVATE onnx-proto "-Wl,--exclude-libs,ALL")
+if(WIN32)
+    target_link_libraries(migraphx_onnx PRIVATE onnx-proto)
+else()
+    target_link_libraries(migraphx_onnx PRIVATE onnx-proto "-Wl,--exclude-libs,ALL")
+endif()
```
pfultz2, line 48: You can write it like this since `onnx-proto` is needed for both windows and linux:

```cmake
target_link_libraries(migraphx_onnx PRIVATE onnx-proto)
if(NOT WIN32)
    target_link_libraries(migraphx_onnx PRIVATE "-Wl,--exclude-libs,ALL")
endif()
```

src/CMakeLists.txt
apwojcik: The problem we observe - clang compiler crashing due out of memory - is when we compile CPU and GPU targets only. If I switch the option only for the mentioned targets, the compilation of `migraphx` takes a very long time, especially the linking stage. When I enabled that switch globally, the compilation of crashing targets passed, and the overall compilation time significantly shortened.

pfultz2: Ah sorry, it should be `target_compile_definitions(migraphx PUBLIC MIGRAPHX_USE_TYPE_ERASED_MATCHERS=1)`, this will set it for any cmake target that uses the `migraphx/matchers.hpp` header including `migraphx` itself.

pfultz2 (approved):

PR #2403, opened 2023-11-02
Disable python tests for Python 3.6 temporarily

ahsan-ca:
We need to disable python tests on Python 3.6 temporarily till we remove numpy dependencies from the tests using them.
Different versions of Python are compatible with different versions of numpy, so we need to remove numpy dependency.


TedThemistokleous (approved):

umangyadav (approved):

ahsan-ca:
@causten This PR is ready to be merged in. Jenkins errors are not related to the changes in this PR (I tried re-running them but ran into the same errors: `get_device_id: No device`) .

Once merged, we'd need this backported into rocm6.0. This fixes test failures due to incompatible version of numpy.

pfultz2:
Since we arent fixing the tests then we should make a release note that the python API is not supported on SLES, since we wont be testing on this platform.

test/py/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ function(add_py_venv_fixture FIXTURE_NAME VIRTUAL_ENV_DIR REQUIREMENTS_FILE)
         set(PYTHON_EXECUTABLE ${PYTHON_${PYTHON_VERSION}_EXECUTABLE})
 
         if(NOT TEST py_${PYTHON_VERSION}_${FIXTURE_NAME}_initialize_env)
-            if (NOT (${FIXTURE_NAME} STREQUAL "onnx" AND ${PYTHON_VERSION} STREQUAL ${PYTHON_VERSION_TO_DISABLE_ONNX}))
+            if (NOT (${PYTHON_VERSION} STREQUAL ${PYTHON_VERSION_TO_DISABLE_ONNX}))
```
pfultz2, line 44: We should only disable the test when `MIGRAPHX_DISABLE_VIRTUAL_ENV=Off`.

ahsan-ca: Thanks for the feedback Paul. I will make the changes as you suggested in a follow up PR.

ahsan-ca: I looked at initially disabling the tests using `MIGRAPHX_DISABLE_VIRTUAL_ENV=On` for Python 3.6 but I had some concerns regarding the code flow if we disable it for a particular Python version inside CMakeLists.txt. I can discuss this on Monday and I can proceed based on the feedback.

PR #2404, opened 2023-11-02
stop caching unused models

causten:
In ONNX 1.14.0 PR [4861](https://github.com/onnx/onnx/pull/4861) stopped using external models for testing and went to a very light version of each.  https://github.com/onnx/onnx/blob/rel-1.14.1/onnx/backend/test/case/model/__init__.py#L40

I am removing any hint of download_models.sh since Jenkins is failing a lot recently with download_models.sh timing out and causing Jenkins to fail.


umangyadav (approved):
Nice. This would speed up building dockers

Dockerfile
```diff
@@ -80,6 +80,10 @@ ADD rbuild.ini /rbuild.ini
 # Temporarily install a new cmake until switching to ubuntu 22.04
 RUN pip3 install cmake==3.22.1
 
+# Location where onnx unit tests models are cached
+ENV ONNX_HOME=/.onnx
+RUN mkdir -p $ONNX_HOME/.onnx && chmod 777 $ONNX_HOME/.onnx
```
pfultz2, line 85: This can be removed since we are no longer caching models.

Also, why are you creating the `/.onnx/.onnx` directroy?

pfultz2: It look like it copies these files around to a temporary and then it copies it to this directory. I wonder if we can point `ONNX_HOME` to the onnx files in the python package.

causten: `/.onnx/.onnx` yeah total mistake.  I fixed that

So it turns out ONNX test models still copy the files to the test dir.  What seems to be happening is a permission issue between building the docker (install onnx1.14.1) and running the docker (docker run -t -d -u 1001:1001).  

https://github.com/onnx/onnx/blob/2bb6e5f46f46c71f5ad25c6ee7307885e896f081/onnx/backend/test/runner/__init__.py#L234

PR #2413, opened 2023-11-08
Enable running individual passes in the driver

pfultz2:


src/driver/passes.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <migraphx/simplify_reshapes.hpp>
+
+#include <migraphx/ranges.hpp>
+#include <unordered_map>
+
+namespace migraphx {
+namespace driver {
+inline namespace MIGRAPHX_INLINE_NS {
+
+std::unordered_map<std::string, pass> create_passes_lookup()
```
umangyadav, line 58: Enumerating passes would add more maintenance. Can we create Registry of available passes ?

src/driver/main.cpp
```diff
@@ -123,6 +125,7 @@ struct loader
            ap.append(),
            ap.nargs(2));
         ap(optimize, {"--optimize", "-O"}, ap.help("Optimize when reading"), ap.set_value(true));
+        ap(passes, {"--apply-pass", "-p"}, ap.help("Passes to apply to model"), ap.append());
```
umangyadav, line 128: Need to add this option in documents

src/driver/main.cpp
```diff
@@ -330,6 +333,8 @@ struct loader
                                      migraphx::dead_code_elimination{},
                                  });
         }
+        if(not passes.empty())
```
umangyadav, line 336: Side note : `optimize` option can utilize `optimize_module` passs.

src/driver/passes.cpp
pfultz2: We can try to do that in the future, but we would need to make the passes serializable, but I think thats beyond the scope of this PR.

src/driver/main.cpp
pfultz2: Updated.

umangyadav (approved):

PR #2414, opened 2023-11-08
Port test_gpu.py to not use numpy and re-enable Python tests for Python 3.6

ahsan-ca:
This PR removes numpy dependancy from `test_gpu.py` test and re-enables Python tests for Python 3.6.


test/py/test_gpu.py
```diff
@@ -120,9 +118,8 @@ def test_if_pl():
 
     params = {}
     shapes = p.get_parameter_shapes()
-    params["x"] = np.ones(6).reshape(shapes["x"].lens()).astype(np.float32)
-    params["y"] = np.array([2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0
-                            ]).reshape(shapes["y"].lens()).astype(np.float32)
+    params["x"] = migraphx.fill_argument(migraphx.shape(type='float_type', lens=shapes["x"].lens()), 1)
+    params["y"] = migraphx.fill_argument(migraphx.shape(type='float_type', lens=shapes["y"].lens()), 2.0)
     params["cond"] = np.array([1]).reshape(()).astype(bool)
```
ahsan-ca, line 123: Working on `params["cond"] = np.array([1]).reshape(()).astype(bool)`...

ahsan-ca:
@pfultz2 A couple of things:
1. I don't see `test_numpy.py` being used in `CMakeLists.txt`. Do you think I should delete this test or leave it as is?

2. I have reverted the recent change to disable non-onnx tests for Python 3.6. Now only onnx tests are disabled for Python 3.6 i.e. we do not run onnx tests when we encounter Python 3.6.
Another approach can be to not use virtual envs to run onnx tests for Python 3.6 (use `MIGRAPHX_DISABLE_VIRTUAL_ENV` flag). However, it might not be good to make this change now since, earlier, we had notified the QA team that they could remove any Python related packages since we had moved to use virtual environments. So my concern is if we disable the virtual env (using disable flag) at this time, QA might run into errors because of this. With the release close to being finalized, it might create hurdles in that process.

If we think we should make the change to disable virtual environments for Python onnx We can make this change once 6.0 is released and then let the QA team know that we would need certain packages for Python 3.6 in order to run onnx tests outside the virtual environment. This is provided we would like to disable virtual environments for Python 3.6. If we are ok with keeping the onnx tests disabled for Python 3.6, then we do not need to make any changes. I would love to hear your thoughts on this. Thanks!

FYI @causten This PR will enable running non-onnx tests for Python 3.6 as it removes the numpy dependency in the test that needed different numpy versions for different Python versions.

pfultz2 (approved):

test/py/requirements.txt
```diff
@@ -22,4 +22,3 @@
 # THE SOFTWARE.
 #####################################################################################
 
-numpy==1.21.6
```
umangyadav, line 25: This empty file should be removed.

ahsan-ca: The function `add_py_venv_fixture` uses this requirements file for onnx and non-onnx invocations in `COMMAND ${PYTHON_EXECUTABLE} -m pip install -r ${REQUIREMENTS_FILE})`. Also, for future use in the virtual environments, its better to have a requirements file where packages can be added as needed by tests.

ahsan-ca: If you feel strongly about removing this file, I could add a check in CMakeLists.txt to see if the requirements files exists and only then run `COMMAND ${PYTHON_EXECUTABLE} -m pip install -r ${REQUIREMENTS_FILE})`
In future, if a package needs to be added, the developer can create the requirements file and add in the package.
It might just be easier to leave the empty file, but I do not have a very strong preference about it. So open to your suggestions on this. @umangyadav @pfultz2

causten: yeah lets not keep empty files lying around

ahsan-ca: Sounds good. Done.

umangyadav (approved):

PR #2416, opened 2023-11-08
Fix sphinx documentation

CharlieL7:
Fixes some bugs in the documentation


TedThemistokleous (approved):

kahmed10 (approved):

docs/driver/compile.rst
```diff
@@ -1,5 +1,3 @@
-.. include:: ./driver/read.rst
```
pfultz2, line 1: Why is this removed here?

CharlieL7: Leads to a weird error with the current documentation where it tries to find the read.rst at `../../driver/driver/read.rst`.

pfultz2: Maybe it should just be `read.rst`.

umangyadav (approved):

docs/driver/compile.rst
CharlieL7: Doesn't work, then the `make doc` complains about:
```
../../docs/driver/compile.rst:1: CRITICAL: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: '../../docs/read.rst'.
```

PR #2418, opened 2023-11-09
Update SHA-1 to point past Eigen 3.4 fix

TedThemistokleous:
Force SHA-1 to be one after point in onnxruntime where the upstream changes on the Eigen 3.4 library are fixed. This occurred due to Eigen updating their library which changed the SHA-1 and onxnrt has an expected SHA-1 baked in for the binary they're pulling down during compile. 

For us this means our CI will constantly fail on the onnxruntime step for all our CI

Push was fixed yesturday with these two changes that are off the SHA-1 I've manually updated Onnxrt too.
https://github.com/microsoft/onnxruntime/pull/18301
https://github.com/microsoft/onnxruntime/pull/18308

### DIFF
--- a/test/onnx/.onnxrt-commit
+++ b/test/onnx/.onnxrt-commit
@@ -1,1 +1,1 @@
-b7b8b5b2ce80edb33990c7ae0fedac6ae3c623f4
+7a3da4526f98c9cfc6387a5faa1edeec7d88ef17


TedThemistokleous:
Already updated past eigen issue. Need to update PR i was seeing this failure on.

PR #2428, opened 2023-11-09
[MLIR] Updates needed for general stride support

krzysz00:
This is a companion PR to
https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1312 . The updated commit hash points onto that PR branch, so coordinated merges are advised.

With the above rocMLIR changes, the MLIR MIGraphX dialect now represents both the dimensions and strides of tensors inside MLIR, thus allowing NHWC convolutions to be correctly offloaded.

In this PR, we:
- Remove special handling for the case where non-standard shapes become input to MLIR modules
- Fold broadcast and multibroadcast operations into the input size od MLIR modules
- Update tests
- Add an extra TRACE_MLIR print to help debug crashes in the high-level pipeline


test/gpu/mlir.cpp
```diff
@@ -160,15 +160,38 @@ module {
     EXPECT(verify_mlir(m));
 }
 
+TEST_CASE(conv_nhwc)
```
krzysz00, line 163: For unclear reasons, this testcase fails verification. Do we know if there's something I need to do to make it pass? I'm not sure if the test verifier respects strides

kahmed10: I believe we don't have NHWC convolution support for ref. Ideally you would write the test with the NCHW strides, and then we would have MIGRAPHX_ENABLE_NHWC=1 for the GPU which should insert layouts and run in NHWC.

krzysz00: ... how much work will it be to add NHWC support for ref?

... why is it not there?

(I can, in the meantime, comment out the verify part of this test and leave a comment about why we had to disable it)

pfultz2: > I'm not sure if the test verifier respects strides

What do you mean by verifier? the `verify_args_with_tolerance`?(that actually should work with any layout).  

> I believe we don't have NHWC convolution support for ref.

The ref supports NHWC, it should support any layout, actually.

pfultz2: I think its an issue with `auto_contiguous` and the fact that mlir op doesnt support non standard shape. So it makes the output standard shape but then the final output is standard shape instead of NHWC. I think removing the `standard()` requirement from `check_shapes` should fix the issue for this test.

kahmed10: > > I'm not sure if the test verifier respects strides
> 
> What do you mean by verifier? the `verify_args_with_tolerance`?(that actually should work with any layout).
> 
> > I believe we don't have NHWC convolution support for ref.
> 
> The ref supports NHWC, it should support any layout, actually.

Yeah you're right. Although we don't do any transformations to NHWC on ref if the inputs are NCHW.

krzysz00: The point is, I want to explicitly have a test that makes MLIR create a NHWC convolution and verifies it against the result of doing that same convolution with a CPU reference. I don't know what I need to do do to make that possible, but obviously it's more than "nothing", because uncommenting this call makes the test fail

pfultz2: You need remove the check for standard shape here: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/fuse_mlir.cpp#L97

krzysz00: ... update, found the bug on our end

pfultz2 (approved):

causten:
A lot of models took a performance hit with this PR.  Need to resolve that before merging

PR #2436, opened 2023-11-10
Use double buffer for block_scan

pfultz2:
This avoids incorrect concurrent writes.

### DIFF
--- a/src/targets/gpu/device/include/migraphx/gpu/device/scan.hpp
+++ b/src/targets/gpu/device/include/migraphx/gpu/device/scan.hpp
@@ -47,7 +47,7 @@ __device__ void block_scan(index idx, Op op, T init, ForStride fs, Input input, 
     type x = init;
     fs([&](auto i) {
         int iout = 0;
-        int iin = 1;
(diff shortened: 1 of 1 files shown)


src/targets/gpu/device/include/migraphx/gpu/device/scan.hpp
```diff
@@ -43,24 +43,32 @@ template <index_int N,
 __device__ void block_scan(index idx, Op op, T init, ForStride fs, Input input, Output output)
 {
     using type = decltype(input(deduce_for_stride(fs)));
-    MIGRAPHX_DEVICE_SHARED type buffer[N];
+    MIGRAPHX_DEVICE_SHARED type buffer[2][N];
     type x = init;
     fs([&](auto i) {
+        int iout = 0;
```
umangyadav, line 49: Use index_int

pfultz2: Done

umangyadav (approved):

PR #2442, opened 2023-11-13
Add a tuning limit for mlir

pfultz2:
This adds `MIGRAPHX_MLIR_TUNE_LIMIT` env variable that will limit the number of tunings MLIR will use. This can be used for debugging or figuring out a better limit for quick tuning.

### DIFF
--- a/docs/dev/env_vars.rst
+++ b/docs/dev/env_vars.rst
@@ -260,8 +260,9 @@ Appends to tuning cfg file that could be used with rocMLIR tuning scripts.
 Set to "1", "enable", "enabled", "yes", or "true" to use.
 Do exhaustive tuning for MLIR.
 
-**MIGRAPHX_MLIR_TUNE_LIMIT**
+.. envvar:: MIGRAPHX_MLIR_TUNE_LIMIT
 
+Set to an integer greater than 1.
 Limits the number of solutions that MLIR will use for tuning.
 
 CK vars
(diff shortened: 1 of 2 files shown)


src/targets/gpu/mlir.cpp
```diff
@@ -73,6 +73,7 @@ namespace gpu {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_MLIR);
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_TUNE_EXHAUSTIVE);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_TUNE_LIMIT);
```
umangyadav, line 76: Need to document this inside https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/docs/dev/env_vars.rst

bpickrel (approved):
Update date in copyright notice

src/targets/gpu/mlir.cpp
causten: @pfultz2 can you document this new env please

docs/dev/env_vars.rst
```diff
@@ -260,6 +260,9 @@ Appends to tuning cfg file that could be used with rocMLIR tuning scripts.
 Set to "1", "enable", "enabled", "yes", or "true" to use.
 Do exhaustive tuning for MLIR.
 
+**MIGRAPHX_MLIR_TUNE_LIMIT**
+
+Limits the number of solutions that MLIR will use for tuning.
```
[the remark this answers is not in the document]
pfultz2: That would be `than` not `then`.

umangyadav (approved):

PR #2443, opened 2023-11-14
Add QLinearSigmoid and QLinearLekyRelu op

gyulaz-htec:
[QLinearSigmoid](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.QLinearSigmoid) operator is required for [Faster-RCNN-int8](https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/faster-rcnn/model/FasterRCNN-12-int8.onnx) and [Mask-RCNN-int8](https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/mask-rcnn/model/MaskRCNN-12-int8.onnx) onnx zoo models.

Fixes: https://github.com/migraphx-benchmark/AMDMIGraphX/issues/151 and https://github.com/migraphx-benchmark/AMDMIGraphX/issues/156


src/onnx/parse_qlinearsigmoid.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+Type Constraints
+T : tensor(uint8), tensor(int8)
+Constrain input and output types to 8 bit tensors.
+
+*/
+
+struct parse_qlinearsigmod : op_parser<parse_qlinearsigmod>
+{
+    std::vector<op_desc> operators() const { return {{"QLinearSigmoid"}}; }
```
pfultz2, line 81: You can do `{{"QLinearSigmoid", "sigmoid"}}`

src/onnx/parse_qlinearsigmoid.cpp
```diff
@@ -0,0 +1,129 @@
...
+Y : T
+Output tensor
+
+Type Constraints
+T : tensor(uint8), tensor(int8)
+Constrain input and output types to 8 bit tensors.
+
+*/
+
+struct parse_qlinearsigmod : op_parser<parse_qlinearsigmod>
```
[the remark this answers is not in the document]
gyulaz-htec: There is currently only one other unary qlineary operator in migraphx [QLinearglobalAveragePool](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.QLinearGlobalAveragePool), but that requires more detailed input checking (it has no optional inputs) also calling the underlying migraphx operator for that is more complex than a simple  `add_common_op` call.
On the other hand I'm planning to add `QLinearLeakyRelu` and `QLinearAveragePool` ops which are unary operators with similar requirements. I will try to update this to `parse_qlinearunary`.

gyulaz-htec:
@pfultz2 I've updated the PR added comon parsing of qlinearunary ops and moved them in  the same file. Also Added QLinearLeakyRelu to the batch, because it is also required for ONNX zoo models and is a unary op.

src/onnx/parse_qlinearsigmoid.cpp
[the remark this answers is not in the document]
pfultz2: `QLinearAveragePool` should be handled by a different parser, but the `QLinearLeakyRelu` can reuse the same parse.

src/onnx/parse_qlinearunary.cpp
```diff
@@ -0,0 +1,299 @@
...
+    const auto& in_x = args[0];
+
+    auto sh_x   = in_x->get_shape();
+    auto type_x = sh_x.type();
+    if(type_x != migraphx::shape::int8_type and type_x != migraphx::shape::uint8_type)
+        MIGRAPHX_THROW(opd.op_name + ": unsupported input type");
+}
+
+template <class F>
+instruction_ref parse_qlinearunary(const onnx_parser::node_info& info,
```
pfultz2, line 52: Move this function to a seperate header, similiar to the other helper functions.

src/onnx/parse_qlinearunary.cpp
```diff
@@ -0,0 +1,299 @@
...
+Y : T
+Output tensor
+
+Type Constraints
+T : tensor(uint8), tensor(int8)
+Constrain input and output types to 8 bit tensors.
+
+*/
+
+struct parse_qlinearsigmod : op_parser<parse_qlinearsigmod>
```
pfultz2, line 119: This should be name `parse_qlinearunary`.

src/onnx/parse_qlinearunary.cpp
pfultz2: Also this function should be named `add_qlinearunary` not with `parse_` which is generally reserved for the parser classes.

src/onnx/parse_qlinearunary.cpp
```diff
@@ -0,0 +1,299 @@
...
+{
+    std::vector<op_desc> operators() const { return {{"QLinearSigmoid", "sigmoid"}}; }
```
pfultz2, line 121: This should be `{"QLinearSigmoid", "sigmoid"}, {"QLinearLeakyRelu", "leaky_relu"}}`.

src/onnx/parse_qlinearunary.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    instruction_ref parse(const op_desc& opd,
+                          const onnx_parser& /*parser*/,
+                          const onnx_parser::node_info& info,
+                          const std::vector<instruction_ref>& args) const
+    {
+        check_qlinearunary_inputs(opd, args);
+        return parse_qlinearunary(info, args, [&](instruction_ref dquant_x) {
+            return info.add_instruction(migraphx::make_op(opd.op_name), dquant_x);
```
pfultz2, line 130: This should do:

```cpp
auto op = parser.load(opd.op_name, info);
return info.add_instruction(op, dquant_x);
```

So it works for leaky_relu as well.
```

File: src/onnx/parse_qlinearunary.cpp [hunk 7]
```diff
@@ -0,0 +1,299 @@
...
+Outputs
+Y : T
+Output data tensor from pooling across the input tensor. The output tensor has the same rank as the
+input. with the N and C value keep it value, while the other dimensions are all 1. Type Constraints
+T : tensor(uint8), tensor(int8)
+Constrain input and output types to signed/unsigned int8 tensors.
+
+*/
+
+struct parse_qlinearglobalaveragepool : op_parser<parse_qlinearglobalaveragepool>
```
pfultz2, line 245: This should be name `qlinearpool` and it should be moved to a separate cpp file.

gyulaz-htec: This was previously in the deleted `parse_qlinearglavgpool.cpp`, but if this should be in a separate file I will restore `parse_qlinearglavgpool.cpp`. As a follow up I can change it to `parse_qlinearpooling` when adding QLinearAveragePool.

src/onnx/parse_qlinearunary.cpp
gyulaz-htec: Removed this function because the change you mentioned below makes this solution obsolete.

src/onnx/parse_qlinearunary.cpp
gyulaz-htec: Renamed it to parse_qlinearunary

gyulaz-htec:
@pfultz2 I've made the requested changes, as I've mentioned in one of the comments I've reset the `parse_qlinearglavgpool.cpp` and moving it under parse_qlinearpooling will be done in a seperate PR.

pfultz2 (approved):

gyulaz-htec:
@causten the Performance results seem to be okay now.

PR #2445, opened 2023-11-15
Simplify constant variable input slices

CharlieL7:
* Updates the `simplify_dyn_ops` pass to handle the updated slices
* Depends on #2429 (don't merge this till #2429 is merged)


src/simplify_dyn_ops.cpp
```diff
@@ -61,8 +66,65 @@ struct find_static_2in_broadcasts
...
+    void apply(module& m, const match::matcher_result& mr) const
+    {
+        auto ins       = mr.result;
+        auto inputs    = ins->inputs();
+        auto slice_op  = any_cast<op::slice>(ins->get_operator());
+        auto set_attrs = slice_op.get_set_attributes();
+        std::vector<int64_t> starts_vec;
+        std::vector<int64_t> ends_vec;
+        std::vector<int64_t> axes_vec;
+        if(set_attrs == slice_op.ends_axes)
```
pfultz2, line 91: I think this needs to be `op::slice::ends_axes` to fix the tidy warning.

src/simplify_dyn_ops.cpp
```diff
@@ -77,27 +139,51 @@ struct find_const_3in_slice
...
-            ends_arg.visit([&](auto output) { ends_vec.assign(output.begin(), output.end()); });
-            auto slice_val = ins->get_operator().to_value();
-            auto axes_vec  = slice_val.at("axes").to_vector<int64_t>();
-            m.replace_instruction(
-                ins,
-                make_op("slice", {{"starts", starts_vec}, {"ends", ends_vec}, {"axes", axes_vec}}),
-                inputs.at(0));
+            // slice(data, starts, ends)
+            inputs.at(1)->eval().visit(
+                [&](auto output) { starts_vec.assign(output.begin(), output.end()); });
```
pfultz2, line 151: Not directly related to this PR, but could we add a `to_vector` method to the `raw_data` class that would do this conversion:

```cpp
    template <class T>
    std::vector<T> to_vector() const
    {
        std::vector<T> result;
        visit([&](auto x) { result.assign(x.begin(), x.end()); });
        return result;
    }
```

Then you could just write `inputs.at(1)->eval().to_vector<int64_t>()`.

bpickrel (approved):

src/simplify_dyn_ops.cpp
bpickrel: If you do, comment it with Paul's usage example so other programmers will know how to use it.

pfultz2: >  Do we want to make that enhancement?

If you want to make it in the PR you could or we can open another PR with the update.

pfultz2 (approved):

PR #2448, opened 2023-11-16
Simplify constant of shape

CharlieL7:
* Simplify what we parse the ONNX operator `ConstantOfShape` into a literal for a static shape
* Depends on #2445 (don't merge this till #2445 is merged)


src/simplify_dyn_ops.cpp
```diff
@@ -297,7 +326,8 @@ void simplify_dyn_ops::apply(module& m) const
                         find_static_2in_broadcasts{},
                         find_const_2in_slice{},
                         find_const_3in_slice{},
-                        find_const_4in_slice{});
+                        find_const_4in_slice{},
```
lakhinderwalia, line 329: Should all these find** operations happen in the sequence listed here..? Or can they be any amongst these..? Thanks.

lakhinderwalia (commented):
I find names of these structs: find_***() confusing a little bit, as their apply() methods are not find_ but find_n_replace()..

pfultz2:
> I find names of these structs: find_***() confusing a little bit, as their apply() methods are not find_ but find_n_replace()..

We name all the finders like this. It describes the matcher not the apply method as that would make the name to long `find_const_4in_slice_and_replace_with_static_slice`.

We should be consistent here for now, and we can bikeshed a better name in another PR where we rename the other 70 or so matchers. Also, the call to these classes is called `find_matches` as well not `replace_matches`, but we can bikeshed this in another PR.

src/simplify_dyn_ops.cpp
pfultz2: > Should all these find** operations happen in the sequence listed here..?

What do you mean by "happen"? The `find_matches` will pick the first one to match an instruction(skipping the rest of the matchers for that instruction). And, its still possible that all of these match in a `module` among other instructions. However, none of these matchers overlap in `simplify_dyn_ops` so order doesn't matter here, but it can matter if you have overlapping matchers.

lakhinderwalia: Thanks.

lakhinderwalia (approved):

umangyadav (approved):

pfultz2 (approved):

PR #2455, opened 2023-11-18
FP8 GPU implementation

umangyadav:
Tested on MI300. 

Added verify tests for Math ops, reduce ops and a couple/few pointwise ops. 

Some FP8 tests are commented out because they run into compile errors. Need to fix and enable them


src/targets/gpu/jit/reduce.cpp
```diff
@@ -169,13 +170,20 @@ struct simple_reduce_compiler : compiler<simple_reduce_compiler>
...
+                                                   {"algo", algo},
+                                                   {"transformers", make_transformer_args(vec)},
+                                                   {"preamble", v.get("preamble", std::string{})}});
+        // disable DPP for FP8 for now,, TODO: need to disable for Any FP8 types
+        if(std::any_of(inputs.begin(), inputs.end(), [](const auto& s) {
+               return s.type() == migraphx::shape::fp8e4m3fnuz_type;
+           }))
+        {
+            options.params += "-DMIGRAPHX_HAS_DPP=0 ";
+        }
```
pfultz2, line 186: Is this needed? I would have thought the dpp reductions would've worked.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
```diff
@@ -0,0 +1,530 @@
...
+ * ************************************************************************ */
+
+#ifndef MIGRAPHX_GUARD_KERNELS_FLOAT8_HPP
+#define MIGRAPHX_GUARD_KERNELS_FLOAT8_HPP
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wfloat-equal"
+#pragma clang diagnostic ignored "-Wold-style-cast"
+#endif // __clang__
+#define MIGRAPHX_HIP_DEVICE __device__
```
pfultz2, line 30: Just use `__device__` directly.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
```diff
@@ -0,0 +1,530 @@
...
+ * CTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * ************************************************************************ */
+
+#ifndef MIGRAPHX_GUARD_KERNELS_FLOAT8_HPP
+#define MIGRAPHX_GUARD_KERNELS_FLOAT8_HPP
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wfloat-equal"
+#pragma clang diagnostic ignored "-Wold-style-cast"
```
pfultz2, line 28: Are these needed?

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
```diff
@@ -0,0 +1,530 @@
...
+
+    // constructor from bool
+    explicit constexpr MIGRAPHX_HIP_DEVICE
+    float8(const bool v, rounding_mode rm = rounding_mode::standard, uint32_t rng = 0)
+        : float8((float)(v), rm, rng)
+    {
+    }
+    // convert to float
+// #if defined(__gfx940__) || defined(__gfx941__) || defined(__gfx942__)
+#if 0 // need constexpr operator(). This version can't be constexpr // NOLINT
```
pfultz2, line 216: You can do `if(__builtin_is_constant_evaluated())` to check if its in a `constexpr` context or not. Thats what I do here to use vector operations when not using `constexpr`:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/array.hpp#L101

src/targets/gpu/kernels/include/migraphx/kernels/float8_impl.hpp
```diff
@@ -0,0 +1,344 @@
...
+struct conditional
+{
+    using type = T;
+};
+
+template <class T, class F>
+struct conditional<false, T, F>
+{
+    using type = F;
+};
```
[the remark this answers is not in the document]
umangyadav: Can't include type_traits.hpp since that itself includes `float8.hpp` leading into circular dependency

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
```diff
@@ -195,6 +239,9 @@ MIGRAPHX_DEVICE_MATH_BINARY_FOR(double, min, ::min)
 MIGRAPHX_DEVICE_MATH_BINARY_FOR(migraphx::half, max, ::__hmax)
 MIGRAPHX_DEVICE_MATH_BINARY_FOR(migraphx::half, min, ::__hmin)
 
+MIGRAPHX_DEVICE_MATH_BINARY_FOR_FP8(max, ::max)
+MIGRAPHX_DEVICE_MATH_BINARY_FOR_FP8(min, ::min)
```
pfultz2, line 243: We already define max and min for all data types below. Why is this needed?

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
umangyadav: Thanks. It works made some changes.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
umangyadav: for the `operator ==` , it needs to disable `-Wfloat-equal` . 
Removed old-style-cast. 
Added C++20-extensions.  need that for `asm` inside `constexpr`

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
umangyadav: Done.

src/targets/gpu/jit/reduce.cpp
umangyadav: Removed. It works.

src/targets/gpu/kernels/include/migraphx/kernels/math.hpp
umangyadav: Removed. Not needed

src/targets/gpu/kernels/include/migraphx/kernels/type_traits.hpp
```diff
@@ -246,6 +248,9 @@ constexpr T numeric_max()
         return __FLT_MAX__;
     else if constexpr(is_same<T, migraphx::half>{})
         return __FLT16_MAX__;
+    // TODO: Do it generically for all fp8 types
+    else if constexpr(is_same<T, migraphx::fp8::fp8e4m3fnuz>{})
+        return migraphx::fp8::numeric_limits<T>::max();
```
pfultz2, line 253: Just add a float8 overload for `numeric_max` and `numeric_lowest` in the float8 header.

umangyadav: Done.

src/targets/cpu/dnnl.cpp
```diff
@@ -67,7 +67,8 @@ dnnl::memory::data_type to_dnnl_memory_data_type(shape::type_t t)
     case st::float_type: return dt::f32;
     case st::int32_type: return dt::s32;
     case st::int8_type: return dt::s8;
-    case st::uint8_type: return dt::u8;
+    case st::uint8_type:
+    case st::fp8e4m3fnuz_type: return dt::u8;
```
pfultz2, line 71: This seems wrong, `dt::u8` is an integer not a float. We should throw an error for fp8, when using zendnn. It should already be removed since we just whitelist the supported operators [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/cpu/target.cpp#L69).

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
```diff
@@ -56,27 +56,28 @@ struct avg_pool
     template <class T>
     MIGRAPHX_DEVICE_CONSTEXPR T operator()(T x, T y)
     {
-        return x + y;
+        return static_cast<T>(x + y);
```
pfultz2, line 59: Why is a `static_cast` needed? Shouldn't addition of fp8 produce an fp8?

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
```diff
@@ -56,27 +56,28 @@ struct avg_pool
...
     }
 
     template <class T>
     MIGRAPHX_DEVICE_CONSTEXPR T final(T x, index_int y)
     {
-        return (y == 0) ? 0.0 : (x / y);
+        return (y == 0) ? static_cast<T>(0.0) : static_cast<T>(x / y);
```
pfultz2, line 65: Use `T{0.0}` instead.

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
```diff
@@ -56,27 +56,28 @@ struct avg_pool
...
 {
+    using ret_type = typename Iterator::value_type;
     array<int, 2> low{};
     array<int, 2> high{};
     for(index_int ii = 0; ii < xy.size(); ++ii)
     {
         if(xy[ii] < -1.0f or xy[ii] > dims[ii])
         {
-            return 0;
+            return static_cast<ret_type>(0);
```
pfultz2, line 80: Use `ret_type{0}`.

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
```diff
@@ -92,11 +93,14 @@ MIGRAPHX_DEVICE_CONSTEXPR typename Iterator::value_type bilinear_interpolate(
...
-    float ly                                   = xy[0] - low[0];
-    float lx                                   = xy[1] - low[1];
-    float hy                                   = 1.0f - ly;
-    float hx                                   = 1.0f - lx;
-    array<typename Iterator::value_type, 4> ws = {hy * hx, hy * lx, ly * hx, ly * lx};
+    float ly              = xy[0] - low[0];
+    float lx              = xy[1] - low[1];
+    float hy              = 1.0f - ly;
+    float hx              = 1.0f - lx;
+    array<ret_type, 4> ws = {static_cast<ret_type>(hy * hx),
```
pfultz2, line 100: We should probably just make the array a `float`, and then cast back to the original type at the end.

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
```diff
@@ -199,25 +206,25 @@ __device__ void roialign(const T& x_t, const U& rois_t, const V& ind_t, W& y_t,
...
         {
-            y_t[i] = calc_pooling(offset_x,
-                                  roi_starts,
-                                  bin_size,
-                                  {ph, pw},
-                                  bin_grid_size,
-                                  in_dims,
-                                  s.roi_offset,
-                                  avg_pool{});
+            y_t[i] = static_cast<ytype>(calc_pooling(offset_x,
```
pfultz2, line 209: Use `ytype{}` instead of `static_cast`.

pfultz2: Actually, why is this even needed? `calc_pooling` return `op.final` which you already added casts for.

src/targets/gpu/kernels/include/migraphx/kernels/softmax.hpp
```diff
@@ -43,7 +44,7 @@ __device__ void softmax(Input input1, Output output)
         auto exp_in = r.inner([&](auto x) { return migraphx::exp(x - c); })(input);
         auto batch_sum =
             r.reduce(op::sum{}, 0, [](auto x) { return migraphx::convert<float>(x); })(exp_in);
-        r.inner([&](auto& y, auto x) { y = x / batch_sum; })(output, exp_in);
+        r.inner([&](auto& y, auto x) { y = static_cast<otype>(x / batch_sum); })(output, exp_in);
```
pfultz2, line 47: `otype{x / batch_sum}`.

pfultz2: Actually, thinking about it more, this should just call the `implicit_conversion` function. We probably need to update it to use the explicit constructor.

src/targets/cpu/dnnl.cpp
umangyadav: fp8 is eliminated through eliminiate_dtype pass with converts to float.  
but later simplify_algebra is removing convert and turning test back to fp8. 

e.g. 

```
290 Validate ...
291
292 Pass: dead_code_elimination
293 module: "main"
294 z = @param:z -> fp8e4m3fnuz_type, {1, 2}, {2, 1}, target_id=0
295 y = @param:y -> fp8e4m3fnuz_type, {3, 2}, {2, 1}, target_id=0
296 x = @param:x -> fp8e4m3fnuz_type, {2, 2}, {2, 1}, target_id=0
297 @3 = convert[target_type=2](x) -> float_type, {2, 2}, {2, 1}, target_id=0
298 @4 = convert[target_type=2](y) -> float_type, {3, 2}, {2, 1}, target_id=0
299 @5 = convert[target_type=2](z) -> float_type, {1, 2}, {2, 1}, target_id=0
300 @6 = concat[axis=0](@3,@4,@5) -> float_type, {6, 2}, {2, 1}, target_id=0
301 @7 = convert[target_type=12](@6) -> fp8e4m3fnuz_type, {6, 2}, {2, 1}, target_id=0
302
303
304 Pass: simplify_algebra
305 z = @param:z -> fp8e4m3fnuz_type, {1, 2}, {2, 1}, target_id=0
306 y = @param:y -> fp8e4m3fnuz_type, {3, 2}, {2, 1}, target_id=0
307 x = @param:x -> fp8e4m3fnuz_type, {2, 2}, {2, 1}, target_id=0
308 @3 = concat[axis=0](x,y,z) -> fp8e4m3fnuz_type, {6, 2}, {2, 1}, target_id=0
309 @4 = convert[target_type=2](@3) -> float_type, {6, 2}, {2, 1}, target_id=0
310 @5 = convert[target_type=12](@4) -> fp8e4m3fnuz_type, {6, 2}, {2, 1}, target_id=0

```

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
umangyadav: Done

src/targets/gpu/kernels/include/migraphx/kernels/softmax.hpp
umangyadav: changed it to `otype{x/batch_sum}`. 

`implict_conversion` is the reason i had to add static_cast in many places. It can't call to explict constructors.

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
umangyadav: Removed. Thanks.

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/softmax.hpp
umangyadav: I need to use static_cast otherwise it complains about narrowing conversion for __Float16. 

```
In file included from /tmp/comgr-491967/input/main.cpp:3:
/tmp/comgr-491967/include/migraphx/kernels/softmax.hpp:47:50: error: non-constant-expression cannot be narrowed from type 'type' (aka 'float') to 'otype' (aka '_Float16') in initializer list [-Wc++11-narrowing]
        r.inner([&](auto& y, auto x) { y = otype{x / batch_sum}; })(output, exp_in);
```

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/softmax.hpp
pfultz2: I mean the `implicit_conversion` function [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/vec.hpp#L215). You would need to update it [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/vec.hpp#L210) to use `U(x)` instead of `{x}`.

src/simplify_algebra.cpp
```diff
@@ -653,7 +653,7 @@ struct find_concat_op
     static bool is_valid_op(const operation& op)
     {
         return contains({"broadcast", "multibroadcast"}, op.name()) or
-               op.attributes().contains("pointwise");
+               (op.attributes().contains("pointwise") and op.name() != "convert");
```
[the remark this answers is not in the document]
umangyadav: convert is pointwise op. Do we even have horizontal fusion for "convert" ? I don't see how that would helpful.

src/targets/gpu/kernels/include/migraphx/kernels/softmax.hpp
umangyadav: I tried that with `U{x}` 
but for the _FLoat16 it requires explicit cast still. 

tmp/comgr-55f84c/include/migraphx/kernels/vec.hpp:210:18: error: non-constant-expression cannot be narrowed from type 'float' to '_Float16' in initializer list [-Wc++11-narrowing]
        return U{x};
                 ^

src/simplify_algebra.cpp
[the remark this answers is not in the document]
umangyadav: Yes it does skip fusion of converts.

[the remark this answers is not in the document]
umangyadav: Added test just in case.

test/verify/test_fmod_mod.cpp
```diff
@@ -71,3 +71,4 @@ struct test_mod : verify_program<test_mod>
         return p;
     }
 };
+// TODO: check if requires FP8 test
```
TedThemistokleous, line 74: It appears we can do fmod with different types as long as fp8 follows is arithmetic
https://cplusplus.com/reference/type_traits/is_arithmetic/

https://cplusplus.com/reference/cmath/fmod/

test/verify/test_acosh.cpp
```diff
@@ -23,20 +23,23 @@
...
-        auto* mm = p.get_main_module();
-        migraphx::shape s{migraphx::shape::float_type, {16}};
+        auto* mm                      = p.get_main_module();
+        migraphx::shape::type_t dtype = migraphx::shape::get_type<CType>();
+        migraphx::shape s{dtype, {16}};
         auto x       = mm->add_parameter("x", s);
-        auto min_val = mm->add_literal(1.1f);
-        auto max_val = mm->add_literal(100.0f);
+        auto min_val = mm->add_literal(migraphx::literal{migraphx::shape{dtype}, {1.1}});
+        auto max_val = mm->add_literal(migraphx::literal{migraphx::shape{dtype}, {100.0}});
```
TedThemistokleous, line 42: There a reason why you dropped the  .f here vs some of the other test? Does this need to be converted from double->fp8?

test/verify/test_atanh.cpp
```diff
@@ -23,20 +23,24 @@
...
-        auto* mm = p.get_main_module();
-        migraphx::shape s{migraphx::shape::float_type, {16}};
+        auto* mm                      = p.get_main_module();
+        migraphx::shape::type_t dtype = migraphx::shape::get_type<CType>();
+        migraphx::shape s{dtype, {16}};
         auto x       = mm->add_parameter("x", s);
-        auto min_val = mm->add_literal(-0.95f);
-        auto max_val = mm->add_literal(0.95f);
+        auto min_val = mm->add_literal(migraphx::literal{migraphx::shape{dtype}, {-0.95f}});
+        auto max_val = mm->add_literal(migraphx::literal{migraphx::shape{dtype}, {0.95f}});
```
TedThemistokleous, line 43: Same as my above comment, do we need to drop the "f" here for these decimal values and do  a double->fp8 or should the conversion be from float -> fp8 or it doesn't matter?

test/verify/test_acosh.cpp
umangyadav: No reason.  I probably removed it because literal also had dtype associated. therefore it will get interpreted based on it. 
added back.

test/verify/test_atanh.cpp
umangyadav: Doesn't matter.

test/verify/test_fmod_mod.cpp
umangyadav: Yes it should work. 
But currently for some reason even `half_type` is breaking.  I'll have to check.

umangyadav: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2488

Not sure what is going on with `HalfType`.  Remainder seems to be returning Float vec. 

cc : @pfultz2

umangyadav: Added test for `fp8` `fmod` and `mod`

src/targets/gpu/kernels/include/migraphx/kernels/ops.hpp
```diff
@@ -118,7 +118,7 @@ struct highest
     template <class T>
     constexpr operator T() const
     {
-        return numeric_max<vec_type<T>>();
+        return numeric_max<vec_type<T>, void>();
```
TedThemistokleous, line 121: Is there a numeric_min we also have to adjust here?

umangyadav: numeric_min is negative of numeric_max. 
It is already defined in type_traits.hpp

TedThemistokleous: ok perfect

test/verify/test_fmod_mod.cpp
TedThemistokleous: I think we observed this too when we were doing mod operator with the fmod flag, i think we did determine things were being cast to float instead of half, as part of the half ops since it was just reusing the std::fmod call still.

test/gpu/jit.cpp
```diff
@@ -348,18 +348,18 @@ TEST_CASE(compile_math)
...
+        if(contains({migraphx::shape::bool_type, migraphx::shape::tuple_type}, t))
             continue;
         auto name = migraphx::shape::cpp_type(t);
         if(t == migraphx::shape::half_type)
             name.insert(0, "migraphx::");
         data_types.push_back(name);
-        migraphx::transform(vec_sizes, std::back_inserter(data_types), [&](auto i) {
-            return "migraphx::vec<" + name + ", " + std::to_string(i) + ">";
-        });
+        if(t != migraphx::shape::fp8e4m3fnuz_type)
```
TedThemistokleous, line 357: Naive question why skip this for fp8? More stuff to add in jit later?

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
```diff
@@ -92,15 +92,15 @@ MIGRAPHX_DEVICE_CONSTEXPR typename Iterator::value_type bilinear_interpolate(
...
-    float ly                                   = xy[0] - low[0];
-    float lx                                   = xy[1] - low[1];
-    float hy                                   = 1.0f - ly;
-    float hx                                   = 1.0f - lx;
-    array<typename Iterator::value_type, 4> ws = {hy * hx, hy * lx, ly * hx, ly * lx};
+    float ly           = xy[0] - low[0];
+    float lx           = xy[1] - low[1];
+    float hy           = 1.0f - ly;
+    float hx           = 1.0f - lx;
+    array<float, 4> ws = {hy * hx, hy * lx, ly * hx, ly * lx};
```
TedThemistokleous, line 99: Why change this from Iterator::value_type to float? Is the expectation that this will get converted down to fp8 easier?

umangyadav: hy, hx, lx , ly all are float. Therefore better to do comptuation in float and then convert final result to expected return type.

test/gpu/jit.cpp
umangyadav: fp8 doesn't have vectorization support yet. Added a comment.

src/include/migraphx/bit_cast.hpp
```diff
@@ -21,18 +21,23 @@
...
 
 // NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
 #define MIGRAPHX_CONST_FOLD(x) (__builtin_constant_p(x) ? (x) : (x))
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
-template <typename To, typename From>
+template <typename To,
+          typename From,
+          MIGRAPHX_REQUIRES(std::is_trivially_copyable_v<To>and std::is_trivially_copyable_v<From>)>
```
pfultz2, line 40: I would prefer it written as `std::is_trivially_copyable<To>{} and std::is_trivially_copyable<From>{}` to be consistent.

src/targets/cpu/lowering.cpp
```diff
@@ -376,13 +375,24 @@ struct cpu_apply
         // Apply these operators first so the inputs can be const folded
         for(auto it : iterator_for(*modl))
         {
-            if(it->name() == "pow")
+            // skip lowering if input has fp8 as one of the inputs since oneDNN doesn't have fp8
+            // supported yet
+            if(it->name() == "pow" and
+               (std::none_of(it->inputs().begin(), it->inputs().end(), [](const auto& i) {
+                   return i->get_shape().type() == migraphx::shape::fp8e4m3fnuz_type;
+               })))
```
pfultz2, line 383: Can you just do `if(std::any_of(...)) continue;` like you did below?

src/targets/gpu/kernels/include/migraphx/kernels/ops.hpp
pfultz2, line 121: Why do you need `void` here?

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
```diff
@@ -0,0 +1,564 @@
...
+
+} // namespace fp8
+
+// NOLINTNEXTLINE
+#define MIGRAPHX_FP8_MIN_MAX(T)                  \
+    template <>                                  \
+    constexpr T numeric_max<T, void>()           \
+    {                                            \
+        return fp8::numeric_limits<T>::max();    \
+    }                                            \
```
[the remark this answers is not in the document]
umangyadav: Doesn't work it, It would be considered as redefinition of `numeric_max()`. I need to speicalize `numeric_max` template that is defined in `type_traits.hpp`.

src/targets/cpu/lowering.cpp
umangyadav: Done. 
But i don't see why that would make difference.

src/include/migraphx/bit_cast.hpp
umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
```diff
@@ -0,0 +1,564 @@
...
+    template <>                                  \
+    constexpr T numeric_lowest<T>()              \
+    {                                            \
+        return fp8::numeric_limits<T>::lowest(); \
+    }
```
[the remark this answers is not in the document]
umangyadav: removed it. It it taken care of by `numeric_lowest` inside type_traits.

test/simplify_algebra_test.cpp
```diff
@@ -1017,6 +1017,40 @@ TEST_CASE(simplify_concat_add_relu_broadcast_same_axis)
...
+        auto concat  = m2.add_instruction(migraphx::make_op("concat", {{"axis", 0}}), x, y);
+        auto concath = m2.add_instruction(
+            migraphx::make_op("convert",
+                              {{"target_type", migraphx::to_value(migraphx::shape::half_type)}}),
+            concat);
+        m2.add_instruction(pass_op{}, concath);
+    }
+    EXPECT(m1 == m2);
+}
+
```
CharlieL7, line 1053: Why this test case in this PR?

CharlieL7 (commented):
What's the usage of `implicit_conversion`?
Looks like a lot of the kernels were changed to be more explicit about the types.

test/simplify_algebra_test.cpp
umangyadav: It shouldn't be in this PR ideally. 

But i was trying to make DNNL work with FP8 and as part of that trial. i made some changes in simplify_algebra that was pessimizing convert fusion and none of the unit-tests were catching it. 

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2455#discussion_r1408490543

src/targets/cpu/dnnl.cpp
```diff
@@ -68,6 +68,7 @@ dnnl::memory::data_type to_dnnl_memory_data_type(shape::type_t t)
     case st::int32_type: return dt::s32;
     case st::int8_type: return dt::s8;
     case st::uint8_type: return dt::u8;
+    case st::fp8e4m3fnuz_type:
```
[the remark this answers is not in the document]
umangyadav: This would have been handled through default case. But would  be good to have better message. Commited suggested changes. Thanks.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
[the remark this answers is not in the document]
pfultz2: Ok then you should be able to specialize it like this:

```cpp
template <migraphx::fp8::f8_type T = migraphx::fp8::f8_type::fp8, bool FNUZ, class Enable=void>
constexpr float8<T, FNUZ> numeric_max<float8<T, FNUZ>, Enable>()
{
    return fp8::numeric_limits<float8<T, FNUZ>>::max();
}
```

That should work for all fp8 types. Plus the `class Enable=void` wont require us to pass `void` into the function.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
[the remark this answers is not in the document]
pfultz2: Ok, you can specialize this using `float8<T, FNUZ>` template.

umangyadav:
> What's the usage of implicit_conversion?
Looks like a lot of the kernels were changed to be more explicit about the types. 

From what i understand, some of the internal functions/kernels return output in different type compared to what is desired from  JIT kernel's output type.  and implicit conversion takes care of such types mismatches. 

Float8 constructors are marked as explicit to avoid implict conversions and therefore had to call constructor explicitly in many places explicitly.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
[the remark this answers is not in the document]
umangyadav: `float8<T, FNUZ>` Partial specialization that way is not allowed. Runs into compilation errors.

src/targets/gpu/kernels/include/migraphx/kernels/roialign.hpp
TedThemistokleous: Ah I see you want to preserve precision till the last convert. Got it.

src/targets/cpu/lowering.cpp
pfultz2: It makes it easier to add operators to the first loop, and its consistent with the second loop.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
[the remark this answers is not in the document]
umangyadav: Made changes.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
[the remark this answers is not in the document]
umangyadav: Made changes.

src/targets/gpu/kernels/include/migraphx/kernels/ops.hpp
umangyadav: Made changes.

src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp
```diff
@@ -0,0 +1,566 @@
...
+    static constexpr __device__ fp8e4m3fnuz quiet_NaN()
+    {
+        return fp8e4m3fnuz(0x80, fp8e4m3fnuz::from_bits());
+    }
+
+    static constexpr __device__ fp8e4m3fnuz max()
+    {
+        return fp8e4m3fnuz(0x7F, fp8e4m3fnuz::from_bits());
+    }
+    // this is min value that is not DeNorm. DeNorm min is 0x01
```
TedThemistokleous, line 448: Say this is denormalized instead of denorm. Don't abbreviate it.

TedThemistokleous: Just make a comment at the top or something. DeNorm = Denormalized so you can abbreviate it. Not a make or break though.

umangyadav: Done.

src/targets/gpu/kernels/include/migraphx/kernels/float8_impl.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+// NOLINTBEGIN
+template <int Wm, int We, typename T, bool NegativeZeroNan, bool Clip>
+__device__ constexpr uint8_t cast_to_f8(T f_x, bool stoch = false, uint32_t rng = 0)
+{
+    constexpr bool is_float = true;
+    // half is not supported for now
+    constexpr bool is_half = false;
+    static_assert(Wm + We == 7, "Wm+We==7");
+    static_assert(is_float or is_half, "Only float can be cast to f8");
```
TedThemistokleous, line 45: If half isn't supported  for the cost_to_fp8 we shouldn't include it as a passing condition for the static assert

TedThemistokleous: Disregard

TedThemistokleous (approved):
LGTM. No more questions/concerns and looks like you've handled all comments

pfultz2 (approved):
Looks good! We should probably look into having an env variable to enable fp8 emulation since its probably much slower on unsupported hardware then just using fp32 or fp16 directly.

TedThemistokleous:
> Looks good! We should probably look into having an env variable to enable fp8 emulation since its probably much slower on unsupported hardware then just using fp32 or fp16 directly.

like this idea too but lets get this in so we can hammer the others out. Good for a smaller final PR

PR #2457, opened 2023-11-19
License stamper update

NBolade-224:


kahmed10:
It might not be passing the licensing check because the fork commit years are somehow all in 2023.

causten:
I love that you are trying to contribute, but just remember I'm not merging code changes for code change sake.  Copyrights do not need to change unless significant changes have occurred to the file.

NBolade-224:
> It might not be passing the licensing check because the fork commit years are somehow all in 2023.

You're right (not sure why it does that though). The work around I put in place is to have the check_stamp.py only look at files that have been modified/created in comparison to the up-to-date version of the MIGraphX develop branch (git diff --name-only --diff-filter=d FETCH_HEAD). 

The script then makes sure that all these files have the current year in the license stamp, with the logic being that any file modified (no matter how small the change) or created should have its license stamp year updated. 

============================================================

> I love that you are trying to contribute, but just remember I'm not merging code changes for code change sake. Copyrights do not need to change unless significant changes have occurred to the file.

Its no problem! I'm happy to help and I understand. The current updates are files that have a commit year of 2023 but still have the 2022 year in the license stamp. I amended the license stamper to check the commit year of all the files and then update the license year to the commit year for each file (only if the commit year is greater than 2022, since anything before that will already have the 2022 year stamp). 

If preferred, I can have the license_stamper.py use the same logic as the (newly amended) check_stamp.py instead. In which case, the license_stamper.py will only look for files that have changed in comparison to the latest develop branch of MIGraphX and then update the license stamp year in those files to the current year. 

I can also implement a 2nd filter based on lines added/deleted so that only files that have more than X amount of additions/deletes are updated instead. 

Which one would be best to implement? (If the latter, how many lines deleted or added should be the trigger to update the year?)

TedThemistokleous:
I think this is sufficient based on this condition

_The current updates are files that have a commit year of 2023 but still have the 2022 year in the license stamp. I amended the license stamper to check the commit year of all the files and then update the license year to the commit year for each file (only if the commit year is greater than 2022, since anything before that will already have the 2022 year stamp)._

I think outside of the merge conflicts this captures the changes done in 2023

TedThemistokleous (changes requested):
Changes look good. Fix your merge conflicts so we can ensure this runs in CI correctly.

kahmed10 (approved):

PR #2460, opened 2023-11-20
Add concat fusions

pfultz2:
This will fuse the inputs to concat if they are pointwise operators.


src/fuse_concat.cpp
```diff
@@ -0,0 +1,172 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
kahmed10, line 4: Is the old license stamper script being run? The date should be 2023

src/include/migraphx/fuse_concat.hpp
```diff
@@ -0,0 +1,43 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
kahmed10, line 4: same as the other file

test/fuse_concat.cpp
```diff
@@ -0,0 +1,152 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
kahmed10, line 4: same as the other file

test/verify/test_pooling_add_concat_relu.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
kahmed10, line 4: same as the other file

src/include/migraphx/stringutils.hpp
```diff
@@ -177,6 +177,18 @@ inline std::string interpolate_string(const std::string& input,
         std::move(end));
 }
 
+inline std::string to_c_id(const std::string& name, char rep = '_')
```
TedThemistokleous, line 180: to_c_id = to_concat_id?

src/targets/gpu/target.cpp
```diff
@@ -167,6 +168,8 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         dead_code_elimination{},
         enable_pass(not enabled(MIGRAPHX_DISABLE_REDUCE_FUSION{}), fuse_reduce{}),
         dead_code_elimination{},
+        fuse_concat{},
```
TedThemistokleous, line 171: Asking the question here more out of curiosity. I assume order doesn't matter here for fuse_pointwise vs fuse_concat? Is there a benefit to swapping order ever around the fuse_reduce?

src/include/migraphx/stringutils.hpp
pfultz2: No, it converts the string to a C identifier, hence `to_c_id`. It was moved from module.cpp which was used to print the IR. Its moved here so it can also be used for codegen to ensure that all the variables and template parameters we have are all valid C identifiers as well.

src/targets/gpu/target.cpp
pfultz2: It needs to run after `fuse_pointwise` otherwise there will be no pointwise modules to fuse with. 

> Is there a benefit to swapping order ever around the fuse_reduce?

I dont think so. In either case we will have two kernels that need to be run if these two passes overlap.

TedThemistokleous (approved):

TedThemistokleous:
Fix CI but otherwise I get the idea.

PR #2469, opened 2023-11-23
fix bugs found in the process class on Windows

apwojcik:
While working on the enablement of tests, we discovered a few issues in the `process` class on Windows. This PR fixes them and makes the HIPRTC driver Windows-compatible, too.
In contrast to Linux, the `CreateProcess` function does not start a process in a separate command shell session on Windows. Hence, we cannot use `cd` to change to the provided working directory. Instead, we need to use the `CreateProcess` function argument. Furthermore, the process being spawned is not the `cmd` provided, but `cd`, and since on Windows, `cd` is a shell command and not an application, all is failing miserably.

### DIFF
--- a/src/targets/gpu/hiprtc/main.cpp
+++ b/src/targets/gpu/hiprtc/main.cpp
@@ -31,10 +31,31 @@
 #include <iostream>
 #include <cstring>
 
+#ifdef _WIN32
+// cppcheck-suppress definePrefix
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#endif
+
 std::vector<char> read_stdin()
 {
     std::vector<char> result;
+#ifdef _WIN32
+    HANDLE std_in = GetStdHandle(STD_INPUT_HANDLE);
+    if(std_in == INVALID_HANDLE_VALUE)
+        MIGRAPHX_THROW("STDIN invalid handle (" + std::to_string(GetLastError()) + ")");
+    constexpr std::size_t BUFFER_SIZE = 1024;
+    DWORD bytes_read;
+    TCHAR buffer[BUFFER_SIZE];
+    for(;;)
+    {
+        BOOL status = ReadFile(std_in, buffer, BUFFER_SIZE, &bytes_read, nullptr);
+        if(status == FALSE or bytes_read == 0)
+            break;
 
+        result.insert(result.end(), buffer, buffer + bytes_read);
+    }
+#else
     std::array<char, 1024> buffer;
     std::size_t len = 0;
     while((len = std::fread(buffer.data(), 1, buffer.size(), stdin)) > 0)
@@ -44,6 +65,7 @@ std::vector<char> read_stdin()
 
         result.insert(result.end(), buffer.data(), buffer.data() + len);
     }
+#endif
     return result;
 }
 
--- a/src/process.cpp
+++ b/src/process.cpp
@@ -88,21 +88,38 @@ int exec(const std::string& cmd, std::function<void(process::writer)> std_in)
 
 constexpr std::size_t MIGRAPHX_PROCESS_BUFSIZE = 4096;
 
+enum class direction
+{
+    input,
+    output
+};
+
+template <direction dir>
 class pipe
 {
     public:
-    explicit pipe(bool inherit_handle = true)
+    explicit pipe()
     {
         SECURITY_ATTRIBUTES attrs;
         attrs.nLength              = sizeof(SECURITY_ATTRIBUTES);
-        attrs.bInheritHandle       = inherit_handle ? TRUE : FALSE;
+        attrs.bInheritHandle       = TRUE;
         attrs.lpSecurityDescriptor = nullptr;
 
         if(CreatePipe(&m_read, &m_write, &attrs, 0) == FALSE)
             throw GetLastError();
 
-        if(SetHandleInformation(&m_read, HANDLE_FLAG_INHERIT, 0) == FALSE)
-            throw GetLastError();
+        if(dir == direction::output)
+        {
+            // Do not inherit the read handle for the output pipe
+            if(SetHandleInformation(m_read, HANDLE_FLAG_INHERIT, 0) == 0)
+                throw GetLastError();
+        }
+        else
+        {
+            // Do not inherit the write handle for the input pipe
+            if(SetHandleInformation(m_write, HANDLE_FLAG_INHERIT, 0) == 0)
+                throw GetLastError();
+        }
     }
 
     pipe(const pipe&)            = delete;
@@ -112,10 +129,36 @@ class pipe
 
     ~pipe()
     {
-        CloseHandle(m_read);
-        m_read = nullptr;
-        CloseHandle(m_write);
-        m_write = nullptr;
+        if(m_write != nullptr)
+        {
+            CloseHandle(m_write);
+        }
+        if(m_read != nullptr)
+        {
+            CloseHandle(m_read);
+        }
+    }
+
+    bool close_write_handle()
+    {
+        auto result = true;
+        if(m_write != nullptr)
+        {
+            result  = CloseHandle(m_write) == TRUE;
+            m_write = nullptr;
+        }
+        return result;
+    }
+
+    bool close_read_handle()
+    {
+        auto result = true;
+        if(m_read != nullptr)
+        {
+            result = CloseHandle(m_read) == TRUE;
+            m_read = nullptr;
+        }
+        return result;
     }
 
     std::optional<std::pair<bool, DWORD>> read(LPVOID buffer, DWORD length) const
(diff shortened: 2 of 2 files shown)


src/targets/gpu/hiprtc/main.cpp
```diff
@@ -31,10 +31,31 @@
...
+    TCHAR buffer[BUFFER_SIZE];
+    for(;;)
+    {
+        BOOL status = ReadFile(std_in, buffer, BUFFER_SIZE, &bytes_read, nullptr);
+        if(status == FALSE or bytes_read == 0)
+            break;
 
+        result.insert(result.end(), buffer, buffer + bytes_read);
+    }
+#else
```
pfultz2, line 58: Why is this needed? The below code should work on windows.

apwojcik: It is not working. IPC on Windows differs from Linux. Pipes on Windows are sections of shared memory and not artificial I/O channels. The only way to make it work is to use `_pipe` or `_popen` in MIGraphX, which emulates POSIX functions on Windows. However, as we already discussed, `_pipe`, `_popen`, etc., have severe limitations and only can be used in console applications. However, most use cases are Windows programs, Windows Runtime applications, etc. According to the MSFT documentation, in such a scenario, the functions return an invalid file pointer that causes the program to stop responding indefinitely. This is what we are experiencing here, for example, when executing `test_gpu_quantization`.

pfultz2: > IPC is not part of the C/C++ standard but is OS-specific based on the OS standard implemented. 

This is just using `stdin` which is part of standard C/C++. Why does`stdin` not read from `GetStdHandle(STD_INPUT_HANDLE)`? From the documentation [here](https://learn.microsoft.com/en-us/windows/console/getstdhandle) they look like they are reading from the same place. Is there another documentation that explains why and what `stdin` actually reads?

> Pipes on Windows are sections of shared memory and not artificial I/O channels.

You reading from it like a file with `ReadFile`, so dont understand why this implementation detail matters. Furthermore, `FILE*` or iostreams can read from memory buffers not just files as well. 

> The only way to make it work is to use _pipe or _popen

Why doesnt `CreatePipe` work?

Also curious why you dont pass `CREATE_NO_WINDOW` to `CreateProcess` as well? I would assume this would open a console window in GUI applications(which we dont want).

pfultz2: > have severe limitations and only can be used in console applications (require command shell support).

So from what I am reading [here](https://learn.microsoft.com/en-us/windows/console/console-handles), [`GetStdHandle`](https://learn.microsoft.com/en-us/windows/console/getstdhandle) can only be used from a console application. So to make this work we would need to launch this as a console application anyways, which is why we would need `CREATE_NO_WINDOW` to prevent a console popup happening in a GUI application.

I still dont understand why `stdin` doesn't work for this case.

apwojcik: The `CREATE_NO_WINDOWS` cannot be used when the parent process has captured the `stdin` or `stdout` of a child process. This is the case here, and the windows will not appear because both streams have been captured.

apwojcik: It turned out the mode of `stdin` was `_O_TEXT`, and since Message Pack generates binary data, there were errors in data transmission between processes. When I switched `stdin` to `_O_BINARY` mode, everything started to work. So, I removed Win32-specific code from the HIPRTC driver.
Because Windows 'end-of-line' comprises two characters (\r\n), it is essential to differentiate between text and binary when sending data.

apwojcik: I added the unit test for the 'process' class in this PR.

apwojcik: I am composing an email regarding `tiny-process-library`.

apwojcik: Additionally, I have separated process arguments from a module (path to executable) and added support for environmental variables.

src/include/migraphx/process.hpp
```diff
@@ -50,8 +52,14 @@ struct MIGRAPHX_EXPORT process
 
     process& cwd(const fs::path& p);
 
-    void exec();
-    void write(std::function<void(process::writer)> pipe_in);
+    void exec(std::string_view args = "", std::string_view envs = "");
```
pfultz2, line 55: `args` should be passed to the constructor, and it should be `std::vector<std::string>`. Also, there should be an `env` method similar to `cwd` to set the environment.

src/include/migraphx/process.hpp
```diff
@@ -50,8 +52,14 @@ struct MIGRAPHX_EXPORT process
...
+    void write(std::function<void(process::writer)> pipe_in,
+               std::string_view args = "",
+               std::string_view envs = "");
+
+#ifdef _WIN32
```
pfultz2, line 60: No windows ifdefs. This class is meant to be portable across OSs.

src/include/migraphx/process.hpp
```diff
@@ -50,8 +52,14 @@ struct MIGRAPHX_EXPORT process
...
+    process& redirect_std_out(void* rout);
```
pfultz2, line 61: I think this should be `process& read(const process::writer& pipe_out)`. For consistency, it would be good to change `write` to `process& write(std::function<void(process::writer)> pipe_in)` and then require `exec` to run the process instead of having `write` call `exec`.

src/targets/gpu/hiprtc/main.cpp
```diff
@@ -31,12 +31,21 @@
...
+#endif
+
 std::vector<char> read_stdin()
 {
     std::vector<char> result;
 
     std::array<char, 1024> buffer;
     std::size_t len = 0;
+#ifdef _WIN32
+    if(_setmode(_fileno(stdin), _O_BINARY) == -1)
```
pfultz2, line 46: Can you add a comment here to explain the reason for this?

pfultz2: Actually, maybe we should add a static method to do this so we dont need ifdefs here:

```cpp
struct process
{
    static void use_binary_for_stdin();
};
```

test/process_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+#endif
+#include "test.hpp"
+
+#ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#include <io.h>
+#include <fcntl.h>
+#undef getpid
+#define getpid _getpid
```
pfultz2, line 51: I dont think we should be including this in the test. The process class should abstract out all of that.

test/process_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    auto tid = std::this_thread::get_id();
+    auto clk = std::chrono::steady_clock::now().time_since_epoch().count();
+    std::stringstream ss;
+    ss << std::hex << "migx-test-" << pid << '-' << tid << '-' << clk << '-' << test_id << ".txt";
+    return ss.str();
+}
+
+TEST_CASE(string_stdin)
+{
+    auto out = migraphx::fs::temp_directory_path() / get_temp_file("string");
```
pfultz2, line 95: Use `tmp_dir` to create the temp file. It will handle the cleanup automatically.

johnnynunez:
??

src/targets/gpu/compile_hip.cpp
```diff
@@ -194,32 +194,32 @@ struct hiprtc_program
...
-        options.push_back("-Wno-reserved-identifier");
-        options.push_back("-Wno-unused-parameter");
-        options.push_back("-Wno-gnu-line-marker");
-        options.push_back("-Wno-old-style-cast");
+        options.emplace_back("-DMIGRAPHX_HAS_DPP=0");
+        options.emplace_back("-DMIGRAPHX_ENABLE_HIPRTC_WORKAROUNDS=1");
+        options.emplace_back("-Wno-reserved-identifier");
+        options.emplace_back("-Wno-unused-parameter");
+        options.emplace_back("-Wno-gnu-line-marker");
+        options.emplace_back("-Wno-old-style-cast");
```
pfultz2, line 211: There's a lot of changes like this that are unrelated to updating the process class. Can you put these unrelated changes in seperate PR? It will make it easier to review.

src/include/migraphx/filesystem.hpp
```diff
@@ -64,6 +64,18 @@
...
 
+// NOLINTBEGIN
+#ifdef _WIN32
+#define MIGRAPHX_EXE_POSTFIX ".exe"
+#define MIGRAPHX_LIB_PREFIX ""
+#define MIGRAPHX_LIB_POSTFIX ".dll"
+#else
+#define MIGRAPHX_EXE_POSTFIX ""
+#define MIGRAPHX_LIB_PREFIX "lib"
+#define MIGRAPHX_LIB_POSTFIX ".so"
```
pfultz2, line 75: First, this shouldn't go in this header(which is just used to include std::filesystem or the filesystem TS). Secondly, we shouldn't use macros for this. We should have functions that can append the extension to the name.

src/include/migraphx/process.hpp
```diff
@@ -49,9 +54,12 @@ struct MIGRAPHX_EXPORT process
     ~process() noexcept;
 
     process& cwd(const fs::path& p);
+    process& env(const std::vector<std::string>& envs);
+    process& launcher(const fs::path& launcher);
 
     void exec();
     void write(std::function<void(process::writer)> pipe_in);
+    void read(std::string& buffer) const;
```
pfultz2, line 62: This should take `process::writer` so we can output directly to iostream without needing to go into a string buffer.

src/include/migraphx/process.hpp
```diff
@@ -49,9 +54,12 @@ struct MIGRAPHX_EXPORT process
     ~process() noexcept;
 
     process& cwd(const fs::path& p);
+    process& env(const std::vector<std::string>& envs);
+    process& launcher(const fs::path& launcher);
```
pfultz2, line 58: What is this? If we are using a launcher like `ccache` then we will set `cmd` to `ccache` and then pass the executable as an argument.

src/targets/gpu/compile_hip.cpp
```diff
@@ -313,12 +310,16 @@ bool has_compiler_launcher()
 src_compiler assemble(src_compiler compiler)
 {
     compiler.out_ext = ".S";
-    compiler.flags   = replace_string(compiler.flags, " -c", " -S");
+    for(auto& flag : compiler.flags)
+    {
+        if(flag == "-c")
+            flag = "-S";
+    }
```
pfultz2, line 317: Use `std::replace` here.

src/compile_src.cpp
```diff
@@ -46,31 +47,24 @@ std::vector<char> src_compiler::compile(const std::vector<src_file>& srcs) const
...
 
-    if(not launcher.empty())
-    {
-        td.execute(launcher, compiler + " " + params);
-    }
-    else
-    {
-        td.execute(compiler, params);
-    }
+    td.execute(compiler, params, launcher);
```
pfultz2, line 61: This should declare the `params` with the compiler if using the launcher:

```cpp
std::vector<std::string> params;
if(not launcher.empty())
    params.push_back(launcher);
params.insert(params.end(), flags.begin(), flags.end());
const std::string& exe = launcher.empty() ? compiler : launcher;
```

Then you can do `td.execute(exe, params)`.

src/targets/gpu/compile_hip.cpp
```diff
@@ -331,21 +332,25 @@ std::vector<std::vector<char>> compile_hip_src(const std::vector<src_file>& srcs
         compiler.launcher = MIGRAPHX_HIP_COMPILER_LAUNCHER;
 #endif
 
-    if(params.find("-std=") == std::string::npos)
-        compiler.flags += " --std=c++17";
-    compiler.flags += " -fno-gpu-rdc";
+    if(std::find_if(params.begin(), params.end(), [](const auto& s) { return s.find("-std="); }) ==
```
pfultz2, line 335: Use `std::none_of`.

src/targets/gpu/hiprtc/main.cpp
```diff
@@ -58,12 +73,20 @@ int main(int argc, char const* argv[])
...
-    auto out = migraphx::gpu::compile_hip_src_with_hiprtc(
-        std::move(srcs), v.at("params").to<std::string>(), v.at("arch").to<std::string>());
-    if(not out.empty())
-        migraphx::write_buffer(output_name, out.front());
+    try
+    {
+        auto v = migraphx::from_msgpack(read_stdin());
+        std::vector<migraphx::gpu::hiprtc_src_file> srcs;
+        migraphx::from_value(v.at("srcs"), srcs);
+        // clang-format off
```
pfultz2, line 81: Why is clang-format turned off here?

test/process_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+    "deserunt mollit anim id est laborum.";
+
+std::vector<char> read_stdin()
+{
+    std::vector<char> result;
+    std::array<char, 1024> buffer{};
+    std::size_t len = 0;
+#ifdef _WIN32
+    if(_setmode(_fileno(stdin), _O_BINARY) == -1)
+        throw std::runtime_error{"failure setting IO mode to binary"};
```
pfultz2, line 69: Dont you need to reset the mode at the end of the function?

apwojcik: Yes, I do. Thanks for spotting this gap!

src/targets/gpu/hiprtc/main.cpp
apwojcik: As far as remember, I turned it off because I found the auto-formatting ugly (I think that was the reason)
```c++
        auto out =
            migraphx::gpu::compile_hip_src_with_hiprtc(std::move(srcs),
                                                       v.at("params").to_vector<std::string>(),
                                                       v.at("arch").to<std::string>());
```
I turned auto-formating on again.

PR #2475, opened 2023-11-27
Split multiple same dyn_dim

CharlieL7:
* Updates the split_single_dyn_dim pass to handle when there are multiple parameters with the same non-fixed dynamic_dimension
* This is what we see in bert_base_uncased for batch_size


src/split_single_dyn_dim.cpp
```diff
@@ -36,36 +36,79 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+        {
+            if(not dd.is_fixed())
+            {
+                num_non_fixed += 1;
+                // catch more than one non-fixed dynamic_dimension
+                if(num_non_fixed > 1)
+                {
+                    return std::nullopt;
+                }
+                ret.push_back(dynamic_dimensions_check{param.first, dd});
```
umangyadav, line 76: This can technically return true if there are more than one non-fixed dynamic dimensions in all parameters with same shape. If so docstring need to be modified.

src/split_single_dyn_dim.cpp
```diff
@@ -36,36 +36,79 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+/**
+ * Check the parameters in std::vector<dynamic_dimensions_check> object to see if any of the
+ * parameters outputs to a select_module operator.
+ */
+bool any_sm_next(const_module_ref mm, const std::vector<dynamic_dimensions_check>& ddcs)
+{
+    for(const auto& ddc : ddcs)
+    {
+        auto p_outputs  = mm->get_parameter(ddc.dyn_param_str)->outputs();
+        bool is_sm_next = std::any_of(p_outputs.cbegin(), p_outputs.cend(), [](auto ins) {
```
umangyadav, line 103: You can just do `return std::any_of`

src/split_single_dyn_dim.cpp
CharlieL7: I think you're saying that the docstring should be updated to make it more clear that the same non-fixed `dynamic_dimension` can appear within multiple parameters?

src/split_single_dyn_dim.cpp
CharlieL7: Cannot, need to continue the loop if a select_module is not found. I could move the `std::any_of` call into the `if`, but that looks pretty messy to me.

TedThemistokleous (approved):

src/split_single_dyn_dim.cpp
```diff
@@ -36,36 +36,81 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+    if(dyn_params.empty())
         return std::nullopt;
-    // Check if there is a second dynamic parameter
-    if(std::any_of(std::next(ps_it), param_shapes.end(), is_dynamic))
+    std::vector<dynamic_dimensions_check> ret{};
+    // get non-fixed dynamic_dimension from all parameters
+    for(const auto& param : dyn_params)
+    {
+        const auto& dds   = param.second.dyn_dims();
+        int num_non_fixed = 0;
```
pfultz2, line 67: Use `count_if` to get the number of non fixed dimensions.

src/split_single_dyn_dim.cpp
```diff
@@ -36,36 +36,81 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+ * In other words, each parameter can have one non-fixed dynamic_dimension `x` where `x` is the same
+ * between all of the parameters with a non-fixed dynamic_dimension.
+ * Returns the parameters and the dynamic dimension in a vector of dynamic_dimensions_check objects.
+ */
+optional<std::vector<dynamic_dimensions_check>>
+has_one_unique_dyn_dim(const std::unordered_map<std::string, shape>& param_shapes)
 {
-    // True if parameters contain exactly one dynamic shape with exactly one non-fixed
-    // dynamic_dimension.
+    if(param_shapes.empty())
```
lakhinderwalia, line 52: This special check isn't required. The return at line 60 should do just fine, without this clause.

bpickrel: or vice versa.

bpickrel (approved):
Those tests that "does nothing" are actually testing that running a pass over the test program doesn't change it, right?

On the other hand, if you're writing tests of code that truly can't be tested just to suppress a CodeCov message, isn't there another way to suppress them?

pfultz2:
@causten This shouldn't have been merged as the comments from reviewers were not addressed.

@CharlieL7 The review comments should be addressed in a followup PR.

CharlieL7:
> Those tests that "does nothing" are actually testing that running a pass over the test program doesn't change it, right?

Yes

PR #2476, opened 2023-11-28
Add Qlinearconcat op

gyulaz-htec:
[QlinearConcat](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.QlinearConcat) operator is required for the following int8 onnx zoo models:
* YOLOv3
* Mask-RCNN
* Inception v1
* DenseNet-121
* Ultra-lightweight face detection model

Fixes: https://github.com/migraphx-benchmark/AMDMIGraphX/issues/158


test/onnx/gen_onnx.py
```diff
@@ -6251,6 +6251,31 @@ def qlinearaveragepool_nt_cip_test():
...
+    zp0 = helper.make_tensor('4', TensorProto.INT8, [], [1])
+
+    t1 = helper.make_tensor_value_info('t1', TensorProto.INT8, [3])
+    s1 = helper.make_tensor('5', TensorProto.FLOAT, [], [0.25])
+    zp1 = helper.make_tensor('6', TensorProto.INT8, [], [0])
+
+    y = helper.make_tensor_value_info('out', TensorProto.INT8, [5])
+
+    node = onnx.helper.make_node(
+        'QLinearConcat',
```
lakhinderwalia, line 6270: Is this checked in, please? Thanks.

src/onnx/parse_qlinearconcat.cpp
```diff
@@ -0,0 +1,134 @@
...
+    // basic type checking for QLinearConcat Operator
+    void check_inputs(const std::vector<instruction_ref>& args) const
+    {
+        auto args_size = args.size();
+        // at least 5 input tensors, can be 5, 8, 11 ...
+        if((args_size < 5) or ((args_size - 2) % 3 != 0))
+            MIGRAPHX_THROW("QLINEARCONCAT: missing inputs");
+
+        auto t0_shape = args[2]->get_shape();
+        auto t0_type  = t0_shape.type();
```
lakhinderwalia, line 94: Please also check for valid type allowed for t0_type.

lakhinderwalia: Ideally, checking for Y related values as well..

src/onnx/parse_qlinearconcat.cpp
```diff
@@ -0,0 +1,134 @@
...
+    {
+        check_inputs(args);
+        if(not contains(info.attributes, "axis"))
+            MIGRAPHX_THROW("QLINEARCONCAT: missing axis attribute");
+
+        auto axis = parser.parse_value(info.attributes.at("axis")).template at<float>();
+        std::vector<instruction_ref> tmp;
+        for(auto idx = 2; idx < args.size(); idx += 3)
+        {
+            auto l       = args[idx];
```
lakhinderwalia, line 118: Suggestion: `l` is a very confusing name, right next to a `1`. And what does it mean, please?

src/onnx/parse_qlinearconcat.cpp
lakhinderwalia: What are the valid values allowed for `axis`, please?

test/onnx/gen_onnx.py
gyulaz-htec: I forgot to add it, thanks for noticing. I've added it.

src/onnx/parse_qlinearconcat.cpp
gyulaz-htec: changed it to a more dscriptive name.

src/onnx/parse_qlinearconcat.cpp
gyulaz-htec: Added the checks for t0_type and y related values. Axis is an int value, I've parsed it as a float mistakenly, Iv1 fixed that as well.

src/onnx/parse_qlinearconcat.cpp
```diff
@@ -0,0 +1,140 @@
...
+        if((args_size < 5) or ((args_size - 2) % 3 != 0))
+            MIGRAPHX_THROW("QLINEARCONCAT: missing inputs");
+
+        auto y_zp      = args[1];
+        auto y_zp_type = y_zp->get_shape().type();
+        if(y_zp_type != migraphx::shape::int8_type and y_zp_type != migraphx::shape::uint8_type)
+            MIGRAPHX_THROW("QLINEARCONCAT: unsupported input type");
+
+        auto t0_type = args[2]->get_shape().type();
+        if(t0_type != y_zp_type)
```
lakhinderwalia, line 99: The output type doesn't have to match the input(s) type -- They both have to be either int8 or uint8, so inputs can be int8, and output be uint8. So this comparison on line 99 should be replaced by something similar to line 95 above.. Thanks.

test/onnx/verify_onnx.cpp
```diff
@@ -1932,6 +1932,28 @@ TEST_CASE(qlinearaveragepool_nt_cip_test)
     EXPECT(migraphx::verify::verify_rms_range(result_vector, gold));
 }
```
lakhinderwalia, line 1934: A request: could you please create a test for a 3D tensor concatenation. Axis should be 1. Input types: int8. Output type uint8: 

The Input (two) tensors could be of dimension (just to illustrate) say: 3x4x2 & 3x2x2. Output should be shaped: 3x6x2.  You can keep the input values simple. All same (e.g. 10) in one input tensor. All the same (e.g. 25) in another input tensor. I am just suggesting values just to give an idea, so please write the test the way you like, based on your scale etc. Let the scale of input_1, input_2 and of the output be all different, with different zero points as well. I suggest this, because our testing of the original 'concat' operator are all along axis=0, and we should do more, ideally. If it is already tested somewhere else for the concat, please ignore my suggestion.

src/onnx/parse_qlinearconcat.cpp
```diff
@@ -0,0 +1,140 @@
...
+    {
+        auto args_size = args.size();
+        // at least 5 input tensors, can be 5, 8, 11 ...
+        if((args_size < 5) or ((args_size - 2) % 3 != 0))
+            MIGRAPHX_THROW("QLINEARCONCAT: missing inputs");
+
+        auto y_zp      = args[1];
+        auto y_zp_type = y_zp->get_shape().type();
+        if(y_zp_type != migraphx::shape::int8_type and y_zp_type != migraphx::shape::uint8_type)
+            MIGRAPHX_THROW("QLINEARCONCAT: unsupported input type");
```
lakhinderwalia, line 96: "unsupported output type", please.

src/onnx/parse_qlinearconcat.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+        auto t0_type = args[2]->get_shape().type();
+        if(t0_type != y_zp_type)
+            MIGRAPHX_THROW("QLINEARCONCAT: mismatching type");
+        for(auto idx = 2; idx < args.size(); idx += 3)
+        {
+            if((args[idx]->get_shape().type() != t0_type) or
+               (args[idx + 2]->get_shape().type() != t0_type))
+            {
+                MIGRAPHX_THROW("QLINEARCONCAT: mismatching types");
```
lakhinderwalia, line 106: "mismatching input types"

src/onnx/parse_qlinearconcat.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+    std::vector<op_desc> operators() const { return {{"QLinearConcat"}}; }
+
+    // basic type checking for QLinearConcat Operator
+    void check_inputs(const std::vector<instruction_ref>& args) const
+    {
+        auto args_size = args.size();
+        // at least 5 input tensors, can be 5, 8, 11 ...
+        if((args_size < 5) or ((args_size - 2) % 3 != 0))
+            MIGRAPHX_THROW("QLINEARCONCAT: missing inputs");
```
CharlieL7, line 91: Aren't the inputs supposed to be tuples? As in each input with type `TV` is a tuple of `(Tensor, Scale, ZeroPoint)`. The spec says `(3 - inf)` inputs, so that's what I would expect.

CharlieL7 (changes requested):

src/onnx/parse_qlinearconcat.cpp
gyulaz-htec: I've checked the models in the [description](https://github.com/ROCm/AMDMIGraphX/pull/2476#issue-2014622506) and those inputs are parsed just a simple sequence of tensors, I can't see any tuple type for shapes there.
For MaskRCNN-int8 I can see the following inputs for QLinearConcat operator:
```
shape: float_type, {1}, {0}
shape: uint8_type, {1}, {0}
shape: uint8_type, {1000, 4}, {4, 1}
shape: float_type, {1}, {0}
shape: uint8_type, {1}, {0}
shape: uint8_type, {1000, 4}, {4, 1}
shape: float_type, {1}, {0}
shape: uint8_type, {1}, {0}
shape: uint8_type, {1000, 4}, {4, 1}
shape: float_type, {1}, {0}
shape: uint8_type, {1}, {0}
shape: uint8_type, {1000, 4}, {4, 1}
shape: float_type, {1}, {0}
shape: uint8_type, {1}, {0}
shape: uint8_type, {507, 4}, {4, 1}
shape: float_type, {1}, {0}
shape: uint8_type, {1}, {0}
```
Is there any specific parsing option for tuple_type?
ONNX Runtime also parses these inputs as a sequence of tensors:
https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cpu/quantization/qlinear_concat.cc#L18-L19

gyulaz-htec: Removed the docstring and added comments about the actual input tensor layout in the input checking part.

CharlieL7 (approved):

PR #2481, opened 2023-11-29
Fix agent model and multinomial parsing

umangyadav:


pfultz2 (approved):

CharlieL7 (commented):
So this changes the buffer to random_uniform to have a shape of [batch_size, sample_size]. What difference does this make?

CharlieL7 (approved):
So this fixes the static shape version, got it.

bpickrel (approved):

PR #2482, opened 2023-11-29
Test commit for whether the NHWC mlir changes are the problem

krzysz00:
This shouldn't be merged

It bumps the commit pointer to check perf


causten:
I bumped the pointer to dad4460fafb19edab342c8b86ea0a751845bd6b0 to test fix for base performance

PR #2493, opened 2023-11-30
Update onnx proto and onnx parser to handle fp8 types

umangyadav:
updates `onnx.proto` files to be same as `onnx==1.14.1` that is used by MIGraphX. File is copied from [here](https://github.com/onnx/onnx/blob/rel-1.14.1/onnx/onnx.proto).
Add more switch cases in onnx parser to handle FP8 dtypes. 

Testing : Reading of quantized fp8 onnx model works. it runs on "ref" alright as well.

Question : Are there any backend tests that need to be enabled ?

### DIFF
--- a/tools/format.py
+++ b/tools/format.py
@@ -63,7 +63,8 @@ def clang_format(against, apply=False, path=CLANG_FORMAT_PATH):
         print(f"{git_clang_format} not installed. Skipping format.")
         return
     diff_flag = "" if apply else "--diff"
-    run(f"{git_clang_format} --binary {clang_format} {diff_flag} {base}")
+    run(f"{git_clang_format} --extensions c,cpp,hpp,h,cl,hip,in --binary {clang_format} {diff_flag} {base}"
+        )
 
 
 def get_files_changed(against, ext=('py')):


src/onnx/onnx_parser.cpp
```diff
@@ -609,6 +625,13 @@ shape::type_t get_type(int dtype)
     case 11: return shape::double_type;
     case 12: return shape::uint32_type;
     case 13: return shape::uint64_type;
+    case 18: return shape::fp8e4m3fnuz_type;
+    case 14:
+    case 15:
+    case 16:
+    case 17:
+    case 19:
+    case 20:
```
CharlieL7, line 634: Why have these if we don't support those datatypes? We also don't support sparse_tensor_type which is 8.

umangyadav: I have only added `E4M3FNUZ` type inside the MIGraphX IR so far. 

That's why only `case 18` is supported.

umangyadav: `sparse_tensor_type` is for something else. It is TypeProto something not related to data type.

attila-dusnoki-htec:
> Question : Are there any backend tests that need to be enabled ?

There are some:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/py/onnx_backend_test.py#L871-L921

umangyadav:
> > Question : Are there any backend tests that need to be enabled ?
> 
> There are some: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/py/onnx_backend_test.py#L871-L921

The other things is that  many of the tests are cast and castlike. Which would require some update in parsing logic of MIGraphX to account for new "saturate" attribute.

pfultz2:
#2504 should fix the format issue.

CharlieL7 (approved):

PR #2503, opened 2023-12-01
Enable standalone convolution for mlir by default

pfultz2:
Only use miopen for windograd.


umangyadav:
> Only use miopen for windograd. 

There are still some checks here. 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/eafd55de0b4bebc26cd090c0d10a3b5d739d6a56/src/targets/gpu/fuse_mlir.cpp#L210

umangyadav (approved):

PR #2504, opened 2023-12-01
Only format certain extensions

pfultz2:


umangyadav (approved):

tools/format.py
```diff
@@ -63,7 +63,7 @@ def clang_format(against, apply=False, path=CLANG_FORMAT_PATH):
         print(f"{git_clang_format} not installed. Skipping format.")
         return
     diff_flag = "" if apply else "--diff"
-    run(f"{git_clang_format} --binary {clang_format} {diff_flag} {base}")
+    run(f"{git_clang_format} --extensions c,cpp,hpp,h,cl,hip,in --binary {clang_format} {diff_flag} {base}")
```
umangyadav, line 66: `.hip` does it even exists ?

pfultz2: It could in the future, so we could format it.

PR #2505, opened 2023-12-02
Lazy reshape fixes

pfultz2:
There was many cases where we would do the reshape across the non-packed axis, which would produce incorrect results. This redoes the algorithm to do a fold instead of trying to do a equality comparison. 

These cases are broken for reshape copy as well. I plan to address that in another PR.


CharlieL7 (approved):
LGTM, errors out several cases of reshape_lazy that we cannot alias for.

TedThemistokleous (approved):

PR #2473, opened 2023-11-26
Add FP8 rocblas gemm support

umangyadav:
Depends on #2455 

- Checks if BETA API for FP8 rocblas gemm exists. 
- If it does and if the hardware is supported (>=gfx940) then calls into FP8 API. 
- If hardware is not supported or BETA API doesn't exist then it uses `eliminate_fp8` pass that adds converts around unsupported operator to emulate them into Float and convert back result to FP8. 
- Adds verify tests for FP8
- tested this on gfx942, it passes verification.

PR #2506, opened 2023-12-02
FP8 QuantDot operation

umangyadav:
this version of Dot/GEMM is different in the sense that, it uses FP8 dtype for the input args but computes fp32 output. 

need to disable two of the verify test for the CPU backend because they are failing. Failure is because of lossy cast from (Float->fp8->float) back inside the "ref" implementation while CPU backend optimizes out `float -> fp8 ->  float` converts to `no-op`. 

Depends on #2473


TedThemistokleous:
@umangyadav retargeting this on top of the dependent branch so once the next PR gets merged, it should retarget back to main. That'll help us see changes between the two.

src/include/migraphx/op/quant_dot.hpp
```diff
@@ -73,6 +75,10 @@ struct quant_dot
 
         auto out_lens   = a.lens();
         out_lens[dim_1] = b.lens()[dim_1];
+        if(t == shape::fp8e4m3fnuz_type)
+        {
+            return {shape::float_type, out_lens};
```
TedThemistokleous, line 80: Guessing this is to stop the noop lossy conversion?

src/eliminate_data_type.cpp
```diff
@@ -120,6 +120,22 @@ void eliminate_data_type::apply(module& m) const
...
+            while(convert_input->name() == "convert")
+            {
+                convert_input = convert_input->inputs().front();
+            }
+            if(convert_input->get_shape() == ins->get_shape())
+            {
+                m.replace_instruction(ins, convert_input);
+            }
+        }
+    }
```
pfultz2, line 138: I think this should go into a seperate pass like `eliminate_convert`.

src/simplify_reshapes.cpp
```diff
@@ -183,6 +183,11 @@ struct find_nested_convert
         auto x     = ins->inputs().front();
         auto input = x->inputs().front();
 
+        while(input->name() == "convert")
+        {
+            input = input->inputs().front();
+        }
```
pfultz2, line 189: Can you add a unit test for this?

Also, if we add a `eliminate_convert` pass, we can probably remove it from here.

src/targets/ref/target.cpp
```diff
@@ -42,17 +43,20 @@ std::string target::name() const { return "ref"; }
...
-            dead_code_elimination{},
-            auto_contiguous{},
-            dead_code_elimination{},
-            lowering{},
-            dead_code_elimination{}};
+    return {
+        normalize_ops{},
+        eliminate_pad{},
+        dead_code_elimination{},
+        eliminate_data_type{{migraphx::shape::fp8e4m3fnuz_type}, shape::float_type, {"quant_dot"}},
```
pfultz2, line 50: Why wouldn't this be set to `all`?

umangyadav: I didn't find it necessary for any other Ops

src/simplify_reshapes.cpp
umangyadav: Can i keep it for some other PR ?

umangyadav: opened #2520

src/targets/ref/target.cpp
TedThemistokleous: I think its better to be granular here instead of heavy-handed. Atleast it's explicit what's being removed incase there's issues down the line.

TedThemistokleous (approved):

src/targets/ref/lowering.cpp
```diff
@@ -307,20 +307,46 @@ struct ref_quant_gemm
...
+                });
+            });
 
+            arg_1.visit([&](auto output) {
+                args.at(1).visit([&](auto input) {
+                    std::transform(input.begin(), input.end(), output.begin(), [&](const auto x) {
+                        return static_cast<int32_t>(x);
+                    });
+                });
+            });
```
pfultz2, line 347: We shouldn't be converting the input here. [`gemm`](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/include/migraphx/gemm.hpp#L52) is already using a `double` accumulator. We could also convert the `amat` and `bmat` to double to avoid loss with the multiply as well.

src/targets/ref/lowering.cpp
```diff
@@ -307,20 +307,46 @@ struct ref_quant_gemm
...
+            migemm(result, arg_0, arg_1, int32_t{1}, int32_t{0});
```
pfultz2, line 348: This should just call our [`gemm`](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/include/migraphx/gemm.hpp#L36) function instead of `migemm`.

ahsan-ca (approved):
Looks good besides Paul comments.

src/targets/ref/lowering.cpp
umangyadav: Done.

src/targets/ref/lowering.cpp
umangyadav: Done.

pfultz2 (approved):

PR #2507, opened 2023-12-03
FP8 convolution 

umangyadav:
FP8 convolutions that uses rocMLIR. 

Uses rocMLIR from SHA : 5085343bca363109ae9ebabb7ca2b65c52bc861c

https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1336

regular `convolution` takes in both inputs as FP8 and generates FP8 output. Internally on hardware it will do accumulation in Fp32 but final result is converted back to fp8 using downcasting. 

`quant_convolution` takes in both inputs as fp8 and generates FP32 output.  This version can utilize QDQ quantization scheme to use scales to downcast FP32 output to fp8. 

rocMLIR fp8 convolution are limited to 2d forward convolutions only.  Therefore only have added tests for those. 
3D convolutions, 1D convolution, backwards convolution (transposed convolutions) are not enabled. 

`convert` fusion with `mlir-convolution` is not compiling on non-MI300 hardware therefore disabled it.

For now i've kept both versions of convolutions. In future if there is no use for regular fp8 convolution then it can be removed. 

depends on #2473


requirements.txt
```diff
@@ -29,4 +29,4 @@ pybind/pybind11@d159a563383d10c821ba7b2a71905d1207db6de4 --build
 msgpack/msgpack-c@cpp-3.3.0 -DMSGPACK_BUILD_TESTS=Off
 sqlite3@3.43.2 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 ROCmSoftwarePlatform/composable_kernel@70eefcf4f263aa5c25f3c9ff0db8f6f199ef0fb9 -DCK_BUILD_JIT_LIB=On -DCMAKE_POSITION_INDEPENDENT_CODE=On
-ROCmSoftwarePlatform/rocMLIR@9e66e8050209f03349a41b6b497f0da2b285a53b -DBUILD_FAT_LIBROCKCOMPILER=On
+ROCmSoftwarePlatform/rocMLIR@5085343bca363109ae9ebabb7ca2b65c52bc861c -DBUILD_FAT_LIBROCKCOMPILER=On
```
causten, line 32: Does this hash make sense still?  There have been an update since this work started

umangyadav: I need to update this hash.

krzysz00: The tip of our `develop` has all the fp8 fixes and your PR

umangyadav: Reverted this change. SHA from develop branch works well now.

TedThemistokleous (approved):

TedThemistokleous:
LGTM

PR #2508, opened 2023-12-03
Make reshape copy always have standard shape as output

pfultz2:


CharlieL7 (approved):
LGTM. This is how I originally thought reshape_copy would have worked. What benefit did the always merged strides version have?

TedThemistokleous (approved):

TedThemistokleous:
> LGTM. This is how I originally thought reshape_copy would have worked. What benefit did the always merged strides version have?

I kept things the same so they were consistent between the alias + copy versions. Maybe that was an error on my part.

PR #2509, opened 2023-12-04
Remove unused device kernels

umangyadav:
Gather and Pad operators have JIT kernels. 
Device kernels are no longer in use hence removing them.


TedThemistokleous (approved):

TedThemistokleous:
I get cleaning these up but do we ever want to keep this around for debug purposes?

umangyadav:
> I get cleaning these up but do we ever want to keep this around for debug purposes?
I see value in that but we can use git to retrieve older files. Also we had `gather` and `pad` JIT kernels for a while now. So unlikely it has any bugs now.

PR #2511, opened 2023-12-05
Resize op

bpickrel:
Creates a new Resize op.


bpickrel:
The test ` bin/test_ref resize_test_1` causes an error that dyn_out.computed_shape is a dynamic shape on line src/include/migraphx/op/resize.hpp:159

test/onnx/where_test.onnx
```diff
@@ -1,4 +1,4 @@
-
+	
```
bpickrel, line 1: this should not be changed.

bpickrel:
@CharlieL7 @pfultz2  I'd like to nudge you guys for a review on PR 2511.  It's got at least one trivial test error--I just re-merged it and the copyright dates have to be updated--and I may need to add some picayune tests to satisfy code coverage, but it's not too early to start looking over my implementation of Resize.  I'm just getting over being sick for a week so I hope I can remember what I did!

src/onnx/parse_resize.cpp
```diff
@@ -287,61 +336,102 @@ struct parse_resize : op_parser<parse_resize>
...
+                //         , {"coordinate_transformation_mode", coord_trans_mode}, {"sizes",
+                //         in_lens}}), args[0]);
+                // }
+                // else
+                // {
+                //     return info.add_instruction(make_op("resize", { {"nearest_mode",
+                //     nearest_mode}
+                //         , {"coordinate_transformation_mode", coord_trans_mode}, {"scales",
+                //         vec_scale}}), args[0]);
+                // }
```
CharlieL7, line 417: Note that this can only be done if a compiler pass is made to convert to the gather version from the Resize op. Additionally the current Resize op will need to handle all the needed cases.

CharlieL7 (changes requested):
Major comments:
* 0 dimension assumption is not applicable for this implementation.
* Current usage of `dyn_output` in the `op.compute()` is very convoluted and needs to be refactored.

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,331 @@
...
+                                   [](auto scale_i, size_t in_len) {
+                                       return static_cast<size_t>(scale_i * in_len);
+                                   });
+                }
+            });
+        }
+        if(dyn_out.computed_shape.dynamic())
+        {
+            // at this time, case of single, dynamic input is not allowed.
+            output_shape = {args[0].get_shape().type(), out_lens};
```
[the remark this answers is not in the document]
bpickrel: I don't think this line is even reachable, since it should throw in compute_shape()

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,331 @@
...
+        shape output_shape;
+        auto in_lens = args[0].get_shape().to_static(1).lens();
+        std::vector<size_t> out_lens(in_lens.size());
+
+        // Scales are either given, or calculated from output shape
+        std::vector<float> vec_scale(in_lens.size(), 1.0f);
+
+        if(dyn_out.computed_shape.dynamic() and args.size() > 1)
+        {
+            args[1].visit([&](auto input) {
```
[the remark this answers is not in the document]
CharlieL7: Need to look at the values inside of `scales` or `sizes` so cannot use the `dyn_output` struct.

src/onnx/parse_resize.cpp
bpickrel: Added your remark to the code comments.

src/include/migraphx/op/resize.hpp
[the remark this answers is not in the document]
bpickrel: I'm not sure any significant improvement is possible.  Either shapes or scales can be specified, and each can appear either as an attribute (constant) or as an input (runtime).  So the handling logic has to have lots of permutations.  If you can think of something that's clearer, I'd like to hear it but any change may be error-prone.

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,331 @@
...
+                using type = typename decltype(input)::value_type;
+                if constexpr(std::is_integral<type>{})
+                {
+                    // Copy the output size from args[1].
+                    std::transform(input.begin() + 1,
+                                   input.end(),
+                                   out_lens.begin() + 1,
+                                   [](auto size_i) { return size_i; });
+                    //   Assume dimension 0 is
+                    // batch size and leave input size alone.
```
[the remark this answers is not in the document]
bpickrel: ok, removed

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,331 @@
...
+            std::vector<shape::dynamic_dimension> dyn_dims(inputs.back().lens().at(0),
+                                                           shape::dynamic_dimension{0, max_val});
+            return {inputs.front().type(), dyn_dims};
+        }
+    }
+
+    argument compute(const dyn_output& dyn_out, std::vector<argument> args) const
+    {
+        shape output_shape;
+        auto in_lens = args[0].get_shape().to_static(1).lens();
```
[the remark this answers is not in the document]
bpickrel: changed

src/include/migraphx/op/resize.hpp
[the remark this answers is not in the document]
CharlieL7: * Do not use `dyn_output.computed_shape` it reruns compute_shape() with the shapes of the input arguments.
* Have two cases one with 1 input argument (constant scales or sizes) and one with 2 input arguments (variable scales or sizes).
* You can directly use the output shape calculated by compute_shape() for the 1 input argument version since this is the static shape case. 
* All dynamic shape cases that are used in current models will have >=2 inputs. In this case you have to calculate the output shape at runtime with the values in scales or sizes.

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,328 @@
...
+                const auto& inputs = output_shape.lens();
+                std::transform(inputs.begin(),
+                               inputs.end(),
+                               scales.begin(),
+                               out_lens.begin(),
+                               [](auto scale_i, size_t in_len) {
+                                   return static_cast<size_t>(scale_i * in_len);
+                               });
+            }
+        }
```
CharlieL7, line 300: Given that this is a static input tensor we should be able to calculate the scales given the output sizes or vice versa at parsing. So we could just have both scales and sizes set during parsing and not have to do this calculation at runtime?

test/onnx/gen_onnx.py
```diff
@@ -7417,22 +7438,132 @@ def resize_downsample_f_test():
 
 
 @onnx_test()
-def resize_downsample_c_test():
-    scales = np.array([1.0, 1.0, 0.6, 0.6], dtype=np.float32)
+def resize_downsample_f_dyn_test():
+    # scales is an input
+    # TODO:  How to create an Onnx file where scales is a runtime input?
```
CharlieL7, line 7443: See the variable input slice onnx tests: https://github.com/ROCm/AMDMIGraphX/blob/05642ec7807e08e27a0efec3ddb1b364cd2cddbb/test/onnx/gen_onnx.py#L8738

src/onnx/parse_resize.cpp
```diff
@@ -287,61 +336,102 @@ struct parse_resize : op_parser<parse_resize>
...
+            }
         }
         // linear mode
         else
         {
+            // out_lens and other variables can't be populated if non-static inputs.
+            if(not is_static_scale_input)
+                MIGRAPHX_THROW("PARSE_" + opd.op_name +
+                               ": linear mode not supported for non-static inputs");
```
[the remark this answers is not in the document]
bpickrel: done

src/onnx/parse_resize.cpp
```diff
@@ -287,61 +336,102 @@ struct parse_resize : op_parser<parse_resize>
         std::vector<double> vec_scale = get_scales(info.attributes);
 
         // If `scales` was not an attribute, it must be an input
-        if(vec_scale.empty())
+        // bool is_scale_input{true};
+        instruction_ref scales_sizes_arg(args[0]);
+        bool is_static_scale_input(not vec_scale.empty());
```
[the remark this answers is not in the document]
bpickrel: done

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,331 @@
...
+                    "RESIZE: One and only one of sizes or scales attributes must be given");
+            if(not sizes.empty())
+            {
+                if(not(sizes.size() == input_s.ndim()))
+                    MIGRAPHX_THROW("RESIZE: sizes attribute's size must match rank of input X");
+                std::vector<size_t> lens;
+                std::transform(sizes.begin(),
+                               sizes.end(),
+                               std::back_inserter(lens),
+                               [](auto in_len) { return static_cast<size_t>(in_len); });
```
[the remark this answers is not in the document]
bpickrel: You can't do typecasting with std::copy, can you?

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace op {
+
+/**
+ * The Resize operation mirrors the Onnx Resize operation with some differences.
+ * Currently, only Nearest mode is supported.  "Axes" and "ROI" attributes not recognized.
+ * When in put shape is a dynamic shape, the 0'th dimension is presumed to represent
+ * batch size and is not resized.
```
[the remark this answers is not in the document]
bpickrel: changed

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,331 @@
...
+                return shape{input_s.type(), lens};
+            }
+        }
+        else
+        {
+            // Dynamic output shape.
+            if(inputs.back().ndim() != 1)
+                MIGRAPHX_THROW("RESIZE: size/scale input must have rank 1");
+            if(inputs.back().dynamic() and not inputs.back().dyn_dims()[0].is_fixed())
+                MIGRAPHX_THROW("RESIZE: size/scale input must be fixed size");
```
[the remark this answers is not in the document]
bpickrel: changed

src/include/migraphx/op/resize.hpp
bpickrel: If I remember, my original reason for doing it this way was to populate the local variable vec_scale from either inputs or attributes, while keeping the code that does it together.  Whether or not it's really more clear, I don't think the difference is worth refactoring at this point.  Computationally it's insignificant since the number of dimensions is small.

Also, recall that in the current configuration this code branch will never be exercised--the parser converts a Resize instruction with static inputs to the Gather op, not Resize.

src/onnx/parse_resize.cpp
```diff
@@ -287,61 +336,109 @@ struct parse_resize : op_parser<parse_resize>
...
+                //         , {"coordinate_transformation_mode", coord_trans_mode}, {"sizes",
+                //         in_lens}}), args[0]);
+                // }
+                // else
+                // {
+                //     return info.add_instruction(make_op("resize", { {"nearest_mode",
+                //     nearest_mode}
+                //         , {"coordinate_transformation_mode", coord_trans_mode}, {"scales",
+                //         vec_scale}}), args[0]);
+                // }
```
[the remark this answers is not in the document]
bpickrel: moved

test/onnx/gen_onnx.py
bpickrel: ok, new test added

src/include/migraphx/op/resize.hpp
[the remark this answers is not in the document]
bpickrel: ok, this change is made although the difference seems insignificant.

src/include/migraphx/op/resize.hpp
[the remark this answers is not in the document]
CharlieL7: Aren't `sizes` already `size_t` type?

src/include/migraphx/op/resize.hpp
```diff
@@ -0,0 +1,317 @@
...
+                        input.begin(),
+                        input.end(),
+                        in_lens.begin(),
+                        vec_scale.begin(),
+                        [](auto sz, size_t in_len) { return static_cast<float>(sz) / in_len; });
+                }
+                else
+                {
+                    // read the scale from args[1]
+                    //
```
CharlieL7, line 274: whitespace

src/onnx/parse_resize.cpp
```diff
@@ -192,71 +193,117 @@ static std::vector<double> get_scales(const onnx_parser::attribute_map& attr)
...
+                                            const std::vector<double>& vec_scale,
+                                            instruction_ref args_0) const
+    {
+        std::string nearest_mode = get_nearest_mode(info.attributes);
+        std::vector<int> ind(out_elements);
+
+        // map out_idx to in_idx
+        auto nearest_op              = get_nearest_op(nearest_mode);
+        std::string coord_trans_mode = get_coord_trans_mode(info.attributes);
+        auto idx_op                  = get_original_idx_op(coord_trans_mode);
```
CharlieL7, line 283: Can we use the functions in the `resize` operator so we don't have repeated code?

src/onnx/parse_resize.cpp
```diff
@@ -287,61 +336,85 @@ struct parse_resize : op_parser<parse_resize>
...
+            is_constant_scale_input =
+                not parse_args(args, in_lens, opd.op_name, vec_scale, out_lens, scales_sizes_arg);
         }
 
-        if(in_lens.size() != vec_scale.size())
+        if(is_constant_scale_input)
         {
-            MIGRAPHX_THROW("PARSE_" + opd.op_name + ": ranks of input and scale are different!");
-        }
+            // TODO:  How to make a test case where the scale input is not constant?
```
CharlieL7, line 356: I think we addressed this comment already

test/onnx/parse/resize_downsample_f_dyn3_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+#include <onnx_test.hpp>
+
+TEST_CASE(resize_downsample_f_dyn3_test)
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape ss{migraphx::shape::float_type, {4}};
+    mm->add_instruction(migraphx::make_op("undefined"));
```
CharlieL7, line 32: Why do we have `undefined` instructions?

test/onnx/verify/resize_downsample_f_dyn2_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    EXPECT(migraphx::verify::verify_range_with_tolerance(result_vector,
+                                                         migraphx::verify::expected{gold}));
+
+    auto reference_result = reference_p.eval(pp).back();
+    std::vector<float> reference_vector;
+    reference_result.visit(
+        [&](auto output) { reference_vector.assign(output.begin(), output.end()); });
+    EXPECT(migraphx::verify::verify_range_with_tolerance(
+        result_vector, migraphx::verify::expected{reference_vector}));
```
CharlieL7, line 73: This looks to be checking that the result is the same after a visit? What's the purpose of this check that is different from the previous check?

test/onnx/parse/resize_downsample_f_dyn3_test.cpp
bpickrel: I think this is a pre-existing behavior I never resolved.  I tried to analyze it but never really understood what was the point  so I didn't dare try to undo it.

test/onnx/verify/resize_downsample_f_dyn3_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+    EXPECT(migraphx::verify::verify_range_with_tolerance(result_vector,
+                                                         migraphx::verify::expected{gold}));
+
+    auto reference_result = reference_p.eval(pp).back();
+    std::vector<float> reference_vector;
+    reference_result.visit(
+        [&](auto output) { reference_vector.assign(output.begin(), output.end()); });
+
+    EXPECT(migraphx::verify::verify_range_with_tolerance(
+        result_vector, migraphx::verify::expected{reference_vector}));
```
CharlieL7, line 77: same question

test/onnx/verify/resize_downsample_f_dyn_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+    EXPECT(migraphx::verify::verify_range_with_tolerance(result_vector,
+                                                         migraphx::verify::expected{gold}));
+
+    auto reference_result = reference_p.eval(pp).back();
+    std::vector<float> reference_vector;
+    reference_result.visit(
+        [&](auto output) { reference_vector.assign(output.begin(), output.end()); });
+
+    EXPECT(migraphx::verify::verify_range_with_tolerance(
+        result_vector, migraphx::verify::expected{reference_vector}));
```
CharlieL7, line 72: same question

src/onnx/parse_resize.cpp
bpickrel: done.

src/include/migraphx/op/resize.hpp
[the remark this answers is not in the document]
bpickrel: You are right, except I don't even need to copy it.  fixed

test/onnx/verify/resize_downsample_f_dyn2_test.cpp
bpickrel: The first check compares locally generated p's result with the gold values.  The second checks it against the program created by the parser.

test/onnx/verify/resize_downsample_f_dyn3_test.cpp
bpickrel: `resize_downsample_f_dyn3_test.onnx` defines `scales` as a runtime input with `return ([node], [X, scalesX], [Y])`.   In `resize_downsample_f_ref_test.onnx` , scales  is used as `return ([node], [X], [Y], [scale_tensor])` and so it is set in the parsing process.  Comparing the results of both kinds of inputs checks that the result through different code paths is the same.

test/onnx/verify/resize_downsample_f_dyn_test.cpp
bpickrel: See above.  I think this covers all possible input cases and code paths.

CharlieL7 (approved):

TedThemistokleous (approved):

Issue #2512, opened 2023-12-05
MatMulInteger op is partially supported


gyulaz-htec:
[MatMulInteger](https://github.com/onnx/onnx/blob/main/docs/Operators.md#MatMulInteger) op can accept uint8 and int8 types as inputs, but only int8 is supported.
MatMulInteger is parsed as [quant_dot](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/op/quant_dot.hpp) which has a [restriction](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/op/quant_dot.hpp#L43) about that inputs should be int8, but it should allow uint8 as well.
Also the `a_zero_point` and `b_zero_point` optional inputs are not handled in MIGraphX.

MatMulInteger is needed for the  [BERT-Squad int8](https://github.com/onnx/models/blob/main/text/machine_comprehension/bert-squad/model/bertsquad-12-int8.onnx) onnx zoo model.
At the moment it fails with the following (I've used the changes from this [PR](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2489) to pass through an unimplemented operator):
`what():  /code/AMDMIGraphX/src/include/migraphx/check_shapes.hpp:210: same_type: quant_dot: Types do not match`

Actions required:
- [x] Update MatMulInteger implementation to accept uint8 type for T1 and T2 inputs
- [x] Update MatMulInteger implementation to process `a_zero_point` and `b_zero_point` input tensors
- [x] Don't forget to enable test_matmulinteger_cpu in [onnx_backend:test.py](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/py/onnx_backend_test.py#L137)

gyulaz-htec:
Fixed in https://github.com/ROCm/AMDMIGraphX/pull/2513

PR #2513, opened 2023-12-05
Update MatMulInteger op parsing

gyulaz-htec:
Add missing uint8 input support and zero point tensor support to [MatMulInteger](https://github.com/onnx/onnx/blob/main/docs/Operators.md#MatMulInteger).
For the detailed description see: https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2512

Failing back to `dot` operator when zero point attributes are present, because those can lead out of the limits of uint8 and int8.


TedThemistokleous (changes requested):
Fix conflicts. Changes for this look good, and I think we're seeing this as part of some other models being read in with the missing uint8 support

gyulaz-htec:
> Fix conflicts. Changes for this look good, and I think we're seeing this as part of some other models being read in with the missing uint8 support

@TedThemistokleous I've fixed the conflicts also found the code path with uint8 inputs and no zero points is not covered with test, so I've added a verify test for that.

kahmed10 (approved):

TedThemistokleous (approved):

PR #2519, opened 2023-12-06
Move some cppcheck rules to cppcheck addon instead

pfultz2:
Some other rules can be converted as well but it finds many more cases, so I will enable them in seperate PRs to reduce down the diff.


src/program.cpp
```diff
@@ -507,10 +507,8 @@ std::vector<argument> generic_eval(const module* mod,
                 }));
         }
         assert(results.find(ins) != results.end());
-        if(not ins->get_shape().any_of_dynamic())
-        {
-            assert(results.at(ins).get_shape() == ins->get_shape());
-        }
+        assert(ins->get_shape().any_of_dynamic() or
```
TedThemistokleous, line 510: why drop not here?

TedThemistokleous (commented):
Minor question thats it.

src/program.cpp
pfultz2: So the `assert` will only assert `results.at(ins).get_shape() == ins->get_shape()` when `ins->get_shape().any_of_dynamic()` is false, to match the previous behavior.

TedThemistokleous: AH you're trying to save a compare. Got it.

PR #2525, opened 2023-12-06
ref_use_double and eliminate_convert pass

CharlieL7:
* `ref` is used as our target values for `verify` tests. It should have higher accuracy than our GPU tests. Accuracy is muddled for fp16 values when comparing `gpu` to `ref`
* Adds new verify_option struct
* Can use driver flag `--ref-use-double` to convert all floating point on ref target to double precision.
* Creates a new eliminate_convert pass that handles more cases of nested converts.


src/driver/verify.cpp
```diff
@@ -86,6 +88,7 @@ std::vector<argument> run_target(program p,
                                  precision quantize,
                                  const parameter_map& inputs)
 {
+    // NOTE: why do we have this here when driver/main.cpp:compiler already quantized
```
pfultz2, line 91: It doesn't quantize it for `verify`:

https://github.com/ROCm/AMDMIGraphX/blob/develop/src/driver/main.cpp#L579

So the ref is not quantized whereas the target is quantized.

src/driver/verify.cpp
```diff
@@ -74,6 +75,7 @@ verify::tolerance get_tolerances(const program& p,
 
 std::vector<argument> run_ref(program p, const parameter_map& inputs)
 {
+    fp_to_double(p);
```
pfultz2, line 78: Do we want to do this unconditionally?

CharlieL7: We could add an option to the driver, but then need another flag to be passed.

CharlieL7:
Figured out the loop_test bug. Was a logic mistake on this PR. This PR actually removes more nested converts now.
On develop loop_test we have this fused pointwise instruction:
```
loop_module:@10 = gpu::code_object[code_object=3168,symbol_name=equal_convert_convert_not_kernel,global=1,local=1024,](loop_module:@5,loop_module:loop_module:#output_0) -> bool_type, {1}, {0}, target_id=0
```
With this PR we get:
```
loop_module:@7 = gpu::code_object[code_object=3152,symbol_name=equal_convert_not_kernel,global=1,local=1024,](loop_module:@5,loop_module:loop_module:#output_0) -> bool_type, {1}, {0}, target_id=0
```

TedThemistokleous (approved):

src/include/migraphx/compile_options.hpp
```diff
@@ -40,6 +40,12 @@ struct compile_options
 
     bool fast_math       = true;
     bool exhaustive_tune = false;
+
+    /**
+     * Converts floating point values to double on the ref target.
+     * Currently only used by the driver.
+     */
+    bool fp_to_double = false;
```
pfultz2, line 48: `compile_options` are for the `compile` function. If its not being used by `compile` then it shouldn't be there.

src/driver/main.cpp
```diff
@@ -467,6 +467,10 @@ struct compiler
            {"--exhaustive-tune"},
            ap.help("Exhastively search for best tuning parameters for kernels"),
            ap.set_value(true));
+        ap(co.fp_to_double,
```
pfultz2, line 470: This should go under `verify` not `compile` as its only used with `verify` command,

src/driver/main.cpp
```diff
@@ -467,6 +467,10 @@ struct compiler
...
+           {"--fp-to-double"},
```
pfultz2, line 471: This should also be named `--ref-use-double`.

src/include/migraphx/pass.hpp
```diff
@@ -333,6 +333,21 @@ inline const ValueType& any_cast(const pass& x)
 
 #endif
 
+/// Dummy pass for default return
+struct id_pass
+{
+    std::string name() const { return "id"; }
+    void apple(const module&) const {}
```
pfultz2, line 340: It should be named `apply` not `apple`.

src/include/migraphx/pass.hpp
```diff
@@ -333,6 +333,21 @@ inline const ValueType& any_cast(const pass& x)
...
+};
+
+/// Used in the targets to enable/disable compiler passes
+inline pass enable_pass(bool enabled, pass p)
```
pfultz2, line 344: This should be defined in a pass.cpp fle. And the `id_pass` class should be moved to the .cpp file as well.

src/include/migraphx/compile_options.hpp
CharlieL7: The option is used to enable compiler passes, so it makes sense as a compile_option like disable_fast_math.

pfultz2: But we ignore the option for the GPU/CPU targets.

pfultz2:
I think it would be better to create a `verify_options` struct that can store things like `precision` and whether ref will use `double`:

```cpp
struct verify_options
{
    precision quantize = precision::fp32;
    bool ref_use_double = false;
};
```

And you can pass that in place of the `quantize` parameter.

src/driver/verify.cpp
```diff
@@ -72,9 +73,16 @@ verify::tolerance get_tolerances(const program& p,
...
+std::vector<argument> run_ref(program p,
+                              const compile_options& options,
+                              const verify_options& vo,
+                              const parameter_map& inputs)
 {
-    p.compile(migraphx::make_target("ref"));
+    p.compile(migraphx::make_target("ref"), options);
+    if(vo.ref_use_double)
+    {
+        run_passes(p, {fp_to_double{}});
```
pfultz2, line 84: Can we run this pass first before `compile`?

pfultz2 (approved):

causten:
I can merged this with the performance numbers looking the way they are

CharlieL7:
> I can merged this with the performance numbers looking the way they are

I don't see why this PR's changes would affect performance at all.

PR #2527, opened 2023-12-06
Autocast_fp8 pass

ahsan-ca:
This PR creates a pass to autocast FP8 parameters and return values to handle FP8 data type.


src/autocast_fp8.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+void autocast_fp8_pass::apply(module& m) const
+{
+    for(auto ins : iterator_for(m))
+    {
+        const auto& ins_name = ins->name();
+        if(ins_name == "@param" && ins->get_shape().type() == fp8_type)
```
umangyadav, line 38: For the future keep `fp8_type` as a set of `fp8` and check if `ins` has type as one of the FP8.

src/autocast_fp8.cpp
```diff
@@ -0,0 +1,77 @@
...
+        if(ins_name == "@return")
+        {
+            std::vector<instruction_ref> orig_inputs = ins->inputs();
+            std::vector<instruction_ref> new_inputs;
+            bool is_input_fp8 = false;
+            for(const auto& i : orig_inputs)
+            {
+                if (i->get_shape().type() == fp8_type)
+                {
+                    is_input_fp8 = true;
```
umangyadav, line 59: You can do something like this instead of keeping a flag
https://github.com/ROCm/AMDMIGraphX/blob/fe61d940c420f6e0692b90b65cb1eb3013fde312/src/eliminate_data_type.cpp#L55

src/include/migraphx/autocast_fp8.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+struct program;
+struct module;
+
+/**
+This will cast input paramater data types to fp8 and return data types to fp32.*/
+struct MIGRAPHX_EXPORT autocast_fp8_pass
+{
+    //std::vector<shape::type_t> orig_fp8_type;
+    shape::type_t fp8_type = migraphx::shape::fp8e4m3fnuz_type;
```
umangyadav, line 40: Make this a set of `fp8` types

src/include/migraphx/autocast_fp8.hpp
```diff
@@ -0,0 +1,49 @@
...
+    shape::type_t fp32_type = migraphx::shape::float_type;
```
umangyadav, line 41: Perhaps better to keep name as `target_type` in case we want to auto cast to `fp16` instead of fp32.

src/module.cpp
```diff
@@ -262,7 +262,6 @@ instruction_ref module::insert_instruction(instruction_ref ins,
                                            std::vector<instruction_ref> args)
 {
     assert(has_instruction(ins) or is_end(ins, this->end()));
-    assert(not starts_with(op.name(), "@"));
```
umangyadav, line 265: Allow parameter but not anything else. 

perhaps we should have something like `replace_return`. `replace_parameter` to make it  explicit and not allow `builitin` inside `insert_instruction`.  @pfultz2  what are your thoughts ?

umangyadav: Need to create `replace_parameter_shape()` function. 

Names of the parameter needs to remain constant.

ahsan-ca: Used remove instruction function to remove unused parameters, as Paul suggested.

src/autocast_fp8.cpp
ahsan-ca: Added check for new_inputs.size() in the if condition for this.

src/include/migraphx/autocast_fp8.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <migraphx/shape.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+struct program;
+struct module;
+
+/**
+This will cast input paramater data types to fp8 and return data types to fp32.*/
```
umangyadav, line 36: Docstring doesn't sound accurate. 

this pass will convert model with fp8 input parameter to model with fp32 input parameter and intenrally add casts to fp8 for those converted params.

src/autocast_fp8.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <migraphx/program.hpp>
+#include <migraphx/instruction.hpp>
+#include <migraphx/iterator_for.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+bool autocast_fp8_pass::is_fp8_type(const shape::type_t& s) const
+{
+    return fp8_types.find(s) != fp8_types.end();
```
umangyadav, line 35: you can use `contains()` utility method. 
e.g. 
https://github.com/ROCm/AMDMIGraphX/blob/a60bdb672f685cdfaf52a497be5aaa23eed0e0ba/src/eliminate_data_type.cpp#L42

src/autocast_fp8.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+    std::vector<instruction_ref> remove_parameters;
+    for(auto ins : iterator_for(m))
+    {
+        const auto& ins_name = ins->name();
+        if(ins_name == "@param" and is_fp8_type(ins->get_shape().type()))
+        {
+            shape::type_t fp8_type = ins->get_shape().type();
+            migraphx::shape new_shape = ins->get_shape().with_type(target_type);
+            std::string new_param_name =
```
umangyadav, line 48: is this some new name ?

ahsan-ca: Not really, I think I can change it to `param_name`.

src/autocast_fp8.cpp
```diff
@@ -0,0 +1,87 @@
...
+            {
+                if (is_fp8_type(i->get_shape().type()))
+                {
+                    new_inputs.push_back(m.insert_instruction(
+                        ins,
+                        migraphx::make_op("convert", {{"target_type", migraphx::to_value(target_type)}}),
+                        i));
+                }
+            }
+            if(new_inputs.size())
```
umangyadav, line 72: this won't work. new_inputs would only have converted instructions. 

if `return` is let's say returning mixed types 
fp32 and fp8 then new_inputs would only have fp8 one. 

also there could be multiple elements inside `new_inputs` vector.

test/autocast_fp8.cpp
```diff
@@ -0,0 +1,142 @@
...
+        auto x   = m2.add_parameter("x", {migraphx::shape::float_type, {1}});
+        auto y   = m2.add_parameter("y", {migraphx::shape::float_type, {1}});
+        auto sum = m2.add_instruction(migraphx::make_op("add"), x, y);
+        m2.add_return({sum});
+
+    }
+    EXPECT(m1 == m2);
+}
+
+int main(int argc, const char* argv[]) { test::run(argc, argv); }
```
umangyadav, line 142: add test with "return" statement that returns multiple instructions but each or some of them with different types.

src/include/migraphx/autocast_fp8.hpp
ahsan-ca: Updated string.

test/autocast_fp8.cpp
ahsan-ca: Added test for returning results of multiple instructions.

src/autocast_fp8.cpp
ahsan-ca: You are right, thanks for catching that. I have made the required changes.

src/autocast_fp8.cpp
```diff
@@ -64,8 +64,12 @@ void autocast_fp8_pass::apply(module& m) const
                         migraphx::make_op("convert", {{"target_type", migraphx::to_value(target_type)}}),
                         i));
                 }
+                else
+                {
+                    new_inputs.push_back(m.insert_instruction(ins, migraphx::make_op("identity"), i));
```
umangyadav, line 69:
```suggestion
                    new_inputs.push_back(i);
```

ahsan-ca: Done.

test/autocast_fp8.cpp
```diff
@@ -0,0 +1,172 @@
...
+    migraphx::module m1;
+    {
+        auto x   = m1.add_parameter("x", {migraphx::shape::float_type, {1}});
+        auto y   = m1.add_parameter("y", {migraphx::shape::float_type, {1}});
+        auto sum = m1.add_instruction(migraphx::make_op("add"), x, y);
+        m1.add_return({sum});
+    }
+    run_pass(m1);
+
+    migraphx::module m2;
```
umangyadav, line 161: you can remove this m2 and 
do `run_pass(m1) == m1`

test/autocast_fp8.cpp
```diff
@@ -0,0 +1,172 @@
...
+        auto y2      = m1.add_parameter("y2", {migraphx::shape::float_type, {1}});
+        auto sum1    = m1.add_instruction(migraphx::make_op("add"), x1, y1);
+        auto sum2    = m1.add_instruction(migraphx::make_op("add"), x2, y2);
+        m1.add_return({sum1, sum2});
+    }
+    run_pass(m1);
+
+    migraphx::module m2;
+    {
+        auto x2          = m2.add_parameter("x2", {migraphx::shape::float_type, {1}});
```
umangyadav, line 135: maybe change one of the shapes inside the test to be transposed. 
like `lens= {3, 4}, strides={1, 3}` 
Reason is that pass is using `with_type` somewhere. It should be fine but would be good for tests instead of having all the tests with same shapes.

umangyadav (approved):
Looks good except minor comments.

test/autocast_fp8.cpp
ahsan-ca: Done.

test/autocast_fp8.cpp
ahsan-ca: Not sure if I am missing something but I tried this and get the error ` error: invalid operands to binary expression ('void' and 'migraphx::module')
    run_pass(m1) == m1;`.

umangyadav: oh i see. you would need to copy module m1 first. 

so `module m2 = m1` and then do `run_pass(m1)` and do `m1 == m2`.

src/autocast_fp8.cpp
```diff
@@ -0,0 +1,85 @@
...
+                        ins,
+                        migraphx::make_op("convert", {{"target_type", migraphx::to_value(target_type)}}),
+                        i));
+                }
+                else
+                    new_inputs.push_back(i);
+            }
+            if(new_inputs != inputs)
+            {
+                auto new_ins = m.insert_instruction(ins, ins->get_operator(), {new_inputs});
```
pfultz2, line 72: I think `m.replace_return({new_inputs})` can be used here instead inserting and replacing.

pfultz2: And it can be done unconditionally.

src/autocast_fp8.cpp
```diff
@@ -0,0 +1,85 @@
...
+                if (contains(fp8_types, i->get_shape().type()))
+                {
+                    new_inputs.push_back(m.insert_instruction(
+                        ins,
+                        migraphx::make_op("convert", {{"target_type", migraphx::to_value(target_type)}}),
+                        i));
+                }
+                else
+                    new_inputs.push_back(i);
+            }
```
pfultz2, line 69: This can use `std::transform`:

```cpp
std::transform(inputs.begin(),
               inputs.end(),
               std::back_inserter(new_inputs),
               [&](auto i) {
                 if (contains(fp8_types, i->get_shape().type()))
                 {
                   return m.insert_instruction(
                       ins,
                       migraphx::make_op(
                           "convert",
                           {{"target_type", migraphx::to_value(target_type)}}),
                       i);
                 } else
                   return i;
               });
```

test/autocast_fp8.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <basic_ops.hpp>
+#include <migraphx/autocast_fp8.hpp>
+#include <migraphx/dead_code_elimination.hpp>
+#include <migraphx/eliminate_identity.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/instruction.hpp>
+#include <migraphx/iterator_for.hpp>
+#include <migraphx/make_op.hpp>
+#include <migraphx/pass_manager.hpp>
+#include <migraphx/permutation.hpp>
```
pfultz2, line 33: I this header needed?

test/autocast_fp8.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <migraphx/ranges.hpp>
```
pfultz2, line 34: Is this header needed as well?

test/autocast_fp8.cpp
ahsan-ca: Good catch, these are really not needed. Removed headers that are not needed.

test/autocast_fp8.cpp
```diff
@@ -0,0 +1,164 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#include <basic_ops.hpp>
+#include <migraphx/autocast_fp8.hpp>
+#include <migraphx/dead_code_elimination.hpp>
+#include <migraphx/eliminate_identity.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/instruction.hpp>
+#include <migraphx/iterator_for.hpp>
```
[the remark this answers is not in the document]
ahsan-ca: Removed headers that were not needed.

test/autocast_fp8.cpp
ahsan-ca: Removed as well.

src/autocast_fp8.cpp
ahsan-ca: Used `std::transform`. Thanks for the suggestion.

src/autocast_fp8.cpp
ahsan-ca: That makes it simpler, thanks for the suggestion. I have made the change.

src/module.cpp
```diff
@@ -476,7 +476,6 @@ instruction_ref module::insert_literal(instruction_ref ins, literal l)
 
 instruction_ref module::insert_parameter(instruction_ref ins, std::string name, shape s)
 {
-    assert(get_parameter_shape(name) == shape{});
```
ahsan-ca, line 479: I missed to include this earlier. The assert needed to be removed because we are adding a parameter with the same name as an existing one with a non-default shape. Just wanted to draw your attention to this to confirm that its ok to be removed. @pfultz2

pfultz2 (approved):

src/module.cpp
pfultz2: We shouldn't remove this assert. We should probably rename the parameter before we insert the new one.

pfultz2, line 479: We shouldn't remove this assert. We should probably rename the parameter before we insert the new one.

pfultz2 (changes requested):

src/module.cpp
ahsan-ca: I was looking and there does not seem to be existing functionality to change parameter names. Any suggestions on how to approach this?

ahsan-ca: I have added `rename_parameter` function to the module class.

src/module.cpp
```diff
@@ -554,6 +554,15 @@ instruction_ref module::get_parameter(std::string name) const
         return this->end();
 }
 
+void module::rename_parameter(instruction_ref ins, const std::string name)
+{
+    assert(ins->name() == "@param");
+    auto& op      = const_cast<operation&>(ins->get_operator());
```
pfultz2, line 560: Dont use `const_cast` here, this is UB.

src/module.cpp
```diff
@@ -554,6 +554,15 @@ instruction_ref module::get_parameter(std::string name) const
...
+    auto param_op = any_cast<builtin::param>(&op);
+    assert(param_op != nullptr);
+    param_op->parameter = name;
```
pfultz2, line 563: After you change the name then do `ins->replace(op)` to update the operator in the instruction.

src/module.cpp
```diff
@@ -554,6 +554,15 @@ instruction_ref module::get_parameter(std::string name) const
         return this->end();
 }
 
+void module::rename_parameter(instruction_ref ins, const std::string name)
+{
+    assert(ins->name() == "@param");
+    auto& op      = const_cast<operation&>(ins->get_operator());
+    auto param_op = any_cast<builtin::param>(&op);
```
pfultz2, line 561: Just do the `any_cast` in the same line: `auto op = any_cast<builtin::param>(ins->get_operator())`.

ahsan-ca: I have made the changes to the `rename_parameter` function and it looks like this now:
```
void module::rename_parameter(instruction_ref ins, const std::string name)
{
    assert(ins->name() == "@param");
    auto op = any_cast<builtin::param>(ins->get_operator());
    op.parameter = name;
    ins->replace(op);
}
```
But I get the following runtime error:
```
[   RUN    ] autocast_fp8_1
terminate called after throwing an instance of 'migraphx::version_2_9_0::exception'
  what():  /code/AMDMIGraphX/AMDMIGraphX/src/include/migraphx/builtin.hpp:78: compute_shape: builtin
Aborted (core dumped)
```
Just wanted to make sure I am not missing any changes that you suggested @pfultz2.

ahsan-ca: The function `get_operator()` does not allow modification to the operation object that it returns (defined as: `const operation& instruction::get_operator() const { return op; }`). I think this is what is causing the issue.

If I have another function that allows the modification, then the tests run fine without any issues. How do you think we should approach this in light of the changes you suggested @pfultz2?

pfultz2: We could just use assignment here:

```cpp
    auto op = any_cast<builtin::param>(ins->get_operator());
    op.parameter = name;
    *ins = instruction{op, ins->get_shape(), {}};
```

ahsan-ca: With that change, I see the following error:
```
[   RUN    ] autocast_fp8_1
terminate called after throwing an instance of 'migraphx::version_2_9_0::exception'
  what():  /code/AMDMIGraphX/AMDMIGraphX/src/pass_manager.cpp:56: validate_pass: autocast_fp8_pass
pass produces invalid program at instruction 4: add
Aborted (core dumped)
```

pfultz2: Can you push your changes?

ahsan-ca: Sure, done. The last two commits should have the changes you suggested.

pfultz2: Theres a use-after-free error here.

pfultz2: So this clears the outputs so the `replace_instruction` doesnt do anything. We need to add back the outputs:

```cpp
    auto outputs = ins->outputs();
    *ins = instruction{op, ins->get_shape(), {}};
    for(auto output:outputs)
        ins->add_output(output);
```

ahsan-ca: Ah, I see. I have pushed the commit with this change. I do not see an error with the tests now. Thanks a lot for the help, Paul!

pfultz2 (approved):

Issue #2520, opened 2023-12-06
Add `eliminate_nested_converts` pass and add unit-tests


umangyadav:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/2506#discussion_r1416435324

`eliminate_data_type` pass requires removal of nested converts and 
there is already a matcher inside `simplify_reshapes` that does exactly that.  need to move this matcher into a separate pass and use that instead. 

Add unit-tests . Unit-tests should cover cases like those described in #2517

Some initial work is already here : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/compare/quant_gemm_fp8...eliminate_nested_converts?expand=1

CharlieL7:
Working on this, fixing bug

umangyadav:
@CharlieL7  is this fixed ?

umangyadav:
Fixed with https://github.com/ROCm/AMDMIGraphX/pull/2525/files#diff-25ecf97dd6d0374c3b442647c1d8bdcb48be882a5842ccccff7e15409572081a

PR #2530, opened 2023-12-07
Print Ref compiled program in verify tests

umangyadav:
Solution to #2525  and #2520 would mean changing compilation pipeline for ref program. 
It is easier for debug purpose to print compiled ref program for comparison. 

This PR allows to print it. Previously verify test printed uncompiled program for "ref".

### DIFF
--- a/test/verify/run_verify.cpp
+++ b/test/verify/run_verify.cpp
@@ -136,16 +136,18 @@ void run_verify::validate(const migraphx::target& t,
         ti.validate(p, m);
 }
 
-std::vector<migraphx::argument> run_verify::run_ref(migraphx::program p,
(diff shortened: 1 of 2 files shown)


TedThemistokleous (approved):

umangyadav:
> LGTM, doesn't solve the issue from #2525 however because the problem is that when `ref` runs with fp16 the accuracy is inherently lower than when running with higher floating point precision. Comparing other targets to ref in the `verify` tests then becomes as wild goose chase to figure out where the error is coming from.

Yes this doesn't solve anything. This is just useful for debugging purposes.

pfultz2 (approved):

PR #2531, opened 2023-12-07
Add warnings for FP8 onnx

umangyadav:
Need to put some warnings. 

Can't find a better place other than parser to put these warnings. Ideally warnings should come from API side. 

This would work for both C/C++ and Python. 

But this wouldn't work if users were to create python or C program themselves. I assume it is not a usecase other than for testing purpose.


TedThemistokleous (approved):

TedThemistokleous (approved):

pfultz2:
You could put this warning in the compile step, that way it warns whether it was created manually, onnx or tensorflow.

PR #2534, opened 2023-12-07
prep for rocm6.0

causten:
Dry run to capture errors not seen in local tests


hip-clang.docker
```diff
@@ -54,9 +54,6 @@ ADD dev-requirements.txt /dev-requirements.txt
 ADD requirements.txt /requirements.txt
 ADD rbuild.ini /rbuild.ini
 
-# Temporarily install a new cmake until switching to ubuntu 22.04
-RUN pip3 install cmake==3.22.1
```
pfultz2, line 58: You need to update ubuntu to 22.04 if you want to remove cmake install.

pfultz2 (approved):

PR #2535, opened 2023-12-07
Add `--fp8` option to quantize models in FP8 inside `migraphx-driver`

umangyadav:
Depends on #2506 

Follows same scheme as Int8 quantization except it uses different scales compared to Int8.


shivadbhavsar:
We should also expose `quantize_fp8` to the APIs in the same way we have quantize_int8 and quantize_fp16

CharlieL7 (approved):
LGTM

src/include/migraphx/quantization.hpp
```diff
@@ -41,11 +41,19 @@ struct program;
 MIGRAPHX_EXPORT void quantize_fp16(program& prog,
                                    const std::vector<std::string>& ins_names = {"all"});
 
+MIGRAPHX_EXPORT void quantize_8bits(program& prog,
```
pfultz2, line 44: I dont think this needs to be declared in the header, its only used internally for `quantize_int8` and `quantize_fp8`.

src/quantization.cpp
```diff
@@ -57,29 +57,23 @@ void quantize_fp16(program& prog, const std::vector<std::string>& ins_names)
...
     run_passes(prog, {optimize_module{}});
 
-    std::shared_ptr<std::vector<std::pair<float, float>>> int8_quant_params =
+    std::shared_ptr<std::vector<std::pair<float, float>>> quant_8bit_params =
         std::make_shared<std::vector<std::pair<float, float>>>();
     std::shared_ptr<std::vector<float>> max_abs_vals = std::make_shared<std::vector<float>>();
 
-    auto calc_quant_params = [int8_quant_params, max_abs_vals, &t](std::size_t ins_index,
-                                                                   std::vector<argument> args) {
+    float quantized_range  = (precision == shape::type_t::int8_type) ? 127.0 : 240.0;
```
pfultz2, line 74: Can you put this in another function? Ideally we should use a visit+numeric_limits to get the quantized range, but we can leave it as is for now.

src/quantization.cpp
```diff
@@ -57,29 +57,23 @@ void quantize_fp16(program& prog, const std::vector<std::string>& ins_names)
...
+    auto calc_quant_params = [quant_8bit_params, max_abs_vals, quantized_range, &t](
```
pfultz2, line 75: This can use `[&]` capture for the lambda.

src/quantize_8bits.cpp
```diff
@@ -66,9 +64,10 @@ void quantize_int8_pass::apply(module& m) const // NOLINT
 
         auto input = ins->inputs().front();
         auto s     = input->get_shape();
-        if(contains(quantizable_types, s.type()) and s.type() != shape::int8_type)
+        if(contains(quantizable_types, s.type()) and s.type() != precision)
         {
-            auto zero_point  = m.add_literal(static_cast<int8_t>(param.second));
+            auto zero_point = m.add_literal(
+                migraphx::literal{migraphx::shape{precision}, {static_cast<int8_t>(param.second)}});
```
pfultz2, line 70: The cast is not needed anymore.

umangyadav: Removed.

src/quantization.cpp
umangyadav: Done.

src/quantization.cpp
umangyadav: NOt making change for now.

src/include/migraphx/quantization.hpp
umangyadav: Done.

src/quantization.cpp
```diff
@@ -134,23 +126,58 @@ void quantize_int8(program& prog,
...
+
+    std::vector<std::string> supported_ins_names;
+    auto* mm                                     = prog.get_main_module();
+    for(auto ins : iterator_for(*mm))
+    {
+        if(ins->name() == "convert")
+        {
+            continue;
+        }
+        else if(not starts_with(ins->name(), "@"))
```
pfultz2, line 175: The `else` is redundant here.

pfultz2 (approved):

ahsan-ca (approved):

PR #2536, opened 2023-12-08
Fix using the LD to embed files and enable by default when building shared libraries on linux

pfultz2:


kahmed10 (approved):

kahmed10:
Seems to fail on SLES.

PR #2540, opened 2023-12-08
BUMP rocMLIR SHA

umangyadav:
Points to `release/rocm-6.1` branch of rocMLIR which has fix for quantizelinear and dequantizelinear.


umangyadav:
Dup of #2539

PR #2541, opened 2023-12-08
Adding issue template

abhimeda:
Hello, this PR is introducing a new issue template for ROCm public repositories to make it easier for users to submit the details for issues and bugs. Problems pertaining to documentation or non-bug related, users have the ability to open a blank issue template to submit their report. 

Thank you,
Community Support Team (ML SW SDK)


.github/ISSUE_TEMPLATE/issue_report.yml
```diff
@@ -0,0 +1,216 @@
...
+        - ROCm 5.5.0
+  validations:
+    required: true
+- type: dropdown
+  attributes:
+    label: ROCm Component
+    description: (Optional) If this issue relates to a specific ROCm component, it can be mentioned here.
+    multiple: true
+    options:
+        - Other
```
causten, line 101: Having a list of other teams doesn't make sense,  We shouldn't be encouraging issues in one project that are actually a problem in a sibling project.  This list seems better at the ROCm/ROCm issue tracker

causten (changes requested):

.github/ISSUE_TEMPLATE/issue_report.yml
abhimeda: Thank you for pointing this out! What we can do is, set the default component to match the repo name. Would that help?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: true
```
pfultz2, line 1: What does this mean? Does this mean we can still open issues without the template?

.github/ISSUE_TEMPLATE/issue_report.yml
```diff
@@ -0,0 +1,216 @@
...
+  attributes:
+    label: ROCm Version
+    description: What version(s) of ROCm did you encounter the issue on?
+    multiple: true
+    options:
+        - ROCm 5.7.1
+        - ROCm 5.7.0
+        - ROCm 5.6.0
+        - ROCm 5.5.1
+        - ROCm 5.5.0
```
pfultz2, line 92: We dont really want to maintain a list of rocm versions here. How would a user report the issue with rocm 6.0 or 6.1 that is not on the list? It should just be `type: input` to write in the rocm version.

.github/ISSUE_TEMPLATE/issue_report.yml
pfultz2: It shouldn't just be the default, it should be the only option.

.github/ISSUE_TEMPLATE/issue_report.yml
abhimeda: I added ROCm 6.0.0

We've opted for a drop down instead of an input field because it standardizes how it is represented. If we pulled an issue from this repo into a dashboard, we can easily search for ROCm 6.0.0 vs trying to find rocm 6, or ROCm 6, ROCM 6, etc.

pfultz2: You cant do a regex  like `[0-9]+\.[0-9]+\.[0-9]+` to validate the format? I would assume the user would just write `6.0.1`(no need to repeat rocm).

abhimeda: This decision was based on how users submitted issues in the RadeonOpenCompute organization. There was too much variation and grouping related issues was difficult.

causten (approved):

.github/ISSUE_TEMPLATE/issue_report.yml
abhimeda: I removed the component list after speaking to my manager. The component will be implicitly set to AMDMIGraphX because when issues come in from this repo, we have access to the repo name.

.github/ISSUE_TEMPLATE/config.yml
abhimeda: yes for issues that don't align with the template

causten (approved):

.github/ISSUE_TEMPLATE/issue_report.yml
pfultz2: So how does a user report an issue with 6.1 or 6.2 when they are released? There is no option for those versions.

abhimeda: I would have to submit another PR

pfultz2: But how does the user report it when the PR hasn't been merged yet? Such PRs would be lower priority to merge if we are needing to get features or bug fixes in, so this is a likely scenario. Very likely users will probably pick a random version and then add a comment to which rocm version they are actually using. So this field seems completely useless.

PR #2544, opened 2023-12-12
refactoring quantization passes

umangyadav:
`quantize_8bits_pass` wasn't really using `ins_names` field. Removed it. 

using `std::set` for keeping names of the operators to be quantized instead of `std::vector` for easy lookup.


shivadbhavsar (approved):

src/include/migraphx/quantize_8bits.hpp
```diff
@@ -41,7 +41,7 @@ struct module;
  */
 struct MIGRAPHX_EXPORT capture_arguments_pass
 {
-    std::vector<std::string> ins_names = {"dot", "convolution"};
+    std::set<std::string> ins_names = {"dot", "convolution"};
```
pfultz2, line 44: Do you think `unordered_set` would be better?

umangyadav: yes for sure. I can make change

umangyadav: done

pfultz2 (approved):

PR #2545, opened 2023-12-12
Improvements to the test driver

pfultz2:
- Use glob expressions to select test cases
- Print the elapsed time for each test


test/include/test.hpp
```diff
@@ -402,6 +403,29 @@ auto within_abs(T px, U py, double ptol = 1e-6f)
         px, py, ptol);
 }
 
+template <class Iterator1, class Iterator2>
+bool glob_match(Iterator1 start, Iterator1 last, Iterator2 pattern_start, Iterator2 pattern_last)
+{
+    std::tie(start, pattern_start) =
+        std::mismatch(start, last, pattern_start, pattern_last, [](auto c, auto m) {
+            if(m == '?')
```
lakhinderwalia, line 411: Why would a match to '?' return a true here? It should behave like '*', but just for one character.

pfultz2: We need a loop for `*` which `std::mismatch` can't handle which is why we bail here.

test/include/test.hpp
```diff
@@ -606,6 +631,10 @@ struct driver
             {
             }
         }
+        auto finish = std::chrono::steady_clock::now();
+        auto elapsed_ms =
+            std::chrono::duration_cast<std::chrono::duration<double, std::milli>>(finish - start)
```
lakhinderwalia, line 636: Milliseconds may be too coarse sometimes. Should we allow (as a configurable option, if that is a possibility) a higher granularity here..? Thanks.

test/include/test.hpp
lakhinderwalia: Suggestion: please document what is this function trying to do here.

test/include/test.hpp
pfultz2: Milliseconds is a good default as it can handle fast times and slow times without getting to large for most cases. 

We can add a flag to change the unit in the future after we have an idea of which type of units we would like to use.

test/include/test.hpp
pfultz2: Added a comment.

test/include/test.hpp
```diff
@@ -402,6 +403,30 @@ auto within_abs(T px, U py, double ptol = 1e-6f)
         px, py, ptol);
 }
 
+template <class Iterator1, class Iterator2>
```
lakhinderwalia, line 406: (I have some overall concerns..)

1. Is this really globbing (which works on files) or is it regexp (which works on strings)? The api name vs what is inside makes me wonder.
2. If it is for regexp, we could use a standard libary maybe..? https://en.cppreference.com/w/cpp/regex. The code below is fairly complex, and I am wondering do we need this afresh ..? 
3. If this is a general purpose API, this shouldn't be in test.hpp
4. Handling of '*' can be fairly complex. I don't know how it is handling complex usage (because '*' would need some amount of backtracking etc..) 
5. One would need a unit test for this kind of api to test its functionality through various complexities of handling '*'. 
Thanks.

pfultz2: > Is this really globbing (which works on files) or is it regexp (which works on strings)? The api name vs what is inside makes me wonder.

It is globbing on string. It is not a regex. Its similar to the [`fnmatch`](https://www.man7.org/linux/man-pages/man3/fnmatch.3.html) function.

> If it is for regexp, we could use a standard libary maybe..? https://en.cppreference.com/w/cpp/regex. The code below is fairly complex, and I am wondering do we need this afresh ..?

Regex in the C++ standard library is really slow(slower than calling to perl or python), which is why I didnt use it. Also globbing syntax is much simpler and its backwards compatible with the exact matching we were using. A regex most likely would need to be a seperate flag. 

> If this is a general purpose API, this shouldn't be in test.hpp

The test.hpp header is meant to be a standalone header file, which is why it wasn't added to our migraphx/algorithm.hpp header. 

> Handling of * can be fairly complex. I don't know how it is handling complex usage (because * would need some amount of backtracking etc..)

It uses recursion to handle the backtracking. 

> One would need a unit test for this kind of api to test its functionality through various complexities of handling *.

We dont have any unit tests for the test.hpp header, but some unit tests would be a good idea.

lakhinderwalia: The need for regexp (glob is the incorrect name here in this code, and should be fixed anyways..) looks like a very basic one that the c++ regexp library shouldn't be discounted without some comparative test results -- I am uncomfortable with the idea of home-brewed globbing/regexp.

pfultz2: This is using a globbing expression or wildcard expression not a regular expression(which is why it is named as such). The globbing expression can be converted to a regex but it is not trivial conversion(see a C++ 17 translation [here](https://github.com/p-ranav/glob/blob/master/source/glob.cpp#L21)). Since we are only supporting a smaller subset(ie just `*` and `?`), its a fairly simple algorithm to implement glob matching directly, and I think much simpler than translating to a regex.

pfultz2: Tests added for this function.

test/test.cpp
```diff
@@ -0,0 +1,39 @@
...
+    // Repeated star
+    EXPECT(glob_match("aabaabqqbaab", "a****baab"));
+    EXPECT(glob_match("abcdd", "***d"));
+    EXPECT(glob_match("abcdd", "***d****"));
+
+    // Single wildcard
+    EXPECT(glob_match("abc", "a?c"));
+}
+
+int main(int argc, const char* argv[]) { test::run(argc, argv); }
```
lakhinderwalia, line 39:
```
#include <iostream>
#include <string>
#include <regex>

std::string do_regexp(const std::string & str, const std::string & rstr) {
  std::string rv;
  std::regex r(rstr);
  std::smatch m;
  std::cout << str << "  [" << rstr << "] ---> " << std::flush;
  auto match = std::regex_search(str, m, r);
  if(match) 
    std::cout << m.str();	// the matched part
  std::cout << '\n';

  return m.str();
}

int main()
{
  do_regexp("/foo/bar/baz/blig/fig/blig", "/foo/.*/blig");
  do_regexp("/foo/bar/baz/xlig/fig/blig", "/foo/.*/blig");
  do_regexp("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaab", "a*a*a*a*a*a*a*a*b");
  do_regexp("aabaabaab", "a.*");
  do_regexp("aabaabaab", "a.*b.*ab");
  do_regexp("aabaabaab", "a.*baab");
  do_regexp("aabaabaab", "aa.*");
  do_regexp("aabaabaab", "aaba*");
  do_regexp("aabaabqqbaab", "a*baab");
  do_regexp("aabaabqqbaab", "a*baab");
  do_regexp("abcdd", ".*d");
  do_regexp("abcdd", ".*d*");
  do_regexp("daaadabadmanda", "da*da*da*");
  do_regexp("mississippi", "m*issip*");
  do_regexp("a big cat", "zz");
  do_regexp("a big cat", "cat*");     
  return 0;
}
```

test/include/test.hpp
lakhinderwalia: PTAL at the sample code that I provided, just for comparison, without writing all the globbing stuff. We are doing string related operations here. Thanks.

test/test.cpp
lakhinderwalia: Please note, the `*` has to be written as `.*`, because `*` operates on something.. Thanks.

pfultz2: Yea there are problem with using regex directly like this:

1) Running `./bin/my_test test_foo` will run both `test_foo` and `test_foo_with_something`, when it should only run `test_foo`.
2) The test name can't always be copied directly to the command-line, since many characters need to be escaped in the regex.

This is why I am using globbing instead of a regex as it is mostly backwards compatible with what we are doing currently.

lakhinderwalia: > Yea there are problem with using regex directly like this:
> 
> 1. Running `./bin/my_test test_foo` will run both `test_foo` and `test_foo_with_something`, when it should only run `test_foo`.

Not really. You have many ways to solve it. One is to compare the original string with the matched results. Other is to search for the correct regexp:   `do_regexp("test_foo_with_something", "test_foo$")`

> 2. The test name can't always be copied directly to the command-line, since many characters need to be escaped in the regex.

The test cases that you listed are so easy to _escape_. I of course can't comment on sequences not provided in the example. Thanks.
> 
> This is why I am using globbing instead of a regex as it is mostly backwards compatible with what we are doing currently.

pfultz2: > One is to compare the original string with the matched results.

That can become ambiguous as the meaning changes if there a test with the exact name or not.

> Other is to search for the correct regexp: do_regexp("test_foo_with_something", "test_foo$")

Thats not backwards compatible. We should be able to just paste the name of the test to run it.

> The test cases that you listed are so easy to escape. I of course can't comment on sequences not provided in the example. 

Those test cases weren't added since there isn't any special characters needed to escape, but I can add some test cases like `glob_match("test.foo[gpu]", "test.foo[gpu])`, which the regex would fail. 

From an interface perspective the globbing is much simpler than a regex and it can be a drop in replacement. This is why globbing(and not a regex) is used for referencing files in the shell as well as for other things like checks for [clang tidy](https://clang.llvm.org/extra/clang-tidy/) or test names for [Catch2](https://github.com/catchorg/Catch2).

lakhinderwalia: Please add all the test cases that you think you need to address in that test file. Else this would be a discussion of hypotheticals. Thanks.

pfultz2: Sure tests are added. 

Also I checked the performance and using `std::regex_search` is 100x slower than `glob_match` for the same test cases, so its also a very significant slowdown:

```
[   RUN    ] globbing
[ COMPLETE ] globbing (1.31127ms)
[   RUN    ] regex
[ COMPLETE ] regex (137.184ms)
[==========] 2 tests ran
```

And, `glob_match` is written for simplicity and not performance, it could be optimized more.

test/test.cpp
```diff
@@ -0,0 +1,78 @@
+/*
```
lakhinderwalia, line 1: Here is an equivalent sample, with _additional more-powerful matching patterns_ than (currently implemented in this PR) in the file test.cpp. We don't have to implement regexp related stuff at all or enhance it in future. It is all there. BTW, it is fast, numbers are below.

```
#include <iostream>
#include <string>
#include <regex>

std::string do_regexp(const std::string & str, const std::string & rstr) {
  std::regex r(rstr);
  std::smatch m;
  std::cout << str << "  [" << rstr << "] ---> " << std::flush;
  auto match = std::regex_search(str, m, r);
  if(match) 
    std::cout << m.str();	// the matched part
  std::cout << '\n';

  return m.str();
}

int main(int argc, char * argv[])
{
  int ct = 1;
  if(argc > 1) ct = atoi(argv[1]);
  std::cout << "Count: " << ct << std::endl;

  if(ct < 1 || ct > 5000) return -1; // too much.. ?
  
  for(int i = 0; i < ct; i++) {
    // next 3: not matching fully, which can be checked for in many ways:
    do_regexp("ab", "a");
    do_regexp("ba", "a");
    do_regexp("bac", "a");

    // exact match..
    do_regexp("ab", "ab");
    
    do_regexp("/foo/bar/baz/blig/fig/blig", "/foo/.*/blig");
    do_regexp("/foo/bar/baz/xlig/fig/blig", "/foo/.*/blig");
    do_regexp("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaab", "a*a*a*a*a*a*a*a*b");
    do_regexp("aabaabaab", "a.*");
    do_regexp("aabaabaab", "a.*b.*ab");
    do_regexp("aabaabaab", "a.*baab");
    do_regexp("aabaabaab", "aa.*");
    do_regexp("aabaabaab", "aaba.*");
    do_regexp("aabaabqqbaab", "a.*baab");
    do_regexp("aabaabqqbaab", "a.*baab");
    do_regexp("abcdd", ".*d");
    do_regexp("abcdd", ".*d*");
    do_regexp("daaadabadmanda", "da.*da.*da.*");
    do_regexp("mississippi", "m*issip*");
    do_regexp("a big cat", "zz"); // no match

    // Some fun with regexp for non-spaces or spaces:
    // non spaces is a combo which * can't handle..
    do_regexp("a catalog for my bobcat", "\\S+cat$");
    do_regexp("a catalog for my bobcat", "\\s+cat$");
    do_regexp("a catalog for my bob cat", "\\s+cat$");

    // a notion of end of a line:
    do_regexp("a big cat", "a big$");
    do_regexp("a big", "a big$");

    // a notion of begining of a line:
    do_regexp("a big cat", "^a big");  // match
    do_regexp("a big mess", "a big$"); // no match
    

    do_regexp("aabaabqqbaab", "a.*baab");
    do_regexp("abcdd", ".*d");
    do_regexp("abcdd", ".*d.*");

    // Single wildcard
    do_regexp("abc", "a.?c");

    // Special characters
    do_regexp("test.foo[gpu]", "test.foo\\[gpu\\]");
    do_regexp("test.foo[gpu]", "test.foo\\[.*\\]");
    do_regexp("test.foo[gpu]", ".*\\[.*\\]");

    do_regexp("test.foo(gpu)", "test.foo\\(gpu\\)");
    do_regexp("test.foo(gpu)", "test.foo\\(.*\\)");
    do_regexp("test.foo(gpu)", ".*\\(.*\\)");

    do_regexp("test.foog", "test.foo\\[gpu\\]");
    do_regexp("test.foogpu", "test.foo\\[gpu\\]");
    do_regexp("test_foo", "test.foo");
  }    
  return 0;
}
```

Compiled with -O2 with gcc on rome-6 it just took 4 milli seconds for all this stuff in this file. It is pointless to compare performances here -- for deployment of test scripts -- this _4 ms_ or something alike is too little to matter, even if the other approach might be faster:
```
real    0m0.004s
user    0m0.004s
sys     0m0.000s
```
Please also see the additional built-in regexp capabilities that are highlighted above, just to give an example of the powerful library that we should not ignore, and write our own code.

test/test.cpp
lakhinderwalia: I have provided the sample equivalent code (_with more powerful rules than that are implemented in this PR_), and the execution took just 4 milli seconds. Thus performance is not an issue. Please use the built-in super powerful standard regexp library.

pfultz2: Perf was just one issue. the more important issue is usability. Sure a regex is much more powerful but introduces a lot more complexity in the interface. I dont want to have insert `^` or `$` or add in escape characters, instead of passing in `test.foo[gpu]` I need pass in `"^test\\.foo\\[gpu\\]\$"`.

This also changes the behavior of the test drivers which takes a test case name as an input. The globbing is much simpler and maintains backwards compatibility. Its also how its done in other test frameworks such as [Catch2](https://catch2-temp.readthedocs.io/en/latest/command-line.html#specifying-which-tests-to-run) and [googletest](https://google.github.io/googletest/advanced.html#running-a-subset-of-the-tests). I dont think the extra capability of regexes in this case is useful(otherwise popular test frameworks like Catch2 or googletest would use it) or worth the extra complexity(such as escaping or using `^` or `$`).

lakhinderwalia (changes requested):
As recommended, backed with ample examples here, please use std::regexp, a feature rich library, in place of your own homebrew. Thanks.

test/test.cpp
pfultz2: You are missing the case for `EXPECT(glob_match("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaab", "a*a*a*a*a*a*a*a*a*a*a*a*a*b"));`. And, a lot of the regexes are not correct translations(the first three test cases match when they shouldn't match, you are missing escaping `.`, `m*issip*` does not work the same, etc).  You can see the test cases I did for the regex version here:

https://github.com/ROCm/AMDMIGraphX/blob/test-driver-show-time-and-glob-regex/test/test.cpp#L88

I translated each globbing expression to a regex manually, and I ran the cases with rocm clang with -O3.

pfultz2:
> As recommended, backed with ample examples here, please use std::regexp, a feature rich library, in place of your own homebrew. Thanks.

Regexes changes the behavior of the test drivers which takes a test case name as an input. The globbing is much simpler and maintains backwards compatibility. Its also how its done in other test frameworks such as [Catch2](https://catch2-temp.readthedocs.io/en/latest/command-line.html#specifying-which-tests-to-run) and [googletest](https://google.github.io/googletest/advanced.html#running-a-subset-of-the-tests). I dont think the extra capability of regexes in this case is useful(otherwise popular test frameworks like Catch2 or googletest would use it) or worth the extra complexity(such as escaping or having to insert `^` or `$`).

pfultz2:
Also, if you want to use regex, you can already use `grep` to filter the tests with the regex: `$TEST_EXE -l | grep $REGEX | xargs -d'\n' $TEST_EXE`. So I dont see a need to support regex in the driver in directly. Globbing handles 99.9% of the use case without the hassle of escaping characters, and for the 0.1% case `grep` can be used.

TedThemistokleous (approved):
I think this is good enough for now, and we can look at improvements later since this is a lower priority. There is a good discussion above for and for not regexes but we can iterate on this if we need to add additional pieces. 

If anyone else on the reviewers wants to chime in too by all  means.

test/include/test.hpp
TedThemistokleous: Worth adding as a TODO?

kahmed10 (approved):

CharlieL7 (approved):
Works well on my local machine

lakhinderwalia (changes requested):
As also requested earlier, please use std::regexp, a feature rich library, in place of your own homebrew. Thanks.

pfultz2:
> As also requested earlier, please use std::regexp, a feature rich library,

You seem to be skip over the concerns I raised with using regexes. Mainly, it makes it more complicated to pass test names to the test driver.

> in place of your own homebrew. Thanks.

I am not implementing a regex, so there is no homebrewing. This is an implementation of globbing, which the standard library does not provide support for. `regex != globbing `.

lakhinderwalia:
> 
> You seem to be skip over the concerns I raised with using regexes. Mainly, it makes it more complicated to pass test names to the test driver.

No concerns were skipped. Enough examples were given, where the reviewer has spent so much time! The concerns presented just went from one subjective point to other, with no architectural justification to do the homebrew, and all concerns were answered. Replacing standard library functionality, just because one thinks it is 'complicated' is ridiculous. The regexp library replaced all what you did -- while you still insist to call your stuff as 'glob'.

pfultz2:
> No concerns were skipped. Enough examples were given, where the reviewer has spent so much time! 

The test cases clearly show how it is much more complicated to write, which you are ignoring again. The glob version of `test.foo[*]` is much easier and simpler to write than the equivalent regex version of `^test\\.foo\\[.*\\]$`.

> The concerns presented just went from one subjective point to other

 It not just my personal preference. Other team members prefer globbing instead of regex. Not only with our team, but the industry seems to prefer to use globbing over regexes as the popular test frameworks use globbing with no support for regexes. 

> with no architectural justification to do the homebrew, and all concerns were answered. Replacing standard library functionality

The justification is the standard library doesn't support globbing so thats why I implemented it. It cannot replace standard library functionality that doesn't exist in the first place.

> just because one thinks it is 'complicated' is ridiculous

Not just one but many think its more complicated. Dismissing it as ridiculous doesnt help solve the issue. Its just you are choosing to ignore the concerns instead of working on a solution.

> The regexp library replaced all what you did 

No it didnt. All the patterns had to be rewritten since there is no standard library support for globbing. 

> while you still insist to call your stuff as 'glob'.

'glob', 'globbing', or 'wildcard expression' are what they are commonly called. Globs do not include syntax for the [Kleene star](https://en.wikipedia.org/wiki/Kleene_star) which allows multiple repetitions of the preceding part of the expression; thus they are not considered a regex, which can describe the full set of [regular languages](https://en.wikipedia.org/wiki/Regular_language) over any given finite alphabet. Furthermore, none of the regex libraries or languages with regexes(like Perl) support globbing patterns either.

lakhinderwalia:
Please update the help message with how should the 'glob' operation be run.

lakhinderwalia:
Am I doing something wrong below:

root@438cd1deea1b:/code/AMDMIGraphX/build# bin/test_fuse_pointwise  si*
[   RUN    ] single
[ COMPLETE ] single (7.41811ms)
[==========] 1 tests ran
root@438cd1deea1b:/code/AMDMIGraphX/build# bin/test_fuse_pointwise  s*
[  ERROR   ] Test case 'src' not found.
[==========] 0 tests ran
[  FAILED  ] 1 tests failed
[  FAILED  ] src

pfultz2:
> Am I doing something wrong below:

Yes, you need to either put quotes around it(ie `test_fuse_pointwise 's*'`) or escape the star(ie `test_fuse_pointwise s\*`) otherwise bash will expand it to filenames if that matches. `si*` works since there are no files that match, but doing `s*` will match directories like `src` which are in the build directory. This same issue exists with gtest, catch2, and clang tidy as well.

lakhinderwalia:
> > Am I doing something wrong below:
> 
> Yes, you need to either put quotes around it(ie `test_fuse_pointwise 's*'`) or escape the star(ie `test_fuse_pointwise s\*`) otherwise bash will expand it to filenames if that matches. `si*` works since there are no files that match, but doing `s*` will match directories like `src` which are in the build directory. This same issue exists with gtest, catch2, and clang tidy as well.

Please update the help line to show the proper usage of 'glob'. Thanks.

pfultz2:
I added a 'help' text for the globbing.

test/include/test.hpp
```diff
@@ -654,20 +737,38 @@ struct driver
...
+                for(auto&& pattern : get_case_names(iname))
                 {
-                    auto f = m.find(name);
+                    auto f = m.find(pattern);
                     if(f == m.end())
                     {
-                        out() << color::fg_red << "[  ERROR   ] Test case '" << name
-                              << "' not found." << color::reset << std::endl;
-                        failed.push_back(name);
+                        std::copy_if(get_test_cases().begin(),
```
lakhinderwalia, line 749: Since you said I wasn't focused on the solution (!!)
Here is a quick replacement for all your custom 'glob' code, with what the standard library offers. About **half** the size of your code -- not including comments. And it can always be tweaked depending on your requirements, which I see are using only '*' and '?'. I did not run the test.cpp, but that should likely be fine.

```
                    if(f == m.end())
                    {
                        std::string esc_pattern;
                        const auto meta = {
                            '.', '^', '$', '*', '+', '?', '(', ')', '[', ']', '{', '}', '\\', '|'};

                        for(auto c : pattern)
                        {
                            if(c == '*' || c == '?')
                                esc_pattern.append(1, '.');
                            else if(std::find(meta.begin(), meta.end(), c) != meta.end())
                                esc_pattern.append("\\");
                            esc_pattern.append(1, c);
                        }

                        std::regex rex_pattern(esc_pattern);
                        for(const auto& test_name : get_test_cases())
                            if(std::regex_match(test_name.first, rex_pattern))
                                found_cases.push_back(test_name);
                    }
```

lakhinderwalia (changes requested):
_For other reviewers._ Please stress on the use of standard library, not custom code which you have approved. In the replacement code that I have pasted, the standard library does it in about half the size of the custom code. This updated code can be tweaked based on the exact specification, and is easily enhance-able, without writing new brittle code. All the arguments about what is 'difficult' should be laid to rest.

test/include/test.hpp
pfultz2: I dont think this handles the translation correctly. I already found one test case where it fails. It can be easily fixed, but my concern is that there is more cases that may not be handled correctly either. A compete translation that is used in libraries like p-ranav/glob [here](https://github.com/p-ranav/glob/blob/master/source/glob.cpp#L21) are much bigger, so there could defintately be some cases we are still missing. 

Furthermore, fixing those issue makes the function almost as big as the `glob_match` so its not really that much simpler.

pfultz2:
> the standard library does it in about half the size of the custom code.

This is completely false. Its 12 lines vs 18 lines of code. And its actually closer to 15 after fixing some incorrect test cases. There may be more bugs in your implementation too.

Also, `std::regex` has really bad implementations of it in the standard libraries. [It is currently faster to launch PHP to execute a regex than it is to use `std::regex`](https://cor3ntin.github.io/posts/abi/#why-do-we-want-to-break-abi). An ABI break is needed for it to improve performance to make it closer to boost.regex(which is about 20x faster on average).

All things considered its best to avoid `std::regex` until things get fixed. If we need to use regexes in the future we can use a third-party library like boost.regex or pcre2. However, I dont want to add a third-party dependency on the test drivers especially when there is a very simple implementation using `std::mismatch`. 

As a tech lead, its important for me to evaluate the dependencies we have in this project, so I dont want to have a dependency on something that has a bad implementation that either wont be fixed or if it were fixed it would cause other breakage.

lakhinderwalia:
> > the standard library does it in about half the size of the custom code.
> 
> This is completely false. Its 12 lines vs 18 lines of code. And its actually closer to 15 after fixing some incorrect test cases. There may be more bugs in your implementation too.

Total code removed: 29 lines of glob_match() + 9 in run(). 
New lines: 1 include line + 17 in run().

38 old vs. new 18. All in test.hpp.

If there are bugs in this suggested implementation, as you hint :-), like any other code in MigraphX, it can be fixed. That is what we always do -- fix bugs as they come up. The reviewers can and should review this code, which firstly, hasn't happened :-) 

> 
> Also, `std::regex` has really bad implementations of it in the standard libraries. [It is currently faster to launch PHP to execute a regex than it is to use `std::regex`](https://cor3ntin.github.io/posts/abi/#why-do-we-want-to-break-abi). An ABI break is needed for it to improve performance to make it closer to boost.regex(which is about 20x faster on average).
> 
> All things considered its best to avoid `std::regex` until things get fixed. If we need to use regexes in the future we can use a third-party library like boost.regex or pcre2. However, I dont want to add a third-party dependency on the test drivers especially when there is a very simple implementation using `std::mismatch`.
> 
_There is no third party. Same standard library from C++._  

> As a tech lead, its important for me to evaluate the dependencies we have in this project, so I don't want to have a dependency on something that has a bad implementation that either won't be fixed or if it were fixed it would cause other breakage. 

Ours is a reference code, and MiGraphX should do it the right way. Speed of testing is not a consideration here -- and if one can show that in our CI, as test cases are run, it can be worked upon.

The reviewers should be reviewing this suggested code, and we can address their concerns. We write code for the MiGraphX -- this is not a personal project.

Thanks.

pfultz2:
> The reviewers can and should review this code, which firstly, hasn't happened :-)

Yea because you dropped this code at the last minute after everyone had reviewed the code, and I made fixes as well. You really need to make the case why we should scrap what I did and use your implementation instead. I haven't see any advantages yet.

>Total code removed: 29 lines of glob_match() + 9 in run().
> New lines: 1 include line + 17 in run().
> 
> 38 old vs. new 18.

First, you are including comments which is not code. Secondly, we still need a `glob_match` function to run the unit tests. What you are writing is dishonest.

So there is only 3 lines of difference after fixing a bug. If we run into more bugs in the future this difference could be even less.

> Speed of testing is not a consideration here

Speed of testing is important for development. Its important to have a quick turn around when tweaking code and running tests. This is why I refactored the onnx tests into seperate files. 

> MiGraphX should do it the right way. 

Yes of course, but with all engineering decisions there are tradeoffs. You have not really explained a real advantage to use `std::regex` over `std::mismatch`. However, I see a lot of disadvantages with `std::regex`:

* 100x slower matching at runtime
* Slower compile-time(about a minute longer to compile all the tests with `std::regex`)
* Concerns for more bugs in the future due to incorrect translation
* ABI breakage

If we want so support more globbing features such as character classes, the `std::regex` version will become a lot more complicated, because there are different escaping rules based on the context and the character used. The `std::mismatch` version would be much simpler to implement it because we dont need to add state to a for loop. 

If we want to support things like excluding tests with `-`, `~` or `exclude:` prefix, than this will be done outside of the `glob_match` function so using `std::regex` vs `std::mismatch` is irrelevant.

I dont think these disadvantages is worth saving 3 lines of code.

lakhinderwalia:
> > The reviewers can and should review this code, which firstly, hasn't happened :-)
> 
> Yea because you dropped this code at the last minute after everyone had reviewed the code, and I made fixes as well. You really need to make the case why we should scrap what I did and use your implementation instead. I haven't see any advantages yet.

My last comment on this PR. The code should have come from the coder, the reviewer gives suggestions, and enough sample code was actually provided. No ball was dropped by this reviewer. I just saw a meandering list of concerns, and all were addressed! I finally presented the latest code when it was suggested by you that I was not focused on solution -- this should not happen in a review process. Thanks and Best wishes.

pfultz2:
> My last comment on this PR.

You last comment just claims your solution is the right solution, while also falsely claiming it uses a lot less code. You do not explain any advantages to your solution or why it is better.

> The code should have come from the coder,

I did talk about that solution early on(see [here](https://github.com/ROCm/AMDMIGraphX/pull/2545#discussion_r1432067118)), and I explained how it isn't a trivial solution by looking at other libraries that have implemented it. Then you circled back to this solution after the work was already done and updates done by other reviewers. You are running in circles here. The PR should be approved at this point.

> I just saw a meandering list of concerns, and all were addressed

And thats how you ignore my concerns that I have with your solution while also not explaining why your solution is better.

test/include/test.hpp
sean-parent: [ external commentary] The duration representation is `double` - you will get fractional milliseconds https://godbolt.org/z/8e5zY3qW7.

test/include/test.hpp
sean-parent: [external commentary] FWIW - I'm okay with this implementation if it meets your needs and prefer it to the regex suggestion because of the complexity of transforming a glob pattern to a regex pattern and making sure the escaping is happening correctly. The function needs a specification. The operation should also return the position of the last matched item (which makes the bool redundant).

pfultz2: @sean-parent Thanks for the commentary.

> The function needs a specification

Can you explain more how the specification should be defined?

> The operation should also return the position of the last matched item (which makes the bool redundant).

I actually considered returning an iterator to the first mismatch when I was implementing this originally.(I dont think returning the last matched item would work for cases where the first element doesnt match). However, this is not as simple as returning `start`. 

After the call to `std::mismatch` it is the first mismatch unless its a glob. After the star loop, `start` will point to the iterator that did match, which seem ambiguous to return. I could try to keep track of the mismatches and return it but it would make the implementation a little more complicated.

I preferred to keep it simpler and just return a bool. We aren't going to use the iterator for this context, and if we need it in the future we could look at updating it.

test/include/test.hpp
lakhinderwalia: Thanks @sean-parent . Good catch, and the URL gives a perfect illustration.

test/include/test.hpp
lakhinderwalia: > [external commentary] FWIW - I'm okay with this implementation if it meets your needs and prefer it to the regex suggestion because of the complexity of transforming a glob pattern to a regex pattern and making sure the escaping is happening correctly. The function needs a specification. The operation should also return the position of the last matched item (which makes the bool redundant).

Thanks, @sean-parent. 

The representation of the glob pattern here is _trivial_. These are test names, not even file names, as far as I have seen. Tests so far I have seen are named as `[a-z|A-Z|0-9|_]+`. There was a hint somewhere in the tests that maybe '[' and ']' might also be present in these strings, but haven't found it. For all this basic stuff , a simple 6 liner escaping of all the regex meta characters is a good start, as shown in this code:  https://github.com/ROCm/AMDMIGraphX/pull/2545#discussion_r1442390538 . And of course, plenty to room to enhance this, as required.. BTW, as a backdrop, from performance point of view, this is simply about filtering (via regex or something else..) a list of maybe up to 250 test-names, and then running those tests -- running those tests takes waaay longer than just filtering these names. So tests might run for minutes, and the regex solution might take say 1 ms longer :-) 

Some of my thoughts I would like to share with you, as a reviewer:

Had some crazy complex pattern been presented in the problem statement, or in the course of review, I would be the first to say: yes, we should escape from the escape sequences. A sample test file presented is also trivial. 

I too first thought that maybe (as you point out) there is some need to look at the matching sub-patterns etc. Hence, I gave some sample regex code to try. But so far, no such requirement for sub-patterns has emerged. This is a rather simple use-case for std::regex.

The recommendation for using std::regex was given after I saw the attempted custom 'globbing' code here is simply a confusing use of std::mismatch -- with two meta characters `*` and `?` kicking control of there -- and then onto some recursive loopy stuff to take care of basic handling of string. IMHO, this kind of confusing stuff isn't required, when simple alternative exist in the standard library. I, as a reviewer, couldn't be sure looking at this custom code, exactly what is the requirement it is trying to fulfil.

Had a complex requirement been presented, definitely a custom solution can be justified. There is no such requirement, what-so-ever. The standard library offers such a rich set of options to drive std::regex -- which include sub-patterns, back-tracking and what not, that may not be considered important today, but is a solid way to plan the future requirements. I am also uncomfortable in enhancing this little custom code in future, as it always comes up -- developers will easily break it, if not the original developer, who BTW, is very bright.

Therefore, once we outrun the usefulness of std::regex, I would be the first to suggest that we should write custom code -- now or later. I would apply that logic not just to std::regex, but also to other parts of the standard library, but need driven.

Thanks.

sean-parent: For the specification - specify the constraints on your iterators (input? forward? random access?). Specify that `[start, last)` is a valid range (`first` and `last` are the strong convention). The value type must be char (or convertible to char?, or comparable with char?). What is the complexity (in operation counts)? What is the maximum recursion depth? Globbing typically excludes path separators, so note if that is not the case here. The old [SGI STL documentation](https://www.boost.org/sgi/stl/search.html) is a great example of solid specification practices.

test/include/test.hpp
lakhinderwalia: @sean-parent, thanks to your hint at _the complexity of transforming a glob pattern to a regex pattern_ it made me do some reading. My simple transform logic presented above doesn't properly handle (though easily fixable) an expected 'failure' case very well, as regex engine will become very slow: For example: to match "aaaaaaaaaaaaaaaaaaaaa" with a pattern "a**z" should fail, but with a N**2 slowdown, and a***z with N**3 slowdown. N being the length of the input string..  

Therefore, we should add this (or similar) negative test to our test cases. Thanks.

lakhinderwalia (approved):
Approving the PR, as the discussion of glob to regex can be separated. Please do add a negative test case, as suggested. Thanks.

sean-parent:
[a final comment from the peanut gallery] I'm glad this discussion has settled. A question was asked as to when I would recommend `std::regex`.  For the cited reasons, I have not encountered a good use case (difficult to specify behavior, slow performance, and slow compile time). If I had a good regex problem, I would probably use [CTRE](https://github.com/hanickadot/compile-time-regular-expressions). If I were to use regex for this problem, I would factor out the glob-to-regex function. `std::string glob_to_regex(std::string_view glob).` There are many ways this could be implemented - I would probably use `fold_left()` with a switch statement but the most efficient solution is an interesting problem.

I will also comment that lines of code are a poor measure for any solution. Define the priorities for your project so you make consistent tradeoffs. My default priorities are correctness, efficiency, and build-times; your project priorities may differ.

lakhinderwalia:
> A question was asked as to when I would recommend `std::regex`. For the cited reasons, I have not encountered a good use case (difficult to specify behavior, slow performance, and slow compile time). 

@sean-parent, Thanks for your wisdom. Call me naive, but I somehow always expect that a standards body would clearly un-bless a std library if it has (multiple) unsatisfactory facets, so that an average user won't depend upon say a reddit thread to decide to handcraft core functions. Else, this makes the role of a reviewer very hard :-)

pfultz2:
> Call me naive, but I somehow always expect that a standards body would clearly un-bless a std library if it has (multiple) unsatisfactory facets

You should read Corentin's post ["The Day The Standard Library Died"](https://cor3ntin.github.io/posts/abi/) that I linked earlier. 

> so that an average user won't depend upon say a reddit thread to decide to handcraft core functions.

No need for a reddit thread. I am mostly up to date about whats going on with C++, and so I explained these things.

PR #2547, opened 2023-12-12
Improve performance on small reductions by doing multiple reduction per wavefront

pfultz2:
- Add `subwave` functions to get indices per the subdivided wavefront. 
- Add a `subwave<>` algorithm for reductions
- Update `dpp_reduce` to do partial reduce for smaller sizes
- Update `compute_block` to account for different wavefront sizes across different hardware
- Update `find_div_const` to skip for integrals since this can lead to incorrect results(ie `x/y` is not the same as `x*(1/y)` when using integers). This was causing some of the newer test cases to fail, which is why its included in this PR.

Locally testing this, I see a 2x perf boost on small reduction sizes.


src/targets/gpu/jit/reduce.cpp
```diff
@@ -145,18 +157,29 @@ struct simple_reduce_compiler : compiler<simple_reduce_compiler>
         auto faxis             = find_fast_axis({options.virtual_inputs.front()});
         vectorize vec{};
         auto nelements = options.virtual_inputs.back().elements();
-        auto algo      = v.get("algo", get_reduce_algo(options.virtual_inputs));
-        if(algo == "block")
+        auto algo      = v.get("algo", get_reduce_algo(ctx, options.virtual_inputs));
```
kahmed10, line 160: I think it would be better to get the wavefront size before calling into all of these functions. Then pass in the wavefront size everywhere else instead of `ctx`.

src/targets/gpu/kernels/include/migraphx/kernels/pp.hpp
```diff
@@ -0,0 +1,51 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
kahmed10, line 4: update to 2023

src/targets/gpu/jit/reduce.cpp
pfultz2: We could depend on other hardware attributes in the future, so I dont really want to expose such details in the function interface.

causten:
Tests and formatting are failing

src/simplify_algebra.cpp
```diff
@@ -1200,6 +1200,9 @@ struct find_div_const
         auto ins   = r.result;
         auto c_ins = r.instructions["c"];
 
+        if(shape::is_integral(ins->get_shape().type()))
+            return;
```
umangyadav, line 1204: i don't get why it would skip on integral types ? `find_div_const` should work for all types.

test/verify/main.cpp
```diff
@@ -81,8 +80,18 @@ int main(int argc, const char* argv[])
 #else
                 "batch_quant_dot_1<migraphx::fp8::fp8e4m3fnuz, float>",
                 "quant_dot_3args_4<migraphx::fp8::fp8e4m3fnuz, float>",
-                "quant_dot_3args_5<migraphx::fp8::fp8e4m3fnuz, float>"
+                "quant_dot_3args_5<migraphx::fp8::fp8e4m3fnuz, float>",
 #endif
+            "test_block_reduce_small<3, migraphx::shape::int8_type>",
+            "test_block_reduce_small<4, migraphx::shape::int8_type>",
+            "test_block_reduce_small<8, migraphx::shape::int8_type>",
+            "test_block_reduce_small<16, migraphx::shape::int8_type>",
```
umangyadav, line 88: These works for GCC ? I had to insert #ifdefs earlier.

src/simplify_algebra.cpp
pfultz2: Because the `1/x`(with `x > 0`) is always 0 for integral types. So its not a valid transformation.

test/verify/main.cpp
pfultz2: It doesnt matter if its gcc or clang. We convert it to a float on the cpu backend which produces different results than just using int8.

test/verify/test_block_reduce_small.cpp
```diff
@@ -0,0 +1,68 @@
...
+};
+
+template <int N>
+struct test_block_reduce_small_for_size : test_block_reduce_small<N, migraphx::shape::half_type>,
+                                          test_block_reduce_small<N, migraphx::shape::float_type>,
+                                          test_block_reduce_small<N, migraphx::shape::int8_type>
+{
+};
+
+template struct test_block_reduce_small_for_size<2>;
```
CharlieL7, line 58: Shouldn't we confirm that these tests are using the reduce kernel we expect it use? This verify test checks that the results are correct, but we don't check if the better `algo` is used for a given case.

src/targets/gpu/kernels/include/migraphx/kernels/pp.hpp
```diff
@@ -0,0 +1,51 @@
...
+#define MIGRAPHX_PP_REPEAT4(m, ...) MIGRAPHX_PP_REPEAT3(m, __VA_ARGS__) m(4, __VA_ARGS__)
+#define MIGRAPHX_PP_REPEAT5(m, ...) MIGRAPHX_PP_REPEAT4(m, __VA_ARGS__) m(5, __VA_ARGS__)
+#define MIGRAPHX_PP_REPEAT6(m, ...) MIGRAPHX_PP_REPEAT5(m, __VA_ARGS__) m(6, __VA_ARGS__)
+#define MIGRAPHX_PP_REPEAT7(m, ...) MIGRAPHX_PP_REPEAT6(m, __VA_ARGS__) m(7, __VA_ARGS__)
+#define MIGRAPHX_PP_REPEAT8(m, ...) MIGRAPHX_PP_REPEAT7(m, __VA_ARGS__) m(8, __VA_ARGS__)
+#define MIGRAPHX_PP_REPEAT9(m, ...) MIGRAPHX_PP_REPEAT8(m, __VA_ARGS__) m(9, __VA_ARGS__)
+#define MIGRAPHX_PP_REPEAT10(m, ...) MIGRAPHX_PP_REPEAT9(m, __VA_ARGS__) m(10, __VA_ARGS__)
+
+#define MIGRAPHX_PP_REPEAT(n, m, ...) \
+    MIGRAPHX_PP_PRIMITIVE_CAT(MIGRAPHX_PP_REPEAT, n)(m, __VA_ARGS__)
```
CharlieL7, line 46: Checking my understanding: This code looks to repeat the given argument a set number of times using a macro. I'm guessing that doing it this way makes it a compile-time static number of function calls? It's not fully clear to me what we're doing.

test/verify/test_block_reduce_small.cpp
pfultz2: We dont have a way to check that.

src/targets/gpu/kernels/include/migraphx/kernels/pp.hpp
pfultz2: This will invoke the macro `m` for `n` times passing in the count and the additional arguments. So `MIGRAPHX_PP_REPEAT(2, m, x)` will do `m(0, x) m(1, x)`. 

> I'm guessing that doing it this way makes it a compile-time static number of function calls?

These are preprocessor macros. There is no function calls or callstack.

>  It's not fully clear to me what we're doing.

This is used to generate the assembly for the different `SubWaveSize`. Generating a single assembly block for each step would allow the compiler to insert instructions inbetween the assembly blocks which we dont want. So instead I generate all the steps for each `SubWaveSize`(which is just string literals), by using the `MIGRAPHX_PP_REPEAT` to repeat the steps.

src/targets/gpu/kernels/include/migraphx/kernels/pp.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace migraphx {
+
+} // namespace migraphx
```
umangyadav, line 50: Can this be removed ?

pfultz2: Yes

causten:
@pfultz2  format issue

PR #2549, opened 2023-12-13
Allow reshape ops between dq and quant_op

shivadbhavsar:
Add "reshape" to the list of ops allowed between a `dequantize` and `quant_x` op. Also changed the naming of the associated matcher since this is the only place it is used (open to suggestions on the naming).

This is required for running quantized gpt2 via torch_migraphx

### DIFF
--- a/src/include/migraphx/matcher.hpp
+++ b/src/include/migraphx/matcher.hpp
@@ -858,9 +858,9 @@ auto skip_broadcasts_converts(Ms... ms)
 }
 
 template <class... Ms>
-auto skip_broadcasts_transposes_contiguous(Ms... ms)
+auto skip_post_dq_ops(Ms... ms)
 {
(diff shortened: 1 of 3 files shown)


TedThemistokleous (approved):

umangyadav (approved):

src/include/migraphx/matcher.hpp
```diff
@@ -858,9 +858,9 @@ auto skip_broadcasts_converts(Ms... ms)
 }
 
 template <class... Ms>
-auto skip_broadcasts_transposes_contiguous(Ms... ms)
+auto skip_post_dq_ops(Ms... ms)
```
pfultz2, line 861: This should probably just be moved to simplify_qdq since it will only be used there.

shivadbhavsar: Done

pfultz2 (approved):

PR #2553, opened 2023-12-14
Fixed iterator_out_of_range error in test_op_shape

aljanjic-amd:
Fixed error iterator_out_of_range in test_op_shape using `std::next(data_dims.begin(), k)` function in if condition where the test was falling instead of operator+ `data_dims.begin() + k`. Similar changes are made on two more operator+ places in the same if condition, `std::next(ind_dims.begin(), q - 1)` instead of `ind_dims.begin() + q - 1` and `std::next(upd_dims.begin(), q - 1)` instead of `upd_dims.begin() + q - 1`.

### DIFF
--- a/src/include/migraphx/op/scatternd_op.hpp
+++ b/src/include/migraphx/op/scatternd_op.hpp
@@ -91,11 +91,10 @@ struct scatternd_op : op_name<Derived>
         auto data_dims = data_shape.to_dynamic().dyn_dims();
 
         // Check that corresponding portions of tensor shapes match.
-        if(not(std::equal(
-                   ind_dims.begin(), std::next(ind_dims.begin(), q - 1), upd_dims.begin()) and
-               std::equal(std::next(data_dims.begin(), k),
-                          data_dims.end(),
-                          std::next(upd_dims.begin(), q - 1))))
+        // Brackets around q - 1 are placed for safeguarding against the breaking iterator out of
+        // vector range.
+        if(not(std::equal(ind_dims.begin(), ind_dims.begin() + (q - 1), upd_dims.begin()) and
+               std::equal(data_dims.begin() + k, data_dims.end(), upd_dims.begin() + (q - 1))))
             MIGRAPHX_THROW("ScatterND: incorrect update shape. Update dimensions must match "
                            "indices and data.");
 
--- a/src/include/migraphx/op/scatternd_op.hpp
+++ b/src/include/migraphx/op/scatternd_op.hpp
@@ -91,8 +91,8 @@ struct scatternd_op : op_name<Derived>
         auto data_dims = data_shape.to_dynamic().dyn_dims();
 
         // Check that corresponding portions of tensor shapes match.
-        if(not(std::equal(ind_dims.begin(), ind_dims.begin() + q - 1, upd_dims.begin()) and
-               std::equal(data_dims.begin() + k, data_dims.end(), upd_dims.begin() + q - 1)))
+        if(not(std::equal(ind_dims.begin(), std::next(ind_dims.begin(), q - 1), upd_dims.begin()) and
+               std::equal(std::next(data_dims.begin(), k), data_dims.end(), std::next(upd_dims.begin(), q - 1))))
             MIGRAPHX_THROW("ScatterND: incorrect update shape. Update dimensions must match "
                            "indices and data.");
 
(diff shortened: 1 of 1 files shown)


pfultz2:
If its out of bounds then `std::next(data_dims.begin(), k)` will still access out of bounds as well. Dont we need a bounds check?

bpickrel:
What is the failure case here?  The value of q and the check on k a few lines up should have ensured that these iterators are always in range.

pfultz2:
> What is the failure case here? The value of q and the check on k a few lines up should have ensured that these iterators are always in range.

Good point.

aljanjic-amd:
Without any changes, all Test cases contained in _test_op_shape_test_ are passing on Linux, while 5 of them are failing on the same if condition and with the same error on Windows. The error occurred in these Test cases when trying to iterate through vectors past their end. In the mentioned if condition, it should be iterated through vector _data_dims_ for _k_ times, starting from _data_dims.begin()_ element of the vector (`data_dims.begin() + k`), even after many if checkings before this where are variables like _k_ computed. For example, in failing Test case test_scatternd0, size of _data_dims_ vector is 1, and _k_ is also equal to 1, so the code breaks when the operator+ is applied to the segment above, because it couldn't be approached to the _k_-th element after _data_dims.begin()_ element, because vector _data_dims_ has only one element. Similar errors occur in other failing Test cases. After modification with use od `std::next(data_dims.begin(), k)` function which returns the _k_-th element of vector _data_dims_ if it's possible to iterate _k_ times or the last element of vector if it's not possible, all of this Test cases are passing both on Windows and Linux. @pfultz2 @bpickrel

pfultz2:
> For example, in failing Test case test_scatternd0, size of data_dims vector is 1, and k is also equal to 1, so the code breaks when the operator+ is applied to the segment above, because it couldn't be approached to the k-th element after data_dims.begin() element, because vector data_dims has only one element.

When doing `data_dims.begin() + 1` when `data_dims` is size 1 should result in `data_dims.begin() + 1 == data_dims.end()`. And calling `std::equal` with an empty range(ie both iterators are `data_dims.end()`) is completely valid. So, I dont understand what the issue is here.

Furthermore the iterators to `data_dims` are random access so `std::next` will just call the `+` or `+=`, so the fix wont do anything different here.

bpickrel:
Are the tests that led to this error ones that are supposed to throw a shape error?  Some of those contain inputs with deliberate indexing errors.  `std::equal()` should be able to detect iterator ranges that don't match before trying to dereference `end()`, but could you find out if it's doing it wrong?  I hope this would be easy to check with some debug code.

aljanjic-amd:
> When doing `data_dims.begin() + 1` when `data_dims` is size 1 should result in `data_dims.begin() + 1 == data_dims.end()`. And calling `std::equal` with an empty range(ie both iterators are `data_dims.end()`) is completely valid. So, I dont understand what the issue is here.
> 
> Furthermore the iterators to `data_dims` are random access so `std::next` will just call the `+` or `+=`, so the fix wont do anything different here.

The problem is not with `std::equal()` function, because that function works well in the first part of the mentioned `if` condition, the problem occurred in the second part of the `if` condition, at segment `data_dims.begin() + k`. The functions `data_dims.begin()` and `data_dims.end()` returns the first and the last element of the vector `data_dims` and that's the problem with the vector size 1, where `data_dims.begin() == data_dims.end()`. For this reason, I changed the code with the `std::next()` function, because the behavior is the same as the operator+ has, but the errors don't occur anymore. Why is operator+ working well on Linux and not on Windows, I'm not sure. @pfultz2 

> Are the tests that led to this error ones that are supposed to throw a shape error? Some of those contain inputs with deliberate indexing errors. std::equal() should be able to detect iterator ranges that don't match before trying to dereference end(), but could you find out if it's doing it wrong? I hope this would be easy to check with some debug code.

The same error occurred in 5 Test cases in the `expect_shape()` function and none of these errors occurred in the Test cases where is function `throws_shape()` tested, which should manage incorrect shapes and handle the exceptions. After my modification of the code, all Test cases pass successfully. @bpickrel 
@apwojcik

bpickrel:
1.  Could you please also try the following @aljanjic-amd ? In the original code, change `q-1` to `(q-1)` each place it appears.  I am wondering if left-to-right evaluation leads to a temporary out of bounds reference.
2.   What compiler does the Windows build use?  If I had that, I could try it myself instead of asking you to do it.

aljanjic-amd:
In the last commit, after @bpickrel's comment and yesterday's meeting, in the original code I added brackets in order to "safeguard" hitting error iterator out of range, so in the original code in the `std::equal` function, `ind_dims.begin() + q - 1` was replaced with `ind_dims.begin() + (q - 1)` and the same modification was applied to the `upd_dims.begin() + q - 1` which was also replaced with `upd_dims.begin() + (q - 1)`. After this changes, all tests are passing. @pfultz2

pfultz2 (approved):

bpickrel (approved):
: )

PR #2554, opened 2023-12-14
Change ccache location in CI servers

causten:
Move location of ccache to the `workspace` area on the servers and avoid the /var/jenkins directory tree.  At the time of install /var/jenkins is owned by the root user.  There were scripts that changed the ownership.  Since the user 'jenkins' is running in the container and all job work is done under /home/jenkins/workspace/... the ccache dir would be more logically located there


pfultz2 (approved):

TedThemistokleous (approved):

Jenkinsfile
```diff
@@ -12,7 +11,7 @@ def rocmtestnode(Map conf) {
     def docker_args = conf.get("docker_args", "")
     def docker_build_args = conf.get("docker_build_args", "")
     def pre = conf.get("pre", {})
-    def ccache = "/var/jenkins/.cache/ccache"
+    def ccache = "/home/jenkins/workspace/.cache/ccache"
```
pfultz2, line 14: It seems it would be better to use `/home/jenkins/ccache`. I assume the workspace directory is managed by jenkins so I dont know if there is a conflict.

PR #2561, opened 2023-12-19
Split onnx verify tests

pfultz2:


test/onnx/verify_onnx.cpp
```diff
@@ -623,9 +623,8 @@ template <typename T = float>
 std::vector<T> norm_test(const std::vector<size_t>& x_dims,
                          std::vector<T>& scale,
                          std::vector<T>& bias,
-                         const std::string& onnx_file)
+                         migraphx::program p)
```
lakhinderwalia, line 626: Shouldn't these go into their own file? Thanks.

pfultz2: They are in their own file here: https://github.com/ROCm/AMDMIGraphX/pull/2561/files#diff-7f2673ea4663cdcf9824170130c611738ecb7c982897704dd28f88c0df4e62d0R33

This was just changed so the split tests can run correctly. This file will be removed after this is merged. Its not before to avoid difficult merge conflicts.

lakhinderwalia: OK. Very confusing, because I do see some changes in this file, a small number. I don't see it is being deleted -- perhaps you will do it in a different PR. I am going to approve it.

lakhinderwalia (approved):
Approving this PR. I was expecting verify_onnx.cpp to be deleted here, or maybe soon after this PR. Thanks.

Issue #2570, opened 2023-12-21
too many warnings during a compile


causten:
There are over 10k warnings about `warning: unsafe pointer arithmetic [-Wunsafe-buffer-usage]` during the compilation of MIGraphX.  If this is a bug let's fix it, if the code is working as designed and the warning is not needed let's suppress it for the specific lines with a pragma.  

line 345 and 351 of src/include/migraphx/shape.hpp makes up the bulk of the problem.  

To recreate the problem simply compile the code

pfultz2:
> line 345 and 351 of src/include/migraphx/shape.hpp makes up the bulk of the problem.

Where are the rest of the warning coming from? Can you show the output from `make |& grep warning: | sort -u`?

causten:
other heavy offenders... 

src/include/migraphx/literal.hpp:103:63
test/include/test.hpp:531:39
src/include/migraphx/op/loop.hpp:112:56
src/include/migraphx/run_loop.hpp:106:53

pfultz2:
We should disable the warning globally(in cmake [here](https://github.com/ROCm/AMDMIGraphX/blob/develop/cmake/EnableCompilerWarnings.cmake#L103)). Its not useful, its warning when accessing array in bounds using constants.

pfultz2:
Also, this is an experimiental diagnostic clang, so the recommened approach is to disable it in the build: https://www.devgem.io/posts/how-to-address-the-wunsafe-buffer-usage-warning-in-clang

PR #2578, opened 2023-12-28
add "-Wno-unsafe-buffer-usage"

NBolade-224:
Add flag to supress warnings from #2570


cmake/EnableCompilerWarnings.cmake
```diff
@@ -101,6 +101,7 @@ else()
                 -Wno-unused-command-line-argument
                 -Wno-weak-vtables
                 -Wno-c99-extensions
+                -Wunsafe-buffer-usage
```
pfultz2, line 104: This should be `-Wno-unsafe-buffer-usage`.

kahmed10 (approved):

PR #2579, opened 2023-12-29
To rocm6.0 onnxrt whl

TedThemistokleous:
Update the Jenkinsfile to use a prebuild wheel file instead of always performing a full onnxruntime during our CI.

Should speed up CI by  ~30 minutes per run as we're not building Onnxruntime anymore. 

Also moved CI dockerfile to 6.0 here to match with the prebuild onnxruntime wheel


causten (approved):

TedThemistokleous:
this still needs a way to add the build folder so we can pull that blob in to run our CI.

causten:
> this still needs a way to add the build folder so we can pull that blob in to run our CI.

I saw that.  I approved because I was thinking the Jenkins file is part of AMDs CI infrastructure.  Ideally the test binaries would be published in a location that is publicly available

causten:
@TedThemistokleous  should this be updated to support the current 6.1.3 CI environment? It has published whl and zip now.  The move to rocm 6.2 wouldn't likely think of this if it is not there for 6.1 already

TedThemistokleous:
@Causten looks like we still don't have the .zip files on repo.radeon even for rocm 6.2. Ill need to figure out why. whls are up though

PR #2587, opened 2024-01-03
Run simplify_qdq before quantize_fp16

shivadbhavsar:
Issue that need to be fixed:
- Pre-quantized models (models that already have qdq instructions) cannot be quantized to fp16 (for non dot/conv ops) with the current workflow
- When `quantize_fp16` is called, the qdq ops for weights are evaluated and converted to literals before reaching the `simplify_qdq` pass in the compile pipeline

Alternate Fixes:
1) Skip `quantizelinear` and `dequantizelinear` ops when running const_prop. Issue here is that weights would now get converted to fp16 first and then be further quantized to int8 instead of being directly quantized down to int8 from full precision. Also, something about the quantize_fp16 pass seems to still be removing qdq instructions associated with weights.
2) Add quantize_fp16 in compile pipeline (somewhere after simplify_qdq). Requires adding compile option to enable this pass, and needs to deal with different targets.

Proposed Fix:
- Simply run `simplify_qdq` before running the current `quantize_fp16` passes. This ensures any instructions that can be reduced to int8/fp8 ops are handled first and any weights are quantized directly from full precision.


TedThemistokleous (approved):

pfultz2 (changes requested):
Removing the qdq should be decided by the backend target based on what quantized operators it supports. We can't run this pass in quantize_fp16 because backends can't undo the quantization without losing precision.

shivadbhavsar:
> Removing the qdq should be decided by the backend target based on what quantized operators it supports. We can't run this pass in quantize_fp16 because backends can't undo the quantization without losing precision.

Ok fair enough. So how should we deal with this scenario?
I attempted to manipulate the quantize_fp16 pass so that it would preserve the qdq ops, but doing this is problematic in many ways and also introduces a lot of edge cases. I would say that if we are doing both int8/fp8 quantization + fp16 quantization, it should always do int8/fp8 first (ie. resolve qdqs) and then convert everything to half. Does it make sense to add fp16 quantization in the compilation passes then?

TedThemistokleous:
> > Removing the qdq should be decided by the backend target based on what quantized operators it supports. We can't run this pass in quantize_fp16 because backends can't undo the quantization without losing precision.
> 
> Ok fair enough. So how should we deal with this scenario? I attempted to manipulate the quantize_fp16 pass so that it would preserve the qdq ops, but doing this is problematic in many ways and also introduces a lot of edge cases. I would say that if we are doing both int8/fp8 quantization + fp16 quantization, it should always do int8/fp8 first (ie. resolve qdqs) and then convert everything to half. Does it make sense to add fp16 quantization in the compilation passes then?

So is the intent here to support mixed precision?

shivadbhavsar:
> > > Removing the qdq should be decided by the backend target based on what quantized operators it supports. We can't run this pass in quantize_fp16 because backends can't undo the quantization without losing precision.
> > 
> > 
> > Ok fair enough. So how should we deal with this scenario? I attempted to manipulate the quantize_fp16 pass so that it would preserve the qdq ops, but doing this is problematic in many ways and also introduces a lot of edge cases. I would say that if we are doing both int8/fp8 quantization + fp16 quantization, it should always do int8/fp8 first (ie. resolve qdqs) and then convert everything to half. Does it make sense to add fp16 quantization in the compilation passes then?
> 
> So is the intent here to support mixed precision?

Ya I guess its more like mixed quantization (technically int8 +fp32 is already mixed precision). But this came about from me trying to replicate some of the driver based tests where we pass both fp16 and int8 flags, and realizing its not possible to do via the APIs on pre-quantized models

shivadbhavsar:
Closing this for now, will solve issue in a different PR

Issue #2542, opened 2023-12-11
Testing other GELU approximations


CharlieL7:
* We previously tested the sigmoid and tanh GELU approximations and found the sigmoid approximation to be faster
  * `GELU ~= 0.5 * x( 1 + tanh( sqrt(2 / pi) * ( x + 0.044715 * x ^ 3) )`
  * `GELU ~= x * sigmoid(1.702 * x) `
* There are other ways to arrange the approximations however using algebraic tricks that may lead to better numerical characteristics.
* One reformulation of the tanh approximation is used by the ROCM EP
* We should test these other methods and examine how other libraries like CK do these approximations.

CharlieL7:
Example from ORT ROCM EP
```
struct FastGeLU {
  template <typename T>
  __host__ __device__ __forceinline__ void operator()(T& y, const T& x) const {
    constexpr const float b = 0.7978845608028654f;  // sqrt(2.0/M_PI)

    // const T cdf = a + a * _Tanh(in * (c * in * in + b));
    const T xb = x * T(b);
    const T u = xb * T(0.044715f) * x * x + xb;
    const T emu = __expf(-u - u);
    const T cdf = T(1.0f) / (T(1.0f) + emu);
    y = x * cdf;
  }
};
```

CharlieL7:
Version of `find_tanh_fast_gelu` matcher that replaces with `erf()` :
```
        auto ins      = r.result;
        auto x        = r.instructions["x"];
        auto sqrt1_2  = m.add_literal(literal{shape{x->get_shape().type()}, {M_SQRT1_2}});
        auto one      = m.add_literal(literal{shape{x->get_shape().type()}, {1.0f}});
        auto one_half = m.add_literal(literal{shape{x->get_shape().type()}, {0.5f}});
        auto a        = insert_common_op(m, ins, make_op("mul"), {x, sqrt1_2});
        auto erf      = m.insert_instruction(ins, make_op("erf"), a);
        auto add_erf  = insert_common_op(m, ins, make_op("add"), {one, erf});
        auto b        = insert_common_op(m, ins, make_op("mul"), {one_half, add_erf});
        auto y        = m.insert_instruction(ins, make_op("mul"), x, b);
        m.replace_instruction(ins, y);
```

rewrite_gelu test:
```
        auto ins      = r.result;
        auto x        = r.instructions["x"];
[...]

PR #2589, opened 2024-01-03
Tanh GELU matcher and matcher.has_value() fix

CharlieL7:
* Adds a matcher that finds the tanh GELU approximation found in distillgpt2
  * Note v0 vs v1 matchers that are slightly different 
* Replaces the tanh GELU approximation with a more numerically stable tanh GELU approximation
  * Algebraically equivalent, just rearranged operations
* Changes `matcher.has_value()` to convert the type of the literal before trying to compare the values
  * Was having an issue with matching fp16 constants in a fp16 model (fp16 constants pre-quantized in the model and no longer possible to compare float to fp16 value)
* Adds a new version of `matcher.has_value()` that uses multiples of the floating point epsilon values so that smaller floating point values will have reasonable tolerances.


src/rewrite_gelu.cpp
```diff
@@ -53,7 +58,60 @@ struct find_gelu_erf
...
+        auto sqrt1_2  = m.add_literal(literal{shape{x->get_shape().type()}, {M_SQRT1_2}});
+        auto one      = m.add_literal(literal{shape{x->get_shape().type()}, {1.0f}});
+        auto one_half = m.add_literal(literal{shape{x->get_shape().type()}, {0.5f}});
+        auto a        = insert_common_op(m, ins, make_op("mul"), {x, sqrt1_2});
+        auto erf      = m.insert_instruction(ins, make_op("erf"), a);
+        auto add_erf  = insert_common_op(m, ins, make_op("add"), {one, erf});
+        auto b        = insert_common_op(m, ins, make_op("mul"), {one_half, add_erf});
+        auto y        = m.insert_instruction(ins, make_op("mul"), x, b);
+        m.replace_instruction(ins, y);
+        */
```
TedThemistokleous, line 107: Do we want to have this as a comment or as another option for implementations to via some MIGRAPHX_XXX flags we can toggle?

causten (commented):
/data/src/include/migraphx/match/gelu_tanh.hpp:48:14: error: invalid case style for local variable 'either_SQRT2RPI_add' [readability-identifier-naming,-warnings-as-errors]
        auto either_SQRT2RPI_add = either_arg(0, 1)(has_value(sqrt(M_2_PI), 10, 10), add_any_mul);
             ^~~~~~~~~~~~~~~~~~~
             either_sqr_t2_rpi_add

src/include/migraphx/match/gelu_tanh.hpp
```diff
@@ -36,31 +36,44 @@ template <class F>
...
     auto tanh_fn() const
     {
-        return f("tanh")(
-            used_once(),
-            arg(0)(f("mul")(either_arg(0, 1)(has_value(sqrt(M_2_PI), 1e-3),
-                                             f("add")(any_arg(0, 1)(f("mul")(either_arg(0, 1)(
-                                                 has_value(0.044715f), pow_fn()))))))));
+        auto mul_const_pow = f("mul")(either_arg(0, 1)(has_value(0.044715f, 10, 10), pow_fn()));
+        auto add_any_mul         = f("add")(any_arg(0, 1)(mul_const_pow));
+        auto either_SQRT2RPI_add = either_arg(0, 1)(has_value(sqrt(M_2_PI), 10, 10), add_any_mul);
```
pfultz2, line 48: This should include the multiply. It looks strange to seperate the arguments from the operator especially since this is only valid for commutative operators like `mul`.

src/include/migraphx/matcher.hpp
```diff
@@ -857,23 +857,69 @@ auto skip_broadcasts_converts(Ms... ms)
...
+        return f(l);
+    }));
+}
+
+/**
+ * Uses integer multiples of the corresponding floating point epsilon and
+ * compares with abs(y - x) < eps * (atol_mult + rtol_mult * abs(y))
+ */
+template <class T>
+inline auto has_value(T x, std::size_t atol_mult, std::size_t rtol_mult)
```
pfultz2, line 878: I dont think its a good idea to have an absolute and relative tolerance like this since they are very different range of numbers. The `atol_mult` is a multiple of epsilon so you would expect a large number like 80 or so, but `rtol_mult` is a multiply of the absolute value so it should be a very small number like `1e-6`. You make this same mistake in the gelu matcher as well.

src/include/migraphx/match/gelu_tanh.hpp
```diff
@@ -36,31 +36,44 @@ template <class F>
...
-    auto matcher() const
+    /// x * 0.5 * (1.0 + tanh( sqrt(2/M_PI) * (x + 0.044715 * x ^ 3) ) )
+    auto matcher_v1() const
     {
-        return f("mul")(used_once(),
-                        either_arg(0, 1)(any().bind("x"),
-                                         f("add")(any_arg(0, 1)(f("mul")(
-                                             either_arg(0, 1)(has_value(0.5f), tanh_fn()))))));
+        auto add_one_tanh =
+            f("add")(used_once(), either_arg(0, 1)(has_value(1.0, 10, 10), tanh_fn()));
```
pfultz2, line 65: The value of `10` is way too large for a relative tolerance.

src/include/migraphx/match/gelu_tanh.hpp
```diff
@@ -36,31 +36,44 @@ template <class F>
...
-                                                 has_value(0.044715f), pow_fn()))))))));
+        auto mul_const_pow = f("mul")(either_arg(0, 1)(has_value(0.044715f, 10, 10), pow_fn()));
+        auto add_any_mul         = f("add")(any_arg(0, 1)(mul_const_pow));
+        auto either_SQRT2RPI_add = either_arg(0, 1)(has_value(sqrt(M_2_PI), 10, 10), add_any_mul);
+        return f("tanh")(used_once(), arg(0)(f("mul")(either_SQRT2RPI_add)));
+    }
+
+    /// x * (0.5? + 0.5 * tanh( sqrt(2/M_PI) * (x? + 0.044715 * x? ^ 3) ) )
+    /// <item>? question mark means it doesn't explicitly match that item (anything will work)
+    auto matcher_v0() const
```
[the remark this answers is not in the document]
CharlieL7: the V0 version is what used to be there. It doesn't match to anything that I know of since we were not using it before. The V1 version is what I found in distilgpt2

src/include/migraphx/matcher.hpp
```diff
@@ -857,23 +857,69 @@ auto skip_broadcasts_converts(Ms... ms)
...
+            }
         });
         return b;
-    }));
+    });
+}
+
+/// Check literal value with an explicit floating point tolerance
+template <class T, class F, MIGRAPHX_REQUIRES(std::is_floating_point<F>{})>
+inline auto has_value(T x, F tolerance)
```
pfultz2, line 900: I think this should be removed.

src/include/migraphx/matcher.hpp
pfultz2, line 878: Can we set default values for this?

src/rewrite_gelu.cpp
```diff
@@ -53,7 +58,60 @@ struct find_gelu_erf
...
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        // Algebraically rearranged version of the tanh GELU approximation that is more numerically
+        // stable
+        auto ins        = r.result;
+        auto x          = r.instructions["x"];
+        auto sqrt_2_rpi = m.add_literal(
+            literal{shape{x->get_shape().type()},
+                    {0.7978845608028653558798921198687637369517172623298693153318516593}});
```
pfultz2, line 79: Can we write this as `sqrt(2.0f) / M_PI` for the value?

src/rewrite_gelu.cpp
```diff
@@ -53,7 +58,60 @@ struct find_gelu_erf
...
+        auto fit_const = m.add_literal(literal{shape{x->get_shape().type()}, {0.044715f}});
```
pfultz2, line 80: What is this value? Can it be computed as well?

src/rewrite_gelu.cpp
```diff
@@ -53,7 +58,60 @@ struct find_gelu_erf
...
+        auto b        = insert_common_op(m, ins, make_op("mul"), {one_half, add_erf});
+        auto y        = m.insert_instruction(ins, make_op("mul"), x, b);
+        m.replace_instruction(ins, y);
+        */
+    }
+};
+
+void rewrite_gelu::apply(module& m) const
+{
+    match::find_matches(m, find_gelu_erf{}, find_tanh_fast_gelu{});
```
pfultz2, line 113: `find_gelu_erf` should only be enabled when fast math is enabled. `find_tanh_fast_gelu` should always be used.

src/rewrite_gelu.cpp
```diff
@@ -53,7 +58,60 @@ struct find_gelu_erf
...
-void rewrite_gelu::apply(module& m) const { match::find_matches(m, find_gelu_erf{}); }
+/**
+ * Find fastGELU blocks (where the graph already does a GELU approximation) and replace them
+ * with an alternative approximation that is less likely to overflow.
+ * The replacement approximation is equivalent to:
+ * GELU(x) ~= 0.5 * x * ( 1 + tanh( sqrt(2/M_PI) * (x + 0.044715 * x^3)))
+ */
+struct find_tanh_fast_gelu
+{
+    auto matcher() const { return match::gelu_tanh(); }
```
pfultz2, line 69: Is this only more numerically stable for `gelu_tanh` or is it more stable then `gelu_erf`?

src/rewrite_gelu.cpp
CharlieL7: Sure, probably want to do the calculation with doubles.

src/rewrite_gelu.cpp
CharlieL7: This is a fitting constant, so it cannot be computed.

src/rewrite_gelu.cpp
CharlieL7: It's definitely more numerically stable that the original `gelu_tanh`.

src/include/migraphx/matcher.hpp
```diff
@@ -857,23 +857,69 @@ auto skip_broadcasts_converts(Ms... ms)
...
-                   v.begin(), v.end(), [&](auto val) { return std::fabs(val - x) < tolerance; }))
+            // cast to the literal's data type before comparing
+            using type = typename decltype(v)::value_type;
+            auto eps   = std::numeric_limits<type>::epsilon();
+            if(std::all_of(v.begin(), v.end(), [&](auto val) {
+                   return std::fabs(val - static_cast<type>(x)) <
+                          eps * (atol_mult + rtol_mult * std::fabs(val));
+               }))
+            {
                 b = true;
```
lakhinderwalia, line 891: Programming style (optional): return true here (and a false later).

src/rewrite_gelu.cpp
```diff
@@ -53,7 +58,60 @@ struct find_gelu_erf
...
+            literal{shape{x->get_shape().type()},
+                    {0.7978845608028653558798921198687637369517172623298693153318516593}});
+        auto fit_const = m.add_literal(literal{shape{x->get_shape().type()}, {0.044715f}});
+        auto one       = m.add_literal(literal{shape{x->get_shape().type()}, {1.0f}});
+        auto xb        = insert_common_op(m, ins, make_op("mul"), {x, sqrt_2_rpi});
+        auto a         = insert_common_op(m, ins, make_op("mul"), {xb, fit_const});
+        auto b         = m.insert_instruction(ins, make_op("mul"), a, x);
+        auto c         = m.insert_instruction(ins, make_op("mul"), b, x);
+        auto u         = m.insert_instruction(ins, make_op("add"), c, xb);
+        auto neg_u     = m.insert_instruction(ins, make_op("neg"), u);
```
lakhinderwalia, line 87: I don't understand how tanh is computed here, please. How are you calculating e**(x) and e**(-x) here? Thanks.

src/include/migraphx/matcher.hpp
pfultz2: You cant return values from a visitor.

src/rewrite_gelu.cpp
pfultz2: How is it fitted? Is it using a lagrange polynomial or taylor series?

lakhinderwalia: Specified in https://onnx.ai/onnx/operators/onnx__Gelu.html

src/rewrite_gelu.cpp
```diff
@@ -53,7 +58,60 @@ struct find_gelu_erf
     }
 };
 
-void rewrite_gelu::apply(module& m) const { match::find_matches(m, find_gelu_erf{}); }
+/**
+ * Find fastGELU blocks (where the graph already does a GELU approximation) and replace them
+ * with an alternative approximation that is less likely to overflow.
+ * The replacement approximation is equivalent to:
+ * GELU(x) ~= 0.5 * x * ( 1 + tanh( sqrt(2/M_PI) * (x + 0.044715 * x^3)))
```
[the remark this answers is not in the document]
lakhinderwalia: (Not your code. But in a similar light of documentation: what does 1.702f do for find_gelu_erf().. If it is a replacement for 1/sqrt(2), then it is incorrect -- perhaps it is something else..)

src/include/migraphx/matcher.hpp
lakhinderwalia: I see. Thanks.

src/rewrite_gelu.cpp
CharlieL7: [tanh_gelu_scan2.pdf](https://github.com/ROCm/AMDMIGraphX/files/13832504/tanh_gelu_scan2.pdf)
I've attached scanned document with the derivation of the rearranged approximation.

src/rewrite_gelu.cpp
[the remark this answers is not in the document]
CharlieL7: I can write out an expanded form or the steps from the document shown in the last response I made.

src/include/migraphx/matcher.hpp
```diff
@@ -857,23 +857,69 @@ auto skip_broadcasts_converts(Ms... ms)
...
             return false;
         auto l = ins->get_literal();
         if(l.empty())
             return false;
+        return f(l);
+    }));
+}
+
+/**
+ * Uses integer multiples of the corresponding floating point epsilon and
```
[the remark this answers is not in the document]
CharlieL7: Uses a multiple of floating point epsilon for the absolute and relative tolerances to compare the floating point values against. `x` is the value of of the literal in the model. `y` is the value that we are looking for.

src/rewrite_gelu.cpp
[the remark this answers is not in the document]
CharlieL7: `1.702f` should be the same fitting constant for the sigmoid approximation.

src/rewrite_gelu.cpp
CharlieL7: The fitting is based on the original GELU paper by Hendrycks and Gimpel (https://arxiv.org/abs/1606.08415) who reference a paper by Choudhry (https://www.hrpub.org/download/20140305/MS7-13401470.pdf). They probably used some measure of difference between the original erf() version and `0.5(1+tanh(y))` and did a fitting.

src/include/migraphx/matcher.hpp
CharlieL7: Yes, if the other versions are removed.

src/include/migraphx/matcher.hpp
CharlieL7: We can remove this, but what would be the desired functionality of previous calls that used a fixed tolerance?

src/rewrite_gelu.cpp
CharlieL7: I kept this as a comment in case we want to switch the behavior in the future. Not sure how we want to do it.

src/include/migraphx/matcher.hpp
pfultz2: This seems like a much better implementation than the older tolerance version. We never used the tolerance parameter except for gelu, most likely because of fp16 which this implementation should handle better.

src/rewrite_gelu.cpp
causten: Can you add a comment in the code where you got the number from

src/rewrite_gelu.cpp
CharlieL7: Sure, I can place the code and test into an issue. I don't wan't to expand the scope of this PR any more than it already has been expanded.

src/include/migraphx/match/gelu_tanh.hpp
TedThemistokleous: Can we also make sqrt(M_2_PI)) a const?

src/rewrite_gelu.cpp
TedThemistokleous: Completely agree.  I'm more concerned with separating it out here, tracking this for a later PR/work and to reduce scope of what we want to be reviewing.

src/include/migraphx/match/gelu_tanh.hpp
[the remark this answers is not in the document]
pfultz2: Can we just combine them with:

```cpp
auto matcher() const
{
    return any_of(matcher_v0(), matcher_v1());
}
```

?

lakhinderwalia (approved):
Approved. 
It would be nice to document the alternate replacement of tanh (based on the identity function that you have mentioned.). Also, my personal style would be: atol_xx could be called abs_tol_xx, and rtol_xx could be called rel_tol_xx.

src/rewrite_gelu.cpp
CharlieL7: I added the code to #2542. Will be removing from this PR.

src/include/migraphx/match/gelu_tanh.hpp
```diff
@@ -36,31 +36,43 @@ template <class F>
...
     auto tanh_fn() const
     {
-        return f("tanh")(
-            used_once(),
-            arg(0)(f("mul")(either_arg(0, 1)(has_value(sqrt(M_2_PI), 1e-3),
-                                             f("add")(any_arg(0, 1)(f("mul")(either_arg(0, 1)(
-                                                 has_value(0.044715f), pow_fn()))))))));
+        auto mul_const_pow    = f("mul")(either_arg(0, 1)(has_value(0.044715), pow_fn()));
+        auto add_any_mul   = f("add")(any_arg(0, 1)(mul_const_pow));
+        auto mul_SQRT2RPI_add = f("mul")(either_arg(0, 1)(has_value(sqrt(M_2_PI)), add_any_mul));
```
pfultz2, line 48: `mul_SQRT2RPI_add` doesn't match the casing style. It should be `mul_sqrt2rpi_add`. I think clang tidy is complaining about that.

src/include/migraphx/match/gelu_tanh.hpp
```diff
@@ -36,31 +36,43 @@ template <class F>
...
     }
 };
 } // namespace detail
 
 template <class F>
 auto gelu_tanh(F f)
 {
-    return detail::gelu_tanh_matcher<F>{f}.matcher();
+    return any_of(detail::gelu_tanh_matcher<F>{f}.matcher_v0(),
+                  detail::gelu_tanh_matcher<F>{f}.matcher_v1());
```
pfultz2, line 75: You should just construct `gelu_tanh_matcher` and call a `matcher` method. Inside the `matcher` method you can do `any_of`.

pfultz2 (approved):

src/include/migraphx/matcher.hpp
```diff
@@ -857,23 +857,56 @@ auto skip_broadcasts_converts(Ms... ms)
...
+                       v.begin(), v.end(), [&](auto val) { return val == static_cast<type>(x); }))
+                {
+                    b = true;
+                }
+            }
+            else
+            {
+                auto eps = std::numeric_limits<type>::epsilon();
+                if(std::all_of(v.begin(), v.end(), [&](auto val) {
+                       return std::fabs(val - static_cast<type>(x)) <
```
umangyadav, line 900: Comment: All values inside literal are compare against `x`.  That would probably mean all values inside literal `y` are of same value.

pfultz2: We dont know if they are all the same value, that is why we check that they are the same value.

PR #2590, opened 2024-01-03
Make the tbb dependency conditional

pfultz2:


TedThemistokleous (approved):

TedThemistokleous:
I guess todo enable another CI pass later with Executors on if we know the distro we're running supports tbb2 or equivalent? @pfultz2

PR #2602, opened 2024-01-05
allow skipping ops in const prop

shivadbhavsar:
Solving issue described in #2587 

Changes:
1) Define set of ops to be skipped with propagating constants
2) Skip qdq ops in `optimize_module`  when calling `quantize_fp16`


umangyadav (approved):

src/include/migraphx/optimize_module.hpp
```diff
@@ -38,6 +39,7 @@ struct module_pass_manager;
  */
 struct MIGRAPHX_EXPORT optimize_module
 {
+    std::unordered_set<std::string> const_prop_skip_ops = {};
```
pfultz2, line 42: Can we prefix the variable with the name of the pass? Like `propagate_constant_skip_ops`?

pfultz2 (approved):

PR #2608, opened 2024-01-05
Use mlir_mode::fast for dot operations on Navi and reduce k threshold to 1024

turneram:


pfultz2:
This looks like this is missing the k<1024 change.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -416,14 +416,15 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
...
+        mpm,
+        find_mlir_fused_ops{.conv_mode = get_mode("fused", mlir_mode::fast),
+                            .dot_mode  = is_navi ? mlir_mode::fast : get_mode("fused", mode)});
 
     match::find_matches(
         mpm,
-        find_mlir_standalone_convolution_op{get_mode("convolution", mlir_mode::int8)},
-        find_mlir_standalone_dot_op{get_mode("dot", mlir_mode::none)});
+        find_mlir_standalone_convolution_op{get_mode("convolution", mlir_mode::fast)},
+        find_mlir_standalone_dot_op{is_navi ? mlir_mode::fast : get_mode("dot", mlir_mode::none)});
```
pfultz2, line 427: This should probably be `get_mode("dot", mode)` so we use MLIR for dot fusions when using the `MIGRAPHX_ENABLE_EXTRA_MLIR` setting.

pfultz2: Sorry thats not the case. This is standalone.

pfultz2 (approved):

pfultz2:
Can this be targeted to develop and then cherry picked into the release branch?

causten:
This can be closed.  We won't be using this technique

causten:
There will not be a need to target this against rocm-rel-6.0.  If you have other uses in 6.1 then go for it but a new PR might be easier

PR #2613, opened 2024-01-09
Use the quantized output when an operator is used again

pfultz2:
This reduces the standalone qlinear ops in resnet50.


pfultz2:
This also reduces some of the qlinear ops in bert as well.

test/simplify_qdq_test.cpp
```diff
@@ -1148,4 +1154,64 @@ TEST_CASE(dot_correctness)
...
+        auto out_scale1 = add_scale_mul(m2, scale, scale, 1, 1, sh.lens());
+        auto d1         = add_quantize_op(m2, "dequantizelinear", dot1, out_scale1);
+        auto add1       = m2.add_instruction(migraphx::make_op("add"), d1, y);
+
+        auto q3         = add_quantize_op(m2, "quantizelinear", add1, scale, zero);
+        auto q4         = add_quantize_op(m2, "quantizelinear", w2, scale, zero);
+        auto dot2       = m2.add_instruction(migraphx::make_op("quant_dot"), q3, q4);
+        auto out_scale2 = add_scale_mul(m2, scale, scale, 1, 1, sh.lens());
+        auto d2         = add_quantize_op(m2, "dequantizelinear", dot2, out_scale2);
+        auto d3   = add_quantize_op(m2, "dequantizelinear", q3, q3->inputs()[1], q3->inputs()[2]);
```
shivadbhavsar, line 1208: I dont quite understand why this is desirable. Seems to me like this just degrades accuracy by using a "fake quantized" result. Why do we want to add `d3` rather than just adding `add1` (which is already float)?

pfultz2: Because then the qlinear op cannot be fused with add. Ideally we want the output of the activations to be int8 which will go into the next convolution or another pointwise if thats the case.

shivadbhavsar: Ok I see, I'm not totally familiar with all the fusion passes but I think this makes sense

shivadbhavsar (approved):

causten:
tidy and format errors

src/simplify_qdq.cpp
```diff
@@ -38,6 +38,7 @@
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
+namespace {
```
umangyadav, line 41: were there any symbol conflicts that required adding another namespace ?

src/simplify_qdq.cpp
```diff
@@ -298,6 +299,46 @@ bool compare_literals(instruction_ref ins1, instruction_ref ins2)
...
+    {
+        return match::name("quantizelinear")(
+            match::used_once(), match::arg(0)(match::none_of(match::used_once()).bind("x")));
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins   = r.result;
+        auto x_ins = r.instructions["x"];
+        assert(ins != x_ins);
```
umangyadav, line 321: This seems impossible.

src/simplify_qdq.cpp
pfultz2: Not that I know, but this will help prevent ODR issues(which is difficult to debug) in the future. We should use this for all classes define in .cpp files though.

src/simplify_qdq.cpp
pfultz2: I had misplaced the bind in the matcher which caused this failure. This is just to catch such mistakes in the future.

umangyadav (approved):

causten:
@pfultz2  this cant go in until the unit tests pass

pfultz2:
@causten Fixed test case.

causten:
The failing Perf run of resnet50v1_int8 is...

terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 2) >= this->size() (which is 2)
timeout: the monitored command dumped core

Can't merge until this is resolved.  Don't worry about the bert case, there is a path error in the script

PR #2617, opened 2024-01-09
Always use mlir for quantized convolution

pfultz2:
Previously the logic checked for the output type of convolution which was incorrect since its int32 and not int8. Instead this checks the inputs, so now `quant_convolution` will always run with mlir.


krzysz00 (approved):
Seems fine to me (assuming the commit pointer's been bumped to include Yi's group conv2d support in a different PR)

umangyadav (approved):

causten:
@pfultz2 there appears to be some CI failures with performance

PR #2618, opened 2024-01-09
Add the missing cmake file for parallel stl

pfultz2:


cmake/FindParallelSTL.cmake
```diff
@@ -0,0 +1,45 @@
...
+        if(ParallelSTL_TBB_HAS_EXECUTION_PAR)
+            set(ParallelSTL_USES_TBB On)
+            set(ParallelSTL_LIBRARIES TBB::tbb)
+            message(STATUS "Using TBB for parallel execution")
+        endif()
+    endif()
+endif()
+
+foreach(VAR ParallelSTL_FOUND ParallelSTL_LIBRARIES ParallelSTL_USES_TBB)
+    string(TOUPPER ${VAR} ParallelSTL_VAR)
```
umangyadav, line 42: why is this necessary ?

pfultz2: Just in case we use the all caps version of the variable name.

umangyadav (approved):

PR #2620, opened 2024-01-09
Fix mlir submodule input parameters

pfultz2:
The parameters should always be standard shape as we will add in the strides later when we know the layout.


umangyadav (approved):

causten:
Performance CI failures need to be reviewed

pfultz2:
I dont see a perf difference locally.

PR #2626, opened 2024-01-11
Cleanup LSTM and RNN activation functions

nives-vukovic:
- Align and add check for valid number of activation functions in LSTM and RNN parser
- Cleanup LSTM and RNN operator calls with invalid number of activation functions

Resolves: https://github.com/migraphx-benchmark/AMDMIGraphX/issues/160


nives-vukovic:
@TedThemistokleous  Additional tests have been added to improve coverage.

CharlieL7 (approved):

TedThemistokleous (approved):

TedThemistokleous:
lgtm now too. had to rekick ci for jenkins

src/onnx/parse_lstm.cpp
```diff
@@ -194,7 +122,10 @@ struct parse_lstm : op_parser<parse_lstm>
             MIGRAPHX_THROW("LSTM: incorrect direction attribute");
         }
 
+        // set default activation functions
         std::vector<std::string> vec_names = {"sigmoid", "tanh", "tanh"};
+        lstm_actv_functions(dirct, vec_names);
```
lakhinderwalia, line 127: You could easily get rid of lstm_actv_functions, by default filling line 126, above. Just as in parse_rnn.cpp.

src/onnx/parse_lstm.cpp
```diff
@@ -194,7 +122,10 @@ struct parse_lstm : op_parser<parse_lstm>
...
+
```
lakhinderwalia, line 128: This could be an issue with many operators, I am realizing, and I should have been more careful in the past perhaps even with my own code, as I look at the default attribute specifications of string type: Please note, that an empty string can be passed for an activation function each, in the list of the required 3 or 6 activation function names. Example: ['', 'tanh', ''] is likely valid. In that case, the default should ideally be used for those empty strings. It is easy to do here, actually, just by updating vec_names[i] only if names[i] is not empty.

If this assumption is not valid (about empty strings), you can always ignore this comment. Thanks.

src/onnx/parse_lstm.cpp
nives-vukovic: @lakhinderwalia  It is not completely the same as in rnn default is 'tanh','tanh' (as for bidirectional) and here it is 'sigmoid','tanh','tanh' as for one-directional. However, I did want to get rid of the lstm_actv_functions and place the code in the parse function, but the tidy check was complaining about exceeding recommended complexity (having 16 branches instead of maximal 15), so I left that part in this shortened version of  lstm_actv_functions. If you prefer changing it somehow, it is not an issues, but I didn't find a way not to exceed it.

src/onnx/parse_lstm.cpp
nives-vukovic: I believe we are in line with the specification regarding this point. In case a user specifies ['', 'tanh', ''], we will report an error as the program will not recognize "" as a valid activation function. This is the same approach onnx runtime uses as far as I could see, first the parsing:https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cpu/rnn/lstm_base.h and the activation function validation https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cpu/rnn/rnn_helpers.cc#L714

src/onnx/parse_lstm.cpp
lakhinderwalia: This is not a big issue. Simplification of the code can be done in many ways, but that shouldn't block this PR at this point, IMO.

src/onnx/parse_lstm.cpp
lakhinderwalia: Just to be clear, from the spec it implies that ['', 'tanh', ''] should fill in as ['sigmoid', 'tanh', 'sigmoid']. But it is an edge case, so let us not worry about it here for now. Thanks for your response.

lakhinderwalia (approved):
Thanks for doing these changes, and responding to my questions.

PR #2629, opened 2024-01-11
Dynamic broadcasting update and `dot_broadcast` operator

CharlieL7:
* Updates the broadcasting rules for dynamic shapes:
  *  Allows for broadcasting between `dynamic_dimensions` where one contains the other
  * For example:
```
a = shape{ dyn_dims={{1, 100}, {4, 4}} }
b = shape{ dyn_dims={{4, 20}, {0, max_int} }
broadcast(a, b) -> shape{ dyn_dims={{4, 20}, {4, 4}} }
```
  * This change is needed because of how certain operators can create dynamic shapes with `dynamic_dimensions` `{0, max_int}` and other mismatches (ex: `allocate`, `resize`).
* Adds the `dot_broadcast` operator to handle broadcasting dynamic shapes for the `dot` operator.
  * It essentially broadcasts dimensions other than the last two.


src/include/migraphx/op/dot_broadcast.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * This operator is only needed if one of the shapes are dynamic.
+ * Example:
+ * a = shape[{1, 4}, 3, 248, 248]
+ * b = shape[248, 365]
+ * dot_broadcast(a, b) => shape[{1, 4}, 3, 248, 248] (no change)
+ * dot_broadcast(b, a) => shape[{1, 4}, 3, 248, 365]
+ */
+struct dot_broadcast
+{
+    std::string name() const { return "dot_broadcast"; }
```
pfultz2, line 48: Maybe we should call this `broadcast_for_dot`. I thought at first this was doing a dot and broadcast together.

src/common.cpp
```diff
@@ -217,5 +225,20 @@ instruction_ref add_common_op(module& m, const operation& op, std::vector<instru
     return insert_common_op(m, m.end(), op, std::move(inputs));
 }
 
+shape make_bcast_shape(const shape& input_shape,
+                       const std::vector<std::size_t>& bcast_lens,
+                       const std::size_t& offset)
+{
+    std::vector<size_t> bcast_strides(bcast_lens.size(), 0);
+    for(std::ptrdiff_t i = input_shape.ndim() - 1; i >= 0; i--)
```
pfultz2, line 233: I think it would be nicer to write it as `for(std::ptrdiff_t i  : reverse(range(input_shape.ndim())))`.

src/common.cpp
```diff
@@ -217,5 +225,20 @@ instruction_ref add_common_op(module& m, const operation& op, std::vector<instru
...
+    {
+        if(bcast_lens[i + offset] == input_shape.lens()[i])
```
pfultz2, line 235: We should probably have an `assert` that `offset` wont go out of range.

bpickrel:
The new broadcasting rule doesn't seem to be documented anywhere, either in code comments or in a *.rst file.  The description for this PR is just a starting point.  As I understand it, if there are two dynamic dimensions and one's range is contained within the other, then the broader range is clamped to the narrower one.  We need to get the full logic laid out clearly in case new problems crop up and the rule has to be revisited.  Also, I don't understand WHY the changed rule makes models work when they didn't before.

CharlieL7:
> The new broadcasting rule doesn't seem to be documented anywhere, either in code comments or in a *.rst file. The description for this PR is just a starting point. As I understand it, if there are two dynamic dimensions and one's range is contained within the other, then the broader range is clamped to the narrower one. We need to get the full logic laid out clearly in case new problems crop up and the rule has to be revisited. Also, I don't understand WHY the changed rule makes models work when they didn't before.

The previous broadcasting rule was simply that everything had to match exactly or the other dimension was 1. This change most importantly allows for handling the {0, max_int} dynamic_dimension case.

src/common.cpp
CharlieL7: Made it call `.at()` for size checking

src/common.cpp
```diff
@@ -76,16 +74,32 @@ std::vector<shape::dynamic_dimension> compute_broadcasted_dyn_dims(shape s0, sha
                        {
                            return b;
                        }
+                       else if(a.within_range(b))
+                       {
+                           return shape::dynamic_dimension{a.min, a.max};
```
umangyadav, line 79: It seems strange a bit. Broadcasts are generally used to expand dimensions's range. But here for the dynamic case it picks one that is smaller in range. 
is there a good explanation on why this would be necessary ?

src/include/migraphx/op/broadcast_for_dot.hpp
```diff
@@ -0,0 +1,94 @@
...
+        if(s0.dynamic() or s1.dynamic())
+        {
+            s0           = s0.to_dynamic();
+            s1           = s1.to_dynamic();
+            auto dds0_it = s0.dyn_dims().end() - 2;
+            auto dds1_it = s1.dyn_dims().end() - 2;
+            std::vector<shape::dynamic_dimension> sliced_dds0{s0.dyn_dims().begin(), dds0_it};
+            std::vector<shape::dynamic_dimension> sliced_dds1{s1.dyn_dims().begin(), dds1_it};
+            auto output_dyn_dims = compute_broadcasted_dyn_dims(sliced_dds0, sliced_dds1);
+            output_dyn_dims.insert(output_dyn_dims.end(), dds0_it, s0.dyn_dims().end());
```
umangyadav, line 66: Why is it just picking last two dims from `s0` ?

src/simplify_dyn_ops.cpp
```diff
@@ -318,6 +319,39 @@ struct find_const_alloc_fill
...
+                                                match::arg(1)(match::static_shape()));
+    }
+
+    void apply(module& m, const match::matcher_result& mr) const
+    {
+        auto broadcast_for_dot_ins = mr.result;
+        auto inputs                = broadcast_for_dot_ins->inputs();
+        auto s0                    = inputs.at(0)->get_shape();
+        auto s1                    = inputs.at(1)->get_shape();
+        auto l0_it                 = s0.lens().begin() + s0.ndim() - 2;
```
umangyadav, line 343:
```suggestion
        auto l0_it                 = s0.lens().end() - 2;
```

src/common.cpp
umangyadav: I read earlier comments that it's mainly about handling {0, max_int} case. I agree with @bpickrel  that we should document this somewhere that explains why  {0, max_int} shape arises in certain cases and why broadcasting this way would be correct solution.

src/include/migraphx/op/broadcast_for_dot.hpp
CharlieL7: It's broadcasting the leading dimensions for the `dot` operator

src/common.cpp
umangyadav: docstring explains the logic. I was looking more of answer to design decision. "broadcast" is not an appropriate word because the context in which the word "broadcast" is used is different compared to logic used here. I don't have better alternative suggestion though. may be call it `compute_common_dyn_dims_for_dot` ? but then `broadcast_for_dot` would require a new name as well.

umangyadav: >  of answer to design decision 

I mean it would be helpful to put some sentences in what specific case this type of broadcasting is useful and why it's necessary to do it this way.

CharlieL7: > broadcast" is not an appropriate word because the context in which the word "broadcast" is used is different compared to logic used here. 
It actually is broadcasting, it just so happens that matching `dynamic_dimensions` can be more lenient. At runtime the static dimensions have to be the same (unless you have a fixed 1 dimension). The only way that works is if the output dynamic_dimension range is the smaller one. There's technically an edge case with ranges that include 1, but it's pretty contrived.

umangyadav: I like this explanation

umangyadav (approved):

src/common.cpp
```diff
@@ -217,5 +225,20 @@ instruction_ref add_common_op(module& m, const operation& op, std::vector<instru
     return insert_common_op(m, m.end(), op, std::move(inputs));
 }
 
+shape make_bcast_shape(const shape& input_shape,
```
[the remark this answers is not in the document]
CharlieL7: I'll add a docstring to explain this function.

bpickrel, line 228:
```suggestion
/** Make a broadcasting shape from a subset of input shape's dimensions and strides.

 input_shape   Input shape.  Must be static.
 bcast_lens      Dimensions of output shape.
 offset              Index offset to use.  bcast_lens.size() + offset must be no more than input_shape.ndim()
  

*/
shape make_bcast_shape(const shape& input_shape,
```

bpickrel (approved):
See one last suggestion for a code comment.

src/common.cpp
```diff
@@ -51,21 +51,19 @@ std::vector<std::size_t> compute_broadcasted_lens(std::vector<std::size_t> s0,
...
     {
-        std::swap(s0, s1);
+        std::swap(dds0, dds1);
     }
-    auto offset = s1.ndim() - s0.ndim();
-    std::vector<shape::dynamic_dimension> out_dims(s1.dyn_dims());
-    std::transform(s0.dyn_dims().cbegin(),
-                   s0.dyn_dims().cend(),
-                   s1.dyn_dims().cbegin() + offset,
+    auto offset = dds1.size() - dds0.size();
```
lakhinderwalia, line 62: Style: you could iterate from the back of these shapes, without calculating/using the offset..

src/common.cpp
```diff
@@ -51,21 +51,19 @@ std::vector<std::size_t> compute_broadcasted_lens(std::vector<std::size_t> s0,
...
+    std::vector<shape::dynamic_dimension> out_dims(dds1);
+    std::transform(dds0.cbegin(),
+                   dds0.cend(),
+                   dds1.cbegin() + offset,
```
lakhinderwalia, line 66: Does this take care of multiple '1' in a dynamic dimension, or that will not occur, as a pre-condition? Thanks.

src/common.cpp
lakhinderwalia, line 79: why not return just 'a' ?

CharlieL7: don't want to include the optimals value if it exists

src/include/migraphx/op/broadcast_for_dot.hpp
```diff
@@ -0,0 +1,94 @@
...
+        check_shapes{inputs, *this, true}.has(2);
+        check_shapes{{inputs.at(0)}, *this, true}.min_ndims(2);
+        check_shapes{{inputs.at(1)}, *this, true}.min_ndims(2);
+        auto s0 = inputs.at(0);
+        auto s1 = inputs.at(1);
+        if(s0.dynamic() or s1.dynamic())
+        {
+            s0           = s0.to_dynamic();
+            s1           = s1.to_dynamic();
+            auto dds0_it = s0.dyn_dims().end() - 2;
```
lakhinderwalia, line 61: s0.resize(s0.size() -2);

src/simplify_dyn_ops.cpp
```diff
@@ -319,6 +320,39 @@ struct find_const_alloc_fill
...
+        return match::name("broadcast_for_dot")(match::arg(0)(match::static_shape()),
+                                                match::arg(1)(match::static_shape()));
+    }
+
+    void apply(module& m, const match::matcher_result& mr) const
+    {
+        auto broadcast_for_dot_ins = mr.result;
+        auto inputs                = broadcast_for_dot_ins->inputs();
+        auto s0                    = inputs.at(0)->get_shape();
+        auto s1                    = inputs.at(1)->get_shape();
```
lakhinderwalia, line 343: Style: a resize() operation is what you are emulating..

lakhinderwalia:
Left some suggestions/comments, in case you end up creating another PR for this.. Else please ignore. Thanks.

src/simplify_dyn_ops.cpp
CharlieL7: I'm not following what you mean here. There's no resize() function for `shape`.

lakhinderwalia: As in, l0_broadcasted_lens is simply a resize of s0 lens. l1_broadcasted_lens too.. similar to https://github.com/ROCm/AMDMIGraphX/pull/2629/files/1adde18494a9918d855ce3ffc27f42b848bd89d5#r1500019530

PR #2630, opened 2024-01-12
Adding additional check, if the graph is empty

ivarusic-amd:
Test_noop from test suite was failing on Windows Debug configuration, because graph in this case is empty and `std::prev(mm->end)` was causing problems.
After adding additional check in tf_test.cpp and tf_parser.cpp:
`auto last_ins = std::prev(mm->end());` 
This line was changed with this:
`auto last_ins = (mm->size() > 0) ? std::prev(mm->end()) : mm->end();`

It passes on both, Windows and Linux. 
@pfultz2 @apwojcik


src/tf/tf_parser.cpp
```diff
@@ -298,7 +298,7 @@ void tf_parser::parse_graph(const tensorflow::GraphDef& graph)
     {
         this->parse_node(p.first);
     }
-    auto last_ins = std::prev(mm->end());
+    auto last_ins = (mm->size() > 0) ? std::prev(mm->end()) : mm->end();
```
pfultz2, line 301: You should do an early return here:

```cpp
if (mm->size() == 0)
    return;
```

Also the `if(last_ins != mm->end())` can be removed as well as its completely redundant.

test/tf/tf_test.cpp
```diff
@@ -63,7 +63,7 @@ migraphx::program optimize_tf(const std::string& name, bool is_nhwc)
                               migraphx::eliminate_identity{}});
 
     // remove the last return instruction
-    auto last_ins = std::prev(mm->end());
+    auto last_ins = (mm->size() > 0) ? std::prev(mm->end()) : mm->end();
```
pfultz2, line 66: Change the condition of `if(last_ins != mm->end())` to `if(mm->size() > 0)`. 

```cpp
if(mm->size() > 0)
{
    auto last_ins = std::prev(mm->end());
    if(last_ins->name() == "@return")
    {
        mm->remove_instruction(last_ins);
    }
}
```

PR #2639, opened 2024-01-12
Fuse group convolution

pfultz2:


src/simplify_algebra.cpp
```diff
@@ -725,6 +725,54 @@ struct find_concat_op
...
+        return match::name("concat")(
+            match::all_of[match::inputs()](match::used_once(), match::name("convolution")));
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins  = r.result;
+        auto axis = ins->get_operator().to_value()["axis"].to<int>();
+        if(axis != 1)
+            return;
```
TedThemistokleous, line 741: Add test case to cover this return path.

TedThemistokleous (changes requested):
Just add a test case to cover the axis != case in your apply. Otherwise LGTM

src/simplify_algebra.cpp
```diff
@@ -725,6 +725,54 @@ struct find_concat_op
...
+        if(ins->inputs().empty())
+            return;
```
umangyadav, line 743: Would this case occure at all ? All of the inputs to the concat must be convolution. It can't be empty

CharlieL7 (approved):
LGTM

TedThemistokleous (approved):
Just fix licensing now. LGTM too

PR #2647, opened 2024-01-16
Fixing export API symbols from dynamic library

urpetkov-amd:
Export API symbol for `migraphx`. API symbol exporting affects only Windows. It is transparent on Linux.

@pfultz2 @apwojcik

### DIFF
--- a/tools/include/pass.hpp
+++ b/tools/include/pass.hpp
@@ -90,7 +90,7 @@ interface('pass',
 %>
 
 /// Used in the targets to enable/disable compiler passes
-pass enable_pass(bool enabled, pass p);
+MIGRAPHX_EXPORT pass enable_pass(bool enabled, pass p);
 
 #endif
 
--- a/src/include/migraphx/pass.hpp
+++ b/src/include/migraphx/pass.hpp
@@ -332,7 +332,7 @@ inline const ValueType& any_cast(const pass& x)
 #endif
 
 /// Used in the targets to enable/disable compiler passes
-pass enable_pass(bool enabled, pass p);
+MIGRAPHX_EXPORT pass enable_pass(bool enabled, pass p);
 
 #endif
 


pfultz2 (approved):

tools/include/pass.hpp
```diff
@@ -81,16 +81,19 @@ void module_pass_manager_apply(const T& x, module_pass_manager& mpm)
...
-%>
+<%interface(
+    'pass',
+    virtual('name', returns = 'std::string', const = True),
+    virtual('apply',
+            returns = 'void',
+            mpm     = 'module_pass_manager &',
+            const   = True,
+            default = 'migraphx::detail::module_pass_manager_apply'),
+    virtual('apply', returns = 'void', p = 'program &', const = True, default = 'migraphx::nop'))%>
```
apwojcik, line 92: Please revert that change, and do not run clang-format on this file - it does not know how to format it correctly because it is a file template and not the regular header file.

tools/include/pass.hpp
```diff
@@ -81,16 +81,19 @@ void module_pass_manager_apply(const T& x, module_pass_manager& mpm)
...
 
 /// Used in the targets to enable/disable compiler passes
-pass enable_pass(bool enabled, pass p);
+MIGRAPHX_EXPORT pass
+enable_pass(bool enabled, pass p);
```
apwojcik, line 96: Make this line look identical to the one from src/include/migrpahx/pass.hpp

apwojcik (changes requested):

tools/include/pass.hpp
urpetkov-amd: Thanks for suggestion, I changed it.

tools/include/pass.hpp
urpetkov-amd: Thanks for suggestion, I changed it.

apwojcik (approved):

apwojcik:
Interestingly, the generated file passed the clang-format check, but the source file failed. @pfultz2, what do you think about this? Should we turn off clang-format checking for the template and, while generating the resulting file, remove the appropriate comments for turning off/on the checking?

umangyadav (approved):

pfultz2:
This should just be merged in as is, and we can try and address the formatting issue later. We might need to use something other than `<%` `%>` for the templated part.

causten:
I'm going to approve this with the known format failure.  We don't like the suggestion

PR #2648, opened 2024-01-17
Enable Asymmetric Quantization

shivadbhavsar:
This PR enables parsing asymmetrically quantized inputs. It also supports asymmetrically quantized  weights (this is not really done in practice, but it's good to have for the sake of completeness).


TedThemistokleous:
Not sure why Jenkins isn't running for this @shivadbhavsar . Maybe it didn't kick in the builds?

TedThemistokleous (approved):

src/simplify_qdq.cpp
```diff
@@ -55,12 +55,12 @@ std::unordered_set<std::string> get_quantizable_op_names()
 struct match_find_quantizable_ops
 {
     static bool
-    is_valid_scale(instruction_ref scale, std::vector<std::size_t> lens, std::size_t axis)
+    is_valid_qparam(instruction_ref scale, std::vector<std::size_t> lens, std::size_t axis)
```
umangyadav, line 58: nit: this is used for both scale and zp. having scale as the parameter name is not a good fit.  same with `qparam_broadcast_op`

umangyadav (approved):

src/simplify_qdq.cpp
shivadbhavsar: ah forgot to update the param name, Ill update it

PR #2656, opened 2024-01-18
use eliminate_convert for the fp8 JIT compilation 

umangyadav:
This PR fixes two things and adds one feature.

1.  Compilation of FP8 on non-MI300 machines. 
2.  Bug in quantizelinear which uses `min()` for the FP8. For the floating points it should use `lowest()`

- Adds capibility to set tolerances case by case basis for the verify tests. 

On non-MI300 machines when compiling model using `--fp8` flag, it generates following kind of JIT kernel. Notice it has quantizelinear + convert as the pattern. This pattern generates two redundant converts. One of the convert is coming from `rewrite_quantization`. https://github.com/ROCm/AMDMIGraphX/blob/3fd8abf5c383cae44bea27a93b72c1c7c90ef3d3/src/targets/gpu/compile_gen.cpp#L186

```
template<class Tx0, class Tx1>
__device__ __attribute__((const)) auto inner_pointwise(Tx0 x0,Tx1 x1) {
// @literal -> float_type, {1}, {0}
auto z0 = float(240);
// @literal -> float_type, {1}, {0}
auto z1 = float(-240);
// @literal -> fp8e4m3fnuz_type, {1}, {0}
auto z2 = migraphx::fp8::fp8e4m3fnuz(0);
// @literal -> float_type, {1}, {0}
auto z3 = float(0.19705340266227722);
// @param:x1 -> float_type, {1}, {0}
// @literal -> float_type, {1}, {0}
auto z5 = float(0.00023211138613987714);
// @param:x0 -> float_type, {1}, {0}
// convert[target_type=2] -> float_type, {1}, {0}
......
// add -> float_type, {1}, {0}
auto z14 = migraphx::convert<float>(z12 + z13);
// clip -> float_type, {1}, {0}
auto z15 = migraphx::convert<float>(migraphx::min(migraphx::max(z1, z14), z0));
// convert[target_type=12] -> fp8e4m3fnuz_type, {1}, {0}
auto z16 = migraphx::convert<migraphx::fp8::fp8e4m3fnuz>(migraphx::convert<migraphx::fp8::fp8e4m3fnuz>(z15));
// convert[target_type=2] -> float_type, {1}, {0}
auto z17 = migraphx::convert<float>(migraphx::convert<float>(z16));
// @return -> float_type, {1}, {0}
return z17;

}
```

This can't be compiled as MIGraphX would try to vectorize based on input and output types of the kernel which is Float but internally it uses FP8.  FP8 can't be vectorized. 

Converts can be eliminated in such cases as they are redundant.

### DIFF
--- a/src/include/migraphx/op/quantizelinear.hpp
+++ b/src/include/migraphx/op/quantizelinear.hpp
@@ -77,7 +77,7 @@ struct quantizelinear
         visit_all(result, y_zero_point)([&](auto output, auto zero_pts) {
             visit_all(x, y_scale)([&](auto input, auto scales) {
                 using quant_type = typename decltype(output)::value_type;
-                auto min_value   = std::numeric_limits<quant_type>::min();
+                auto min_value   = std::numeric_limits<quant_type>::lowest();
                 auto max_value   = std::numeric_limits<quant_type>::max();
                 par_for(output_shape.elements(), [&](auto i) {
                     double quantized = static_cast<double>(std::nearbyint(input[i] / scales[i])) +
--- a/src/targets/gpu/compile_gen.cpp
+++ b/src/targets/gpu/compile_gen.cpp
@@ -21,6 +21,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+#include <migraphx/eliminate_convert.hpp>
 #include <migraphx/gpu/compile_gen.hpp>
 #include <migraphx/gpu/context.hpp>
 #include <migraphx/shape.hpp>
@@ -183,7 +184,10 @@ void generate_pointwise(cpp_generator& gg, const module& pm, const std::string& 
 {
     module m = pm;
     run_passes(m,
-               {rewrite_quantization{}, eliminate_common_subexpression{}, dead_code_elimination{}});
+               {rewrite_quantization{},
+                eliminate_convert{},
+                eliminate_common_subexpression{},
+                dead_code_elimination{}});
     cpp_generator g;
     g.fmap([](const std::string& fname) { return "migraphx::" + fname; });
     g.add_point_op("where", "${function:where}(${0}, ${1}, ${2})");
--- a/test/verify/main.cpp
+++ b/test/verify/main.cpp
@@ -89,5 +89,6 @@ int main(int argc, const char* argv[])
                          "batch_quant_dot_1<migraphx::fp8::fp8e4m3fnuz, float>",
                          "quant_dot_3args_4<migraphx::fp8::fp8e4m3fnuz, float>",
                          "quant_dot_3args_5<migraphx::fp8::fp8e4m3fnuz, float>"});
+    rv.change_tolerance_for("test_quantizelinear_convert", 100000);
     rv.run(argc, argv);
 }
--- a/test/verify/run_verify.cpp
+++ b/test/verify/run_verify.cpp
@@ -188,7 +188,8 @@ auto get_hash(const T& x)
 
 void run_verify::verify(const std::string& name,
                         const migraphx::program& p,
-                        const migraphx::compile_options& c_opts) const
+                        const migraphx::compile_options& c_opts,
+                        size_t tolerance) const
 {
     using result_future =
         std::future<std::pair<migraphx::program, std::vector<migraphx::argument>>>;
@@ -254,7 +255,7 @@ void run_verify::verify(const std::string& name,
             for(std::size_t i = 0; ((i < num) and passed); ++i)
             {
                 passed &= migraphx::verify_args_with_tolerance(
-                    tname, result[i], migraphx::verify::expected{gold[i]});
+                    tname, result[i], migraphx::verify::expected{gold[i]}, tolerance);
             }
 
             if(not passed or migraphx::enabled(MIGRAPHX_TRACE_TEST{}))
@@ -276,7 +277,17 @@ void run_verify::run(int argc, const char* argv[]) const
     for(auto&& p : get_programs())
     {
         labels[p.section].push_back(p.name);
-        test::add_test_case(p.name, [=] { verify(p.name, p.get_program(), p.compile_options); });
+        if(tolerances.find(p.name) != tolerances.end())
+        {
+            test::add_test_case(p.name, [=] {
+                verify(p.name, p.get_program(), p.compile_options, tolerances.at(p.name));
+            });
+        }
+        else
+        {
+            test::add_test_case(p.name,
+                                [=] { verify(p.name, p.get_program(), p.compile_options); });
+        }
     }
     test::driver d{};
     d.get_case_names = [&](const std::string& name) -> std::vector<std::string> {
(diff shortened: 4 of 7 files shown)


test/verify/main.cpp
```diff
@@ -89,5 +89,6 @@ int main(int argc, const char* argv[])
                          "batch_quant_dot_1<migraphx::fp8::fp8e4m3fnuz, float>",
                          "quant_dot_3args_4<migraphx::fp8::fp8e4m3fnuz, float>",
                          "quant_dot_3args_5<migraphx::fp8::fp8e4m3fnuz, float>"});
+    rv.change_tolerance_for("test_quantizelinear_convert", 100000);
```
pfultz2, line 92: Tolerance should be a method on the `verify_program` class. And then `test_quantizelinear_convert` can override it. We can add a `tolerance` member variable to the `program_info` class to store the result of the method.

pfultz2: And does the `tolerance` need to be that high?

test/verify/run_verify.cpp
```diff
@@ -188,7 +188,8 @@ auto get_hash(const T& x)
 
 void run_verify::verify(const std::string& name,
                         const migraphx::program& p,
-                        const migraphx::compile_options& c_opts) const
+                        const migraphx::compile_options& c_opts,
+                        size_t tolerance) const
```
pfultz2, line 192: We should probably pass `program_info` struct directly to this method instead of so many parameters.

src/targets/gpu/compile_gen.cpp
```diff
@@ -183,7 +184,10 @@ void generate_pointwise(cpp_generator& gg, const module& pm, const std::string&
 {
     module m = pm;
     run_passes(m,
-               {rewrite_quantization{}, eliminate_common_subexpression{}, dead_code_elimination{}});
+               {rewrite_quantization{},
+                eliminate_convert{},
+                eliminate_common_subexpression{},
```
pfultz2, line 189: We should just use `optimize_module` here.

test/verify/main.cpp
umangyadav: > And does the tolerance need to be that high? 

Yes, test_verify takes floating point epsilon based on final output. Which is ~1e-7, RMS Error is ~0.01. Therefore need tolerance for 1e5. 

```
FAILED: cpu
RMS Error: 0.0104129
ref:2.75, 1.75, -1.75, 2.5, -1, 1, 0.625, 9
target:2.6875, 1.75, -1.75, 2.5, -1.0625, 1.0625, 0.625, 9.25
Max diff: 0.25
Mismatch at 0: 2.75 != 2.6875
```

Mismatch between numbers are coming due to the fact that For the "ref" implementation it downcasts to "fp8" first and then upcasts to float. 

For the CPU and GPU it optimizes away this converts. 

e.g. 

```
ref:
module: "main"
y_zero_point = @param:y_zero_point -> fp8e4m3fnuz_type, {2, 2, 2}, {4, 2, 1}, target_id=0
y_scale = @param:y_scale -> float_type, {2, 2, 2}, {4, 2, 1}, target_id=0
x = @param:x -> float_type, {2, 2, 2}, {4, 2, 1}, target_id=0
@3 = ref::quantizelinear(x,y_scale,y_zero_point) -> fp8e4m3fnuz_type, {2, 2, 2}, {4, 2, 1}, target_id=0
@4 = ref::convert[target_type=2](@3) -> float_type, {2, 2, 2}, {4, 2, 1}, target_id=0
@5 = @return(@4), target_id=0
```

for the GPU it is : 
```
template<class Tx0, class Tx1, class Tx2>
__device__ __attribute__((const)) auto inner_pointwise(Tx0 x0,Tx1 x1,Tx2 x2) {
// @literal -> float_type, {1}, {0}
auto z0 = float(240);
// @literal -> float_type, {1}, {0}
auto z1 = float(-240);
// @param:x2 -> fp8e4m3fnuz_type, {1}, {0}
// @param:x1 -> float_type, {1}, {0}
// @param:x0 -> float_type, {1}, {0}
// div -> float_type, {1}, {0}
auto z5 = migraphx::convert<float>(x0 / x1);
// nearbyint -> float_type, {1}, {0}
auto z6 = migraphx::convert<float>(migraphx::nearbyint(z5));
// convert[target_type=2] -> float_type, {1}, {0}
auto z7 = migraphx::convert<float>(migraphx::convert<float>(x2));
// add -> float_type, {1}, {0}
auto z8 = migraphx::convert<float>(z6 + z7);
// clip -> float_type, {1}, {0}
auto z9 = migraphx::convert<float>(migraphx::min(migraphx::max(z1, z8), z0));
// @return -> float_type, {1}, {0}
return z9;

}


extern "C" {
MIGRAPHX_GLOBAL void quantizelinear_convert_kernel(void * private_p0,void * private_p1,void * private_p2,void * private_p3)
{
    auto idx = make_index();
    pointwise(idx, vectorize<1, 0>())(MIGRAPHX_LIFT(inner_pointwise), private_p0,private_p1,private_p2,private_p3);
}

}

```

On the sidenote : now that we are calling `optimize_module`, naming of JIT kernel is not accurate. e.g. For this example it is `quantizelinear_convert` but `convert` is optimized out.

src/targets/gpu/compile_gen.cpp
umangyadav: Done.

test/verify/run_verify.cpp
umangyadav: Done.

test/verify/run_verify.hpp
```diff
@@ -24,9 +24,11 @@
 #ifndef MIGRAPHX_GUARD_TEST_RUN_VERIFY_HPP
 #define MIGRAPHX_GUARD_TEST_RUN_VERIFY_HPP
 
+#include "verify_program.hpp"
```
pfultz2, line 27: We can forward declare the `program_info` class instead of including the header.

test/verify/run_verify.hpp
```diff
@@ -24,9 +24,11 @@
...
 #include <migraphx/program.hpp>
 #include <functional>
 #include <map>
+#include <unordered_map>
```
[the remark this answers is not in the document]
umangyadav: Removed, it is from earlier change.

test/verify/run_verify.hpp
umangyadav: Done.

shivadbhavsar (approved):

pfultz2 (approved):

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -77,7 +77,7 @@ struct quantizelinear
         visit_all(result, y_zero_point)([&](auto output, auto zero_pts) {
             visit_all(x, y_scale)([&](auto input, auto scales) {
                 using quant_type = typename decltype(output)::value_type;
-                auto min_value   = std::numeric_limits<quant_type>::min();
+                auto min_value   = std::numeric_limits<quant_type>::lowest();
```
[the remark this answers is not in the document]
umangyadav: I added test inside `ref/quantizelinear.cpp` for this.

CharlieL7 (approved):

PR #2657, opened 2024-01-18
Resize matcher

bpickrel:
resize matcher/pass optimizer code is required to convert any Resize op. with dynamic batch inputs into gpu-usable code.  This is a workaround for the fact that we don't yet have a GPU implementation of Resize.  See Issue 2658.


src/simplify_dyn_ops.cpp
```diff
@@ -26,10 +26,131 @@
...
+    auto matcher() const
+    {
+        std::cout << "calling resize matcher...........................\n";
+
+        // Will this work with either dynamic or static input 0?
+        return match::name("resize")(match::nargs(2), 
+                match::arg(0)(match::static_shape()),
+                match::arg(1)(match::is_constant()));  
+
+        // How to add the other case:  only 1 input arg
```
pfultz2, line 68: You can use the `any_of` matcher to list out the different case.

bpickrel: Comment removed because at this point, we don't need to match the 1-argument case in operational use.  The Onnx parser doesn't make a Resize op. when it only gets 1 argument, and Resize ops with 1 argument can only be created programmatically (i.e. by a unit test).

CharlieL7 (commented):
Is this ready to review? Looks like there are TODOs and codecoverage that needs udpates.

bpickrel:
> Is this ready to review? Looks like there are TODOs and codecoverage that needs udpates.

That's all test stuff.  It ought to be possible to review the functional code while I clean it up.

CharlieL7 (approved):

PR #2660, opened 2024-01-19
Fixed promote_literals_test by using additional if condition

aljanjic-amd:
After `out_ins->replace_argument(out_ins, ins, new_lit);`, structures `out_ins` and `ins` are modified which causes that `ins->outputs()` is returning an empty vector, so the for loop `for(auto out_ins : ins->outputs())` after first iteration of the for loop breaks. Therefore, the additional code serves as a safeguard for handling this zero values. @pfultz2 @apwojcik


pfultz2:
> After out_ins->replace_argument(out_ins, ins, new_lit);, structures out_ins and ins are modified which causes that ins->outputs()

We should not be modifying the vector while we are iterating, that is UB. We should probably assign `outputs` to a variable and then iterate the copy.

src/promote_literals.cpp
```diff
@@ -41,8 +41,9 @@ void promote_literals::apply(module_pass_manager& mpm) const
     {
         if(ins->name() == "@literal")
         {
-            auto new_lit = root_module->add_literal(ins->get_literal());
-            for(auto out_ins : ins->outputs())
+            auto new_lit    = root_module->add_literal(ins->get_literal());
+            auto ins_output = ins->outputs();
```
pfultz2, line 45: Can you make the variable plural?

aljanjic-amd: I created a new commit with the new variable name.

pfultz2 (approved):

CharlieL7 (approved):

CharlieL7 (commented):
Not sure I understand the bug. Why wasn't this caught by the `promote_and_ecs1` test?

CharlieL7:
Can we add a test checking for the bug?

aljanjic-amd:
> Can we add a test checking for the bug?

After this modification, all of the Test cases pass both on Windows and Linux, but without this modification, the error occurs in all of the Test cases on Windows.

CharlieL7:
> > Can we add a test checking for the bug?
> 
> Without this modification, these three test cases in `promote_literals_test.cpp` fail on Windows, but not on Linux. After this modification, both on Windows and Linux all tests pass. I'm not quite sure how could I create a test where we should check for the bug and handle that bug, because after this modification, there are no more bugs. @CharlieL7

I did some testing and found that it doesn't seem to fail on Linux with Clang. Probably because Clang handles the UB in some way that happens to work. I tried to enable libc++ debug checking but it doesn't seem to have worked or caught the UB.

CharlieL7 (approved):

PR #2662, opened 2024-01-19
Revert "Add the missing cmake file for parallel stl"

causten:
The introduction of 2618 caused a significant slow down in graph compile times.  Running resnet50 went from 50 seconds to 5 minutes between 538ffcce339ec76c06ffa703cf3368fddad21c95 and 733bd87bb547a35b68ad89ad338d0b1bede98503.


umangyadav:
AFAIU this change only affects the compile time of migraphx not the runtime. I don't know why you would see change in resnet50 compile time.

pfultz2 (changes requested):
We should not remove a file we are referencing in our cmake build scripts.

PR #2666, opened 2024-01-19
Temporarily disable parallel executors

causten:
The introduction of 2618 caused a significant slow down in graph compile times. Running resnet50 went from 50 seconds to 5 minutes between https://github.com/ROCm/AMDMIGraphX/commit/538ffcce339ec76c06ffa703cf3368fddad21c95 and https://github.com/ROCm/AMDMIGraphX/commit/733bd87bb547a35b68ad89ad338d0b1bede98503.


pfultz2 (approved):

umangyadav:
I don't understand the root cause, Parallelization should help in compile times of models. Is it because the TBB or C++ exectution.hpp itself is slow ?

PR #2669, opened 2024-01-19
Add Dot, Convolution, and Fused as default options in fuse_mlir

turneram:


pfultz2:
So this is slower on mi250 for bert base case. 1.8ms vs 2.1ms.

causten:
> So this is slower on mi250 for bert base case. 1.8ms vs 2.1ms.

It's actually faster then what we have today in develop, but the 1.8 was from 5.7.1.   This PR gets us back to 2.05ms.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -61,12 +61,19 @@ bool mlir_enabled()
...
+
 static bool is_requested(std::string_view option, bool fallback = false)
 {
     auto string_value = string_value_of(MIGRAPHX_MLIR_USE_SPECIFIC_OPS{}, "");
-    if(string_value.empty())
+    if(string_value.empty() and get_default_options().empty())
         return fallback;
-    const auto options = split_string(string_value, ',');
+    auto options = split_string(string_value, ',');
+    options.insert(options.begin(), get_default_options().begin(), get_default_options().end());
```
manupak, line 76: This should not be concatenating default options with user provided ones as this will mandate "dot", "convolution", "fused" always. I think what you want it use is either a non-empty MIGRAPHX_MLIR_USE_SPECIFIC_OPS{} *or* get_default_options().

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -61,12 +61,19 @@ bool mlir_enabled()
...
+{
+    static std::vector<std::string> default_options{"dot", "convolution", "fused"};
+    return default_options;
+}
+
 static bool is_requested(std::string_view option, bool fallback = false)
 {
     auto string_value = string_value_of(MIGRAPHX_MLIR_USE_SPECIFIC_OPS{}, "");
-    if(string_value.empty())
+    if(string_value.empty() and get_default_options().empty())
```
manupak, line 73: > get_default_options().empty()

would always be non-empty, why bother with this if ?

manupak (commented):
I think default case should only be used when MIGRAPHX_MLIR_USE_SPECIFIC_OPS{} is not specified. 

The current version of PR blocks us from experimenting with individual offload types.

src/targets/gpu/fuse_mlir.cpp
turneram: Missed your comment until now. Just updated to only use get_default_options when MIGRAPHX_MLIR_USE_SPECIFIC_OPS is empty.

pfultz2:
There is no way to use the fast mode for dot operators anymore.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -61,12 +61,18 @@ bool mlir_enabled()
...
+    return default_options;
+}
+
+static bool is_requested(std::string_view option)
 {
     auto string_value = string_value_of(MIGRAPHX_MLIR_USE_SPECIFIC_OPS{}, "");
     if(string_value.empty())
-        return fallback;
-    const auto options = split_string(string_value, ',');
+        return contains(get_default_options(), option);
```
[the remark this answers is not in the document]
turneram: Could we just add a "none" option?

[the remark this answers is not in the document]
pfultz2: Yea we could do that.

causten (approved):

PR #2671, opened 2024-01-22
Integrate hipblaslt calls in gemm op

mqhc2020:


pfultz2:
This should be a different operator. We can add a `compile_blas` similar to `compile_miopen` that can do the tuning for convolution through the [`compile`](https://github.com/ROCm/AMDMIGraphX/blob/e1f75de84e231de6673e31fcd9f4b5ffb7946c68/src/include/migraphx/operation.hpp#L519) function in operation interface. We just insert a placeholder operator during lowering, and then `compile_blas` will run the `compile` function to tune which operator is the fastest. Then this function can return the time and needed workspace.

The workspace should be inserted as an `allocate` instruction so the memory can be reused by non-hipblas kernels. So by returning the workspace we can insert the needed `allocate` instruction and then pass it as a parameter to the hipblas operator.

PR #2672, opened 2024-01-23
Use problem_cache for gemm tuning

pfultz2:


src/targets/gpu/problem_cache.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+void problem_cache::insert(const std::string& name, const value& problem, const value& solution)
+{
+    assert(not solution.is_null());
+    cache[create_key(name, problem)] = solution;
+}
+void problem_cache::mark(const std::string& name, const value& problem)
+{
+    cache.insert(std::make_pair(create_key(name, problem), value{}));
+}
```
TedThemistokleous, line 25: Minor nitpick. Add spaces between these functions

src/targets/gpu/gemm_impl.cpp
```diff
@@ -574,17 +574,45 @@ void gemm_compute(context& ctx,
...
+    ctx.get_problem_cache().insert(
+        "rocblas", gemm_problem(output_shape, input_shapes), solution_idx);
+}
+
+int32_t gemm_default_solution(context& ctx,
+                              const shape& output_shape,
+                              const std::vector<shape>& input_shapes)
+{
+    std::vector<shape> gemm_shapes{input_shapes[0], input_shapes[1], output_shape};
+    auto problem = to_value(gemm_shapes);
```
TedThemistokleous, line 597: This intended to be fed into gemm_problem as the shapes?

src/targets/gpu/gemm_impl.cpp
```diff
@@ -593,27 +621,36 @@ int32_t gemm_finalize(context& ctx,
...
     }
     else
     {
         // If a tuned solution index is already given, don't tune again but validate
         // in case the data was tuned with a different rocBLAS version
-        auto gemm_item = gemm_impl<float>(output_shape, input_shapes, alpha, beta, compute_fp32);
+        auto gemm_item = gemm_impl<T>(output_shape, input_shapes, alpha, beta, compute_fp32);
         solution_idx   = gemm_item.validate(ctx, input_shapes, solution_idx);
     }
+    gemm_save_solution(ctx, output_shape, input_shapes, solution_idx);
```
umangyadav, line 634: Saving should be after tuning inside `if` clause.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -132,6 +132,8 @@ struct rocblas_gemm
     void finalize(context& ctx, const shape& output_shape, const std::vector<shape>& input_shapes)
     {
 #ifdef MIGRAPHX_USE_ROCBLAS_TUNING_API
+        if(solution_idx == 0)
+            solution_idx = gemm_default_solution(ctx, output_shape, input_shapes);
```
umangyadav, line 136: From what i understand, This is where problem cache would be helpful. So that it doesn't tune same gemm configs again and again and instead would return saved solution idx from previous tuning ? Is that correct ? 

But then,

umangyadav (commented):
Looks good overall. Please fix licensing and then i can kick jenkins

src/targets/gpu/include/migraphx/gpu/gemm.hpp
pfultz2: Yes thats correct.

src/targets/gpu/gemm_impl.cpp
pfultz2: What do you mean?

src/targets/gpu/gemm_impl.cpp
pfultz2: Done

umangyadav (approved):

src/targets/gpu/gemm_impl.cpp
TedThemistokleous: the varaible problem isn't used but I assume captures shapes and would be used in place of gemm_problem(output, input)?

pfultz2: Ah thats a mistake. Good catch. I guess cppcheck caught it as well. The code just needs to be deleted, as I replaced it with a function since the problem needs to be computed in more than one place.

TedThemistokleous: All good. LGTM after fix.

TedThemistokleous (approved):
Fix format but otherwise makes sense.

umangyadav:
@pfultz2  agent model showing perf drop.  Looks unrelated. Can you please confirm ?

PR #2673, opened 2024-01-23
Rewrite softmax

pfultz2:
This rewrites softmax to lower set of operators so we can use our reduction fusion pipeline. It does this after flash attention fusion.


src/rewrite_reduce.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+namespace {
+struct find_softmax
+{
+    auto matcher() const { return match::name("softmax"); }
+
+    void apply(module& m, const match::matcher_result& r) const
```
umangyadav, line 14: Does this make softmax JIT kernel redundant ?

pfultz2: Yes it will.

pfultz2:
> Wouldn't it better to parse softmax directly as set of individual operators then ?

This would make matching flash attention more complicated, which is why I kept it. We could consider removing this in the future.

pfultz2:
So agent model improves quite a bit because we are able to fuse softmax with log+reduce_max that happens. The distilgpt2 model also improves because there is a `where` clause used for the softmax scale. We can probably rewrite though to use flash attention.

src/rewrite_reduce.cpp
pfultz2: This would make matching flash attention more complicated, which is why I kept it. We could consider removing this in the future.

Also, we might want to calibrate this operator in the future so we can get rid of the reduce_max.

umangyadav (approved):

TedThemistokleous (approved):

TedThemistokleous:
LGTM. heads up if CI doesn't fire off you may need to kick it for this. 

Fix conflicts before merge too.

causten:
@pfultz2 Can you investigate the accuracy drop

pfultz2:
The max diff goes from 3.9007 to 3.9173. So its a small difference. Most likely due to doing the reduction in fp16, but its so small of difference that probably the tolerance just needs to be updated in the accuracy checks.

Issue #2674, opened 2024-01-23
Dynamic shape with unspecified range


CharlieL7:
There are operators in MIGraphX that can produce dynamic shapes with an unspecified range such as `allocate` or `resize`. For these operators we have been using `dynamic_dimension`s with range `{0, max_int}` as a stand-in for an unknown range. This leads to a clarity issue and an issue with the `shape.element_space()` calculation for rank > 1.
* We need a way to specify that a dimension is fully unknown without a specific minimum or maximum.
* This will need logic changes throughout the dynamic shape handling.

bpickrel:
My first Verify test for PR #2657 failed because a program containing an indefinite-sized dynamic shape caused the GPU to run out of memory.  Even relatively modest ranges of `{0,64}, {0,64}, {0, 4096}, {0, 4096}`  (which multiplies out to 16G; not known how many bytes this equates to) lead to a failure to load the program on aus-navi3x-02.  This tells us the current unspecified range logic is not just unclear but unworkable.

pfultz2:
Why cant `{0, max_int}` mean an unspecified range? And what about the cases where the min is known but the max is not known? We are never going to use `max_int` as a real upper limit so we can interpret it as meaning the upper range is not known. 

> This leads to a clarity issue and an issue with the shape.element_space() calculation for rank > 1.

And for that case we should just return `max_int`.

pfultz2:
> My first Verify test for PR https://github.com/ROCm/AMDMIGraphX/pull/2657 failed because a program containing an indefinite-sized dynamic shape caused the GPU to run out of memory.

I dont understand why you are trying to run dynamic shapes on the gpu. That isn't supported.

Even more so this seems irrelevant to this issue. Even if we can represent an unknown dimension we have no way to allocate a buffer for it because we dont know its size.

CharlieL7:
Clipping is a lot less work, so I'm going to make `shape.element_space()` do that for now.

bpickrel:
> > My first Verify test for PR #2657 failed because a program containing an indefinite-sized dynamic shape caused the GPU to run out of memory.
> 
> I dont understand why you are trying to run dynamic shapes on the gpu. That isn't supported.

For starters, that is still a draft PR so it makes sense to try a test to see what happens.  What I learned is that 

1.  the test does do the Resize pass substitution and run the results on the GPU; 
2.  the test turned up a memory error and so is proven useful.

[...]

PR #2675, opened 2024-01-23
Clip to max_int on overflow of element_space for dynamic shape

CharlieL7:
* Found an error while going through code with Brian, this will make it so we throw with an descriptive error.
* Made an issue #2674 that outlines a better fix of having some way to create a dynamic shape with unknown range.

### DIFF
--- a/src/shape.cpp
+++ b/src/shape.cpp
@@ -127,9 +127,10 @@ struct shape_impl
 
             return std::accumulate(
                 maxes.begin(), maxes.end(), std::size_t{1}, [&](std::size_t x, std::size_t y) {
+                    // overflow check and clip
                     if(x != 0 and y > max_val / x)
                     {
-                        MIGRAPHX_THROW("SHAPE: possible overflow of dynamic element_space");
+                        return max_val;
                     }
                     return x * y;
                 });
--- a/src/shape.cpp
+++ b/src/shape.cpp
@@ -123,7 +123,16 @@ struct shape_impl
         if(not m_dyn_dims.empty())
         {
             auto maxes = max_lens();
-            return std::accumulate(maxes.begin(), maxes.end(), std::size_t{1}, std::multiplies<>());
+            std::size_t max_val = std::numeric_limits<std::size_t>::max();
+
+            return std::accumulate(
+                maxes.begin(), maxes.end(), std::size_t{1}, [&](std::size_t x, std::size_t y) {
+                    if(x != 0 and y > max_val / x)
+                    {
+                        MIGRAPHX_THROW("SHAPE: possible overflow of dynamic element_space");
+                    }
+                    return x * y;
+                });
         }
 
         assert(m_lens.size() == m_strides.size());
--- a/test/shape_test.cpp
+++ b/test/shape_test.cpp
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
- * Copyright (c) 2015-2023 Advanced Micro Devices, Inc. All rights reserved.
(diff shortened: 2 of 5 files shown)


pfultz2:
Would it be better to clip rather than throw an error?

bpickrel:
I don't agree with this PR unless you apply it everywhere.  If we're going to check for 64-bit integer overflows in this case, we ought to check for them consistently in many other places in the code base.  I see that std::accumulate is used in dozens of different places.

As for the specific problem you're addressing here, I don't think it will be a pressing issue once we address the underlying issue of dynamic element spaces getting too large.  The practical limit on their size is--as far as I can tell--the amount of memory that can be allocated on the GPU, which is hardware-dependent but always far less than 2^64.  It's not clear to me that all of the memory for a dynamic element space really needs to be allocated, but apparently that's what it's trying to do right now.

I say this because I've been doing some trial and error checking of the workable limits for that Verify test I wrote for dynamic Resize op inputs.  It failed at 64x64x2kx2k.  The total element space blows up very quickly for any tensor of rank higher than 2 dimensions.

test/shape_test.cpp
```diff
@@ -160,6 +160,34 @@ TEST_CASE(test_shape_dynamic_compares)
...
+    migraphx::shape s{migraphx::shape::float_type, {{0, max_val}, {0, max_val}}};
+    EXPECT(s.element_space() == max_val);
+}
+
+TEST_CASE(dynamic_shape_element_space_overflow1)
+{
+    std::size_t max_val   = std::numeric_limits<std::size_t>::max();
+    std::size_t large_val = max_val / 10;
+    migraphx::shape s{migraphx::shape::float_type, {{0, large_val}, {0, large_val}}};
+    EXPECT(s.element_space() == max_val);
```
TedThemistokleous, line 181: If we overflow do we not expect to error out/ signal an overflow as a warning/error?

src/shape.cpp
```diff
@@ -123,7 +123,17 @@ struct shape_impl
...
             auto maxes = max_lens();
-            return std::accumulate(maxes.begin(), maxes.end(), std::size_t{1}, std::multiplies<>());
+            std::size_t max_val = std::numeric_limits<std::size_t>::max();
+
+            return std::accumulate(
+                maxes.begin(), maxes.end(), std::size_t{1}, [&](std::size_t x, std::size_t y) {
+                    // overflow check and clip
+                    if(x != 0 and y > max_val / x)
+                    {
+                        return max_val;
```
TedThemistokleous, line 133: add a warning about clipping that we can turn on in debug builds

TedThemistokleous (commented):
I think its better to clip but then add in a warning/error during debug builds to show that's happened. This sounds like a good topic to go over for Monday

src/shape.cpp
```diff
@@ -123,7 +123,17 @@ struct shape_impl
         if(not m_dyn_dims.empty())
         {
             auto maxes = max_lens();
-            return std::accumulate(maxes.begin(), maxes.end(), std::size_t{1}, std::multiplies<>());
+            std::size_t max_val = std::numeric_limits<std::size_t>::max();
```
pfultz2, line 126: This can be declared as a constant.

pfultz2 (approved):

pfultz2:
> As an example there is this MIGRAPHX_ASSERT in src/targets/gpu/kernels/include/migraphx/kernels/gemm_batcher.hpp

That is using a static shape not a dynamic shape.

TedThemistokleous (approved):

src/shape.cpp
umangyadav: Warning would be good to have

PR #2679, opened 2024-01-24
Add support for Hardmax operator

nives-vukovic:
Add support for Hardmax operator.

Add corresponding reference, onnx, and verify tests.

Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/97


bpickrel (approved):
Looked over tests, looks correct.  Approved, but please add a test for the one line not covered in Codecov.

nives-vukovic:
@bpickrel As far as I can see updated Codecov report does not show any missing lines in implemented and otherwise affected files.

src/onnx/parse_hardmax.cpp
```diff
@@ -38,35 +38,57 @@ struct parse_hardmax : op_parser<parse_hardmax>
...
 
         if(contains(info.attributes, "axis"))
         {
             axis = parser.parse_value(info.attributes.at("axis")).at<int>();
         }
 
-        auto indices = info.add_instruction(make_op("argmax", {{"axis", axis}}), args);
+        if(parser.opset_version < 13)
+        {
+            // input is coerced into a 2D matrix of size NxD
```
bpickrel, line 62: This effectively means that the output will contain a single 1, at the location of the global max value--is that correct?

bpickrel:
> @bpickrel As far as I can see updated Codecov report does not show any missing lines in implemented and otherwise affected files.

I see yesterday's commit takes care of it.  I did add one question about a potential divide by 0.

src/onnx/parse_hardmax.cpp
nives-vukovic: It depends on the axis value, in the older version if the input shape is e.g. [3,2,2] and axis is 0, it will coerce it to [1,12] in which case we will indeed have only single 1. If axis is 1 (defult), it will coerce it to [3,4] in which case we will have 3 ones for each of the batches. As described https://onnx.ai/onnx/operators/onnx__Hardmax.html#hardmax-11, and implementation in python backend tests: https://github.com/onnx/onnx/blob/rel-1.7.0/onnx/backend/test/case/node/hardmax.py.
As for division by zero, I guess it could happen if one of the input lens is zero. Solution can be to calculate d in the same way as n is calculated:
size_t d = 1;
for(int i = axis; i < input_lens.size(); i++)
{
  d *= input_lens[i];
}

nives-vukovic: @bpickrel  As for division by zero, I guess it could happen if one of the input lens is zero. After some analysis, I believe the best way would be to handle it as a special case scenario as in onnxruntime, as in that case number of elements() for the input shape would be zero. Therefore returning an output shape with same dimensions filled with zeroes would be the correct result for both versions.

nives-vukovic:
@bpickrel 

> > @bpickrel As far as I can see updated Codecov report does not show any missing lines in implemented and otherwise affected files.
> 
> I see yesterday's commit takes care of it. I did add one question about a potential divide by 0.

As for division by zero, I guess it could happen if one of the input lens is zero. If that is a possible input, after some analysis, I believe the best way would be to handle it as a special case scenario as in onnxruntime, as in that case number of elements() for the input shape would be zero. Therefore returning an output shape with same dimensions filled with zeroes would be the correct result for both versions.

CharlieL7 (approved):

pfultz2 (approved):

PR #2687, opened 2024-01-25
Add optional fp16 rmsnorm conversion pass to fix fp16 accuracy

### LINKED ISSUE #2556: FP16 inaccuracy with Llama-2
FP16 version of llama has strange output.

I could not managed to pinpoint the exac cause yet, but here are my findings:

The following logs contains 5 steps of generation:

`MIGRAPHX_TRACE_EVAL=2 python3 txtgen.py -p "Where is Szeged?" -l -s 256` with and without `--fp16`.
And the timesteps were limited to `5`.

[llama2-fp16-256-gpu-t5.log](https://github.com/ROCm/AMDMIGraphX/files/13686883/llama2-fp16-256-gpu-t5.log)
[llama2-fp32-256-gpu-t5.log](https://github.com/ROCm/AMDMIGraphX/files/13686885/llama2-fp32-256-gpu-t5.log)

Note: it can be reproduced with [this](https://github.com/ROCm/AMDMIGraphX/tree/llama_example/examples/transformers/python_llama2#console-application)

The results with FP32:
```log
1) token=13     with prob=18.465343475341797
2) token=13     with prob=12.884403228759766
3) token=29903  with prob=20.014484405517578
4) token=28530  with prob=22.900659561157227
5) token=287    with prob=26.00668716430664
```

The results with FP16:
```log
1) token=13     with prob=9.03125
2) token=0      with prob=0.0
3) token=18627  with prob=6.38671875
4) token=29901  with prob=9.8359375
5) token=910    with prob=9.671875
```

Note: `13` is a linebreak.

The problem is at the second step, where the full output for that timestep is 0. After that, the model starts to generate nonsense.

attila-dusnoki-htec:
Fixes https://github.com/ROCm/AMDMIGraphX/issues/2556
RMSNorm is used in LLMs like Llama2. Currently the fp16 version can overflow during normalization. This change try to addess it byconvert normalization to fp32.

### DIFF
--- a/docs/dev/env_vars.rst
+++ b/docs/dev/env_vars.rst
@@ -131,6 +131,10 @@ Enables additional opportunities to use MLIR that may improve performance.
 Set to "1", "enable", "enabled", "yes", or "true" to use.
 Use ``hip_copy_to_gpu`` with a new ``literal`` instruction rather than use ``hip_copy_literal{}``.
 
+.. envvar:: MIGRAPHX_ENABLE_RMSNORM_FP16_REWRITE
+Set to "1", "enable", "enabled", "yes", or "true" to use.
+Enables RMSNorm to be converted to fp32 precision to handle large numbers.
+
 Compilation traces
 ----------------------
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -96,6 +96,7 @@ add_library(migraphx
     rewrite_gelu.cpp
     rewrite_pooling.cpp
     rewrite_quantization.cpp
+    rewrite_rmsnorm.cpp
     rewrite_rnn.cpp
     schedule.cpp
     serialize.cpp
--- a/src/driver/passes.cpp
+++ b/src/driver/passes.cpp
@@ -42,6 +42,7 @@
 #include <migraphx/promote_literals.hpp>
 #include <migraphx/propagate_constant.hpp>
 #include <migraphx/rewrite_gelu.hpp>
+#include <migraphx/rewrite_rmsnorm.hpp>
 #include <migraphx/rewrite_pooling.hpp>
 #include <migraphx/rewrite_quantization.hpp>
 #include <migraphx/rewrite_rnn.hpp>
@@ -80,6 +81,7 @@ std::unordered_map<std::string, pass> create_passes_lookup()
         promote_literals{},
         propagate_constant{},
         rewrite_gelu{},
+        rewrite_rmsnorm{},
         rewrite_pooling{},
         rewrite_quantization{},
         rewrite_rnn{},
--- a/src/include/migraphx/match/rmsnorm.hpp
+++ b/src/include/migraphx/match/rmsnorm.hpp
@@ -1,0 +1,100 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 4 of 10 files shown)


pfultz2:
I dont think this is exactly the right approach. I would prefer we update the `rewrite_reduce` pass to lower `reduce_mean` to `reduce_sum`(for all cases). And then we can have a pass to help increase precision by finding patterns like `x^2 / n` and rewrite them to `(x / sqrt(n))^2`.

attila-dusnoki-htec:
@pfultz2  I agree that a general approach for the rewrite could help with more cases. I was just not sure that it should be enabled for all occurences. Also, we need to be careful with it to not intruduce any regressions, since other matcher can depend on that specific structure, so they will need to be extended as well probably.

> I dont think this is exactly the right approach. I would prefer we update the `rewrite_reduce` pass to lower `reduce_mean` to `reduce_sum`(for all cases).

I don't see any `rewrite_reduce` pass. There are some rewrites scattered throughout the code (e.g. [isntance_norm](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/onnx/parse_instancenorm.cpp#L96-L98), [compile_gen](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/targets/gpu/compile_gen.cpp#L219-L233), [layernorm](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/targets/gpu/kernels/include/migraphx/kernels/layernorm.hpp#L64).
The idea would be to have a dedicated `rewrite_reduce` instead of the specific `rewrite_rmsnorm`?

---

Also, this math rewrite only a partial solution, since larger numbers still can cause problem. I would suggest having a dedicated rmsnorm fp16-to-fp32 convert optional pass for those cases.

pfultz2:
> since other matcher can depend on that specific structure, so they will need to be extended as well probably.

This same rewriting is needed for layernorm, Right now we use a custom kernel that does it, but I want to deprecate to use our reduction fusion pipeline though.

Also, this pass can run after prefuse_ops so it shouldn't affect the current layernorm fusion. 

> I don't see any rewrite_reduce pass.

The pass is added in #2673, which hasnt been merged yet. It runs after prefuse ops so the flash attention can still fuse with a softmax.

pfultz2:
> Also, this math rewrite only a partial solution, since larger numbers still can cause problem. I would suggest having a dedicated rmsnorm fp16-to-fp32 convert optional pass for those cases.

I think with fast math we can use fp16 but when its disabled we should promote it to fp32.

attila-dusnoki-htec:
@pfultz2 Updated to the `rewrite_reduce` pass.

> I think with fast math we can use fp16 but when its disabled we should promote it to fp32.

The fp16->fp32 is still behind a flag. I have to check the fast math part first.

> And then we can have a pass to help increase precision by finding patterns like x^2 / n and rewrite them to (x / sqrt(n))^2

Also, this does not have it yet.

src/rewrite_reduce.cpp
```diff
@@ -0,0 +1,92 @@
...
+    {
+        auto reduce_mean = r.result;
+        auto op          = reduce_mean->get_operator().to_value();
+        auto axes        = op["axes"].to_vector<std::int64_t>();
+        auto input       = reduce_mean->inputs().front();
+
+        auto dims = input->get_shape().lens();
+        double n  = std::accumulate(
+            dims.begin() + 2, dims.end(), 1, [&](const auto& i, const auto& j) { return i * j; });
+        auto n_literal = m.add_literal(literal{{input->get_shape().type(), {1}}, {1.0 / n}});
```
pfultz2, line 71: This wont work for integers. This should be `n` and not `1/n` to make it simpler for the integer path(simplify_algebra can rewrite to the reciprocal).

src/rewrite_reduce.cpp
```diff
@@ -0,0 +1,92 @@
...
+        m.replace_instruction(reduce_mean, reduce_sum);
+    }
+};
+
+} // namespace
+
+void rewrite_reduce::apply(module& m) const
+{
+    match::find_matches(m, find_softmax{});
+    match::find_matches(m, rewrite_reduce_mean{});
```
pfultz2, line 88: Dont call another `find_matches` function.

src/rewrite_reduce.cpp
```diff
@@ -0,0 +1,92 @@
...
+        auto sub  = m.insert_instruction(ins, make_op("sub"), input, maxb);
+        auto exp  = m.insert_instruction(ins, make_op("exp"), sub);
+        auto sum  = m.insert_instruction(ins, make_op("reduce_sum", {{"axes", {axis}}}), exp);
+        auto sumb = m.insert_instruction(
+            ins, make_op("multibroadcast", {{"out_lens", input->get_shape().lens()}}), sum);
+        m.replace_instruction(ins, make_op("div"), exp, sumb);
+    }
+};
+
+struct rewrite_reduce_mean
```
pfultz2, line 57: This should be done in a seperate PR.

attila-dusnoki-htec:
@pfultz2 Sorry about `reduce_rewrite`, I misunderstood that part. I removed that and will move it to a separate PR.

attila-dusnoki-htec:
Converted it to draft. It will be updated after https://github.com/ROCm/AMDMIGraphX/issues/2710 added to use this logic instead.

PR #2690, opened 2024-01-26
Fix compute_shape of `mlir_op`

umangyadav:
Fixes compilation of `yolov4.onnx --int8`.

### DIFF
--- a/src/targets/gpu/fuse_mlir.cpp
+++ b/src/targets/gpu/fuse_mlir.cpp
@@ -120,7 +120,21 @@ struct mlir_op
                 mod_ins_shapes[ins] =
                     adjusted_mod_param_shapes[any_cast<builtin::param>(ins->get_operator())
                                                   .parameter];
-                assert(ins->get_shape().type() == mod_ins_shapes[ins].type());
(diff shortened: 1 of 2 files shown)


pfultz2 (approved):

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -104,25 +104,42 @@ struct mlir_op
...
         {
-            if(ins->name() == "@literal" or ins->name() == "@param")
+            if(ins->name() == "@param")
             {
-                ins_shapes[ins] = ins->get_shape();
-                continue;
+                mod_ins_shapes[ins] =
+                    adjusted_mod_param_shapes[any_cast<builtin::param>(ins->get_operator())
+                                                  .parameter];
+                assert(ins->get_shape().type() == mod_ins_shapes[ins].type());
```
pfultz2, line 123: This should be a throw or hard error. Also we should probably check the `.lens()` match as well.

umangyadav: Done.

krzysz00 (approved):
Seems fine to me!

PR #2693, opened 2024-01-29
Handle TopK fp16 quantization

attila-dusnoki-htec:
TopK op can't be quantized with the generic logic, because it depends on the get_tuple_elem op to access the output. This change handle this case for fp16 quantization.

### DIFF
--- a/test/verify/test_fp32_fp16_topk.cpp
+++ b/test/verify/test_fp32_fp16_topk.cpp
@@ -1,0 +1,49 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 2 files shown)


pfultz2:
I dont really like hardcoding operators int the quantizer. Cant we just handle tuple types generically?

pfultz2:
> Sure, I can take a look how convert and shape work more closely with tuples.

Do we need to modify the `convert` operator? This operator is generically written for all pointwise operators.

Cant we just modify the graph to convert each tuple element?

attila-dusnoki-htec:
The idea here is to change input of topk, so `input -> topk` becomes `input -> convert -> topk`, meaning `[float, int]` -> `[half, int]`.
Usually you convert it back to the original dtype, but in this case the dtype is `tuple`.
Then when the `get_tuple_elem` comes in, you also can't convert the input, since the dtype is `tuple`, and still points to the `[float, int]` version. And I'm not sure how to connect the 2 ops correctly, if not in one step during topk.

So in order to work, it does not have to be `if(ins->name() == "topk")` i think, we could check for `tuple` type instead. That might not be considered hardcoding?

Another possible workaround to avoid adding the proposed change to `quantize_fp16`, is to skip the `topk` (or `tuple`) type in general. That will compile and will leave the `topk` in float, which would resolve the model compile issue.
And maybe add another quantize or rewrite pass to handle the specific tuple case.

pfultz2:
> Then when the get_tuple_elem comes in, you also can't convert the input, since the dtype is tuple, and still points to the [float, int] version. And I'm not sure how to connect the 2 ops correctly, if not in one step during topk.

I assume you would get the outputs of the instruction that are `get_tuple_elem` and add conversions back to the original type. 

> Another possible workaround to avoid adding the proposed change to quantize_fp16, is to skip the topk (or tuple) type in general. That will compile and will leave the topk in float, which would resolve the model compile issue.
And maybe add another quantize or rewrite pass to handle the specific tuple case.

That is a quick to get the model running, and then we can look how we can quantize it exactly later on.

attila-dusnoki-htec:
> I assume you would get the outputs of the instruction that are `get_tuple_elem` and add conversions back to the original type.

That is exactly what happens in the PR.

> Cant we just handle tuple types generically?

I replaced any topk mentions with generic tuple type.

> That is a quick to get the model running, and then we can look how we can quantize it exactly later on.

Please take another look. If this still not the correct way, i will change the logic to just ignore topk to at least get the model compile.

test/verify/test_fp32_fp16_topk.cpp
```diff
@@ -0,0 +1,49 @@
...
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        migraphx::shape s{migraphx::shape::float_type, {3, 5}};
+        auto data = mm->add_parameter("data", s);
+        auto r    = mm->add_instruction(
+            migraphx::make_op("topk", {{"axis", -2}, {"k", 3}, {"largest", 0}}), data);
+        auto r0 = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), r);
+        auto r1 = mm->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 1}}), r);
+        mm->add_return({r0, r1});
+        migraphx::quantize_fp16(p, {"topk"});
```
umangyadav, line 45: Correct place for testing this would be inside `test/quantization.cpp` instead of a verify test.

umangyadav (approved):

pfultz2 (approved):

Issue #2695, opened 2024-01-29
Add simplify_qlinear_ops pass


umangyadav:
### Problem Description

DeQuantizeLinear and QuantizeLinear Ops takes scales and zero points as the parameters. 

Zero point is an optional parameter. 

This issue about task of adding `simplify_qlinear_ops` pass which should do two things for now. 

1. Many times when doing quantization using `migraphx-driver` which doesn't use real data, "zero point" comes out to be literal zero value.  This extra zero point parameter can be cleaned up completely from graph before calling into `compile_ops` pass. It could be problematic during JIT compilaton. 
2.  If scales comes to te literal value `1` then qlinear ops can further be optimized out to just be "convert" operation (assuming zero point is zero as well). 

### Operating System

22.04

### CPU

Any

### GPU

AMD Instinct MI300

### ROCm Version

ROCm 6.0.0

umangyadav:
- quantizelinear op requires zero point parameter to determine output shape. Therefore even if zero point's value is zero, it can't be eliminated. 
- If we want to eliminate zero point then need to add "convert" after quantizelinear. But then quantizelinear would be computed as shape of "uint8" and then additional convert may introduce unnecessary bugs and complexity. 
e.g. QuantizeLinear op is : `convert(clip(nearbyint(x / scale) + zp))` Clip operation would use max and min value of uint8 v/s original data type. Let's say original output type was FP8E4M3FNUZ then it would have min max +/-240.0f  v/s uint8 would use 0, 255. 

- Other option is to decompose quantizelinear which is what rewrite_quantization does currently. But that is deferred until JIT compilation to allow MLIR to fuse ops. 

For the DeQuantizeLinear, zero point can be eliminated without any of above issues.

umangyadav:
solved with #2728

PR #2696, opened 2024-01-30
use lower tolerance for zero detection in simplify_algebra

umangyadav:
Fixes compilation of `yolov4.onnx --fp16 --int8` 

As a follow up work created #2695

### DIFF
--- a/src/include/migraphx/matcher.hpp
+++ b/src/include/migraphx/matcher.hpp
@@ -886,34 +886,21 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
         l.visit([&](auto v) {
             // cast to the literal's data type before comparing
             using type = typename decltype(v)::value_type;
-            if constexpr(std::is_integral<type>{})
+            auto tolerance = atol_mult + rtol_mult * std::fabs(x);
+            if(migraphx::float_equal(tolerance, 0) or std::is_integral<type>{})
             {
-                if(std::all_of(
-                       v.begin(), v.end(), [&](auto val) { return val == static_cast<type>(x); }))
-                {
+                if(std::all_of(v.begin(), v.end(), [&](auto val) {
+                       return migraphx::float_equal(val, static_cast<type>(x));
+                   }))
                     b = true;
-                }
             }
             else
             {
                 auto eps = std::numeric_limits<type>::epsilon();
-                auto tolerance = atol_mult + rtol_mult * std::fabs(x);
-                if(migraphx::float_equal(tolerance, 0.0))
-                {
-                    // do exact match using float_equal if tolerance is zero
-                    if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                           return migraphx::float_equal(val, static_cast<type>(x));
-                       }))
-                    {
-                        b = true;
-                    }
-                }
-                else if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                            return std::fabs(val - static_cast<type>(x)) < (eps * tolerance);
-                        }))
-                {
+                if(std::all_of(v.begin(), v.end(), [&](auto val) {
+                       return std::fabs(val - static_cast<type>(x)) < (eps * tolerance);
+                   }))
                     b = true;
-                }
             }
         });
         return b;
--- a/src/include/migraphx/matcher.hpp
+++ b/src/include/migraphx/matcher.hpp
@@ -873,7 +873,7 @@ inline auto literal_value_checker(F f)
 
 /**
  * Uses integer multiples of the corresponding floating point epsilon and
- * compares with abs(y - x) < eps * (atol_mult + rtol_mult * abs(y)).
+ * compares with abs(y - x) < eps * (atol_mult + rtol_mult * abs(x)).
  * atol_mult controls the absolute tolerance.
  * rtol_mult controls the relative tolerance.
  * Uses no tolerance for integral types.
@@ -897,19 +897,19 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
             else
             {
                 auto eps = std::numeric_limits<type>::epsilon();
-                if(atol_mult == 0 and rtol_mult == 0)
+                auto tolerance = atol_mult + rtol_mult * std::fabs(x);
+                if(migraphx::float_equal(tolerance, 0.0))
                 {
-                    // do exact match using float_equal if atol and rtol both are set to zero
+                    // do exact match using float_equal if tolerance is zero
                     if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                           return migraphx::float_equal(std::fabs(val), static_cast<type>(x));
+                           return migraphx::float_equal(val, static_cast<type>(x));
                        }))
                     {
                         b = true;
                     }
                 }
                 else if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                            return std::fabs(val - static_cast<type>(x)) <
-                                   eps * (atol_mult + rtol_mult * std::fabs(x));
+                            return std::fabs(val - static_cast<type>(x)) < (eps * tolerance);
                         }))
                 {
                     b = true;
--- a/src/include/migraphx/matcher.hpp
+++ b/src/include/migraphx/matcher.hpp
@@ -24,6 +24,7 @@
 #ifndef MIGRAPHX_GUARD_RTGLIB_MATCHER_HPP
 #define MIGRAPHX_GUARD_RTGLIB_MATCHER_HPP
 
+#include <migraphx/float_equal.hpp>
 #include <migraphx/functional.hpp>
 #include <migraphx/ranges.hpp>
 #include <migraphx/instruction.hpp>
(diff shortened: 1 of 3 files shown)


src/include/migraphx/matcher.hpp
```diff
@@ -878,7 +878,7 @@ inline auto literal_value_checker(F f)
  * Uses no tolerance for integral types.
  */
 template <class T>
-inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 10)
+inline auto has_value(T x, float atol_mult = 10, float rtol_mult = 10)
```
pfultz2, line 881: I think its better to have that as a integer to make it clearer its a multiple of epsilon and not epsilon. You can set `atol_mul` to 0 to handle small numbers.

src/simplify_algebra.cpp
```diff
@@ -1237,11 +1237,11 @@ struct find_neg_unit_ops
...
         auto mul_neg_1 = match::name("mul")(
-            match::either_arg(0, 1)(match::has_value(-1.0f), match::any().bind("x")));
+            match::either_arg(0, 1)(match::has_value(-1.0f, 1, 0), match::any().bind("x")));
         auto div_neg_1 =
-            match::name("div")(match::args(match::any().bind("x"), match::has_value(-1.0f)));
-        auto sub_0 =
-            match::name("sub")(match::args(match::has_value(0.0f), match::any().bind("x")));
+            match::name("div")(match::args(match::any().bind("x"), match::has_value(-1.0f, 1, 0)));
+        auto sub_0 = match::name("sub")(
+            match::args(match::has_value(0.0f, 1e-3, 0), match::any().bind("x")));
```
pfultz2, line 1244: `atol_mul` should be 0. Also there is no need to set rtol.

src/include/migraphx/matcher.hpp
```diff
@@ -895,9 +895,9 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
             }
             else
             {
-                auto eps = std::numeric_limits<type>::epsilon();
+                float eps = std::numeric_limits<type>::epsilon();
                 if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                       return std::fabs(val - static_cast<type>(x)) <
+                       return std::fabs(static_cast<float>(val) - static_cast<float>(x)) <
                               eps * (atol_mult + rtol_mult * std::fabs(val));
```
CharlieL7, line 901: Static casting to a float is problematic. I specifically had it cast to the type of the literal because of how fp16 cannot represent all of the values fp32 can. This leads to problems with constants that are already quantized to fp16 in pre-quantized models.

src/include/migraphx/matcher.hpp
```diff
@@ -895,9 +895,9 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
             }
             else
             {
-                auto eps = std::numeric_limits<type>::epsilon();
+                float eps = std::numeric_limits<type>::epsilon();
```
pfultz2, line 898: This should be double if you dont use `auto`.

src/include/migraphx/matcher.hpp
pfultz2: Casting to float is problematic because it cant represent values in `double`. 

> Ex: model quantized a value x to fp16 which becomes a value y. If you upcast value y to fp32, it will stay as value y when comparing to the target value.

I dont really understand what the issue is here.

CharlieL7: https://github.com/ROCm/AMDMIGraphX/blob/432e4e526a9fe9bf73af57b8f2148c9a0f0e20d0/test/rewrite_gelu_test.cpp#L127

There's an example of the `fit_const` value being slightly off in the prequantized model. It's exactly the same value in fp16 however.

src/include/migraphx/matcher.hpp
umangyadav: Reverted

src/include/migraphx/matcher.hpp
umangyadav: Reverted change

src/simplify_algebra.cpp
umangyadav: That works too.

src/include/migraphx/matcher.hpp
umangyadav: I  was interpreting `eps` as scaling factor for different dtypes.  What you are suggesting is interpreting atol and rtol as multiples/scaling factors.  I don't have problem with either.

pfultz2 (approved):

CharlieL7 (approved):

src/include/migraphx/matcher.hpp
umangyadav: > Ex: model quantized a value x to fp16 which becomes a value y. If you upcast value y to fp32, it will stay as value y when comparing to the target value. 

I don't understand.  It is doing comparing like `a < b` where `b` would i suppose be 64-bit floating point value because atol and rtol are std::size_t 

now `a` is `val -x`. It doesn't matter because it is upcasting.

CharlieL7: > now `a` is `val -x`. It doesn't matter because it is upcasting.
It can matter. In the example I linked to this number in fp16: `0.044708251953125` is already the closest representation of `0.044715f`, which is what we're searching for. It's not "close enough" it's exactly the right representation.

src/include/migraphx/matcher.hpp
```diff
@@ -896,10 +897,20 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
...
                 auto eps = std::numeric_limits<type>::epsilon();
-                if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                       return std::fabs(val - static_cast<type>(x)) <
-                              eps * (atol_mult + rtol_mult * std::fabs(val));
-                   }))
+                if(atol_mult == 0 and rtol_mult == 0)
+                {
+                    // do exact match using float_equal if atol and rtol both are set to zero
+                    if(std::all_of(v.begin(), v.end(), [&](auto val) {
+                           return migraphx::float_equal(std::fabs(val), static_cast<type>(x));
```
lakhinderwalia, line 904: Question: wouldn't it be logically correct, and more efficient, to do float_equal() using val -- instead of using std::fabs(val) ?

src/include/migraphx/matcher.hpp
```diff
@@ -896,10 +897,20 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
             else
             {
                 auto eps = std::numeric_limits<type>::epsilon();
-                if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                       return std::fabs(val - static_cast<type>(x)) <
-                              eps * (atol_mult + rtol_mult * std::fabs(val));
-                   }))
+                if(atol_mult == 0 and rtol_mult == 0)
```
pfultz2, line 900: This should check that `atol_mult + rtol_mult * std::fabs(x)` is equal to 0. This can be computed as a variable so it doesnt need to be recomputed everytime in the `all_of` loop.

src/include/migraphx/matcher.hpp
pfultz2: Can you add a test for `has_value(5.0, 0, 0)` that checks against `-5.0`(which should fail to match)?

umangyadav: > Can you add a test for has_value(5.0, 0, 0) that checks against -5.0(which should fail to match)? 

added

> ) using val -- instead of using std::fabs(val) ? 

Yes fixed that.

src/include/migraphx/matcher.hpp
umangyadav: Done

CharlieL7 (approved):

pfultz2 (approved):

src/include/migraphx/matcher.hpp
```diff
@@ -896,10 +897,20 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
             else
             {
                 auto eps = std::numeric_limits<type>::epsilon();
-                if(std::all_of(v.begin(), v.end(), [&](auto val) {
-                       return std::fabs(val - static_cast<type>(x)) <
-                              eps * (atol_mult + rtol_mult * std::fabs(val));
-                   }))
+                auto tolerance = atol_mult + rtol_mult * std::fabs(x);
+                if(migraphx::float_equal(tolerance, 0.0))
```
lakhinderwalia, line 901: Confusing logic. `atol_mult== 0 && rtol_mult == 0` is all that should be checked here on line 901 -- and tolerance (or its equivalent) shouldn't be checked for 0.0 here, as that gives the impression that somehow atol_mult == -1 * rtol_mult * x is a valid use_case to execute this if clause.

Nitpicking: If you want to optimize and save the result of variable `tolerance`, please do by creating an else clause on line 911. Also, in that case, move variable `eps` into that else clause -- better combine tolerance and eps.

src/include/migraphx/matcher.hpp
```diff
@@ -896,10 +897,20 @@ inline auto has_value(T x, std::size_t atol_mult = 10, std::size_t rtol_mult = 1
...
+                {
+                    // do exact match using float_equal if tolerance is zero
+                    if(std::all_of(v.begin(), v.end(), [&](auto val) {
+                           return migraphx::float_equal(val, static_cast<type>(x));
```
lakhinderwalia, line 905: `migraphx::float_equal` is also an incorrect implementation. That should be fixed (either in this PR, or separately..) 

float_equal_fn::apply() should be rewritten.
It is:    
```
static bool apply(T x, T y)
    {
        return std::isfinite(x) and std::isfinite(y) and
               std::nextafter(x, std::numeric_limits<T>::lowest()) <= y and
               std::nextafter(x, std::numeric_limits<T>::max()) >= y;
    }
```
And it should be:
```
   static bool apply(T x, T y)
    {
        return std::isfinite(x) and std::isfinite(y) and
               y >= std::nextafter(x, std::numeric_limits<T>::lowest()) and
               y <= std::nextafter(x, std::numeric_limits<T>::max());
    }
```
Thanks.

src/include/migraphx/matcher.hpp
umangyadav: > atol_mult == -1  
Atol is std::size_t it can't be -1

lakhinderwalia: Well, the code and the intent should match, really that is the point. You can of course ignore this suggestion, and I am perfectly fine with it. Thanks.

lakhinderwalia (approved):

PR #2698, opened 2024-01-30
Improve reduction fusion with reshape operators

pfultz2:
This will rewrite the reshapes to improve fusions for reductions. There is a `rewrite_reshapes`pass that resuses the same logic for `fuse_pointwise` and `fuse_reduce`. Right now, it will only work for reshape and multibroadcast. I plan to make it work for any arbitrary reshape/broadcast/transpose operators in the future.


bpickrel:
Would you please add code comments explaining what is being done and why?  This code will be very difficult to maintain for anyone other than the author.

test/fuse_reduce.cpp
```diff
@@ -327,4 +327,158 @@ TEST_CASE(reduce_reduce_broadcast)
...
+    }
+    run_pass(p1);
+    migraphx::program p2;
+    {
+        auto* mm = p2.get_main_module();
+        auto x1  = mm->add_parameter("x1", s1);
+        auto x2  = mm->add_parameter("x2", s1r);
+        auto y   = mm->add_parameter("y", s2);
+        auto x1r = mm->add_instruction(migraphx::make_op("reshape", {{"dims", s3.lens()}}), x1);
+        auto x2r = mm->add_instruction(migraphx::make_op("reshape", {{"dims", s3r.lens()}}), x2);
```
umangyadav, line 455: I don't understand why it would require s3 of shape {2, 2, 16, 64, 64}. To me it seems like it can just be {4, 16, 64, 64} with reduce axes {2, 3}.

src/common_dims.cpp
```diff
@@ -152,5 +146,42 @@ common_dims common_dims::compute(const std::vector<std::size_t>& dims1,
     return cd;
 }
 
+const std::vector<std::vector<std::size_t>>* common_dims::get_axes_map(std::size_t n) const
```
umangyadav, line 149: can you add docstring on what this is supposed to do ? It seems strange from looking at logic that it return either axes_map1 or map2 just on the basis of size. 

Also requires unit-tests inside `test/common_dims.cpp`

test/fuse_reduce.cpp
pfultz2: This uses the common_shape to compute these shapes which avoids collapsing dimensions. 

The first shape is `{2, 32, 4096}`. If the axes was 1 (instead of 2) then using `{2, 2, 16, 64, 64}` we can change the axes to `{1, 2}`, however if we used `{4, 16, 64, 64}` it would be impossible to set the axis for reduction.

Furthermore, even if we are not reducing on axes 1, one of the inputs could be broadcasting on this axis(ie the shape is `{2, 32, 4096}, {0, 1, 0}`) as well which will lead to extra contiguous being inserted.

src/common_dims.cpp
```diff
@@ -152,5 +146,42 @@ common_dims common_dims::compute(const std::vector<std::size_t>& dims1,
...
+{
+    if(axes_map1.size() == n)
+        return &axes_map1;
+    if(axes_map2.size() == n)
+        return &axes_map2;
+    return nullptr;
+}
+
+std::vector<std::size_t>
+common_dims::get_dimensions_for(const std::vector<std::size_t>& idims) const
```
umangyadav, line 159: Can you add unit-tests for this ?

src/common_dims.cpp
```diff
@@ -152,5 +146,42 @@ common_dims common_dims::compute(const std::vector<std::size_t>& dims1,
...
+        auto dim         = idims[i];
+        const auto& axes = (*axes_map)[i];
+        if(axes.size() == 1)
+        {
+            xdims[axes.front()] = dim;
+        }
+        else if(dim == 1)
+        {
+            for(auto axis : axes)
+                xdims[axis] = 1;
```
umangyadav, line 178: I think setting it as 1 would ensure `elements(xdims) != elements(idims)`.

src/include/migraphx/rewrite_reshapes.hpp
```diff
@@ -0,0 +1,182 @@
...
+    std::string name() const { return "rewrite_reshapes"; }
+    struct find_op_reshape_op
+    {
+        std::string op1;
+        std::string op2;
+
+        auto matcher() const
+        {
+            auto reshape =
+                match::name("reshape", "squeeze", "unsqueeze", "flatten")(match::used_once());
```
umangyadav, line 52: add `reshape_lazy` too

src/common_dims.cpp
pfultz2: No it wont. So say an axis of 2 is mapped to axes of `{3, 4}`, then if the dimension is `1` then the number of elements in axes `{3, 4}` should total `1`, so we can easily set it to `1`. If the dimension wasn't 1 and something like 8 then there are several combinations(ie `{1, 8}`, `{4, 2}`, `{8, 1}`, `{2, 4}`, etc), so that is why we dont update it.

src/common_dims.cpp
pfultz2: added.

src/common_dims.cpp
pfultz2: When we are computing a common dimension we have two shapes that are usually different ranks this lets us get the axes map without needing to know whether it was passed as the first or second parameter.

I guess there could be a case where it expands and collpases different parts of the tensor(ie `{2, 32, 4}` -> `{64, 2, 2}` resulting in two shapes that have the same rank. We should bail on that case, let me add a test case for that.

umangyadav (approved):

src/include/migraphx/rewrite_reshapes.hpp
umangyadav: `reshape_lazy` doesn't work ?

src/fuse_reduce.cpp
```diff
@@ -129,17 +130,37 @@ static auto insert_ins_in_submodule(module_ref sm, instruction_ref ins)
...
+                           module::inserter insert = nullptr)
+{
+    return insert_module_in_submodule(
+        sm, ins->inputs(), ins->module_inputs().front(), map_ins, std::move(insert));
+}
+
+static auto insert_module_in_submodule(module_ref sm,
+                                       const std::vector<instruction_ref>& inputs,
+                                       module_ref m,
+                                       module::inserter insert)
```
TedThemistokleous, line 160: Why no default assignment for insert here?

src/fuse_reduce.cpp
```diff
@@ -129,17 +130,37 @@ static auto insert_ins_in_submodule(module_ref sm, instruction_ref ins)
...
-    insert_params(sm, ins, map_ins);
-    auto* m        = ins->module_inputs().front();
-    auto param_map = get_ins_param_map(ins->inputs(), m);
+    insert_params(sm, inputs, map_ins);
+    auto param_map = get_ins_param_map(inputs, m);
     for(auto&& [input, param] : param_map)
     {
         map_ins[param] = map_ins.at(input);
     }
-    return sm->add_instructions(m, map_ins);
```
TedThemistokleous, line 142: Can we add an debug assert here if insert is never assigned to something other than nullptr?

src/fuse_reduce.cpp
```diff
@@ -332,6 +353,62 @@ struct find_reduce_reduce
...
+        }
+        std::sort(axes.begin(), axes.end());
+        auto* oldm = ins->module_inputs().front();
+        auto* sm   = mpm.create_module(oldm->name() + "_reshape");
+        insert_module_in_submodule(
+            sm, inputs, oldm, transform_op([&](const operation& sop) {
+                if(contains(sop.name(), "reduce"))
+                    return make_op(sop.name(), {{"axes", axes}});
+                if(sop.name() == "multibroadcast")
+                    return make_op("multibroadcast", {{"out_lens", ins->get_shape().lens()}});
```
TedThemistokleous, line 394: add coverage for return path + above set of transform_ops insert done

src/include/migraphx/rewrite_reshapes.hpp
TedThemistokleous: Maybe it requires another pass to add the copies back in and convert it to reshape after things are rewritten?

src/module.cpp
```diff
@@ -401,50 +424,61 @@ instruction_ref module::move_instructions(instruction_ref src, instruction_ref d
...
-    return this->insert_instructions(this->end(), instructions, std::move(map_ins));
+    return this->insert_instructions(
+        this->end(), instructions, std::move(map_ins), std::move(insert));
 }
 
 std::vector<instruction_ref>
 module::add_instructions(const_module_ref m,
-                         std::unordered_map<instruction_ref, instruction_ref> map_ins)
+                         std::unordered_map<instruction_ref, instruction_ref> map_ins,
+                         module::inserter insert)
```
TedThemistokleous, line 437: More a readability thing; show default assignment of nullptr here too on all the insert_instruction signatures you changed so the signatures appear the same as inmodule.hpp

src/fuse_reduce.cpp
pfultz2: Its never used optionally, but I can add it.

src/fuse_reduce.cpp
pfultz2: What do you mean?

pfultz2: > "Hey add/mull/sub/etc op that takes args now has null arguments"

`compute_shape` should already throw an error on invalid inputs.

TedThemistokleous: Fair point.

src/include/migraphx/rewrite_reshapes.hpp
pfultz2: There is no `reshape_lazy` until after lowering.

src/include/migraphx/rewrite_reshapes.hpp
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        template <class F>
+        static instruction_ref find_input_if(instruction_ref start, instruction_ref last, F f)
+        {
+            while(start != last)
+            {
+                if(f(start))
+                    return start;
+                if(start->inputs().size() != 1)
```
TedThemistokleous, line 93: Add additional test case for this early return condition.  Part of me wants to say make this a break instead as you're returning the same thing as the  outside of the loop.

src/include/migraphx/rewrite_reshapes.hpp
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        static bool match_input(instruction_ref ins, instruction_ref x_ins)
+        {
+            if(ins->inputs().empty())
+                return false;
+            auto input = ins->inputs().front();
+            if(input->name() == "contiguous")
+                return match_input(input, x_ins);
+            return x_ins == input;
```
TedThemistokleous, line 107: Add simple case where we hit this return.

src/include/migraphx/rewrite_reshapes.hpp
pfultz2: Looks like there is coverage for that. Also early return seems to make it clearer what will happen when this condition is met rather having to follow where the break will end up.

TedThemistokleous: fair point

TedThemistokleous (approved):

TedThemistokleous (commented):
Still seeing some minor CI fails/coverage

TedThemistokleous (changes requested):
Fix licensing + add more coverage

src/include/migraphx/rewrite_reshapes.hpp
pfultz2: There isnt coverage for this line. We actually wont hit this with the way its currently called, but if this was called in a different context we wouldn't want it to keep traversing here which is why the condition is here.

src/include/migraphx/rewrite_reshapes.hpp
pfultz2: There is coverage for this line. Did you mean the return above?

TedThemistokleous: Yes

src/include/migraphx/rewrite_reshapes.hpp
TedThemistokleous: I understand all good.

src/include/migraphx/rewrite_reshapes.hpp
pfultz2: Test added.

TedThemistokleous (approved):

pfultz2:
The perf report is a fluke. I dont see a difference in performance between develop and this branch.

PR #2714, opened 2024-02-01
Simplify `select_module` after `simplify_dyn_ops` pass

CharlieL7:
* This pass recalculates the `output_dyn_shapes` attribute for `select_module` operators after the rest of `simplify_dyn_ops`.
* Recalculating the outputs shapes is required if the other matchers change the output shape of the submodules.
  * For example, the `Resize` operator can output `{0, max_int}` dynamic dimensions, but will be simplified for dynamic batch to a static shape.


pfultz2:
We cant update `compute_shape` in `select_module` to compute the correct dynamic shape output?

CharlieL7:
> We cant update `compute_shape` in `select_module` to compute the correct dynamic shape output?

We could if we use `shape.compute_shape(shape_inputs, module_inputs)`. Would have to change `select_module` to remove the `output_dyn_shapes` attribute and related changes to tests/IR.

src/simplify_dyn_ops.cpp
```diff
@@ -328,6 +464,7 @@ void simplify_dyn_ops::apply(module& m) const
                         find_const_3in_slice{},
                         find_const_4in_slice{},
                         find_const_alloc_fill{});
+    match::find_matches(m, simplify_select_module_output_shape{});
```
bpickrel, line 467: Does adding a second find_matches() call have a different effect than adding the new pass at the end of the preceding list?

CharlieL7: Yes, tuns the matcher after running all of the prior through the model

src/simplify_dyn_ops.cpp
```diff
@@ -318,6 +319,141 @@ struct find_const_alloc_fill
...
+        auto sm_ins           = mr.result;
+        auto sm_module_inputs = sm_ins->module_inputs();
+        std::vector<std::vector<shape>> all_output_shapes(sm_module_inputs.size());
+        std::transform(sm_module_inputs.begin(),
+                       sm_module_inputs.end(),
+                       all_output_shapes.begin(),
+                       [](auto submod) { return submod->get_output_shapes(); });
+        // check that all of the submodules have the same number of outputs and all respective
+        // outputs have the same rank and type
+        auto shapes_ndim  = get_shapes_ndim(all_output_shapes.front());
```
bpickrel, line 343: Can this ever get called with empty `all_output_shapes`?

test/simplify_dyn_ops_test.cpp
```diff
@@ -571,4 +571,254 @@ TEST_CASE(const_alloc_fill)
...
+        // create batch submodules
+        auto create_submodule = [&](std::size_t batch_size, const std::string& module_name) {
+            auto* submod = p1.create_module(module_name);
+            migraphx::shape sm_shape{migraphx::shape::float_type,
+                                     {{batch_size, batch_size}, {4, 20}}};
+            auto sm_input = submod->add_parameter("data", sm_shape);
+            migraphx::shape lit_s{migraphx::shape{migraphx::shape::float_type, {1}}};
+            auto literal_ins = submod->add_literal(migraphx::literal{lit_s, {6}});
+            auto broadcast_lit =
+                submod->add_instruction(migraphx::make_op("multibroadcast"), literal_ins, sm_input);
```
umangyadav, line 716: why wasn't this changed ? other tests have it changed.

src/simplify_dyn_ops.cpp
```diff
@@ -318,6 +319,141 @@ struct find_const_alloc_fill
...
+        auto shapes_types = get_shapes_types(all_output_shapes.front());
+        bool check = std::all_of(
+            all_output_shapes.begin() + 1, all_output_shapes.end(), [&](auto out_shapes) {
+                bool same_types = get_shapes_types(out_shapes) == shapes_types;
+                bool same_ndim  = get_shapes_ndim(out_shapes) == shapes_ndim;
+                return same_types and same_ndim;
+            });
+        if(not check)
```
umangyadav, line 351: you can  do if(std::any_of()) and then return

test/simplify_dyn_ops_test.cpp
CharlieL7: Not simplify-able in this case

src/simplify_dyn_ops.cpp
CharlieL7: It doesn't seem possible since that would mean there are no submodules or none of the submodules have any output.

test/simplify_dyn_ops_test.cpp
```diff
@@ -571,4 +571,254 @@ TEST_CASE(const_alloc_fill)
...
+            {dim1, dim2, dim3, dim4});
+        auto ret =
+            mm1->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), sm_ins);
+        mm1->add_return({ret});
+    }
+
+    EXPECT(p0 == p1);
+}
+
+TEST_CASE(select_module_update1)
```
bpickrel, line 659: Could you please comment how this case differs from select_module_update1?  The only difference I can see is that one fixed dimension value is 20 instead of 4--is that meaningful?

test/simplify_dyn_ops_test.cpp
```diff
@@ -571,4 +571,254 @@ TEST_CASE(const_alloc_fill)
...
+                                           {{"output_dyn_shapes", migraphx::to_value(out_attr)}}),
+            {input0},
+            {dim1, dim2, dim3, dim4});
+        auto ret =
+            mm0->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), sm_ins);
+        mm0->add_return({ret});
+    }
+    migraphx::run_passes(p0, {migraphx::simplify_dyn_ops{}, migraphx::dead_code_elimination{}});
+
+    // difference is `output_dyn_shapes` attribute in `select_module`
```
bpickrel, line 784: I don't think this comment is correct

test/simplify_dyn_ops_test.cpp
```diff
@@ -571,4 +571,254 @@ TEST_CASE(const_alloc_fill)
...
+        auto ret =
+            mm1->add_instruction(migraphx::make_op("get_tuple_elem", {{"index", 0}}), sm_ins);
+        mm1->add_return({ret});
+    }
+
+    EXPECT(p0 == p1);
+}
+
+// contrived example where each submodule to select_module outputs the same static shape.
+TEST_CASE(select_module_update2)
```
bpickrel, line 746: The two programs are identical as far as I can tell.  What are we testing?

src/simplify_dyn_ops.cpp
```diff
@@ -318,6 +319,141 @@ struct find_const_alloc_fill
...
+     * Assuming all shapes have the same ndim.
+     */
+    shape dyn_shape_from_shapes(std::vector<shape> shape_vec) const
+    {
+        // making 2D matrices of min_lens and max_lens
+        // specifically using uint64_t because we're going to put the values into a tensor_view
+        // later
+        std::vector<uint64_t> all_min_lens;
+        std::vector<uint64_t> all_max_lens;
+        for(int i : range(shape_vec.size()))
```
bpickrel, line 405:
```suggestion
                for(auto s : shape_vec)
        {
            // auto s        = shape_vec.at(i);

```

test/simplify_dyn_ops_test.cpp
bpickrel: Could you add a comment?  It's very hard to spot by code inspection if you don't know what you're looking for.

umangyadav (approved):

CharlieL7:
Keeping this PR for now and pushing back the refactor to pass_manager for a later date. Tracking the pass_manager refactor as issue #2714 .

bpickrel (approved):

PR #2726, opened 2024-02-06
Rewrite reduce mean

attila-dusnoki-htec:
Depends on https://github.com/ROCm/AMDMIGraphX/pull/2673

This change rewrites every `reduce_mean` to `reduce_sum` if possible.
Skips it if the `n` literal would be larger than half max and when it is integral type to avoid integer division.


src/rewrite_reduce.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+        bool is_integral = false;
+        double max_n     = 0;
+        input->get_shape().visit_type([&](auto t) {
+            is_integral = t.is_integral();
+            max_n       = t.max();
+        });
+
+        // avoid integer division
+        if(is_integral)
```
pfultz2, line 76: We shouldn't bail for integer. Instead we should just do the division after the sum when its an integer.

src/rewrite_reduce.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+        auto dims = input->get_shape().lens();
+        double n  = std::accumulate(
+            axes.begin(), axes.end(), 1, [&](const auto& prod_value, const auto& axis) {
+                auto norm_axis = axis < 0 ? dims.size() + axis : axis;
+                return prod_value * dims.at(norm_axis);
+            });
+
+        // avoid overflow
+        if(n >= max_n)
```
pfultz2, line 87: Ideally, we should use a higher precision for these cases. For now you could just change it to `n >= max_n / 4`(so the logic [here](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/targets/gpu/compile_gen.cpp#L227) will still be applied).

attila-dusnoki-htec:
> We shouldn't bail for integer. Instead we should just do the division after the sum when its an integer.

Any suggestion how to proceed with int division?
If have it, `div_const` simplify will make a `mul-recip` transformation. Which leads to `multiply-with-zero`, and even that will be optimized out.

I added an extra check for integer division, but it seems from the `simplify_div_const` test fail, that it should do `1/INT -> 0` always. Is that really the intended behaviour? Because i see no dtype conversion to float.

src/rewrite_reduce.cpp
```diff
@@ -54,9 +54,71 @@ struct find_softmax
...
+
+        auto n_literal = m.add_literal(literal{{input->get_shape().type(), {1}}, {n}});
+        if(is_integral)
+        {
+            auto out_lens = reduce_mean->get_shape().lens();
+            for(const auto axis : axes)
+            {
+                auto norm_axis      = axis < 0 ? out_lens.size() + axis : axis;
+                out_lens[norm_axis] = 1;
+            }
```
pfultz2, line 94: This loop can be removed as `reduce_mean` has already computed the output lens.

src/rewrite_reduce.cpp
```diff
@@ -54,9 +54,71 @@ struct find_softmax
...
+
+            auto reduce_sum =
+                m.insert_instruction(reduce_mean, make_op("reduce_sum", {{"axes", axes}}), input);
+            auto n_bcast = m.insert_instruction(
+                reduce_mean, make_op("multibroadcast", {{"out_lens", out_lens}}), n_literal);
+            auto div = m.insert_instruction(reduce_mean, make_op("div"), {reduce_sum, n_bcast});
```
pfultz2, line 100: Use `insert_common_op(reduce_mean, make_op("div"), {reduce_sum, n_literal})` instead of inserting `multibroadcast` manually.

src/rewrite_reduce.cpp
```diff
@@ -54,9 +54,71 @@ struct find_softmax
...
+            m.replace_instruction(reduce_mean, div);
+        }
+        else
+        {
+            auto n_bcast = m.insert_instruction(
+                reduce_mean, make_op("multibroadcast", {{"out_lens", dims}}), n_literal);
+            auto new_input = m.insert_instruction(reduce_mean, make_op("div"), {input, n_bcast});
```
pfultz2, line 107: Use `insert_common_op` here as well.

src/simplify_algebra.cpp
```diff
@@ -1200,6 +1200,12 @@ struct find_div_const
         auto ins   = r.result;
         auto c_ins = r.instructions["c"];
 
+        bool is_integral = false;
+        c_ins->get_shape().visit_type([&](auto t) { is_integral = t.is_integral(); });
+        // avoid 1 / integer
+        if(is_integral)
+            return;
```
pfultz2, line 1207: A similiar fix was added in #2547. We should merge #2547 in first before merging in this PR:

https://github.com/ROCm/AMDMIGraphX/pull/2547/files#diff-0e2e45cd7b8d50c1e0790164b152eb4c8ef7c58c3baffd88e3bed1cb2a754d8bR1203

attila-dusnoki-htec:
Depends on https://github.com/ROCm/AMDMIGraphX/pull/2547

src/rewrite_reduce.cpp
```diff
@@ -54,9 +55,59 @@ struct find_softmax
...
+        auto dims = input->get_shape().lens();
+        double n  = std::accumulate(
+            axes.begin(), axes.end(), 1, [&](const auto& prod_value, const auto& axis) {
+                auto norm_axis = axis < 0 ? dims.size() + axis : axis;
+                return prod_value * dims.at(norm_axis);
+            });
+
+        // avoid overflow (the larger value will be later handled)
+        if(n >= max_n / 4)
+            return;
```
umangyadav, line 85: This could still degrade accuracy. 

If some values inside the inputs are closer to `numeric_limits<type>::max()` or `min()` then it can possibly result in inaccurate results. 
e.g. Let's say as the worst case all the  values inside input are `min()` for floating point and `n == max_n / 4` then current logic would underflow to zero i think. v/s in theory it should result in `min()` as the average value. 

Ideally any accumulation should happen in higher precision.  not sure how to best handle it though.

umangyadav (commented):
LGTM but not sure how it affects the accuracy.

umangyadav (approved):

src/rewrite_reduce.cpp
```diff
@@ -54,9 +55,59 @@ struct find_softmax
...
+
+        bool is_integral = false;
+        double max_n     = 0;
+        input->get_shape().visit_type([&](auto t) {
+            is_integral = t.is_integral();
+            max_n       = t.max();
+        });
+
+        auto dims = input->get_shape().lens();
+        double n  = std::accumulate(
```
pfultz2, line 77: This can be calculated as `auto n = input->get_shape().elements() / reduce_mean->get_shape().elements()`.

pfultz2 (approved):

PR #2727, opened 2024-02-06
fix test_ref makelist

bpickrel:
One-line bug fix.  Adds a `target_link_libraries` statement to the CMakeLists file for the test_ref target.

Without this, trying to make `test_ref` without building any other targets leads to runtime error because a needed library `libmigraphx_ref.so` isn't created.  The problem can be worked around by making nearly any other Migraphx target program.


ahsan-ca (approved):
Thanks for fixing this.

umangyadav (approved):

PR #2728, opened 2024-02-07
Eliminate zero points for Q/DQLinearOps

umangyadav:
solves #2695


src/include/migraphx/simplify_qlinear_ops.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <string>
+#include <migraphx/config.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+struct module;
+
+/**
+Eliminates zero point if it's zero
```
shivadbhavsar, line 36: Does this need a whole new pass? Why not just add it to simplify_qdq?

umangyadav: It can be part of simplify_qdq. 

Initially plan was also to look for scales and if they are 1.0 then eliminating them. But i find it very unlikely case and won't be beneficial. Also we have `find_unit_ops` inside simplify_algebra that would take care of it.  There could be opportunities in future to do more processing in this pass.
Tests would require a lot of changes because simplify_qdq is used inside test/quantization.cpp also.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -38,6 +38,7 @@ namespace op {
 struct quantizelinear
 {
     std::string name() const { return "quantizelinear"; }
+    std::optional<migraphx::shape> output_shape;
```
pfultz2, line 41: This should be `std::optional<migrapx::shape::type_t>`, we only need the output data type. We dont need to the dimensions or strides as we will get that from the inputs.

umangyadav: Done.

src/targets/gpu/mlir.cpp
```diff
@@ -459,9 +460,12 @@ struct mlir_program
     {
         std::vector<MlirNamedAttribute> attributes;
         attributes.reserve(v.size());
-        std::transform(v.begin(), v.end(), std::back_inserter(attributes), [&](const value& x) {
-            return name_attribute(x.get_key(), x.without_key());
-        });
+        migraphx::transform_if(
```
umangyadav, line 463: @krzysz00 FYI

krzysz00: Agreed

src/simplify_qlinear_ops.cpp
```diff
@@ -0,0 +1,78 @@
...
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+auto get_qlinear_ops_names()
+{
+    static std::unordered_set<std::string> qdq_names = {"quantizelinear", "dequantizelinear"};
+    return qdq_names;
+}
+
+struct eliminate_zero_point
```
pfultz2, line 45: Can we move this to simplify_algebra?

umangyadav: Moved it. but kept tests separate.

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -58,6 +65,10 @@ struct quantizelinear
         {
             return {inputs[2].type(), inputs[0].lens(), inputs[0].strides()};
         }
+        if(out_type.has_value())
```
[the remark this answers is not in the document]
umangyadav: output type of quantizelinear is always type of zero point. therefore even if dev has set out_type it should match exactly as zero_point's type. I don't see need to first check for `out_type`.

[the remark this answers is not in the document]
shivadbhavsar: Its fine as it is for now. I guess the real issue is there is nothing preventing us from setting the `out_type` in a pass even if there is a non-zero zero point input. In that case out_type would have no meaning.

shivadbhavsar (approved):

src/include/migraphx/op/quantizelinear.hpp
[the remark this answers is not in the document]
umangyadav: > . In that case out_type would have no meaning. 
That is fine. In that case output type of quantizelinear would be output type of zero point. That's how it should be.

src/simplify_algebra.cpp
```diff
@@ -1255,6 +1255,37 @@ struct find_neg_unit_ops
...
+    {
+        static std::unordered_set<std::string> qdq_names = {"quantizelinear", "dequantizelinear"};
+        return qdq_names;
+    }
+    auto matcher() const
+    {
+        return match::name(get_qlinear_ops_names())(
+            match::arg(0)(match::any().bind("x")),
+            match::arg(1)(match::any().bind("scale")),
+            match::arg(2)(match::has_value(0.0f, 0, 0).bind("zp")));
```
CharlieL7, line 1270: minor: this bind looks unnecessary

test/op_shape_test.cpp
```diff
@@ -2525,6 +2525,15 @@ TEST_CASE(qlinear_fp16)
...
 
+TEST_CASE(qlinear_output_shape)
+{
+    migraphx::shape scales{migraphx::shape::half_type, {2, 4}};
+    migraphx::shape input{migraphx::shape::half_type, {2, 4}};
+    migraphx::shape result{migraphx::shape::int8_type, {2, 4}};
+    expect_shape(
+        result, migraphx::make_op("quantizelinear", {{"out_type", result.type()}}), input, scales);
+}
+
```
CharlieL7, line 2536: minor: could add a test that sets the `out_type` attribute.

CharlieL7 (approved):

test/op_shape_test.cpp
umangyadav: Done.

PR #2730, opened 2024-02-07
PackInt4 Operator

umangyadav:
Packs Int8 instruction into 2xint4 along specified axis. 

Example Ref implementation: 
NCHW weights: 
Lens = {1, 3, 16, 16}, Strides {768, 256, 16, 1}, Data Type = Int8
Would be encoded as following shape: 
Lens = {1, 3, 16, 8}, Strides {384, 128, 8, 1} Data Type = Int8
Two consecutive Int8 values would get packed into one int8 value: 
```
x[i] = 0x0F & Input[j]
x[i] = (input[j]+1)) << 4) | x[i] 
```


src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,104 @@
...
+    {
+        argument result{output_shape};
+        auto in_shape = args.front().get_shape();
+        visit_all(result, args[0])([&](auto output, auto input) {
+            par_for(output_shape.elements(), [&](auto i) {
+                auto data_idx          = output_shape.multi(i);
+                auto in_data_multi_idx = data_idx;
+                in_data_multi_idx[axis] *= 2;
+                // mask first 4 bits, keep it little endian.
+                output[i] = 0x0F & static_cast<int8_t>(input[in_shape.index(in_data_multi_idx)]);
```
pfultz2, line 90: `input[in_shape.index(...)]` is wrong. `.index` computes the offset and then you pass it to `input[]` which takes the position(not the offset). You can do `input(in_data_multi_idx.begin(), in_data_multi_idx].end())` to read the value which is a little ugly. Although, I think it would be better to add a `operator[]` overload to `tensor_view` take a range:

```cpp
    template <class Range, MIGRAPHX_REQUIRES(not std::is_integral<Range>{})>
    T& operator[](const Range& r)
    {
        return (*this)(r.begin(), r.end());
    }
```

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,104 @@
...
+        }
+        auto strides  = in_shape.strides();
+        auto lens     = in_shape.lens();
+        auto new_lens = lens;
+        if(lens[axis] % 2 != 0)
+        {
+            MIGRAPHX_THROW("PACK_INT4: Can not pack axis that has odd lengths");
+        }
+        new_lens[axis] /= 2;
+        return {migraphx::shape::int8_type, new_lens};
```
pfultz2, line 77: The output shape should probably be uint8 since bit fiddling with signed integers is borderline UB.

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
+    {
+        argument result{output_shape};
+        auto in_shape = args.front().get_shape();
+        visit_all(result, args[0])([&](auto output, auto input) {
```
pfultz2, line 84: I wouldnt put them in the same visit as the `input` and `output` could be different types. You dont really need to visits since the `output` will always have the same type so you could do `auto output = result.get<uint8>()`.

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,104 @@
...
+            par_for(output_shape.elements(), [&](auto i) {
+                auto data_idx          = output_shape.multi(i);
+                auto in_data_multi_idx = data_idx;
+                in_data_multi_idx[axis] *= 2;
+                // mask first 4 bits, keep it little endian.
+                output[i] = 0x0F & static_cast<int8_t>(input[in_shape.index(in_data_multi_idx)]);
+                in_data_multi_idx[axis] += 1;
+                // static_cast to avoid ambiguous call during visit_all
```
pfultz2, line 92: I would use `if constexpr(size(T) == 1)`  to avoid compiling for invalid data types.

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,104 @@
...
+    {
+        return pack(f(self.axis, "axis"));
+    }
+
+    migraphx::shape normalize_compute_shape(std::vector<migraphx::shape> inputs) const
+    {
+        // remove standard requirement later.
+        check_shapes{inputs, *this}.same_dims().has(1).standard();
+        auto in_shape = inputs.front();
+        if(in_shape.type() != migraphx::shape::int8_type)
```
pfultz2, line 65: You should probably start with unsigned since thats easier.

src/include/migraphx/op/pack_int4.hpp
pfultz2: I dont think this will work for negative integers. You need to encode the two's complement for signed integers:

```cpp
auto input1 = input[in_data_multi_idx];
auto raw_input1 = bit_cast<uint8>(abs(input1));
if (input1 < 0) raw_input1 = ~raw_input1;
output[i] = 0x0F & raw_input1;
```

src/include/migraphx/op/pack_int4.hpp
krzysz00: Yeah, I think I can agree with that.
If we end up seeing these in MLIR modules it'll require some fiddling (MLIR technically has unsigned types, but they're a "we put this here back when you couldn't do custom types well for the the convenience of translating out of C/C++, they're barely integrated into the standard library at best")

src/include/migraphx/op/pack_int4.hpp
umangyadav: > I dont think this will work for negative integers. 

I'd be assuming Uint8 as the input type for the pack_int4. So i don't need to worry about negative numbers.

src/include/migraphx/op/pack_int4.hpp
umangyadav: how do i get that `T` ? i can get type of `input` or `output` tensorview. but it can't be used inside `constexpr`.

umangyadav:
``` 
  using type = typename decltype(input)::value_type;
            if constexpr(std::is_same<type, uint8_t>{})
```
Need to do this otherwife FP8 is also byte size one.

src/include/migraphx/op/pack_int4.hpp
pfultz2: But you are only accepting `int8_type`, this should be changed to only accept `uint8_type`.

umangyadav: Pushed changes to use uint8_type

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,107 @@
...
+            if constexpr(std::is_same<type, uint8_t>{})
+            {
+                auto output = result.get<uint8_t>();
+                par_for(output_shape.elements(), [&](auto i) {
+                    auto data_idx          = output_shape.multi(i);
+                    auto in_data_multi_idx = data_idx;
+                    in_data_multi_idx[axis] *= 2;
+                    auto input_val = input[in_data_multi_idx];
+                    // mask first 4 bits, keep it little endian.
+                    output[i] = static_cast<uint8_t>(0x0F) & input_val;
```
pfultz2, line 93: I think you can write it as `0x0Fu` or `uint8_t(0x0F)`.

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,107 @@
...
+                    in_data_multi_idx[axis] += 1;
+                    input_val = input[in_data_multi_idx];
+                    output[i] = (input_val << 4) | (output[i]); // NOLINT
+                });
+            }
```
pfultz2, line 98: There should be an else that throws an errror.

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,107 @@
...
+                par_for(output_shape.elements(), [&](auto i) {
+                    auto data_idx          = output_shape.multi(i);
+                    auto in_data_multi_idx = data_idx;
+                    in_data_multi_idx[axis] *= 2;
+                    auto input_val = input[in_data_multi_idx];
+                    // mask first 4 bits, keep it little endian.
+                    output[i] = static_cast<uint8_t>(0x0F) & input_val;
+                    in_data_multi_idx[axis] += 1;
+                    input_val = input[in_data_multi_idx];
+                    output[i] = (input_val << 4) | (output[i]); // NOLINT
```
pfultz2, line 96: Why is `NOLINT` added here?

src/include/migraphx/tensor_view.hpp
```diff
@@ -119,6 +119,18 @@ struct tensor_view
         return m_data[m_shape.index(i)];
     }
 
+    template <class Range>
```
pfultz2, line 122: This should be `template <class Range, MIGRAPHX_REQUIRES(not std::is_integer<Range>{})>` so other integer indices types(besides `std::size_t`) wont use this overload.

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+        new_lens[axis] /= 2;
+        return {migraphx::shape::uint8_type, new_lens};
+    }
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
+    {
+        argument result{output_shape};
+        auto in_shape = args.front().get_shape();
+        args[0].visit([&](auto input) {
```
pfultz2, line 82: I dont think you need a visit for the input since it should always be `uint8`. You can do `auto input = args[0].get<uint8_t>()` instead.

src/include/migraphx/op/pack_int4.hpp
umangyadav: tidy for somereason complains binary_op being used with signed operands (even if i statically cast 4 and input_val them to unsigned, output_val is declared as unsigned).

src/include/migraphx/tensor_view.hpp
umangyadav: I have `decltype()` which would fail if it is integer.

src/include/migraphx/op/pack_int4.hpp
umangyadav: 57017 warnings generated.
/home/umayadav/repo/AMDMIGraphX/src/include/migraphx/op/pack_int4.hpp:93:25: error: use of a signed integer operand with a binary bitwise operator [hicpp-signed-bitwise,-warnings-as-errors]
            output[i] = (uint8_t(input_val) << uint8_t(4)) | output[i];
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ~

CMake Error at tidy-target-migraphx-_home_umayadav_repo_AMDMIGraphX_build_57_src_ops_migraphx_op_pack_int4_hpp_cpp.cmake:82 (message):
  Clang tidy failed.

src/include/migraphx/op/pack_int4.hpp
umangyadav: Done.

src/include/migraphx/op/pack_int4.hpp
umangyadav: Done.

shivadbhavsar (approved):

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -0,0 +1,102 @@
...
+        par_for(output_shape.elements(), [&](auto i) {
+            auto data_idx          = output_shape.multi(i);
+            auto in_data_multi_idx = data_idx;
+            in_data_multi_idx[axis] *= 2;
+            auto input_val = input[in_data_multi_idx];
+            // mask first 4 bits, keep it little endian.
+            output[i] = uint8_t(0x0F) & input_val;
+            in_data_multi_idx[axis] += 1;
+            input_val = input[in_data_multi_idx];
+            output[i] = (input_val << 4u) | output[i]; // NOLINT
```
pfultz2, line 93: I dont think the `NOLINT` is needed here, this should all be unsigned.

umangyadav: tidy is complaining even though all are unsigned.

pfultz2: Can you change at least to `NOLINT(...)` with the name of the check that is complaining: 

https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics

umangyadav: Done

umangyadav: I still see tidy warning despite having // NOLINT(hipcc-signed-bitwise)

```
/home/umayadav/repo/AMDMIGraphX/src/include/migraphx/op/pack_int4.hpp:93:25: error: use of a signed integer operand with a binary bitwise operator [hicpp-signed-bitwise,-warnings-as-errors]
            output[i] = (input_val << 4u) | output[i]; // NOLINT(hipcc-signed-bitwise)
                        ^~~~~~~~~~~~~~~~~ ~

CMake Error at tidy-target-migraphx-_home_umayadav_repo_AMDMIGraphX_build_src_ops_migraphx_op_pack_int4_hpp_cpp.cmake:83 (message):
  Clang tidy failed.

```

pfultz2 (approved):

pfultz2 (approved):

PR #2731, opened 2024-02-07
Add dim_param parsing to onnx_parser and driver

turneram:


pfultz2:
I think this looks good. In the future we could extend to handle cases like `x + 1` and such.

src/include/migraphx/onnx.hpp
```diff
@@ -41,6 +41,8 @@ struct onnx_options
     shape::dynamic_dimension default_dyn_dim_value = {1, 1};
     /// Explicitly specify the dims of an input
     std::unordered_map<std::string, std::vector<std::size_t>> map_input_dims = {};
+    /// Explicitly specify a symbolic named parameter dimension
+    std::unordered_map<std::string, shape::dynamic_dimension> map_dim_params = {};
```
pfultz2, line 45: I would just call this `dim_params`.

pfultz2 (approved):

PR #2733, opened 2024-02-08
fixed msgpack_binary2 large buffer test

aljanjic-amd:
Test msgpack fails on Windows in both Release and Debug version in test_msgpack_binary2 Test case. In this Test case are compared two different functions to_msgpack(bin) and msgpack_buffer(bin) where is expected that these two functions have the same behavior. The reason why the test is failing is because the size of bin is 4Gb and the mgspack_buffer(bin) function cuts the length to 2Gb on the line std::string str = buffer.str(); in the definition of the function where the type of variable buffer is std::stringstream. The upper limit of the size of std::string on Windows is 2Gb, so practically for every size greater than 2Gb, line of code std::string str = buffer.str();  won’t work properly. In Linux CI large buffer tests are disabled, therefore, this test is also disabled on Windows. @apwojcik @pfultz2


apwojcik:
We will turn off large buffer tests for now when testing MIGraphX on Windows; later, we will implement verifying data with chunks written to the file, as we talked about in the meeting.

PR #2740, opened 2024-02-08
Create CODEOWNERS file

dgaliffiAMD:
Define global CODEOWNERS: @causten @TedThemistokleous @pfultz2 @umangyadav.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @causten @TedThemistokleous @pfultz2 @umangyadav
+* @causten


causten (changes requested):
Since these code owner names will be applied to any PR I don't want it to globally add 4 reviewers per PR.  I'd like to get a little more accurate.  The team can chime in about which directory folders/files they would like to be automatic reviewers on.

dgaliffiAMD:
> Since these code owner names will be applied to any PR I don't want it to globally add 4 reviewers per PR. I'd like to get a little more accurate. The team can chime in about which directory folders/files they would like to be automatic reviewers on.

No problem. I'll gladly update this PR once that assignment is available.

causten (approved):

PR #2741, opened 2024-02-08
Show mlir program when tracing benchmarking

pfultz2:
This will show the mlir program when using `MIGRAPHX_TRACE_BENCHMARKING=3`.


src/targets/gpu/mlir.cpp
```diff
@@ -960,6 +952,24 @@ void adjust_param_shapes(module& m, const std::vector<shape>& inputs)
...
 }
 
+std::string dump_mlir(const module& m, const std::vector<shape>& inputs)
+{
+    module mm;
+    const_module_ref mr = &m;
+    if(not inputs.empty())
+    {
+        mm = m;
+        mr = &mm;
```
umangyadav, line 962: Why does it require const_ref ? Shouldn't just copied module `mm` work ?

pfultz2: I only want to copy the module if there is no input shapes because we wont be adjusting the parameters.

umangyadav: Module with empty input shape is unlikely case it would have been const-folded.

umangyadav: I think it is better to assume inputs are not empty. it would be simpler

pfultz2: There is two overloads to `dump_mlir`. One just takes the module(which we dont want to copy) and the other overload takes the input shapes, which might be different than the shapes in the module so we need to modify the module which we will use a copy for this case.

src/targets/gpu/compile_ops.cpp
```diff
@@ -179,6 +185,8 @@ struct compile_plan
                                    std::cout << "No binary" << std::endl;
                                return std::numeric_limits<double>::max();
                            }
+                           if(trace_level > 2)
```
umangyadav, line 188: add this option to documents

src/targets/gpu/mlir.cpp
umangyadav: > and the other overload takes the input shapes, which might be different than the shapes in the module so we need to modify the module which we will use a copy for this case. 

Yes but those input shapes parameter would be input arguments to the precompile_op instruction. If they are empty that means MLIR module also doesn't take any inputs.

pfultz2: When the inputs are empty it means skip doing param adjustments.

src/targets/gpu/compile_ops.cpp
```diff
@@ -153,22 +159,33 @@ struct compile_plan
...
+    {
+        if(config)
+            return to_string(config->problem);
+        return "<no problem key>";
+    }
+
     const compiled_result& benchmark() const
     {
         const auto trace_level = value_of(MIGRAPHX_TRACE_BENCHMARKING{});
+        if(trace_level > 0 and not results.empty())
```
bpickrel, line 172: Suggest you consolidate some of these checks with `if ... or ...`

bpickrel (approved):

PR #2748, opened 2024-02-10
match gemm_softmax_gemm when there is no scale

shivadbhavsar:
Make the scale mul op optional when looking for attention. Required to get mlir attention for for SDXL workflow.


src/targets/gpu/prefuse_ops.cpp
```diff
@@ -181,17 +182,20 @@ struct find_gemm_softmax_gemm
...
         float scale = 1.0;
-        scale_lit->eval().visit([&](const auto s) {
-            // CK only supports single-valued scale
-            if(std::all_of(
-                   s.begin() + 1, s.end(), [&](auto v) { return float_equal(v, s.front()); }))
-                scale = s.front();
-            else
-                return;
-        });
+        if(r.instructions.find("scale") != r.instructions.end())
```
pfultz2, line 187: Use `contains(r.instructions, "scale")` instead.

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -181,17 +182,20 @@ struct find_gemm_softmax_gemm
...
-                scale = s.front();
-            else
-                return;
-        });
+        if(contains(r.instructions, "scale"))
+        {
+            auto scale_lit = r.instructions["scale"];
+            scale_lit->eval().visit([&](const auto s) {
+                // CK only supports single-valued scale
+                if(std::all_of(
```
kahmed10, line 192: It might be better to flip the logic so that it returns if scale values are different.

kahmed10:
Can you add a test case?

shivadbhavsar:
> Can you add a test case?

This pass only really runs with special flags. How would the test cases work for this? 
There are no existing cases for this pass

src/targets/gpu/prefuse_ops.cpp
shivadbhavsar: added

kahmed10: I was thinking something like this
```suggestion
                if(not std::all_of(....) return;
                scale = s.front();
```

shivadbhavsar: But thats inside a visit, that return will not exit the apply method for this matcher

test/gpu/prefuse_ops.cpp
```diff
@@ -0,0 +1,204 @@
...
+        auto attn =
+            mm->add_instruction(pre_gemm_softmax_gemm{migraphx::make_op("dot"), 2}, x, y, z);
+
+        mm->add_return({attn});
+        return p;
+    };
+
+    migraphx::program p1 = create_program();
+    migraphx::program p2;
+    if(migraphx::gpu::mlir_attention_enabled())
```
pfultz2, line 86: Dont rely on this global setting. Instead add a flag to the `prefuse_ops` class like `enable_attention` that can be set to true during tests. And then you pass that along to the `find_gemm_softmax_gemm` so the matcher will match it.

src/targets/gpu/prefuse_ops.cpp
kahmed10: oh yeah I saw that too. But looks like you rewrote it and looks fine now.

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -162,36 +162,53 @@ auto is_mlir_gemm()
...
+    return match::make_basic_pred_matcher([=](instruction_ref ins) {
+        if(ins->name() != "dot")
+            return false;
+        return enable_attention;
+    });
+}
+
 struct find_gemm_softmax_gemm
 {
+    bool enable_attention;
```
pfultz2, line 176: Set this to `false`, so its not uninitialized when default constructing.

shivadbhavsar: done

pfultz2 (approved):

PR #2772, opened 2024-02-14
Add support for the Einsum operator

mirza-halilcevic:
- Implement ONNX parsing support for the [Einsum](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Einsum) operator
- Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/92


src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+    {
+        std::tuple<string_vec, std::string, char_int_map, bool> ret;
+        // cppcheck-suppress variableScope
+        auto& [input_terms, output_term, label_count, explicit_form] = ret;
+
+        std::string term;
+        bool has_ellipsis = false;
+        explicit_form     = false;
+
+        for(int i = 0; i < equation.size(); ++i)
```
pfultz2, line 122: I think we should do lexing to parse the equation. We have some lexing functions here for the json:

https://github.com/ROCm/AMDMIGraphX/blob/develop/src/convert_to_json.cpp#L58

We could move this to a header file so we can reuse that to lex the einsum. We could also refactor it to return `std::vector<string_view>` instead of using the `token` typedef. A `lex_equal` function could be added to do comparisons:

```cpp
auto lex_equal(std::string s)
{
    return [=](const char* start, const char* last) {
        auto n = last - start;
        if (n < s.size())
            return start;
        if (std::equal(start, start+s.size(), s))
            return start+s.size();
        return start;
    };
}

auto einsum_tokenize(const std::string& s)
{
    std::vector<lexer> lexers;

    lexers.push_back(lex_while(&isspace));
    lexers.push_back(lex_while(&isalpha));
    lexers.push_back(lex_equal("->"));
    lexers.push_back(lex_equal("..."));
    lexers.push_back(lex_equal(","));

    return tokenize(s.data(), s.data() + s.length(), lexers);
}
```

This should make it simpler to loop over the terms.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+        terms = std::move(input_terms);
+        terms.emplace_back(std::move(output_term));
+        for(const auto [l, _] : label_count)
+            unique_labels += l;
+
+        return ret;
+    }
+
+    std::tuple<string_vec, std::string, char_int_map, bool>
```
pfultz2, line 111: Use a struct for this.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+            last_op     = cur_op;
+            cur_pair[0] = cur_pair[1];
+        }
+
+        return finalize_output(info, cur_op, map_mat, cur_pair);
+    }
+
+    // Equation Parsing
+
+    std::tuple<string_vec, std::string, size_t>
```
pfultz2, line 89: Use a struct for this.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+namespace onnx {
+
+struct parse_einsum : op_parser<parse_einsum>
+{
+    using string_vec = std::vector<std::string>;
+    using int_vec    = std::vector<int>;
+    using int_mat    = std::vector<std::vector<int>>;
+    template <typename T>
+    using char_map     = std::map<char, T>;
+    using char_int_map = char_map<int>;
```
pfultz2, line 43: Remove all these typedefs except for the `int_mat`. It makes clearer seeing what the actual types are.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+            for(auto i = 0; i < term.size(); ++i)
+                counts[term[i]].push_back(i);
+
+            for(auto it = counts.begin(); it != counts.end();)
+            {
+                if(it->second.size() < 2)
+                    it = counts.erase(it);
+                else
+                    ++it;
+            }
```
pfultz2, line 308: Use `migraphx::erase_if`:

https://github.com/ROCm/AMDMIGraphX/blob/develop/src/include/migraphx/erase.hpp#L53

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+            duplicates.push_back(counts);
+        }
+
+        return duplicates;
+    }
```
pfultz2, line 314: It seems like it would be more effecient if we use the `group_uniqe` and `std::sort` algorithm to find the duplicates, especially since you dont even need a map just the counts in sorted order.

```cpp
std::vector<std::vector<int>> find_duplicates(const string_vec& terms) const
{
    std::vector<std::vector<int>> result;
    std::transform(terms.begin(), terms.end(), std::back_inserter(result), [](auto term) {
        std::vector<int> counts;
        std::sort(term.begin(), term.end());
        group_unique(term.begin(), term.end(), [](auto start, auto last) {
            auto n = last-start;
            if (n < 2)
                return;
            counts.push_back(n);
        }, std::equal_to<>{});
        return counts;
    });
    return result;
}
```

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+                                   static_cast<ssize_t>(calc_dim(sum_axes, op2_lens))};
+
+        op1 = info.add_instruction(make_op("reshape", {{"dims", dims1}}), op1);
+        op2 = info.add_instruction(make_op("reshape", {{"dims", dims2}}), op2);
+        op2 = info.add_instruction(make_op("transpose", {{"permutation", {0, 2, 1}}}), op2);
+        instruction_ref op = info.add_instruction(make_op("dot"), op1, op2);
+
+        std::vector<size_t> new_lens(op1_lens.size(), 1);
+        for(auto i = 0; i < new_lens.size() - sum_axes.size(); ++i)
+            new_lens[i] = std::max(op1_lens[i], op2_lens[i]);
```
pfultz2, line 500: Use `std::transform` here.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+        for(auto i = 0; i < cur_pair[1].size(); ++i)
+        {
+            if(cur_pair[1][i] == -1)
+                unsq_axes.push_back(i);
+            else
+                perm.push_back({cur_pair[1][i], i});
+        }
+        op = info.add_instruction(make_op("unsqueeze", {{"axes", unsq_axes}}), op);
+
+        std::sort(perm.begin(), perm.end(), [](auto lhs, auto rhs) {
```
pfultz2, line 553: Use `by(std::less<>{}, [](auto x) { return std::get<0>(x); })` instead for the compare op.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+        std::vector<std::tuple<int, int>> perm;
+
+        for(auto i = 0; i < cur_pair[1].size(); ++i)
+        {
+            if(cur_pair[1][i] == -1)
+                unsq_axes.push_back(i);
+            else
+                perm.push_back({cur_pair[1][i], i});
+        }
+        op = info.add_instruction(make_op("unsqueeze", {{"axes", unsq_axes}}), op);
```
pfultz2, line 551: Assign to a new variable: `auto unsqueeze = ...`.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+        {
+            if(cur_pair[1][i] == -1)
+                continue;
+
+            new_perm[std::get<1>(perm[p++])] = i;
+        }
+
+        op = apply_transpose_op(info, op, new_perm, cur_pair[1]);
+
+        return op;
```
pfultz2, line 570: Just do `return apply_transpose_op(info, op, new_perm, cur_pair[1])`

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+            }
+
+            op = apply_reduce_sum_op(info, op, red, cur_pair[1]);
+        }
+
+        return transpose_squeeze(info, cur_pair, op, map_mat.back());
+    }
+
+    instruction_ref unsqueeze_transpose(const onnx_parser::node_info& info,
+                                        int_mat& cur_pair,
```
pfultz2, line 538: Shouldn't this be `const`?

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+        return true;
+    }
+
+    instruction_ref apply_transpose_op(const onnx_parser::node_info& info,
+                                       instruction_ref op,
+                                       const int_vec& perm,
+                                       int_vec& row) const
+    {
+        if(is_transpose_identity(perm))
```
pfultz2, line 631: I dont think you need to check for this. We already have a simplification to remove noop transposes later.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+            {
+                rbc      = true;
+                rlens[l] = llens[l];
+            }
+        }
+
+        if(lbc)
+            opl = info.add_instruction(make_op("multibroadcast", {{"out_lens", llens}}), opl);
+        if(rbc)
+            opr = info.add_instruction(make_op("multibroadcast", {{"out_lens", rlens}}), opr);
```
pfultz2, line 677: Cant you use `insert_common_args` for this?

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+        return int_mat(cur_pair, int_vec(cols, fill_value));
+    }
+
+    int_vec extract_column(int_mat map_mat, int col_idx, int row_begin, int row_end) const
+    {
+        int_vec ret;
+        ret.reserve(row_end - row_begin);
+
+        for(int i = row_begin; i < row_end; ++i)
+            ret.push_back(map_mat[i][col_idx]);
```
pfultz2, line 711: This can be written using `std::transform`:

```cpp
std::transform(map_mat.begin()+row_begin, map_mat.begin()+row_end, std::back_inserter(ret), [&](auto& x) {
    return x[col_idx];
});
```

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+    // Equivalent to numpy.arange without the step parameter
+    int_vec arange(int start_value, int end_value) const
+    {
+        int_vec ret(end_value - start_value);
+        std::iota(ret.begin(), ret.end(), start_value);
+        return ret;
+    }
+
+    template <typename... Vecs>
```
pfultz2, line 741: Use `class`.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+    std::decay_t<std::tuple_element_t<0, std::tuple<Vecs...>>> concat_vectors(Vecs&&... vecs) const
```
pfultz2, line 742: Take the first vector as a seperate parameter instead of using `tuple_element`:

```cpp
template<class Vector, class... Vectors>
Vector concat_vectors(Vector vec, Vectors&&... vecs)
{
    ...
    return vec;
}
```

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+    {
+        size_t reserve_size = 0u;
+        ([&](auto&& vec) { reserve_size += vec.size(); }(std::forward<Vecs>(vecs)), ...);
```
pfultz2, line 745: Use `each_args`.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+        std::decay_t<std::tuple_element_t<0, std::tuple<Vecs...>>> ret;
+        ret.reserve(reserve_size);
+
+        ([&](auto&& vec) { ret.insert(ret.end(), vec.begin(), vec.end()); }(
```
pfultz2, line 750: Use `each_args`.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+        int_vec new_perm(cur_pair[1].size());
+        std::iota(new_perm.begin(), new_perm.end(), 0);
+
+        for(auto i = 0, p = 0; i < row_output.size(); ++i)
+        {
+            if(row_output[i] == -1)
+                continue;
+
+            new_perm[i] = std::get<1>(perm[p++]);
+        }
```
pfultz2, line 602: Use `transform_if` for this.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+        int_vec new_perm(cur_pair[1].size());
+        std::iota(new_perm.begin(), new_perm.end(), 0);
+
+        for(auto i = 0, p = 0; i < cur_pair[1].size(); ++i)
+        {
+            if(cur_pair[1][i] == -1)
+                continue;
+
+            new_perm[std::get<1>(perm[p++])] = i;
```
pfultz2, line 565: So `std::get<1>(perm[p++])` should be the same value as `i`. What is this trying to do?

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: The naming here is misleading since we don't store the counts, we store the axes on which the duplicated labels occur. I renamed the map to `duplicated_axes` and added a comment explaining the function.

e.g. For equation "iikjj,jkj", the result is a vector containing the two following maps:
result[0]: {'i': [0, 1], 'j': [3, 4]}
result[1]: {'j': [0, 2]}

This is also implemented keeping in mind that in the future we might need to support multiple duplicates per input.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+        auto indices_arg = info.add_literal(
+            migraphx::literal{migraphx::shape{migraphx::shape::int64_type, indices_lens}, indices});
+
+        op = info.add_instruction(
+            migraphx::make_op("gathernd", {{"batch_dims", batch_axes.size()}}), op, indices_arg);
+
+        // compute output row
+        for(auto& r : cur_pair[1])
+            if(contains(axes, r))
```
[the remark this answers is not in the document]
mirza-halilcevic: Not sure how to make use of `any_of` here. Are you suggesting something like this `if(any_of(axes, [r](auto a){ return a == r; }))`?

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Ok, that makes sense. I removed the check.

src/onnx/parse_einsum.cpp
mirza-halilcevic: It is passed to `apply_transpose_op` which modifies it.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,766 @@
...
+                                       int_vec& row) const
+    {
+        if(is_transpose_identity(perm))
+            return op;
+
+        op = info.add_instruction(make_op("transpose", {{"permutation", perm}}), op);
+        // compute output row
+        auto cpy = row;
+        for(auto i = 0u; i < perm.size(); ++i)
+            row[i] = cpy[perm[i]];
```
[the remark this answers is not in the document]
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: The problem is that `transform_if` doesn't advance the output iterator unless the predicate is satisfied. Besides that, a separate counter is used to index `perm`.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Since `perm` is sorted before this, it won't be the same value.
The idea behind this entire `unsqueeze_transpose` function is to expand the input dimensions to match the number of unique labels in the entire equation. Then we permute the labels so they are in alphabetical order, which is what this part of the code does. This makes the logic afterward significantly easier.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

src/onnx/parse_einsum.cpp
mirza-halilcevic: Done.

randyh62 (approved):

test/onnx/gen_onnx.py
```diff
@@ -2009,6 +2009,751 @@ def dynamicquantizelinear_2d_test():
...
+    x = helper.make_tensor_value_info('x', TensorProto.FLOAT, [2, 3])
+    v = helper.make_tensor_value_info('v', TensorProto.FLOAT, [1, 3])
+    y = helper.make_tensor_value_info('y', TensorProto.FLOAT, [2, 1])
+
+    node = onnx.helper.make_node('Einsum',
+                                 inputs=['x', 'v'],
+                                 outputs=['y'],
+                                 equation='ij,kj->ik')
+
+    return ([node], [x, v], [y])
```
CharlieL7, line 2075: I think we discussed this during a meeting but I've forgotten what was said.
In my head a matrix-vector product is `A_ij v_j = C_i`, so that would be `ij,j->i` in einsum notation.

test/onnx/gen_onnx.py
```diff
@@ -2009,6 +2009,751 @@ def dynamicquantizelinear_2d_test():
...
+@onnx_test()
+def einsum_matrix_dot_product_test():
+    x1 = helper.make_tensor_value_info('x1', TensorProto.FLOAT, [2, 3])
+    x2 = helper.make_tensor_value_info('x2', TensorProto.FLOAT, [2, 3])
+    y = helper.make_tensor_value_info('y', TensorProto.FLOAT, [1])
+
+    node = onnx.helper.make_node('Einsum',
+                                 inputs=['x1', 'x2'],
+                                 outputs=['y'],
+                                 equation='ij,ij->')
```
CharlieL7, line 2115: Do we have tests without the `->` to make sure that works?

test/onnx/gen_onnx.py
```diff
@@ -2009,6 +2009,751 @@ def dynamicquantizelinear_2d_test():
...
+    x1 = helper.make_tensor_value_info('x1', TensorProto.FLOAT, [2, 3])
+    x2 = helper.make_tensor_value_info('x2', TensorProto.FLOAT, [2, 3])
+    y = helper.make_tensor_value_info('y', TensorProto.FLOAT, [2, 3])
+
+    node = onnx.helper.make_node('Einsum',
+                                 inputs=['x1', 'x2'],
+                                 outputs=['y'],
+                                 equation='ij,ij->ij')
+
+    return ([node], [x1, x2], [y])
```
CharlieL7, line 2131: How is this currently parsed? The simplest parsing would be an element-wise `mul`, but I don't think that's how it currently works. I'm guessing it's current using `dot` with something like: `C_ij = A_ij B_ij = A_ij1 B_j1i`?

CharlieL7 (commented):
Awesome work, I'm glad this parser handles all of these cases. I have a couple of questions. Once those are answers I will then approve.

test/onnx/gen_onnx.py
mirza-halilcevic: Good point, I changed the test case. The parser will unsqueeze the vector in that case so it will end up the same.

test/onnx/gen_onnx.py
mirza-halilcevic: Yes, we have a few:
- https://github.com/ROCm/AMDMIGraphX/blob/3dceee176d4beb3e5877f85a9772ef6ffca2ec93/test/onnx/gen_onnx.py#L2258
- https://github.com/ROCm/AMDMIGraphX/blob/3dceee176d4beb3e5877f85a9772ef6ffca2ec93/test/onnx/gen_onnx.py#L2271
- https://github.com/ROCm/AMDMIGraphX/blob/3dceee176d4beb3e5877f85a9772ef6ffca2ec93/test/onnx/gen_onnx.py#L2413
- https://github.com/ROCm/AMDMIGraphX/blob/3dceee176d4beb3e5877f85a9772ef6ffca2ec93/test/onnx/gen_onnx.py#L2427

test/onnx/gen_onnx.py
mirza-halilcevic: At the moment, we would do the following for inputs of shape `{2, 3}`:
```
@8 = ref::reshape[dims={6, -1, 1}](@6) -> float_type, {6, 1, 1}, {1, 1, 1}, target_id=0
@9 = ref::reshape[dims={6, -1, 1}](@7) -> float_type, {6, 1, 1}, {1, 1, 1}, target_id=0
@10 = ref::transpose[permutation={0, 2, 1}](@9) -> float_type, {6, 1, 1}, {1, 1, 1}, target_id=0
@11 = ref::dot(@8,@10) -> float_type, {6, 1, 1}, {1, 1, 1}, target_id=0
@12 = ref::reshape[dims={2, 3}](@11) -> float_type, {2, 3}, {3, 1}, target_id=0
```
So basically a batch matmul with 6 batches and 1x1 matrices.

CharlieL7: OK, that's fine. We should make an issue to track that this can be simplified by a later compiler pass to just a `mul` operator. If there are any other known simplifications that can be done at a later date we should also have those in an issue. This way we don't have to rediscover those simplifications later on.

CharlieL7 (approved):

test/onnx/gen_onnx.py
pfultz2: A more general approach I did here to convert skinny matrices to mul+reduce:

https://github.com/ROCm/AMDMIGraphX/compare/develop...skinny-dot-2-reduce

Although the dot -> mul directly should probably be done in simplify_algebra.

src/include/migraphx/lexing.hpp
```diff
@@ -0,0 +1,98 @@
...
+        auto n = end - start;
+        if(n < s.size())
+            return start;
+        if(std::equal(start, start + s.size(), s.data()))
+            return start + s.size();
+        return start;
+    };
+}
+
+inline std::vector<std::string_view>
```
pfultz2, line 67: The non-template function should be defined in a .cpp file.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,774 @@
...
+    // exactly once in the equation.
+    // If ellipsis are present in the left hand side of the equation, the ellipsis dimensions are
+    // set to the beginning of the output term.
+    std::string generate_output_term(const std::map<char, int>& label_count,
+                                     size_t ellipsis_ndim) const
+    {
+        std::string output_term = ellipsis_ndim == 0 ? "" : "*";
+        for(const auto [label, count] : label_count)
+            if(count == 1)
+                output_term += label;
```
pfultz2, line 264: Use `transform_accumulate` here.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,774 @@
...
+                                     const int_mat& map_mat,
+                                     size_t input_idx,
+                                     int_mat& cur_pair) const
+    {
+        if(not duplicates.empty())
+        {
+            std::vector<std::vector<int>> diag;
+            diag.reserve(duplicates.size());
+            for(const auto& [_, v] : duplicates)
+                diag.push_back(v);
```
pfultz2, line 355: Use `std::transform`.

src/onnx/parse_einsum.cpp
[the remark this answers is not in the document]
pfultz2: Ah I misread the loop. It should be `std::replace_if` instead.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,774 @@
...
+        op = info.add_instruction(
+            migraphx::make_op("gathernd", {{"batch_dims", batch_axes.size()}}), op, indices_arg);
+
+        // compute output row
+        for(auto& r : cur_pair[1])
+            if(contains(axes, r))
+                r = first_axis;
+
+        std::vector<int> to_remove(axes.begin() + 1, axes.end());
+        for(auto t : to_remove)
```
pfultz2, line 429: Just do `range(axes.begin() + 1, axes.end())` instead of constructing the intermediate vector.

src/onnx/parse_einsum.cpp
```diff
@@ -0,0 +1,774 @@
...
+        auto new_sum_axes   = arange(perm.size() - sum_axes.size(), perm.size());
+
+        auto common_labels = set_union(new_batch_axes, new_sum_axes);
+        std::tie(op1, op2) = apply_broadcast_op(info, op1, op2, common_labels);
+
+        auto op = batch_dot(info, cur_pair, op1, op2, new_batch_axes, new_sum_axes);
+
+        auto perm_cpy = perm;
+        for(auto i = 0; i < perm.size(); ++i)
+            perm[perm_cpy[i]] = i;
```
pfultz2, line 496: I think we have the `invert_permutation` function which will do this.

src/onnx/parse_einsum.cpp
pfultz2: It seems like you could do something like `auto new_perm = sort_permutation(perm, by(std::less<>{}, [](auto x) { return x; })`(or maybe `std::greater<>{}` needs to be used). And then you can extend the vector with `iota` to add the missing axes.

src/onnx/parse_einsum.cpp
pfultz2: I think `sort_permutation` can be used here as well.

src/onnx/parse_einsum.cpp
pfultz2: It also would be good to have a comment explaining what this function is trying to do as well.

mirza-halilcevic: I added some comments, hope they are helpful. Regarding `sort_permutation`, it is tricky to make use of it here since the unsqueezed dims have to remain at their original axis while the other dims have to be sorted. The unsqueezed dims represent labels missing from the input term and can appear anywhere, we can't just `iota` them at the end.

e.g. `2 -1 0 -1 1` has to become `0 -1 1 -1 2` where `-1` is an unsqueezed dim, so the permutation produced needs to be `4 1 0 3 2`.

pfultz2: So why not insert a transpose first before the unsqueeze? Then you dont need to compute the permutation after the unsqueeze.

pfultz2: Also, if you must do the transpose after the unsqueeze(which seems more complicated), you could add a `sort_if` function to do the sorting but skip over some elements(this will at least make the intention more clear):

```cpp
template<class Iterator, class Predicate>
void sort_if(Iterator first, Iterator last, Predicate pred) {
    std::vector<typename std::iterator_traits<Iterator>::value_type> temp;
    std::copy_if(first, last, std::back_inserter(temp), pred);
    std::sort(temp.begin(), temp.end());
    auto temp_start = temp.begin();
    std::for_each(first, last, [&](auto& elem) {
        if (pred(elem))
        {
            elem = std::move(*temp_start);
            temp_start++;
        }
    });
}
```

mirza-halilcevic: I refactored both `unsqueeze_transpose` and `transpose_squeeze` and renamed them to reflect the operation order (to `transpose_unsqueeze` and `squeeze_transpose`). Let me know if there's anything else that needs to be addressed.

pfultz2 (approved):

causten:
Windows CI is failing.  Seems like...
src\onnx\parse_einsum.cpp:516:21: error: use of undeclared identifier 'ssize_t'

mirza-halilcevic:
> Windows CI is failing. Seems like... src\onnx\parse_einsum.cpp:516:21: error: use of undeclared identifier 'ssize_t'

I replaced `ssize_t` with `int64_t`. Seems like it's still failing for some other reason.

PR #2773, opened 2024-02-14
Add filesystem utility functions

apwojcik:
It is extracted from #2469 - the functions for creating executable file names, shared object/dynamic library file names, etc. Additionally, it provides overloaded the `+` operator for concatenating `fs::path` with `std::string` and appending the filename with an additional extension.


src/compile_src.cpp
```diff
@@ -26,6 +26,7 @@
 #include <migraphx/tmp_dir.hpp>
 #include <migraphx/stringutils.hpp>
 #include <migraphx/errors.hpp>
+#include <migraphx/fileutils.hpp>
```
apwojcik, line 29: Needed for overloaded `+` operator.

test/fileutils.cpp
```diff
@@ -0,0 +1,86 @@
+/*
+* The MIT License (MIT)
+*
+* Copyright (c) 2015-2022 Advanced Micro Devices, Inc. All rights reserved.
```
umangyadav, line 4:
```suggestion
* Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
```

src/fileutils.cpp
```diff
@@ -0,0 +1,65 @@
...
+constexpr std::string_view library_prefix{""};
+constexpr std::string_view library_postfix{".dll"};
+constexpr std::string_view static_library_postfix{".lib"};
+constexpr std::string_view object_file_postfix{".obj"};
+#else
+constexpr std::string_view executable_postfix{""};
+constexpr std::string_view library_prefix{"lib"};
+constexpr std::string_view library_postfix{".so"};
+constexpr std::string_view static_library_postfix{".a"};
+constexpr std::string_view object_file_postfix{".o"};
```
pfultz2, line 41: These could just be `const char*`. It will make concatenating with `std::string` simpler.

src/include/migraphx/fileutils.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+MIGRAPHX_EXPORT fs::path make_executable_filename(std::string_view name);
+MIGRAPHX_EXPORT fs::path make_shared_object_filename(std::string_view name);
+MIGRAPHX_EXPORT fs::path make_object_file_filename(std::string_view name);
+MIGRAPHX_EXPORT fs::path make_static_library_filename(std::string_view name);
+
+inline std::string operator+(std::string_view l, const fs::path& r)
+{
+    return std::string{l} + r.string();
+}
```
pfultz2, line 42: Can we remove this overload?

src/include/migraphx/fileutils.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+inline fs::path append_extension(const fs::path& l, std::string_view r)
+{
+    return fs::path{l}.replace_extension(l.extension().string() + r.data());
+}
```
pfultz2, line 47: This could be move to the cpp file instead of inline.

test/fileutils.cpp
```diff
@@ -0,0 +1,92 @@
...
+#define MIGRAPHX_LIB_PREFIX
+#define MIGRAPHX_DYNAMIC_POSTFIX ".dll"
+#define MIGRAPHX_STATIC_POSTFIX ".lib"
+#define MIGRAPHX_OBJECT_POSTFIX ".obj"
+#else
+#define MIGRAPHX_EXE_POSTFIX
+#define MIGRAPHX_LIB_PREFIX "lib"
+#define MIGRAPHX_DYNAMIC_POSTFIX ".so"
+#define MIGRAPHX_STATIC_POSTFIX ".a"
+#define MIGRAPHX_OBJECT_POSTFIX ".o"
```
pfultz2, line 48: These can be define as constants instead of macros.

src/fileutils.cpp
apwojcik: `std::string` appending/concatenation with `std::string_view` is 25% faster than with bear `const char *` due to the internal representation of a memory in the `std::string_view.` However, I need to make a small change to use that benefit.

src/include/migraphx/fileutils.hpp
apwojcik: Can we keep it? Future PRs not yet submitted will change from `std::string` to `fs::path` for filenames (needed for unified path handling between Linux and Windows - `std::string` does not do the job). If the overload is missing, there will be many more changes, e.g., when logging from `MIGRAPHX_...("..." + filename);` to `... + filename.string()`. Furthermore, newer C++ standards have this overload implemented.

test/fileutils.cpp
apwojcik: The goal is to concatenate at the compilation time, not the runtime.
For example, the shared object file name on Linux is `libsomething.so`, and the same dynamic library on Windows is named `something.dll`. By defining those macros, I can do `MIGRAPHX_LIB_PREFIX "something" MIGRAPHX_DYNAMIC_POSTFIX` which will result in the respective library names.
When I define them as constants, e.g., as `const char *`, I need to concatenate the string at runtime. There is a way of concatenating `std::string_view` at compile time, but it requires variable templated structures. I don't see any benefit here.

apwojcik:
_MIGraphX Performance Tests / release / MIGraphX Performance (pull_request)_ is failing because of this:

Step 1/10 : ARG DOCKERBASE=rocm-migraphx:5.1
Step 2/10 : FROM $DOCKERBASE
pull access denied for rocm-migraphx, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
Error: Process completed with exit code 1.

test/fileutils.cpp
pfultz2: This is for tests, I dont see any benefit in concating at compile time.

test/fileutils.cpp
apwojcik: Done

apwojcik:
Once again MIGraphX Performance Tests failed because of this error
```
pull access denied for rocm-migraphx, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

test/fileutils.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+template <const std::string_view& S, const std::string_view&... R>
+struct join_strings_compile_time<S, R...>
+{
+    static constexpr std::string_view value =
+        join_strings_compile_time<S, join_strings_compile_time<R...>::value>::value;
+};
+
+template <const std::string_view&... Strings>
+static constexpr auto join_strings_v = join_strings_compile_time<Strings...>::value;
```
pfultz2, line 88: There is no need for all this. We can make the variables `std::string` and then just use the string concatenation for the tests.

src/fileutils.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+fs::path make_object_file_filename(std::string_view name)
+{
+    return std::string{name}.append(object_file_postfix);
+}
+
+fs::path make_static_library_filename(std::string_view name)
+{
+    return std::string{library_prefix}.append(name).append(static_library_postfix);
+}
```
pfultz2, line 62: These functions should probably take `std::string` as a parameter instead because a string can be moved directly(whereas we cant move into a `string_view`) and you already constructing a `std::string`.

src/include/migraphx/fileutils.hpp
pfultz2: Can you then add an overload for `operator+(const fs::path& r, std::string l)` as well?

src/include/migraphx/fileutils.hpp
```diff
@@ -0,0 +1,48 @@
...
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+MIGRAPHX_EXPORT fs::path make_executable_filename(std::string_view name);
+MIGRAPHX_EXPORT fs::path make_shared_object_filename(std::string_view name);
+MIGRAPHX_EXPORT fs::path make_object_file_filename(std::string_view name);
+MIGRAPHX_EXPORT fs::path make_static_library_filename(std::string_view name);
+MIGRAPHX_EXPORT fs::path append_extension(const fs::path& path, std::string_view ext);
+
+inline std::string operator+(std::string_view l, const fs::path& r)
```
pfultz2, line 40: Use `std::string` here so the string can be moved.

pfultz2 (approved):

PR #2775, opened 2024-02-15
Add vim syntax file for MIGX IR

CharlieL7:
* I created a syntax highlighting file for vim for the MIGX IR
* You can add the syntax file to your syntax directory for vim or neovim to install
* Here is an example of the highlighting:
2024-02-14-191219_1276x1381_scrot

### DIFF
--- a/tools/syntax/migx.vim
+++ b/tools/syntax/migx.vim
@@ -1,0 +1,30 @@
+" Vim syntax file
+" Language: MIGraphX Intermediate Representation
+" Current Maintainer: Charlie Lin (https://github.com/CharlieL7)
+" Previous Maintainer:
+" Last Change: 14 Feb 2024
+
+" quit when a syntax file was already loaded
+if exists("b:current_syntax")
+  finish
+endif
+
+syn keyword migx_keyword param
+syn keyword migx_keyword return
+syn keyword migx_keyword target_id
+syn match migx_ins_number "@\d\+"
+syn match migx_instruction "\s=\s\zs.\+\ze\[" nextgroup=migx_attributes
+syn match migx_attributes "\(\[.\{-}\]\)"
+syn match migx_output_type "\s->\s\zs.\{-}\ze," nextgroup=migx_output_dims
+syn match migx_output_dims "\s\zs{.\{-}}\ze,\starget_id"
+syn match migx_exec_time ":\s\zs.\{-}ms\ze"
+
+let b:current_syntax = "migx"
+
+hi def link migx_keyword NonText
+hi def link migx_ins_number Number
+hi def link migx_instruction Operator
+hi def link migx_attributes Comment
+hi def link migx_output_type Type
+hi def link migx_output_dims Normal
+hi def link migx_exec_time NonText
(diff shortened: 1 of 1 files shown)


tools/syntax/migx.vim
```diff
@@ -0,0 +1,30 @@
...
+syn match migx_output_type "\s->\s\zs.\{-}\ze," nextgroup=migx_output_dims
+syn match migx_output_dims "\s\zs{.\{-}}\ze,\starget_id"
+syn match migx_exec_time ":\s\zs.\{-}ms\ze"
+
+let b:current_syntax = "migx"
+
+hi def link migx_keyword NonText
+hi def link migx_ins_number Number
+hi def link migx_instruction Operator
+hi def link migx_attributes Comment
```
lakhinderwalia, line 27: migx_attributes isn't always present, from what I see: example, the mul operator:
```
@13514 = reshape[dims={-1}](@13513) -> int64_type, {1}, {1}, target_id=0
@13515 = mul(@11820,@11819) -> int64_type, {1}, {1}, target_id=0
```

lakhinderwalia (approved):
Thanks for doing this. Very useful.

tools/syntax/migx.vim
```diff
@@ -0,0 +1,30 @@
...
+" Current Maintainer: Charlie Lin (https://github.com/CharlieL7)
+" Previous Maintainer:
+" Last Change: 14 Feb 2024
+
+" quit when a syntax file was already loaded
+if exists("b:current_syntax")
+  finish
+endif
+
+syn keyword migx_keyword param
```
umangyadav, line 12: is literal covered ?

umangyadav (approved):
Thanks charlie, this is useful. 

I wonder if there a way to add "go to definition" kind of jumps using ctags or other tools.

umangyadav:
Also would it be possible to use this kind of highlighting inside vscode ? let's say i save a file as `gelu.mgx` can it be automatically highlighted ?

tools/syntax/migx.vim
CharlieL7: Let me make an addition for literal

CharlieL7:
> I wonder if there a way to add "go to definition" kind of jumps using ctags or other tools.

Yes, that's possible to do with ctags in vim. I might look into it later.

> Also would it be possible to use this kind of highlighting inside vscode ? let's say i save a file as gelu.mgx can it be automatically highlighted ?

Probably, I don't use vscode however. I know how to do the auto filetype detection for vim.

> migx_attributes isn't always present, from what I see: example, the mul operator:

Good catch, latest update should fix it.

bpickrel (commented):
I wonder if the boilerplate copyright header is required?

tools/syntax/migx.vim
```diff
@@ -0,0 +1,31 @@
+" Vim syntax file
+" Language: MIGraphX Intermediate Representation
+" Current Maintainer: Charlie Lin (https://github.com/CharlieL7)
+" Previous Maintainer:
+" Last Change: 15 Feb 2024
+
```
bpickrel, line 6: Is this intended to go into output files with a specific filename extension?  If so, you should describe a procedure here for registering this syntax file.  If it's just for use from the command line, then I suggest you add the following comment:

" To use from command line:  vim -u tools/syntax/migx.vim myfile

CharlieL7: I'm pretty sure it's the vim tradition to have the syntax files like this. The user can decide how they want to use the file or use filetype detection.

PR #2779, opened 2024-02-15
Unpack int4

umangyadav:
Depends on #2730


src/include/migraphx/op/unpack_int4.hpp
```diff
@@ -0,0 +1,102 @@
...
+        auto new_lens = in_shape.lens();
+        new_lens[axis] *= 2;
+        return {migraphx::shape::uint8_type, new_lens};
+    }
+
+    argument compute(const shape& output_shape, std::vector<argument> args) const
+    {
+        argument result{output_shape};
+        auto in_shape = args.front().get_shape();
+        result.visit([&](auto output) {
```
pfultz2, line 78: I dont think you need a visit for the output, since it should always be `uint8_t`.

test/ref/unpack_int4.cpp
```diff
@@ -0,0 +1,158 @@
...
+    migraphx::shape s{migraphx::shape::uint8_type, {4}, {1}};
+    auto l0  = mm->add_literal(migraphx::literal{s, {0x1A, 0x2B, 0x3C, 0x4D}});
+    auto l0b = mm->add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", {4, 4}}}), l0);
+    mm->add_instruction(migraphx::make_op("unpack_int4"), l0b);
+    p.compile(migraphx::make_target("ref"));
+    auto result = p.eval({}).back();
+    std::vector<uint8_t> results_vector(32);
+    result.visit([&](auto output) { results_vector.assign(output.begin(), output.end()); });
+    std::vector<uint8_t> gold{
+        static_cast<uint8_t>(0x0A), static_cast<uint8_t>(0x01), static_cast<uint8_t>(0x0B),
```
lakhinderwalia, line 86: Unless there is some specific compiler issue, there should be no need for all the static_cast<uint8_t> for individual elements.. the vector type is already specified as uint8_t. There are many such instances..

src/CMakeLists.txt
```diff
@@ -252,6 +252,7 @@ register_migraphx_ops(
     undefined
     unique
     unknown
+    unpack_int4
```
lakhinderwalia, line 255: Please add a URL in comments, if this operator is based on some external specification. Thanks.

lakhinderwalia (approved):

test/ref/unpack_int4.cpp
umangyadav: removed them

src/include/migraphx/op/unpack_int4.hpp
umangyadav: done.

src/CMakeLists.txt
umangyadav: This is not based on any external specification.

shivadbhavsar (approved):

pfultz2 (approved):

PR #2781, opened 2024-02-15
Remove same dynamic output shapes check

CharlieL7:
* Removes the output shape check for dynamic shapes because it can now change after the `simplify_dyn_ops` pass


umangyadav (approved):

bpickrel (approved):
This fixes a test that would have failed for my `resize_matcher` changes.

PR #2782, opened 2024-02-15
Enable mlir by default for gemms with small K

pfultz2:


umangyadav (approved):

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -555,9 +555,6 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
         return std::max(m1, m2);
     };
 
-    mlir_mode mode =
```
umangyadav, line 558: This ENV Var can be removed and also need to update docs

umangyadav (changes requested):

umangyadav (approved):

TedThemistokleous (approved):

TedThemistokleous:
Fix the compile errors otherwise this makes sense

PR #2785, opened 2024-02-16
Pointwise + Concat fusion

### LINKED ISSUE #2735: Fuse "convert + concat"
### Problem Description

This pattern is from SD-XL.
```
text_embeds = @param:text_embeds -> float_type, {1, 1280}, {1280, 1}, target_id=0: 0.0007269ms, 1%
@34 = load[offset=12544,end=15104](@1) -> half_type, {1, 1280}, {1280, 1}, target_id=0: 0.00091438ms, 1%
@35 = gpu::code_object[code_object=9032,symbol_name=convert_kernel,global=640,local=1024,](text_embeds,@34) -> half_type, {1, 1280}, {1280, 1}, target_id=0: 0.0199863ms, 1%

@52 = gpu::code_object[code_object=9224,symbol_name=concat_kernel,global=1408,local=1024,](@35,@50,@51) -> half_type, {1, 2816}, {2816, 1}, target_id=0: 0.0195763ms, 1%
```

@35 is not used anywhere else in the graph.  Therefore it can be fused with @52 which is concat. 

MIGraphX already has "fuse_concat" pass but it currently looks for concat that goes into pointwise op.  

https://github.com/ROCm/AMDMIGraphX/blob/4c4474b7882d315560cc386df74cac87a3d77f20/src/fuse_concat.cpp#L109

Additional matchers can be added in that pass to handle case mentioned in this issue.

umangyadav:
Fixes #2735


test/verify/test_pooling_add_concat_relu.cpp
```diff
@@ -28,23 +28,57 @@
...
+{
+    migraphx::program create_program() const { return create_concat_fusion_program(true); }
+};
+
+struct test_pooling_add_concat : verify_program<test_pooling_add_concat>
+{
+    migraphx::program create_program() const { return create_concat_fusion_program(false); }
+};
+
+struct test_add_sub_concat_slice_mul : verify_program<test_add_sub_concat_slice_mul>
```
pfultz2, line 65: This should go into a seperate cpp file.

umangyadav: Done.

src/fuse_concat.cpp
```diff
@@ -98,6 +104,54 @@ struct fused_concat
...
+                               return mpm.create_module("concat:" + pm->name(), *pm);
+                           }
+                           auto* pm = mpm.create_module("concat:noop" +
+                                                        std::to_string(get_noop_counter()));
+                           auto x   = pm->add_parameter("x0", shape{input->get_shape().type()});
+                           pm->add_return({x});
+                           return pm;
+                       });
+        auto* post_pm = mpm.create_module("noop:concat" + std::to_string(get_noop_counter()));
+        auto x        = post_pm->add_parameter("!x0", shape{concat_ins->get_shape().type()});
```
pfultz2, line 145: Why is this prefixed with a `!`? I would prefer not to have it start with a special character because this will become `_x0` in the C++.

umangyadav: I wasn't sure. I copied same logic from `pointwise_concat_pointwise` fusion.  I can remove it.

shivadbhavsar (approved):

src/fuse_concat.cpp
pfultz2: It uses `x0` in that pass.

pfultz2:
We should probably only do this input fusion if the number of noops is low. I would use something like `max_noops = min(1, concat_ins->inputs().size() / 4)` to calculate the max number of noops we allow.

src/fuse_concat.cpp
umangyadav: https://github.com/ROCm/AMDMIGraphX/blob/b1a24e5420ba49a08f8530cbb2e66264ae145d29/src/fuse_concat.cpp#L158

I am referring this line.

umangyadav:
> We should probably only do this input fusion if the number of noops is low. I would use something like `max_noops = min(1, concat_ins->inputs().size() / 4)` to calculate the max number of noops we allow.

This would simply mean allowing at max 1 no-op. 

Added that rule and a test.

pfultz2 (approved):

PR #2791, opened 2024-02-20
Bump MIGraphX version

umangyadav:
6.1 has branched out. Bump minor version of MIGraphX. I assume we haven't had any breaking API change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ include(ROCMSetupVersion)
 
 option(BUILD_DEV "Build for development purpose only" OFF)
 
-rocm_setup_version(VERSION 2.9.0)
+rocm_setup_version(VERSION 2.10.0)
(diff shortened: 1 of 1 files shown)


umangyadav:
> @umangyadav Will it be on windows?

This is just updating migraphx's version. It applies to windows too.

johnnynunez:
> > @umangyadav Will it be on windows?
> 
> This is just updating migraphx's version. It applies to windows too.



> > @umangyadav Will it be on windows?
> 
> This is just updating migraphx's version. It applies to windows too.

But on the AMD page there is only HIP for 5.7.1 on windows.

umangyadav:
> But on the AMD page there is only HIP for 5.7.1 on windows.

5.7.1 i assume is the rocm version. This PR is bumping migraphx's version which is independent from rocm version. MIGraphX version is to version MIGraphX's internal IR or operators.

causten (approved):

PR #2805, opened 2024-02-21
rewrite_gelu to use tanh_exp approximation for fp16

CharlieL7:
Replace the `gelu_erf` with the  `tanh_exp` approximation rather than the sigmoid approximation.


src/rewrite_gelu.cpp
```diff
@@ -33,74 +33,75 @@ namespace migraphx {
...
-    auto matcher() const { return match::gelu_erf(); }
+    auto matcher() const { return match::any_of(match::gelu_erf(), match::gelu_tanh()); }
 
     void apply(module& m, const match::matcher_result& r) const
     {
-        auto ins = r.result;
         auto x   = r.instructions["x"];
-        if(x->get_shape().type() != migraphx::shape::half_type)
+        auto input_type                              = x->get_shape().type();
+        std::set<decltype(input_type)> convert_types = {migraphx::shape::half_type};
```
bpickrel, line 78: Why do this comparison in such an indirect way?

src/rewrite_gelu.cpp
```diff
@@ -33,74 +33,75 @@ namespace migraphx {
...
+    double sqrt_2_rpi   = sqrt(M_2_PI);
+    auto sqrt_2_rpi_lit = m.add_literal(literal{shape{x->get_shape().type()}, {sqrt_2_rpi}});
+    auto fit_const      = m.add_literal(literal{shape{x->get_shape().type()}, {0.044715f}});
+    auto one            = m.add_literal(literal{shape{x->get_shape().type()}, {1.0f}});
+    auto xb             = insert_common_op(m, ins, make_op("mul"), {x, sqrt_2_rpi_lit});
+    auto a              = insert_common_op(m, ins, make_op("mul"), {xb, fit_const});
+    auto b              = m.insert_instruction(ins, make_op("mul"), a, x);
+    auto c              = m.insert_instruction(ins, make_op("mul"), b, x);
+    auto u              = m.insert_instruction(ins, make_op("add"), c, xb);
+    auto neg_u          = m.insert_instruction(ins, make_op("neg"), u);
```
lakhinderwalia, line 57: These two steps, 57 and 58 can be optimized out.. Please adjust the constants.. Please see, as an example: https://github.com/ROCm/composable_kernel/pull/1162/files

lakhinderwalia (changes requested):

src/rewrite_gelu.cpp
lakhinderwalia: I would be curious to know the performance improvements with reduced steps (I found the gain to be 5%), and also its comparison to the erf() version: I found the performance improvement to be approx. 15%. My tests were all in FP, however. Thanks.

src/rewrite_gelu.cpp
```diff
@@ -33,74 +33,75 @@ namespace migraphx {
...
- * A. Choudhury, ‘A simple approximation to the area under standard normal curve’, Mathematics and
- * Statistics, vol. 2, no. 3, pp. 147–149, 2014.
+ * Find tanhGELU blocks and replace them with a rearranged version that is less likely to overflow
+ * and is more performant.
  */
 struct find_tanh_fast_gelu
 {
     auto matcher() const { return match::gelu_tanh(); }
 
     void apply(module& m, const match::matcher_result& r) const
```
bpickrel, line 94: This struct and similar ones with a `matcher()` and `apply()` method seem to be set up for a polymorphic code design, without actually using polymorphism or any other kind of class interrelationship.  Is there a reason it's done this way?

bpickrel (approved):
This look OK to me.

src/rewrite_gelu.cpp
pfultz2: > method seem to be set up for a polymorphic code design, without actually using polymorphism

It uses polymorphism. 

> or any other kind of class interrelationship.

You mean a type-erased interface?

>  Is there a reason it's done this way?

Versus type-erasure? Because the matchers can be inlined and with CSE the matchers that check for the same things can be done only once.

src/rewrite_gelu.cpp
umangyadav, line 78: std::set<migraphx::shape::type_t>` does not work ?

src/rewrite_gelu.cpp
```diff
@@ -33,74 +33,75 @@ namespace migraphx {
...
+    auto b              = m.insert_instruction(ins, make_op("mul"), a, x);
+    auto c              = m.insert_instruction(ins, make_op("mul"), b, x);
+    auto u              = m.insert_instruction(ins, make_op("add"), c, xb);
+    auto neg_u          = m.insert_instruction(ins, make_op("neg"), u);
+    auto d              = m.insert_instruction(ins, make_op("sub"), neg_u, u);
+    auto emu            = m.insert_instruction(ins, make_op("exp"), d);
+    auto e              = insert_common_op(m, ins, make_op("add"), {one, emu});
+    auto cdf            = insert_common_op(m, ins, make_op("div"), {one, e});
+    auto y              = m.insert_instruction(ins, make_op("mul"), x, cdf);
+    m.replace_instruction(ins, y);
```
umangyadav, line 63: Wouldn't this need to be added with `0.5 * x`  ? 

 `GELU(x) ~= 0.5 * x * ( 1 + tanh( sqrt(2/M_PI) * (x + 0.044715 * x^3)))`
`GELU(x) ~= 0.5 * x + 0.5 * x * tanh(sqrt(2/M_PI) * (x + 0.044715 * x^3)))`

src/rewrite_gelu.cpp
```diff
@@ -33,74 +33,75 @@ namespace migraphx {
...
+    auto sqrt_2_rpi_lit = m.add_literal(literal{shape{x->get_shape().type()}, {sqrt_2_rpi}});
+    auto fit_const      = m.add_literal(literal{shape{x->get_shape().type()}, {0.044715f}});
+    auto one            = m.add_literal(literal{shape{x->get_shape().type()}, {1.0f}});
+    auto xb             = insert_common_op(m, ins, make_op("mul"), {x, sqrt_2_rpi_lit});
+    auto a              = insert_common_op(m, ins, make_op("mul"), {xb, fit_const});
+    auto b              = m.insert_instruction(ins, make_op("mul"), a, x);
+    auto c              = m.insert_instruction(ins, make_op("mul"), b, x);
+    auto u              = m.insert_instruction(ins, make_op("add"), c, xb);
+    auto neg_u          = m.insert_instruction(ins, make_op("neg"), u);
+    auto d              = m.insert_instruction(ins, make_op("sub"), neg_u, u);
```
umangyadav, line 58: factor of `-2` in in denominator of tanh exponential can be folded with `M_PI` 
`tanh(sqrt(2/M_PI) * (x + 0.044715 * x^3)))`

`2 / (1 + exp(-2 * sqrt(2/M_PI) * (x + 0.044715 * x^3)))`

`2 / (1 + exp(1.59576912161 * (x + 0.044715 * x^3)))`

pfultz2: This should already be const folded by propagate_constant.

umangyadav: I don't think so because it will be as 
`a = mul(x , c1)` where `c1 = sqrt(2/M_PI)` 
and then it would 
b = mul(a , fitting_const)
c = mul(b, x)
d = mul(c, x)

pfultz2: Ah I see, we might rewrite it simplify algebra to fold the constant though, but I am not sure. So it might be better to write it in the better order, or use a constant.

src/rewrite_gelu.cpp
```diff
@@ -33,74 +33,75 @@ namespace migraphx {
...
+        replace_with_tanh_exp_gelu(m, r);
     }
 };
 
 void rewrite_gelu::apply(module& m) const
 {
     if(fast_math)
     {
-        match::find_matches(m, find_gelu_erf{}, find_tanh_fast_gelu{});
+        match::find_matches(m, find_gelu_erf{});
```
lakhinderwalia, line 104: BTW, if we match fast_math clause, I would expect tanh_fast_gelu to be used..  I am getting mixed up here.. sorry.

CharlieL7: We have a flag `--disable-fast-math` that's for preventing replacing erf -> tanh_exp approx. `find_gelu_erf` by itself already matches both erf and the slower tanh approximations.

lakhinderwalia: My expectation would be that `disable-fast-math` should logically turn off the `fast_math` variable here.. so taking out the `find_tanh_fast_gelu()` from this matcher clause is very confusing to me, when it is listed under `if(fast_math)`.

src/rewrite_gelu.cpp
CharlieL7: No, the 1+tanh() -> exp() transformation handles it already.

src/rewrite_gelu.cpp
CharlieL7: Should work fine, I was lazy

src/rewrite_gelu.cpp
CharlieL7: I multiplied out the factor of 2. I did not do more performance tests though

umangyadav (approved):

src/rewrite_gelu.cpp
CharlieL7: I agree it's somewhat confusing. I changed L70 in rewrite_gelu since it was requested during an earlier meeting:
> auto matcher() const { return match::any_of(match::gelu_erf(), match::gelu_tanh()); }
We still want to replace the slower tanh approximations if found, because we don't lose any accuracy.

test/rewrite_gelu_test.cpp
```diff
@@ -106,17 +117,28 @@ TEST_CASE(non_bias_gelu)
...
-        auto sub = m2.add_instruction(migraphx::make_op("sub"), a, b);
-        auto l1  = m2.add_literal(migraphx::literal{s2, {1.702f}});
-        auto mul = add_common_op(m2, migraphx::make_op("mul"), {sub, l1});
-        auto sig = m2.add_instruction(migraphx::make_op("neg"), mul);
-        sig      = m2.add_instruction(migraphx::make_op("exp"), sig);
-        auto l2  = m2.add_literal(migraphx::literal{s2, {1.0f}});
-        sig      = add_common_op(m2, migraphx::make_op("add"), {sig, l2});
-        sig      = m2.add_instruction(migraphx::make_op("div"), sub, sig);
-        m2.add_return({sig});
+        using migraphx::literal;
```
umangyadav, line 120: Are there very tests with GELU ? with FP16 ?

src/rewrite_gelu.cpp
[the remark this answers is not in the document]
CharlieL7: Code residual from trying to match to multiple types. I figured that we'll expand to additional types so left the code there.

lakhinderwalia (approved):
Approving this PR. Thanks for optimizing the replacement logic... Please do address any testing questions from @umangyadav. Thanks.

lakhinderwalia:
BTW, the matchers in https://github.com/ROCm/AMDMIGraphX/blob/0eb1c1dc9ca820d7ff058dccb532f0134eeda467/src/include/migraphx/match/gelu_tanh.hpp#L45 should be consolidated. One looks for 0.5, the other looks for the constants inside tanh. Perhaps both could look for just the constants inside tanh. _Doesn't have to this PR, however.._ Thanks.

test/rewrite_gelu_test.cpp
CharlieL7: Yeah, there are verify tests `test_gelu`, `test_add_gelu`, `test_add_gelu_half`.

src/rewrite_gelu.cpp
```diff
@@ -33,74 +33,73 @@ namespace migraphx {
...
+    auto lit0           = m.add_literal(literal{shape{x->get_shape().type()}, {const0}});
+    auto lit1           = m.add_literal(literal{shape{x->get_shape().type()}, {const1}});
+    auto one            = m.add_literal(literal{shape{x->get_shape().type()}, {1.0}});
+    auto xb             = insert_common_op(m, ins, make_op("mul"), {x, lit1});
+    auto a              = m.insert_instruction(ins, make_op("mul"), x, xb);
+    auto b              = insert_common_op(m, ins, make_op("add"), {a, lit0});
+    auto u              = m.insert_instruction(ins, make_op("mul"), x, b);
+    auto emu            = m.insert_instruction(ins, make_op("exp"), u);
+    auto c              = insert_common_op(m, ins, make_op("add"), {one, emu});
+    auto cdf            = insert_common_op(m, ins, make_op("div"), {one, c});
```
lakhinderwalia, line 59: Please divide x by emu. Avoid line 59.

PR #2809, opened 2024-02-21
mul add transpose dot matcher

### LINKED ISSUE #2736: Rewrite reshapes to improve const folding of mul_add
In unet, there is a pattern like this:

```
@218 = hip::hip_copy_literal[id=main:@literal:902] -> half_type, {640, 1, 1}, {1, 1, 1}
@219 = hip::hip_copy_literal[id=main:@literal:755] -> half_type, {640, 640}, {640, 1}
@220 = hip::hip_copy_literal[id=main:@literal:784] -> half_type, {640, 1, 1}, {1, 1, 1}
@224 = multibroadcast[out_lens={2, 640, 64, 64},out_dyn_dims={}](@220) -> half_type, {2, 640, 64, 64}, {0, 1, 0, 0}
@225 = reshape_lazy[dims={2, 640, 64, 64}](@223) -> half_type, {2, 640, 64, 64}, {2621440, 4096, 64, 1}
@226 = load[offset=62969600,end=73455360](@1) -> half_type, {2, 640, 64, 64}, {2621440, 4096, 64, 1}
@227 = multibroadcast[out_lens={2, 640, 64, 64},out_dyn_dims={}](@218) -> half_type, {2, 640, 64, 64}, {0, 1, 0, 0}
@228 = gpu::code_object[code_object=9352,symbol_name=mul_add_kernel,global=655360,local=1024,](@225,@227,@224,@226) -> half_type, {2, 640, 64, 64}, {2621440, 4096, 64, 1}
@229 = transpose[permutation={0, 2, 3, 1}](@228) -> half_type, {2, 64, 64, 640}, {2621440, 64, 1, 4096}
@230 = reshape_lazy[dims={2, 4096, 640}](@229) -> half_type, {2, 4096, 640}, {2621440, 1, 4096}
@231 = multibroadcast[out_lens={2, 640, 640},out_dyn_dims={}](@219) -> half_type, {2, 640, 640}, {0, 640, 1}
@232 = load[offset=140435200,end=150920960](@1) -> half_type, {2, 4096, 640}, {2621440, 640, 1}
@233 = gpu::gemm[alpha=1,beta=0,compute_fp32=1,trans_batch=0,solution_idx=0](@230,@231,@232) -> half_type, {2, 4096, 640}, {2621440, 640, 1}
```

We already have a pass to do the rewriting needed to const fold it(see [`find_dot_add`](https://github.com/ROCm/AMDMIGraphX/blob/4c4474b7882d315560cc386df74cac87a3d77f20/src/simplify_algebra.cpp#L366) and [`find_dot_mul`](https://github.com/ROCm/AMDMIGraphX/blob/4c4474b7882d315560cc386df74cac87a3d77f20/src/simplify_algebra.cpp#L269)), but it doesnt work with these extra operators between them and the gemm. To improve it we can:

- In simplify_reshapes, move the transpose+reshape to the input of multiply, and then adjust the broadcast for the constants to use the correct axis.
- We will still have an extra `multibroadcast` to broadcast the batch dimensions. This probably can't be removed so we will need adjust [`find_dot_add`](https://github.com/ROCm/AMDMIGraphX/blob/4c4474b7882d315560cc386df74cac87a3d77f20/src/simplify_algebra.cpp#L366) and [`find_dot_mul`](https://github.com/ROCm/AMDMIGraphX/blob/4c4474b7882d315560cc386df74cac87a3d77f20/src/simplify_algebra.cpp#L269) to account for an extra broadcast.

shivadbhavsar:
This partially resolves #2736.

It turns out that just moving the transpose gives a sizable perf boost for the sdxl unet model. Doing the same for the reshape op as indicated in that issue actually causes a perf degradation instead and so it will need to be investigated further.


pfultz2:
> Doing the same for the reshape op as indicated in that issue actually causes a perf degradation

Are you preserving the broadcasted shape when you move the reshape?

pfultz2:
Also, I dont know if this is good to do in general. The pointwise/reduce fusions can fuse around a reshape but it doesnt work with a `transpose` so if we just move the transpose but we dont do const folding of the operators then the transpose could prevent us from fusing the operators.

shivadbhavsar:
> > Doing the same for the reshape op as indicated in that issue actually causes a perf degradation
> 
> Are you preserving the broadcasted shape when you move the reshape?

What do you mean? 
If you mean preventing const prop from turning the broadcasted shape into a contiguous one, then ya I tested it with that change. In terms of the implementation of the matcher, it was exactly the same as this but looks for "reshape" op name rather than "transpose"

shivadbhavsar:
> Also, I dont know if this is good to do in general. The pointwise/reduce fusions can fuse around a reshape but it doesnt work with a `transpose` so if we just move the transpose but we dont do const folding of the operators then the transpose could prevent us from fusing the operators.

I see, is there anything additional I can look for in the matcher to prevent interfering with fusions? Also, I have a PR im working on that takes care of the reshape for this case. We can evaluate if this makes sense alongside that change once I have it up.

pfultz2:
> If you mean preventing const prop from turning the broadcasted shape into a contiguous one, then ya I tested it with that change.

You tested with which change? And what was the result?

Because if you move the reshape to the inputs of the `mul` or `add` then it will be `reshape(multibroadcast(...))` but reshape doesnt propagate the broadcast by default. So we need to rewrite it in a way where it can be `multibroadcast(reshape(...))`(maybe that could be another matcher).

> I see, is there anything additional I can look for in the matcher to prevent interfering with fusions?

Probably should only look for a `mul` or `add` and check that the `mul` is broadcasting across a matrix dimension. 

> Also, I have a PR im working on that takes care of the reshape for this case. We can evaluate if this makes sense alongside that change once I have it up.

I would assume we would do this in the same matcher.

shivadbhavsar:
> You tested with which change? And what was the result?
> 
> Because if you move the reshape to the inputs of the `mul` or `add` then it will be `reshape(multibroadcast(...))` but reshape doesnt propagate the broadcast by default. So we need to rewrite it in a way where it can be `multibroadcast(reshape(...))`(maybe that could be another matcher).

The results where much worse after adding the pass for moving the reshape in the same way. Writing that matcher for switching order of reshape and multibroadcast is but complicated for some cases, so for testing I just modified the const prop pass to not simplify the `const -> multibroadcast -> reshape` pattern. 


> Probably should only look for a `mul` or `add` and check that the `mul` is broadcasting across a matrix dimension.

Sure thats easy enough to do, my concern was that I didnt want to have the transpose end up in the middle of a series of pointwise ops. Wouldn't that also prevent fusions?


> I would assume we would do this in the same matcher.

I separated it out because the version parsed directly from torch doesn't even have a reshape in this spot and so that only needs the transpose moved. Also, this reshape is packing some batch dims I was planning to handle that by the matcher in #2823.

pfultz2:
> The results where much worse after adding the pass for moving the reshape in the same way. Writing that matcher for switching order of reshape and multibroadcast is a bit complicated for some cases

Yea we probably need a more general solution for this. I have an idea for a data structure we can use to help with this.

> so for testing I just modified the const prop pass to not simplify the const -> multibroadcast -> reshape pattern.

That helps the const propagation for the `mul` input, but it doesnt help when const folding the `mul` into the `B` matrix though.

So the B matrix is broadcasted across the batch, which means this operator will use a non-batched gemm which is generally faster. So when we const fold the mul we might be using non-broadcasted constant for the B matrix. If the constant to `mul` isnt a `multibroadcast` then `find_inner_broadcast` wont be used, which is needed to preserve the broadcast across the batch.

> Sure thats easy enough to do, my concern was that I didnt want to have the transpose end up in the middle of a series of pointwise ops. Wouldn't that also prevent fusions?

But the extra checks improves the likeliness that this can be const-folded.

> I separated it out because the version parsed directly from torch doesn't even have a reshape in this spot and so that only needs the transpose moved.

So does the torch version const fold the mul_add?

> Also, this reshape is packing some batch dims I was planning to handle that by the matcher in https://github.com/ROCm/AMDMIGraphX/pull/2823.

So using #2823, I assume the `mul_add` gets const folded, but there is a perf drop.

src/simplify_reshapes.cpp
```diff
@@ -949,6 +949,43 @@ struct find_reshape_reshape_dot
...
+{
+    auto matcher() const
+    {
+        auto reshapes = reshaper_names();
+        reshapes.insert("transpose");
+        auto const_pw = match::pointwise()(
+            match::either_arg(0, 1)(match::is_constant().bind("const"), match::any().bind("inp")));
+        auto match_shape_op = match::name("transpose")(match::args(const_pw.bind("pw")));
+        auto skip_pointwise_reshape_outputs =
+            match::skip_output(match::any_of(match::name(reshapes), match::pointwise()));
```
umangyadav, line 964: Logical question about `skip_output` does it only skip one time or one op or can it skip multiple ops ?

e.g. 
if 
let's say graph is `mul --> add -- sub --> transpose --> dot` then after two applications of this matcher it would become 

`mul --> transpose --> add --> sub --> dot` 

can `skip_output` skip two pointwises and match with `dot` in this case ?

shivadbhavsar:
> That helps the const propagation for the `mul` input, but it doesnt help when const folding the `mul` into the `B` matrix though.
> 
> So the B matrix is broadcasted across the batch, which means this operator will use a non-batched gemm which is generally faster. So when we const fold the mul we might be using non-broadcasted constant for the B matrix. If the constant to `mul` isnt a `multibroadcast` then `find_inner_broadcast` wont be used, which is needed to preserve the broadcast across the batch.
> 
> But the extra checks improves the likeliness that this can be const-folded.
> 
> So does the torch version const fold the mul_add?
> 
> So using #2823, I assume the `mul_add` gets const folded, but there is a perf drop.


Ok I see, that all makes sense. Im going to run and re-capture a few traces and perf numbers with and without these passes and then we can talk about it further. Issue is I switched MLIR branch in the middle of all this work which is making it hard to compare the results I have atm. I'll get back to you on this.

src/simplify_reshapes.cpp
```diff
@@ -949,6 +949,43 @@ struct find_reshape_reshape_dot
...
+// eg. mul -> add -> transpose -> dot becomes transpose -> mul -> add -> dot
+struct find_pointwise_transpose_dot
+{
+    auto matcher() const
+    {
+        auto reshapes = reshaper_names();
+        reshapes.insert("transpose");
+        auto const_pw = match::pointwise()(
+            match::either_arg(0, 1)(match::is_constant().bind("const"), match::any().bind("inp")));
+        auto match_shape_op = match::name("transpose")(match::args(const_pw.bind("pw")));
```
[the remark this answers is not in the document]
shivadbhavsar: Initially i was using this to also move reshapes so I kept it generic, but I'll update it

src/simplify_reshapes.cpp
```diff
@@ -949,6 +949,43 @@ struct find_reshape_reshape_dot
...
+        auto pw_op     = pw->get_operator();
+        auto new_inp   = m.insert_instruction(trans, shape_op, inp);
+        auto new_const = m.insert_instruction(trans, shape_op, constant);
+
+        m.replace_instruction(trans, pw_op, new_inp, new_const);
+
+        // Check if matcher is applicable to moved reshape
+        auto new_r = match::match_instruction(m, new_inp, matcher());
+        if(new_r.result != m.end())
+            apply(m, new_r);
```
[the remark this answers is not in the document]
shivadbhavsar: Ya that was my thought too but in theory there could be more than 4 consecutive pw ops

src/simplify_reshapes.cpp
shivadbhavsar: Thats a good point actually, my code is assuming that it can skip multiple times. @pfultz2 is this true?

shivadbhavsar: actually it has to be working that way, otherwise the test case wouldn't pass. Test case only runs the pass once and the transpose moves before both mul and add

pfultz2: >  Test case only runs the pass once and the transpose moves before both mul and add

The pass itself is ran 4 times.

shivadbhavsar: Oh right thats true. But in any case what I meant is that it wouldn't match in the second iteration if `skip_ouput` didnt skip multiple instructions. Just want to make sure Im understanding that matcher correctly

umangyadav: > actually it has to be working that way, otherwise the test case wouldn't pass. Test case only runs the pass once and the transpose moves before both mul and add

which test case are you referring to ? test case i see only has two pointwise e.g mul and add

pfultz2: > Thats a good point actually, my code is assuming that it can skip multiple times. @pfultz2 is this true?

Yes this is true, its called recursively in the matcher:

https://github.com/ROCm/AMDMIGraphX/blob/f522bb9d41a3ee11124c02babbb510b6274ea41e/src/include/migraphx/matcher.hpp#L671

src/simplify_reshapes.cpp
```diff
@@ -1065,16 +1065,14 @@ struct find_reshape_reshape_dot
 // const folding simplifications
 struct find_mul_add_shape_op_dot
 {
-    std::string op_name;
-
     auto matcher() const
     {
-        auto const_mul_add = match::name("mul", "add")(
+        const std::unordered_set<std::string> shape_ops = {"transpose", "convert"};
```
bpickrel, line 1070: Could you add a comment as to what logic this is doing?  I'm trying to work it out just by reading the code, but functions such as `match::skip_output` are just so cryptic that I can't tell if this is a functional change or merely style.

src/simplify_reshapes.cpp
```diff
@@ -1121,6 +1121,36 @@ struct find_reshape_dot
     }
 };
 
+// Remove transposes and converts between mul/add -> dot so simplify_algebra can perform
+// const folding simplifications
+struct find_mul_add_shape_op_dot
+{
+    auto matcher() const
+    {
+        const std::unordered_set<std::string> shape_ops = {"transpose", "convert"};
```
pfultz2, line 1130: You could do `auto shape_ops = match::name("transpose", "convert")` instead.

src/simplify_reshapes.cpp
```diff
@@ -1121,6 +1121,36 @@ struct find_reshape_dot
...
+        auto const_mul_add                              = match::name("mul", "add")(
+            match::either_arg(0, 1)(match::is_constant().bind("const"), match::any().bind("inp")));
```
pfultz2, line 1132: Can you name it `input` instead of `inp`?

shivadbhavsar:
**Needs #3174 to be merged first**

pfultz2 (approved):

PR #2810, opened 2024-02-22
Fix resetting rounding mode for nearbyint

gyulaz-htec:
Rounding mode was not reset after nearbyint because of the `return` before ` fesetround(rounding_mode);`:
https://github.com/ROCm/AMDMIGraphX/blob/9d8dc8696b457efcdc578111f0963df3c0b85c38/src/include/migraphx/op/nearbyint.hpp#L39-L42

### DIFF
--- a/src/include/migraphx/op/nearbyint.hpp
+++ b/src/include/migraphx/op/nearbyint.hpp
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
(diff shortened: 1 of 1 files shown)


umangyadav (approved):

lakhinderwalia (approved):
Thanks. Am surprised the compiler didn't warn of this...

PR #2815, opened 2024-02-22
Use clang++ from HIP SDK on Windows as a C++ compiler

apwojcik:
The `c++` command for the C++ compiler is unavailable on Windows. In general, installed by default of any compiler on Windows is not the standard. We assume HIP SDK is installed in the system at least, so we can use that as the C++ compiler.
The installation directory of HIP SDK is provided through the HIP_PATH environment variable (the HIP SDK installer creates that variable).


pfultz2:
We should use whatever the `CMAKE_CXX_COMPILER` is set to by default on windows. I dont want to rely on setting environment variables to have tests work correctly.

apwojcik:
> We should use whatever the `CMAKE_CXX_COMPILER` is set to by default on windows. I dont want to rely on setting environment variables to have tests work correctly.

We cannot rely on build system configuration for the production package. The reason is that compilers on Windows may be installed in nonstandard locations (usually they are; for instance, we are installing compilers at `c:\optr\rocm` and the HIP SDK installer by default at `C:\Program Files\AMD\HIP_SDK`).
The HIP SDK on Windows reveals its location by setting the HIP_PATH environment variable. Users may install to an arbitrary directory (the default is just a suggestion). Most of the tools do that, including MSVC. Alternatively, they may provide Windows registry entries with relevant information. So, we cannot rely on CMAKE_CXX_COMPILER on Windows because it will not match the deployment system.

pfultz2:
> We cannot rely on build system configuration for the production package.

Yes of course, thats why we dont call to the compiler during the runtime, we use hipRTC for that(the option to use clang is for dev only).

> The reason is that compilers on Windows may be installed in nonstandard locations

Compilers maybe installed in nonstandard locations on linux as well(it could be installed with a source-based package manager like spack) or not installed at all since its a dev package not a runtime package. This is why we compile with hipRTC. 

> it is not only to pass tests.

I dont understand, the only place the `c++` compiler is used is in a unit test. Its not used anywhere else.

pfultz2 (approved):

PR #2816, opened 2024-02-22
Retry tmp_dir removal on failure

apwojcik:
On Windows, the first attempt to remove the directory might not be successful (primarily if a child process uses it, too). Therefore, this PR changes the `tmp_dir` class to retry removal up to 5 times.


src/tmp_dir.cpp
```diff
@@ -88,9 +88,24 @@ void tmp_dir::execute(const std::string& exe, const std::string& args) const
...
+    constexpr int MAX_RETRIES_COUNT = 5;
     if(not enabled(MIGRAPHX_DEBUG_SAVE_TEMP_DIR{}))
     {
-        fs::remove_all(this->path);
+        int count = 0;
+        while(count < MAX_RETRIES_COUNT)
+        {
+            try
+            {
+                fs::remove_all(this->path);
```
pfultz2, line 99: Wouldn't it be better to use the non-throwing version of this call?

src/tmp_dir.cpp
```diff
@@ -88,9 +88,24 @@ void tmp_dir::execute(const std::string& exe, const std::string& args) const
 
 tmp_dir::~tmp_dir()
 {
+    constexpr int MAX_RETRIES_COUNT = 5;
     if(not enabled(MIGRAPHX_DEBUG_SAVE_TEMP_DIR{}))
     {
-        fs::remove_all(this->path);
+        int count = 0;
+        while(count < MAX_RETRIES_COUNT)
```
pfultz2, line 95: This could be written as `for(auto count:range(max_retries_count))` instead.

src/tmp_dir.cpp
```diff
@@ -88,9 +88,24 @@ void tmp_dir::execute(const std::string& exe, const std::string& args) const
...
+        {
+            try
+            {
+                fs::remove_all(this->path);
+                break;
+            }
+            catch(const fs::filesystem_error& err)
+            {
+                std::cerr << err.what() << std::endl;
```
pfultz2, line 104: This should probably say `std::cerr << "Failed to remove " << path << ": " << err.what() << std::endl`.

pfultz2 (approved):

PR #2817, opened 2024-02-22
Fix reading cached kernels from Windows resource file

apwojcik:
The resources are in `migraphx_gpu.dll`. The call to `GetModuleHandle(nullptr)` returns the handle to the executable, not the DLL.


cmake/Embed.cmake
```diff
@@ -130,10 +130,20 @@ ${RC_FILE_MAPPING}
 namespace resource {
 std::string_view read(int id)
 {
-    HMODULE handle = GetModuleHandle(nullptr);
-    HRSRC rc = FindResource(handle, MAKEINTRESOURCE(id), MAKEINTRESOURCE(TEXTFILE));
-    HGLOBAL data = LoadResource(handle, rc);
-    return {static_cast<const char*>(LockResource(data)), SizeofResource(handle, rc)};
+    HMODULE handle = GetModuleHandle(\"migraphx_gpu.dll\");
```
pfultz2, line 133: We should not be hard-coding the name here. I would like to eventually move this module to rocm-cmake, but for now we just copy it between us and ck(which also makes this not possible either).

apwojcik: CK uses static libraries, so on Windows, the compiled resource file (if used) is a *.res file - it is not being linked to the static library. It should be installed separately. The change is correct if CK is meant to be used with GPU only. It is currently impossible to deduce the linked-in library name because it is not provided at this compilation stage.

pfultz2 (approved):

causten (approved):

PR #2821, opened 2024-02-22
Fix dyn broadcasting

CharlieL7:
* Fixes a bug in the previous PR #2629 and makes dynamic shape broadcasting more generalized.
* the bug: `dot` operator should not accept non-matching outer dimensions. The outer dimensions should have already been made the same by a previous `broadcast_for_dot` instruction.
* Updated the dynamic dimension broadcasting to use the intersection of the ranges rather than if one is the subset of the other. Makes it more logical to understand.
* Added a docstring for `make_bcast_shape()`

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -74,19 +74,19 @@ compute_broadcasted_dyn_dims(std::vector<shape::dynamic_dimension> dds0,
                        {
                            return b;
                        }
-                       else if(a.within_range(b))
-                       {
-                           return shape::dynamic_dimension{a.min, a.max};
-                       }
-                       else if(b.within_range(a))
-                       {
-                           return shape::dynamic_dimension{b.min, b.max};
-                       }
                        else
                        {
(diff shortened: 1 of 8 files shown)


src/common.cpp
```diff
@@ -74,19 +74,19 @@ compute_broadcasted_dyn_dims(std::vector<shape::dynamic_dimension> dds0,
...
                        {
-                           MIGRAPHX_THROW("COMPUTE_BROADCASTED_DYN_DIMS: dynamic shapes {" +
-                                          migraphx::to_string_range(dds0) + "} and {" +
-                                          migraphx::to_string_range(dds1) + "} mismatch!");
+                           auto intersect = a.intersection(b);
+                           if(intersect.has_value())
+                           {
+                               return intersect.value();
+                           }
+                           else
```
lakhinderwalia, line 84: Nitpicking. The else clause is not required.

lakhinderwalia (approved):

umangyadav (approved):

PR #2823, opened 2024-02-23
Prevent collapsing batch dims in dot ops with constants

shivadbhavsar:
This simplifies many `reshape -> dot -> reshape` patterns that are not handled by the `find_reshape_reshape_dot` pass (ie. in gemms where one input is a constant).

This also simplifies the reshape found in #2736


shivadbhavsar:
SDXL Pref results for reference:

Torch-MIGraphX (end to end):
Before PR: 2850 ms
With PR: 2801 ms

ONNX Unet (4x attn trim):
Before PR: 5.54 ms
After PR:  5.52 ms

As expected, it doesnt affect the onnx version much because there is an extra convert in the middle. Once the convert is handled, the perf number reduces to 5.47ms.

src/simplify_reshapes.cpp
```diff
@@ -949,6 +981,76 @@ struct find_reshape_reshape_dot
...
 };
 
+struct find_reshape_const_dot
+{
+    auto matcher() const
+    {
+        return match::name("dot")(
+            match::used_once(),
+            match::either_arg(0, 1)(match::name("reshape").bind("rsp"),
+                                    match::skip_broadcasts(match::is_constant().bind("constant"))));
```
pfultz2, line 991: I dont think `skip_broadcasts` is needed when using `is_constant`. However, is it necessary to check for a const? Can we just match anything?

src/simplify_reshapes.cpp
```diff
@@ -900,6 +900,38 @@ struct find_scalar_multibroadcast_reshape_or_transpose
     }
 };
 
+// Remove unnecessary preceeding size 1 dims for constants
+struct find_const_multibroadcast
+{
+    auto matcher() const
+    {
+        return match::name("multibroadcast")(
+            match::arg(0)(match::is_constant()(match::used_once()).bind("constant")));
```
umangyadav, line 909: constant doesnot need to be used_once. `apply()` is not removing or replacing but rather inserting squeeze on it.

src/simplify_reshapes.cpp
```diff
@@ -949,6 +981,76 @@ struct find_reshape_reshape_dot
...
+            return;
+
+        auto rsp_lens = rsp->get_shape().lens();
+        auto inp      = rsp->inputs().front();
+        auto inp_lens = inp->get_shape().lens();
+
+        // Gemm axis should not be altered by the reshape
+        bool flipped    = rsp == dot->inputs().back();
+        size_t dot_axis = (flipped) ? -2 : -1;
+        if(rsp_lens.end()[dot_axis] != inp_lens.end()[dot_axis])
```
umangyadav, line 1011: I don't get what this doing. `end()[dot_axis]`  What does is it  doing ?

src/simplify_reshapes.cpp
```diff
@@ -900,6 +900,38 @@ struct find_scalar_multibroadcast_reshape_or_transpose
...
+        auto naxes = std::distance(const_lens.begin(), it);
+        if(naxes == 0)
+            return;
+
+        std::vector<std::size_t> sq_axes(naxes);
+        std::iota(sq_axes.begin(), sq_axes.end(), 0);
+
+        auto sq_const =
+            m.insert_instruction(mbr, make_op("squeeze", {{"axes", sq_axes}}), constant);
+        m.replace_instruction(mbr, mbr->get_operator(), sq_const);
```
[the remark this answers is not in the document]
shivadbhavsar: This is just removing any unnecessary preceding dims in literals eg.  {1, 1, 640, 640) which are later broadcasted to something like {2, 32, 640, 640}. Would broadcast work for this?  I thought it only does 1 axis

src/simplify_reshapes.cpp
shivadbhavsar: well, I want to bind "constant" to whatever was before the broadcast if there is a broadcast, if I dont do it I'll have to add a couple if statements in the apply method to check when the input is broadcasted vs not. 

I guess it doesnt have to be a const as long as any batch dimensions of the input are broadcasted. Otherwise, this logic for modifying the broadcast wouldnt work. Ie. for the gemm thats part of the attention, this does not work (for that scenario you would have to reshape the other input)

src/simplify_reshapes.cpp
shivadbhavsar: Thats true but in the unlikely case that somehow its used elsewhere without broadcast, it would duplicate the literal (there would be the original and a copy with the squeezed shape). I dont think this ever happens in reality though, so Im happy to remove that constraint

src/simplify_reshapes.cpp
shivadbhavsar: The goal here is to check that the gemm axis is unaffected by the reshape. So it A is the input and W is the weight, then for A*W, the last axis of A shouldnt be manipulated by the reshape, and if its W*A, then the second last axis should be untouched.

src/simplify_reshapes.cpp
pfultz2: I dont see a reason for it to be a constant. As long as we are expanding the batch dimension there shouldn't be a problem. The only case is that we would need to rewrite `reshape(multibroadcast(...))` if we are doing calculations to this matrix so broadcasting would be preserved.

src/simplify_reshapes.cpp
umangyadav: I just didn't understand syntax. That's all.  
Also this should do some bound checking e.g. what is `end()[dot_axis]` can go out of bound for some weird case.

src/simplify_reshapes.cpp
```diff
@@ -949,6 +981,76 @@ struct find_reshape_reshape_dot
...
+            return;
+
+        std::vector<size_t> new_const_lens{inp_lens.begin(), inp_lens.end() - 2};
+        operation new_bc_op;
+
+        auto bc_const      = (flipped) ? dot->inputs().front() : dot->inputs().back();
+        auto bc_const_lens = bc_const->get_shape().lens();
+        new_const_lens.insert(new_const_lens.end(), bc_const_lens.end() - 2, bc_const_lens.end());
+
+        // if the orignal weight is one dimensional, look at the original broadcast
```
umangyadav, line 1021:
```suggestion
        // if the original weight is one dimensional, look at the original broadcast
```

src/simplify_reshapes.cpp
shivadbhavsar: Ya it doesnt have to be a constant, updated.

src/simplify_reshapes.cpp
shivadbhavsar: rsp_lens cant be out of bounds because dot op operands need to have dimns >= 2, but the input to that can definitely be and issue so added a check for that.

src/simplify_reshapes.cpp
```diff
@@ -900,26 +900,80 @@ struct find_scalar_multibroadcast_reshape_or_transpose
...
+            return;
+
+        std::vector<std::size_t> sq_axes(naxes);
+        std::iota(sq_axes.begin(), sq_axes.end(), 0);
+
+        auto sq_const =
+            m.insert_instruction(mbr, make_op("squeeze", {{"axes", sq_axes}}), constant);
+        m.replace_instruction(mbr, mbr->get_operator(), sq_const);
+    }
+};
```
pfultz2, line 932: Why is this needed?

src/simplify_reshapes.cpp
```diff
@@ -900,26 +900,80 @@ struct find_scalar_multibroadcast_reshape_or_transpose
...
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto convert = r.result;
+        auto rsp     = r.instructions["rsp"];
+
+        auto inp         = rsp->inputs().front();
+        auto new_convert = m.insert_instruction(rsp, convert->get_operator(), inp);
+        m.replace_instruction(convert, rsp->get_operator(), new_convert);
+    }
+};
```
pfultz2, line 953: First, we should look for any unary operator not just convert. Secondly we should move past all reshape/transpose/broadcast operators. Thirdly, we may want to move it to the input or the output depending on what is on the input or output. There are 3 cases I see:

* `pointwise/reduce/constant -> reshapes -> unary -> reshapes -> non-pointwise/reduce`: we should move the unary operator to the left(ie the inputs).
* `non-pointwise/reduce -> reshapes -> unary -> reshapes -> pointwise/reduce`: we should move the unary operator to the right(ie the output).
* `pointwise/reduce/constant -> reshapes -> unary -> reshapes -> pointwise/reduce`: we could move the unary to either side but I guess we could prefer the input which might be better if its an input.

pfultz2: This is implemented in #2958.

pfultz2: So I think this can be removed.

shivadbhavsar: Removed

src/simplify_reshapes.cpp
shivadbhavsar: removed matcher

Issue #1904, opened 2023-06-29
Enable ORT accuracy tests to verify int8


TedThemistokleous:
rocMLIR will be added to migraphx.  This will be for all the data types but this issue will be to ensure existing tests in DLM pass using int8

TedThemistokleous:
Seems like #2300 will aid in accuracy for these runs.

TedThemistokleous:
Reopenning, need to rerun testing with these changes + update ORT EP

TedThemistokleous:
Resnet50 runs, need to do further analysis to compare to fp16 through same pipeline. 

Plan of attack

- [ ] Add changes to DLM to run e2e pipeline with int8 <- has accuracy result using real dataset
- [ ] Storage for imagenet data set for DLM runs
- [ ] Add test for bert
- [ ] Add test for distilgpt

Added models will leverage existing e2e code, may need to write/borrow code for benchmark.py thats used in parity checks.

TedThemistokleous:
Doing this part of QA validation for the resnet50 pipeline added in onnxruntime-inference-examples

TedThemistokleous:
Got DLM changes for benchmark.py but failing off 6.0 sorting out issues with run scripts.

TedThemistokleous:
Tracked by JIRA: https://ontrack-internal.amd.com/browse/SWDEV-410597

TedThemistokleous:
Got a gpt2 run with int8 quant here.

```
quantized model saved to:./onnx_models/gpt2_1_int8_gpu.onnx
Size of quantized ONNX model(MB):157.34468364715576
Finished quantizing model: ./onnx_models/gpt2_1_int8_gpu.onnx
Run onnxruntime on gpt2 with input shape [1, 32]
[...]

PR #2826, opened 2024-02-23
Add pass to convert Uint8 to int8 across operators

TedThemistokleous:
~Related to #1904 ~

Originally from when we were parsing in dynamicquantizelinear but we've found other operators weren't handling inputs of uint8 correctly and thus we need to add a pass to properly handle conversion.

This is dependent on the more recent #2888 so that we can then handle MatmulInteger,ConvInteger and DynamicQuantizeLinear and uint8 during compile time rather than in the parser


pfultz2:
> Without this we attempt to get mlir to try to create int8 x uint8 which fail at compile.

Can you explain this in more detail?

TedThemistokleous:
> > Without this we attempt to get mlir to try to create int8 x uint8 which fail at compile.
> 
> Can you explain this in more detail?

When speaking with MLIR they said they don't support uint8 x int8 quant_dots along with just uint8 in general for their kernels. I was seeing failures when using models optimized by Onnxruntime and using int8. This is related to the drop in performance is we would fail to compile the model in onnxruntime and then default to ROCm EP and or CPU

pfultz2:
But why are we getting mixed types like this in the first place?

I am wondering if we should have a pass to fix this up instead of doing this in the frontend parser.

TedThemistokleous:
> But why are we getting mixed types like this in the first place?


Because the implimentation of dynamicquantizelinear breaks it up into separate instructions which hardcode uint8 as part of the output type for the zero point which is defined by the onnx standard. 

> I am wondering if we should have a pass to fix this up instead of doing this in the frontend parser.
Doing more tests on this seems to break backend onnx tests since there is an expectation that all zero point output will be uint8.  So it appears we should be doing a pass then if one of the inputs to a quantize linear is then a uint8_t type to add in the convert

src/onnx/parse_dynamicquantizelinear.cpp
```diff
@@ -139,10 +139,25 @@ struct parse_dynamicquantizelinear : op_parser<parse_dynamicquantizelinear>
         auto y_zero_point = info.add_instruction(
             migraphx::make_op("convert", {{"target_type", migraphx::shape::uint8_type}}), round);
 
+        // 2.5 Adjust so output is int8 instead of uint due to MLIR being unable to handle int8 *
```
lakhinderwalia, line 142: Question. Couldn't the previous step to round etc. be directly giving out int8? Thanks.

TedThemistokleous: No, unfortunetly the output from additional onnx tests as part of the standard dyanicQuantizelinear needs to always output uint8 here. That seems to be what's also failing in CI right now eg)

```

 ACTUAL: dtype('uint8')
 DESIRED: dtype('int8')

======================================================================
FAIL: test_dynamicquantizelinear_min_adjusted_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/workspace/AMDMIGraphX/build_602/test/py/venv-onnx/3.10/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 290, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/workspace/AMDMIGraphX/build_602/test/py/venv-onnx/3.10/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 467, in run
    self.assert_similar_outputs(
  File "/workspace/AMDMIGraphX/test/py/onnx_backend_test.py", line 51, in assert_similar_outputs
    np.testing.assert_equal(ref_outputs[i].dtype,
  File "/workspace/AMDMIGraphX/build_602/test/py/venv-onnx/3.10/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 425, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:

```

TedThemistokleous:
Moving to draft. Will craft better matcher after tomorrow morning's coffee. @lakhinderwalia  I see your point from Monday's meeting. Looks like we can just target and replace q_min/q_max as well as the convert at the end to adjust the data range without requiring the extra shift add and converts.

TedThemistokleous:
There appears to have been an bug in how we were parsing which was causing the failures in the verify tests when that datatype is moved from uint8_t to int8_t. Change has been mirrored in the parser tests

src/onnx/parse_dynamicquantizelinear.cpp
```diff
@@ -131,8 +134,8 @@ struct parse_dynamicquantizelinear : op_parser<parse_dynamicquantizelinear>
         // intermediate_zero_point = qmin - min(x) / y_scale
         auto q_min     = info.add_literal(migraphx::literal{migraphx::shape{x_type}, {x_min}});
         auto q_max     = info.add_literal(migraphx::literal{migraphx::shape{x_type}, {x_max}});
-        auto sub1      = info.add_common_op("sub", q_min, min_x);
-        auto interm_zp = info.add_common_op("div", sub1, y_scale);
+        auto div1      = info.add_common_op("div", min_x, y_scale);
```
lakhinderwalia, line 137: The variables x_min & min_x are very confusing.. (same for x_max & max_x).. One set  could be renamed to [t_max & t_min] the other to maybe [cur_max & cur_min].. ? Please choose some names you like..

TedThemistokleous: Done

src/onnx/parse_dynamicquantizelinear.cpp
```diff
@@ -110,29 +110,32 @@ struct parse_dynamicquantizelinear : op_parser<parse_dynamicquantizelinear>
...
         auto q_range =
             info.add_literal(migraphx::literal{migraphx::shape{x_type}, {x_max - x_min}});
 
+        std::vector<size_t> len_vec(x_reshaped->get_shape().lens().size());
+        std::iota(len_vec.begin(), len_vec.end(), 0);
+
         // maximum(0, max(x))
         auto max_x =
-            info.add_instruction(migraphx::make_op("reduce_max", {{"axes", {0}}}), x_reshaped);
+            info.add_instruction(migraphx::make_op("reduce_max", {{"axes", len_vec}}), x_reshaped);
```
lakhinderwalia, line 124: Could you please check. I think this was supposed to implement maximum (0, max()). And that is now lost. The 0 was where the len_vec has been provided, and it had perhaps a different intent.. Thanks.

lakhinderwalia: Agree, @pfultz2. I just don't see the outer max being implemented anywhere (referring to line 122: `// maximum(0, max(x))` )

pfultz2: You mean we should be doing `max(0, reduce_max(x))` not just `reduce_max`?

src/simplify_dynamicquantizelinear.cpp
```diff
@@ -0,0 +1,93 @@
...
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        /* Need to modify the uint8 min/max range as well as final convert to convert to int8 */
+        auto convert_op = r.instructions["convert"];
+
+        // Ops to get q_min/q_max quickly
+        auto round_op    = convert_op->inputs().at(0);
+        auto saturate_op = round_op->inputs().at(0);
+        auto q_min       = saturate_op->inputs().at(1);
+        auto q_max       = saturate_op->inputs().at(2);
```
shivadbhavsar, line 65: I think its better to match for these instructions rather than relying on them always being is this order. As it is, this will match with any quantizelinear where the zero point is a convert instruction, and then assume these set of instructions exist in this format. Lets not assume its from the dynamicquantizelinear op and check for it in the matcher

test/matcher.cpp
```diff
@@ -174,6 +174,118 @@ TEST_CASE(match_arg8)
...
+    migraphx::module mm;
+    auto one   = mm.add_literal(1.0);
+    auto two   = mm.add_literal(2.0);
+    auto one_f = mm.add_literal(1.0f);
+    auto two_f = mm.add_literal(2.0f);
+
+    auto sum   = mm.add_instruction(sum_op{}, one, two);
+    auto sum_f = mm.add_instruction(sum_op{}, one_f, two_f);
+    mm.add_instruction(pass_op{}, sum);
+    mm.add_instruction(pass_op{}, sum_f);
```
umangyadav, line 280: This may not be necessary 

Also you don't need to create a Sum operation for all these tests. 
You can just create a test like following 

```
TEST_CASE(match_type)
{
    migraphx::module mm;
    auto one   = mm.add_literal(1.0);
    auto two   = mm.add_literal(2.0);
    auto one_f = mm.add_literal(1.0f);
    auto m = mathc::has_type(migraphx::shape::float_type);
    auto r = find_match(mm, m);
    EXPECT(bool{r.result == one_f})
}
```

src/simplify_dynamicquantizelinear.cpp
TedThemistokleous: I'm checking based on the type being uint8_type though as part of the matcher though. Wouldn't that omit any convert->quantizelinear sequence which don't have the correct type?

TedThemistokleous: Ah fair. So I guess I'll search for the other ops in tandem then

src/onnx/parse_dynamicquantizelinear.cpp
TedThemistokleous: Added.

test/matcher.cpp
TedThemistokleous: The intent with this test was to find the op with the correct input but would have the same name as there's two ops named "sum". This was particularly important in my current use case as I wanted to target converts with a certain type

src/simplify_dynamicquantizelinear.cpp
TedThemistokleous: Added two more ops in the search for the matcher, the nearbyint and clip into the convert should be sufficient, unless you want me to check if the clip also has two inputs as well? 

I figured we know this is some kind of narrowing of data range is occuring with the additional aps and then the has_type() check should handle the type we want to look at. I don't believe the static quantization has the clip/nearbyint before a quantizelinear as well

src/onnx/parse_dynamicquantizelinear.cpp
```diff
@@ -98,41 +98,40 @@ struct parse_dynamicquantizelinear : op_parser<parse_dynamicquantizelinear>
...
         auto y_scale = info.add_common_op("div", sub0, q_range);
 
         // 2. Computing y_zero_point
         // intermediate_zero_point = qmin - min(x) / y_scale
-        auto q_min     = info.add_literal(migraphx::literal{migraphx::shape{x_type}, {x_min}});
-        auto q_max     = info.add_literal(migraphx::literal{migraphx::shape{x_type}, {x_max}});
-        auto sub1      = info.add_common_op("sub", q_min, min_x);
-        auto interm_zp = info.add_common_op("div", sub1, y_scale);
+        auto div1      = info.add_common_op("div", min_x, y_scale);
+        auto interm_zp = info.add_common_op("sub", q_min, div1);
```
lakhinderwalia, line 134: Shouldn't the Zero Point be also readjusted in simplify_dynamicquantizelinear.cpp?

lakhinderwalia: I think this is fine..

test/onnx/verify/dynamicquantizelinear_1d_test.cpp
```diff
@@ -79,3 +80,61 @@ TEST_CASE(dynamicquantizelinear_1d_max_adjusted_test)
...
+}
+
+TEST_CASE(dynamicquantizelinear_1d_max_adjusted_simplify_test)
+{
+    auto p   = migraphx::parse_onnx("dynamicquantizelinear_1d_test.onnx");
+    auto* mm = p.get_main_module();
+    migraphx::run_passes(*mm, {migraphx::simplify_dynamicquantizelinear{}});
+    p.compile(migraphx::make_target("ref"));
+
+    std::vector<float> data{-1.0, -2.1, -1.3, -2.5, -3.34, -4.0};
```
lakhinderwalia, line 120: Could we please verify the results of both: int_8 & uint_8 quantization..? From the same original graph: i.e. before modification, and post modification.. Here the gold vector is verifying the modification results. Thanks.

test/onnx/verify/dynamicquantizelinear_2d_test.cpp
```diff
@@ -52,3 +53,32 @@ TEST_CASE(dynamicquantizelinear_2d_test)
...
 }
+
+TEST_CASE(dynamicquantizelinear_2d_simplify_test)
+{
+    auto p   = migraphx::parse_onnx("dynamicquantizelinear_2d_test.onnx");
+    auto* mm = p.get_main_module();
+    migraphx::run_passes(*mm, {migraphx::simplify_dynamicquantizelinear{}});
+    p.compile(migraphx::make_target("ref"));
+
+    std::vector<float> data{1.0, 2.1, 1.3, 2.5, 3.34, 4.0, 1.5, 2.6, 3.9, 4.0, 3.0, 2.345};
```
lakhinderwalia, line 64: Please add a couple of negative values as well.. unless there is some other logic here..

test/onnx/verify/dynamicquantizelinear_1d_test.cpp
lakhinderwalia: Also, a couple of positive values to line 120, please

lakhinderwalia (approved):
Approving this PR. Please try to add a couple of tests that I have suggested, if you could. Thanks.

test/onnx/verify/dynamicquantizelinear_2d_test.cpp
TedThemistokleous: Thats fine, no harm in adding the additional test. the intent with these was to copy the non simplified test with the same input data, add the simplify pass  and get a result that's just shifted by -128 after all the quantization is done. It confirms that the same input + a pass shifts the data correctly and as expected.

src/onnx/parse_dynamicquantizelinear.cpp
```diff
@@ -98,41 +98,40 @@ struct parse_dynamicquantizelinear : op_parser<parse_dynamicquantizelinear>
...
         // 1. Computing y_scale
         // Note: currently, DynamicQuantizeLinear only has uint8 quantization:
-        const auto x_max = std::numeric_limits<uint8_t>::max();
-        const auto x_min = std::numeric_limits<uint8_t>::min();
-
-        auto q_range =
-            info.add_literal(migraphx::literal{migraphx::shape{x_type}, {x_max - x_min}});
+        const auto type_max = std::numeric_limits<uint8_t>::max();
+        const auto type_min = std::numeric_limits<uint8_t>::min();
+        std::vector<size_t> len_vec(x_shape.lens().size());
```
pfultz2, line 106: Can you name this `axes` since its used for the axes to the reducitons?

test/onnx/verify/dynamicquantizelinear_2d_test.cpp
lakhinderwalia: > It confirms that the same input + a pass shifts the data correctly and as expected.
Thanks, that is why I had suggested if you could in a single test verify two golden sets: one before the modification, and one after.

src/include/migraphx/matcher.hpp
```diff
@@ -689,6 +689,12 @@ inline auto var(std::string s)
         });
 }
 
+inline auto has_type(shape::type_t t)
+{
+    return make_basic_pred_matcher(
+        [=, m_t = t](instruction_ref ins) { return ins->get_shape().type() == m_t; });
```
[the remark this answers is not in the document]
TedThemistokleous: Done

src/onnx/parse_dynamicquantizelinear.cpp
TedThemistokleous: Done

test/onnx/verify/dynamicquantizelinear_2d_test.cpp
TedThemistokleous: Yup all good. I added these as a side by side and also compared scale values as well to verify that if we parse in the same graph, perform a pass on one, the outputs should be shifted to the correct range, scales shouldn't change and zero point should be shifted to the proper value.

test/onnx/verify/dynamicquantizelinear_2d_test.cpp
```diff
@@ -52,3 +53,57 @@ TEST_CASE(dynamicquantizelinear_2d_test)
     std::vector<uint8_t> y_zpt_gold = {0};
     EXPECT(migraphx::verify::verify_rms_range(y_zpt, y_zpt_gold));
 }
+
+TEST_CASE(dynamicquantizelinear_2d_dot_simplify_test)
```
TedThemistokleous, line 57: @lakhinderwalia used the values you mentioned here.

src/simplify_dynamicquantizelinear.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return s;
+}
+
+/*
+ *  Dynamicquantizelinear by default adds uint8_t typed zero point into a quantize linear
+ *  which needs to converted to int8 in order to avoid uint8 x int8 operations or uint8 operations
+ *  from occuring on the backend as this isn't supported by MLIR nor how we simplify our quantizable
+ *  ops.
+ */
+struct match_find_dynamicquantizelinear_convert_int8_zp
```
umangyadav, line 51: @nives-vukovic  this could be helpful for your problem.

TedThemistokleous:
Updated things here for the failing UTs, turns out if it was flagging overflow with dot (y,y) with some test cases. 

Alternatively, It looks like I can't get this to update without removing the same_type() check in dot. @pfultz2 let me know if there's an alternative way we should handle this.

src/simplify_dynamicquantizelinear.cpp
TedThemistokleous: Its not a complete solution. Ive got changes built on this PR right now testing this out. It looks like the encoder models that are prequantized also have quant_dot/quant_convolution already added after the fact too so I'll need to add a check as well for them. 

This will be in a separate PR linked to @nives-vukovic issue. I've got one of the encoders reading in correctly, its just failing on compile now after I've added this updated pass.

src/include/migraphx/op/dot.hpp
```diff
@@ -39,7 +39,7 @@ struct dot
     std::string name() const { return "dot"; }
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this, true}.same_type().same_ndims().has(2);
+        check_shapes{inputs, *this, true}.same_ndims().has(2);
```
umangyadav, line 42: I think they both need to be same for GPU backend atleast. For the Ref we  can do it, but if we are going to allow it, we better test it. Can you explain why you made this change ?

test/onnx/parse/dynamicquantizelinear_simplfy_2d_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+        mm->add_instruction(migraphx::make_op("quantizelinear"), x, scale_y_bcast, y_pt_c_bcast);
+
+    auto sub = mm->add_instruction(migraphx::make_op("sub"), y_out, y_out);
+    mm->add_instruction(migraphx::make_op("dot"), y_out, sub);
+
+    auto prog = optimize_onnx("dynamicquantizelinear_2d_dot_test.onnx");
+    auto* m   = prog.get_main_module();
+    migraphx::run_passes(*m, {migraphx::simplify_dynamicquantizelinear{}});
+    p.debug_print();
+    p.debug_print();
```
umangyadav, line 82: remove these

test/onnx/parse/dynamicquantizelinear_simplfy_2d_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    auto y_out =
+        mm->add_instruction(migraphx::make_op("quantizelinear"), x, scale_y_bcast, y_pt_c_bcast);
+
+    auto sub = mm->add_instruction(migraphx::make_op("sub"), y_out, y_out);
+    mm->add_instruction(migraphx::make_op("dot"), y_out, sub);
+
+    auto prog = optimize_onnx("dynamicquantizelinear_2d_dot_test.onnx");
+    auto* m   = prog.get_main_module();
+    migraphx::run_passes(*m, {migraphx::simplify_dynamicquantizelinear{}});
```
umangyadav, line 80: You are testing results for the pass which should go under `test/simplify_dynamicquantizelinear_test.cpp`.
under `/test/onnx/parse` it should test whether parsing is correct ornot.

test/onnx/verify/dynamicquantizelinear_1d_test.cpp
```diff
@@ -79,3 +80,61 @@ TEST_CASE(dynamicquantizelinear_1d_max_adjusted_test)
     std::vector<uint8_t> y_zpt_gold = {255};
     EXPECT(migraphx::verify::verify_rms_range(y_zpt, y_zpt_gold));
 }
+
+TEST_CASE(dynamicquantizelinear_1d_dot_simplify_test)
+{
+    auto p   = migraphx::parse_onnx("dynamicquantizelinear_1d_dot_test.onnx");
+    auto* mm = p.get_main_module();
+    migraphx::run_passes(*mm, {migraphx::simplify_dynamicquantizelinear{}});
```
umangyadav, line 88: `dot` if it doesn't allow different types for the inputs then you should also put this pass inside `ref/target.cpp` directly.

TedThemistokleous: If I add this to ref target, it will break the other tests that expect the output of dynamicquantizelinear to be uint8 not int8.

test/onnx/parse/dynamicquantizelinear_simplfy_2d_test.cpp
TedThemistokleous: This was done to verify a pass after parse yields a predictable result as we don't support uint8 x int8 quantization.

test/onnx/verify/dynamicquantizelinear_1d_test.cpp
TedThemistokleous: Ive added back the same_types() check as part of dot. its required me to splice the change as part of the pass instead

test/onnx/parse/dynamicquantizelinear_simplfy_2d_test.cpp
TedThemistokleous: Removed

src/include/migraphx/op/dot.hpp
TedThemistokleous: It was due to be using add_instruction and replace_instruction as part of the pass. I've used inserts and moves instead to get the desired result for the 1-d and 2d cases now, and it appears that our quantizelinear op was incorrect as only the first two args (data, scale) need be the same time and zero point must be assigned to the desired output type as zero point and output match. 

My recent changes should have fixed this.

TedThemistokleous:
@pfultz2 you were right I needed to use insert here instead of add along with looking at the quantizelinear as well as part of the update. 

Removed any changes to our dot() but also found bug in how we handled input to quantizelinear. Updated parse test since I'm now adding convert on the input to either dot/conv prior to adding in the slice to avoid type errors which should get optimized out later once the types match up.

src/simplify_dynamicquantizelinear.cpp
```diff
@@ -0,0 +1,148 @@
...
+                        match::arg(2)(skip_broadcasts(
+                            match::name("convert")(
+                                match::has_type(migraphx::shape::uint8_type),
+                                match::arg(0)(
+                                    match::name("nearbyint")(
+                                        match::arg(0)(match::name("clip").bind("saturate")))
+                                        .bind("round")))
+                                .bind("convert"))))
+                        .bind("quant_lin"))))
+                .bind("target"));
```
pfultz2, line 69: Use variables/functions to improve the readability of the matcher.

src/simplify_dynamicquantizelinear.cpp
```diff
@@ -0,0 +1,148 @@
...
+ *  Dynamicquantizelinear by default adds uint8_t typed zero point into a quantize linear
+ *  which needs to converted to int8 in order to avoid uint8 x int8 operations or uint8 operations
+ *  from occuring on the backend as this isn't supported by MLIR nor how we simplify our quantizable
+ *  ops.
+ */
+struct match_find_dynamicquantizelinear_convert_int8_zp
+{
+    auto matcher() const
+    {
+        return match::any_arg(0, 1)(
```
pfultz2, line 55: This seems problematic. If the conv/dot is used more than once, we would create multiple conv/dots. Also, if its not a binary operator that follows it would fail also.

src/simplify_dynamicquantizelinear.cpp
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        auto new_target_op =
+            m.insert_instruction(target_op, target_op->get_operator(), converted_inputs);
+
+        auto data  = quant_op->inputs().at(0);
+        auto scale = quant_op->inputs().at(1);
+
+        m.replace_instruction(target_op, new_target_op);
+        m.remove_instruction(target_op);
```
pfultz2, line 128: Dont call `remove_instruction`. DCE will remove unused instructions. There can be problems with removing while iterating(such as infintie loops, etc). The DCE pass will do this correctly so we should just rely on it to remove the unused instructions.

TedThemistokleous: What alternative would you suggest here then? A replace will still call compute_shapes() and then propagate through and generate an issue with dot/conv on the same_type() check

src/simplify_dynamicquantizelinear.cpp
TedThemistokleous: Sure I can clean that up

TedThemistokleous:
Split out changes for parse_dynamcquantizelinear for this one. Retargeted branch to that PR until that changeset gets in. Will revisit one that's merged

causten:
@TedThemistokleous to review and resolve conflicts

TedThemistokleous:
DQL fix + Pass changes
```
Summary:
gpu::code_object::reduce_min_min_kernel: 2.04166ms / 49 = 0.0416666ms, 13%
gpu::code_object::reduce_max_max_sub_mul_kernel: 2.03947ms / 49 = 0.0416219ms, 13%
gpu::code_object::mul_quantizelinear_kernel: 1.66411ms / 48 = 0.0346689ms, 10%
gpu::code_object::mlir_quant_dot: 1.35967ms / 47 = 0.0289292ms, 9%
gpu::code_object::convert_kernel: 1.15707ms / 50 = 0.0231415ms, 7%
gpu::code_object::quantizelinear_convert_sub_quantizelinear_kernel: 1.15445ms / 49 = 0.0235602ms, 7%
gpu::code_object::div_neg_clip_nearbyint_kernel: 1.14034ms / 49 = 0.0232722ms, 7%
gpu::code_object::mul_kernel: 1.12784ms / 49 = 0.0230171ms, 7%
gpu::code_object::contiguous_kernel: 0.854442ms / 36 = 0.0237345ms, 6%
gpu::code_object::quantizelinear_kernel: 0.833104ms / 36 = 0.0231418ms, 5%
gpu::code_object::layernorm_mul_add_kernel: 0.580536ms / 24 = 0.024189ms, 4%
gpu::code_object::dequantizelinear_add_add_kernel: 0.534086ms / 23 = 0.0232211ms, 4%
gpu::code_object::mlir_quant_dot_dequantizelinear_add: 0.38627ms / 13 = 0.0297131ms, 3%
load: 0.316312ms / 537 = 0.000589036ms, 2%
gpu::code_object::dequantizelinear_mul_where_reduce_max_sub_exp_reduce_sum_div_quantizelinear_kernel: 0.293268ms / 12 = 0.024439ms, 2%
gpu::code_object::dequantizelinear_add_mul_mul_add_mul_exp_add_div_kernel: 0.287277ms / 12 = 0.0239397ms, 2%
gpu::code_object::mlir_quant_dot_dequantizelinear: 0.280463ms / 12 = 0.0233719ms, 2%
multibroadcast: 0.245326ms / 295 = 0.000831613ms, 2%
hip::hip_copy_literal: 0.0990334ms / 150 = 0.000660223ms, 1%
gpu::code_object::mlir_quant_dot_dequantizelinear_mul: 0.0969119ms / 1 = 0.0969119ms, 1%
reshape_lazy: 0.0966945ms / 131 = 0.000738126ms, 1%
transpose: 0.0604593ms / 48 = 0.00125957ms, 1%
slice: 0.04459ms / 36 = 0.00123861ms, 1%
gpu::code_object::add_layernorm_mul_add_kernel: 0.0248186ms / 1 = 0.0248186ms, 1%
gpu::code_object::dequantizelinear_add_kernel: 0.0238949ms / 1 = 0.0238949ms, 1%
gpu::code_object::gather_kernel: 0.0233956ms / 1 = 0.0233956ms, 1%
@param: 0.00996924ms / 26 = 0.000383432ms, 1%
hip::hip_allocate_memory: 0.00088522ms / 1 = 0.00088522ms, 1%
check_context::migraphx::gpu::context: 0.00072764ms / 1 = 0.00072764ms, 1%

Batch size: 1
Rate: 189.117 inferences/sec
Total time: 5.28774ms
Total instructions time: 16.7771ms
Overhead time: 0.371628ms, -11.4893ms
Overhead: 7%, -217%
[ MIGraphX Version: 2.10.0. ] Complete: bin/driver perf ../int8_models/gpt2_1_int8_gpu.onnx --input-dim @input_ids 1 32 --fill1 input_ids --disable-fast-math --int8


Summary:
gpu::code_object::reduce_max_max_sub_mul_kernel: 2.05055ms / 49 = 0.0418479ms, 13%
gpu::code_object::reduce_min_min_kernel: 2.04459ms / 49 = 0.0417264ms, 13%
gpu::code_object::mul_quantizelinear_kernel: 1.66972ms / 48 = 0.0347858ms, 10%
gpu::code_object::mlir_quant_dot: 1.36225ms / 47 = 0.028984ms, 9%
gpu::code_object::convert_kernel: 1.16309ms / 50 = 0.0232618ms, 7%
gpu::code_object::quantizelinear_convert_sub_quantizelinear_kernel: 1.16019ms / 49 = 0.0236774ms, 7%
gpu::code_object::div_neg_clip_nearbyint_kernel: 1.14452ms / 49 = 0.0233575ms, 7%
gpu::code_object::mul_kernel: 1.13133ms / 49 = 0.0230883ms, 7%
gpu::code_object::contiguous_kernel: 0.860616ms / 36 = 0.023906ms, 6%
gpu::code_object::quantizelinear_kernel: 0.833249ms / 36 = 0.0231458ms, 5%
gpu::code_object::layernorm_mul_add_kernel: 0.583202ms / 24 = 0.0243001ms, 4%
gpu::code_object::dequantizelinear_add_add_kernel: 0.537737ms / 23 = 0.0233799ms, 4%
gpu::code_object::mlir_quant_dot_dequantizelinear_add: 0.385738ms / 13 = 0.0296722ms, 3%
load: 0.314656ms / 537 = 0.000585952ms, 2%
gpu::code_object::dequantizelinear_add_pow_mul_add_mul_tanh_add_mul_mul_kernel: 0.311504ms / 12 = 0.0259587ms, 2%
gpu::code_object::dequantizelinear_mul_where_reduce_max_sub_exp_reduce_sum_div_quantizelinear_kernel: 0.295388ms / 12 = 0.0246157ms, 2%
gpu::code_object::mlir_quant_dot_dequantizelinear: 0.285528ms / 12 = 0.023794ms, 2%
multibroadcast: 0.257628ms / 295 = 0.000873314ms, 2%
gpu::code_object::mlir_quant_dot_dequantizelinear_mul: 0.100579ms / 1 = 0.100579ms, 1%
hip::hip_copy_literal: 0.0988059ms / 150 = 0.000658706ms, 1%
reshape_lazy: 0.0981314ms / 131 = 0.000749094ms, 1%
transpose: 0.0633442ms / 48 = 0.00131967ms, 1%
slice: 0.0449541ms / 36 = 0.00124872ms, 1%
gpu::code_object::add_layernorm_mul_add_kernel: 0.0250007ms / 1 = 0.0250007ms, 1%
gpu::code_object::dequantizelinear_add_kernel: 0.0242269ms / 1 = 0.0242269ms, 1%
gpu::code_object::gather_kernel: 0.0238603ms / 1 = 0.0238603ms, 1%
@param: 0.0102222ms / 26 = 0.000393161ms, 1%
check_context::migraphx::gpu::context: 0.0009512ms / 1 = 0.0009512ms, 1%
hip::hip_allocate_memory: 0.0009052ms / 1 = 0.0009052ms, 1%

Batch size: 1
Rate: 190.495 inferences/sec
Total time: 5.24948ms
Total instructions time: 16.8825ms
Overhead time: 0.37222ms, -11.633ms
Overhead: 7%, -222%
[ MIGraphX Version: 2.10.0. ] Complete: bin/driver perf ../int8_models/gpt2_1_int8_gpu.onnx --input-dim @input_ids 1 32 --fill1 input_ids --int8
```

TedThemistokleous:
Not required but can still be used as perf improvement now 2903 has been added. Will need to determine speedup after rebase

src/include/migraphx/op/quantizelinear.hpp
```diff
@@ -56,7 +56,7 @@ struct quantizelinear
 
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.same_dims().has(2, 3);
+        check_shapes{inputs, *this}.has(2, 3);
```
pfultz2, line 59: Why is this removed? They should have the same dimensions.

src/simplify_dynamicquantizelinear.cpp
pfultz2: You need to be replacing the dot/conv instead.

causten:
@TedThemistokleous  Is this PR still needed ?

Issue #2827, opened 2024-02-23
Fix for prefix_sum


TedThemistokleous:
Fixes issue found with prefix sum list initialization of single input variables found here: 
https://ontrack-internal.amd.com/browse/SWDEV-443633

Fix is in a branch and PR is open. Running test_verify locally in the container specified in the ticket before moving PR out of draft status

PR #2828, opened 2024-02-23
Remove list initialization for prefix_scan_sum

TedThemistokleous:
This seems to cause an error as seen in : https://ontrack-internal.amd.com/browse/SWDEV-443633

Issue with using list initialization for inputs in the compute shape function of prefix_scan_sum.

related to #2827


pfultz2:
I think it would be better to make the `shape(const std::vector<shape>& subs)` explicit, that might fix the issue which seems due to implicit conversions.

src/shape.cpp
```diff
@@ -95,7 +95,9 @@ struct shape_impl
         }
     }
 
-    shape_impl(const std::vector<shape>& subs) : m_type(shape::tuple_type), m_shapes(subs) {}
+    explicit shape_impl(const std::vector<shape>& subs) : m_type(shape::tuple_type), m_shapes(subs)
```
pfultz2, line 98: This shouldn't need explicit here, this is just a private class.

TedThemistokleous: Sure I'll change.

TedThemistokleous: Done

src/targets/gpu/include/migraphx/gpu/prefix_scan_sum.hpp
```diff
@@ -54,7 +54,7 @@ struct hip_prefix_scan_sum : oper<hip_prefix_scan_sum>
 
     shape compute_shape(const std::vector<shape>& inputs) const
     {
-        std::vector<shape> in_shapes{inputs};
+        std::vector<shape> in_shapes(inputs);
```
umangyadav, line 57: is this necessary change now that constructor is marked as explicit ?

TedThemistokleous: Changed it back to braced construction

pfultz2 (approved):

umangyadav (approved):

PR #2830, opened 2024-02-23
Fix split multiple same dyn_dim

CharlieL7:
* addresses previous PR review comments that were not completed yet

### DIFF
--- a/src/split_single_dyn_dim.cpp
+++ b/src/split_single_dyn_dim.cpp
@@ -49,10 +49,6 @@ struct dynamic_dimensions_check
 optional<std::vector<dynamic_dimensions_check>>
 has_one_unique_dyn_dim(const std::unordered_map<std::string, shape>& param_shapes)
 {
-    if(param_shapes.empty())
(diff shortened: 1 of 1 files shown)


lakhinderwalia (approved):

src/split_single_dyn_dim.cpp
```diff
@@ -64,19 +60,18 @@ has_one_unique_dyn_dim(const std::unordered_map<std::string, shape>& param_shape
...
             if(not dd.is_fixed())
             {
-                num_non_fixed += 1;
-                // catch more than one non-fixed dynamic_dimension
-                if(num_non_fixed > 1)
-                {
-                    return std::nullopt;
-                }
                 ret.push_back(dynamic_dimensions_check{param.first, dd});
+                return true;
```
pfultz2, line 67: Thats an interesting way to use `count_if`, not what I had originally in mind, but definitely works.

pfultz2 (approved):

PR #2832, opened 2024-02-24
Temporarily disable Navi in CI

causten:


TedThemistokleous:
Not sure whats going on here, continous CI says fail but its all green. 

Also do we want to be pulling coverage for the Navi32 on our other checks or is this related to other things we'e seen?

TedThemistokleous (approved):

PR #2841, opened 2024-02-27
Suppress compilation warning about result not read

apwojcik:
The official documentation allows us to use `std::ignore` for suppressing compilation warnings about return values not being read (even originally, it was designed to be used together with `std::tie`).
See more info here: https://en.cppreference.com/w/cpp/utility/tuple/ignore.

### DIFF
--- a/test/ref/resize.cpp
+++ b/test/ref/resize.cpp
@@ -421,7 +421,7 @@ TEST_CASE(resize_fail_test_4)
                                            {"coordinate_transformation_mode", "invalid"}}),
                         {a0});
     p.compile(migraphx::make_target("ref"));
-    EXPECT(test::throws([&] { std::ignore = p.eval({}); }));
+    EXPECT(test::throws([&] { p.eval({}); }));
 }
 
 TEST_CASE(resize_fail_test_5)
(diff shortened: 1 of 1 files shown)


pfultz2 (approved):

test/ref/resize.cpp
```diff
@@ -444,7 +444,7 @@ TEST_CASE(resize_fail_test_5)
                            {"coordinate_transformation_mode", "pytorch_half_pixel"}}),
         {a0});
     p.compile(migraphx::make_target("ref"));
-    EXPECT(test::throws([&] { p.eval({}).back(); }));
+    EXPECT(test::throws([&] { std::ignore = p.eval({}).back(); }));
```
pfultz2, line 447: Actually I think we can remove the call to `.back()` for this test. Its only checking that `eval` throws, it doesnt make sense to call `.back()` here.

pfultz2 (changes requested):

umangyadav (approved):

test/ref/resize.cpp
```diff
@@ -421,7 +421,7 @@ TEST_CASE(resize_fail_test_4)
                                            {"coordinate_transformation_mode", "invalid"}}),
                         {a0});
     p.compile(migraphx::make_target("ref"));
-    EXPECT(test::throws([&] { p.eval({}).back(); }));
+    EXPECT(test::throws([&] { std::ignore = p.eval({}); }));
```
pfultz2, line 424: We shouldn't need the `std::ignore` as `p.eval` does not have the `no_discard` attribute.

pfultz2 (approved):

PR #2844, opened 2024-02-29
Add pass to rewrite pow2 div

gyulaz-htec:
This pass is needed for Llama2 fp16 where RMSNorm calc can go out of bounds outputting inf values.
The pass rewrites `x^2/n` to `(x/sqrt(n))^2`.

### DIFF
--- a/src/include/migraphx/matcher.hpp
+++ b/src/include/migraphx/matcher.hpp
@@ -594,6 +594,14 @@ MIGRAPHX_PRED_MATCHER(same_input_shapes, instruction_ref ins)
         ins->inputs().begin(), ins->inputs().end(), [&](auto x) { return x->get_shape() == s; });
 }
 
+MIGRAPHX_PRED_MATCHER(same_inputs, instruction_ref ins)
+{
+    if(ins->inputs().empty())
+        return false;
+    auto s = ins->inputs().front();
+    return std::all_of(ins->inputs().begin(), ins->inputs().end(), [&](auto x) { return x == s; });
+}
+
 MIGRAPHX_PRED_MATCHER(has_same_value, instruction_ref ins)
 {
     if(ins->name() != "@literal")
--- a/src/include/migraphx/matcher.hpp
+++ b/src/include/migraphx/matcher.hpp
@@ -598,8 +598,9 @@ MIGRAPHX_PRED_MATCHER(same_inputs, instruction_ref ins)
 {
     if(ins->inputs().empty())
         return false;
-    auto s = ins->inputs().front();
-    return std::all_of(ins->inputs().begin(), ins->inputs().end(), [&](auto x) { return x == s; });
+    auto input = ins->inputs().front();
+    return std::all_of(
+        ins->inputs().begin(), ins->inputs().end(), [&](auto x) { return x == input; });
 }
 
 MIGRAPHX_PRED_MATCHER(has_same_value, instruction_ref ins)
--- a/src/quantization.cpp
+++ b/src/quantization.cpp
@@ -41,6 +41,7 @@
 #include <migraphx/make_op.hpp>
 #include <migraphx/pass_manager.hpp>
 #include <migraphx/normalize_ops.hpp>
+#include <migraphx/rewrite_low_precision.hpp>
 #include <set>
 
 namespace migraphx {
@@ -57,6 +58,7 @@ void quantize_fp16(program& prog, const std::vector<std::string>& ins_names)
 {
     run_passes(prog,
                {normalize_ops{},
+                rewrite_low_precision{},
                 optimize_module{{"quantizelinear", "dequantizelinear"}},
                 quantize_fp16_pass{ins_names},
                 optimize_module{{"quantizelinear", "dequantizelinear"}}});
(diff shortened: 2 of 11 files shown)


test/rewrite_pow2_div_test.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include <migraphx/make_op.hpp>
+#include <migraphx/pass_manager.hpp>
+#include <migraphx/program.hpp>
+#include <migraphx/register_target.hpp>
+#include <migraphx/common.hpp>
+#include <migraphx/verify.hpp>
+
+#include <test.hpp>
+
+void run_pass(migraphx::module& m)
```
umangyadav, line 36: can this go in the same test file as simplify_algebra.cpp ?

test/simplify_algebra_test.cpp
```diff
@@ -3711,4 +3711,32 @@ TEST_CASE(conv_concat)
     EXPECT(m1.sort() == m2.sort());
 }
 
+TEST_CASE(simplify_pow2_div)
```
umangyadav, line 3714: add another test for integral type

pfultz2:
This should go into a seperate pass. Maybe it could be called `rewrite_low_precision`?

src/simplify_algebra.cpp
```diff
@@ -1628,6 +1628,35 @@ struct find_split_transpose
     }
 };
 
+struct find_pow2_div
+{
+    auto matcher() const
+    {
+        return match::name("div")(
+            match::arg(0)(match::name("pow")(match::arg(0)(match::any().bind("x")),
```
pfultz2, line 1636: This should also look for `mul(x, x)` as well.

src/simplify_algebra.cpp
```diff
@@ -1628,6 +1628,35 @@ struct find_split_transpose
...
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins = r.result;
+        auto pow = r.instructions["pow"];
+        auto n   = r.instructions["n"];
+        auto x   = r.instructions["x"];
+
+        // sqrt(n) with integral numbers can cause huge accuracy loss due to float->int rounding
+        if(shape::is_integral(n->get_shape().type()))
```
pfultz2, line 1650: We should also be checking for low precision. We probably shouldn't do this rewrite with fp32.

gyulaz-htec:
@pfultz2 I've updated the PR
* Moved the rewrite in the a pass (`rewrite_low_precision`).
* Updated the matcher to match `mul(x,x)`.
* Enabled only `fp16` for the rewrite, should I include `fp8` as well?
* Added test to cover `mul(x,x)` and some more to test that the new pass don't apply to other types or to `mul(x,y)`.

**I've updated the source branch, but github didn't sync my changes in the PR. If this doesn't change druing the weekend I can reopen it.**

src/quantization.cpp
```diff
@@ -57,6 +58,7 @@ void quantize_fp16(program& prog, const std::vector<std::string>& ins_names)
 {
     run_passes(prog,
                {normalize_ops{},
+                rewrite_low_precision{},
```
pfultz2, line 61: This pass should run after `rewrite_reduce`.

src/include/migraphx/matcher.hpp
```diff
@@ -594,6 +594,14 @@ MIGRAPHX_PRED_MATCHER(same_input_shapes, instruction_ref ins)
         ins->inputs().begin(), ins->inputs().end(), [&](auto x) { return x->get_shape() == s; });
 }
 
+MIGRAPHX_PRED_MATCHER(same_inputs, instruction_ref ins)
+{
+    if(ins->inputs().empty())
+        return false;
+    auto s = ins->inputs().front();
```
pfultz2, line 601: Can you name this variable `input`? The `s` seems to imply a shape which it is not.

src/quantization.cpp
gyulaz-htec: Thanks! That makes more sense.

gyulaz-htec:
@pfultz2 I've made the proposed chanegs

pfultz2 (approved):

src/rewrite_low_precision.cpp
```diff
@@ -0,0 +1,69 @@
...
+        return match::name("div")(match::arg(0)(pow2()),
+                                  match::arg(1)(match::is_constant().bind("n")));
+    }
+
+    void apply(module& m, const match::matcher_result& r) const
+    {
+        auto ins = r.result;
+        auto n   = r.instructions["n"];
+        auto x   = r.instructions["x"];
+
```
lakhinderwalia, line 56: If instead of:
  x^2/n --> (x/sqrt(n))^2, 
The following is applied, is there any loss of accuracy? 
  x^2/n --> (x/n) * n
Thanks.

gyulaz-htec: Yes, that's actually works, the accuracy is the same with that. This solution saves us one instruction, also removes the `sqrt`. Thanks for the idea.

gyulaz-htec:
@pfultz2 @umangyadav @lakhinderwalia I've updated the PR with what @lakhinderwalia [proposed](https://github.com/ROCm/AMDMIGraphX/pull/2844#discussion_r1512126948) to use x^2/n --> (x/n) * x. The tests are passing with that change.

src/rewrite_low_precision.cpp
lakhinderwalia: Thanks.

lakhinderwalia (approved):
Thanks.

umangyadav (approved):

PR #2853, opened 2024-03-02
Add propagate_precision pass

pfultz2:


src/include/migraphx/propagate_precision.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef MIGRAPHX_GUARD_MIGRAPHX_PROMOTE_PRECISION_HPP
```
umangyadav, line 1: License

src/include/migraphx/shape.hpp
```diff
@@ -354,9 +354,9 @@ struct MIGRAPHX_EXPORT shape
 
         std::size_t size(std::size_t n = 1) const { return sizeof(type) * n; }
 
-        auto is_integral() const { return std::is_integral<type>{}; }
-        auto is_signed() const { return std::is_signed<type>{}; }
-        auto is_unsigned() const { return std::is_unsigned<type>{}; }
+        bool is_integral() const { return std::is_integral<type>{}; }
+        bool is_signed() const { return std::is_signed<type>{}; }
+        bool is_unsigned() const { return std::is_unsigned<type>{}; }
```
umangyadav, line 359: Curious why `auto` was changed to `bool` ?

src/propagate_precision.cpp
```diff
@@ -0,0 +1,191 @@
...
+    shape::type_t type;
+
+    friend bool operator==(const precision& xp, const precision& yp) { return xp.type == yp.type; }
+    friend bool operator<(const precision& xp, const precision& yp)
+    {
+        bool is_less = false;
+        shape::visit(xp.type, [&](auto x) {
+            shape::visit(yp.type, [&](auto y) {
+                if(x.is_integral() != y.is_integral())
+                    return;
```
umangyadav, line 28: For readability i think it is better to write it as
`if(not x.is_integral() or not y.is_integral())`

src/propagate_precision.cpp
```diff
@@ -0,0 +1,191 @@
...
+    friend bool operator>(const precision& xp, const precision& yp) { return yp < xp; }
+    // This is not totally ordered
+    friend bool operator<=(const precision& xp, const precision& yp)
+    {
+        return (xp < yp) or (xp == yp);
+    }
+    friend bool operator>=(const precision& xp, const precision& yp)
+    {
+        return (xp > yp) or (xp == yp);
+    }
```
TedThemistokleous, line 54: This might seem like an odd ask but why not make these xor vs or? If one is true then the other doesn't matter as the result shouldn't be true. Anyway adding xor instead of or here can speed things up / check for errors?

src/include/migraphx/shape.hpp
TedThemistokleous: I think this is for readability/clarity. I don't see why these shouldn't resolve to anything but bool? Unless we want to use value here specified by stl?

https://en.cppreference.com/w/cpp/types/is_unsigned

src/propagate_precision.cpp
```diff
@@ -0,0 +1,191 @@
+#include <migraphx/propagate_precision.hpp>
```
TedThemistokleous, line 1: Add license here too

test/propagate_precision.cpp
```diff
@@ -0,0 +1,158 @@
+#include <migraphx/propagate_precision.hpp>
```
TedThemistokleous, line 1: License

test/propagate_precision.cpp
```diff
@@ -0,0 +1,158 @@
...
+        auto y        = m2.add_parameter("y", s2);
+        auto convert1 = m2.add_instruction(
+            migraphx::make_op("convert", {{"target_type", migraphx::shape::double_type}}), x);
+        auto two  = m2.add_literal(migraphx::literal{{migraphx::shape::half_type}, {2}});
+        auto div  = migraphx::add_common_op(m2, migraphx::make_op("div"), {convert1, two});
+        auto sqrt = m2.add_instruction(migraphx::make_op("sqrt"), div);
+        auto mul  = m2.add_instruction(migraphx::make_op("mul"), sqrt, y);
+        m2.add_return({mul});
+    }
+    EXPECT(m1.sort() == m2.sort());
```
TedThemistokleous, line 118: I understand this is done to preserve precision throughout all these divides/square roots but are we not worried about the added overhead here now? we've just converted the fp16 set of ops to double, or is compute not a concern here?

test/propagate_precision.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+TEST_CASE(propagate_reduce)
```
umangyadav, line 121: can you add a test where the pass doesn't do anything ?

src/include/migraphx/shape.hpp
pfultz2: Doing `x.is_integral() != y.is_integral()` will fail to compile because they will be different types using `auto`. So I explicitly convert it to `bool` instead.

test/propagate_precision.cpp
pfultz2: But we are converting to double anyways. For elementwise, the compute shouldn't be that much overhead since these are all essentially unary operators.

CharlieL7 (commented):
For some background: where are we failing accuracy because of precision changes?

CharlieL7 (approved):
Code looks good. This looks like a slight performance decrease however? Are there any cases where we don't want to do this?

causten:
@pfultz2  can you fix the failing test and address review comments

pfultz2:
> This looks like a slight performance decrease however?

There could be, but it would be negligible. There is already a conversion to a higher precision so it will just do additional unary operators in that higher precision because it wont increase any data movement which is the expensive part.

> Are there any cases where we don't want to do this?

I dont think so, as this only applies when there is conversion to higher precision, and such a conversion would probably only be there because better accuracy is needed over performance, so this PR will help improve that.

ahsan-ca:
https://github.com/ROCm/AMDMIGraphX/pull/4009 should fix Jenkins failure.

test/propagate_precision.cpp
ahsan-ca: Added `propagate_reduce_float_to_double` test case that does not promote reduce from float to double since we do not have Data Parallel Primitives for reduce of double type.

src/propagate_precision.cpp
```diff
@@ -0,0 +1,240 @@
...
+void propagate_precision::apply(module_pass_manager& mpm) const
+{
+    auto upgrade = find_instruction_to_upgrade(mpm.get_module());
+    for(const auto& p : upgrade)
+    {
+        auto ins = p.first;
+        auto t   = p.second;
+
+        // Skip promoting reduce instruction from float to double
+        if(contains(ins->name(), "reduce") and
```
pfultz2, line 218: We should not be skipping float to double conversions.

ahsan-ca: Thanks for taking a look Paul. I was meaning to discuss this with you when you got back.

The reason why I added change to skip conversions from float to double was because DPP for reduction do not support the double type. So if we promote reduction to double from float, we won't be able to use DPP support. In other words, using float should make the reduction operation faster due to leveraging DPP.

I might be missing/misunderstanding how reduction is handled for this case. We can talk when you are back to understand how reduction is working for this case. 

I have reverted the change that skipped over float to double conversions.

ahsan-ca: For reference: the [Jenkins CI logs](http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/AMDMIGraphX/branches/PR-2853/runs/8/nodes/165/steps/167/log/?start=0) of the test case (test_einsum_sum_cpu) failing:
```
[2025-05-13T17:12:11.846Z] ./migraphx/kernels/reduce.hpp:165:1: error: not a valid operand.
[2025-05-13T17:12:11.846Z]   165 | MIGRAPHX_DPP_REDUCE(op::sum, v_add, _u)
[2025-05-13T17:12:11.846Z]       | ^
[2025-05-13T17:12:11.846Z] ./migraphx/kernels/reduce.hpp:157:5: note: expanded from macro 'MIGRAPHX_DPP_REDUCE'
[2025-05-13T17:12:11.846Z]   157 |     MIGRAPHX_DPP_REDUCE_ASM_FUN(double, op, prefix##_f64);      \
[2025-05-13T17:12:11.846Z]       |     ^
[2025-05-13T17:12:11.846Z] ./migraphx/kernels/reduce.hpp:131:13: note: expanded from macro 'MIGRAPHX_DPP_REDUCE_ASM_FUN'
[2025-05-13T17:12:11.846Z]   131 |             MIGRAPHX_DPP_REDUCE_ASM(1, x, ins, );                                     \
[2025-05-13T17:12:11.846Z]       |             ^
[2025-05-13T17:12:11.847Z] ./migraphx/kernels/reduce.hpp:110:60: note: expanded from macro 'MIGRAPHX_DPP_REDUCE_ASM'
[2025-05-13T17:12:11.847Z]   110 |         __asm__ volatile("s_nop 4\n" MIGRAPHX_PP_REPEAT(n, MIGRAPHX_DPP_REDUCE_ASM_REPEAT, ins) \
[2025-05-13T17:12:11.847Z]       |                                                            ^
[2025-05-13T17:12:11.847Z] <inline asm>:2:32: note: instantiated into assembly here
[2025-05-13T17:12:11.847Z]     2 | v_add_f64 v[2:3] v[2:3] v[2:3] row_shr:1
```

PR #4009 was put up to remove DPP for double type.

pfultz2: DPP can still be used for reduction we just cant use the assembly version. If you remove the macro call it will fallback to using the intrinsics which will work.

ahsan-ca:
Jenkins CI onnx pipeline failure is known and due to another issue, not related to this PR.

PR #2862, opened 2024-03-06
Sleep for 50ms after tuning to give the GPU a break

pfultz2:


kahmed10:
Are we seeing any noticeable difference here?

PR #2866, opened 2024-03-07
Fix targets registration on Windows

apwojcik:
Unfortunately, automatic target registration does not work on Windows.
The linker will not add a specified DLL as a dependency if at least a symbol from the dynamic library is not used. Therefore, linking to `migrpahx_all_targets` does not link dynamic libraries on Windows as intended.

This PR changes automatic registration to runtime manual registration. Registration is done when the target library is loaded by the `register_target()` function exported from the target library. It is called from the `migraphx::make_target()` function.


pfultz2:
Can we just use the `-Wl,--no-as-needed` flag instead on windows?

pfultz2:
Also, we could make `migraphx_all_targets` a static library and then have it load the all targets before `main`:

```cpp
__attribute__((constructor)) void load_targets()
{
    #ifdef HAVE_CPU
            make_target("cpu");
    #endif
    #ifdef HAVE_GPU
            make_target("gpu");
    #endif
    #ifdef HAVE_FPGA
            make_target("fpga");
    #endif
}
```

apwojcik:
> Can we just use the `-Wl,--no-as-needed` flag instead on windows?

It is not supported on Windows. It is only applicable to shared objects.

apwojcik:
> Also, we could make `migraphx_all_targets` a static library and then have it load the all targets before `main`:
> 
> ```c++
> __attribute__((constructor)) void load_targets()
> {
>             make_target("ref");
>     #ifdef HAVE_CPU
>             make_target("cpu");
>     #endif
>     #ifdef HAVE_GPU
>             make_target("gpu");
>     #endif
>     #ifdef HAVE_FPGA
>             make_target("fpga");
>     #endif
> }
> ```

This will now work on Windows. DLLs are not the same as shared objects. Furthermore, MSVC/Clang on Windows does not recognize the constructor attribute because Windows does not support it. One must implement DllMain() to serve a similar purpose. However, that is more complicated because it is process-attached and thread-attached aware—the auto registration of a target creates thread-local storage, making it unavailable after exiting from a library loading thread. A DLL on Windows should generate and return an execution context to a caller to store the per-process state.

apwojcik:
I managed to simplify the PR and limit the number of required changes.

pfultz2:
> Furthermore, MSVC/Clang on Windows does not recognize the constructor attribute because Windows does not support it. 

Sure, we can write it using standard C++ as well:

```cpp
struct auto_load_targets
{
    auto_load_targets()
    {
        make_target("ref");
#ifdef HAVE_CPU
        make_target("cpu");
#endif
#ifdef HAVE_GPU
        make_target("gpu");
#endif
#ifdef HAVE_FPGA
        make_target("fpga");
#endif
    }
};

static auto load_targets = auto_load_targets();
```

And this does work on windows because that is how we auto register unit tests. 

> One must implement DllMain() to serve a similar purpose. 

Why would we call `DllMain`? This is not run from a dll. This is a static library being linked into an executable.

apwojcik:
@pfultz2 it is done

src/register_target.cpp
```diff
@@ -32,6 +32,28 @@
...
+#ifdef HAVE_GPU
+        make_target("gpu");
+#endif
+#ifdef HAVE_FPGA
+        make_target("fpga");
+#endif
+    }
+};
+[[maybe_unused]] static auto load_targets = auto_load_targets{};
+} // namespace
```
pfultz2, line 54: This needs to go into a static/object library that we only link in for the tests. This prevents lazy loading the targets.

src/register_target.cpp
```diff
@@ -44,8 +66,6 @@ std::unordered_map<std::string, target>& target_map()
     return m;
 }
 
-void register_target_init() { (void)target_map(); }
```
pfultz2, line 47: Dont remove this, I think this was necessary to ensure targets are loaded in some cases.

src/include/migraphx/register_target.hpp
```diff
@@ -66,9 +65,6 @@ struct register_target_action
     }
 };
 
-template <class T>
-using auto_register_target = auto_register<register_target_action, T>;
```
pfultz2, line 70: Dont remove this, We still want to provide the option to register a target with CRTP.

test/CMakeLists.txt
```diff
@@ -33,7 +33,11 @@ if(MIGRAPHX_DISABLE_LARGE_BUFFER_TESTS)
     add_compile_definitions(MIGRAPHX_DISABLE_LARGE_BUFFER_TESTS)
 endif()
 
+add_library(register_targets STATIC register_target.cpp)
+target_link_libraries(register_targets PRIVATE migraphx)
```
pfultz2, line 37: I think you need to link `migraphx_all_targets` here as well to get the macros defines such as `HAVE_GPU`, `HAVE_CPU`, etc.

apwojcik: Done

src/register_target.cpp
```diff
@@ -86,16 +84,5 @@ target make_target(const std::string& name)
     return it->second;
 }
 
-std::vector<std::string> get_targets()
```
pfultz2, line 89: This function should not be removed.

apwojcik: Something is not right; I reverted that change. I will double-check.

apwojcik: I fixed it.

pfultz2 (approved):

PR #2870, opened 2024-03-07
use mlir for smaller 3x3 kernels

umangyadav:


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -269,7 +269,28 @@ auto is_mlir_conv(mlir_mode mode)
...
+        else
+        {
+            auto padding = v["padding"].to_vector<size_t>();
+            // if padding is not (1, 1), use MLIR
+            if(not std::any_of(padding.begin(), padding.end(), [](const auto& pad_value) {
+                   return pad_value != 1;
+               }))
+            {
+                return true;
+            }
```
pfultz2, line 286: You should check that stride is 1 as well.

umangyadav:
Closing this one for now untill we have more data or need

PR #2880, opened 2024-03-12
Supported ONNX operators list

attila-dusnoki-htec:
List of the Opset 19 operators and their state in MIGraphX parsing.


randyh62 (approved):
Seems fine.

PR #2882, opened 2024-03-12
Mlir disable specific ops

pfultz2:
Specific ops can be disabled using the `!` or `~` prefix to disable it.

### DIFF
--- a/src/targets/gpu/fuse_mlir.cpp
+++ b/src/targets/gpu/fuse_mlir.cpp
@@ -28,6 +28,7 @@
 #include <migraphx/make_op.hpp>
 #include <migraphx/register_op.hpp>
 #include <migraphx/env.hpp>
+#include <migraphx/algorithm.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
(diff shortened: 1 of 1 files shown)


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -61,12 +62,46 @@ bool mlir_enabled()
...
+
+template <class Action>
+static std::vector<std::string> get_usage()
+{
+    static const auto options =
+        split_string(string_value_of(MIGRAPHX_MLIR_USE_SPECIFIC_OPS{}, ""), ',');
+    static const auto enabled = std::is_same<Action, requested>{};
+    std::vector<std::string> result;
+    auto remove_not_symbol = [&](const std::string& s) {
+        if(starts_with(s, "!"))
```
umangyadav, line 82: Do you plan to add `~` symbol as well?

causten: so MIGRAPHX_MLIR_USE_SPECIFIC_OPS="!attention,dot" would not do any attentions through MLIR but would use MLIR for all dots ?

umangyadav (approved):

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -61,12 +62,55 @@ bool mlir_enabled()
...
+struct rejected
+{
+};
+} // namespace
+
+static bool is_negated_op(const std::string& s)
+{
+    if(s.empty())
+        return false;
+    return contains({'!', '~'}, s[0]);
```
umangyadav, line 78: you don't necessarily need to use `contains` `s[0] == '!'` or s[0] == `~` should suffice.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -61,12 +62,55 @@ bool mlir_enabled()
...
+    static const auto options =
+        split_string(string_value_of(MIGRAPHX_MLIR_USE_SPECIFIC_OPS{}, ""), ',');
+    static const bool enabled = std::is_same<Action, requested>{};
+    std::vector<std::string> result;
+    auto remove_not_symbol = [&](const std::string& s) {
+        if(is_negated_op(s))
+            return s.substr(1);
+        return s;
+    };
+    transform_if(
```
umangyadav, line 93: can it use copy_if instead ?

kahmed10:
We should update the documentation showing how to disable the specific ops

PR #2883, opened 2024-03-12
Rewrite reduce mean/variance

pfultz2:
Rewrites mean/variance to use `reduce_mean(x)` and `reduce_mean(x*x)` so it can be fused in the same reduction.


src/targets/gpu/compile_gen.cpp
```diff
@@ -298,7 +317,7 @@ std::string generate_reduce(module m, const std::string& name)
     auto f        = g.generate_module(m, [&](instruction_ref ins, const auto& names) {
         if(contains(ins->name(), "reduce"))
         {
-            return reduce_op::generate(ins, names.at(ins->inputs().front()));
+            return reduce_op::generate(ins, cpp_generator::to_args(ins->inputs(), names));
         }
         else if(ins->name() == "pointwise")
         {
```
TedThemistokleous, line 323: Why is this else case empty? Remove this if not needed

pfultz2: What else case is empty?

TedThemistokleous: Disregard refreshed github didn't load correctly on this for some reason

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -38,6 +38,8 @@ namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_LAYERNORM_FUSION);
```
umangyadav, line 41: add this to env var docs

src/fuse_reduce.cpp
```diff
@@ -210,6 +210,32 @@ static void create_reduce_modules(module_pass_manager& mpm)
     }
 }
 
+namespace {
+MIGRAPHX_PRED_MATCHER(used_once_except_broadcast, instruction_ref ins)
```
umangyadav, line 214: add unit-tests for this one

src/fuse_reduce.cpp
```diff
@@ -210,6 +210,32 @@ static void create_reduce_modules(module_pass_manager& mpm)
...
+            });
+        if(non_broadcast == ins->outputs().end())
+            return false;
+        return std::count_if(
+                   ins->outputs().begin(), ins->outputs().end(), [&](instruction_ref output) {
+                       if(not contains(output->name(), "broadcast"))
+                           return true;
+                       if(output->outputs().size() != 1)
+                           return true;
+                       return output->outputs().front() != *non_broadcast;
```
umangyadav, line 232: what's the reason for checking output's outputs ? 

name `used_once_except_broadcast` to me suggest that instruction can have more than 2 outputs where all but one are broadcasts.

pfultz2: So this is checking that broadcasted output goes into the same instruction as the other output. I rewrote it to make it clearer(the `count_if` is confusing).

src/fuse_reduce.cpp
pfultz2: I added a test for this case.

umangyadav: Where ? I expected test inside `test/matcher.cpp`

src/targets/gpu/prepare_reduce.cpp
```diff
@@ -0,0 +1,126 @@
...
+        return;
+    auto last = reduces.front();
+    auto op   = last->get_operator();
+    std::vector<instruction_ref> inputs;
+    std::transform(reduces.begin(), reduces.end(), std::back_inserter(inputs), [&](auto reduce) {
+        return reduce->inputs().front();
+    });
+    auto preduce = m.insert_instruction(last, parallel_reduce{op}, inputs);
+    int i        = 0;
+    for(auto reduce : reduces)
```
TedThemistokleous, line 112: Might be a slight nit-pick in naming but make "reduces" into something like "reductions" Just makes this a but more readable and then be clearer when you grab a reduce from the list of reductions.

Same goes for "preduce" make this into parallel_reductions since this might get interpreted as pointer-to-reduce with the name preduce.

test/fuse_reduce.cpp
```diff
@@ -213,6 +213,98 @@ TEST_CASE(reduce_reduce)
...
+                    migraphx::make_op("multibroadcast", {{"out_lens", s.lens()}}), add);
+                return add_pointwise(
+                    p2, rm, "main:pointwise3", {inputs[0], rsum1b, addb}, single_pointwise("clip"));
+            });
+        mm->add_return({clip});
+    }
+    EXPECT(p1.sort() == p2.sort());
+}
+
 TEST_CASE(reduce_reduce_mismatch_axis)
```
TedThemistokleous, line 308: This test case is good and I'm sure these can be massaged to create the other tests to handle cases where matching fails and some of the other coverage error warnings

TedThemistokleous (changes requested):
- [ ] Add additional test coverage rewrite_reduce.cpp
- [x] Readability for prepare_reduce.cpp

src/fuse_reduce.cpp
pfultz2: The test case was added to fuse_reduce.cpp. `test/matcher.cpp` is for testing the `migraphx/matcher.hpp`. This matcher does not belong there, its specific to fuse_reduce which is why its a internal/private function.

TedThemistokleous (approved):
Looks good.

kahmed10 (approved):

PR #2886, opened 2024-03-13
Fuse layout with pointwise op

umangyadav:


src/targets/gpu/fuse_ops.cpp
```diff
@@ -767,11 +767,15 @@ struct find_contiguous
     }
 };
 
-struct find_pointwise_contiguous
+struct find_pointwise_layout
```
pfultz2, line 770: Maybe this should be named `find_pointwise_layout_contiguous` to make it clear it works for contiguous as well as layout.

test/gpu/fuse_ops.cpp
```diff
@@ -160,6 +160,87 @@ TEST_CASE(pointwise_contiguous)
     EXPECT(p1 == p2);
 }
 
+TEST_CASE(pointwise_layout_convolution)
+{
+    migraphx::shape s1{migraphx::shape::float_type, {2, 320, 128, 128}};
+    migraphx::shape s2{migraphx::shape::float_type, {320, 320, 3, 3}, {2880, 1, 960, 320}};
+    migraphx::shape s3{migraphx::shape::float_type, {2, 320, 128, 128}, {5242880, 1, 40960, 320}};
+    migraphx::shape s4{migraphx::shape::int8_type, {41943040}}; //  workspace for gpu::convolution
```
pfultz2, line 168: I dont think its a good idea to hardcode the workspace which could change. Maybe its better to run lowering(and possibly `fuse_pointwise`).

src/targets/gpu/fuse_ops.cpp
umangyadav: Done.

kahmed10 (approved):

pfultz2 (approved):

Issue #2752, opened 2024-02-12
normalize_compute_shape: QUANT_CONVOLUTION: only accept input and weights of type int8_t or fp8e4m3fnuz_type


attila-dusnoki-htec:
- MIGraphX commit: [431e714](https://github.com/ROCm/AMDMIGraphX/commit/431e714ae6bd40e8793b5a2ab455bfbc45f1679e)
- GPU: MI210 (gfx90a)
- ROCm version: 6.0.1

Note: The problematic parts use `uint8`.

Model(s) to reproduce:
- https://huggingface.co/distil-whisper/distil-medium.en/blob/main/onnx/encoder_model_quantized.onnx
- https://huggingface.co/distil-whisper/distil-large-v2/blob/main/onnx/encoder_model_quantized.onnx
- https://huggingface.co/distil-whisper/distil-small.en/blob/main/onnx/encoder_model_quantized.onnx
- https://huggingface.co/jonathandinu/face-parsing/blob/main/onnx/model_quantized.onnx
- https://huggingface.co/enzostvs/hair-color/blob/main/onnx/model_quantized.onnx

Fails at: `/code/AMDMIGraphX/src/include/migraphx/op/quant_convolution.hpp:96: normalize_compute_shape: QUANT_CONVOLUTION: only accept input and weights of type int8_t or fp8e4m3fnuz_type`

TedThemistokleous:
Looks like the encoder has dynamicquantizelinear in it. I'll run this through the changes in my PR to take a look if that fixes this error



![Image](https://github.com/ROCm/AMDMIGraphX/assets/107195283/925b5029-6fff-4cb8-8733-9793b435167f)

TedThemistokleous:
Have an initial fix off develop called fix_encoder_errors, but there still seems to be an issue with MIOpen when running this as an int8 instead of uint8

TedThemistokleous:
@attila-dusnoki-htec  what settings did you use for these models?

I'm getting past the error for uint8 now but I'm seeing likely an issue related to dimensions for these runs

```
@1355 = sub(@1351,@1354) -> float_type, {1, 1, 1}, {1, 1, 1}, target_id=0
@1356 = div(@1355,@3) -> float_type, {1, 1, 1}, {1, 1, 1}, target_id=0
@1357 = div(@1354,@1356) -> float_type, {1, 1, 1}, {1, 1, 1}, target_id=0
[...]

PR #2888, opened 2024-03-15
Handle weight and data input zero points for ConvInteger

TedThemistokleous:
Wasn't being handled correctly. Required to correctly parse in ConvInteger especially in cases with DynamicQuantizeLinear feeding the resulting zero point

Onnx spec mentions we have potentially 4 inputs intead of 3 like regular Conv. This is cropping up as part of DynamicQuantizeLinear when used in networks.

We're currently not handling the case where there's a bias on the weights as seen in #2752 with models that use DynamicQuantizeLinear to determine a new zero point for the convolution


src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +149,40 @@ struct parse_convolution : op_parser<parse_convolution>
...
-        return info.add_bias(args, l1, 1);
+        if (op.name() == "quant_convolution") {
+            if(args.size() > 2)
+            {
+                l0_zp = args[2];
+                if (l0_zp->get_shape().type() != l0_shape.type()) {
+                    MIGRAPHX_THROW("PARSE: ConvInteger Data and Data Zero Point must have same type");
+                }
+
+                l0_zp = info.add_common_op("sub", l0, l0_zp);
```
lakhinderwalia, line 166: Could you please also check that it is not zero.

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +149,40 @@ struct parse_convolution : op_parser<parse_convolution>
...
+
+                if(args.size() > 3)
+                {
+                    w_zp = args[3];
```
lakhinderwalia, line 170: Are we handling both cases where w_zero_point is a scalar or a tensor..? Thanks.

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +149,40 @@ struct parse_convolution : op_parser<parse_convolution>
 
         recalc_conv_attributes(values, kdims);
 
+        instruction_ref ret;
+        // parse a_zero_point and b_zero_point values
+        auto l0_zp = l0;
```
lakhinderwalia, line 154: Thanks for doing this PR. 
You don't have to do anything about the pre-existing names ..
This is somehow the simplest and yet trickiest naming and assignment... I can remember..:  Is this 10 or l0, I need a bigger monitor to find that out. :-)

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +149,40 @@ struct parse_convolution : op_parser<parse_convolution>
...
+                ret = info.add_instruction(op, l0_zp, w_zp);
+            }
+        }
+        else {
+            auto l1 = info.add_instruction(op, l0, args[1]);
+            ret = info.add_bias(args, l1, 1);
+        }
+        return ret;
     }
 };
```
lakhinderwalia, line 188: If you end up modifying this file, could you please also cut-n-paste the operator documentation from https://onnx.ai/onnx/operators/onnx__ConvInteger.html. 
Thanks.

TedThemistokleous: The problem is this parser contains code to parse both Conv and Conv Integer. I can add it to quant convolution though

pfultz2: Dont paste it for quant_convolution. They are not the same operator. They are similar in some ways, but they are also different. 

I also dont see a reason to paste such documentation since it can be easily referenced.

src/onnx/parse_convolution.cpp
TedThemistokleous: We need to as per onnx spec. I can handle that

src/onnx/parse_convolution.cpp
TedThemistokleous: I agree with @pfultz2 since this is a more generic parser for convolution where functionality diverges.

src/onnx/parse_convolution.cpp
TedThemistokleous: Just to be clear, your concern is if the zero point exists but it itself is zero thus avoid adding subtraction node + broadcasts?

src/onnx/parse_convolution.cpp
TedThemistokleous: Should be handled now. I've added common op on the subtract to the bias to handle this.

src/onnx/parse_convolution.cpp
lakhinderwalia: Yes, please. It is as if no Zero Point was specified, and the default case would kick in. Thanks.

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +159,45 @@ struct parse_convolution : op_parser<parse_convolution>
...
+            MIGRAPHX_THROW("PARSE:Conv Data and Data Zero Point must have same type");
+        }
+
+        if(is_quant_conv && (w_zp->get_shape().type() != w_shape.type()))
+        {
+            MIGRAPHX_THROW("PARSE:ConvInteger Weights and Weights Zero Point must have same type");
+        }
+
+        // Assign input bias on weights or data accordingly
+        if(is_quant_conv && (l0_zp != l0))
```
lakhinderwalia, line 182: The idea of comparing L_0 to L_0_zp is not logical. We end up here because of set_bias(), above -- because of it,  L_0 compares to itself when no bias is present, at lines 171  -- or at 176. 
Suggestion: 1. Please separate the error handling, maybe as in https://github.com/ROCm/AMDMIGraphX/blob/77177af25ea1280af552f761293a352fa1c6e0f0/src/onnx/parse_qlinearmatmul.cpp#L96, 
Suggestion: 2. Remove comparisons like 'l0_zp != l0', as they can be equivalently handled by directly checking for index > 2 or index > 3 --- as that is all that set_bias() does!

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +159,45 @@ struct parse_convolution : op_parser<parse_convolution>
...
+        // parse a_zero_point and b_zero_point values
+        auto l0_zp         = set_bias(l0, 2, args);
+        auto w_zp          = set_bias(weights, 3, args);
+        auto is_quant_conv = opd.op_name == "quant_convolution";
+
         op.from_value(values);
-        auto l1 = info.add_instruction(op, l0, args[1]);
-        return info.add_bias(args, l1, 1);
+
+        // Check for type mismatch on parse
```
lakhinderwalia, line 170: Do we check somewhere if the input type of L0 should be one of: int8, uint8. Same for Weights. Also, if the output should be int32 only? Thanks.

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +159,45 @@ struct parse_convolution : op_parser<parse_convolution>
 
         recalc_conv_attributes(values, kdims);
 
+        instruction_ref ret;
+        // parse a_zero_point and b_zero_point values
+        auto l0_zp         = set_bias(l0, 2, args);
```
lakhinderwalia, line 164: Referencing https://onnx.ai/onnx/operators/onnx__ConvInteger.html, we could call l0 as x, or something like that..

Request: please rename l0 as x. I am having a very hard time differentiating between l0 and 10. _I know it is not your code, but it is time to perhaps clean it, as it is making code reviews harder._

test/onnx/gen_onnx.py
```diff
@@ -1315,6 +1315,22 @@ def conv_bias_test():
     return ([node], [x, y, z], [out])
 
 
+@onnx_test()
```
lakhinderwalia, line 1318: Could you please also add a test to verify the calculations, with these biases. These newly added tests check the expected graph, I would recommend we should also verify the numerical calculations.

src/onnx/parse_convolution.cpp
TedThemistokleous: Output should only be int32. I'm leaving the input to be checked by quant_convolution operator itself which checks weights and handles the proper output size of int32

https://github.com/ROCm/AMDMIGraphX/blob/develop/src/include/migraphx/op/quant_convolution.hpp#L91

src/onnx/parse_convolution.cpp
TedThemistokleous: Done

test/onnx/gen_onnx.py
TedThemistokleous: Yup understood

src/onnx/parse_convolution.cpp
TedThemistokleous: Done

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +168,25 @@ struct parse_convolution : op_parser<parse_convolution>
...
+        // parse a_zero_point and b_zero_point values
+        auto is_quant_conv = opd.op_name == "quant_convolution";
+        auto x_zp          = set_bias(x, info, 2, is_quant_conv, args);
+        auto w_zp          = set_bias(weights, info, 3, is_quant_conv, args);
+
         op.from_value(values);
-        auto l1 = info.add_instruction(op, l0, args[1]);
-        return info.add_bias(args, l1, 1);
+
+        x_zp->debug_print();
```
shivadbhavsar, line 179: remove debug  print

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +168,25 @@ struct parse_convolution : op_parser<parse_convolution>
...
+
+        ret = info.add_instruction(op, x_zp, w_zp);
```
shivadbhavsar, line 181: I am not sure we should be passing in x and zp thats been offset by the zero points here, in most cases this will cause the original weight and input to overflow. 
The proper way to handle zero points would be to do the computation using the original input/weight and then computing the offset to be applied to the result and then apply that offset.

test/onnx/gen_onnx.py
TedThemistokleous: There are tests in the onnx_backend_test that should cover this

seeing test_convInteger + test+convinteger_without_padding in onnx_backend_test.py 

I can add a verify test with the pad though to ensure this is fine

src/onnx/parse_convolution.cpp
TedThemistokleous: Conv(x,w) - Conv(x_zp, w_zp) is what you mean? Is the goal then to hope these get fused?

TedThemistokleous: @shivadbhavsar added this and updated parser coverage to handle.

src/onnx/parse_convolution.cpp
TedThemistokleous: Removed

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +193,58 @@ struct parse_convolution : op_parser<parse_convolution>
...
-        return info.add_bias(args, l1, 1);
+
+        ret = info.add_instruction(op, x, weights);
+
+        // Handle quant_conv residuals between input/weights to avoid overflow
+        if(is_quant_conv)
+        {
+            if(not is_symmetric_zero_point(x_zp) and is_input_bias)
+            {
+                auto out_zp_1 = info.add_common_op("quant_convolution", x_zp, weights);
```
shivadbhavsar, line 214: these additional quant_conv ops should use the same attributes (ie. padding, etc.) as the original quant_conv

src/onnx/parse_convolution.cpp
```diff
@@ -149,9 +193,58 @@ struct parse_convolution : op_parser<parse_convolution>
 
         recalc_conv_attributes(values, kdims);
 
+        instruction_ref ret;
+        // parse a_zero_point and b_zero_point values
+        auto is_quant_conv = opd.op_name == "quant_convolution";
+        auto is_input_bias  = args.size() > 2;
+        auto is_weight_bias = args.size() > 3;
```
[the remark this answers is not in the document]
TedThemistokleous: So don't check it based on arg size but instead check based on eval(). ok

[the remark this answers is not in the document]
TedThemistokleous: Actually, I think we have to also check the type based on the literal here. For quant conv that means it needs to be non zero if int8, and 128 if uint8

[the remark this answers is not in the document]
shivadbhavsar: yes thats true, in simplify_qdq we skip over uint8 so the code from there only accounts for the int8 case

src/onnx/parse_convolution.cpp
TedThemistokleous: TODO: Share same attributes between quant_convolutions for asymmetric  inputs

src/onnx/parse_convolution.cpp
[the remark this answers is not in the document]
TedThemistokleous: Added handling of the uint8 case for now similar to how I handled in in 2903 with a shift. Should shift the literal 0 to the proper 128 midpoint value for uint8 and then we can apply our exclusion of zero points accordingly based on symmetry. 

Still need to sus out the mismatched cases and oddball verify tests with stuff like uint8 data, int8 weights and visa versa.

src/onnx/parse_convolution.cpp
```diff
@@ -42,18 +43,153 @@ struct parse_convolution : op_parser<parse_convolution>
...
+                                             const instruction_ref& x,
+                                             const instruction_ref& weights,
+                                             const instruction_ref& x_zp,
+                                             const instruction_ref& w_zp,
+                                             onnx_parser::node_info& info)
+    {
+        instruction_ref ret = input;
+        if(not is_symmetric_zero_point(x_zp))
+        {
+            auto out_zp_1 = info.add_common_op("quant_convolution", x_zp, weights);
```
shivadbhavsar, line 156: I think these are still not using the attributes from the original quant conv. You should use `op` from L295

TedThemistokleous: Done

src/onnx/parse_convolution.cpp
TedThemistokleous: Done

shivadbhavsar (approved):

src/onnx/parse_convolution.cpp
```diff
@@ -42,18 +43,201 @@ struct parse_convolution : op_parser<parse_convolution>
...
+        if(args.size() > index)
+        {
+            // Check for type mismatch on parse
+            if(input->get_shape().type() != args[index]->get_shape().type())
+                MIGRAPHX_THROW("PARSE:Conv Data and Data Zero Point must have same type");
+
+            ret = args[index];
+            if(is_symmetric_zero_point(ret))
+            {
+                std::cout << "Hit symmetric input" << std::endl;
```
lakhinderwalia, line 118: Is this a debug statement? Thanks.

src/onnx/parse_convolution.cpp
```diff
@@ -42,18 +43,201 @@ struct parse_convolution : op_parser<parse_convolution>
...
+                                     instruction_ref& input_bias)
+    {
+        input = add_int8_shift(info, offset_op, input);
+        if(has_bias)
+        {
+            input_bias = add_int8_shift(info, offset_op, input_bias);
+        }
+    }
+
+    static float get_symmetric_value(const instruction_ref& input)
```
lakhinderwalia, line 76: This function is basically a one-liner if-statement.

src/onnx/parse_convolution.cpp
```diff
@@ -42,18 +43,201 @@ struct parse_convolution : op_parser<parse_convolution>
...
+
+            ret = args[index];
+            if(is_symmetric_zero_point(ret))
+            {
+                std::cout << "Hit symmetric input" << std::endl;
+                ret = gen_symmetric_literal(ret, is_quant_conv, info);
+            }
+            ret->debug_print();
+        }
+        else
```
lakhinderwalia, line 123: Both the if/else look identical..

src/onnx/parse_convolution.cpp
```diff
@@ -42,18 +43,201 @@ struct parse_convolution : op_parser<parse_convolution>
...
+
+        auto input_type  = input->get_shape().type();
+        auto weight_type = weights->get_shape().type();
+
+        // Handle uint8 bias and input shifts
+        instruction_ref offset_op;
+        if(((input_type == migraphx::shape::uint8_type) or
+            (weight_type == migraphx::shape::uint8_type)))
+        {
+            offset_op = info.add_literal(
```
lakhinderwalia, line 212: Is the offset_op always -128? Would it not depend upon the inputted zero_point of the input tensor (the zero point of uint8 can be anywhere between 0 to 255.. right?)

src/onnx/parse_convolution.cpp
```diff
@@ -42,18 +43,201 @@ struct parse_convolution : op_parser<parse_convolution>
...
+        }
+        else
+        {
+            ret = gen_symmetric_literal(ret, is_quant_conv, info);
+        }
+
+        return ret;
+    }
+
+    static bool is_symmetric_zero_point(instruction_ref zp)
```
lakhinderwalia, line 131: why should one care if it is symmetrical or not.. a simple add operator works with or without symmetry, Thanks..

test/onnx/gen_onnx.py
```diff
@@ -1315,6 +1315,22 @@ def conv_bias_test():
...
+def conv_bad_bias_test():
```
lakhinderwalia, line 1319: Not sure I follow the negative test here. Is it the wrong type of tensors x & y? Thanks.

test/onnx/gen_onnx.py
```diff
@@ -1545,6 +1591,72 @@ def convinteger_bias_test():
     return ([node], [x, y, z], [out])
 
 
+@onnx_test()
```
lakhinderwalia, line 1594: Could you please change one of the inputs, x or y to be UINT8. Mixed types are allowed by the standard, I think. You can either modify the test, or add a fresh one. Thanks.

src/onnx/parse_convolution.cpp
```diff
@@ -42,18 +43,201 @@ struct parse_convolution : op_parser<parse_convolution>
         return {{"Conv", "convolution"}, {"ConvInteger", "quant_convolution"}};
     }
 
+    // Convert to half prior to a shift to ensure we preserve accuracy here then
+    // convert back to int8
+    static instruction_ref add_int8_shift(const onnx_parser::node_info& info,
```
lakhinderwalia, line 48: Suggestion: a name like uint8_to_int8 might be more functionally informative. But this is highly subjective, so you don't have to change anything really.. up to you. Thanks.

src/onnx/parse_convolution.cpp
TedThemistokleous: Ah yes something I missed earlier. Will remove

src/onnx/parse_convolution.cpp
TedThemistokleous: I'd rather have this as a function call incase other types are required in the future

src/onnx/parse_convolution.cpp
TedThemistokleous: The intent here is to add a valid literal that's symmetric scalar value so that other parts of the code don't get triggered generating residual convolution blocks.

src/onnx/parse_convolution.cpp
TedThemistokleous: This is relevant as we're avoiding overflow + reducing terms created when we parse in a Convolution and this leads to a few simplifications + helps accuracy .

In this operator we're performing convolution of two sums which simplifies too

`Conv(X - A, Y - B) =  Conv(X,Y) - Conv(X, B) - Conv(Y,A) + Conv(A,B)`

Convolution becomes a a set of matrix multiplication and additions of terms.

For the int8 case, the multiply in the convolution in Int8, MatMulInteger(X,Y) between terms must not overflow but the accumulation can as each MatMulInteger op accumulates in int32. Expanding the terms out ensures we wont roll over/have accuracy issues when performing a bias before convolution.

`MatMulInteger(X - A, Y - B) =  MatMulInteger(X,Y) - MatMulinteger(X, B) - MatMulinteger(Y,A) + MatMulInteger(A,B)`

If a zero point is symmetric (128 in uint8, and 0 in int8) one of the four terms before the convolution is performed becomes zero, thus we don't create two of the terms in the expansion.

That's why in the code we're checking for symmetric with respect to datatype and I've expanded it from a regular inline if to function calls to get and generate symmetric zero points. Should we start adding other quantized types, we can just add the data type to the list with is specific zero point.

test/onnx/gen_onnx.py
TedThemistokleous: This was added as I'm modifying parse_convolution which handles both quantized convolution and regular convolution. 

I noticed we were missing a type check on the bias used for convolution and added the test case. In short, bias arguments must be the identical type to which input they apply too. We weren't covering that case. Here Z, is int32 instead of FLOAT like the X input which should be an error.

test/onnx/gen_onnx.py
TedThemistokleous: I have the case below with convinteger_mismatched_input_types_test() but I can add a case for dual bias of different types

src/onnx/parse_convolution.cpp
TedThemistokleous: The intent here is to shift everything to int8 as most of our backend only works smoothly with int8 and not uint8, primarily MLIR with their kernels

test/onnx/verify/quant_convolution_dual_bias_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+ * THE SOFTWARE.
+ */
+
+#include <migraphx/register_target.hpp>
+#include <migraphx/verify.hpp>
+#include <onnx_test.hpp>
+
+TEST_CASE(quant_convolution_dual_zero_bias_test)
+{
+    // github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.QLinearMul
```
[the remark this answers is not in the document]
TedThemistokleous: Removed

src/onnx/parse_convolution.cpp
TedThemistokleous: Yes since we're moving everything from uint8 to int8 we need to subtract 128

src/onnx/parse_convolution.cpp
TedThemistokleous: Removed

test/onnx/gen_onnx.py
TedThemistokleous: Added the extra case of uint8 for one input and bias and int8 for another by reusing one of the test and changing the types. Gold data should be the same since the quantizations between int8/uint8 shouldn't effect the intended information/accumulation at the end when we get an int32 value

TedThemistokleous:
seeing this locally when I enabled hipRTC

```
/workspace/AMDMIGraphX/test/onnx
root@aus-navi3x-02:/workspace/AMDMIGraphX/test/onnx# ../../build/bin/test_verify "ck_gemm_softmax_gemm"
[   RUN    ] ck_gemm_softmax_gemm
[ COMPLETE ] ck_gemm_softmax_gemm (3361.42ms)
[==========] 1 tests ran
root@aus-navi3x-02:/workspace/AMDMIGr
```

Not sure what's going on with CI here. @lakhinderwalia  was there anymore comments you had for this one?

pfultz2:
Why is this being done in the onnx parser? Why cant we parse the quant_convolution and then have a pass to trasnform it to int8 afterwards?

TedThemistokleous:
> Why is this being done in the onnx parser? Why cant we parse the quant_convolution and then have a pass to transform it to int8 afterwards?

This is being converted to int8 since pushing things to uint8 breaks MLIR from my understanding. I've also handled mismatches on this end.  I figured after this I can write a pass for a further optimization to remove/bundle these since the intent here was to get things to run and not fail. 

When adding the uint8 datatype it had implications with MLIR and lowering when I tried it when dealing with MatMulInteger and I'd assume the same is true for ConvInteger.

pfultz2:
> This is being converted to int8 since pushing things to uint8 breaks MLIR from my understanding. 

The lack of support for uint8 is specific to the gpu-backend, so the frontend parsers(such as onnx) should not be compensating for that. Rather the backend(such as the gpu targets should have passes that transform it into something that it can support. We already do this for eliminate_data_type to handle data types that the frontend onnx parses but the gpu target does not support.

TedThemistokleous:
> > This is being converted to int8 since pushing things to uint8 breaks MLIR from my understanding.
> 
> The lack of support for uint8 is specific to the gpu-backend, so the frontend parsers(such as onnx) should not be compensating for that. Rather the backend(such as the gpu target) should have passes that transform it into something that it can support. We already do this for eliminate_data_type to handle data types that the frontend onnx parses but the gpu target does not support.

I agree, but this was added in to get support for the missing input + uint8 data types in uint8 for ConvInteger.

. What do you suggest I do here? This PR's been open since march, and I've handled all the comments.

src/onnx/parse_convolution.cpp
lakhinderwalia: Yes, agreed on that: to convert to int8. But exactly what is to be (permissibly) converted? Only uint8. I assume this is to convert only from uint8 to int8. But never mind..

src/onnx/parse_convolution.cpp
lakhinderwalia: Thanks.  The correct functional boundary for this constant adjustment: it should be contained inside add_int8_shift().

src/onnx/parse_convolution.cpp
lakhinderwalia: Pardon me, what I meant was, there is an 'add' step (of a constant value 128) for `bias` and `input` (or whatever..) in `add_int8_shift()`. Then we have a 'subract' step in `handle_quant_bias()`. Couldn't there be just one 'subract' of a value = zp - 128 ? And avoid the 'add' step for a constant '128'.

lakhinderwalia: BTW, it can be a simple 'subtract' as I mention above.. Or just an 'add' if the quantities were reversed. But the point is, do we need both: a 'subtract' with the ZP + an 'add' with '128'?

lakhinderwalia:
> The lack of support for uint8 is specific to the gpu-backend, so the frontend parsers(such as onnx) should not be compensating for that. Rather the backend(such as the gpu target) should have passes that transform it into something that it can support. We already do this for eliminate_data_type to handle data types that the frontend onnx parses but the gpu target does not support.

@pfultz2, isn't the zp for weights known at compile/parse time, so that could be adjusted (uint8<-> int8) at that stage? But we would need to add this conversation operation in the GPU pipeline at the boundary of inputs (uint8->int8) to a convinteger.. Thanks.

src/onnx/parse_convolution.cpp
TedThemistokleous: Ah I think the confusion lies in the fact this parser handles both Conv and ConvInteger inputs

src/onnx/parse_convolution.cpp
TedThemistokleous: OH I see what you're saying. There's two things going on here that are not related

1. The add_int8_shift() shifts the input (data + bias value) so that it operates in int8 if its uint8 if that uint8 is a symmetric zero point (128) thus we subtract and do the add of a negative -128 value to data and zero point.

2. the handle_quant_bias() applies the bias  shifted convolution step (after the datatype is shifted from uint8 to int8) with the post shifted biases since the result is expanded as I explained above. To expand things it requires us to have identical datatypes as each Conv resolves to a GEMM, and mismatching GEMMs (uint8 and int8) aren't permitted. The result of each of the expanded convolutions is int32 and thus we dont need to worry as much about overflow of the result when we add the resulting convolutions

The subtraction isn't for the bias here, its to subtract Convolutions between biases and zero points which would be negative when we expand out Conv(X - a, Y - b)

pfultz2:
> I agree, but this was added in to get support for the missing input + uint8 data types in uint8 for ConvInteger.

Updating the onnx parser should first just get the ref version working and then we can investigate how we want to support it for the gpu. The problem is that when MLIR support uint8 a lot of work is needed to remove this code and update all the tests whereas if we do this as a pass we can just disable the transformation to enable the support.

But one thing I am not clear on, are the weights and inputs different data types so we need to do the conversion? Or are they both uint8?

> What do you suggest I do here? This PR's been open since march

The PR title or the description doesnt say anything about converting the data type. I have mentioned before in other PRs that such conversions should be done as a pass. The only case possibly we should being doing the conversion in the parser is when it uses different data types, since our ref version doesnt support it(although it could be extended which might be a better options).

>  Is there a specific set of review comments you want to add here to fix things?

Remove shifting the uint8 data in the parser, and add uint8 as a supported type in the `quant_convolution::normalize_compute_shape` method.

pfultz2:
> isn't the zp for weights known at compile/parse time, so that could be adjusted (uint8<-> int8) at that stage? 

At what stage? We shouldn't adjust it at the parsing stage, it should be adjusted during the compile stage.

TedThemistokleous:
> > I agree, but this was added in to get support for the missing input + uint8 data types in uint8 for ConvInteger.
> 
> Updating the onnx parser should first just get the ref version working and then we can investigate how we want to support it for the gpu. The problem is that when MLIR support uint8 a lot of work is needed to remove this code and update all the tests whereas if we do this as a pass we can just disable the transformation to enable the support.
> 
> But one thing I am not clear on, are the weights and inputs different data types so we need to do the conversion? Or are they both uint8?
> 
> > What do you suggest I do here? This PR's been open since march
> 
> The PR title or the description doesnt say anything about converting the data type. I have mentioned before in other PRs that such conversions should be done as a pass. The only case possibly we should being doing the conversion in the parser is when it uses different data types, since our ref version doesnt support it(although it could be extended which might be a better options).
> 
The concern here and why I've done things with the shift in the parser is that there exists cases where a model has given us a uint8 input for weights with zero point along with int8 for the data and its bias value. 

> > Is there a specific set of review comments you want to add here to fix things?
> 
> Remove shifting the uint8 data in the parser, and add uint8 as a supported type in the `quant_convolution::normalize_compute_shape` method.

The corresponding bias for the data/weight must be the same type, but the data/weights can be different types. Even if I added the support this doesn't handle cases where one is uint8 and int8 without running into accuracy issues with zero points.

TedThemistokleous:
@pfultz2 more to your comment, on removing the shift. We already do a shift on MatMulInteger, and I figured I'd fix up some existing work to handle both MatMulinteger + ConvInteger with different inputs to do the proper shift if we leave the shifts in here for now and remove them later.

pfultz2:
I can see a reason to do this for mismatched types, but if both are uint8 there shouldn't be shifts inserted. 

The `MatMulInteger` seems to insert shifts for both cases, which seems unfortunate, but I guess since its already done there, you could do the same here if you want, although it might create more work when we want to actually want to support uint8. 

If you do decide to remove the shifts, it would be good to have `MatMulInteger` updated as well(not in this PR but at at least an issue to track it).

src/include/migraphx/op/quant_convolution.hpp
```diff
@@ -75,7 +75,7 @@ struct quant_convolution
 
     shape normalize_compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.has(2).same_type().same_ndims().min_ndims(3);
+        check_shapes{inputs, *this}.has(2).same_ndims().min_ndims(3);
```
pfultz2, line 78: If we want to support different types, the compute method needs to be updated to use a different visit for input and weights(instead of using visit_all for them).

TedThemistokleous:
@pfultz2 reverted shift changes. I'll handle pass as part of 2826 for DynamicQuantize + MatMulInteger as well. Making this an issue as well to track

src/onnx/parse_convolution.cpp
lakhinderwalia: As per our discussion, this is a const-folding topic, so not an issue for me that there are effectively two 'add' operation, instead of just one.

lakhinderwalia (approved):

pfultz2 (approved):

PR #2893, opened 2024-03-18
Enable rocgdb in the dockerfile

pfultz2:
rocgdb is not supported on 22.04. This works around the issue by installing the older python version used on ubuntu 20.04(the newer python is still install in the dockerfile).


Dockerfile
```diff
@@ -15,6 +15,9 @@ RUN sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/6.0
 # From docs.amd.com for installing rocm. Needed to install properly
 RUN sh -c "echo 'Package: *\nPin: release o=repo.radeon.com\nPin-priority: 600' > /etc/apt/preferences.d/rocm-pin-600"
 
+# rocgdb doesn't work on 22.04, workaround by installing the older python packages that are in 20.04
+RUN add-apt-repository -y ppa:deadsnakes/ppa
```
umangyadav, line 19: What's the connection between rocgdb and python version ? 
rocgdb uses python3.8 internally ?

pfultz2: Yea, it uses libpython3.8 directly which is not available on ubuntu 22.04.

umangyadav (approved):

Dockerfile
causten: You need to install "software-properties-common" before running add-apt-repository

causten (changes requested):

Dockerfile
pfultz2: I added the missing package.

causten (approved):

PR #2903, opened 2024-03-19
Fix parsing of MatMulInteger

TedThemistokleous:
Originally we were using dot and converting the input for MatMulinteger() as it was parsed in.  As a result we were inserting dot() instead of quant_dot into the model.

I've added a change to check for similar data types and then add a shift to an input  where there's a mismatch of int8 and uint8 on input. This is usually seen when dynamicquantizelinear is used as the input as it inserts a uint8 as the zero point which in some models (like gpt2) is operated on.

With this I'm seeing around a 30% boost in the GPT run.

```
Summary:
gpu::code_object::reduce_min_kernel: 2.06349ms / 49 = 0.042112ms, 14%
gpu::code_object::reduce_max_sub_mul_kernel: 2.05264ms / 49 = 0.0418906ms, 13%
gpu::code_object::mlir_quant_dot: 1.87745ms / 61 = 0.0307779ms, 12%
gpu::code_object::concat_kernel: 1.15502ms / 49 = 0.0235718ms, 8%
gpu::code_object::quantizelinear_sub_convert_add_convert_kernel: 1.14867ms / 49 = 0.0234422ms, 8%
gpu::code_object::mul_kernel: 1.13639ms / 49 = 0.0231916ms, 8%
gpu::code_object::neg_div_clip_nearbyint_convert_kernel: 1.13261ms / 49 = 0.0231144ms, 8%
gpu::code_object::contiguous_kernel: 1.12907ms / 48 = 0.0235223ms, 8%
gpu::code_object::quantizelinear_kernel: 0.836898ms / 36 = 0.0232472ms, 6%
gpu::code_object::layernorm_mul_add_kernel: 0.585637ms / 24 = 0.0244015ms, 4%
gpu::code_object::convert_mul_add_add_kernel: 0.545402ms / 23 = 0.0237131ms, 4%
gpu::code_object::convert_mul_add_kernel: 0.310343ms / 13 = 0.0238725ms, 2%
load: 0.307436ms / 515 = 0.000596963ms, 2%
gpu::code_object::dequantizelinear_mul_where_reduce_max_sub_exp_reduce_sum_div_quantizelinear_kernel: 0.294008ms / 12 = 0.0245007ms, 2%
gpu::code_object::convert_mul_add_mul_mul_add_mul_exp_add_div_kernel: 0.28876ms / 12 = 0.0240634ms, 2%
gpu::code_object::mlir_quant_dot_dequantizelinear: 0.285526ms / 12 = 0.0237939ms, 2%
multibroadcast: 0.220696ms / 246 = 0.000897138ms, 2%
reshape_lazy: 0.119711ms / 180 = 0.000665061ms, 1%
hip::hip_copy_literal: 0.100321ms / 151 = 0.00066438ms, 1%
transpose: 0.0685118ms / 48 = 0.00142733ms, 1%
slice: 0.0486096ms / 36 = 0.00135027ms, 1%
gpu::code_object::convert_mul_kernel: 0.0291348ms / 1 = 0.0291348ms, 1%
gpu::code_object::add_layernorm_mul_add_kernel: 0.0249158ms / 1 = 0.0249158ms, 1%
gpu::code_object::dequantizelinear_add_kernel: 0.0237881ms / 1 = 0.0237881ms, 1%
gpu::code_object::gather_kernel: 0.023749ms / 1 = 0.023749ms, 1%
gpu::code_object::convert_kernel: 0.0230378ms / 1 = 0.0230378ms, 1%
@param: 0.00976646ms / 26 = 0.000375633ms, 1%
hip::hip_allocate_memory: 0.000916ms / 1 = 0.000916ms, 1%
check_context::migraphx::gpu::context: 0.000771ms / 1 = 0.000771ms, 1%

Batch size: 1
Rate: 212.998 inferences/sec
Total time: 4.69488ms
Total instructions time: 15.8433ms
Overhead time: 0.376437ms, -11.1484ms
Overhead: 8%, -237%
[ MIGraphX Version: 2.10.0. ] Complete: bin/driver perf ../int8_models/gpt2_1_int8_gpu.onnx --input-dim @input_ids 1 32 --fill1 input_ids --disable-fast-math --int8


```

### DIFF
--- a/src/onnx/parse_matmul.cpp
+++ b/src/onnx/parse_matmul.cpp
@@ -91,46 +91,101 @@ struct parse_matmul : op_parser<parse_matmul>
             auto s1_lens        = a1->get_shape().lens();
             instruction_ref ba0 = a0;
             instruction_ref ba1 = a1;
-            // try broadcasting if dimensions other than last two do not match
-            if(not std::equal(
-                   s0_lens.rbegin() + 2, s0_lens.rend(), s1_lens.rbegin() + 2, s1_lens.rend()))
-            {
-                auto l0_it = s0_lens.begin() + s0_lens.size() - 2;
-                std::vector<std::size_t> l0_broadcasted_lens(s0_lens.begin(), l0_it);
-                auto l1_it = s1_lens.begin() + s1_lens.size() - 2;
-                std::vector<std::size_t> l1_broadcasted_lens(s1_lens.begin(), l1_it);
-                auto output_lens =
-                    compute_broadcasted_lens(l0_broadcasted_lens, l1_broadcasted_lens);
-                l0_broadcasted_lens = output_lens;
-                l0_broadcasted_lens.insert(l0_broadcasted_lens.end(), l0_it, s0_lens.end());
-                l1_broadcasted_lens = output_lens;
-                l1_broadcasted_lens.insert(l1_broadcasted_lens.end(), l1_it, s1_lens.end());
-                if(s0_lens != l0_broadcasted_lens)
-                {
-                    ba0 = info.add_instruction(
-                        make_op("multibroadcast", {{"out_lens", l0_broadcasted_lens}}), a0);
-                }
-                if(s1_lens != l1_broadcasted_lens)
-                {
-                    ba1 = info.add_instruction(
-                        make_op("multibroadcast", {{"out_lens", l1_broadcasted_lens}}), a1);
-                }
-            }
 
             // parse a_zero_point and b_zero_point values
             if(args.size() > 2)
             {
-                ba0 = info.add_instruction(
-                    make_op("convert", {{"target_type", migraphx::shape::float_type}}), ba0);
+                if(args[2]->get_shape().type() != a0->get_shape().type())
+                {
+                    MIGRAPHX_THROW("PARSE_QUANT_DOT: A zero point must be the same type as A data");
+                }
 
-                ba0 = info.add_common_op("sub", ba0, args[2]);
+                ba0 = args[2];
+                ba0 = info.add_common_op("sub", a0, ba0);
+
                 if(args.size() > 3)
                 {
-                    ba1 = info.add_instruction(
-                        make_op("convert", {{"target_type", migraphx::shape::float_type}}), ba1);
-                    ba1 = info.add_common_op("sub", ba1, args[3]);
+                    if(args[3]->get_shape().type() != a1->get_shape().type())
+                    {
+                        MIGRAPHX_THROW(
+                            "PARSE_QUANT_DOT: B zero point must be the same type as B data");
+                    }
+
+                    ba1 = args[3];
+                    ba1 = info.add_common_op("sub", a1, ba1);
                 }
-                dot_res = info.add_instruction(make_op("dot"), ba0, ba1);
+
+                auto ba0_type = ba0->get_shape().type();
+                auto ba1_type = ba1->get_shape().type();
+
+                // determine if we need to add a shift to ensure items get shifted to the same int8
+                // type if we see uint8/int8 mismatches
+                if(ba0_type != ba1_type)
+                {
+                    instruction_ref unshifted_input;
+                    if(ba0_type == migraphx::shape::int8_type)
+                    {
+                        unshifted_input = ba1;
+                    }
+                    if(ba1_type == migraphx::shape::int8_type)
+                    {
+                        unshifted_input = ba0;
+                    }
+
+                    // Convert to fp16 prior to a shift to ensure we preserve accuracy here then
+                    // convert back to int8
+                    auto int8_shift = info.add_literal(
+                        migraphx::literal{migraphx::shape{migraphx::shape::int16_type}, {-128}});
+
+                    auto unshifted_input_int16 = info.add_instruction(
+                        migraphx::make_op("convert",
+                                          {{"target_type", migraphx::shape::int16_type}}),
+                        unshifted_input);
+
+                    auto input_shifted_int16 =
+                        info.add_common_op("add", unshifted_input_int16, int8_shift);
+
+                    auto shifted_input = info.add_instruction(
+                        migraphx::make_op("convert", {{"target_type", migraphx::shape::int8_type}}),
+                        input_shifted_int16);
+
+                    if(ba0_type == migraphx::shape::int8_type)
+                    {
+                        ba1 = shifted_input;
+                    }
+                    if(ba1_type == migraphx::shape::int8_type)
+                    {
+                        ba0 = shifted_input;
+                    }
+                }
+
+                // try broadcasting if dimensions other than last two do not match
+                if(not std::equal(
+                       s0_lens.rbegin() + 2, s0_lens.rend(), s1_lens.rbegin() + 2, s1_lens.rend()))
+                {
+                    auto l0_it = s0_lens.begin() + s0_lens.size() - 2;
+                    std::vector<std::size_t> l0_broadcasted_lens(s0_lens.begin(), l0_it);
+                    auto l1_it = s1_lens.begin() + s1_lens.size() - 2;
+                    std::vector<std::size_t> l1_broadcasted_lens(s1_lens.begin(), l1_it);
+                    auto output_lens =
+                        compute_broadcasted_lens(l0_broadcasted_lens, l1_broadcasted_lens);
+                    l0_broadcasted_lens = output_lens;
+                    l0_broadcasted_lens.insert(l0_broadcasted_lens.end(), l0_it, s0_lens.end());
+                    l1_broadcasted_lens = output_lens;
+                    l1_broadcasted_lens.insert(l1_broadcasted_lens.end(), l1_it, s1_lens.end());
+                    if(s0_lens != l0_broadcasted_lens)
+                    {
+                        ba0 = info.add_instruction(
+                            make_op("multibroadcast", {{"out_lens", l0_broadcasted_lens}}), a0);
+                    }
+                    if(s1_lens != l1_broadcasted_lens)
+                    {
+                        ba1 = info.add_instruction(
+                            make_op("multibroadcast", {{"out_lens", l1_broadcasted_lens}}), a1);
+                    }
+                }
+
+                dot_res = info.add_instruction(make_op("quant_dot"), ba0, ba1);
                 dot_res = info.add_instruction(
                     make_op("convert", {{"target_type", migraphx::shape::int32_type}}), dot_res);
             }


src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,100 @@ struct parse_matmul : op_parser<parse_matmul>
...
+                             instruction_ref& bias_arg)
+    {
+        if(args.size() > index)
+        {
+            if(args[index]->get_shape().type() != input->get_shape().type())
+            {
+                MIGRAPHX_THROW("PARSE_QUANT_DOT: zero point must be the same type as data");
+            }
+
+            bias_arg = args[index];
```
umangyadav, line 130:
```suggestion
            bias_arg = args[index];
            if(bias_arg->get_shape().type() != input->get_shape().type())
            {
                MIGRAPHX_THROW("PARSE_QUANT_DOT: zero point must be the same type as data");
            }

            
```

TedThemistokleous: Done

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,100 @@ struct parse_matmul : op_parser<parse_matmul>
...
+        *shifted_output = shifted_input;
+    }
+
+    static void set_bias_arg(const onnx_parser::node_info& info,
+                             const std::vector<instruction_ref>& args,
+                             const int index,
+                             const instruction_ref& input,
+                             instruction_ref& bias_arg)
+    {
+        if(args.size() > index)
```
umangyadav, line 123: For the else case add a THROW

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,100 @@ struct parse_matmul : op_parser<parse_matmul>
...
+        {
+            bias_arg = args[index];
+            if(bias_arg->get_shape().type() != input->get_shape().type())
+            {
+                MIGRAPHX_THROW("PARSE_QUANT_DOT: zero point must be the same type as data");
```
umangyadav, line 128: if not too much of work add a test coverage for this line.

src/onnx/parse_matmul.cpp
```diff
@@ -91,51 +185,21 @@ struct parse_matmul : op_parser<parse_matmul>
...
-                        make_op("multibroadcast", {{"out_lens", l0_broadcasted_lens}}), a0);
-                }
-                if(s1_lens != l1_broadcasted_lens)
-                {
-                    ba1 = info.add_instruction(
-                        make_op("multibroadcast", {{"out_lens", l1_broadcasted_lens}}), a1);
-                }
-            }
+                set_bias_arg(info, args, 2, a0, ba0);
+                set_bias_arg(info, args, 3, a1, ba1);
```
umangyadav, line 193: `ba0` and `ba1` are used for broadcasted `a0` and `a1`. 
Do not mix them with bias_arg. 

You can make `set_bias_arg` to return `instruction_ref`

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,100 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            return;
+
+        instruction_ref unshifted_input;
+        instruction_ref* shifted_output = nullptr;
+        if(ba0_type == migraphx::shape::int8_type)
+        {
+            unshifted_input = ba1;
+            shifted_output  = &ba1;
+        }
+        if(ba1_type == migraphx::shape::int8_type)
```
umangyadav, line 93: make it `else if` and add `else` clause that throws. Right now `MatMulInteger` is only supporting int8/uint8 but that may change in future to include FP8.

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,100 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            migraphx::make_op("convert", {{"target_type", migraphx::shape::int16_type}}),
+            unshifted_input);
+
+        auto input_shifted_int16 = info.add_common_op("add", unshifted_input_int16, int8_shift);
+
+        auto shifted_input = info.add_instruction(
+            migraphx::make_op("convert", {{"target_type", migraphx::shape::int8_type}}),
+            input_shifted_int16);
+
+        *shifted_output = shifted_input;
```
umangyadav, line 114: why do you require setting this pointer ? can you instead make it return `instruction_ref` instead of void ?

src/onnx/parse_matmul.cpp
TedThemistokleous: You want to see a supported types lookup then? That should be on input for the quant_dot then as this should just check if types are the same but would pass if we have A,B inputs + zero points all be fp8s

src/onnx/parse_matmul.cpp
TedThemistokleous: Sure I can do that

src/onnx/parse_matmul.cpp
TedThemistokleous: You shall get a throw

src/onnx/parse_matmul.cpp
TedThemistokleous: Good point. Adding additional coverage then

src/onnx/parse_matmul.cpp
TedThemistokleous: Will do

src/onnx/parse_matmul.cpp
TedThemistokleous: done

src/onnx/parse_matmul.cpp
TedThemistokleous: Added a throw for matmul case outside in if block. Onnx spec mentions this should only have two inputs so no additional args should be parsed in for that op

src/onnx/parse_matmul.cpp
TedThemistokleous: Coverage added

src/onnx/parse_matmul.cpp
TedThemistokleous: Added support for specific types and test coverage on its failure

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,77 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            }
+            if(s1_lens != l1_broadcasted_lens)
+            {
+                ba1 = info.add_instruction(
+                    make_op("multibroadcast", {{"out_lens", l1_broadcasted_lens}}), a1);
+            }
+        }
+    }
+
+    // Convert to fp16 prior to a shift to ensure we preserve accuracy here then
```
umangyadav, line 75:
```suggestion
    // Convert to int16 prior to a shift to ensure we preserve accuracy here then
```

TedThemistokleous: Done

src/eliminate_data_type.cpp
```diff
@@ -108,7 +108,8 @@ void eliminate_data_type::apply(module& m) const
                                                            "scatternd_add",
                                                            "scatternd_mul",
                                                            "scatternd_none",
-                                                           "select_module"};
+                                                           "select_module",
+                                                           "quantizelinear"};
```
pfultz2, line 112: Why is this added here?

TedThemistokleous: Using eliminate_data_type with uint8 to handle inputs adds a convert to dynamicquantizelinear and breaks CI check in the onnxbackend tests

pfultz2: I thought we only did this for `quant_dot`(which also should include `dot`, `convolution`, and `quant_convolution`). Why are we removing uint8 on qlinear?

TedThemistokleous: Hitting fails in CI with Onnxruntime backend with the quant_dot. I can't seem to repro locally but its showing up in jenkins

src/onnx/parse_matmul.cpp
```diff
@@ -89,55 +160,40 @@ struct parse_matmul : op_parser<parse_matmul>
...
-                {
-                    ba0 = info.add_instruction(
-                        make_op("multibroadcast", {{"out_lens", l0_broadcasted_lens}}), a0);
-                }
-                if(s1_lens != l1_broadcasted_lens)
-                {
-                    ba1 = info.add_instruction(
-                        make_op("multibroadcast", {{"out_lens", l1_broadcasted_lens}}), a1);
-                }
+                MIGRAPHX_THROW("PARSE_MATMUL: Bias Args not supported for MatMul");
```
umangyadav, line 174: This throw should happen before call to `set_bias_arg()`

src/onnx/parse_matmul.cpp
```diff
@@ -89,55 +160,40 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            if(is_quant_dot and not is_same_type and (ba0_type == migraphx::shape::uint8_type))
             {
-                dot_res = info.add_instruction(make_op(opd.op_name), ba0, ba1);
+                ba0 = add_int8_shift(info, ba0);
             }
+
+            if(is_quant_dot and not is_same_type and (ba1_type == migraphx::shape::uint8_type))
+            {
+                ba1 = add_int8_shift(info, ba1);
+            }
```
umangyadav, line 193: it would be better if you can simplify it like 
```
if(is_quant_dot and not is_same_type) {
if(ba1_type == migraphx::shape::uint8_type) {}
if(ba2_type == migrpahx::shape::uint8_type)()
}
```

umangyadav:
> With this I'm seeing around a 30% boost in the GPT run.

 Do you have a way to check for accuracy with your changes ?

src/onnx/parse_matmul.cpp
TedThemistokleous: I simplified it this way  because tidy was giving me errors on complexity but I see your point. Less checks better

TedThemistokleous: Done

src/onnx/parse_matmul.cpp
```diff
@@ -99,10 +99,7 @@ struct parse_matmul : op_parser<parse_matmul>
         if(args.size() > index)
         {
             instruction_ref bias_arg = args[index];
-            if(bias_arg->get_shape().type() != input->get_shape().type())
```
umangyadav: Why was this removed ?

src/onnx/parse_matmul.cpp
```diff
@@ -182,14 +180,12 @@ struct parse_matmul : op_parser<parse_matmul>
...
-            if(is_quant_dot and not is_same_type and (ba0_type == migraphx::shape::uint8_type))
-            {
-                ba0 = add_int8_shift(info, ba0);
-            }
-
-            if(is_quant_dot and not is_same_type and (ba1_type == migraphx::shape::uint8_type))
+            if(is_quant_dot and not is_same_type)
             {
-                ba1 = add_int8_shift(info, ba1);
+                if((ba0_type == migraphx::shape::uint8_type))
```
umangyadav, line 185: Extra paranthesis

TedThemistokleous:
> > With this I'm seeing around a 30% boost in the GPT run.
> 
> Do you have a way to check for accuracy with your changes ?

Not yet. Do you want me to run a driver verify on it?

src/onnx/parse_matmul.cpp
TedThemistokleous: Had to reset editor. Fixed this

src/onnx/parse_matmul.cpp
TedThemistokleous: Done

src/onnx/parse_matmul.cpp
TedThemistokleous: Done

umangyadav:
> > > With this I'm seeing around a 30% boost in the GPT run.
> > 
> > 
> > Do you have a way to check for accuracy with your changes ?
> 
> Not yet. Do you want me to run a driver verify on it?

Driver verify would use onnx file both  the "ref" and "gpu".  This PR is modifying parsing and that applies to both Ref and Gpu and therefore both Ref and GPU would generate same results. 

I meant if there is way to run through DLM or UIF and try out with real data and see .

umangyadav (approved):
Changes wise looks good to me. 
Make sure to address paul's comment and also if you have way to verify for accuracy. that's great too.

TedThemistokleous:
> Changes wise looks good to me. Make sure to address paul's comment and also if you have way to verify for accuracy. that's great too.

I was able to run this through Onnxruntime using one of the tests we have in DLM that requires int8 quant

https://github.com/ROCm/AMDMIGraphX/issues/1904#issuecomment-2009595842

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,77 @@ struct parse_matmul : op_parser<parse_matmul>
...
+                                          instruction_ref& unshifted_input)
+    {
+        auto int8_shift = info.add_literal(
+            migraphx::literal{migraphx::shape{migraphx::shape::int16_type}, {-128}});
+
+        auto unshifted_input_int16 = info.add_instruction(
+            migraphx::make_op("convert", {{"target_type", migraphx::shape::int16_type}}),
+            unshifted_input);
+
+        auto input_shifted_int16 = info.add_common_op("add", unshifted_input_int16, int8_shift);
```
bpickrel, line 87: No change needed here, but can you tell me why this op needs `add_common_op()` while the others nearby use `add_instruction()`?

test/onnx/verify/matmulinteger_int8_uint8_one_zp_test.cpp
```diff
@@ -0,0 +1,48 @@
...
+    std::vector<uint8_t> data1 = {128, 129, 126, 131, 124, 133};
+
+    migraphx::parameter_map pp;
+    pp["1"] = migraphx::argument(s0, data0.data());
+    pp["2"] = migraphx::argument(s1, data1.data());
+
+    auto result = p.eval(pp).back();
+    std::vector<int32_t> result_vector;
+    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
+    std::vector<int32_t> gold = {56, -76, -22, 21, 60, -82, 14, -25};
```
bpickrel, line 46: It looks like this multiplies the same two matrices as the previous test and gets completely different results.  What am I missing?

test/onnx/verify/matmulinteger_uns_zp_test.cpp
```diff
@@ -43,6 +43,6 @@ TEST_CASE(matmulinteger_uns_zp_test)
     auto result = p.eval(pp).back();
     std::vector<int32_t> result_vector;
     result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
-    std::vector<int32_t> gold = {-38, -83, -44, -98, -50, -113, -56, -128};
+    std::vector<int32_t> gold = {13, 76, 10, 64, 7, 52, 4, 40};
```
bpickrel, line 46: same question.  Looks like the previous gold value comes from the Onnx doc at https://github.com/onnx/onnx/blob/main/docs/Operators.md#examples-87 .  Why does the answer change?

bpickrel (commented):
I left a couple of comments on low level style and a couple of just questions.  I don't understand why the gold values in the tests changed.  No comments on the logic behind this PR as I'm not familiar enough with it to add anything.

test/onnx/verify/matmulinteger_uns_zp_test.cpp
TedThemistokleous: I wrote my own by hand to validate this output

test/onnx/verify/matmulinteger_int8_uint8_one_zp_test.cpp
TedThemistokleous: Different zero points

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,77 @@ struct parse_matmul : op_parser<parse_matmul>
...
 
+    static void broadcast_dimensions(const onnx_parser::node_info& info,
+                                     const std::vector<size_t>& s0_lens,
+                                     const std::vector<size_t>& s1_lens,
+                                     instruction_ref& a0,
+                                     instruction_ref& a1,
+                                     instruction_ref& ba0,
+                                     instruction_ref& ba1)
+    {
+        // try broadcasting if dimensions other than last two do not match
```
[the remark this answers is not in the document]
TedThemistokleous: No its for the last two. First dimensions should math here.

src/onnx/parse_matmul.cpp
TedThemistokleous: To handle broadcasts/shapes between ops and constants literals such as zero offsets

bpickrel (approved):
I still have 2 or 3 comments that haven't been answered, but they are style matters and not blockers.

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,77 @@ struct parse_matmul : op_parser<parse_matmul>
         return {{"MatMul", "dot"}, {"MatMulInteger", "quant_dot"}};
     }
 
+    static void broadcast_dimensions(const onnx_parser::node_info& info,
+                                     const std::vector<size_t>& s0_lens,
+                                     const std::vector<size_t>& s1_lens,
+                                     instruction_ref& a0,
```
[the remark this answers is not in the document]
TedThemistokleous: I think I see your point here as these are inputs and not modified in the call, only ba0, ba1

src/targets/gpu/target.cpp
```diff
@@ -131,6 +131,7 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         simplify_qdq{},
         enable_pass(not mlir_enabled(), rewrite_quantization{}),
         dead_code_elimination{},
+        eliminate_data_type{{migraphx::shape::uint8_type}, shape::float_type},
```
umangyadav, line 134: You are not specifiying "ops" for which it should eliminate uint8 type. Therefore by default it would eliminate uint8 for all the ops. 

If that is what you intend to do then, you can just add `uint8` in `unsupported_type` and just do one call to `eliminate_data_type{}` instead of three that are happening here.

src/eliminate_data_type.cpp
umangyadav: @TedThemistokleous are you able to reproduce locally ? Or else maybe share error log from Jenkins ? 
IMO we shouldn't require removing uint8 for the quantizelinear if we already handle quant_dot or quant_conv.

src/targets/gpu/target.cpp
TedThemistokleous: That's the intent. Okay will do

src/eliminate_data_type.cpp
TedThemistokleous: Didn't see it locally but I saw it on jenkins. Let me push the target fix first and ensure this does arise. I was happening with compile

src/targets/gpu/target.cpp
TedThemistokleous: Done

src/eliminate_data_type.cpp
TedThemistokleous: Removed this. Seems to be fine working locally.

pfultz2 (approved):

src/targets/gpu/target.cpp
```diff
@@ -131,8 +130,8 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         simplify_qdq{},
         enable_pass(not mlir_enabled(), rewrite_quantization{}),
         dead_code_elimination{},
-        // workaround for rocBLAS unsupported error when using uint8 in quant_dot
-        eliminate_data_type{{migraphx::shape::uint8_type}, shape::float_type, {"quant_dot"}},
+        // workaround for rocBLAS unsupported error when using uint8 in quant_dot & quant_convolution
+        eliminate_data_type{{migraphx::shape::uint8_type}, shape::float_type, {{"quant_convolution"}, {"quant_dot"}}},
```
umangyadav, line 134:
```suggestion
        eliminate_data_type{{migraphx::shape::uint8_type}, shape::float_type, {"quant_convolution", "quant_dot"}},
```

src/targets/gpu/target.cpp
```diff
@@ -92,7 +92,6 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
     unsupported_types.erase(shape::type_t::half_type);
     unsupported_types.erase(shape::type_t::bool_type);
     unsupported_types.erase(shape::type_t::int8_type);
-    unsupported_types.erase(shape::type_t::uint8_type);
```
umangyadav, line 95: Revert this change. 

`uint8_type` is supported type. 

If you erase that from unsupported_types then eliminate_data_type would try to convert quantizelinear with uint8 to float. 

I think We should only convert uint8 type quant_dot and quant_convolution to supported type.

src/eliminate_data_type.cpp
```diff
@@ -108,6 +108,7 @@ void eliminate_data_type::apply(module& m) const
                                                            "scatternd_add",
                                                            "scatternd_mul",
                                                            "scatternd_none",
+                                                           "quantizelinear",
```
umangyadav, line 111: You can revert this change

src/onnx/parse_matmul.cpp
```diff
@@ -38,6 +38,77 @@ struct parse_matmul : op_parser<parse_matmul>
...
+    {
+        if(args.size() > index)
+        {
+            instruction_ref bias_arg = args[index];
+            if(bias_arg->get_shape().type() != input->get_shape().type())
+            {
+                MIGRAPHX_THROW("PARSE_QUANT_DOT: zero point must be the same type as data");
+            }
+
+            return info.add_common_op("sub", input, bias_arg);
```
umangyadav, line 107: This would do "sub" between two uint8_t types which would result in incorrect results. 
It should do subtraction in signed type

umangyadav: https://onnx.ai/onnx/operators/onnx__MatMulInteger.html#matmulinteger

A and AZ will have same type and B and BZ will have same type. 

For the `matmulInteger(A-AZ, B-ZA)`  we can have four cases. 

1. `A : int8` `B : uint8` --> in this case we would need to convert entire operation to float type. Because for the `B - BZ` in worse case can generate `-255` as the value and that would require GEMM to happen in float16/32  to represent all values properly. 
2. `A: int8`, `B:int8` --> we already handle that case but `B - BZ` or `A - AZ` can again give `-128 - 128` = `-255` value which can't be handled. Therefore entire GEMM need to happen in float16/32
3. `A: uint8` and `B:uint8` same case as 2. it may generate `-255` 
4. `A:uint8` and  `B:int8` same as case 1. 
5. 
Here is the reference python implementation from onnx : 
https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_matmul_integer.py#L15

I think what is required is 
1. If any input and arg and corresponding zero point is `uint8` then first convert them to `int8` by adding `-128` shift. 
2. then compute `A - AZ` and `B-BZ`
3. And then do the `int8 * int8 = int32` 

Right now what you are doing is following : 
1.  compute `A-AZ` and `B-BZ` which would suffer from unsigned subtraction issue. Ideally we should throw warning for unsigned subtraction. 
2.  If any of the input is uint8 and then convert to int8 by shifting -128. 
3.  And then do `int8 * int8 = int32`

TedThemistokleous: Done

src/eliminate_data_type.cpp
TedThemistokleous: Reverted

src/targets/gpu/target.cpp
TedThemistokleous: Reverted

src/targets/gpu/target.cpp
TedThemistokleous: Done

src/onnx/parse_matmul.cpp
```diff
@@ -89,55 +182,84 @@ struct parse_matmul : op_parser<parse_matmul>
...
+                                                                 migraphx::shape::int8_type};
+            const auto a0_type                                = a0->get_shape().type();
+            const auto a1_type                                = a1->get_shape().type();
+            const auto ba0_type                               = ba0->get_shape().type();
+            const auto ba1_type                               = ba1->get_shape().type();
+
+            if(is_quant_dot and
+               (not contains(supported_types, ba0_type) or
+                not contains(supported_types, ba1_type) or not contains(supported_types, a0_type) or
+                not contains(supported_types, a1_type)))
```
umangyadav, line 207: You only need to check for valid types for `a1` and `a0` . biases will have same type as a0 and a1 for which there is a check in set_bias_arg

src/onnx/parse_matmul.cpp
```diff
@@ -89,55 +182,84 @@ struct parse_matmul : op_parser<parse_matmul>
...
+
+            if(is_quant_dot)
+            {
+                instruction_ref offset_op;
+                if(a0_type == migraphx::shape::uint8_type or a1_type == migraphx::shape::uint8_type)
                 {
-                    ba0 = info.add_instruction(
-                        make_op("multibroadcast", {{"out_lens", l0_broadcasted_lens}}), a0);
+                    offset_op = info.add_literal(
+                        migraphx::literal{migraphx::shape{migraphx::shape::int16_type}, {-128}});
```
umangyadav: Better make them `float16`. `int16` is unsupported type and would be converted into `float` by eliminate_data_type

src/onnx/parse_matmul.cpp
```diff
@@ -89,55 +182,84 @@ struct parse_matmul : op_parser<parse_matmul>
...
+                if(a1_type == migraphx::shape::uint8_type)
+                {
+                    a1 = add_int8_shift(info, offset_op, a1);
+                    if(has_ba1)
+                    {
+                        ba1 = add_int8_shift(info, offset_op, ba1);
+                    }
+                    else
+                    {
+                        ba1 = a1;
```
umangyadav: Else is unnecessary `ba1 = a1` is there is no bias. It would be set by `set_bias_arg`

src/onnx/parse_matmul.cpp
TedThemistokleous: Removed

test/onnx/verify/matmulinteger_uns_test.cpp
```diff
@@ -43,6 +43,6 @@ TEST_CASE(matmulinteger_uns_test)
     auto result = p.eval(pp).back();
     std::vector<int32_t> result_vector;
     result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
-    std::vector<int32_t> gold = {34, 97, 28, 82, 22, 67, 16, 52};
+    std::vector<int32_t> gold = {45730, 44641, 46108, 45010, 46486, 45379, 46864, 45748};
```
umangyadav, line 46: Can you make sure these golden data are correct by running them through numpy python code ?

umangyadav: You can use 
https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_matmul_integer.py#L15

src/onnx/parse_matmul.cpp
TedThemistokleous: issue is we still need to add the shift so without this we're using the unshifted version of a0/a1

umangyadav: That would be taken care by `has_ba1` `has_ba2`

TedThemistokleous: but a0/a1 get shifted and then ba0/ba1 gets used for the appropriate dot operator to be used

src/onnx/parse_matmul.cpp
TedThemistokleous: Done

umangyadav (approved):

PR #2906, opened 2024-03-19
Repeat loop only while module is being changed

pfultz2:


src/module.cpp
```diff
@@ -1122,6 +1205,25 @@ ins_dep_map module::calc_implicit_deps() const
...
+{
+    if(n == 0)
+        return;
+    if(n == 1)
+    {
+        f();
+        return;
+    }
+    auto has_changed = impl->changed.subscribe();
+    for(auto i : range(n))
```
umangyadav, line 1218:
```suggestion
    for(auto : range(n))
```
Does this work ?

src/module.cpp
```diff
@@ -41,12 +41,83 @@
...
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_FINALIZE)
 
+template <std::size_t N>
+struct bit_signal
+{
+    std::bitset<N> slots;
+    std::bitset<N> allocated;
```
umangyadav, line 55: why do you need bitset<N>. 

Why can't simple `bool` flag is is enough ? 
Is it because of multi-threaded parallel compilation? Then is `64` sufficient ?

src/module.cpp
pfultz2: No it doesnt.

src/module.cpp
pfultz2: There could be multiple subscribers withing the same thread, especially since this is called recursively. Once for the optimize_module loop and then again for the sub-passes like simplify_algebra, simplify_reshapes, etc.

> Is it because of multi-threaded parallel compilation?

The current implementation is not really thread-safe, and it seems unlikely to be used across threads since modifying the module across threads is not thread-safe either.

src/module.cpp
```diff
@@ -41,12 +41,83 @@
...
 #include <unordered_set>
+#include <bitset>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_FINALIZE)
 
+template <std::size_t N>
+struct bit_signal
```
CharlieL7, line 52: Unit tests for this struct would be good

src/module.cpp
```diff
@@ -41,12 +41,83 @@
...
 #include <utility>
 #include <unordered_set>
+#include <bitset>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_FINALIZE)
 
+template <std::size_t N>
```
CharlieL7, line 51:
```suggestion
/**
* Observer pattern for keeping track of if something has changed or been updated.
* Can have up to `N` different subscribers.
* Use by creating a `bit_signal` and adding subscribers with `bit_signal.subscribe()`.
* Use `bit_signal.notify()` to set that subscribers should be notified.
* Get the status of the subscription by checking the `slot` returned by `bit_signal.subscribe()`.
*/
template <std::size_t N>
```

CharlieL7 (changes requested):

src/module.cpp
umangyadav: Didn't know about observer pattern. Thanks!

umangyadav (approved):
Looks good, But address @CharlieL7 comments.

src/include/migraphx/bit_signal.hpp
```diff
@@ -0,0 +1,90 @@
...
+    struct slot
+    {
+        bit_signal* handler = nullptr;
+        std::size_t i       = N;
+
+        slot() = default;
+
+        slot(bit_signal* h, std::size_t x) : handler(h), i(x) {}
+
+        slot(slot&& rhs) : handler(rhs.handler), i(rhs.i)
```
umangyadav, line 33: @pfultz2  tidy fixes required

src/module.cpp
pfultz2: Tests added.

CharlieL7 (approved):

PR #2908, opened 2024-03-20
Add initial splitk reduce

pfultz2:
This adds a pass to do splitk reductions. It also updates the kernel to handle splitk reduciton as well. Right now, this only works for a single reduce, future work will enable this for multiple reductions. This is only enabled when using `MIGRAPHX_ENABLE_SPLIT_REDUCE=1` until we get more testing for this.  

I do see a 2x improvement when using this on large reductions like 327680.


src/targets/gpu/kernels/include/migraphx/kernels/scatter_reduction_modes.hpp
```diff
@@ -42,7 +42,7 @@ struct assign_add
     template <class T, class U>
     MIGRAPHX_DEVICE_CONSTEXPR void operator()(T& x, U y) const
     {
-        atomicAdd(&x, y);
+        unsafeAtomicAdd(&x, T(y));
```
TedThemistokleous, line 45: Why make this now non atomic? Is this due to the async fill added for hip calls?

src/include/migraphx/split_reduce.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <string>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+struct module_pass_manager;
+
+struct MIGRAPHX_EXPORT split_reduce
+{
+    std::size_t split_size = 8192;
```
TedThemistokleous, line 38: How did we determine 8192 as the size here? Is this some number we've hit somewhere? I understand this is just 2^13 but is there a reason for this?

src/module.cpp
```diff
@@ -423,61 +421,62 @@ instruction_ref module::move_instructions(instruction_ref src, instruction_ref d
...
 module::insert_instructions(instruction_ref ins,
                             const std::vector<instruction_ref>& instructions,
-                            std::unordered_map<instruction_ref, instruction_ref> map_ins,
+                            std::unordered_map<instruction_ref, instruction_ref>* map_ins,
                             module::inserter insert)
 {
+    std::unordered_map<instruction_ref, instruction_ref> default_map_ins;
     return insert_generic_instructions(
-        *this, ins, instructions, std::move(map_ins), std::move(insert));
+        *this, ins, instructions, map_ins ? *map_ins : default_map_ins, std::move(insert));
```
TedThemistokleous, line 455: Wouldn't it be safer to do `!map_ins ? default_map_ins: *map_ins ` here since map_ins can be null. I'm assuming pointer is used vs reference here since the initialization for map_ins can potentially be nullptr by default.

Is there a way to not use nullptr in anyway?

src/targets/gpu/kernels/include/migraphx/kernels/scatter_reduction_modes.hpp
pfultz2: Its still atomic. Its just using "unsafe" atomics so it will use the atomic add in the hardware(otherwise this will fallback to compare-and-swap which is really slow).

TedThemistokleous: makes sense

src/include/migraphx/split_reduce.hpp
pfultz2: This is just a number I picked after running some different sizes. I maybe adjusting this number further in the future.

src/module.cpp
pfultz2: > Wouldn't it be safer to do !map_ins ? default_map_ins: *map_ins here since map_ins can be null.

Whats the difference? It seems cleaner to reverse the condition so we dont need the extra `not`.

> Is there a way to not use nullptr in anyway and move these all to references?

No, this is an optional parameter.

src/include/migraphx/split_reduce.hpp
TedThemistokleous: Do you think its worth tying this to some sort of environment var so we have control of this outside of compiling it in? Seems like something we'd want to test for/tune if it'll make a difference between larger/small models unless the intent was "After a certain point in size we should be split_reducing things" 

More so I'm curious if regardless of model size, is there an optimal split_size for each we can set and see a benifit.

src/module.cpp
pfultz2: Ah wait, I need to explictly check for null, so its probably better to invert it then.

src/include/migraphx/split_reduce.hpp
pfultz2: I dont think this will need to be a tuning parameter, and I dont see a reason to expose a env variable for this setting. We can alway add one later if we want to do tweak it. This is similiar to the large K value we use for fusing gemms with mlir/ck. We dont have an env variable for it as we haven't needed to tweak it like that.

src/module.cpp
pfultz2: > Do we not want to just have a default reference to some object instead if its not set just for type safety/needing to worry about null? 

We would still need to check for null though.

> One thing that does irk me is that we're passing around null pointers for map_ins then default setting this after a check, whereas why not just make this a reference to a default empty and pass around const references then? Or is that not possible in this case?

You can't do that with lvalue references. Also as a bonus, with a pointer it will make it explicit from the caller that we will modify the variable.

src/split_reduce.cpp
```diff
@@ -0,0 +1,208 @@
...
+static std::vector<instruction_ref> get_alive(const_module_ref rm,
+                                              const std::vector<instruction_ref>& splits)
+{
+    std::vector<instruction_ref> result;
+    bool stop = false;
+    liveness(*rm, [&](auto ins, const auto& live_set) {
+        if(stop)
+            return;
+        if(not contains(splits, ins))
+            return;
```
umangyadav, line 118: Why do you require a stop flag ? 
`if(not contains(splits,ins))` would take care of early return.

src/split_reduce.cpp
```diff
@@ -0,0 +1,208 @@
...
+static std::string assign_op(const std::vector<instruction_ref>& splits)
+{
+    static std::unordered_map<std::string, std::string> m = {
+        {"reduce_sum", "assign_add"},
+        {"reduce_mean", "assign_add"},
+        {"reduce_prod", "assign_mul"},
+        {"reduce_max", "assign_max"},
+        {"reduce_min", "assign_min"},
+    };
+    return m.at(splits.front()->name());
```
umangyadav, line 137: Doesn't it need to return vector as the output as the inputs are also vector ?

src/targets/gpu/include/migraphx/gpu/hip.hpp
```diff
@@ -81,6 +83,30 @@ struct hip_allocate
...
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.value, "value"));
+    }
+
+    std::string name() const { return "hip::fill"; }
+    shape compute_shape(const std::vector<shape>& inputs) const
+    {
+        check_shapes{inputs, *this}.has(1);
```
umangyadav, line 99: assert for packed tensor. As hipMemsetAsync can't handle non-packed

src/targets/gpu/hip.cpp
```diff
@@ -293,6 +302,12 @@ argument get_preallocation(context& ctx, const std::string& id)
     return ctx.get_current_device().preallocations.at(id);
 }
 
+void gpu_fill(context& ctx, const argument& dst, int value)
+{
+    // TODO: Handle non-packed tensor when value is not 0
```
umangyadav, line 307: What's special about value 0 for non-packed tensor such that it would work ? I thought it wouldn't work even for the zero.

src/targets/gpu/target.cpp
```diff
@@ -77,6 +78,7 @@ namespace gpu {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_SCHEDULE_PASS)
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_REDUCE_FUSION)
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_SPLIT_REDUCE)
```
umangyadav, line 81: add this variable in env var docs

src/module.cpp
```diff
@@ -732,6 +751,194 @@ void module::finalize(std::vector<context>& contexts)
...
+    }
+    else
+    {
+        std::transform(
+            params.begin(),
+            params.end(),
+            inputs.begin(),
+            std::inserter(result, result.end()),
+            [&](instruction_ref param, auto input) { return std::make_pair(input, param); });
+    }
```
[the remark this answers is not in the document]
pfultz2: The parameters are flipped in the `make_pair` call. The reverse will reverse the lookup so instead of `ins -> param_ins` it will be `param_ins -> ins`.

src/module.cpp
```diff
@@ -732,6 +751,194 @@ void module::finalize(std::vector<context>& contexts)
...
+}
+void module_with_inputs::replace(const std::vector<instruction_ref>& keys,
+                                 const std::vector<instruction_ref>& values)
+{
+    for(auto& ins : inputs)
+    {
+        auto it = std::find(keys.begin(), keys.end(), ins);
+        if(it == keys.end())
+            continue;
+        ins = values[it - keys.begin()];
```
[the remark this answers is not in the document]
pfultz2: > This requires implicit ordering inside vector of keys and values.

Yes both vectors need to be in the same order(and they are in the cases that are already used).

> Can they be made a map instead ?

There is already a map overload. I dont see why I need to make a copy to make it a map.

src/split_reduce.cpp
pfultz2: We only want the liveness for the first split found. The `return` is like a `continue` in the loop and the `stop` flag is like a `break`.

src/split_reduce.cpp
pfultz2: Yes but for now we dont support multiple outputs. This will switch to a vector when we support the multiple outputs.

src/targets/gpu/hip.cpp
pfultz2: Yea it wont work for zero either.

pfultz2: Because its writing a float type but using an integer value. So a value of `1` as `int` wont be a value of `1` as a `float`.

src/include/migraphx/module.hpp
```diff
@@ -127,38 +129,38 @@ struct MIGRAPHX_EXPORT module
...
-                     inserter insert                                              = nullptr);
+                     std::unordered_map<instruction_ref, instruction_ref>* map_ins = nullptr,
+                     inserter insert                                               = nullptr);
 
     std::vector<instruction_ref>
     insert_instructions(instruction_ref ins,
                         const std::vector<instruction_ref>& instructions,
-                        std::unordered_map<instruction_ref, instruction_ref> map_ins = {},
-                        inserter insert                                              = nullptr);
+                        std::unordered_map<instruction_ref, instruction_ref>* map_ins = nullptr,
```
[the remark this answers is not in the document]
pfultz2: > Is there a good reason why this need to be made a pointer now ?

You cant default intialize a lvalue reference. 

> map_ins does it have any use after calls to insert or add_instructions ?

Yes it reuses the map in the `split` function. It also reuses it in fuse_reduce.

src/include/migraphx/liveness.hpp
```diff
@@ -0,0 +1,77 @@
...
+            {
+                auto i = instruction::get_output_alias(input);
+                // Skip if variable comes from parent
+                if(not m.has_instruction(i))
+                    continue;
+                live_set.insert(i);
+            }
+        };
+        add_live_variables(ins->inputs());
+        add_live_variables(implicit_deps[ins]);
```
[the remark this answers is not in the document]
pfultz2: > this call may not necessary.

Not for split_reduce since the submodules dont reference the parent, but it is necessary for memory_coloring because there are submodules which will reference the parent(such as when using loop or if).

src/include/migraphx/liveness.hpp
```diff
@@ -0,0 +1,77 @@
...
+template <class F>
+void liveness(const module& m, F f)
+{
+    auto implicit_deps = m.calc_implicit_deps();
+    std::unordered_set<instruction_ref> live_set;
+    auto rp = reverse(m);
+    for(auto rins : iterator_for(rp)) // NOLINT
+    {
+        // The base iterator is one ahead, so we need to use the previous iterator
+        auto ins = std::prev(rins.base());
```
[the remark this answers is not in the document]
pfultz2: Yes because the base iterator is one ahead. That is because the reverse iterator is passed the `end()` iterator for the start iterator.

src/split_reduce.cpp
umangyadav: > and the stop flag is like a break. 

Yeah but i don't see need to have `stop` flag.

src/targets/gpu/include/migraphx/gpu/hip.hpp
umangyadav: Can you check for it ?

src/split_reduce.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+void split_reduce::apply(module_pass_manager& mpm) const
+{
+    for(auto ins : iterator_for(mpm.get_module()))
+    {
+        if(ins->name() != "fused_reduce")
+            continue;
+        auto* rm = ins->module_inputs().front();
+        if(get_reduce_size(rm) < split_size)
+            continue;
```
umangyadav, line 162: So from what i get it would split reduction module into multiple modules if reduction size > 8192. 

1. Why dont' we enforce that during `fuse_reduce` pass itself ? 
2. How does this split reduce pass help with performance ?

umangyadav (commented):
I consider this as a Large PR. I think it can be broken down to following pieces. 

1. Changes for the Param Utils and liveness
2. Adding hip::fill operator
3. Adding Generic Split functionality for the module and adding tests for those (they are missing right now). 
4. Adding Split_reduce pass

src/include/migraphx/module.hpp
```diff
@@ -205,6 +209,28 @@ struct MIGRAPHX_EXPORT module
...
+    std::unordered_map<instruction_ref, instruction_ref>
+    get_ins_param_map(const std::vector<instruction_ref>& inputs, bool reverse = false) const;
+
+    using with_inputs = module_with_inputs;
+
+    /// This will split the module into two parts at the instruction splits.
+    /// Each split instruction becomes an input parameter in the second
+    /// module. As such the inputs instructions to the second module will use
+    /// the split instructions as input placeholders that can be replaced
+    /// later.
```
umangyadav, line 224: Thanks for this. It helped understanding what it is trying to do instead of guessing what it is doing. 

We should add unit-tests for this

src/targets/gpu/kernels/include/migraphx/kernels/scatter_reduction_modes.hpp
umangyadav: https://rocm.docs.amd.com/projects/HIP/en/latest/reference/kernel_language.html#unsafe-floating-point-atomic-rmw-operations

From the description it seems it can produce incorrect results in certain cases. How do we gurantee correct results ?

umangyadav:
> I consider this as a Large PR. I think it can be broken down to following pieces.
> 
> 1. Changes for the Param Utils and liveness
> 2. Adding hip::fill operator
> 3. Adding Generic Split functionality for the module and adding tests for those (they are missing right now).
> 4. Adding Split_reduce pass

I would like it better if we can break down this PR

src/targets/gpu/jit/reduce.cpp
```diff
@@ -131,6 +133,61 @@ static std::size_t compute_subwave_size(context& ctx, std::size_t n)
     return wavefront_size;
 }
 
+static std::vector<shape> split_reduce(const std::vector<shape>& inputs,
```
umangyadav, line 136: Can you add docstring explaining what it is supposed to do ?

src/include/migraphx/liveness.hpp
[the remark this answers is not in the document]
umangyadav: > , but it is necessary for memory_coloring because there are submodules which will reference the parent 

Yeah but they are skipped inside the loop
```
                // Skip if variable comes from parent
                if(not m.has_instruction(i))
                    continue;
```

src/split_reduce.cpp
pfultz2: Then how else can I `break` on the first split?

src/targets/gpu/include/migraphx/gpu/hip.hpp
pfultz2: I added an `assert` in the `hip_fill` function.

src/targets/gpu/hip.cpp
pfultz2: > It writes Int value

It does write the int value, but if the type is float we dont want an in value of 1, we want a float value of 1(which has a different byte representation).

src/split_reduce.cpp
pfultz2: > Why dont' we enforce that during fuse_reduce pass itself ?

We want to fuse as much as possible first, otherwise we will end up even more kernels. 

> How does this split reduce pass help with performance ?

Each reduction is done within a workgroup, for a large reduction its much better to reduce across multiple workgroups which means we cannot fuse operators afterwards as there is no global sync.

I do see an almost 2x improvement with very large reduction sizes. So there is definitely a perf improvement for this.

src/include/migraphx/liveness.hpp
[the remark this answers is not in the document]
pfultz2: > Yeah but they are skipped inside the loop

Yes, its not the same thing. The implict deps do not calculate the dependencies of the parent module. It calculates the dependencies of the submodules of an instruction that reference back to the original module.

pfultz2:
> > I consider this as a Large PR. I think it can be broken down to following pieces.
> > 
> > 1. Changes for the Param Utils and liveness

Why would I create PR to move functions to separate headers when they are only used once?

> > 2. Adding hip::fill operator

Why would I create a PR for an operator that is not used anywhere?

> > 3. Adding Generic Split functionality for the module and adding tests for those (they are missing right now).
> > 4. Adding Split_reduce pass
> 
> I would like it better if we can break down this PR

This will slow down getting splitk reduce implemented as there are a lot more changes to come for this. I stopped here to avoid having a very large PR and avoiding dealing with merge conflicts.

src/include/migraphx/module.hpp
pfultz2: Yea I will add unit tests for this.

src/include/migraphx/liveness.hpp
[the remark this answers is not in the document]
umangyadav: > . It calculates the dependencies of the submodules of an instruction that reference back to the original module. 

Since submodule instruction is using instruction from original/parent module, 
`if(not m.has_instruction(i))` this should be true always.

src/split_reduce.cpp
umangyadav: > Then how else can I break on the first split? 

You don't need to break, `if(not contains(splits, ins))` would break it for you after the first split.

src/split_reduce.cpp
umangyadav: > We want to fuse as much as possible first, otherwise we will end up even more kernels.

But can't we break out from doing fusion if it encounters `relements >= 8192` ?

src/targets/gpu/jit/reduce.cpp
```diff
@@ -301,13 +361,14 @@ struct fused_reduce_compiler : compiler<fused_reduce_compiler>
...
-             {"transformers", make_transformer_args(vec)},
-             {"preamble", v.get("preamble", std::string{})}});
+                        {"params", enum_params(inputs.size(), "void * private_p")},
+                        {"args", enum_params(inputs.size(), "private_p")},
+                        {"assign", assign},
+                        {"algo", algo},
+                        {"reduced", "decltype(" + generate_make_shape(reduce_output_shape) + ")"},
+                        {"lambda", v.at("lambda").to<std::string>()},
+                        {"transformers", make_transformer_args(vec)},
+                        {"preamble", v.get("preamble", std::string{})}});
```
TedThemistokleous, line 371: Nitpick: Not sure why all of this is shifted instead of one change for adding assign.

src/targets/gpu/hip.cpp
```diff
@@ -246,6 +247,14 @@ void gpu_sync()
 
 void gpu_sync(const context& ctx) { ctx.finish(); }
 
+void hip_async_memset(context& ctx, const argument& dst, int value)
+{
+    std::size_t dst_size = dst.get_shape().bytes();
+    auto status          = hipMemsetAsync(dst.data(), value, dst_size, ctx.get_stream().get());
```
TedThemistokleous, line 253: Do we not need to add any sort of event syncronization or are we assuming things are handled by using the context's stream, other items will sync to the end of this MemSet?

src/split_reduce.cpp
```diff
@@ -0,0 +1,208 @@
...
+        auto splits = find_split(rm);
+        if(splits.empty())
+            continue;
+        // Only use split reduce with float for now
+        if(not std::all_of(splits.begin(), splits.end(), [](instruction_ref split) {
```
TedThemistokleous, line 167: Why only use for float? If to reduce scope for now, add a TODO here.

src/split_reduce.cpp
```diff
@@ -0,0 +1,208 @@
...
+               return split->get_shape().type() == shape::float_type;
+           }))
+            continue;
+        auto v    = ins->get_operator().to_value();
+        auto axes = v["axes"].to_vector<std::int64_t>();
```
TedThemistokleous, line 172: Add coverage for continues; non float and empty splits() above.

TedThemistokleous (changes requested):
Some more comments. One question regarding the use of hipMemsetAsync

I understand we want this in sooner than later so if we're reducing scope for now, add in TODO's

Also add in test coverage for additional early returns/continue

src/split_reduce.cpp
pfultz2: The atomic instruction works directly with float. It doesnt work for half, it only works for half2, but to support that will need some non-trivial work.

src/targets/gpu/hip.cpp
pfultz2: The memset should be enqued on the stream so we dont need any extra synchronization(the queue for the stream is always in order).

src/include/migraphx/liveness.hpp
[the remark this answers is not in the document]
pfultz2: > if(not m.has_instruction(i)) this should be true always.

No, an instruction can have input that references a parent module(not in the case of split_reduce since those operators dont rely on lexical scoping, but it is the case when using other operators that do allow lexical scoping such as `if` or `loop`).

src/split_reduce.cpp
pfultz2: Yea I do need a break. I only want the instructions live at the first split found(which is the last split instructions in the module due to how liveness traverses in reverse). I dont care whats live on any previous splits.

If I didnt break then I would be splitting the module at unnecessary points(also there would be duplicates for instructions live for both).

Right now we dont handle multiple outputs so there is only one split so it wouldn't make a difference, but the goal is to support multiple outputs which will require multiple split points.

src/split_reduce.cpp
TedThemistokleous: understood.

src/split_reduce.cpp
pfultz2: We will miss fusions that way. Layernorm has two reductions that later elementwise operators use. If we skip the fusions we will have two seperate reduction kernels, instead of having one kernel with multiple outputs for both reductions.

There could also be other misses with the fusions as we fuse more as more is fused together.

src/split_reduce.cpp
pfultz2: I would prefer not to add tests for those cases, since I hope to support those other cases in the future, and will add test cases for that then.

TedThemistokleous: Then add a TODO there or make it an issue we can track for later improvement.

TedThemistokleous (approved):
Fix format and windows build.

Blue ocean is all green. Not sure why its not reporting correctly here though.

src/targets/gpu/jit/reduce.cpp
pfultz2: Added the docstring.

src/targets/gpu/kernels/include/migraphx/kernels/scatter_reduction_modes.hpp
umangyadav: @pfultz2  `unsafeAtomicAdd` HIP function it seems is only available for float and double.  Does this function require a `std::enable_if` to use different overload for integers and other dtypes ?

src/targets/gpu/jit/reduce.cpp
```diff
@@ -131,6 +133,67 @@ static std::size_t compute_subwave_size(context& ctx, std::size_t n)
     return wavefront_size;
 }
 
+/// This will adjust the input shapes so a partial reduction is done per workgroup.
+/// This is done by splitting the reduction axis so each split group becomes
+/// part of the batch. So if we want to do a split redution of a tensor
+/// {K}, then this will create a tensor of {K/N, N} where N is the number of
+/// split groups. To compute the number of split groups it finds the largets
```
umangyadav, line 140:
```suggestion
/// split groups. To compute the number of split groups it finds the largest
```

src/targets/gpu/kernels/include/migraphx/kernels/scatter_reduction_modes.hpp
pfultz2: > From the description it seems it can produce incorrect results in certain cases. How do we guarantee correct results when using "unsafe"atomic add here in MIGraphX ?

This is only a problem when doing atomics across devices, which we dont do. 

> unsafeAtomicAdd HIP function it seems is only available for float and double. Does this function require a std::enable_if to use different overload for integers and other dtypes ? Or perhaps do an assert.

I will probably need to add an `if constexpr` to check for that and then call the safe version. However, it would still be good to add an assert for debug runs as using the CAS is extremely slow.

umangyadav: > I will probably need to add an if constexpr to check for that and then call the safe version. However, it would still be good to add an assert for debug runs as using the CAS is extremely slow. 

Can you add that ?

umangyadav (approved):

src/targets/gpu/hip.cpp
umangyadav: Would be nice to include some verify tests

PR #2909, opened 2024-03-20
Add support for GridSample operator

gyulaz-htec:
Add support for [GridSample](https://github.com/onnx/onnx/blob/main/docs/Operators.md#gridsample) onnx operator.
Currently the following feature set is supported:
* inputs: 4-D inputs
* align_corners: both `0` and `1`
* mode: `linear` and `nearest`
* padding_mode: `zeros`, `border` and `reflection`
Fixes: https://github.com/ROCm/AMDMIGraphX/issues/2923


gyulaz-htec:
@causten  The PR is ready for review
This sould not be marged yet, because I've found an accuracy issue with `fuse_pointwise` pass with the GPU target when computing GridSample results (with linear inetrploation).
Collected my findings on that issue in https://github.com/ROCm/AMDMIGraphX/issues/2923

gyulaz-htec:
I had to exclude `logical_and` and `where` operators from pointwise fusion, so the GPU tests are passing now.
https://github.com/ROCm/AMDMIGraphX/issues/2923 is fixed by that.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    instruction_ref sample(const onnx_parser::node_info& info) override
+    {
+        setup(info);
+        std::vector<instruction_ref> indices;
+        std::vector<instruction_ref> validation;
+        static auto nhw_shape = migraphx::shape{migraphx::shape::int64_type, {3}};
+        bool validate         = not has_border_padding();
+        for(size_t n = 0; n < m_batch; n++)
```
TedThemistokleous, line 234: Get these to run for pre increment here for these for loops since worst case they're  O(n^4) (N,C, H, W)

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+                    auto nhw = info.add_literal(migraphx::literal{nhw_shape, {n, h, w}});
+                    auto h_t = info.add_instruction(make_op("gathernd"), m_round_y, nhw);
+                    auto w_t = info.add_instruction(make_op("gathernd"), m_round_x, nhw);
+                    for(size_t c = 0; c < m_channel; c++)
+                    {
+                        update_indices(info, h_t, w_t, n, c, indices, validation, validate);
+                    }
+                }
+            }
+        }
```
TedThemistokleous, line 249: I think this can be rewritten and run in parallel here per sample for each batch + height combination instead of done in serial.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+                    for(size_t c = 0; c < m_channel; c++)
+                    {
+                        update_indices(info, y0, x0, n, c, indices_all.at(0), validation_all.at(0));
+                        update_indices(info, y0, x1, n, c, indices_all.at(1), validation_all.at(1));
+                        update_indices(info, y1, x0, n, c, indices_all.at(2), validation_all.at(2));
+                        update_indices(info, y1, x1, n, c, indices_all.at(3), validation_all.at(3));
+                    }
+                }
+            }
+        }
```
TedThemistokleous, line 334: Same comment as above to avoid O(n^4). There seems like we're only doing ops after finding batch, height, width.

TedThemistokleous (changes requested):
- [ ] Add cubic support + onnx backend coverage
- [ ] Add additional test cases for coverage
- [ ] Fix loop to pre increment
- [ ] Look into alternative to avoid O(n^4)  updates and ops.

test/py/onnx_backend_test.py
```diff
@@ -396,14 +396,7 @@ def disabled_tests_onnx_1_10_0(backend_test):
 def disabled_tests_onnx_1_11_0(backend_test):
     # errors
     # from OnnxBackendNodeModelTest
-    backend_test.exclude(r'test_gridsample_aligncorners_true_cpu')
     backend_test.exclude(r'test_gridsample_bicubic_cpu')
```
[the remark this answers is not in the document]
gyulaz-htec: Adding bicubic sample would complicate things further because it would require more computation steps  (e.g. tvm bicubic interpolation is more complex than the other two: https://github.com/apache/tvm/blob/main/python/tvm/topi/image/grid_sample.py#L208-L277 ), and I don't wanted to further bloat this PR, it is already around 2k line changes.
I'm happy to add it in a follow up PR.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+            align_corners = parser.parse_value(info.attributes.at("align_corners")).at<bool>();
+        }
+
+        if(contains(info.attributes, "mode"))
+        {
+            mode = info.attributes.at("mode").s();
+            // Note: can be cubic or bicubic depending on the onnx version
+            if(contains(mode, "cubic"))
+            {
+                MIGRAPHX_THROW("PARSE_GRID_SAMPLE: cubic mode is not supported");
```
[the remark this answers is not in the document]
gyulaz-htec: I've already created one test case for this in `test\onnx\parse`: https://github.com/ROCm/AMDMIGraphX/pull/2909/files#diff-370f8944dad2d8d0077880b181893bd879ecf4830803611ffd33cf09305d67cc
I have no clue why codecov doesn't recognize it.

src/onnx/parse_gridsample.cpp
gyulaz-htec: @TedThemistokleous could you please elaborate more on this. I tried `par_dfor`  to run in parellel for batch+height combination as you suggested, but it fails to compile. From what I can see `add_instruction` calls are not suitable to be executed in parallel, it produces invalid instructions because of the race condition.

gyulaz-htec: As I can see `par_dfor` is used in runtime constructions but this is a compile time problem. Is there any solution for compile time parallelism?

src/onnx/parse_gridsample.cpp
gyulaz-htec: Maybe I don't get it right, but are you hinting that we should iterate over touples of possible combinations instead of the nested for loops?

src/onnx/parse_gridsample.cpp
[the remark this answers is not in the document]
gyulaz-htec: It turned out that the test onnx file was failing with a parse error instead of this. Not sure if it's the cause of the codecov not recognizing. I will update that when we have access to the rack again.

test/py/onnx_backend_test.py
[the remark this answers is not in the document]
gyulaz-htec: Created https://github.com/ROCm/AMDMIGraphX/issues/2960 to track cubic interpolation

src/onnx/parse_gridsample.cpp
TedThemistokleous: Start at -1 and change your loops to perform ++n vs n++
 
`( size_t n=-1; n < m_batch; ++n)`

When you n++ you create a temp variable so nesting this you're doing a bunch of extra create/copy/destroy ops on the background.

From in notes: https://en.cppreference.com/w/cpp/language/operator_incdec

_Because a temporary copy of the object is constructed during post-increment and post-decrement, pre-increment or pre-decrement operators are usually more efficient in contexts where the returned value is not used._

pfultz2: > When you n++ you create a temp variable so nesting this you're doing a bunch of extra create/copy/destroy ops on the background.

For trivial types this shouldn't matter, the compiler will always optimize out the unused temporary.

src/onnx/parse_gridsample.cpp
pfultz2: I dont think par_dfor can be used, but `dfor` can be used to get rid of some of the nesting.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+    instruction_ref m_one_l;
+    instruction_ref m_two_l;
+    instruction_ref m_zero_l;
+    instruction_ref m_minus_half_l;
+    instruction_ref m_width_l;
+    instruction_ref m_width_max_l;
+    instruction_ref m_height_l;
+    instruction_ref m_height_max_l;
+    instruction_ref m_unnorm_x;
+    instruction_ref m_unnorm_y;
```
pfultz2, line 57: These variables are not initialized in the constructor.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+        std::for_each(
+            std::next(instructions.begin()), instructions.end(), [&info, &ret](auto& ins) {
+                ret = info.add_instruction(make_op("concat", {{"axis", 0}}), ret, ins);
+            });
+        return ret;
+    }
+
+    inline bool has_border_padding() const { return m_padding == "border"; }
+
+    virtual instruction_ref sample(const onnx_parser::node_info& info) = 0;
```
pfultz2, line 204: Remove `virtual`, this function is never called from the base class.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+        m_batch      = i_lens.at(0);
+        m_channel    = i_lens.at(1);
+        m_in_height  = i_lens.at(2);
+        m_in_width   = i_lens.at(3);
+        auto g_lens  = grid->get_shape().lens();
+        m_out_height = g_lens.at(1);
+        m_out_width  = g_lens.at(2);
+    }
+
+    virtual ~grid_sampler() {}
```
pfultz2, line 75: Remove this.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+        cond                    = info.add_common_op("equal", cond, m_zero_l);
+        auto extra              = info.add_common_op("mul", size_times, size);
+        extra                   = info.add_common_op("sub", index_align_corner, extra);
+        auto cond_true          = info.add_common_op("add", extra, corner_start);
+        auto cond_false         = info.add_common_op("sub", size, extra);
+        cond_false              = info.add_common_op("add", cond_false, corner_start);
+        return info.add_common_op("where", cond, cond_true, cond_false);
+    }
+
+    virtual void setup(const onnx_parser::node_info& info)
```
pfultz2, line 97: Remove `virtual` here as this is not called from the base class.

pfultz2: Actually this should be moved to the constructor.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+        m_ceil_y               = info.add_common_op("add", m_floor_y, m_one_l);
+        auto fract_x           = info.add_common_op("sub", m_unnorm_x, m_floor_x);
+        auto fract_y           = info.add_common_op("sub", m_unnorm_y, m_floor_y);
+        auto one_minus_fract_x = info.add_common_op("sub", m_one_l, fract_x);
+        auto one_minus_fract_y = info.add_common_op("sub", m_one_l, fract_y);
+        m_corner_weights[0]    = info.add_common_op("mul", one_minus_fract_y, one_minus_fract_x);
+        m_corner_weights[1]    = info.add_common_op("mul", one_minus_fract_y, fract_x);
+        m_corner_weights[2]    = info.add_common_op("mul", fract_y, one_minus_fract_x);
+        m_corner_weights[3]    = info.add_common_op("mul", fract_y, fract_x);
+    }
```
pfultz2, line 303: This should be moved to the constructor.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+    instruction_ref sample(const onnx_parser::node_info& info) override
+    {
+        setup(info);
+        std::array<std::vector<instruction_ref>, 4> indices_all;
+        std::array<std::vector<instruction_ref>, 4> validation_all;
+        std::vector<instruction_ref> weight_indices;
+
+        static auto nhw_shape = migraphx::shape{migraphx::shape::int64_type, {3}};
```
pfultz2, line 312: Add `const` to this.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+    void update_indices(const onnx_parser::node_info& info,
+                        const instruction_ref& h,
+                        const instruction_ref& w,
+                        size_t n,
+                        size_t c,
+                        std::vector<instruction_ref>& indices,
+                        std::vector<instruction_ref>& validation,
+                        bool validate = true) const
+    {
+        static auto nc_shape = migraphx::shape{m_grid->get_shape().type(), {2}};
```
pfultz2, line 173: Dont use `static` for this variable as it relies on member variables.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    static instruction_ref concat_on_first_dim(const onnx_parser::node_info& info,
+                                               std::vector<instruction_ref> instructions)
+    {
+        if(instructions.empty())
+            MIGRAPHX_THROW("PARSE_GRID_SAMPLE: can't concatenate empty list of instructions");
+
+        auto ret = instructions.at(0);
+        std::for_each(
```
pfultz2, line 195: Use `std::accumulate`.

gyulaz-htec:
@TedThemistokleous @pfultz2 I've adressed the majority of you commets except the ` Look into alternative to avoid O(n^4) updates and ops.` part.

gyulaz-htec:
@TedThemistokleous @pfultz2 I've managed to move out the op creation and updates outside the O(n^4) part, so the heavy lifting is done outside of that now. The remaining code there is only for generating literals for `gathernd` indices.

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,433 @@
...
+            info.add_common_op("logical_and", x0_validation, y0_validation),
+            info.add_common_op("logical_and", x1_validation, y0_validation),
+            info.add_common_op("logical_and", x0_validation, y1_validation),
+            info.add_common_op("logical_and", x1_validation, y1_validation)};
+
+        std::vector<instruction_ref> weighted_corners;
+        auto weight_index_t = concat_on_first_dim(info, weight_indices);
+        weight_index_t      = info.add_instruction(
+            make_op("reshape", {{"dims", {weight_indices.size(), 3}}}), weight_index_t);
+        dfor(4)([&](auto corner) {
```
gyulaz-htec, line 346: use `std::transform` here

src/onnx/parse_gridsample.cpp
```diff
@@ -0,0 +1,433 @@
...
+            auto indices = corner_indices.at(corner);
+            auto samples = info.add_instruction(make_op("gathernd"), m_input, indices);
+            samples = info.add_common_op("where", corner_validations.at(corner), samples, m_zero_l);
+            auto weights = info.add_instruction(
+                make_op("gathernd"), m_corner_weights.at(corner), weight_index_t);
+            weighted_corners.push_back(info.add_instruction(make_op("mul"), samples, weights));
+        });
+
+        auto samples = weighted_corners.at(0);
+        std::for_each(std::next(weighted_corners.begin()),
```
gyulaz-htec, line 356: use `std:accumulate` here

src/fuse_pointwise.cpp
```diff
@@ -204,12 +205,17 @@ struct pointwise_reshape : rewrite_reshapes_base
...
+    size_t pointwise_cnt = 0;
+    if(disable_fusion)
     {
+        create_pointwise_modules(mpm, pointwise_cnt, {"layout"});
+        mpm.run_pass(dead_code_elimination{});
         return;
     }
+    // Excluding `logical_and` and `where` from fusion
+    // to fix GPU accuarcy of GridSample: https://github.com/ROCm/AMDMIGraphX/issues/2923
+    create_pointwise_modules(mpm, pointwise_cnt, {"layout", "logical_and", "where"});
```
pfultz2, line 217: This has major perf problems. I dont think we should be skipping fusions. Instead we should fix the GPU kernel.

gyulaz-htec: @pfultz2 GPU kernel debuging is unknown area for me. Can you give me som pointers regaridng how can I start debugging this?

pfultz2: I added some comments to the issue.

gyulaz-htec:
@pfultz2 I've removed the fuse pointwise related changes from the PR in favor of https://github.com/ROCm/AMDMIGraphX/pull/3054

kahmed10 (approved):

kahmed10:
I would like to see additional documentation and possible refactoring since the parse_gridsample.cpp is quite large. Not blocking for merge.

PR #2926, opened 2024-03-26
fix start index of statistic arrays

jasberc:
This fixes corruption of "python resnet50.py" with "--run 1" and correct statistic results.

### DIFF
--- a/examples/onnxruntime/resnet50/resnet50.py
+++ b/examples/onnxruntime/resnet50/resnet50.py
@@ -263,12 +263,12 @@ def main():
 
     if flags.QPS:
         print("resnet50, Rate = {} QPS ".format(
-            format((((flags.batch) / (sum(latency[1:]) / len(latency[1:])))),
+            format((((flags.batch) / (sum(latency[0:]) / len(latency[0:])))),
                    '.2f')))
 
     else:
         print("resnet50, Averaage Execution time = {} ms".format(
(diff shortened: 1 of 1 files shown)


examples/onnxruntime/resnet50/resnet50.py
```diff
@@ -263,12 +263,12 @@ def main():
 
     if flags.QPS:
         print("resnet50, Rate = {} QPS ".format(
-            format((((flags.batch) / (sum(latency[1:]) / len(latency[1:])))),
+            format((((flags.batch) / (sum(latency[0:]) / len(latency[0:])))),
```
umangyadav, line 266: Good find. 
I assume the reason why it started with `[1:]` was that it considered first run as the warmup run. 
If you doing `--run 1` it wouldn't give you the most accurate performance numbers. 

I think it is better to do warm up runs as a separate calls.

TedThemistokleous: Yes, we needed the warmup run since there was overhead during setup. 

This is required as this step also includes the additional run. A better approach here would be to set the amount of runs to `run 1` + 1 so that you still do a warmup run, but run two runs under the hood.

TedThemistokleous:
@jasberc  why was this closed?

jasberc:
> @jasberc why was this closed?

it's reformed in https://github.com/ROCm/AMDMIGraphX/pull/2928 and merged.

PR #2928, opened 2024-03-27
fix corruption when using "--run 1"

jasberc:
set a warm-up run outside of estimation to avoid crashing with "--run 1".


umangyadav:
@jasberc  can you fix formatting error that shows up in Github CI check ? 

you can run `python3 tools/format.py -i` locally but you need to use MIGraphX's Dev Docker, else you can manually fix format.

causten (approved):

umangyadav (approved):

PR #2938, opened 2024-04-01
add changes to update time loop for split-k GEMM 

umangyadav:
Fixes https://github.com/ROCm/AMDMIGraphX/issues/2897


src/targets/gpu/include/migraphx/gpu/compiler.hpp
```diff
@@ -45,18 +45,27 @@ struct compiler_replace
...
+    compiler_replace(const std::vector<operation>& op, F f)
         : code_object{op},
           replace_fn([=](const compiler_replace& cr, module& m, instruction_ref ins) {
               f(m, ins, cr.code_object);
           })
     {
     }
 
-    operation code_object = {};
+    std::vector<operation> code_object = {};
```
pfultz2, line 68: This should be renamed to `code_objects` as well to make it clear it can be multiple code objects.

src/targets/gpu/compile_ops.cpp
```diff
@@ -184,8 +184,15 @@ struct compile_plan
...
                                return std::numeric_limits<double>::max();
                            }
-                           auto t = time_op(
-                               *ctx, cr->replace.code_object, to_shapes(cr->ins->inputs()), 20);
+                           // Time all the code objects for a given perf config and calculate total
+                           // time e.g. in case of split-K GEMM, it may or may not support fusion.
+                           // In that case MLIR compile would return code objects for individual
+                           // GEMM and pre/post fusion code objects.
+                           auto cobjs = cr->replace.code_object;
+                           double t   = std::accumulate(
```
pfultz2, line 192: I think it would be better to use `transform_accumulate`(its under the `migraphx/algorithm.hpp` header).

pfultz2 (approved):

PR #2940, opened 2024-04-02
fix conv_concat matcher when original convolutions are grouped

shivadbhavsar:
The original `find_concat_conv` pass from #2639  has a bug when it matches with original convolutions that have group != 1. This is a fairly simple change to handle that case.


umangyadav (approved):

TedThemistokleous (commented):
Looks good. Will approve once CI finishes

causten:
@shivadbhavsar  can you test performance before and after on a couple of the models listed above that dropped in performance.

PR #2942, opened 2024-04-03
Transpose -> GEMM issue

mirza-halilcevic:
- Fix issue with gemm not detecting transpose in cases where the transpose does not affect the strides
- Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/183


src/targets/gpu/gemm_impl.cpp
```diff
@@ -129,7 +129,27 @@ auto rocblas_invoke(F f, Pack p, Ts... xs)
...
+    if(s.transposed())
+    {
+        return s.strides().back() != 1;
+    }
+    else if(s.strides() != s.as_standard().strides())
+    {
+        // If the strides are not standard, check if there's a permutation that produces those
+        // strides
+        auto lens_perm = s.lens();
+        std::sort(lens_perm.begin(), lens_perm.end());
```
pfultz2, line 143: Why not call `find_permutation`? That should give the correct permutation without needing the loop.

src/targets/gpu/gemm_impl.cpp
```diff
@@ -22,6 +22,7 @@
  * THE SOFTWARE.
  */
 
+#include "migraphx/permutation.hpp"
```
pfultz2: Use angle brackets for includes.

src/targets/gpu/gemm_impl.cpp
mirza-halilcevic: It seems like I misunderstood the implementation of `find_permutation` at first glance. I see now the purpose behind it. I changed the implementation accordingly.

test/verify/test_gemm_transposeb_detect.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+template <migraphx::shape::type_t DType>
+struct test_gemm_transposeb_detect : verify_program<test_gemm_transposeb_detect<DType>>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        auto a   = mm->add_parameter("a", migraphx::shape{DType, {2, 2, 1}});
+        auto b   = mm->add_parameter("b", migraphx::shape{DType, {2, 2, 1}});
```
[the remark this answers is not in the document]
umangyadav: You would need to increase `k > 1024` to make it use rocBLAS by default.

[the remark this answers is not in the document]
mirza-halilcevic: `k` needs to be 1 for this test to recreate the issue. Is there another way to force it to use rocBLAS?

pfultz2 (approved):

PR #2943, opened 2024-04-03
Multibroadcast -> GEMM issue

mirza-halilcevic:
- Fix issue with eliminate_contiguous removing a contiguous after multibroadcast when it's an input to gemm
- Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/184


src/eliminate_contiguous.cpp
```diff
@@ -51,6 +51,14 @@ static bool try_compute_shape(instruction_ref ins,
             return false;
         }
 
+        // If the instruction is a gemm and any of the inputs is broadcasted, the input needs to be
+        // made contiguous
+        if(contains(ins->name(), "gemm") and
+           any_of(inputs, [](const auto& sh) { return sh.broadcasted(); }))
+        {
+            return false;
+        }
```
pfultz2, line 60: This logic should be handled in the `compute_shape` method of the gemm operator. It should throw an error for this case if it can't handle such shapes.

mirza-halilcevic: Done.

src/targets/gpu/include/migraphx/gpu/gemm.hpp
```diff
@@ -75,6 +75,7 @@ struct rocblas_gemm
 
     shape compute_shape(const std::vector<shape>& inputs) const
     {
+        check_shapes{inputs, *this}.not_broadcasted();
```
pfultz2, line 78: This is too broad and will have a negative impact on performance.

mirza-halilcevic: I missed that. I modified the solution to only check the last two dimensions. I also moved the check to `blas_shape`, which I believe is a better place for it than directly in `compute_shape`.

pfultz2 (approved):

test/verify/test_gemm_multibroadcast.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <migraphx/make_op.hpp>
+
+template <migraphx::shape::type_t DType>
+struct test_gemm_multibroadcast : verify_program<test_gemm_multibroadcast<DType>>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        auto a   = mm->add_parameter("a", migraphx::shape{DType, {2, 2, 2}});
```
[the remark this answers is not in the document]
umangyadav: You would need to modify `k` dimension to be greater than 1024 to make it use rocMLIR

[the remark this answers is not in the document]
mirza-halilcevic: Thanks for the hint. I ran the test by manually disabling MLIR. I changed the dims and now it should always use rocBLAS.

umangyadav (approved):

PR #2947, opened 2024-04-03
Bump MLIR for split-k support

causten:


TedThemistokleous (approved):

umangyadav (approved):
I don't think MLIR has fully exposed Split-k gemm support yet. but anyways good to bump SHA

PR #2950, opened 2024-04-04
Fixing mxrfiles Windows

ivarusic-amd:
By adding an additional flag to open the output file in binary mode, we have resolved an issue where .mxr files created on Windows couldn't be opened correctly.


umangyadav (approved):

src/driver/main.cpp
```diff
@@ -396,7 +396,7 @@ struct loader
         std::ofstream fs;
         if(not output.empty())
         {
-            fs.open(output);
+            fs.open(output, std::ios::binary);
```
pfultz2, line 399: Will this cause problems when we output as json or python?

ivarusic-amd: I checked for json, it's working fine. 
command
passed
But, need to check for python.

ivarusic-amd: With this change python works fine also.

pfultz2 (approved):

PR #2951, opened 2024-04-04
Update Stable Diffusion 2.1 example

attila-dusnoki-htec:
Update the Stable Diffusion 2.1 example to use torch gpu buffers.
Also there are some optimization with batching certain tensors.

The notebook is not updated, after it is done, i will mark it ready.

The denoise_step time: ~85.7930 ms -> 77.4980 ms
The others are not really comparable, because there was no warmup, and those are run once.

### DIFF
--- a/examples/diffusion/python_stable_diffusion_21/gradio_reqirements.txt
+++ b/examples/diffusion/python_stable_diffusion_21/gradio_reqirements.txt
@@ -1,7 +1,7 @@
 #####################################################################################
 # The MIT License (MIT)
 #
-# Copyright (c) 2015-2023 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


attila-dusnoki-htec:
TODOs:
- [ ] Fix licensing
- [ ] Update jupyter notebook
- [ ] Extend it with fp16 (per-model to avoid unet nans e.g "--fp16" all or "--fp16 vae clip")
- [ ] Check if hip streams would make much difference instead of sync after copy and run

examples/diffusion/python_stable_diffusion_21/gradio_reqirements.txt
```diff
@@ -1,7 +1,7 @@
 #####################################################################################
```
causten, line 1: The file is spelt incorrectly.  You are missing the "u" in gradio_requirements.txt

examples/diffusion/python_stable_diffusion_21/gradio_reqirements.txt
```diff
@@ -1,7 +1,7 @@
...
 # The MIT License (MIT)
 #
-# Copyright (c) 2015-2023 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
```
causten, line 4: Change from -f requirements.txt to -r requirements.txt

causten (changes requested):

causten (approved):

PR #2957, opened 2024-04-08
Add support for multi outputs in pointwise ops

pfultz2:


src/module.cpp
```diff
@@ -663,6 +663,71 @@ std::vector<shape> module::get_output_shapes() const
     }
 }
 
+std::vector<shape> module::compute_shapes(const std::vector<shape>& inputs,
+                                          compute_shapes_options options) const
+{
+    auto params = this->get_parameter_names();
+    std::sort(params.begin(), params.end());
```
TedThemistokleous, line 670: why is a sort needed here?

src/targets/gpu/compile_gen.cpp
```diff
@@ -180,12 +180,16 @@ std::string make_transformer_args(std::vector<std::string> transformers)
     return join_strings(std::move(transformers), ", ");
 }
 
-void generate_pointwise(cpp_generator& gg, const module& pm, const std::string& name)
+static void generate_pointwise(cpp_generator& gg,
+                               const module& pm,
+                               const std::string& name,
+                               bool always_return_tuple = false)
```
TedThemistokleous, line 186: You don need this if its already defaulted in the header

test/verify/test_pointwise_multi_out.cpp
```diff
@@ -0,0 +1,56 @@
...
+        auto* pm = p.create_module("pointwise");
+        {
+            auto x1   = pm->add_parameter("x1", {migraphx::shape::float_type});
+            auto x2   = pm->add_parameter("x2", {migraphx::shape::float_type});
+            auto add  = pm->add_instruction(migraphx::make_op("add"), x1, x2);
+            auto abs  = pm->add_instruction(migraphx::make_op("abs"), add);
+            auto sqrt = pm->add_instruction(migraphx::make_op("sqrt"), abs);
+            pm->add_return({add, sqrt});
+        }
+        pm->set_bypass();
```
TedThemistokleous, line 48: Why do we set the bypass here?

test/ref/pointwise.cpp
```diff
@@ -49,3 +52,28 @@ TEST_CASE(pointwise_test)
...
+    }
+    mm->add_instruction(migraphx::make_op("pointwise"), {a1, a2}, {pm});
+    p.compile(migraphx::make_target("ref"));
+    auto results = p.eval({}).back().get_sub_objects();
+
+    std::vector<float> gold1 = {0, 16, 4};
+    std::vector<float> gold2 = {0, 4, 2};
+    EXPECT(results[0].to_vector<float>() == gold1);
+    EXPECT(results[1].to_vector<float>() == gold2);
+}
```
TedThemistokleous, line 79: So just as a sanity check to ensure I've understood this. 

You're creating a pointwise module here, and then using it across two different inputs to get two completely different outputs and then verifying those outputs on ref.

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
```diff
@@ -0,0 +1,164 @@
...
+#include <migraphx/kernels/array.hpp>
+
+namespace migraphx {
+
+namespace tuple_detail {
+
+template <class T, index_int N>
+struct element_storage
+{
+    /* [[no_unique_address]] */ T element;
```
TedThemistokleous, line 37: What do you mean by this comment?

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
```diff
@@ -0,0 +1,164 @@
...
+#ifndef MIGRAPHX_GUARD_KERNELS_TUPLE_HPP
+#define MIGRAPHX_GUARD_KERNELS_TUPLE_HPP
+
+#include <migraphx/kernels/array.hpp>
+
+namespace migraphx {
+
+namespace tuple_detail {
+
+template <class T, index_int N>
```
TedThemistokleous, line 34: Why isn't this index_int N class T like the other calls here?

TedThemistokleous (commented):
Few questions. Kicked off CI for you to see if the other stuff is breaking. 

Fix that but I'll take another pass once you've added more changes/answered questions. I think I get the general gist of this one.

src/module.cpp
pfultz2: The order of parameters when passing as arguments is to use alphabetic order.

src/targets/gpu/compile_gen.cpp
pfultz2: This isn't defined in a header.

test/verify/test_pointwise_multi_out.cpp
pfultz2: So it wont apply passes to it until gets to compiling it to C++. Otherwise we would apply lowering to this and its not meant to be lowered in the main pipeline.

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
pfultz2: This attribute is part of C++20, I dont know if we need to enable C++20 or if there is another attribute. Let me look into that.

This attribute is to make empty elements in the tuple to not increase the size of the tuple. So `tuple<empty_type, int> == sizeof(int)`. To do this the compiler needs to assign an address to `element` that is shared with other elements(it doesnt matter that the address is shared since an empty has no data to be read).

C++ requires each member variable to have a unique address, so `[[no_unique_address]]` tells the compiler to let it have a non-unique address that is an address that can be shared with other member variables. There are ways to achieve this without using this attribute through inheritance but this introduces other complexities to deal with(such as ensuring that `tuple<tuple<empty>, tuple<empty>>` is not ambiguous).

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
pfultz2: Thats because for the calls to `get_element` the `N` is passed as an explicit parameter but `T` is not. So `N` needs to appear first in the function declaration.

test/ref/pointwise.cpp
pfultz2: Yea this takes two inputs and has two outputs and we check the ref for both outputs.

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
pfultz2: It works fine without problems so I uncommented it.

src/module.cpp
TedThemistokleous: ah understood. Do we need the sort at all though? Why does alphabetical order matter for the arguments? Does the order matter here?

pfultz2: > Do we need the sort at all though?

Yes of course because we pass them in alphabetic order to the submodule so we need to sort the names so they will be in alphabetic order. 

> Why does alphabetical order matter for the arguments?

That is the convention we have chosen for passing arguments to the submodule parameters.

> Does the order matter here?

Yes of course otherwise we wont know which argument goes with with parameter.

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
TedThemistokleous: Understood. I think it would get ignored though if its not supported whether w have C++20 on or not though wouldn't it? Doesn't hurt to have it in then if it simplifies things.

test/ref/pointwise.cpp
TedThemistokleous: Understood. All good

test/verify/test_pointwise_multi_out.cpp
TedThemistokleous: I see so going forward for all multi output pointwise ops we create we'd need to ensure bypass is set  as part of the program we create and then compile everything together

src/targets/gpu/compile_gen.cpp
TedThemistokleous: [src/targets/gpu/include/migraphx/gpu/compile_gen.hpp](https://github.com/ROCm/AMDMIGraphX/pull/2957/files#diff-f2872ed1a864f2a138f87134aa2b21ed3f82f15db570dc7ac0807cb7e2528158)

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
pfultz2: It says in clang docs that its supported in C++11 or later:

https://releases.llvm.org/15.0.0/tools/clang/docs/AttributeReference.html#no-unique-address

So it should be fine.

src/targets/gpu/compile_gen.cpp
TedThemistokleous: Actually my mistake on this one. But you're still flowing through false by default from the previous generate_pointwise from the header defaults, so is this false needed?

test/verify/test_pointwise_multi_out.cpp
pfultz2: The bypass was always needed even before multi outputs. In the past, we never created the pointwise op directly and instead realied on fuse_pointwise. However, for multi-outputs I created it directly since fuse_pointwise wont create multi-outputs(we can add that later).

src/targets/gpu/compile_gen.cpp
pfultz2: No, its called in two places. It called in `generate_pointwise` but its also called in `generate_reduce` as well, and it doesnt pass the flag there.

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
TedThemistokleous: Yeah I'd say leave it in then

src/targets/gpu/compile_gen.cpp
TedThemistokleous: Fair. I see it. Now I see why you added the default.

TedThemistokleous (approved):
Fix CI otherwise no more questions come to mind. Looks good.

src/include/migraphx/module.hpp
```diff
@@ -202,6 +202,17 @@ struct MIGRAPHX_EXPORT module
...
+    struct compute_shapes_options
+    {
+        std::string name                               = "compute_shapes";
+        bool strict_type                               = false;
+        bool strict_lens                               = false;
+        std::vector<std::size_t> scalar_const_out_lens = {};
+    };
+    std::vector<shape> compute_shapes(const std::vector<shape>& inputs,
+                                      compute_shapes_options options) const;
+    std::vector<shape> compute_shapes(const std::vector<shape>& inputs) const;
```
umangyadav, line 214: Can you add docstring on what these `compute_shapes` for a module ?

src/include/migraphx/raw_data.hpp
```diff
@@ -191,6 +191,14 @@ struct raw_data : raw_data_base
         ss << static_cast<const Derived&>(*this);
         return ss.str();
     }
+
+    template <class T>
+    std::vector<T> to_vector() const
+    {
+        std::vector<T> result(static_cast<const Derived&>(*this).get_shape().elements());
+        this->visit([&](auto x) { result.assign(x.begin(), x.end()); });
```
umangyadav, line 199: This should assert that it is not of type "tuple".

umangyadav: We can make use of this in many of our tests which currently uses visit.  It could be a good first time issue. Can you open an issue and tag as "good first issue" ?

test/ref/pointwise.cpp
```diff
@@ -49,3 +52,28 @@ TEST_CASE(pointwise_test)
     std::vector<float> gold = {0, 2, 4};
     EXPECT(migraphx::verify::verify_rms_range(results_vector, gold));
 }
+
+TEST_CASE(pointwise_multi_out_test)
```
umangyadav, line 56: Would current logic handle case of nested tuples ? 
If so can you add a test for the same ?

src/include/migraphx/op/pointwise.hpp
```diff
@@ -45,23 +45,18 @@ struct pointwise
...
-
-        // Scalar output if all inputs are scalar
-        if(inputs.front().elements() == 1 and
-           all_of(inputs, [](const auto& s) { return s.scalar(); }))
-            return shape{type};
-
-        return shape::from_permutation(type, inputs.front().lens(), find_permutation(inputs));
+        auto result = pm->compute_shapes(
+            inputs,
+            {.name = name(), .strict_type = true, .scalar_const_out_lens = inputs.front().lens()});
```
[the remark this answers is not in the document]
umangyadav: Can you add some tests inside `op_shape_test.cpp` for different cases for the pointwise op ? 

e.g. 
with mutiple outputs, 
with one of the outputs being scalar, 
with strict_type being true and throwing

src/include/migraphx/op/pointwise.hpp
```diff
@@ -86,8 +81,15 @@ struct pointwise
                 [&](auto&& name, auto&& arg) { return std::make_pair(name, arg.element(i)); });
 
             auto results = run(pm, params);
-            assert(results.size() == 1);
-            visit_all(output, results.front())([&](auto out, auto x) { out[i] = x.front(); });
+            assert(results.size() == output.get_sub_objects().size() or
+                   (results.size() == 1 and output.get_sub_objects().empty()));
```
umangyadav, line 85: Can there be a tuple with size 1 ?

TedThemistokleous: I think this handles the single input single output case still is how I interpreted it. Either you get type type and ensure your sub objects match or you have default behavior after a run

src/fuse_pointwise.cpp
```diff
@@ -183,10 +182,11 @@ static bool find_pointwise_modules(module& m)
             continue;
         auto input = *it;
 
-        auto new_inputs = append_pointwise_module(input, ins);
-        m.replace_instruction(input, input->get_operator(), new_inputs, input->module_inputs());
-        m.replace_instruction(ins, input);
-        m.move_instruction(input, ins);
+        auto fused = append_pointwise_module(input, ins);
+        auto name  = fused.mod.name();
+        mpm.rename_module(name, name + ":" + ins->module_inputs().front()->name() + "-deleted");
```
umangyadav, line 187: why `name+:` is prepended on deleted module ?

umangyadav: why does it require renaming ? It's appending current pointwise module to previous pointwise module. Fused module would take name of previous pointwise module. So why is there a need to rename module?

src/cpp_generator.cpp
```diff
@@ -53,12 +54,13 @@ cpp_generator::function::set_body(const module& m, const cpp_generator::generate
         }
         else if(ins->name() == "@return")
         {
-            assert(ins->inputs().size() == 1);
-            return_ins = ins->inputs().front();
+            names[ins] = prefix + "return";
+            ss << "auto " << prefix << "return = " << g(ins, names) << ";\n";
```
umangyadav, line 58:
```suggestion
            ss << "auto " << names[ins] << " = " << g(ins, names) << ";\n";
```

src/include/migraphx/raw_data.hpp
pfultz2: The `visit` already throws an error on tuple type.

test/ref/pointwise.cpp
pfultz2: Right now there isnt a way to create a tuple in the pointwise module.

src/include/migraphx/op/pointwise.hpp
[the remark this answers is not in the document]
pfultz2: All the inputs to pointwise should have the dimensions. We need to create a dimension for the scalar literals in the pointwise module(which only have a single element) so we can compute the shape correctly.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -150,59 +150,11 @@ struct mlir_op
...
-                               [&](auto in) { return mod_ins_shapes[in]; });
-                mod_ins_shapes[ins] = ins->get_operator().compute_shape(input_shapes);
-            }
-        }
-        MIGRAPHX_THROW("No return found in the submodule");
+        auto result =
+            mod->compute_shapes(inputs, {.name = name(), .strict_type = true, .strict_lens = true});
+        if(result.size() == 1)
+            return result.front();
+        return shape{result};
```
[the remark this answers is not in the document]
pfultz2: MLIR might support multiple outputs, but we dont have the support to handle multiple outputs when translating to MLIR dialect. Also, its beyond the scope of this PR.

src/include/migraphx/op/pointwise.hpp
```diff
@@ -45,23 +45,18 @@ struct pointwise
...
         check_shapes{inputs, *this}.has(pnames.size()).same_dims();
 
-        auto type = pm->get_output_shapes().front().type();
-
-        // Scalar output if all inputs are scalar
-        if(inputs.front().elements() == 1 and
-           all_of(inputs, [](const auto& s) { return s.scalar(); }))
-            return shape{type};
-
-        return shape::from_permutation(type, inputs.front().lens(), find_permutation(inputs));
```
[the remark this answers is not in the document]
pfultz2: Each pointwise operator should propagate the transpose. This will ensure the fused pointwise will have the same layout as the non-fused pointwise.

src/include/migraphx/op/pointwise.hpp
pfultz2: > Can there be a tuple with size 1 ?

Yes but for single output we just return it directly instead of a tuple. This is so the single output will still work the same, some more work would be needed to handle tuple outputs in the fuse_pointwise(which outside of the scope of this PR).

src/fuse_pointwise.cpp
pfultz2: Because the fused module uses the same name as the first pointwise, so it needs to rename it so we can add a module with this name, we can remove the module yet until DCE removes the unused function, so thats why we rename it to something that shouldn't conflict with other module names.

src/fuse_pointwise.cpp
```diff
@@ -157,20 +156,20 @@ static std::vector<instruction_ref> append_pointwise_module(instruction_ref ins,
...
         }
     }
-    pm->replace_return(pm->insert_instructions(last, xm, &map_ins));
-    return inputs;
+    pm.replace_return(pm.insert_instructions(last, xm, &map_ins));
+    return {std::move(pm), inputs};
 }
 
-static bool find_pointwise_modules(module& m)
+static bool find_pointwise_modules(module_pass_manager& mpm)
```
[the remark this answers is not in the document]
pfultz2: No this does not handle that yet. Its beyond the scope of this PR, but is something we should work on in the future.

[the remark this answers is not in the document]
TedThemistokleous: Lets track this in an issue then so we don't forget

src/include/migraphx/module.hpp
pfultz2: Added.

src/include/migraphx/op/pointwise.hpp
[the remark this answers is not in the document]
pfultz2: > Can you add some tests inside op_shape_test.cpp for different cases for the pointwise op ?

You can't test operators like pointwise op in op_shape_test.cpp.

>  with mutiple outputs,

There is already a ref test for this.

>  with one of the outputs being scalar,

I am not sure what you are intending for such a test. This wont get coverage for `scalar_const_out_lens` flag. Either way both test cases(ie scalar output and `scalar_const_out_lens` flag) are already in test/fuse_pointwise.cpp.

test/ref/pointwise.cpp
umangyadav: > Right now there isnt a way to create a tuple in the pointwise module. 

I assume you mean there is no way to create "nested" tuples as the outputs of the pointwise module.  Tuples can be created that is what this PR is doing.

src/fuse_pointwise.cpp
umangyadav: i don't understand why it is necessary to rename it. `ins` is being replaced on line 189 and that would remove all its' usages and then DCE runs just after `append_pointwise`

src/include/migraphx/op/pointwise.hpp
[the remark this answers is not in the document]
umangyadav: > . This will ensure the fused pointwise will have the same layout as the non-fused pointwise. 

This may result in different output layout compared to previous logic.  I am wondering about performance impact. 

e.g. pointwise module doing add between two transposed inputs. 
previous logic would return output shape that is transposed. 
Current logic from what i understand would instead return standard shape.

src/targets/gpu/jit/pointwise.cpp
```diff
@@ -70,39 +70,42 @@ struct pointwise_compiler : compiler<pointwise_compiler>
...
         options.set_launch_params(
-            v, compute_global_for(ctx, options.output.elements() / vec.size, 256));
-        auto src = interpolate_string(pointwise_kernel,
-                                      {{"kernel", options.kernel_name},
-                                       {"params", enum_params(inputs.size(), "void * private_p")},
-                                       {"args", enum_params(inputs.size(), "private_p")},
-                                       {"lambda", v.at("lambda").to<std::string>()},
-                                       {"transformers", make_transformer_args(vec)},
-                                       {"preamble", v.get("preamble", std::string{})}});
+            v, compute_global_for(ctx, options.inputs.front().elements() / vec.size, 256));
```
umangyadav, line 81: why did you have to change `options.output` to `options.inputs.front()` ?

src/targets/gpu/jit/pointwise.cpp
```diff
@@ -70,39 +70,42 @@ struct pointwise_compiler : compiler<pointwise_compiler>
...
     {
         if(contains({"layout", "contiguous"}, op.name()))
         {
-            return compile_op(
-                ctx,
-                to_shapes(ins->inputs()),
-                {{"lambda", "[](auto x) { return x; }"}, {"kernel", op.name() + "_kernel"}});
+            return compile_op(ctx,
+                              to_shapes(ins->inputs()),
+                              {{"lambda", "[](auto x) { return make_tuple(x); }"},
```
umangyadav, line 101: Layout and contiguous do not need to return tuple.  `make_tuple()` is that necessary ?

src/include/migraphx/op/pointwise.hpp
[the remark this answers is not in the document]
pfultz2: > You can, we already have some.

Ah I missed that. I will add a test for the `.strict_type` flag. 

> I was looking for test like where it would use .scalar_const_out_lens flag.

Yea we already have that test in test/fuse_pointwise.cpp:

https://github.com/ROCm/AMDMIGraphX/blob/bd52f7172abb76722283fcb9660990cf94db7190/test/fuse_pointwise.cpp#L221

If we dont use the `.scalar_const_out_lens` flag then this will throw an error because the inputs to the `add` operator will be `{2, 3}` and a scalar. 

> > with one of the outputs being scalar,

We already have a test for this as well in test/fuse_pointwise.cpp:

https://github.com/ROCm/AMDMIGraphX/blob/bd52f7172abb76722283fcb9660990cf94db7190/test/fuse_pointwise.cpp#L305

It takes all scalar inputs and checks that we have a scalar output.

src/include/migraphx/op/pointwise.hpp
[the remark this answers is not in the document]
pfultz2: >  Current logic from what i understand would instead return standard shape.

No it wont return standard shape. The `binary` operator will return the permutation of the non-packed or non-broadcasted tensor:

https://github.com/ROCm/AMDMIGraphX/blob/bd52f7172abb76722283fcb9660990cf94db7190/src/include/migraphx/op/binary.hpp#L81

src/fuse_pointwise.cpp
pfultz2: If I replace `input`, then it will use the inputs to `ins` which some could come after the `input` instruction, which will make an invalid module.

src/targets/gpu/jit/pointwise.cpp
pfultz2: Yes the output can be a tuple, so I just use one of the inputs instead. They should all have the same number of elements.

src/targets/gpu/jit/pointwise.cpp
pfultz2: Yes every pointwise needs to return a tuple in the kernel otherwise it wont compile. Thats because the pointwise kernel will unpack the tuple to write to the different output buffers.

src/fuse_pointwise.cpp
umangyadav: > which some could come after the input instruction, which will make an invalid module. 

Looks like current logic on develop branch would fail in that case. I wonder how come current logic on develop didn't stumble into that case.

test/ref/pointwise.cpp
pfultz2: > Tuples can be created that is what this PR is doing.

Sure in general yes, but they can't be created in a pointwise module.

src/targets/gpu/fuse_mlir.cpp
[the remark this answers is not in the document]
pfultz2: There is already an `assert` for this:

https://github.com/ROCm/AMDMIGraphX/blob/bd52f7172abb76722283fcb9660990cf94db7190/src/targets/gpu/mlir.cpp#L643

src/fuse_pointwise.cpp
pfultz2: On develop, it replaced the input instruction, then replaced the usage of `ins`, and then moved it before the `ins`. However, we would still need to rename the module if we did this.

The reason why we dont need to rename it on develop was because we werent checking that the input types matched. So if we replace `input` it can create a different output type(like when we fuse a `convert`). This will still be used by `ins`, so a recompute shape is called, but there is no mismatch error(even though the types no longer match). 

Instead of going through this dance of replace input, then `ins` and move it. I just replace `ins` instead.

src/include/migraphx/op/pointwise.hpp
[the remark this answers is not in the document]
pfultz2: Test added.

umangyadav (approved):

PR #2959, opened 2024-04-09
Stable Diffusion XL Turbo

attila-dusnoki-htec:
With this, we enable 3 versions of SDXL:
- `SDXL` - the base version
- `SDXL-opt` - the tensorrt optimized version
- `SDXL-turbo` - the turbo version

Note: Only the opt version has an onnx refiner model, but that works(?) with all 3 versions. (But only looks good with the opt version 🤷)

|Version|Default|With Refiner|
|:-:|:-:|:-:|
|Base|gradio_sdxl|gradio_sdxl_refiner|
|Opt|gradio_sdxl_opt|gradio_sdxl_opt_refiner|
|Turbo|gradio_sdxl_turbo|gradio_sdxl_turbo_refiner|

Marked as `draft`, because i still want to double check the refiner result, maybe i missed something (e.g. different dtypes)


attila-dusnoki-htec:
The problem was with the refiner, that it used the wrong encoder, because it needs to be the clip2.opt.mod for all example.
Thats why its only worked with the opt version.

The fixed SDXL with refiner:
gradi_sdxl_refiner2

The fixed SDXL-Turbo with refiner:
gradio_sdxl_turbo_refiner2

Also, the gradio app was extended with an output log.

Before run:
gradio_output_log_before

After run:
gradio_output_log_after

Note: this will only show it after the run, not real time.

causten (approved):

PR #2963, opened 2024-04-10
Fix bug with computing output shape with fused_reduce

pfultz2:


causten (approved):
I tested it and this works.  No comment on the code

src/fuse_reduce.cpp
```diff
@@ -65,18 +65,10 @@ struct fused_reduce
...
            }))
             MIGRAPHX_THROW("Dimenstion does not match the submodule.");
-        const auto& s = inputs.at(0);
-        auto lens     = s.lens();
-        if(lens != sm->get_output_shapes().front().lens())
-        {
-            for(const auto& axis : axes)
-            {
-                lens[axis] = 1;
-            }
```
TedThemistokleous, line 75: Are we not concerned about this case anymore then?

TedThemistokleous (changes requested):
Fix CI and just a question about the change in reduce.

src/fuse_reduce.cpp
pfultz2: This was pointless because the output shape already computes the dimensions.

TedThemistokleous (approved):

TedThemistokleous:
Fix CI

PR #2964, opened 2024-04-11
Script optimization: put output memcpy out of session run

jinfeng-t:
1. with io_binding.bind_output('output')
ort1

2. with io_binding.bind_output('output', 'cuda')
ort2

it slightly improves the measured performance results.

### DIFF
--- a/examples/onnxruntime/resnet50/resnet50.py
+++ b/examples/onnxruntime/resnet50/resnet50.py
@@ -116,7 +116,7 @@ def run_sample(session,
                verbose=False):
     io_binding = session.io_binding()
     io_binding.bind_cpu_input('input', inputs.cpu().detach().numpy())
-    io_binding.bind_output('output')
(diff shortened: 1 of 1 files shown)


CharlieL7 (approved):
Weird to me that the GPU option is named `cuda`, but I guess that's how it is.
I did find the snippet about device_type in the API but no other option looks any better.
```
def get_ort_device_type(device_type: str, device_index) -> C.OrtDevice:
    if device_type == "cuda":
        return C.OrtDevice.cuda()
    elif device_type == "cann":
        return C.OrtDevice.cann()
    elif device_type == "cpu":
        return C.OrtDevice.cpu()
    elif device_type == "dml":
        return C.OrtDevice.dml()
    elif device_type == "ort":
        return C.get_ort_device(device_index).device_type()
    else:
        raise Exception("Unsupported device type: " + device_type)
```

PR #2966, opened 2024-04-11
Remove rocblas on windows

tvukovic-amd:
Added MIGRAPHX_USE_ROCBLAS build option which is automatically ON on Windows and OFF on Linux.
When MIGRAPHX_USE_ROCBLAS is OFF, rocBLAS is removed on MIGraphX.

@apwojcik


CMakeLists.txt
```diff
@@ -340,5 +346,10 @@ rocm_create_package(
     MAINTAINER "AMDMIGraphX Maintainer <migraphx-lib.support@amd.com>"
     LDCONFIG
     PTH
-    DEPENDS miopen-hip rocblas ${DEPENDS_HIP_RUNTIME} hip-base half ${PACKAGE_DEPENDS}
+    DEPENDS miopen-hip ${DEPENDS_HIP_RUNTIME} hip-base half ${PACKAGE_DEPENDS}
 )
+
+if(MIGRAPHX_USE_ROCBLAS)
+    list(APPEND PACKAGE_DEPENDS PACKAGE rocblas)
+endif()
```
pfultz2, line 354: This needs to be done before `rocm_create_package`.

src/targets/gpu/compile_miopen.cpp
```diff
@@ -29,7 +29,9 @@
 #include <migraphx/make_op.hpp>
 #include <migraphx/register_op.hpp>
 #include <migraphx/op/identity.hpp>
-#include <migraphx/gpu/rocblas.hpp>
+#ifdef MIGRAPHX_USE_ROCBLAS
```
pfultz2, line 32: This should be defined as 1 or 0, so it should use `#if MIGRAPHX_USE_ROCBLAS`. Also I dont see you adding this define to cmake.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -79,11 +81,23 @@ static bool is_negated_op(const std::string& s)
     return contains({'!', '~'}, s[0]);
 }
 
+#ifndef MIGRAPHX_USE_ROCBLAS
+const std::vector<std::string>& get_default_options()
+{
+    static std::vector<std::string> default_options{"dot", "convolution", "fused"};
+    return default_options;
+}
+#endif
```
pfultz2, line 90: We should not be changing how the environment variable works. This also disables miopen convolutions as well, which seems unrelated.

src/targets/gpu/fuse_ops.cpp
```diff
@@ -893,11 +898,13 @@ void fuse_ops::apply(module& m) const
     match::find_matches(m, find_conv_pointwise{ctx}, find_conv_bias_relu{ctx}, find_conv_bias{ctx});
     run_passes(m, {dead_code_elimination{}});
     match::find_matches(m,
-                        find_gemm_pointwise{},
                         find_layernorm_pointwise{},
                         find_concat_pointwise{},
                         find_contiguous_tranpose_gemm{},
                         find_commutative_broadcast{});
+    #ifdef MIGRAPHX_USE_ROCBLAS
+        match::find_matches(m, find_gemm_pointwise{});
```
pfultz2, line 906: This should not be moved here, it should stay in the list where it was at if rocblas is enabled.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -27,7 +27,9 @@
 #include <migraphx/gpu/export.h>
 #include <migraphx/context.hpp>
 #include <migraphx/gpu/miopen.hpp>
+#if MIGRAPHX_USE_ROCBLAS
 #include <migraphx/gpu/rocblas.hpp>
+#endif
```
pfultz2, line 32: You already ifdef this file so you shouldn't need an ifdef when including it.

src/targets/gpu/lowering.cpp
```diff
@@ -47,7 +47,9 @@
 #include <migraphx/gpu/device_name.hpp>
 #include <migraphx/gpu/gemm.hpp>
 #include <migraphx/gpu/miopen.hpp>
+#if MIGRAPHX_USE_ROCBLAS
 #include <migraphx/gpu/rocblas.hpp>
+#endif
```
pfultz2, line 52: Same here, this doesnt need a ifdef.

src/targets/gpu/compile_miopen.cpp
```diff
@@ -29,7 +29,9 @@
 #include <migraphx/make_op.hpp>
 #include <migraphx/register_op.hpp>
 #include <migraphx/op/identity.hpp>
+#if MIGRAPHX_USE_ROCBLAS
 #include <migraphx/gpu/rocblas.hpp>
+#endif
```
pfultz2, line 34: Same here, the ifdef can be removed.

src/targets/gpu/CMakeLists.txt
```diff
@@ -112,6 +116,10 @@ if(NOT MIGRAPHX_USE_COMPOSABLEKERNEL)
             ${CMAKE_CURRENT_SOURCE_DIR}/jit/ck_gemm_softmax_gemm.cpp)
 endif()
 
+if (MIGRAPHX_USE_ROCBLAS)
+    set(ROCBLAS_SRCS rocblas.cpp gemm_impl.cpp)
```
pfultz2, line 120: You already ifdef these sources, so we dont need to remove them from the list in cmake.

tperry-amd:
Please change the code to default the CMake variable to ON for now.

CMakeLists.txt
```diff
@@ -61,6 +61,12 @@ else()
 option(MIGRAPHX_ENABLE_PYTHON "Enable python bindings" ON)
 endif()
 
+if(WIN32)
```
tperry-amd: Let's change this code to default the variable to ON only right now.

tperry-amd (changes requested):

tperry-amd (approved):

PR #2973, opened 2024-04-15
Add support for MaxPool unit8 type

nives-vukovic:
Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/165


umangyadav (commented):
Is there a test that you want to enable or add ?

nives-vukovic:
@umangyadav That is why I left it in draft state to check with you about this. Currently it fixes crashes for a number of models listed in https://github.com/gyulaz-htec/models/blob/migraphx_testing/MIGRAPHX_fp32-int8.md and also one python backend test that I will remove from disabled list. So this causes issues only on the gpu, is there any other test that you would prefer to have for this kind of fix?

TedThemistokleous:
> @umangyadav That is why I left it in draft state to check with you about this. Currently it fixes crashes for a number of models listed in https://github.com/gyulaz-htec/models/blob/migraphx_testing/MIGRAPHX_fp32-int8.md and also one python backend test that I will remove from disabled list. So this causes issues only on the gpu, is there any other test that you would prefer to have for this kind of fix?

Well if there are any ONNX tests on backend that this breaks on or allows us to run, we should turn those on for coverage and use that to run in our CI if that fixes it.

Secondly what models from that list does this fix? Do you have a comprehensive list? Add that to this ticket and we can determine if we need to add coverage for them here or make them part of another set of runs.

nives-vukovic:
@TedThemistokleous I added a full list of models that don't break due to the unsupported error anymore, they compile and run successfully, but they do have accuracy issues as stated in the comment https://github.com/migraphx-benchmark/AMDMIGraphX/issues/165.
I have also enabled one backend test that passes now.

TedThemistokleous:
> @TedThemistokleous I added a full list of models that don't break due to the unsupported error anymore, they compile and run successfully, but they do have accuracy issues as stated in the comment [migraphx-benchmark#165](https://github.com/migraphx-benchmark/AMDMIGraphX/issues/165). I have also enabled one backend test that passes now.

Accuracy could be related to uint8-int8 related conversion since you need to shift the data and zero points accordingly. See #2903

umangyadav:
> @umangyadav That is why I left it in draft state to check with you about this. Currently it fixes crashes for a number of models listed in https://github.com/gyulaz-htec/models/blob/migraphx_testing/MIGRAPHX_fp32-int8.md and also one python backend test that I will remove from disabled list. So this causes issues only on the gpu, is there any other test that you would prefer to have for this kind of fix?

Having a onnx backend test is good. But in addition to that can you make a few of the verify tests with unit8 type. 
For example : 
You can templatize following tests  to test them with uint8. 
https://github.com/ROCm/AMDMIGraphX/blob/develop/test/verify/test_avg_pooling_3d.cpp
https://github.com/ROCm/AMDMIGraphX/blob/develop/test/verify/test_global_max_pooling.cpp

umangyadav:
> @umangyadav I added two verify tests for MaxPool as requested. Uint8 type or any integer type is not supported for AveragePool, as it can be seen https://onnx.ai/onnx/operators/onnx__AveragePool.html.

ONNX doesn't seem to support that but inside MIGraphX we should be able to create avgpool with uint8 dtype.

nives-vukovic:
@umangyadav  Currently as I see test_avg_pooling_3d_opt passes for uint8_type, while test_avg_pooling_3d fails with 
RMS Error: 0.00927837
ref:14, 16, 12, 18, 22, 13, 12, 17, 14, 18, 15, 14, 12, 18, 13, 15, 10, 16, 18, 19, 15, 18, 13, 14
target:14, 16, 12, 18, 22, 13, 12, 17, 14, 17, 15, 14, 12, 18, 13, 15, 10, 16, 18, 19, 15, 18, 13, 14
Max diff: 1
Mismatch at 9:  !=

umangyadav:
> @umangyadav @TedThemistokleous Currently as I see test_avg_pooling_3d_opt passes for uint8_type, while test_avg_pooling_3d fails with RMS Error: 0.00927837 ref:14, 16, 12, 18, 22, 13, 12, 17, 14, 18, 15, 14, 12, 18, 13, 15, 10, 16, 18, 19, 15, 18, 13, 14 target:14, 16, 12, 18, 22, 13, 12, 17, 14, 17, 15, 14, 12, 18, 13, 15, 10, 16, 18, 19, 15, 18, 13, 14 Max diff: 1 Mismatch at 9: !=
> 
> The test that fails has padding that is not default.

@nives-vukovic  can you open an issue about the avgpool failure with padding ? I'll approve this PR for now.

TedThemistokleous (approved):

causten:
@nives-vukovic there seems to be  failure in the clang_asan run.  Can you investigate

nives-vukovic:
> @nives-vukovic there seems to be failure in the clang_asan run. Can you investigate

Unfortunately, I don't have access to the outputs of these errors. Can those be provided?

nives-vukovic:
@TedThemistokleous @umangyadav  It seems that clang_asan fails for test test_max_pooling_ceil_3d<migraphx::shape::uint8_type>, I managed to reproduce this issue on our system, as far as I can with this cmake configuration it runs all tests on cpu, instead of gpu and fails for this particular test case with error:
`/home/jenkins/workspace/AMDMIGraphX_PR-2973/src/include/migraphx/shape.hpp:338:25: runtime error: -3.40282e+38 is outside the range of representable values of type 'unsigned char'`

Did you encountered such issues before? I see that on cpu, there should be conversion of all types to float, so I am not sure why this is reported. I noticed that in previous PRs where you added support of uint8 type you didn't add verify tests, but only different onnx tests, so is this expected?

TedThemistokleous:
> @TedThemistokleous @umangyadav It seems that clang_asan fails for test `test_max_pooling_ceil_3d<migraphx::shape::uint8_type>`, I managed to reproduce this issue on our system, as far as I can with this cmake configuration it runs all tests on cpu, instead of gpu and fails for this particular test case with error: `/home/jenkins/workspace/AMDMIGraphX_PR-2973/src/include/migraphx/shape.hpp:338:25: runtime error: -3.40282e+38 is outside the range of representable values of type 'unsigned char'`

I haven't seen that before. Likely this implies we're missing some sort of conversion?  That looks like we're getting something smaller than numeric limits off lowest() for float. https://en.cppreference.com/w/cpp/types/numeric_limits



> Did you encountered such issues before? I see that on cpu, there should be conversion of all types to float, so I am not sure why this is reported. I noticed that in previous PRs where you added support of uint8 type you didn't add verify tests, but only different onnx tests, so is this expected?

This isn't expected. We shouldn't be hitting this case. I'll have to dig more

nives-vukovic:
> > @TedThemistokleous @umangyadav It seems that clang_asan fails for test `test_max_pooling_ceil_3d<migraphx::shape::uint8_type>`, I managed to reproduce this issue on our system, as far as I can with this cmake configuration it runs all tests on cpu, instead of gpu and fails for this particular test case with error: `/home/jenkins/workspace/AMDMIGraphX_PR-2973/src/include/migraphx/shape.hpp:338:25: runtime error: -3.40282e+38 is outside the range of representable values of type 'unsigned char'`
> 
> I haven't seen that before. Likely this implies we're missing some sort of conversion? That looks like we're getting something smaller than numeric limits off lowest() for float. https://en.cppreference.com/w/cpp/types/numeric_limits
> 
> > Did you encountered such issues before? I see that on cpu, there should be conversion of all types to float, so I am not sure why this is reported. I noticed that in previous PRs where you added support of uint8 type you didn't add verify tests, but only different onnx tests, so is this expected?
> 
> This isn't expected. We shouldn't be hitting this case. I'll have to dig more

Did you happen to look into this issue or have an advice where to dig for the problem?

umangyadav:
> > > @TedThemistokleous @umangyadav It seems that clang_asan fails for test `test_max_pooling_ceil_3d<migraphx::shape::uint8_type>`, I managed to reproduce this issue on our system, as far as I can with this cmake configuration it runs all tests on cpu, instead of gpu and fails for this particular test case with error: `/home/jenkins/workspace/AMDMIGraphX_PR-2973/src/include/migraphx/shape.hpp:338:25: runtime error: -3.40282e+38 is outside the range of representable values of type 'unsigned char'`
> > 
> > 
> > I haven't seen that before. Likely this implies we're missing some sort of conversion? That looks like we're getting something smaller than numeric limits off lowest() for float. https://en.cppreference.com/w/cpp/types/numeric_limits
> > > Did you encountered such issues before? I see that on cpu, there should be conversion of all types to float, so I am not sure why this is reported. I noticed that in previous PRs where you added support of uint8 type you didn't add verify tests, but only different onnx tests, so is this expected?
> > 
> > 
> > This isn't expected. We shouldn't be hitting this case. I'll have to dig more
> 
> Did you happen to look into this issue or have an advice where to dig for the problem?

Clang ASAN is complaining because, test is using Ceil mode. Last window of maxpool is operation on padded region completely which are initialized with `numeric_limits<float>::Lowest()` . Entire maxpool window is over padded region. 
Later "convert" tries to convert `<float>::Lowest()` to uint8_type and then Clang ASAN complains.  I think this error can be ignored or suppressed but i am not sure how it can be done.

umangyadav:
Discussed potential solutions: 

- Check if MIGraphX is running in sanitizer mode or not by checking some macro. If it is then disable sanitizer on "convert"operation. 
- Do sanitizer safe convert operation for the downcast operations from float to integers when running in sanitizer mode. 
- Disable specific sanitizer check on convert operation. I am not sure if it is possible.

PR #2974, opened 2024-04-15
update SDXL txt2img.py script to support fp16 vae

kahmed10:
Used internally to point to a fp16 vae model.

### DIFF
--- a/examples/diffusion/python_stable_diffusion_xl/txt2img.py
+++ b/examples/diffusion/python_stable_diffusion_xl/txt2img.py
@@ -246,11 +246,17 @@ class StableDiffusionMGX():
         elif "all" in fp16:
             fp16 = ["vae", "clip", "clip2", "unetxl"]
 
+        self.vae_name = "vae_decoder"
+        self.vae_input_name = "latent_sample"
+        if "vae" in fp16:
+            self.vae_name = "vae_fp16_fix"
+            self.vae_input_name = "input.1"
+
         print("Load models...")
         self.models = {
             "vae":
             StableDiffusionMGX.load_mgx_model(
-                "vae_decoder", {"latent_sample": [1, 4, 128, 128]},
+                self.vae_name, {self.vae_input_name: [1, 4, 128, 128]},
                 onnx_model_path,
                 compiled_model_path=compiled_model_path,
                 use_fp16="vae" in fp16,
@@ -512,7 +518,7 @@ class StableDiffusionMGX():
 
     @measure
     def decode(self, latents):
-        copy_tensor_sync(self.tensors["vae"]["latent_sample"], latents)
+        copy_tensor_sync(self.tensors["vae"][self.vae_input_name], latents)
         run_model_sync(self.models["vae"], self.model_args["vae"])
         return self.tensors["vae"][get_output_name(0)]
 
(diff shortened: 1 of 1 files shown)


umangyadav (approved):

examples/diffusion/python_stable_diffusion_xl/txt2img.py
```diff
@@ -246,11 +246,17 @@ def __init__(self, onnx_model_path, compiled_model_path,
...
+        if "vae" in fp16:
+            self.vae_name = "vae_fp16_fix"
+            self.vae_input_name = "input.1"
+
         print("Load models...")
         self.models = {
             "vae":
             StableDiffusionMGX.load_mgx_model(
-                "vae_decoder", {"latent_sample": [1, 4, 128, 128]},
+                self.vae_name, {self.vae_input_name: [1, 4, 128, 128]},
```
umangyadav, line 259: Do you need to update steps in README.md about where to get this model ?

causten: @umangyadav can you point to where to get the model so the README can be adjusted

umangyadav: https://huggingface.co/madebyollin/sdxl-vae-fp16-fix

attila-dusnoki-htec: Can you try to export it like this and check if the result is correct?

```python
import torch
from diffusers import AutoencoderKL
import os

class VAEDecoder(torch.nn.Module):
    def __init__(self, vae):
        super().__init__()
        self.vae = vae

    def forward(self, latent_sample):
        return self.vae.decode(latent_sample)

vae = AutoencoderKL.from_pretrained("madebyollin/sdxl-vae-fp16-fix")
vae.eval()
output = "models/sdxl-1.0-base/vae_decoder_fp16_fix/model.onnx"
os.makedirs(os.path.dirname(output), exist_ok=True)
torch.onnx.export(VAEDecoder(vae),
                  torch.randn(1, 4, 128, 128),
                  output,
                  export_params=True,
                  do_constant_folding=True,
                  input_names=['latent_sample'])
```

If it works, it could be added next to `clip_modifier.py` like `export_vae_fp16_fix.py` or similar, and add a note in the README to use it before running `--fp16 vae`. Also, since the input-name can be set, only `vae_name` needed.
```python
vae_name = "vae_decoder_fp16_fix" if "vae" in fp16 else "vae_decoder"
```

attila-dusnoki-htec: You can also cherry-pick [this](https://github.com/ROCm/AMDMIGraphX/commit/a605f925851e76b63cd9b25210c42f11bd143bdd).

kahmed10: I made slightly different modifications, but used your above suggestion. Please take a look.

examples/diffusion/python_stable_diffusion_xl/README.md
```diff
@@ -99,9 +102,32 @@ Set `pipeline-type` based on which version of models you have: `sdxl` for base,
...
+Quantizing the StabilityAI VAE model produces NaNs in the output, so we need to download a modified version to fix this. Download and export the fp16 fixed version:
+
+```bash
+python vae_fp16.py
+```
+
+> [!NOTE]
+> By default, the path being saved to is `models/sdxl-1.0-base/vae_decoder_fp16_fix/model.onnx`. This can be changed by passing in `--output=<sdxl-directory>/vae_decoder_fp16_fix/model.onnx`.
+
+Run the same text-to-image script as before, except add `--fp6=all` to the command. For example:
```
attila-dusnoki-htec, line 119: `fp6` -> `fp16`

attila-dusnoki-htec (approved):
Only 1 typo that needs addressing. Otherwise look good.
I like these `Note` and `Important` tags, thanks!

examples/diffusion/python_stable_diffusion_xl/README.md
kahmed10: good catch!

umangyadav (approved):

PR #2977, opened 2024-04-17
Moving CI to 6.1

causten:
ROCm 6.1 was released today.  Bumping CI up


pfultz2:
I think there might be another package thats needed to install clang-format.

pfultz2:
I think rocm-llvm-dev is the package that needs to be installed.

Dockerfile
```diff
@@ -47,6 +47,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     rccl-dev \
     rocm-smi-lib \
     rocm-dev \
+    rocm-llvm-dev \
```
pfultz2, line 50: This should be added to install_preqs.sh file instead.

pfultz2:
We should disable the `cppcoreguidelines-missing-std-forward` and `modernize-type-traits` tidy checks and then open an issue where we can address this in a seperate PR.

src/argument.cpp
```diff
@@ -23,7 +23,11 @@
  */
 #include <migraphx/argument.hpp>
 #include <migraphx/functional.hpp>
+#include <migraphx/make_shared_array.hpp>
+#include <migraphx/shape.hpp>
+#include <functional>
 #include <unordered_map>
+#include <utility>
```
pfultz2, line 30: If this is for clang-tidy we should just disable the check for now. I am thinking we might want to disable it completely as this will just add a ton of includes to every file.

CMakeLists.txt
```diff
@@ -203,6 +203,7 @@ rocm_enable_clang_tidy(
         -google-readability-*
         -google-runtime-int
         -google-runtime-references
+	-misc-include-cleaner
```
pfultz2, line 206: You need to disable `cppcoreguidelines-missing-std-forward` and `modernize-type-traits`.

pfultz2:
I think you will need to suppress `cppcoreguidelines-use-default-member-init`, and `bugprone-switch-missing-default-case,-warnings-as-errors`. For the `cppcoreguidelines-noexcept-swap` warning you can just add `noexcept` to the swap function.

TedThemistokleous:
This still needed or do we want to close this and aim to merge in 6.2 once it's ready?

pfultz2:
> This still needed or do we want to close this and aim to merge in 6.2 once it's ready?

We should try to get this merged in. We just need to disable a few more tidy checks:

* cppcoreguidelines-use-default-member-init
* bugprone-switch-missing-default-case

@causten Can you add these exceptions, so we can get this merged?

TedThemistokleous:
> > This still needed or do we want to close this and aim to merge in 6.2 once it's ready?
> 
> We should try to get this merged in. We just need to disable a few more tidy checks:
> 
> * cppcoreguidelines-use-default-member-init
> * bugprone-switch-missing-default-case
> 
> @causten Can you add these exceptions, so we can get this merged?

Can we target 6.1.3 then instead of 6.1?

pfultz2:
> Can we target 6.1.3 then instead of 6.1?

I opened #3335 to upgrade to this version and fix the tidy issues.

PR #2978, opened 2024-04-17
Use number of specified thread for the compilation

umangyadav:
Currently `MIGRAPHX_GPU_COMPILE_PARALLEL` has no effect because `par_for.hpp` is not passing `min_grain` param. This PR fixes that.

### DIFF
--- a/src/include/migraphx/par.hpp
+++ b/src/include/migraphx/par.hpp
@@ -94,7 +94,7 @@ OutputIt par_transform(
 }
 
 template <class InputIt, class UnaryFunction>
-void par_for_each(InputIt first, InputIt last, UnaryFunction f)
+void par_for_each(InputIt first, InputIt last, size_t min_grain, UnaryFunction f)
 {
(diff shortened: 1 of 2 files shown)


pfultz2:
We shouldn't do this. The `min_grain` will be ignored when we are using ParallelSTL. Instead, we should use `simple_par_for` directly to ensure it will always create threads. We could probably have `par_for` with `min_grain` call `simple_par_for` directly and the overload without the `min_grain` can use `par_for_each`.

pfultz2 (approved):

TedThemistokleous:
> We shouldn't do this. The `min_grain` will be ignored when we are using ParallelSTL. Instead, we should use `simple_par_for` directly to ensure it will always create threads. We could probably have `par_for` with `min_grain` call `simple_par_for` directly and the overload without the `min_grain` can use `par_for_each`.

Ah, that makes better sense and good catch. Didn't really think about the parallel STL case.

PR #2982, opened 2024-04-18
Benchmark candidate programs instead of code objects in compile_ops

umangyadav:
Split-k GEMM compilation with MLIR would require unfusing dot + pointwise. 
Dot further requires a "Fill" instruction to instantiate output buffer with zeros.
While benchmarking different candidate during tuning, MIGraphX would need to account for the `hip::fill` call as well for the Splitk-GEMM configuration. 

Creating a program using `compiler_replace` provided by JIT compiler of MLIR would insert all the necessary calls and pointwise op to run Split-K gemm. That way MIGraphX can account for `hip::fill` calls as well.

This PR makes necessary changes to create a program using "compiler_replace" when doing benchmarking.

### DIFF
--- a/src/include/migraphx/program.hpp
+++ b/src/include/migraphx/program.hpp
@@ -51,7 +51,6 @@ struct program_impl;
 
 struct marker;
 
-
 /**
  * @brief Stores the instruction stream
  */
@@ -81,9 +80,7 @@ struct MIGRAPHX_EXPORT program
     std::vector<argument> eval(parameter_map params,
                                execution_environment exec_env = execution_environment{}) const;
 
-    std::vector<argument> eval_with_context(
-                                   std::vector<context>& ctx,
-                                   parameter_map params) const;
+    std::vector<argument> eval_with_context(std::vector<context>& ctx, parameter_map params) const;
 
     void finish() const;
 
--- a/src/include/migraphx/program.hpp
+++ b/src/include/migraphx/program.hpp
@@ -51,9 +51,6 @@ struct program_impl;
 
 struct marker;
 
-std::vector<argument> generic_eval(const migraphx::program& p,
-                                   std::vector<context>& ctx,
-                                   std::unordered_map<std::string, argument> params);
 
 /**
  * @brief Stores the instruction stream
@@ -83,6 +80,11 @@ struct MIGRAPHX_EXPORT program
 
     std::vector<argument> eval(parameter_map params,
                                execution_environment exec_env = execution_environment{}) const;
+
+    std::vector<argument> eval_with_context(
+                                   std::vector<context>& ctx,
+                                   parameter_map params) const;
+
     void finish() const;
 
     std::size_t size() const;
--- a/src/include/migraphx/program.hpp
+++ b/src/include/migraphx/program.hpp
@@ -51,6 +51,10 @@ struct program_impl;
 
 struct marker;
 
+std::vector<argument> generic_eval(const migraphx::program& p,
+                                   std::vector<context>& ctx,
+                                   std::unordered_map<std::string, argument> params);
+
 /**
  * @brief Stores the instruction stream
  */
@@ -79,7 +83,6 @@ struct MIGRAPHX_EXPORT program
 
     std::vector<argument> eval(parameter_map params,
                                execution_environment exec_env = execution_environment{}) const;
-
     void finish() const;
 
     std::size_t size() const;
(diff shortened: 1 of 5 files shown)


src/include/migraphx/program.hpp
```diff
@@ -51,6 +51,10 @@ struct program_impl;
 
 struct marker;
 
+std::vector<argument> generic_eval(const migraphx::program& p,
```
pfultz2, line 54: Make this a member function and just call it `eval`.

src/include/migraphx/program.hpp
```diff
@@ -51,6 +51,10 @@ struct program_impl;
...
+                                   std::vector<context>& ctx,
+                                   std::unordered_map<std::string, argument> params);
```
pfultz2, line 56: Use `parameter_map` typedef here.

src/targets/gpu/include/migraphx/gpu/time_op.hpp
```diff
@@ -35,6 +36,8 @@ namespace gpu {
 MIGRAPHX_GPU_EXPORT double
 time_op(context& ictx, operation op, const std::vector<shape>& inputs, int n = 100);
 
+MIGRAPHX_GPU_EXPORT double time_program(context& ictx, migraphx::program p, int n = 100);
```
TedThemistokleous, line 39: Wouldn't we want something more statistically significant like say 1k runs instead of 100?

src/targets/gpu/time_op.cpp
```diff
@@ -64,6 +65,34 @@ double time_op(context& ictx, operation op, const std::vector<shape>& inputs, in
...
+    run();
+    gctx.get_stream().record(start.get());
+    for(auto i : range(n))
+    {
+        (void)i;
+        run();
+    }
+    gctx.get_stream().record(stop.get());
+    gctx.finish();
+    return context::get_elapsed_ms(start.get(), stop.get()) / n;
```
TedThemistokleous: Are we not concerned about any cold start conditions here if we're taking all the runs or is the assumption here that the initial run before handles cold start?

src/targets/gpu/include/migraphx/gpu/time_op.hpp
umangyadav: this would get called while benchmarking multiple candidate solutions while doing exhaustive tuning. 
There can be hundreds for applicable solutions for a single gemm or other operator. Running each candidate 1k time would make it a lot slower. 

100 is just a default parameter.  While benchmarking it simply uses 20. 

Doing 1k runs may also have adverse effect if GPU becomes hot and it may throttle.

src/targets/gpu/time_op.cpp
umangyadav: It does one run before starting the loop to heat up the GPU.

src/include/migraphx/program.hpp
umangyadav: Done.

src/include/migraphx/program.hpp
umangyadav: Done.

src/targets/gpu/include/migraphx/gpu/time_op.hpp
TedThemistokleous: understood

src/targets/gpu/time_op.cpp
```diff
@@ -41,35 +42,58 @@ std::vector<argument> generate_arguments(const std::vector<shape>& shapes, unsig
...
+    return time_loop(gctx, n, run);
+}
+
+double time_op(const context& ictx, operation op, int n)
 {
     auto inputs = any_cast<migraphx::gpu::code_object_op>(op).expected_inputs;
     return time_op(ictx, op, inputs, n);
 }
 
+double time_program(const context& ictx, migraphx::program p, int n)
```
pfultz2, line 80: You can just use `program` directly instead of `migraphx::program`.

src/targets/gpu/compile_ops.cpp
```diff
@@ -185,17 +186,29 @@ struct compile_plan
...
-                               cobjs.end(),
-                               double{0},
-                               std::plus<>{},
-                               [&](const operation& op) { return time_op(*ctx, op, 20); });
+                           /*
+                           create a small program with insturction being compiled and call "replace"
+                           on that which would insert all the compiled code objects, prefills etc.
+                           necessary to run candidate code object
+                           */
+                           migraphx::program bench_prog;
```
pfultz2, line 194: You can just use `program` directly instead of `migraphx::program`.

src/targets/gpu/time_op.cpp
umangyadav: Done.

src/targets/gpu/compile_ops.cpp
umangyadav: Done.

pfultz2 (approved):

TedThemistokleous (approved):

PR #2987, opened 2024-04-22
Limit parallelism for constant propagation when parallel STL is not enabled

ahsan-ca:
This limits the parallelism for constant propagate pass when parallel STL is not enabled.

### DIFF
--- a/src/propagate_constant.cpp
+++ b/src/propagate_constant.cpp
@@ -28,6 +28,7 @@
 #include <migraphx/functional.hpp>
 #include <migraphx/simple_par_for.hpp>
 #include <migraphx/env.hpp>
+#include <thread>
 #include <unordered_set>
 
 namespace migraphx {
@@ -83,7 +84,12 @@ void propagate_constant::apply(module& m) const
     // Compute literals in parallel
     std::vector<instruction_ref> const_instrs_vec{const_instrs.begin(), const_instrs.end()};
     std::vector<argument> literals(const_instrs_vec.size());
-    simple_par_for(const_instrs_vec.size(), 1, [&](const auto i) {
+    int n = 1;
+#ifndef ParallelSTL_FOUND
+    n = std::max<std::size_t>(
+        std::ceil(static_cast<double>(1024) / std::thread::hardware_concurrency()), 1);
+#endif
+    simple_par_for(const_instrs_vec.size(), n, [&](const auto i) {
         literals[i] = const_instrs_vec[i]->eval();
     });
 


src/propagate_constant.cpp
```diff
@@ -83,7 +84,12 @@ void propagate_constant::apply(module& m) const
     // Compute literals in parallel
     std::vector<instruction_ref> const_instrs_vec{const_instrs.begin(), const_instrs.end()};
     std::vector<argument> literals(const_instrs_vec.size());
-    simple_par_for(const_instrs_vec.size(), 1, [&](const auto i) {
+    int n = 1;
+#ifndef ParallelSTL_FOUND
```
pfultz2, line 88: You should do `#if !MIGRAPHX_HAS_EXECUTORS`. I dont think `ParallelSTL_FOUND` is ever defined.

causten (approved):
Code change works.  Get approval from Umang and Paul, then create a PR for the `release/rocm-rel-6.1` branch.  Not sure why the CI is reporting Perf and Accuracy checks  failing.

src/propagate_constant.cpp
```diff
@@ -83,7 +84,12 @@ void propagate_constant::apply(module& m) const
     // Compute literals in parallel
     std::vector<instruction_ref> const_instrs_vec{const_instrs.begin(), const_instrs.end()};
     std::vector<argument> literals(const_instrs_vec.size());
-    simple_par_for(const_instrs_vec.size(), 1, [&](const auto i) {
+    int n = 1;
```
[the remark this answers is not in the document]
ahsan-ca: Done.

umangyadav (approved):

ahsan-ca:
Stable diffusion perf run time:
```
[ MIGraphX Version: 2.10.0. ] Complete: ./bin/migraphx-driver perf ../model.onnx --input-dim @sample 2 4 64 64 @timestep 1 @encoder_hidden_states 2 64 1024 --fp16

with my change:
real    3m56.950s
user    65m16.201s
sys     2m49.582s

without my change:
real    3m57.585s
user    64m47.968s
sys     3m33.107s
```
The change does not seem to have a significant affect on the compile time.

src/propagate_constant.cpp
```diff
@@ -83,7 +84,12 @@ void propagate_constant::apply(module& m) const
     // Compute literals in parallel
     std::vector<instruction_ref> const_instrs_vec{const_instrs.begin(), const_instrs.end()};
     std::vector<argument> literals(const_instrs_vec.size());
-    simple_par_for(const_instrs_vec.size(), 1, [&](const auto i) {
+    std::size_t n = 1;
+#if !MIGRAPHX_HAS_EXECUTORS
+    n = std::max<std::size_t>(
```
pfultz2, line 89: This is calculating the number of threads, it should be calculating the grainsize. You need to do something like `grainsize = const_instrs_vec.size() / n` to compute the grainsize.

src/propagate_constant.cpp
```diff
@@ -83,7 +84,12 @@ void propagate_constant::apply(module& m) const
     // Compute literals in parallel
     std::vector<instruction_ref> const_instrs_vec{const_instrs.begin(), const_instrs.end()};
     std::vector<argument> literals(const_instrs_vec.size());
-    simple_par_for(const_instrs_vec.size(), 1, [&](const auto i) {
+    std::size_t n = 1;
```
pfultz2, line 87: This variable should be named `grainsize` to make it clear its the grainsize.

src/propagate_constant.cpp
ahsan-ca: `grainsize = const_instrs_vec.size() / n`. If I understand correctly, we should rename `n` here to `grainsize`, what does `n` refer to in the calculation of `grainsize`?

ahsan-ca: Discussed this over the call with Paul and here is what we would like it to be:
```
#if !MIGRAPHX_HAS_EXECUTORS
    std::size_t n = std::max<std::size_t>(2048 / std::thread::hardware_concurrency(), 1);
    grainsize = const_instrs_vec.size()/n;  
#endif
```
I will open up a PR and make the appropriate changes.

src/propagate_constant.cpp
ahsan-ca: Since this PR was merged, I have made this change in a follow up PR: https://github.com/ROCm/AMDMIGraphX/pull/2993

src/propagate_constant.cpp
```diff
@@ -83,7 +84,12 @@ void propagate_constant::apply(module& m) const
...
+#if !MIGRAPHX_HAS_EXECUTORS
+    n = std::max<std::size_t>(
+        std::ceil(static_cast<double>(1024) / std::thread::hardware_concurrency()), 1);
```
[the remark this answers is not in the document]
ahsan-ca: I used `ceil` to be more conservative with limiting the parallelism, models were failing without this. However, I have removed this in the new changes as we limit it in another way: https://github.com/ROCm/AMDMIGraphX/pull/2993

src/propagate_constant.cpp
ahsan-ca: PR opened with the changes: https://github.com/ROCm/AMDMIGraphX/pull/2993

PR #2989, opened 2024-04-22
Fix compute_type_name() when MSVC is used

apwojcik:
The PR is related to the Microsoft Visual C++ compiler when building applications with the MIGraphX C++ interface.
We require that fix for the ONNXRT MIGraphX EP on Windows, which we build with Visual C++ compiler (not HIP SDK C++ compiler). The current implementation of `compute_type_name()` incorrectly decyphers the type name from the `typeid()` operator. Additionally, it does not work on incomplete types. It is hard, however, to obtain the same results as for __PRETTY_FUNCTION__ from the `typeid()` operator. So, on Windows, we use `__FUNCSIG__` instead.
The PR changes the function to resolve type names at compile time instead of runtime. Therefore, there is no need for a static storage duration for type names.

### DIFF
--- a/src/api/include/migraphx/migraphx.hpp
+++ b/src/api/include/migraphx/migraphx.hpp
@@ -34,6 +34,7 @@
 #include <exception>
 #include <array>
 #include <vector>
+#include <string_view>
 #include <cassert>
 #include <iostream>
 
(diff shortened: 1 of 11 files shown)


apwojcik:
CppCheck suggests using `std::string::starts_with()` instead of `std::string::find() == 0`, but this method is a C++20 feature, and we are using C++17.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -34,6 +34,7 @@
 #include <exception>
 #include <array>
 #include <vector>
+#include <string_view>
```
pfultz2, line 37: This cant use `string_view`. This is the API header and should be compatible with older C++ versions that users might be using.

pfultz2:
> CppCheck suggests using std::string::starts_with() instead of std::string::find() == 0, but this method is a C++20 feature, and we are using C++17.

Just add an inline suppression.

pfultz2:
Also it looks like there is an extra space added when using gcc.

src/include/migraphx/type_name.hpp
```diff
@@ -35,8 +35,24 @@ std::string compute_type_name()
...
+    const char function_name[]  = "compute_type_name<";
+    const char parameter_name[] = ">(void)";
+    const char cdecl_name[]     = "__cdecl";
+
+    name = __FUNCSIG__;
+
+    auto begin  = name.find(function_name) + sizeof(function_name) - 1;
+    auto length = name.find(parameter_name) - begin;
+    name        = name.substr(begin, length);
+    if(starts_with(class_name))
```
pfultz2, line 49: Shouldn't this be `starts_with(name, class_name)`? I dont really understand how this even compiles.

apwojcik: Yes, thanks for noticing. Why this has not generated a compile error?

apwojcik: Oh, we do not use MSVC to compile MIGraphX. I will stip out this whole section.

apwojcik: Yes, thanks for noticing. Why this has not generated a compile error?

pfultz2 (approved):

PR #2993, opened 2024-04-23
Limit parallelism more conservatively for constant propagate when parallel STL is not enabled

ahsan-ca:
This is a follow up PR to:
Limit parallelism for constant propagation when parallel STL is not enabled (#2987).
This limits parallelism more conservatively.


ahsan-ca:
Stable diffusion perf run time:
```
[ MIGraphX Version: 2.10.0. ] Complete: ./bin/migraphx-driver perf ../model.onnx --input-dim @sample 2 4 64 64 @timestep 1 @encoder_hidden_states 2 64 1024 --fp16

with my change:
real    4m5.554s
user    66m41.772s
sys     2m51.023s

without my change:
real    3m57.585s
user    64m47.968s
sys     3m33.107s
```
The change does not seem to have a significant affect on the compile time.

pfultz2 (approved):

src/propagate_constant.cpp
```diff
@@ -84,12 +84,12 @@ void propagate_constant::apply(module& m) const
     // Compute literals in parallel
     std::vector<instruction_ref> const_instrs_vec{const_instrs.begin(), const_instrs.end()};
     std::vector<argument> literals(const_instrs_vec.size());
-    std::size_t n = 1;
+    std::size_t grainsize = 1;
 #if !MIGRAPHX_HAS_EXECUTORS
-    n = std::max<std::size_t>(
-        std::ceil(static_cast<double>(1024) / std::thread::hardware_concurrency()), 1);
+    std::size_t n = std::max<std::size_t>(2048 / std::thread::hardware_concurrency(), 1);
+    grainsize     = const_instrs_vec.size() / n;
```
lakhinderwalia, line 90: Can this grainsize be zero? The earlier code seemed to avoid that case.. Thanks.

ahsan-ca: Yes, the grainsize can be zero for the case `const_instrs_vec.size() == 0`. This, however, should not cause any issues. The threadsize in `simple_par_for` has been calculated to zero when `n(const_instrs_vec.size()) == 0`, even before. .
```
    const auto threadsize = std::min<std::size_t>(std::thread::hardware_concurrency(),
                                                  n / std::max<std::size_t>(1, min_grain));
```

PR #2999, opened 2024-04-25
Remove `--model` flag in `migraphx-driver` and instead add `--test` flag

umangyadav:
### **_Update : See recent comment below._** 

Scale is calculated as 
`param_pair.first = quantized_range / max_abs_vals->at(ins_index);`

quantized_range for int8 is : 127

If `max_abs_vals` is inf then `scale` would come out to be zero. 

MIGraphX would detect zero and propagate that which later leads to compilation errors while doing JIT: 

```
root@banff-cyxtera-s81-2:/code/build# bin/driver perf --model resnet50 --fp16 --int8
Running [ MIGraphX Version: 2.10.0. ]: bin/driver perf --model resnet50 --fp16 --int8
Compiling ...
/tmp/comgr-24d775/input/main.cpp:39:56: error: unused parameter '_x0' [-Werror,-Wunused-parameter]
__device__ __attribute__((const)) auto pointwise2(T_x0 _x0) {
```

It is another issue why "infs" are being generated. 

One example is Batchnorm 
```
x3 = pow(x1, x2) 
x4 = recip(x3)
```
The way "resnet.cpp" is written "pow" can produce zeros and "recip" would generate Infs  which are later propagated. 

Solving this generically requires knowing valid data range for all the inputs for each of the operators.


umangyadav:
Update : 

It was decided to remove `--model` completely and add `--test` flag to quickly test migraphx. Changes in this PR reflect that.

kahmed10 (approved):

pfultz2 (approved):

causten (commented):
https://rocm.docs.amd.com/projects/radeon/en/latest/docs/install/install-migraphx.html
- Open an internal ticket to fix this for the next release

PR #3002, opened 2024-04-25
Improvements to inner_broadcast

pfultz2:
This fixes a bug with inner_broadcast and improves the handling for more cases. For the simple cases with the same broadcasts and dimension, it will just insert the broadcasting afterwards.

However, for the case with different dimensions and broadcasts, it will find the axes that are being broadcasted and not broadcasted and it will insert squeezes and unsqueezes as necessary. Before we would just use squeeze with empty axes, which essentially removed all single dimensions, but now we explicitly call out which axes we want to remove.


src/simplify_algebra.cpp
```diff
@@ -506,65 +518,124 @@ struct find_inner_broadcast
...
+                });
+            auto op = insert_common_op(m, ins, ins->get_operator(), inputs);
+            if(iaxes.size() == 1)
+            {
+                m.replace_instruction(
+                    ins,
+                    make_op("broadcast",
+                            {{"axis", iaxes.front()}, {"out_lens", ins->get_shape().lens()}}),
+                    op);
+            }
```
CharlieL7, line 626: Why use `broadcast` rather than `multibroadcast`?
I understand that for one "inner axis" case that `broadcast` can be used directly rather than `multibroadcast` which might need an `unsqueeze` operator. But, this makes the code more complicated for a seemingly unnoticeable benefit. Another option would be to use `broadcast` for everything since the effective difference between the two operators if that broadcast works left-wise with an offset while multibroadcast works right-wise like in numpy.

CharlieL7 (commented):
Would be helpful to have some more comments about what the algo. is trying to do.

This PR brings up a gripe I've had with `broadcast` vs. `multibroadcast`. It's quite unwieldy to have these two slightly different broadcasting instructions that sound like they would do the same thing but actually work quite differently. I think it's better to heavily prefer one over the other (either one). That way we have consistent shape broadcasting rules.

pfultz2:
> This PR brings up a gripe I've had with broadcast vs. multibroadcast. It's quite unwieldy to have these two slightly different broadcasting instructions that sound like they would do the same thing but actually work quite differently. I think it's better to heavily prefer one over the other (either one). That way we have consistent shape broadcasting rules.

Yea I do thinks its annoying to have two different. Originally, `broadcast` was used to broadcast a single dimension, and `multibroadcast` was used when broadcasting a tensor with multiple dimensions. I didnt realize until later that `broadcast` could be used for the same thing as `multibroadcast`, had I knew that back then I would've probably rejected the `multibroadcast` operator. I would love to get rid of it completely but there is still a lot of places that use it and rely on it(like `fuse_reduce` that only looks for `multibroadcastt` [here](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/fuse_reduce.cpp#L230)), so it would be a big refactoring.

src/simplify_algebra.cpp
pfultz2: I think its to keep the same semantics of `broadcast` being for one dimension and `multibroadcast` for multiple. There are places that only look for `multibroadcast` so this could be a problem, and using it without a unsqueeze simplifies the graph for other optimizations. 

Furthermore this reduces the amount of test cases that need to be updated.

src/simplify_algebra.cpp
```diff
@@ -482,17 +482,40 @@ struct find_double_add_lit_broadcast
     }
 };
 
+/// Find elementswise operators that have all broadcast inputs. It then
+/// rewrites the elementwise to do the computation on the non-broadcasted
+/// axes, and then broadcast that result.
```
bpickrel, line 487: Could you add (not necessarily on this line) a definition of an inner broadcast and an inner broadcast axis.  It's not self-evident.

pfultz2: `find_inner_broadcast` is the name of the matcher/finder(and this comment explains what that is looking for). Without a larger context, I dont know what an inner broadcast axis is, and I cant find anywhere this term is used.

CharlieL7 (approved):
just improve test coverage

pfultz2:
Added tests for better coverage.

src/simplify_algebra.cpp
bpickrel: I'm not going to explain what it means, because I'm ***asking*** what it means.

pfultz2: I dont know what inner broadcast axis is either, and I dont even know why its relevant to this PR.

shivadbhavsar (approved):

PR #3008, opened 2024-04-26
Add softmax cross entropy

TedThemistokleous:
Adding softmaxcrossentropy op in MIGraphX. This is used in Llama v2

https://github.com/onnx/onnx/blob/main/docs/Operators.md#softmaxcrossentropyloss

This function is a bit complicated but is able to be built up by smaller Onnx ops

I've added the script I've used to generate the gold data for the verify test (all verify tests pass as of now)

Need to update parser tests to match the result as well as adding in K-Dimension verify tests, but looking to get this into review sooner than later incase there's anything egregious or something that needs to be changed here and I'll be gone this Friday for two weeks. 

The initial motivation to implement this originated from the onnx ref implementation:
https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_softmax_cross_entropy_loss.py

Operators like where and take() were used but it seems I was able to get around this
1. Use the fact that default weights are all 1, thus, for ignore index, apply the corresponding weight for the ignored class index to 0 and scale. This avoids the need for a where operator and reduces things to a simple mul. This also works in the case where we have non default, or 1 weights.

2. Use gather on the weight based on labels to mirror what take() is doing. 

The test verification was done as follows

- Due to softmax being used we can sanity check this mathematically as all zero and all ones resolve to the same values
- Mean and sum shouldn't change based on scatter choices this more tests have been added for the no_reduction cases
- Ignore index and 1 weights are also added to scale classes accordingly
- use softmaxcrossentropyloss_gold_data_check.py to generate the appropriate gold data and verify the result (can be removed before PR is merged)

As some of the banter between @pfultz2 and have suggested, we can also reuse a lot of this and mirror these changes to get NegativeLogLikelihoodLoss (seperate PR) once we decide on this, by removing the softmax frontend to calculate loss.


pfultz2:
So `SoftmaxCrossEntropyLoss` is a onnx function which means it can be implemented using a composite of other onnx operators(like `DynamicQuantizeLinear`). So we shouldn't need to add a new operator for this. I couldn't find a reference of the ONNX operators needed. However, onnxrt has transformation of `LogSoftmax->[Cast]->NegativeLogLikelihoodLoss` to `SoftmaxCrossEntropyLoss`:

https://github.com/microsoft/onnxruntime/blob/3e4db2c6869975a4e34a71a36c67c5a825f669e2/orttraining/orttraining/core/optimizer/loss_rewriter.h#L11

Furthermore `NegativeLogLikelihoodLoss` is also an ONNX function as well, so we dont need to write a custom operator for this as well, although I couldn't find a reference of the ONNX operators, but looking at the ref implementation here:

https://github.com/onnx/onnx/blob/4e7289d05a08453b3f20efeb774d4590b5428903/onnx/reference/ops/op_negative_log_likelihood_loss.py#L74

It looks like this can be implemented using some combination of scatter+where+reshape+reduce.

TedThemistokleous:
> So `SoftmaxCrossEntropyLoss` is a onnx function which means it can be implemented using a composite of other onnx operators(like `DynamicQuantizeLinear`). So we shouldn't need to add a new operator for this. I couldn't find a reference of the ONNX operators needed. However, onnxrt has transformation of `LogSoftmax->[Cast]->NegativeLogLikelihoodLoss` to `SoftmaxCrossEntropyLoss`:
> 
> https://github.com/microsoft/onnxruntime/blob/3e4db2c6869975a4e34a71a36c67c5a825f669e2/orttraining/orttraining/core/optimizer/loss_rewriter.h#L11
> 
Ah I was set on finding a CrossEntropyLoss  and didn't see anything  and that's why I was writing the cross entropy piece.


> Furthermore `NegativeLogLikelihoodLoss` is also an ONNX function as well, so we dont need to write a custom operator for this as well, although I couldn't find a reference of the ONNX operators, but looking at the ref implementation here:
> 
> https://github.com/onnx/onnx/blob/4e7289d05a08453b3f20efeb774d4590b5428903/onnx/reference/ops/op_negative_log_likelihood_loss.py#L74
> 
> It looks like this can be implemented using some combination of scatter+where+reshape+reduce(`np.take` can be implemented using scatter).

Okay, so I guess we'll add a NegativeLossLikelihood Loss then for the SoftmaxCase, leverage the existing ops + softmax on the input?

I know you've mentioned SoftmaxLog() and splitting things by using softmax + log instead so in this case should we do a seperate scatter+where+reshape+reduce for each op or is it better to perform Softmax + NegativeLossLikelihood then?

pfultz2:
> > Furthermore `NegativeLogLikelihoodLoss` is also an ONNX function as well, so we dont need to write a custom operator for this as well, although I couldn't find a reference of the ONNX operators, but looking at the ref implementation here:
> > https://github.com/onnx/onnx/blob/4e7289d05a08453b3f20efeb774d4590b5428903/onnx/reference/ops/op_negative_log_likelihood_loss.py#L74
> > It looks like this can be implemented using some combination of scatter+where+reshape+reduce(`np.take` can be implemented using scatter).
> 
> Okay, so I guess we'll add a NegativeLossLikelihood Loss then for the SoftmaxCase, leverage the existing ops + softmax on the input?

No need to create a new operator such as `NegativeLossLikelihood` since this can also be implemented using our current operators. We may want to make note of this because we could add support for the `NegativeLossLikelihood` operator easily in the future reusing the same set of operators.

TedThemistokleous:
> > > Furthermore `NegativeLogLikelihoodLoss` is also an ONNX function as well, so we dont need to write a custom operator for this as well, although I couldn't find a reference of the ONNX operators, but looking at the ref implementation here:
> > > https://github.com/onnx/onnx/blob/4e7289d05a08453b3f20efeb774d4590b5428903/onnx/reference/ops/op_negative_log_likelihood_loss.py#L74
> > > It looks like this can be implemented using some combination of scatter+where+reshape+reduce(`np.take` can be implemented using scatter).
> > 
> > 
> > Okay, so I guess we'll add a NegativeLossLikelihood Loss then for the SoftmaxCase, leverage the existing ops + softmax on the input?
> 
> No need to create a new operator such as `NegativeLossLikelihood` since this can also be implemented using our current operators. We may want to make note of this because we could add support for the `NegativeLossLikelihood` operator easily in the future reusing the same set of operators.

That's what I figured after going over the links you showed me. I guess this is kind of a 2:1 then.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,316 @@
...
+        std::set<migraphx::shape::type_t> supported_label_types = {migraphx::shape::int32_type,
+                                                                   migraphx::shape::int64_type};
+        if(not contains(supported_label_types, label_shape.type()))
+        {
+            MIGRAPHX_THROW("softmaxcrossentropyloss: Label must be either int32 or int64 type");
+        }
+
+        // TODO: Update this operator when we get bfloat16 support
+        std::set<migraphx::shape::type_t> supported_score_weight_types = {
+            migraphx::shape::half_type, migraphx::shape::float_type, migraphx::shape::double_type};
```
pfultz2, line 183: We should be able to support all float types including fp8(even though onnx spec doesnt mention it). So it would be simpler to just check if its supported with `not shape::is_integral(type)`.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,316 @@
...
+                                       const std::vector<instruction_ref>& args) const
+    {
+        // Get and handle attributes
+        std::string reduction = "mean";
+        if(contains(info.attributes, "reduction"))
+        {
+            std::set<std::string> supported_modes = {"mean", "none", "sum"};
+
+            reduction = info.attributes.at("reduction").s();
+            if(not contains(supported_modes, reduction))
```
pfultz2, line 143: Just put the support modes inline: `if(not contains({"mean", "none", "sum"}, reduction))`

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+        if(scores_shape.ndim() < 2)
+        {
+            MIGRAPHX_THROW("softmaxcrossentropyloss: Scores must be two or more dimensions [batch, "
+                           "class_size, D1...Dk]");
+        }
+
+        std::set<migraphx::shape::type_t> supported_label_types = {migraphx::shape::int32_type,
+                                                                   migraphx::shape::int64_type};
+        if(not contains(supported_label_types, label_shape.type()))
```
pfultz2, line 176: This can be removed. The labels are used for input indices to scatter and gather and we support any type for those indices.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,316 @@
...
+        size_t batch_size = scores_shape.lens().at(0);
+        size_t class_size = scores_shape.lens().at(1);
+
+        auto loss_tensor = info.add_instruction(
+            migraphx::make_op("convert", {{"target_type", scores_shape.type()}}), labels);
+
+        // Default weights will always be 1
+        std::vector<float> weight_vec(class_size, 1.0f);
+        auto weights = info.add_literal(
+            migraphx::literal(migraphx::shape(scores_shape.type(), {class_size}), weight_vec));
```
pfultz2, line 212: I think it would be better to insert a scalar 1 and broadcast it to the tensor size.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: I suppose that makes sense. I can do that

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Okay. Are we not worried about the inputs being to spec regardless if we can handle them? Should I just flag this as a warning in someway?

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: All good I can change this. You're thinking broadcast + next ops can be fused easier than with a large literal?

TedThemistokleous: Done btw

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Removed.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Done

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Done

src/onnx/parse_softmaxcrossentropyloss.cpp
pfultz2: > Are we not worried about the inputs being to spec regardless if we can handle them?

I dont see a problem with it, and if the spec is updated to support more types we dont need to make any changes.

TedThemistokleous: all good well Ive removed it so then we dont need to worry about it then. I think this means I can also remove the test case I added for the valid/invalid input

src/onnx/parse_softmaxcrossentropyloss.cpp
umangyadav: Is this resolved ? I do not see check for shape::is_integral.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,308 @@
...
+        {
+            MIGRAPHX_THROW(
+                "softmaxcrossentropyloss: Score and Labels must identical batch size inputs");
+        }
+
+        if((scores_shape.ndim() - 1) != label_shape.ndim())
+        {
+            MIGRAPHX_THROW(
+                "softmaxcrossentropyloss: Score and Labels must contain identical K-Dimensions");
+        }
```
umangyadav, line 191: You can also check for all the extra dimensions have same number of elements or not. 
E.g. 
Scores = [N, C, D1, D2....., Dk] 
Lables = [N, d1, d2,....,dk] 
then `D1 * D2.. * Dk ==d1 * d2 *.. * dk`

Shouldn't really be necessary to check because otherwise onnx model generation itself should throw because it would be invalid onnx. 

But since we are checking all the constraints, might as well check this one additionally.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            if(weights_shape.type() != scores_shape.type())
+            {
+                MIGRAPHX_THROW(
+                    "softmaxcrossentropyloss: Weight and Scores inputs must be the same type");
+            }
+        }
+
+        // adjust weights based on ignore index is that's set
```
umangyadav, line 232:
```suggestion
        // adjust weights based on ignore index if that's set
```

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,308 @@
...
+        if(has_ignore_index)
+        {
+            auto weights_shape = weights->get_shape();
+            std::vector<float> zero_val_vect(weights_shape.elements(), 0);
+            auto zero_val = info.add_literal(migraphx::literal(weights_shape, zero_val_vect));
+            weights       = info.add_instruction(
+                migraphx::make_op("scatter_none", {{"axis", 0}}), weights, ignore_index, zero_val);
+        }
```
umangyadav, line 240: I do not think this directly corresponds to onnx reference code here 
https://github.com/onnx/onnx/blob/b90e252da11dea9bdc191d6b9b8d01511ef3e3bd/onnx/reference/ops/op_softmax_cross_entropy_loss.py#L33 

Weights have shape `[C]` where C is number of classes. 
From what i understand each element in weight corresponds to weight for that class index's weight. 

e.g. let's say weights  = [3, 2, 1] then weight for class_index = 0 would be 3, weight for class_index = 1 would be 2 weight for class_index = 2 would be 1. 

That is why onnx reference code is first doing `np.take` which would map weights to `target`. 

e.g. if lets' say target_indices = [1, 1, 2] then actual weights to use would be [2, 2, 1]. 

And then ignore_index is ignore_index for the value inside target_indices vector. 

So let's say if `ignore_index = 2` then target_indices would become [1, 1, ignore] and then actual weights to use would become `[2, 2, 0]` 

You are scattering zero values inside the weight matrix directly which is not the same as above and onnx's ref.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,308 @@
...
+            auto lens = label_shape.lens();
+            size_t d =
+                std::accumulate(lens.rbegin(), lens.rend() - 1, 1, std::multiplies<size_t>());
+            scores = info.add_instruction(
+                migraphx::make_op("reshape", {{"dims", {batch_size, class_size, d}}}), scores);
+        }
+
+        // Offload calculation of log(Softmax(scores)) for the input before we perform cross entropy
+        // loss calculation
+        auto softmax_scores = info.add_instruction(migraphx::make_op("softmax"), scores);
```
umangyadav, line 253: You need to do softmax on axis 1.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            if(reduction == "mean")
+            {
+                auto reduced_weights = info.add_instruction(
+                    migraphx::make_op("reduce_sum", {{"axes", {0}}}), weight_tensor);
+                loss_tensor =
+                    info.add_instruction(migraphx::make_op("div"), loss_tensor, reduced_weights);
+            }
+        }
```
umangyadav, line 300: Better to first handle 
`reduction == mean and has_weights` case. 
then you can do do `reduction == mean` case
`reduction == sum` case 
and as the default case just return "None" case.

Similar to onnx ref implementation

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Actually this is a good point. I  was playing with dimensions today and added more batches, or classes and found another issue. Valid or not it's likely worth checking since its assumed based on the spec that D1-Dk are the same, and thus mismatches would be an error

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Ah the intent here was to avoid the use of where here and needing to create a literal of masks to set the zero value for the weight index. At the end I  always do the multiply with the weights so for the default case , everything in weighs is all 1s, 

Scatter_none should do the update of the zero at the value specified by ignore_index which we read in, thus making the weight zero the final result will be zero when scaled out and any loss will be cancelled to zero. I've tried it

I've got this in a test case thats passing based on the input gold data I used and had ignore_index = 2, and I used weights that were different to ensure I'm getting the right weight zeroed out. The weight vector is always a 1d vector in the spec bound by classes so your initial assertion is correct. I didn't see an equivalent with take so I tried to handle it with this. The verify test for this I added is 

`softmaxcrossentropyloss_2d_no_reduction_uneven_weighted_ignore_index_test`

Here's the output up to the scatter with MIGRAPHX_TRACE_EVAL=2
```
[   RUN    ] softmaxcrossentropyloss_2d_no_reduction_uneven_weighted_ignore_index_test
Run instruction: @0 = @literal{0, 0, 0, 0} -> float_type, {4}, {1}, target_id=0
Time: 0.01746ms, 0.018961ms
Run instruction: @1 = @literal{2} -> int64_type, {1}, {0}, target_id=0
Time: 0.00136ms, 0.00146ms
Run instruction: 2 = @param:2 -> float_type, {4}, {1}, target_id=0
Time: 0.01504ms, 0.01515ms
Run instruction: 1 = @param:1 -> int32_type, {4}, {1}, target_id=0
Time: 0.00304ms, 0.00313ms
Run instruction: 0 = @param:0 -> float_type, {4, 4}, {4, 1}, target_id=0
Time: 0.00144ms, 0.00154ms
Run instruction: @5 = ref::convert[target_type=2](1) -> float_type, {4}, {1}, target_id=0
Time: 0.08942ms, 0.08957ms
Output has normal, zero
Output: 0, 3, 1, 2
Min value: 0, Max value: 3, Mean: 1.5, StdDev: 1.11803
Run instruction: @6 = ref::scatter_none[axis=0](2,@1,@0) -> float_type, {4}, {1}, target_id=0
Time: 0.09062ms, 0.0909ms
Output has zero, normal
Output: 1, 0.5, 0, 3
Min value: 0, Max value: 3, Mean: 1.125, StdDev: 1.1388
```

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Ah I can reorder this ok.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Ah Agreed. Looks like default is -1 thus for the 2d case this would always work

TedThemistokleous: Fixed

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Fixed

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,307 @@
...
```
umangyadav, line 299:
```suggestion
        if(reduction == "mean" and has_weights)
        {
             auto reduced_weights = info.add_instruction(
                    migraphx::make_op("reduce_sum", {{"axes", {0}}}), weight_tensor);
             loss_tensor =
                    info.add_instruction(migraphx::make_op("div"), loss_tensor, reduced_weights);
        }
        else if(reduction == "mean")
        {
         loss_tensor = info.add_instruction(migraphx::make_op("reduce_mean", {{"axes", {0}}}),
                                               loss_tensor);                                     
          } else if (reduction == "sum") {
          loss_tensor = info.add_instruction(migraphx::make_op("reduce_sum", {{"axes", {0}}}), loss_tensor);
          
}
```

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,307 @@
...
+        auto ignore_index =
+            info.add_literal(migraphx::literal(migraphx::shape(shape::int64_type, {1}, {0}), {-1}));
+        bool has_ignore_index = contains(info.attributes, "ignore_index");
+        if(has_ignore_index)
+        {
+            auto ignore_index_val =
+                parser.parse_value(info.attributes.at("ignore_index")).at<int64_t>();
+            ignore_index = info.add_literal(migraphx::literal(
+                migraphx::shape(migraphx::shape::int64_type, {1}, {0}), {ignore_index_val}));
+        }
```
umangyadav, line 158: Ignore index can be negative and also out of bound. We should check for negative and convert it to positive value. 

If it is out of bound then make `has_ignore_index = false`.

src/onnx/parse_softmaxcrossentropyloss.cpp
umangyadav: I see what you are trying to do  
It would be better to keep all weights/ignore index related calculation together in the code. 
Weights/ignore indices are not used untill later.

test/onnx/gen_onnx.py
```diff
@@ -9695,6 +9695,663 @@ def softmax_dyn_test():
...
+        ],
+        outputs=["3"],
+        reduction="sum",
+    )
+
+    return ([node], [scores, labels, weights], [loss])
+
+
+@onnx_test()
+def softmaxcrossentropyloss_kd_no_reduction_weighted_test():
```
umangyadav, line 10334: can you add tests

1.  where ignore_index is negative 
2. ignore index is out of bound 
3. matrix is 4D

TedThemistokleous: Yes I can do that

TedThemistokleous: Added three cases for the ignore out of bound  (positive and negative bounds) as well as the negative index.

Will add 4d matrix later this evening.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: This part is done. Sectioned this off as a function called normalize_input_index

test/onnx/gen_onnx.py
TedThemistokleous: Added 4d tests. Had these before but hitting issue with verify when I compare this to ref data. Working on solving these and the asymmetric input cases that I seem to be failing before I'll go back and fix the parser tests. Parser tests sill need an update as now ignore_index literal isn't added when it's not detected on input.

```
[==========] 25 tests ran
[  FAILED  ] 9 tests failed
[  FAILED  ] softmaxcrossentropyloss_kd_no_reduction_even_weighted_ones_test
[  FAILED  ] softmaxcrossentropyloss_kd_no_reduction_uneven_weighted_ones_test
[  FAILED  ] softmaxcrossentropyloss_kd_sum_reduction_weighted_test_ones
[  FAILED  ] softmaxcrossentropyloss_kd_sum_reduction_weighted_test_zeroes
[  FAILED  ] softmaxcrossentropyloss_kd_mean_reduction_weighted_test
[  FAILED  ] softmaxcrossentropyloss_kd_mean_reduction_uneven_weighted_test
[  FAILED  ] softmaxcrossentropyloss_2d_no_reduction_test_asym_test
[  FAILED  ] softmaxcrossentropyloss_2d_no_reduction_uneven_weighted_ones_asym_batch
[  FAILED  ] softmaxcrossentropyloss_2d_no_reduction_uneven_weighted_ones_asym_class
```

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Agreed. I need to use ignore index as soon as I read in a valid input score as that'll give me the max class size. I've modified this to add better handling of out of bounds ignore_index and cleaned it up a bit. I can move things around.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Changed things around

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Done

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Done

TedThemistokleous:
Rebasing this off master to restart work on this tomorrow

TedThemistokleous:
Need to fix parser tests but verify tests all pass and have been validated using the gold data. Gold data check leveraged Onnxruntime and numpy to ensure our answers are consistent for both in each case. 

Let me know if you see/want any additional coverage on verification or if there's things we want to clean up further.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,450 @@
...
+        size_t class_size = scores_shape.lens().at(1);
+        bool is_k_dim  = (ndims > 3);
+        bool sizes_mismatch = batch_size != class_size;
+
+        // ignore_index is optional attribute, assign this as a scalar literal input to the op
+        instruction_ref ignore_index;
+        auto has_ignore_index =
+            normalize_input_index(parser, info, static_cast<int64_t>(class_size), ignore_index);
+
+        bool has_weights = (args.size() > 2);
```
shivadbhavsar, line 385: I think this is problematic in the case where there is no weight input, but the ignore_index attribute is set. In that case you want your weights to be all 1 except at ignore_index, where it should be 0. 
L408 will not apply the 0 weight for ignore_index in this case.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,450 @@
...
+            if (is_k_dim)
+            {
+                std::iota(++perm.begin(), perm.end(), 2);
+                perm.at(class_size - 1) = 1;
+                scores = info.add_instruction(migraphx::make_op("transpose", {{"permutation", perm}}), scores);
+            }
+
+            scores = info.add_instruction(migraphx::make_op("gathernd"), scores, gathernd_indicies);
+
+            if (has_weights) // Saves us a multiply + gather op to gate this as default weights are literal 1's
```
shivadbhavsar, line 408: See above comment, and also you dont have to worry about this optimization in the parser. There is a matcher that looks for a multiplication with a literal of all 1s. That will take care of it.
I suggest getting rid of all the code branches that check for this to make the parser easier to follow.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,450 @@
...
+        auto scores = get_scores(args.at(0));
+        auto scores_shape = scores->get_shape();
+        auto labels  = get_labels(args.at(1), scores_shape);
+
+        // meta parameters based on input scores shape
+        size_t ndims = scores_shape.ndim();
+        size_t batch_size = scores_shape.lens().at(0);
+        size_t class_size = scores_shape.lens().at(1);
+        bool is_k_dim  = (ndims > 3);
+        bool sizes_mismatch = batch_size != class_size;
```
shivadbhavsar, line 378: why does this need to be handled seperately?

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,450 @@
...
+                weights = info.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", scores_shape.lens()}}), weights);
+                if (is_k_dim)
+                   weights = info.add_instruction(migraphx::make_op("transpose", {{"permutation", perm}}), weights);
+                weights = info.add_instruction(migraphx::make_op("gathernd"), weights, gathernd_indicies);
+            }
+        }
+
+        // Create output container which should be the same shape as input labels
+        auto loss_tensor = info.add_instruction(
+            migraphx::make_op("convert", {{"target_type", scores_shape.type()}}), labels);
```
shivadbhavsar, line 422:
```suggestion
        instruction_ref loss_tensor;
```
No sure why this converted labels instruction is needed, its never used.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,450 @@
...
+ 
+        // Do pointwise operators on the final set of indicies and scores we care about rather than
+        // before so that we're not doing a bunch of pointwise on items that aren't part of the loss calulation.
+        // Helps more in the mismatched Batch/Class and multi batch size cases. 
+        auto log_sm_scores  = info.add_instruction(migraphx::make_op("log"), scores);
+        auto neg_lsm_scores = info.add_instruction(migraphx::make_op("neg"), log_sm_scores);
+
+        if (is_k_dim or sizes_mismatch)
+            loss_tensor = handle_reduction(info, neg_lsm_scores, weights, reduction, has_ignore_index, has_weights);
+        else
```
shivadbhavsar, line 432: I dont see why handle_reduction cant just take care of this case too

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <migraphx/verify.hpp>
+#include <onnx_test.hpp>
+
+TEST_CASE(softmaxcrossentropyloss_2d_no_reduction_test_ones)
+{
+    migraphx::program p = migraphx::parse_onnx("softmaxcrossentropyloss_2d_no_reduction_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape score_shape{migraphx::shape::float_type, {4, 4}};
+    std::vector<float> score_data = {1, 1., 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
```
shivadbhavsar, line 35: can you add some tests with varying values in score_data? Having the same values kind of defeats the point of testing gather/scatter ops and you will see false positives. Iota might be a safe bet for this type of thing

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: This was so we didn't use the scatter_none case for mismatched batches and thus flowed into the gathernd logic

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: We still need the output container to be the same shape as the input labels. It was done so I'd create the shape needed. Cheap and dirty way to ensure I had something of the right size for scatter

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: It can, but scatter allowed us to do the assignment based on the input labels vs having to do a bunch of operators for the index selection

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
TedThemistokleous: Yes I can do that. That's easy since I added the script/scratchpad used for data here

TedThemistokleous: removing the scatter path. Its failing on a few of the entries still even though it should technically work for the 2d cases to save us the set of gatherNDs/Broadcasts/etc. I added a test case using std::iota using varying amounts and was seeing a large difference between input/output on the weighted case.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Removed this. Got overly complicated

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Removed this too

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: All good. removed

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Removed the logic so we always give back the default 1's like you mentioned and ignore index modifies this  once weights are added

shivadbhavsar (approved):
Looks good!
Just fix up the two failing tests. (I think the generated onnx file is the issue there?)

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        {
+            std::iota(++perm.begin(), perm.end(), 2);
+            perm.at(class_size - 1) = 1;
+            scores = info.add_instruction(migraphx::make_op("transpose", {{"permutation", perm}}), scores);
+        }
+
+        scores = info.add_instruction(migraphx::make_op("gathernd"), scores, gathernd_indicies);
+
+        std::vector<int64_t> axis_list(ndims-1, 0);
+        std::iota(++(axis_list.begin()), axis_list.end(), 2);
```
lakhinderwalia, line 395: (style, as above).

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        weights = info.add_instruction(migraphx::make_op("unsqueeze", {{"axes", axis_list}}), weights);
+        weights = info.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", scores_shape.lens()}}), weights);
+        if (is_k_dim)
+            weights = info.add_instruction(migraphx::make_op("transpose", {{"permutation", perm}}), weights);
+        weights = info.add_instruction(migraphx::make_op("gathernd"), weights, gathernd_indicies);
+
+        // Do pointwise operators on the final set of indicies and scores we care about rather than
```
lakhinderwalia, line 402: nit: indices

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        // before so that we're not doing a bunch of pointwise on items that aren't part of the loss calulation.
+        // Helps more in the mismatched Batch/Class and multi batch size cases.
```
lakhinderwalia, line 404: Can you please explain what these previous comments mean? 'Helps more', means what? Thanks.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        weights = info.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", scores_shape.lens()}}), weights);
+        if (is_k_dim)
+            weights = info.add_instruction(migraphx::make_op("transpose", {{"permutation", perm}}), weights);
+        weights = info.add_instruction(migraphx::make_op("gathernd"), weights, gathernd_indicies);
+
+        // Do pointwise operators on the final set of indicies and scores we care about rather than
+        // before so that we're not doing a bunch of pointwise on items that aren't part of the loss calulation.
+        // Helps more in the mismatched Batch/Class and multi batch size cases. 
+        auto log_sm_scores  = info.add_instruction(migraphx::make_op("log"), scores);
+        auto neg_lsm_scores = info.add_instruction(migraphx::make_op("neg"), log_sm_scores);
```
lakhinderwalia, line 406: Please document the formulae that is being used here in this operator, to calculate the cross entropy. Thanks.

test/onnx/softmaxcrossentropyloss_gold_data_check.py
```diff
@@ -0,0 +1,98 @@
...
+#               [[1., 1.], [1., 1.]], [[1., 1.], [1., 1.]]],
+#              [[[1., 1.], [1., 1.]], [[1., 1.], [1., 1.]],
+#               [[1., 1.], [1., 1.]], [[1., 1.], [1., 1.]]],
+#              [[[1., 1.], [1., 1.]], [[1., 1.], [1., 1.]],
+#               [[1., 1.], [1., 1.]], [[1., 1.], [1., 1.]]],
+#              [[[1., 1.], [1., 1.]], [[1., 1.], [1., 1.]],
+#               [[1., 1.], [1., 1.]], [[1., 1.], [1., 1.]]]],
+#             dtype=float)
+
+X = np.array([[[[0., 0.], [0., 0.]], [[0., 0.], [0., 0.]],
```
lakhinderwalia, line 25: Pardon me, I don't understand the value in setting X to all zeros. And then weights to all 1's.

test/onnx/verify/softmaxcrossentropyloss_kd_dim_weighted.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include <onnx_verify_utils.hpp>
+
+TEST_CASE(softmaxcrossentropyloss_kd_no_reduction_even_weighted_ones_test)
+{
+    migraphx::program p =
+        optimize_onnx("softmaxcrossentropyloss_kd_no_reduction_weighted_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape score_shape{migraphx::shape::float_type, {4, 4, 2, 2}};
+    std::vector<float> score_data(score_shape.elements(), 1.0f);
```
lakhinderwalia, line 37: All score data shouldn't be 1s. There is no point in testing this degenerative case.

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <migraphx/verify.hpp>
+#include <onnx_test.hpp>
+
+TEST_CASE(softmaxcrossentropyloss_2d_no_reduction_test_ones)
+{
+    migraphx::program p = optimize_onnx("softmaxcrossentropyloss_2d_no_reduction_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape score_shape{migraphx::shape::float_type, {4, 4}};
+    std::vector<float> score_data(score_shape.elements(), 1);
```
lakhinderwalia, line 35: Please use some randomized data.

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
```diff
@@ -0,0 +1,185 @@
...
+    EXPECT(migraphx::verify::verify_rms_range(result_vector, gold));
+}
+
+TEST_CASE(softmaxcrossentropyloss_2d_no_reduction_test_zeros)
+{
+    migraphx::program p = optimize_onnx("softmaxcrossentropyloss_2d_no_reduction_test.onnx");
+    p.compile(migraphx::make_target("ref"));
+
+    migraphx::shape score_shape{migraphx::shape::float_type, {4, 4}};
+    std::vector<float> score_data(score_shape.elements(), 0);
```
lakhinderwalia, line 79: This test has very limited value, IMO. It is same as testing everything with the value of `1`. Unless you could perhaps explain why it is uniquely valuable to test this value..! Thanks.

lakhinderwalia (commented):
Is there any verify test which compares ref value with the equivalent gpu based output? Thanks.

src/onnx/parse_softmaxcrossentropyloss.cpp
lakhinderwalia: If `ignore_index` is out of bound, that doesn't mean `has_ignore_index = false`. One can have such index present `labels`. As specified in the spec:
> l[i][d1][d2]...[dk]  = 0, when labels[n][d1][d2]...[dk] = ignore_index

test/onnx/gen_onnx.py
lakhinderwalia: Please note that an `ignore_index` that is negative is actually the same case as out_of_bounds.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        {
+            auto ignore_index_val =
+                parser.parse_value(info.attributes.at("ignore_index")).at<int64_t>();
+
+            if(ignore_index_val >= input_classes || abs(ignore_index_val) > input_classes)
+            {
+                return false;
+            }
+
+            if(ignore_index_val < 0)
```
lakhinderwalia, line 150: This shouldn't be done. It should be considered as out of bounds, when the index value is not in the range `[0, C)`.

TedThemistokleous:
> Is there any verify test which compares ref value with the equivalent gpu based output? Thanks.

Not needed here as this is an onnx function which is made up of the simpler onnx operator pieces which themselves have been validated between ref and gpu. The comparison I've done is use gold data from Onnxruntime CPU and numpy. This is similar to how we validate output in our driver and are used as part of the verify tests

test/onnx/softmaxcrossentropyloss_gold_data_check.py
```diff
@@ -0,0 +1,98 @@
+import numpy as np
+from onnx.reference import ReferenceEvaluator
+
+#X = np.array([[1.0,2.0,3.,4.], [2.,4.,5.,7.], [11.,13.,23.,17.], [2.,1.,31.,37.]], dtype=float)
```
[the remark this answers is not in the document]
TedThemistokleous: Don't disagree but I may remove this entire file as its never used by anything and was run by hand to aquire and compare output stages vs MIGRAPHX_TRACE_EVAL=3 when ensuring verify tests were 1:1 with Onnxruntime and Numpy for this operator

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+                               const int64_t input_classes,
+                               instruction_ref& ignore_index) const
+    {
+        bool has_ignore_index = contains(info.attributes, "ignore_index");
+        if(has_ignore_index)
+        {
+            auto ignore_index_val =
+                parser.parse_value(info.attributes.at("ignore_index")).at<int64_t>();
+
+            if(ignore_index_val >= input_classes || abs(ignore_index_val) > input_classes)
```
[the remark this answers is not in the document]
TedThemistokleous: Ah I think you're right about this. 

_ Labels element value shall be in range of [0, C). If ignore_index is specified, it may have a value outside [0, C) and the label values should either be in the range [0, C) or have the value ignore_index._

I was assuming this handles the negative case as ignore_index itself specified its an int but no range on what's valid.

lakhinderwalia:
> > Is there any verify test which compares ref value with the equivalent gpu based output? Thanks.
> 
> Not needed here as this is an onnx function which is made up of the simpler onnx operator pieces which themselves have been validated between ref and gpu. The comparison I've done is use gold data from Onnxruntime CPU and numpy. This is similar to how we validate output in our driver and are used as part of the verify tests

There are composite-layered verify tests e.g. `test_transpose_reshape_add_sub_mul`, `test_var_sl_gru_bidirect`  that I just randomly point to. I think verification is important, and worthwhile to add here as well.  Thanks.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+            weights = args.at(2);
+            auto weights_shape = weights->get_shape();
+
+            if(weights_shape.lens()[0] != scores_shape.lens()[1])
+            {
+                MIGRAPHX_THROW("softmaxcrossentropyloss: Invalid weight vector shape. Weight must "
+                               "contain weight for each class");
+            }
+
+            if(migraphx::shape::is_integral(weights_shape.type()))
```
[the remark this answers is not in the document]
TedThemistokleous: Its added more for verbosity incase an input weights are set to the wrong type and a user tries to parse them in say our driver, or a script. We'll catch that during parse.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        if(label_shape.type() != migraphx::shape::int32_type and label_shape.type() != migraphx::shape::int64_type)
+        {
+            MIGRAPHX_THROW(
+                "softmaxcrossentropyloss: Labels must either be int32 or int64 types");
+        }
+
+        if(scores_shape.lens()[0] != label_shape.lens()[0])
+        {
+            MIGRAPHX_THROW(
+                "softmaxcrossentropyloss: Score and Labels must identical batch size inputs");
```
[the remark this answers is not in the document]
TedThemistokleous: Changed

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Good point. I'll add that check and associated tests

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        auto labels_unsq = info.add_instruction(migraphx::make_op("unsqueeze", {{"axes", {-1}}}), labels);
+        auto label_shape = labels->get_shape();
+        auto labels_rank = labels_unsq->get_shape().ndim();
+
+        std::vector<instruction_ref> coordinate_index_literals;
+
+        for (size_t axis = 0; axis < (labels_rank - 1); axis++)
+        {
+            // Trying to replicate torch arrange() here.
+            auto len_val = labels_unsq->get_shape().lens().at(axis);
```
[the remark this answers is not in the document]
TedThemistokleous: Valid nit. I'll change this

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -0,0 +1,420 @@
...
+        scores =
+            info.add_instruction(migraphx::make_op("softmax", {{"axis", 1}}), scores);
+
+        // Index selection before loss calculation completed
+        auto gathernd_indicies = handle_index_selection(info, labels);
+
+        std::vector<int64_t> perm(class_size, 0);
+        if (is_k_dim)
+        {
+            std::iota(++perm.begin(), perm.end(), 2);
```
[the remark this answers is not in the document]
TedThemistokleous: Sure I can change this. I suppose pre increment can be easily missed.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Sure. I'll add as comment

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
TedThemistokleous: This was created as I took a TDD approach to this operator. 

Valuable to test as for softmax giving us another sane known result as the input of zero is that it becomes an 1/N 
as softmax =  [ exp(0) / sum (exp(0)], and N is size of the input vector. With all 1's  you also get  each entry being (e/N(e)) which also simplifies to 1/N for each value. 

This ensures that these two mathematical cases are logically consistent and the operator has no issues with the softmax stage before any complex indexing (2 dimensions or above). I found it useful when using this and checking with MIGRAPHX_TRACE_EVAL=3 to ensure all outputs are updated consistently.

Although it appears trivial its another known point that can be checked by hand and ensure that all numbers will be populated correctly output and give the same correct value. If I used random data only, and got say a zero on output, it wouldn't be clear that a zero is due to the operator putting the right output, or an error with indexing if say that output wasn't populated.

The next test this goes with is with input values that are different, basically an std::iota(0, N) to ensure that each item is indexed to the correct position in the output vector. If that failed, I could use this test and the ones case to ensure with mathematically certainty that it isn't some error numerically that I was seeing.

test/onnx/verify/softmaxcrossentropyloss_kd_dim_weighted.cpp
TedThemistokleous: There is, as it ensures we're getting a valid output from softmax of a known mathematical value. This tests the initial input and goes with the zero case inputs as well

test/onnx/softmaxcrossentropyloss_gold_data_check.py
TedThemistokleous: Tests a known value, see later comment on the zeroes tests I did

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: The odd part about this is that when I test for gold values in numpy and Onnxruntime they both give me consistent and valid outputs when I read in the model with the same inputs/in bound ignore index if I take the absolute value

TedThemistokleous: Nevermind. went over the test. You are correct. Ive changed this to be consistent. Looks like there was a  bad value in the test ( zero) where there should be an entry. numpy and onnxruntime won't error out ,but infact just flat out silently ignore the negative input and operate as if ignore index doesn't exist.

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
lakhinderwalia: If you want to test special cases like `0` or `1`, then please  just set `score_data` elements to `[[0,0,0,0],[1,1,1,1][0.5,0.5,0.5,0.5],[x,x,x,x]]`, as an example. And you can verify all kind of combinations in just one test. Thanks.

lakhinderwalia: What I would recommend is a test beyond the all `1` or `0` case. Please test a test vector like `[[1,0,0,0], [0,1,0,0],[0,0,1,0],[0,0,0,1]]` whose entropy loss is calculated against labels like `[1,1,1,1]`. Thanks.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Added the case and updated tests accordingly. Added another onnx file too

src/onnx/parse_softmaxcrossentropyloss.cpp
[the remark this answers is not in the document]
TedThemistokleous: Changed

src/onnx/parse_softmaxcrossentropyloss.cpp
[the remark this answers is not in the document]
TedThemistokleous: Changed

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Removed this comment

test/onnx/gen_onnx.py
TedThemistokleous: Solved

src/onnx/parse_softmaxcrossentropyloss.cpp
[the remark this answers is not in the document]
TedThemistokleous: Fixed this

test/onnx/softmaxcrossentropyloss_gold_data_check.py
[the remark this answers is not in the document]
TedThemistokleous: Removed this as part of the PR

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Done

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
lakhinderwalia: > What I would recommend is a test beyond the all `1` or `0` case. Please test a test vector like `[[1,0,0,0], [0,1,0,0],[0,0,1,0],[0,0,0,1]]` whose entropy loss is calculated against labels like `[1,1,1,1]`. Thanks.

Please note, this is a one-hot test sample. It isn't the same as some new tests that you have added -- where first four indices are of one value, next 4 of another value, and last 8 values were filled with iota.

src/onnx/parse_softmaxcrossentropyloss.cpp
```diff
@@ -142,16 +142,11 @@ struct parse_softmaxcrossentropyloss : op_parser<parse_softmaxcrossentropyloss>
             auto ignore_index_val =
                 parser.parse_value(info.attributes.at("ignore_index")).at<int64_t>();
 
-            if(ignore_index_val >= input_classes or abs(ignore_index_val) > input_classes)
+            if(ignore_index_val >= input_classes or ignore_index_val < 0)
```
lakhinderwalia, line 145: Pardon me, this thing is likely falling through the cracks. I meant that any index value that is negative is to be considered like the out of bounds. But that doesn't imply that we should ignore it. These `ignore_index` OOB values are still to be processed/compared when looking at `l[i][d1][d2]..[dk]`. As per this specification:

> l[i][d1][d2]...[dk]  = 0, when labels[n][d1][d2]...[dk] = ignore_index

test/onnx/verify/softmaxcrossentropyloss_weighted.cpp
```diff
@@ -304,7 +304,7 @@ TEST_CASE(softmaxcrossentropyloss_2d_no_reduction_uneven_weighted_neg_ignore_ind
     auto result = p.eval(pp).back();
     std::vector<float> result_vector;
     result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
```
lakhinderwalia, line 393: I don't understand, why was this test passing earlier!?

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
TedThemistokleous: Done

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
TedThemistokleous: Oh you wanted a one hot case. Okay. I added the previous of 0000, 1111, a,b,c,d ,e,f,g,h,i
Let me add that case as well

TedThemistokleous: Test is added and called 

softmaxcrossentropyloss_2d_no_reduction_one_hot_test

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: Yup. Handled this. 

What occurs is the ignore index is added to the weights but if its out of bounds, it doesn't make it as part of the weight vector, which is always multiplied at the end once softmax + log + neg is applied to the selected score data. 

If ignore index is out of bounds, then we never update our weight vector with a zero to cancel out the score value before its set for the reduction step

test/onnx/verify/softmaxcrossentropyloss_no_attr.cpp
lakhinderwalia: Thanks.

src/onnx/parse_softmaxcrossentropyloss.cpp
TedThemistokleous: I think this is a bit more complicated than realized. Likely will need to modify parser again since we can't read label on parse but have the ignore index of the operator. Will need to add some additional masking.

TedThemistokleous: Solved. Get equivalent output with using a bitmask and where as we can use the input index and check with equals/where after the loss is calculated. 

Verified the results with Onnxruntime and used that as the gold data in the added verify tests.

TedThemistokleous:
> > > Is there any verify test which compares ref value with the equivalent gpu based output? Thanks.
> > 
> > 
> > Not needed here as this is an onnx function which is made up of the simpler onnx operator pieces which themselves have been validated between ref and gpu. The comparison I've done is use gold data from Onnxruntime CPU and numpy. This is similar to how we validate output in our driver and are used as part of the verify tests
> 
> There are composite-layered verify tests e.g. `test_transpose_reshape_add_sub_mul`, `test_var_sl_gru_bidirect` that I just randomly point to. I think verification is important, and worthwhile to add here as well. Thanks.



> > > Is there any verify test which compares ref value with the equivalent gpu based output? Thanks.
> > 
> > 
> > Not needed here as this is an onnx function which is made up of the simpler onnx operator pieces which themselves have been validated between ref and gpu. The comparison I've done is use gold data from Onnxruntime CPU and numpy. This is similar to how we validate output in our driver and are used as part of the verify tests
> 
> There are composite-layered verify tests e.g. `test_transpose_reshape_add_sub_mul`, `test_var_sl_gru_bidirect` that I just randomly point to. I think verification is important, and worthwhile to add here as well. Thanks.

@lakhinderwalia Added some of the composite tests in for k dimension and 2d cases (Batch, class size,...) & (Batch, class_size)
May need to limit labels to a literal in this case  instead of parameter since I can generate random literals in a certain range as labels have to be within [0, Class_size) . Let me know if you want me to paramaterize these further/ template them also based on reduction type. Will see in the morning.

For the k dimensional case I'd need to sort out the intermediate dimensions/gather calls for larger dimensions since we need to loop over and adjust the indices before the concat. Not hard but wanted to mirror the tests we do already as a baseline verify onnx tests

lakhinderwalia:
> @lakhinderwalia Added some of the composite tests in for k dimension and 2d cases (Batch, class size,...) & (Batch, class_size) May need to limit labels to a literal in this case instead of parameter since I can generate random literals in a certain range as labels have to be within [0, Class_size) . Let me know if you want me to paramaterize these further/ template them also based on reduction type. Will see in the morning.
> 

Thanks, @TedThemistokleous. I was suggesting very simple tests. Take a test graph with an an onnx operator "softmaxcrossentropy" -- no need to expose its entrails. And then just give it a sample data. Pass it through `ref` and `gpu`, and then just compare those results -- instead of a comparison to the (pre-calculated) golden data. Templated tests are just fine, for how you test is programmatically isn't the focus here.

lakhinderwalia (approved):
Thanks for following through my comments, @TedThemistokleous, and adding the new verify tests. They can be simplified, optionally, and I am fine with the new operator :-)

TedThemistokleous:
Got all the end to end stuff working, seeing some odd behavior with the dnn1 based tests for double/float. Looks like their version of log doesn't support double which causes issues, along with  seeing a small difference between mean/sum k -dimension cases which incur enough error to trigger a tolerance error.

PR #3010, opened 2024-04-26
Fuse inputs with mlir

pfultz2:
This will fuse the inputs but only when using the `MIGRAPHX_ENABLE_MLIR_INPUT_FUSION` env variable.


src/module.cpp
```diff
@@ -979,6 +979,61 @@ std::array<module::with_inputs, 3> module::split(const std::vector<instruction_r
...
+                continue;
+            inputs.push_back(input);
+        }
+    }
+    insert_params(*this, inputs, *map_ins);
+    return this->add_instructions(inss, map_ins, std::move(insert));
+}
+
+std::vector<instruction_ref>
+module::fuse(const module& m,
```
bpickrel, line 1020: This function needs a descriptive comment.  Is it an exact replacement for `insert_ins_in_submodule` or is it more general?  And why was it created?  Looks like it's a more widely usable generalization of things that were being done on a case by case basis--is that correct?

src/param_utils.cpp
```diff
@@ -43,5 +46,31 @@ void sort_params(std::vector<instruction_ref>& params)
               }));
 }
 
+std::vector<instruction_ref>
+find_inputs(const std::unordered_map<instruction_ref, instruction_ref>& map_ins,
```
bpickrel, line 50: This function needs a descriptive comment or no one else will ever be able to use it.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -556,6 +558,36 @@ struct find_mlir_attention_fused_ops : public find_mlir_standalone_attention_op
     }
 };
 
+struct find_pointwise_mlir
```
bpickrel, line 561: Will this fusion have the same problems with input ranks or types not matching that I've been seeing in `fuse_reduce.cpp`?  It's not clear to me just what would happen, but I don't see any checks on the compatibility of the instructions being fused.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -591,6 +623,9 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
         mpm,
         find_mlir_standalone_convolution_op{get_mode("convolution", mlir_mode::fast)},
         find_mlir_standalone_dot_op{get_mode("dot", mlir_mode::fast)});
+
+    if(enabled(MIGRAPHX_ENABLE_MLIR_INPUT_FUSION{}))
+        match::find_matches(mpm, find_pointwise_mlir{});
```
krzysz00, line 628: 1. At a minimum, we need to be making sure all the input ops are `is_pointwise_op_supported_by_mlir`
2. More generally, I'd ask that we have a separate filter function for what pointwise functions can be fused on the input, thus allowing us to filter off input fusions that aren't working

krzysz00: 3. The pointwise operation could be separated by a chain of views (reshape/transpose/...) - we should fuse those in too!
4. ... heck, we should do the same thing for non convolution/dot arguments (ex. the bias tensor), which we currently don't do

krzysz00 (commented):
Agreed with Brian about the need for comments and the need for filtering.

(I'm not going to block this because this isn't my neam, but)

src/targets/gpu/fuse_mlir.cpp
pfultz2: > The pointwise operation could be separated by a chain of views (reshape/transpose/...) - we should fuse those in too!

We already fuse those on the inputs. This will fuse additionally on top of that.

src/targets/gpu/fuse_mlir.cpp
pfultz2: No it wont.

src/module.cpp
pfultz2: > Is it an exact replacement for insert_ins_in_submodule?

Yes.

> And why was it created?

So we can reuse it across other passes.

src/targets/gpu/fuse_mlir.cpp
pfultz2: Added a seperate function for input fusion support.

src/include/migraphx/module.hpp
```diff
@@ -245,6 +245,21 @@ struct MIGRAPHX_EXPORT module
                                      const std::vector<instruction_ref>& splits1,
                                      const std::vector<instruction_ref>& splits2) const;
 
+    // Fuse the instruction into the module by inserting the instructions and
```
krzysz00, line 248: Is it obvious to people who know the codebase what the inputs and outputs of these methods are?

pfultz2:
So there is definitely a perf degradation fusing the inputs. I see a perf difference of this:

```
gpu::code_object::mlir_convert_mul_convert_sub_add_rsqrt_sub_mul_mul_add_sigmoid_mul_reshape_convolution_convert: 0.826661ms / 1 = 0.826661ms, 84%
```

vs.

```
gpu::code_object::convert_mul_convert_sub_add_rsqrt_sub_mul_mul_add_sigmoid_mul_kernel: 0.0360864ms / 1 = 0.0360864ms, 8%
gpu::code_object::mlir_reshape_convolution_convert: 0.293353ms / 1 = 0.293353ms, 61%
```

src/fuse_reduce.cpp
```diff
@@ -426,14 +348,14 @@ struct reduce_reshape : rewrite_reshapes_base
...
         auto* sm   = mpm.create_module(oldm->name() + "_reshape");
-        insert_module_in_submodule(sm, inputs, oldm, transform_op([&](const operation& sop) {
-                                       if(contains(sop.name(), "reduce"))
-                                           return make_op(sop.name(), {{"axes", axes}});
-                                       if(sop.name() == "multibroadcast")
-                                           return make_op("multibroadcast", {{"out_lens", dims}});
-                                       assert(sop.name() == "pointwise");
-                                       return sop;
-                                   }));
+        sm->fuse(*oldm, inputs, nullptr, transform_op([&](const operation& sop) {
```
bpickrel, line 351:
```suggestion
        sm->set_bypass();
        sm->fuse(*oldm, inputs, nullptr, transform_op([&](const operation& sop) {
```

src/module.cpp
```diff
@@ -979,6 +979,61 @@ std::array<module::with_inputs, 3> module::split(const std::vector<instruction_r
...
+    auto n = m.get_parameter_shapes().size();
+    for(auto input : inputs)
+    {
+        if(contains(map_ins, input))
+            continue;
+        map_ins[input] = m.add_parameter(param_name(n++), input->get_shape().as_standard());
+    }
+}
+
+std::vector<instruction_ref>
```
bpickrel, line 995:
```suggestion
// General utility for fusing modules.  Inserts instructions into an instruction list.

std::vector<instruction_ref>
```

bpickrel (approved):
I asked for more comments and suggested a change that should simplify merge for my latest PR--otherwise approved.

krzysz00:
Re the perf degradation
1. We've got a known issue with how input fusions are codegen'd
2. Because matrix elements are inherently reused during a gemm - and even more so during an implicit gemm - an input fusion fundamentally costs more floating-point operations than an output fusion. This is why I said that the input fusions need to be limited to, say, type conversions and dequantizations.

That is, while some of that perf loss is on us, some of it is algorithmically inherent to the idea of the input fusion, so I think this PR should be amended to restrict input fusions to cases where they won't be too expensive.

src/module.cpp
pfultz2: There is already a description of the function in the header file.

bpickrel: In `module.hpp`?  `insert_params` isn't even mentioned in that header.

pfultz2: I thought it was for the `fuse` method. Let me add a description for that.

pfultz2:
> That is, while some of that perf loss is on us, some of it is algorithmically inherent to the idea of the input fusion, so I think this PR should be amended to restrict input fusions to cases where they won't be too expensive.

So this isn't enable by default, there is a flag to enable it, so we can test this and try to find when its beneficial or not. This can be solved after this PR is merged(Most likely because I wont be able to address this before leaving on paternity leave).

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -40,6 +42,7 @@ struct module;
 namespace gpu {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_EXTRA_MLIR);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_MLIR_INPUT_FUSION);
```
umangyadav, line 45: Add documentation for this new env var

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -397,14 +400,27 @@ bool is_pointwise_op_supported_by_mlir(const instruction& i)
     return false;
 }
 
+// A seprate function so we can remove operators that are supported by mlir
```
umangyadav, line 403:
```suggestion
// A separate function so we can remove operators that are supported by mlir
```

lpaoletti (approved):

src/include/migraphx/module.hpp
```diff
@@ -245,6 +245,21 @@ struct MIGRAPHX_EXPORT module
...
+    // parameters for any missing inputs.
+    std::vector<instruction_ref>
```
umangyadav, line 250: Need some unit-tests

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -579,6 +595,46 @@ struct find_mlir_attention_fused_ops : public find_mlir_standalone_attention_op
     }
 };
 
+struct find_pointwise_mlir
```
umangyadav, line 598: Need unit-tests

src/targets/gpu/fuse_mlir.cpp
umangyadav: > "pull in the views" logic lives up in fuse_pointwise. 

I don't think we do this at the moment. @pfultz2  looks like it is something we can do on top of fusing pointwise inputs

umangyadav: I mean currently it would fuse `pointwise + reshapes + gemm_based_ins` . We can do `reshapes + pointwise + reshapes + gemm_based_ins.`

src/param_utils.cpp
```diff
@@ -49,5 +52,31 @@ void sort_params(std::vector<instruction_ref>& params)
...
+        if(parent != nullptr and not parent->has_instruction(input))
+            continue;
+        auto v      = param->get_operator().to_value();
+        auto name   = v.at("parameter").to<std::string>();
+        names[name] = input;
+    }
+    std::transform(names.begin(), names.end(), std::back_inserter(result), [](const auto& p) {
+        return p.second;
+    });
+    assert(not sub or result.size() == sub->get_parameter_shapes().size());
```
umangyadav, line 77: Is it using `map` so that instructions comes out in sorted order based on `name` ?

umangyadav, line 77: If `sub == nullptr` you can just do early return

umangyadav (commented):
Looks good overall but need to add unit & verify tests. I am not sure how to do verify tests with ENV flag though.

pfultz2:
> I am not sure how to do verify tests with ENV flag though.

We can add the ENV var to MLIR jenkins job. We already do this to enable MLIR for everything.

src/param_utils.cpp
pfultz2: > Is it using `map` so that instructions comes out in sorted order based on `name` ?

Yes.

pfultz2: > If `sub == nullptr` you can just do early return

Early return where?

src/targets/gpu/fuse_mlir.cpp
pfultz2: > I don't think we do this at the moment. @pfultz2 looks like it is something we can do on top of fusing pointwise inputs

Yes, we should open an issue to handle these cases.

pfultz2: > I mean currently it would fuse pointwise + reshapes + gemm_based_ins . We can do reshapes + pointwise + reshapes + gemm_based_ins.

We could also handle `gemm_based_ins + reshapes + pointwise`, which we do see such patterns in models.

umangyadav: #3211  Opened an issue.

umangyadav (approved):

src/param_utils.cpp
umangyadav: just when it starts the body of `find_inputs()`

PR #3011, opened 2024-04-27
Onehot dyn and refactor

CharlieL7:
* Creates a ref onehot operator that handles dynamic shapes
* New simplify matcher that replaces the new ref onehot operator with other ops and scatter
* Refactored to use scatter rather than gather for this op
* Added new mode to scatter on ref and gpu to skip out of bounds indices


src/include/migraphx/op/scatter_op.hpp
```diff
@@ -67,38 +72,60 @@ struct scatter_op : op_name<Derived>
...
+                if(skip_out_of_bounds)
+                {
+                    if(not output_shape.check_within_bounds(out_idx))
+                    {
+                        return;
+                    }
+                }
+                else
+                {
+                    assert(output_shape.check_within_bounds(out_idx));
```
TedThemistokleous, line 111: Put this assert before the if(skip_out_of_bounds) for release it shouldn't do anything, wheras will trigger on debug, and allow you to not need the else case which would be empty

src/include/migraphx/op/scatter_op.hpp
```diff
@@ -67,38 +72,60 @@ struct scatter_op : op_name<Derived>
...
+
+                // this addition doesn't necessarily make index positive if index was out of bounds
+                index         = (index < 0) ? index + axis_dim_size : index;
+                if(skip_out_of_bounds and index < 0)
+                {
+                    return;
+                }
+                else
+                {
+                    assert(index >= 0);
```
TedThemistokleous, line 98: move this before the if statement and remove the else case

src/targets/gpu/kernels/include/migraphx/kernels/scatter.hpp
```diff
@@ -30,19 +30,37 @@
...
         auto out_idx  = indices_shape.multi(i);
         auto index    = indices_t[i];
         index         = index < 0 ? index + axis_dim_size : index;
+        if constexpr(SkipOutOfBounds)
+        {
+            if(index < 0)
+            {
+                return;
+            }
+        }
```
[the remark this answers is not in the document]
CharlieL7: it's because of constexpr. The value of `index` is not known at compile time, but `SkipOutOfBounds` is known. The compiler will create two kernels one with this if and one without.

src/include/migraphx/op/scatter_op.hpp
CharlieL7: Different behavior, we don't want to assert fail for this case when `skip_out_of_bounds` is enabled. The assert as an else is mainly for debugging purposes. If there is a better way to do this let me know

src/targets/gpu/kernels/include/migraphx/kernels/scatter.hpp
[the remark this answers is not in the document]
TedThemistokleous: ah got it got it.

src/include/migraphx/op/scatter_op.hpp
TedThemistokleous: I think `assert((index >=0) xor (skip_out_of_bounds == True))` works?

src/include/migraphx/op/scatter_op.hpp
TedThemistokleous: Similar to above, use the skip_out_of_bounds_blag

`assert(skip_out_of_bounds or output_shape.check_within_bounds(out_idx));` 

That eliminates the else branch

src/simplify_dyn_ops.cpp
```diff
@@ -450,6 +451,83 @@ struct find_static_broadcast_for_dot
...
+        auto onehot_ins           = mr.result;
+        auto onehot_inputs        = onehot_ins->inputs();
+        auto onehot_op            = any_cast<op::onehot>(onehot_ins->get_operator());
+        auto indices_ins          = onehot_inputs[0];
+        shape indices_shape = indices_ins->get_shape();
+        auto depth_ins            = onehot_inputs[1];
+        auto values_ins           = onehot_inputs[2];
+        shape values_shape        = values_ins->get_shape();
+        std::vector<std::size_t> static_output_lens = indices_shape.lens();
+        auto normalized_axis =
```
lakhinderwalia, line 491: How does this handle axis = -1, the default, please. Also, the legit range of axis is [-r-1, r], r being the rank. Thanks.

src/simplify_dyn_ops.cpp
```diff
@@ -450,6 +451,83 @@ struct find_static_broadcast_for_dot
...
+            (onehot_op.axis < 0) ? onehot_op.axis + indices_shape.ndim() + 1 : onehot_op.axis;
+        depth_ins->eval().visit([&](auto d) {
+            static_output_lens.insert(static_output_lens.begin() + normalized_axis, d[0]);
```
lakhinderwalia, line 494: d[0] has to be casted to int64 for non-integers. Thanks.

TedThemistokleous (changes requested):
Few comments. Just change those If branches for now.

src/shape.cpp
```diff
@@ -420,6 +420,20 @@ void shape::multi_copy(std::size_t idx, std::size_t* start, const std::size_t* e
     *start = tidx;
 }
 
+bool shape::check_within_bounds(std::vector<std::size_t> multi) const
```
pfultz2, line 423: I would just name this `within_bounds`.

src/shape.cpp
```diff
@@ -420,6 +420,20 @@ void shape::multi_copy(std::size_t idx, std::size_t* start, const std::size_t* e
...
+{
+    assert(this->lens().size() == multi.size());
+    bool ret = true;
+    for(int i = 0; i < multi.size(); ++i)
+    {
+        if(multi.at(i) > this->lens().at(i))
+        {
+            ret = false;
+        }
+    }
```
pfultz2, line 433: Use `std::equal` here with `std::less<>{}` for the predicate.

src/shape.cpp
```diff
@@ -420,6 +420,20 @@ void shape::multi_copy(std::size_t idx, std::size_t* start, const std::size_t* e
     *start = tidx;
 }
 
+bool shape::check_within_bounds(std::vector<std::size_t> multi) const
+{
+    assert(this->lens().size() == multi.size());
+    bool ret = true;
+    for(int i = 0; i < multi.size(); ++i)
+    {
+        if(multi.at(i) > this->lens().at(i))
```
pfultz2, line 429: This is a bug, if the index is equal to the lens it will be out of bounds. I think switching to using `std::less<>{}` with `std::equal` will fix this issue.

src/targets/gpu/kernels/include/migraphx/kernels/scatter.hpp
```diff
@@ -30,19 +30,37 @@
...
         out_idx[Axis] = index;
-
+        if constexpr(SkipOutOfBounds)
+        {
+            for(int k = 0; k < out_idx.size(); ++k)
+            {
+                if(out_idx[k] > output_shape.lens[k]){
+                    return;
+                }
+            }
```
pfultz2, line 62: Add `equal` algorithm to the algorithm.hpp header:

```cpp
template <class Iterator1, class Iterator2, class Predcate>
constexpr bool equal(Iterator1 first1, Iterator1 last1, Iterator2 first2,
                     Predcate p)
{
  for (; first1 != last1; ++first1, ++first2)
    if (not p(*first1, *first2))
      return false;
  return true;
}
```

And then use that here.

src/include/migraphx/op/scatter_op.hpp
pfultz2, line 98: Should we throw here?

CharlieL7: That's a question I had. What's our intent between an assert vs. a throw? The meaning I get is that we assert when it's a code error vs. a throw meaning a user or input error.

src/shape.cpp
lakhinderwalia: This API implied to me that two shapes were being considered. If instead it is just an indices vector to compare to, the name (of the API) could be reworded. Thanks.

src/include/migraphx/op/scatter_op.hpp
TedThemistokleous: I think this should be an assert. this isn't in the parser but in the operator itself

src/shape.cpp
CharlieL7: Renamed to multi_within_bounds to address your and Lakinder's comments.

src/shape.cpp
CharlieL7: Renamed to multi_within_bounds to address Paul and Lakhinder's comments.

src/simplify_dyn_ops.cpp
CharlieL7: Should be fine with implicit casting to `std::size_t`.

src/simplify_dyn_ops.cpp
CharlieL7: Adds `rank(indices) + 1` if negative. Handling the range of supported indices in the `scatter` operation.

lakhinderwalia (approved):

src/shape.cpp
pfultz2: For doing a binary `all_of` algorithm, it is named `std::equal` in the standard c++ library. The name `compare` would have been bad since such a comparison is not totally ordered. The name `std::equal` may not be the best name either since its not symmetrical(and the binary predicate is not required to be symmetrical), however that is the name that C++ has chosen for that algorithm(most likely because the common use case of a binary `all_of` is to do equality comparisons). I dont think we should avoid C++ components because they chose a bad or confusing name(ie `vector`, `apply`, etc).

Furthermore, understanding that `std::equal` is a binary `all_of`(C++ developers should be familiar with the standard algorithms), I do find `std::equal` to be clearer here(its also whats suggested [here](https://stackoverflow.com/a/74122451/375343) and its cleaner than using `std::mismatch` or a raw loop).

TedThemistokleous (approved):

Issue #2994, opened 2024-04-24
Fix bug in generic split Module


umangyadav:
Issue came up when trying to unfuse MLIR and pointwise. 
It need to correctly map Split instructions inputs to the second module inputs.

PR #3016, opened 2024-04-29
Fix input args order in split module functionality

umangyadav:
Current logic:

1. Current logic is setting input argument order for the second module based on parameter order in input module. 
2. Next it is creating new parameter by traversing the input module, which doesn't gurantee traversing parameters in same order as input arguments order selected in step 1. 

For example, look the unit-test in this PR.  With current logic `add_1` instruction would do `add(sx4, dot_ins)` instead of original operation of `add(sx1, dot_ins)`. 

Solves #2994

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -922,8 +922,7 @@ generic_split(const module& m,
         instructions2.push_back(ins);
     }
 
-    std::vector<instruction_ref> inputs2 = select_params(instructions2, param_map);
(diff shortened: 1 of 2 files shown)


TedThemistokleous (approved):

pfultz2 (approved):

umangyadav:
@TedThemistokleous  This wasn't a simple bug to figure out 😥 . But bugfix changes are simple enough.

Issue #2898, opened 2024-03-18
Split pointwise from mlir op when splitk is being used


pfultz2:
When compiling for gemm using splitk for MLIR, split the pointwise operators and generate code object using our regular pointwise code generation.

PR #3022, opened 2024-04-30
Add logic to unfuse mlir and pointwise module for the split-k GEMM configs

umangyadav:
Solves #2898


src/targets/gpu/compile_ops.cpp
```diff
@@ -228,10 +228,12 @@ struct compile_plan
...
     }
 
-    void replace(module& m) const
+    instruction_ref
+    replace(module& m,
+            const std::unordered_map<instruction_ref, instruction_ref>& inputs_rep_map) const
     {
         const auto& cr = benchmark();
-        cr.replace.replace(m, cr.ins);
+        return cr.replace.replace(m, cr.ins, inputs_rep_map);
```
pfultz2, line 236: Why are we changing this to take a map and return an instruction?

src/targets/gpu/jit/mlir.cpp
```diff
@@ -21,39 +21,196 @@
...
+            if(i->name() == "@return")
+            {
+                m.add_return(inputs);
+            }
+            else
+            {
+                ins_map[i] = m.add_instruction(i->get_operator(), inputs);
+            }
+        }
+    }
```
pfultz2, line 77: Why doesnt this just do `return *in_mod`?

src/targets/gpu/jit/mlir.cpp
```diff
@@ -21,39 +21,196 @@
...
+            dot_mlir_inputs.push_back(mod_splits[0].mod.get_output_shapes().front());
+            mlir_code_object cop1 = compile_mlir(ctx, mod_splits[0].mod, dot_mlir_inputs, solution);
+            auto pw_inputs        = mod_splits[1].inputs;
+            auto dot_ins_idx      = std::distance(
+                std::find(pw_inputs.begin(), pw_inputs.end(), gemm_ins), pw_inputs.begin());
+            auto pw_shapes         = to_shapes(mod_splits[1].inputs);
+            pw_shapes[dot_ins_idx] = cop1.cop.output;
+            pw_shapes.push_back(mod_splits[1].mod.get_output_shapes().front());
+            assert(pw_shapes.back() == ins->get_shape());
+            auto pw_mod                        = create_pointwise_module(&mod_splits[1].mod);
```
pfultz2, line 113: Why is `create_pointwise_module` called? Why not pass `mod_splits[1].mod` directly to `compile_pointwise`?

src/targets/gpu/jit/mlir.cpp
```diff
@@ -21,39 +21,196 @@
...
+                instruction_ref ins,
+                const std::vector<operation>& ops,
+                const std::unordered_map<instruction_ref, instruction_ref>& inputs_rep_map) {
+                auto dot_inputs        = mods[0].inputs;
+                auto dot_mod_out_shape = mods[0].mod.get_output_shapes().front();
+                auto dot_alloc         = m.insert_instruction(
+                    ins,
+                    migraphx::make_op("hip::allocate", {{"shape", to_value(dot_mod_out_shape)}}));
+                dot_inputs.push_back(dot_alloc);
+                if(not mcos[0].prefill_indices.empty())
```
pfultz2, line 165: This check is redundant.

src/targets/gpu/jit/mlir.cpp
umangyadav: `mod_splits[1].mod` have multibroadcasts instructions which need to be eliminated otherwise compile_pointwise fails.

src/targets/gpu/jit/mlir.cpp
umangyadav: Need to eliminate multibroadcasts

src/targets/gpu/compile_ops.cpp
umangyadav: It is because when compiling mlir dot with split-k config, it would split the input module into two. one with dot and another with pointwise. Split happens when compiling the original instruction. During splits "mod_with_inputs" is cached. This `mod.inputs` would refer to input instructions from original input program to "compile_ops" pass e.g. "precompile_ops". 

Compile happens in parallel. 

Replace happens after everything is compiled. I can't use older cached "precompile_ops" inputs to the dot and pointwise module during the replace. Older instructions are not removed untill Deadcodeelimination runs.  Therefore i need to maintain a map to use the new compiled instructions as the inputs to the "dot" and "pointwise" modules.

src/targets/gpu/jit/mlir.cpp
umangyadav: Removed

src/targets/gpu/compile_ops.cpp
pfultz2: But you have the inputs of the replaced compiled ops as you can get that with `ins->inputs()`. So you just need to map cached inputs from the split module to the new `ins->inputs()`, and this map can be captured by the lambda(so it doesnt need to be passed as a parameter). You dont even need to store a map, just the original `ins->inputs()` vector, and then you can create a map in the replace function using the two vectors.

src/targets/gpu/jit/mlir.cpp
pfultz2: Ah I missed that. You can use `add_instructions` with a custom `inserter` function to eliminate broadcasts, but you need to insert the parameters first though:

```cpp
result.add_instructions(
      in_mod, &map_ins,
      [](module &m, instruction_ref ins, const operation &op,
         const std::vector<instruction_ref> &inputs,
         const std::vector<module_ref> &mod_args) -> instruction_ref {
        if (op.name() == "multibroadcast" and
            inputs.front()->name() == "@literal")
          return inputs.front();
        else
          return m.insert_instruction(ins, op, inputs, mod_args);
      });
```

umangyadav: In order to use `add_instructions`, i need to first create two lists. 
1. instructions that are parameters 
2. instructions that are not parameters.

Step 1. I would need to iterate the the split pointwise module to get that first. 

Step 2. I would need to insert parameters first and maintain a map of old parameter to new parameter. 

Step 3. Go through second list and all remaining instructions while eliminating broadcasts. This would make use of `add_instructions`. 

Instead of going through three steps i find it better to just make use of `create_pointwise_module()` function, which  combines all those three steps into one. 

Instead of going through

src/targets/gpu/compile_ops.cpp
umangyadav: Good idea. Pushed changes. Thanks

src/targets/gpu/jit/mlir.cpp
pfultz2: You dont need to two lists, you just need a list of parameters which you can get with `get_parameters` and then insert them and then insert the module:

```cpp
static module create_pointwise_module(module_ref in_mod) {
  module result;
  std::unordered_map<instruction_ref, instruction_ref> map_ins;
  for (auto param : in_mod->get_parameters())
  {
    map_ins[param] = m.add_parameter(
        any_cast<builtin::param>(param->get_operator()).parameter,
        shape{param->get_shape().type()});
  }
  result.add_instructions(
      in_mod, &map_ins,
      [](module &m, instruction_ref ins, const operation &op,
         const std::vector<instruction_ref> &inputs,
         const std::vector<module_ref> &mod_args) -> instruction_ref {
        if (op.name() == "multibroadcast" and
            inputs.front()->name() == "@literal")
          return inputs.front();
        else
          return m.insert_instruction(ins, op, inputs, mod_args);
      });
  return result;
}
```

umangyadav: `add_instructions` will add parameters again in that case 
https://github.com/ROCm/AMDMIGraphX/blob/2bdd02d38c41802f5883211ffc7d9020b1b9d408/src/module.cpp#L227

and update the `map_ins` map. Therefore i don't think it will work. I need to maintain two lists. 1. parameters and 2. non-parameters instructions and use this overload 
https://github.com/ROCm/AMDMIGraphX/blob/2bdd02d38c41802f5883211ffc7d9020b1b9d408/src/module.cpp#L424

umangyadav: Done.

pfultz2 (approved):

TedThemistokleous (approved):

src/targets/gpu/jit/mlir.cpp
```diff
@@ -21,39 +21,177 @@
...
     {
         auto* smod = ins->module_inputs().front();
         assert(smod->get_parameter_names().size() == ins->inputs().size() - 1);
-        return insert(compile_mlir(ctx, *smod, ins->inputs(), solution));
+        auto gemm_ins = std::find_if(smod->begin(), smod->end(), [&](const auto& i) {
+            return i.name() == "dot" or i.name() == "quant_dot";
+        });
+        // check if (a) module is fused (b) contains a dot instruction and (c) perfConfig can not
+        // allow fused module
+        if(gemm_ins != smod->end() and std::distance(gemm_ins, smod->end()) > 2 and
```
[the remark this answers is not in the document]
umangyadav: smod->end() points to location one past  the Last instruction in the module. 
Last instruction in the fused module is a "Return". 
Therefore if module is fused with "Gemm+Pointwise" then it would be like 
```
Gemm instruction 

Pointwise instruction 

Return 

smod->end() 
```
Therefore i need to check to std::distance counts number of "hops" to reach from start to end. 
In the most simple case that i described above distance would be 3. 
Therefore it need to be >2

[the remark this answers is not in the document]
TedThemistokleous: That clears it up. Thanks for the breakdown. Approved

PR #3031, opened 2024-05-01
Refactor 1 arg dynamic reshape

CharlieL7:
* Allows for more cases of reshaping dynamic input shapes
* For the `-1` case we compare the minimum and maximum shapes to get the new range of values for the output


src/include/migraphx/op/reshape.hpp
```diff
@@ -64,45 +65,83 @@ struct reshape
...
+    // Makes no checks for the validity of the `dims` attribute for the given input shape.
+    shape dyn_1arg_compute_shape(shape s0) const
     {
-        auto dyn_dims = s0.dyn_dims();
-        // track number of fixed elements in input and output
-        std::size_t num_dims_ele = 1;
-        std::size_t num_dd_ele   = 1;
-        for(std::size_t i = 0; i < dyn_dims.size(); ++i)
+        auto input_dyn_dims = s0.dyn_dims();
+        bool has_negative_dim_attr = false;
```
pfultz2, line 73: This can be written as `const bool has_negative_dim_attr = contains(dims, -1);`.

src/include/migraphx/op/reshape.hpp
```diff
@@ -64,45 +65,83 @@ struct reshape
...
             }
         }
-        if(num_dims_ele != num_dd_ele)
+
+        if(has_negative_dim_attr)
         {
-            MIGRAPHX_THROW("Reshape: Number of fixed elements must match. Input: " +
-                           std::to_string(num_dd_ele) + " Output: " + std::to_string(num_dims_ele));
+            // comparing the -1 dimension against the other dimensions
+            auto neg_dim_num = std::distance(this->dims.begin(),
```
pfultz2, line 98: Isn't the negative dimension only allowed as the last dimension?

src/include/migraphx/op/reshape.hpp
pfultz2: Another way since you need the index of the dimension you could search for it first:

```cpp
const auto neg_dim_num = std::distance(this->dims.begin(), std::find(this->dims.begin(), this->dims.end(), -1));
const bool has_negative_dim_attr = neg_dim_num < dims.size();
```

src/include/migraphx/op/reshape.hpp
```diff
@@ -64,45 +65,83 @@ struct reshape
...
+                return a != 0 and c / a != b;
+            };
+
+            // accumulate the minimum and maximum elements in the dimensions before the -1 dimension
+            std::size_t min_cur_elements = 1;
+            std::size_t max_cur_elements = 1;
+            std::size_t max_int          = std::numeric_limits<std::size_t>::max();
+            for(const auto& dd : output_dyn_dims)
+            {
+                min_cur_elements = uint_wraparound_check(min_cur_elements, dd.min)
```
pfultz2, line 113: What you want is [`std::mul_sat`](https://en.cppreference.com/w/cpp/numeric/mul_sat) which unfortunately we cant use that yet, but we can add our own implementation:

```cpp
template<class T>
constexpr T mul_sat(T a, T b) noexcept
{
    T c;
    if (not __builtin_mul_overflow(a, b, &c))
        return c;
    if constexpr(std::is_unsigned<T>{})
        return std::numeric_limits<T>::max();
    if (a < 0 xor b < 0)
        return std::numeric_limits<T>::min();
    return std::numeric_limits<T>::max();
}
```

The builtin should be available in both gcc and clang. Maybe add this under a header such as `sat_ops.hpp`.

src/include/migraphx/op/reshape.hpp
```diff
@@ -64,45 +65,83 @@ struct reshape
...
+        std::vector<shape::dynamic_dimension> output_dyn_dims(dims.size());
+        for(int i = 0; i < dims.size(); ++i)
         {
-            if(dyn_dims[i].is_fixed())
+            int64_t d = dims.at(i);
+            if(d == 0)
+            {
+                output_dyn_dims.at(i) = input_dyn_dims.at(i);
+            }
+            else if(d == -1)
```
lakhinderwalia, line 83: Can there be two indices of value -1 -- maybe checked for error handling? Thanks.

src/include/migraphx/op/reshape.hpp
lakhinderwalia: What about Windows?

pfultz2: > What about Windows?

We compile with clang on windows so the intrinsic shouldn't be a problem.

src/include/migraphx/op/reshape.hpp
CharlieL7: No, it can be any of the dimensions. But there can only be one -1 dimension.

src/include/migraphx/op/reshape.hpp
CharlieL7: No, there can only be one -1 dimension. We check that there is only one -1 dimension earlier during parsing.

lakhinderwalia (approved):

umangyadav (approved):

PR #3034, opened 2024-05-03
`Split` dynamic shape parsing update

CharlieL7:
* Updates parsing of `Split` ONNX operator to support dynamic shapes
  * If splitting over a fixed dynamic dimension we can use the original parsing
  * For splitting over a non-fixed dynamic dimension we must do shape calculations in the graph
    * Currently only supports the case of no `split` attribute or input
    * Should be possible to extend dynamic shape support to other cases in the future


src/onnx/parse_split.cpp
```diff
@@ -49,75 +49,139 @@ struct parse_split : op_parser<parse_split>
...
-                                   std::to_string(info.num_outputs) + "!");
+                    std::size_t chunk_size      = tuned_axis_len / num_outputs + 1;
+                    std::size_t last_chunk_size = tuned_axis_len - chunk_size * (num_outputs - 1);
+                    vec_splits.resize(num_outputs - 1, chunk_size);
+                    vec_splits.push_back(last_chunk_size);
                 }
             }
 
-            if(lens[tuned_axis] % num_outputs == 0)
+            if(std::accumulate(vec_splits.begin(), vec_splits.end(), int64_t(0)) !=
```
lakhinderwalia, line 163: Why should this check on line 163 be there after having split based on prior calculations using  `tuned_axis_len` and `num_outputs` in lines 149 onwards, please?

src/onnx/parse_split.cpp
```diff
@@ -49,75 +49,139 @@ struct parse_split : op_parser<parse_split>
...
+            // {tuned_axis}
+            ret_ins.at(num_outputs - 1) = info.add_instruction(
+                make_op("slice",
+                        {{"axes", {tuned_axis}}, {"ends", {std::numeric_limits<int64_t>::max()}}}),
+                args[0],
+                info.add_instruction(
+                    make_op("mul"),
+                    chunk_size,
+                    info.add_literal(literal{int64_scalar_shape, {num_outputs - 1}})));
+            return ret_ins;
```
lakhinderwalia, line 113: Optional style suggestion. Maybe make lines 60-113 into a separate function.

lakhinderwalia (approved):
Approving this PR. Some stylistic changes can be made, and if the check above is redundant (that I asked about), then please remove it.

src/onnx/parse_split.cpp
```diff
@@ -49,75 +49,139 @@ struct parse_split : op_parser<parse_split>
...
+                        parser.parse_value(info.attributes.at("num_outputs")).at<std::size_t>();
+                    if(num_outputs != info.num_outputs)
+                    {
+                        MIGRAPHX_THROW("PARSE_SPLIT: num_outputs attribute " +
+                                       std::to_string(num_outputs) +
+                                       " doesn't match actual number of outputs " +
+                                       std::to_string(info.num_outputs) + "!");
+                    }
+                }
+                if(tuned_axis_len % num_outputs == 0)
```
lakhinderwalia, line 149: Style: the if/else clause can be compressed by just having a brief if-clause for the last bucket.
```
vec_splits(num_outputs, chunk_size);
auto last = tuned_axis_len%num_outputs;
if(last) vec_splits[num_outputs-1] = last;
```

causten:
Dont forget to fix the format issue

src/onnx/parse_split.cpp
CharlieL7: It's for when the `splits` are given by the user (in the attribute).

src/onnx/parse_split.cpp
CharlieL7: Doesn't work, logic happens such that the last bucket is the smallest and can be 0.

test/onnx/parse/split_dyn_input.cpp
```diff
@@ -0,0 +1,109 @@
...
+        migraphx::make_op("slice", {{"axes", {1}}, {"starts", {0}}, {"ends", {5}}}), input);
+    auto r2 = mm->add_instruction(
+        migraphx::make_op("slice", {{"axes", {1}}, {"starts", {5}}, {"ends", {10}}}), input);
+    auto r3 = mm->add_instruction(
+        migraphx::make_op("slice", {{"axes", {1}}, {"starts", {10}}, {"ends", {15}}}), input);
+    mm->add_return({r1, r2, r3});
+
+    migraphx::onnx_options options;
+    options.default_dyn_dim_value = {10, 30};
+    auto prog = migraphx::parse_onnx("split_dyn_input_fixed_split_axis_test.onnx", options);
```
umangyadav, line 43: use `read_onnx()`

umangyadav (approved):

src/onnx/parse_split.cpp
```diff
@@ -49,75 +174,22 @@ struct parse_split : op_parser<parse_split>
...
-            {
-                std::size_t chunk_size      = lens[tuned_axis] / num_outputs + 1;
-                std::size_t last_chunk_size = lens[tuned_axis] - chunk_size * (num_outputs - 1);
-                vec_splits.resize(num_outputs - 1, chunk_size);
-                vec_splits.push_back(last_chunk_size);
-            }
-        }
+        auto split_axis_is_fixed = [&]() {
+            return input_shape.dyn_dims().at(tuned_axis).is_fixed();
+        };
```
umangyadav, line 183: this can simply be a `bool` variable.

src/onnx/parse_split.cpp
```diff
@@ -34,6 +34,131 @@ namespace migraphx {
...
+        {
+            std::size_t chunk_size = tuned_axis_len / num_outputs;
+            vec_splits.resize(num_outputs, chunk_size);
+        }
+        else
+        {
+            std::size_t chunk_size      = tuned_axis_len / num_outputs + 1;
+            std::size_t last_chunk_size = tuned_axis_len - chunk_size * (num_outputs - 1);
+            vec_splits.resize(num_outputs - 1, chunk_size);
+            vec_splits.push_back(last_chunk_size);
```
umangyadav, line 136: This can potentially lead to resizing and memory allocation twice for the vector. Ideally, you can allocate vector for the splits once for the size (num_outputs) and then put values in that

PR #3036, opened 2024-05-03
Sort parameter inputs into `select_module` from `split_single_dyn_dim` pass

CharlieL7:
* Need to sort the inputs into `select_module` alphabetically by name otherwise there's a possibility that they don't match later because of the `order` attribute of the parameters in the submodules changing.
* I don't know how to make a test that would hit this case, saw the issue with bertsquad-12 and dynamic batch


pfultz2 (approved):

umangyadav:
> I don't know how to make a test that would hit this case 

Is it possible to make a unit-test for the pass itself ? Test that would pass if parameters are sorted and fail if not? You can have expected program with sorted list.

CharlieL7:
> > I don't know how to make a test that would hit this case
> 
> Is it possible to make a unit-test for the pass itself ? Test that would pass if parameters are sorted and fail if not? You can have expected program with sorted list.

Not exactly since this is a compiler pass that makes the `select_module`. I can make a check that essentially confirms that the inputs are ordered alphabetically now.

CharlieL7:
Added the test

test/split_single_dyn_dim_test.cpp
```diff
@@ -287,4 +294,37 @@ TEST_CASE(different_non_fixed_dd)
...
 
+// check that the parameter inputs into select_module are lexiographically ordered
+TEST_CASE(ordered_inputs_to_select_module)
+{
+    migraphx::program p;
+    auto* mm = p.get_main_module();
+    migraphx::shape s{migraphx::shape::float_type, {{1, 4}, {4, 4}}};
+    auto input0 = mm->add_parameter("breadfruit", s);
+    auto input1 = mm->add_parameter("Apricot", s);
+    auto input2 = mm->add_parameter("pineapple", s);
```
umangyadav, line 305: I like these new names :)

umangyadav (approved):

PR #3048, opened 2024-05-07
Improve scalar fusions when not all strides are 0

pfultz2:


umangyadav (approved):

test/fuse_pointwise.cpp
```diff
@@ -243,6 +243,34 @@ TEST_CASE(scalar_input)
...
 
+TEST_CASE(scalar_like_input)
+{
+    migraphx::shape s{migraphx::shape::float_type, {2, 1}};
+    migraphx::program p1;
+    {
+        auto* mm = p1.get_main_module();
+        auto x   = mm->add_parameter("x", s);
+        auto one = mm->add_literal(
+            migraphx::literal{migraphx::shape{migraphx::shape::float_type, {1}}, {1.0f}});
```
TedThemistokleous, line 254: Should you be explicit and make the stride 0 here as well?

umangyadav: `one` is broadcasted later which would introduce strides of zeros.

TedThemistokleous: Ah ok

TedThemistokleous (approved):

PR #3051, opened 2024-05-07
Add hip stream and events to SDXL example

attila-dusnoki-htec:
Streams allows async runs, and events to use more precise timing.
Also fixes some dependency issues.


causten (commented):
Code runs well

causten (approved):

PR #3067, opened 2024-05-09
find_split_concat fix

bpickrel:
Fix a bug that caused an exception in torch-migraphx benchmarks.  Closes Issue https://github.com/ROCm/AMDMIGraphX/issues/3009


src/simplify_algebra.cpp
```diff
@@ -1035,8 +1035,15 @@ struct find_split_concat
             std::find_if(args.begin(), args.end(), [&](auto i) { return i == splits.front(); });
         if(std::distance(it, args.end()) < splits.size())
             return;
+        // Don't do anything if not all operations are slice
+        // if(not std::all_of(it, it + splits.size(), [](instruction_ref x) { 
+        //     return x->get_operator().name() == "slice";
+        // }))
+            // return;
```
pfultz2, line 1042: Why is this commented? That seems like the correct fix.

src/simplify_algebra.cpp
```diff
@@ -1035,8 +1035,15 @@ struct find_split_concat
...
         // If the slices are not in order then stop
         if(not std::is_sorted(it, it + splits.size(), [](instruction_ref x, instruction_ref y) {
+               if(x->get_operator().name() != "slice" or y->get_operator().name() != "slice")
```
pfultz2, line 1045: Dont do this in the comparisons, it doesnt really make sense, and probably can break for a lot of cases.

src/simplify_algebra.cpp
bpickrel: I wanted to show you both ideas I had.  If this looks right to you, I'll comment it in and delete the other.  To confirm:  this matcher should expect that all of the args to the `concat` operator are slice operations--correct?

src/simplify_algebra.cpp
bpickrel: This code would have the effect of ignoring any non-slice operations when checking that slice ops are in order.  If you say that doesn't make sense, that's fine with me.

CharlieL7 (changes requested):
Please add some tests

src/simplify_algebra.cpp
pfultz2: > This code would have the effect of ignoring any non-slice operations when checking that slice ops are in order. 

Thats not what it does. This is a comparison function that returns true when `x` should come before `y`. To say that `x` should come before `y` if either one is not a slice doesnt make sense. Furthermore, if we have vector of `{relu, slice}` then this would be considered in order and we would continue, which is not what we want to do.

pfultz2:
I would assume a test for `concat(relu, slice, relu, slice)` should reproduce the issue. If that doesnt trigger the issue then we can investigate why. We can use `MIGRAPHX_TRACE_MATCHES=1` to see if another matcher is being triggered over this one, and from there we can tweak the test to avoid that.

src/simplify_algebra.cpp
```diff
@@ -1008,40 +1021,58 @@ struct find_split_concat
...
             return;
-        // If the slices are not in order then stop
+        // Don't do anything if the "slice" inputs to the concat op have other operations mixed in
+        // among them
+        if(std::any_of(it, it + splits.size(), [](instruction_ref x) {
+               return x->get_operator().name() != "slice";
+           }))
+            return;
+        // One slice must "start" where the last slice "end"  This parallels the check
+        // in get_splits(), ensuring the inputs to the concat op follow the same rule as the outputs
```
pfultz2, line 1066: This is not the same check that is done in `get_splits`. This is not checking that slice must "start" where the last slice "ends", which is already done in `get_splits`. This is checking that the slices passed to concat are in order.

src/simplify_algebra.cpp
```diff
@@ -1008,40 +1021,58 @@ struct find_split_concat
...
+        // of "ins"
```
pfultz2, line 1067:
```suggestion
        // Check that the slices passed to concat are in order.
```

src/simplify_algebra.cpp
```diff
@@ -1008,40 +1021,58 @@ struct find_split_concat
...
         auto concat_op = any_cast<op::concat>(concat->get_operator());
         auto split_op  = any_cast<op::slice>(splits.front()->get_operator());
         if(split_op.axes.size() != 1)
             return;
         if(split_op.axes.front() != concat_op.axis)
             return;
-        // Replace args
+
+        // Filter the common instruction's inputs (concat can have any number of inputs)
+        // to the ones matching the split list
```
pfultz2, line 1052:
```suggestion
        // Find where the slices are passed to concat
```

test/simplify_algebra_test.cpp
```diff
@@ -2144,6 +2144,34 @@ TEST_CASE(simplify_split_add_relu_multi_axes)
...
+            migraphx::make_op("slice", {{"axes", {2}}, {"starts", {1}}, {"ends", {3}}}), conv);
+        auto concat =
+            m1.add_instruction(migraphx::make_op("concat", {{"axis", 2}}), slice1, relu, slice2);
+
+        m1.add_instruction(pass_op{}, concat);
+    }
+    migraphx::module m2 = m1;
+    run_pass(m1);
+    EXPECT(m1.sort() == m2.sort());
+}
```
pfultz2, line 2173: Move this test to where the `simplify_slice_concat` tests are at.

test/simplify_algebra_test.cpp
```diff
@@ -2144,6 +2144,34 @@ TEST_CASE(simplify_split_add_relu_multi_axes)
     EXPECT(m1.sort() == m2.sort());
 }
 
+TEST_CASE(split_concat_extra_inputs)
```
pfultz2, line 2147: I would call this test `simplify_slice_concat_interleaved_non_slice`.

test/simplify_algebra_test.cpp
```diff
@@ -2144,6 +2144,34 @@ TEST_CASE(simplify_split_add_relu_multi_axes)
...
+    // Matched by matcher find_split_concat, but no substitution because the "slice"
+    // instructions in the input list have a different instruction mixed in
+    migraphx::module m1;
+    {
+        migraphx::shape s{migraphx::shape::float_type, {1, 3, 1, 1}};
+        migraphx::shape ws{migraphx::shape::float_type, {4, 3, 3, 3}};
+
+        auto x    = m1.add_parameter("x", s);
+        auto one  = m1.add_literal(migraphx::generate_literal(ws, 1));
+        auto conv = m1.add_instruction(migraphx::make_op("convolution"), one, x);
```
pfultz2, line 2158: We dont need to add a convolution, it also makes it harder to see what shape is being passed to slice/relu. Instead, just pass the `x` parameter to slice/relu directly and then adjust the shape of `x` parameter to match.

bpickrel: If the convolution isn't there, it doesn't match.

pfultz2: I removed the convolution and it does match:

```cpp
migraphx::shape s{migraphx::shape::float_type, {4, 1, 3, 3}};
auto x    = m1.add_parameter("x", s);
auto slice1 = m1.add_instruction(
    migraphx::make_op("slice", {{"axes", {2}}, {"starts", {0}}, {"ends", {1}}}), x);
auto relu   = m1.add_instruction(migraphx::make_op("relu"), x);
auto slice2 = m1.add_instruction(
    migraphx::make_op("slice", {{"axes", {2}}, {"starts", {1}}, {"ends", {3}}}), x);
auto concat =
    m1.add_instruction(migraphx::make_op("concat", {{"axis", 2}}), slice1, relu, slice2);

m1.add_instruction(pass_op{}, concat);
```

And if I remove the extra check you added, I also get `std::bad_cast` error from this test, so this covers the case without needing the convolution.

bpickrel: You are right.  Updated

pfultz2 (approved):

umangyadav (approved):

PR #3068, opened 2024-05-09
Remove ck_gemm_softmax_gemm test from check CI intermittent failure

lakhinderwalia:


causten (approved):

umangyadav (approved):

src/targets/gpu/jit/ck_gemm_softmax_gemm.cpp
```diff
@@ -228,6 +230,13 @@ struct ck_gemm_softmax_gemm_compiler : compiler<ck_gemm_softmax_gemm_compiler>
         std::vector<shape> gemm_shapes{shapes[0], shapes[1], shapes.back()};
         tc.problem = to_value(gemm_shapes);
         return tc;
+#else
+        (void)ctx;
+        (void)ins;
+        (void)op;
+        (void)exhaustive;
+        return nullopt;
+#endif
```
pfultz2, line 239: This doesnt disable the CK flash attention, it just disables tuning it, and it just so happens that picking `5` by default works for the this test case. The kernel is still offloaded to ck, when running the test:

```
@0 = check_context::migraphx::gpu::context -> float_type, {}, {}, target_id=0
output = @param:output -> half_type, {1, 12, 256, 256}, {786432, 65536, 256, 1}, target_id=0
3 = @param:3 -> half_type, {1, 12, 256, 256}, {786432, 65536, 256, 1}, target_id=0
2 = @param:2 -> half_type, {1, 12, 256, 256}, {786432, 65536, 256, 1}, target_id=0
1 = @param:1 -> half_type, {1, 12, 256, 256}, {786432, 65536, 256, 1}, target_id=0
@5 = transpose[permutation={0, 1, 3, 2}](2) -> half_type, {1, 12, 256, 256}, {786432, 65536, 1, 256}, target_id=0
@6 = gpu::code_object[code_object=25192,symbol_name=ck_gemm_softmax_gemm_kernel,global=24576,local=256,](1,@5,3,output) -> half_type, {1, 12, 256, 256}, {786432, 65536, 256, 1}, target_id=0
```

lakhinderwalia: (This was driven by the logic of fixing our CI issue. So the goal was, not to exhaustive-tune it, and avoid going through the offending series of options. Ideally somebody should be looking at the root cause of it. Else, If there is a functional goal that we should disable this for ck, entirely, then that could be done too in a new PR. Thanks.)

pfultz2: It would be better to either disable offloading to FA(which is more changes), or use an env variable to disable the tuning in the CI. Of course the fastest way to fix the CI is probably to just disable the entire pipeline like in #3087.

PR #3069, opened 2024-05-10
Fix bug in computing dominator

pfultz2:
We were not propagating the dominators for single input instructions, which caused us to miss some dominators.


test/dom.cpp
```diff
@@ -23,30 +23,108 @@
  */
 #include <migraphx/dom_info.hpp>
 #include <migraphx/program.hpp>
+#include <migraphx/ranges.hpp>
+#include <migraphx/iterator_for.hpp>
 #include <basic_ops.hpp>
 #include <test.hpp>
 
+bool dominates_self(const migraphx::dominator_info& dom, const migraphx::module& m)
```
TedThemistokleous, line 31: I don't see where this is used.

umangyadav: It is used in second test. 

@pfultz2  Question for curiosity, Is it ever possible for a node to dominate itself in ML Graph ? I can't think of a case like that.

test/dom.cpp
```diff
@@ -23,30 +23,108 @@
...
+{
+    return migraphx::any_of(migraphx::iterator_for(m), [&](auto ins) {
+        return dom.strictly_dominate(ins, ins);
```
umangyadav, line 34: This would always return false.  https://github.com/ROCm/AMDMIGraphX/blob/cc7c280162e0fcf9e08dce71f8abe6cede1089a4/src/dom_info.cpp#L37

test/dom.cpp
pfultz2: Node cannot strictly dominate itself by [definition](https://en.wikipedia.org/wiki/Dominator_(graph_theory)):

> A node d strictly dominates a node n if d dominates n and d does not equal n.

test/dom.cpp
pfultz2: Yes it should always return false, which is why I added a check for that

umangyadav (approved):

TedThemistokleous (approved):

PR #3070, opened 2024-05-10
Fix bug in finding parallel reductions

### LINKED ISSUE #3046: Infinite recursion in module::sort(), speech_transformer
Model `speech_transformer` in the DLM pytorch performance test causes an exception which is traced to an infinite recursion in the MIGRAPHX method `module::sort()`.  

The error is only seen in branch `fix_find_pointwise_reduce` because in the `develop` branch, the test exits earlier due to a different bug.  The error is not seen in commit `efc01466f` which was created from `develop` commit `ee68f7261f2` but does occur when I merge with commit `2bdd02d38c41` (may 3).  It's not known what commit between those two first introduces the bug.

Steps:

1. Set up DLM performance test environment
2. Check out branch `fix_find_pointwise_reduce` and then merge branch `develop`.
3. Run the test script `python benchmarks/dynamo/torchbench.py  --inference --float16 -dcuda --performance --backend migraphx -k  speech_transformer`
4. Multiple `*.mxr` models are created.  Run MIGraphX driver on the first one:  `bin/driver compile ../../pytorch/fused_0.mxr` to see the fail.

pfultz2:
Using dominators are not the right tool for this, which lead to making operators parallel when they werent leading to infinite recursion. Parallelism can be detected using post dominance frontiers, but it seems much simpler to check for reachability using DFS. 

This fixes #3046.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ARG PREFIX=/usr/local
 RUN dpkg --add-architecture i386
 
 # Install rocm key
-RUN apt-get update && apt-get install -y gnupg2 --no-install-recommends curl && \
+RUN apt-get update && apt-get install -y software-properties-common gnupg2 --no-install-recommends curl && \
     curl -sL http://repo.radeon.com/rocm/rocm.gpg.key | apt-key add -
 
 # Add rocm repository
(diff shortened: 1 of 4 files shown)


TedThemistokleous (approved):

shivadbhavsar (approved):
I am not seeing the segfaults with this PR

PR #3074, opened 2024-05-10
Fix missing/dud parse tests

umangyadav:
Some of the onnx tests are using wrong file names inside the tests that expects "throws".  They pass without testing anything because files themselves were missing because of wrong filenames.


TedThemistokleous (approved):

pfultz2 (approved):
LGTM. We should probably consider creating a function that will abort on missing file instead of throw since we rely on that for testing.

TedThemistokleous:
> LGTM. We should probably consider creating a function that will abort on missing file instead of throw since we rely on that for testing.

Agreed. Likely these were from previous tests I created/deleted too.

PR #3075, opened 2024-05-13
Add onnx tests to rocm test package

umangyadav:
Solves https://github.com/ROCm/AMDMIGraphX/issues/3019

To add onnx tests to test package, it uses embed library to put all the onnx files to one static library. 

A few (3 in total) of the onnx files uses "external" weight files.  Those external files needs to be in same place as the onnx files to be found and loaded. PR uses another embed library for the weights. 

Weights can't be passed as a memory buffers. ONNX does't have an API to directly pass buffer for the external data. Therefore I am writing those weight files from the embed library to the test locations and deleting them once it goes out of scope. 

Testing :
`make package && dpkg -i *.deb`
`cd /opt/rocm/share/test/migraphx`
`ctest -VVV -R test_onnx_test . && ctest -VVV -R test_verify_onnx . && ctest -VVV -R test_onnx_rnn_test .`

As a side benefit all onnx tests can be executed directly from `./build/bin` folder without changing working directory.


pfultz2:
> Weights can't be passed as a memory buffers. ONNX does't have an API to directly pass buffer for the external data. Therefore I am writing those weight files from the embed library to the test locations and deleting them once it goes out of scope.

I dont think we should write temp files to the test directory. This would require running the tests with sudo, which we should not require. Instead we should do either:

1. Create a temp directory to write the weights, and have a onnx option to search for weights in the temp directory.
2. Add onnx option to load weights from a buffer directly instead of loading a file. This most likely could be added as a `std::function<std::string(std::string_view name)>` that takes the filename and returns a `std::string` when the buffer is found.

test/onnx/CMakeLists.txt
```diff
@@ -24,14 +24,19 @@
...
-    add_test(NAME ${TEST_NAME} COMMAND $<TARGET_FILE:${TEST_NAME}> WORKING_DIRECTORY ${TEST_ONNX_DIR})
-    rocm_mark_as_test(${TEST_NAME})
 endfunction()
 
+include(Embed)
+file(GLOB_RECURSE ONNX_FILES CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/*.onnx)
+add_embed_library(onnx_files ${ONNX_FILES} RELATIVE ${CMAKE_CURRENT_SOURCE_DIR})
+
+file(GLOB_RECURSE WEIGHT_FILES CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/*.weight)
+add_embed_library(weight_files ${WEIGHT_FILES} RELATIVE ${CMAKE_CURRENT_SOURCE_DIR})
```
pfultz2, line 38: There is no need to use two different embed libraries. We can distinguish it based on the name. Although, if you pick option 2, this may not even be necessary.

test/onnx/include/onnx_test.hpp
```diff
@@ -35,14 +35,78 @@
...
+    return true;
+}
+
+inline migraphx::program read_onnx(const std::string& name,
+                                   migraphx::onnx_options options = migraphx::onnx_options{})
+{
+    static auto onnx_files{::onnx_files()};
+    static bool read_once = read_weight_files();
+    (void)(read_once);
+    auto prog = migraphx::parse_onnx_buffer(std::string{onnx_files[name]}, options);
```
pfultz2, line 101: This should check that the file exists in the map, and then abort with a message(not throw).

umangyadav: DOne

test/onnx/CMakeLists.txt
umangyadav: Done

test/onnx/include/onnx_test.hpp
```diff
@@ -35,14 +35,55 @@
...
+{
+    static migraphx::tmp_dir td{"weights"};
+    std::vector<std::pair<std::string_view, std::string_view>> weight_files;
+    for(const auto& i : onnx_files)
+    {
+        if(migraphx::ends_with(std::string{i.first}, "weight"))
+        {
+            weight_files.push_back(i);
+        }
+    }
```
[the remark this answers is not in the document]
pfultz2:
```cpp
    for(const auto& i : onnx_files)
    {
        if(migraphx::ends_with(std::string{i.first}, "weight"))
            continue;
        migraphx::fs::path full_path   = td.path / i.first;
        migraphx::fs::path parent_path = full_path.parent_path();
        migraphx::fs::create_directories(parent_path);
        migraphx::write_buffer(full_path, i.second.data(), i.second.size());
    }
```

test/onnx/include/onnx_test.hpp
```diff
@@ -35,14 +35,55 @@
...
+    for(const auto& i : weight_files)
+    {
+        migraphx::fs::path full_path   = td.path / i.first;
+        migraphx::fs::path parent_path = full_path.parent_path();
+        migraphx::fs::create_directories(parent_path);
+        migraphx::write_buffer(full_path, i.second.data(), i.second.size());
+    }
```
pfultz2, line 63:
```suggestion
    for(const auto& i : weight_files)
    {
        if(migraphx::ends_with(std::string{i.first}, "weight"))
            continue;
        migraphx::fs::path full_path   = td.path / i.first;
        migraphx::fs::path parent_path = full_path.parent_path();
        migraphx::fs::create_directories(parent_path);
        migraphx::write_buffer(full_path, i.second.data(), i.second.size());
    }
```

test/onnx/include/onnx_test.hpp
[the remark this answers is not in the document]
umangyadav: Done

pfultz2 (approved):

CharlieL7 (approved):
LGTM

PR #3081, opened 2024-05-14
Add GPU pooling kernel

pfultz2:


CharlieL7 (commented):
Not getting rid of the miopen pooling `src/targets/gpu/pooling.cpp`?

pfultz2:
> Not getting rid of the miopen pooling src/targets/gpu/pooling.cpp?

Not yet, good to have it as a fallback just in case we find a bug.

test/verify/test_lpnorm_pooling_pad.cpp
```diff
@@ -0,0 +1,46 @@
...
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+        auto x   = mm->add_parameter("x", {migraphx::shape::float_type, {1, 3, 5, 5}});
+        mm->add_instruction(migraphx::make_op("pooling",
+                                              {{"mode", migraphx::op::pooling_mode::lpnorm},
+                                               {"padding", {2, 2}},
+                                               {"stride", {1, 1}},
+                                               {"lengths", {3, 3}},
+                                               {"lp_order", 2}}),
```
umangyadav, line 42: What's the difference between this test and `test_lpnorm_pooling` ? i do not find any diff. Can you do count_include_pad test for the maxpool and lpnorm ?

umangyadav: add a few tests with `ceil_mode` enabled

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
```diff
@@ -0,0 +1,184 @@
...
+    MIGRAPHX_DEVICE_CONSTEXPR T final(T x, index_int) const
+    {
+        if constexpr(P == 0)
+            return 1;
+        else if constexpr(P == 1)
+            return x;
+        else if constexpr(P == 2)
+            return migraphx::sqrt(x);
+        else
+            return migraphx::pow(x, 1. / P);
```
umangyadav, line 107: `1/P` would it be considered as constexpr ?

pfultz2: What do you mean?

pfultz2: Ah you mean can it be const-folded?

test/verify/test_lpnorm_pooling_pad.cpp
pfultz2: This is only supposed to have padding, let me update the other tests to remove the padding.

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
umangyadav: Yes i mean if it can be const folded.

pfultz2:
> From what I can figure out, it looks like each thread calculates a window and then moves on to another window depending on the number of threads.

Its a lane reduction, so one workitem does one window for pooling, this is similiar to miopen.

> Have we considered other work distribution schemes?

Yes we can do a  "blockwise"(or it might be smaller than a block or wave) reduction across the window and do a lane reduction down the window. This should be faster, but I plan to implement this as enhancement in the future.

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
pfultz2: Yea it should be const folded by the compiler.

CharlieL7 (approved):
LGTM, do we have an issue tracking the "block-wise reduction" enhancement?

test/verify/test_lpnorm_pooling_pad.cpp
pfultz2: Add `ceil_mode` tests for lpnorm as well.

pfultz2:
> LGTM, do we have an issue tracking the "block-wise reduction" enhancement?

https://github.com/ROCm/AMDMIGraphX/issues/3175

src/targets/cpu/lowering.cpp
```diff
@@ -415,12 +415,23 @@ struct cpu_apply
                        {ins->inputs().front()});
     }
 
+    // TODO:  update lowering to run the reference
+    // code when OneDNN can't execute pooling for a CPU
```
umangyadav, line 419: Lowering to reference backend can be a good first issue.

umangyadav (approved):

src/include/migraphx/op/pooling.hpp
```diff
@@ -344,9 +331,8 @@ struct pooling
                 int end;
                 std::size_t dilated_kernel_dim = dilate_dim(kernel_dims[d_2], dilations[d_2]);
                 // NOLINT
-                if(count_include_pad and ceil_mode and (mode != pooling_mode::max))
+                if(count_include_pad and (mode != pooling_mode::max))
```
umangyadav, line 334: Comments on line 346 
"In non ceiling mode" 
requires update

umangyadav (changes requested):

umangyadav (approved):

lakhinderwalia:
This PR breaks Resnet Model accuracy tests. 
See: https://ontrack-internal.amd.com/browse/SWDEV-471087

PR #3082, opened 2024-05-14
Fix `is_module_fusible()` bug in MIGraphX 

### LINKED ISSUE #3079: Add Splitk GEMM test that would split pointwise module
k<=1024 for MIGraphX to pick MLIR by default. 

Some GEMM configs that has higher probablity of picking split-k config from MLIR are as follows: 

-g 1 -m 2 -n 1280 -k 2816
-g 1 -m 2 -n 1280 -k 1280
-g 1 -m 2 -n 1280 -k 320
-g 1 -m 32 -n 64 -k 1024

Exhaustive-tune needs to be enabled for the split-k to be picked.

umangyadav:
`is_module_fusible()` needs to be run after `run_highlevel_pipeline()` otherwise it always returns True. 

Enable test for the split-k + pointwise by turning on exhaustive-tune for the `gemm_add` test.

Also fixes #3079

### DIFF
--- a/src/targets/gpu/mlir.cpp
+++ b/src/targets/gpu/mlir.cpp
@@ -947,10 +947,12 @@ struct mlir_program
     std::string sym_name;
 };
 
-bool is_module_fusible(const module& m, const value& solution)
+bool is_module_fusible(const module& m, const context& migraphx_ctx, const value& solution)
 {
     mlir_program mp;
+    mp.set_gpu_properties(migraphx_ctx);
     mp.parse(m);
+    mp.run_high_level_pipeline();
     return mlirIsModuleFusible(mp.mmodule.get(), make_mlir_string_ref(*solution.if_string()));
 }
 
--- a/src/instruction.cpp
+++ b/src/instruction.cpp
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
- * Copyright (c) 2015-2023 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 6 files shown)


src/targets/gpu/mlir.cpp
```diff
@@ -947,10 +947,12 @@ struct mlir_program
     std::string sym_name;
 };
 
-bool is_module_fusible(const module& m, const value& solution)
+bool is_module_fusible(const module& m, const context& migraphx_ctx, const value& solution)
 {
     mlir_program mp;
+    mp.set_gpu_properties(migraphx_ctx);
```
pfultz2, line 953: Does gpu properties need to be set for the high level pipeline?

umangyadav: Yes otherwise MLIR prints warnings like 
`Warning: unknown architecture, falling back to defaults:`

TedThemistokleous (approved):
Outside of Paul's question this makes sense.

src/targets/gpu/mlir.cpp
pfultz2: But is that a problem for `is_module_fusible`?

umangyadav: `set_gpu_properties` is only setting the `arch` on the MLIR function. MLIR can potentially run different route while running high level backend pipeline. 

Also we don't want warnings that can be fixed.

pfultz2: I assume the high level pipeline is target independent so it seems unnecessary, but this would be a better question for @sjw36 @krzysz00 @manupak to answer.

umangyadav: MLIR team can answer it, but i don't see why this issue would be a blocker for approval. I don't see any harm in having a call to `set_target_properties()` and setting `arch` attribute.  arch is set later too when compiling MLIR module.

manupak: > I assume the high level pipeline is target independent so it seems unnecessary.

In a practical sense yes, but we do extract features from arch in the high-level; so that further lowerings (low-level) only look at a bag of features that is there on the arch.

pfultz2 (approved):

PR #3086, opened 2024-05-15
Fix mod op compile issue on GPU

gyulaz-htec:
Lifting the failing int8, uint8 and int32 data types to float to workaround the problem described in https://github.com/ROCm/AMDMIGraphX/issues/2268#issuecomment-2112365339


umangyadav (approved):

pfultz2:
We should add a function to do the mod on the GPU correctly instead of falling back to float. Maybe something like:

```cpp
template <class T, class U>
constexpr auto mod(T a, U b)
{
    if constexpr(is_integral<T>{} and is_integral<U>{})
        return a % b;
    return vec_transform(a, b)([](auto x0, auto x1) -> T { 
        return fmod(remainder(x0, x1) + x1, x1); 
    });
}
```

gyulaz-htec:
@umangyadav @pfultz2 I've updated the code according to what @pfultz2 mentioned.
I had to go with `((a % b) + b) % b` instead of `a % b` to match the rounding mode of modulos to what onnx expects, because rounding for modulus in c++ implementation dependant.

Issue #3095, opened 2024-05-16
Concat rewrite to handle more cases


CharlieL7:
Found during investigation for issue #3023 

Issue is in find_concat_multibroadcasts of simplify_reshapes.
What we matched is:
```
param0: lens = [1, 1, 1, 1, 64], strides = [1, 1, 60, 64, 194]
param1: lens = [1, 1, 60, 64, 192], strides = [737280, 737280, 12288, 192 ,1]
mb0: multibroadcast(param0, lens = [1, 12, 60, 64, 64])
mb1: multibroadcast(param1, lens = [1, 12, 60, 64, 192])
concat(mb0, mb1, axis = 4)
```
It currently tries to move the concat before the multibroadcast to:
```
param0: lens = [1, 1, 1, 1, 64], strides = [1, 1, 60, 64, 194]
param1: lens = [1, 1, 60, 64, 192], strides = [737280, 737280, 12288, 192 ,1]
con0: concat(param0, param1, axis = 4)
multibroadcast(con0, lens = [1, 12, 60, 64, 256])
```
This fails because concat does not work if the dimensions other than axis to concat over are not the same.

* We should be able to rewrite `concat` to handle this case also.

PR #3096, opened 2024-05-16
Concat - multibroadcast fix

### LINKED ISSUE #3023: normalize_compute_shape: CONCAT: all input dimensions should match along axis 2
Torch benchmarks fail.
Affected models:
Torchbench: hf_BigBird

CharlieL7:
* Resolves #3023
* Adds checks and their respective tests for cases that the concat-multibroadcast matcher cannot currently handle.
  * See the issue for a breakdown and #3095 to track a rewrite of the this matcher to handle these cases.
* Added documentation and chose new variable names to make what the matcher is doing more clear.

### DIFF
--- a/src/simplify_reshapes.cpp
+++ b/src/simplify_reshapes.cpp
@@ -282,7 +282,7 @@ struct find_concat_multibroadcasts
             return;
         }
 
-        // Get the inputs of multibroadcast ops, will be used as inputs to new concat op
+        // Get the inputs of multibroadcast ops. Will be used as inputs to new concat op
         std::vector<instruction_ref> mb_inputs(concat_inputs.size());
         std::transform(concat_inputs.begin(), concat_inputs.end(), mb_inputs.begin(), [](auto i) {
             return i->inputs().front();
@@ -307,15 +307,9 @@ struct find_concat_multibroadcasts
         const auto& front_in_lens = mb_inputs.front()->get_shape().lens();
         for(std::size_t ax = 0; ax < front_in_lens.size(); ++ax)
         {
-            if(ax != concat_op.axis)
-            {
-                if(not std::all_of(mb_inputs.begin(), mb_inputs.end(), [&](auto input_to_mb) {
-                       return input_to_mb->get_shape().lens()[ax] == front_in_lens[ax];
-                   }))
-                {
-                    return;
-                }
-            }
+			const auto& lens = input_to_mb->get_shape().lens();
+			return std::equal(lens.begin(), lens.begin() + concat_op.axis, front_in_lens.begin()) and 
+				   std::equal(lens.begin() + concat_op.axis + 1, lens.end(), front_in_lens.begin() + concat_op.axis + 1);
         }
 
         auto new_concat_ins = m.insert_instruction(concat_ins, concat_op, mb_inputs);
--- a/src/simplify_reshapes.cpp
+++ b/src/simplify_reshapes.cpp
@@ -299,17 +299,24 @@ struct find_concat_multibroadcasts
 
         // Reduce axis by number of leading broadcasted dimensions
         if(mb_inputs.front()->get_shape().lens().size() < concat_out_lens.size())
+        {
             concat_op.axis -=
                 std::count(front_mb_strides.begin(), front_mb_strides.begin() + concat_op.axis, 0);
+        }
 
         // Inputs to multibroadcasts should have the same dimensions except for the axis to
         // concatenate over
         const auto& front_in_lens = mb_inputs.front()->get_shape().lens();
-        for(std::size_t ax = 0; ax < front_in_lens.size(); ++ax)
+        if(not std::all_of(mb_inputs.begin() + 1, mb_inputs.end(), [&](auto input_to_mb) {
+               const auto& lens = input_to_mb->get_shape().lens();
+               return std::equal(
+                          lens.begin(), lens.begin() + concat_op.axis, front_in_lens.begin()) and
+                      std::equal(lens.begin() + concat_op.axis + 1,
+                                 lens.end(),
+                                 front_in_lens.begin() + concat_op.axis + 1);
+           }))
         {
-			const auto& lens = input_to_mb->get_shape().lens();
-			return std::equal(lens.begin(), lens.begin() + concat_op.axis, front_in_lens.begin()) and 
-				   std::equal(lens.begin() + concat_op.axis + 1, lens.end(), front_in_lens.begin() + concat_op.axis + 1);
+            return;
         }
 
         auto new_concat_ins = m.insert_instruction(concat_ins, concat_op, mb_inputs);
--- a/src/simplify_reshapes.cpp
+++ b/src/simplify_reshapes.cpp
@@ -273,6 +273,7 @@ struct find_concat_multibroadcasts
         auto concat_out_lens  = concat_ins->get_shape().lens();
         auto concat_inputs    = concat_ins->inputs();
         auto front_mb_strides = concat_inputs.front()->get_shape().strides();
+        assert(concat_op.axis >= 0);
 
         // Only apply when concat axis is not a broadcasted dimension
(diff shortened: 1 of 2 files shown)


src/simplify_reshapes.cpp
```diff
@@ -253,32 +268,60 @@ struct find_concat_multibroadcasts
...
+                std::count(front_mb_strides.begin(), front_mb_strides.begin() + concat_op.axis, 0);
+
+        // Inputs to multibroadcasts should have the same dimensions except for the axis to
+        // concatenate over
+        const auto& front_in_lens = mb_inputs.front()->get_shape().lens();
+        for(std::size_t ax = 0; ax < front_in_lens.size(); ++ax)
+        {
+            if(ax != concat_op.axis)
+            {
+                if(not std::all_of(mb_inputs.begin(), mb_inputs.end(), [&](auto input_to_mb) {
```
pfultz2, line 312: You skip the first element: `std::all_of(mb_inputs.begin()+1, mb_inputs.end(), ...)`.

src/simplify_reshapes.cpp
```diff
@@ -253,32 +268,60 @@ struct find_concat_multibroadcasts
...
+                       return input_to_mb->get_shape().lens()[ax] == front_in_lens[ax];
```
pfultz2, line 313: This could be written as:

```cpp
const auto& lens = input_to_mb->get_shape().lens();
return std::equal(lens.begin(), lens.begin()+concat_op.axis, front_in_lens.begin()) and 
       std::equal(lens.begin()+concat_op.axis, lens.end(), front_in_lens.begin()+oncat_op.axis);
```

And then you dont need the outer loop.

test/simplify_reshapes_test.cpp
```diff
@@ -930,6 +931,48 @@ TEST_CASE(concat_multibroadcasts4)
...
+    auto mb_y = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mb_lens1}}), y);
+    auto concat_xy = m.add_instruction(migraphx::make_op("concat", {{"axis", 4}}), mb_x, mb_y);
+    m.add_return({concat_xy});
+    auto m_original = m;
+    run_pass(m);
+    EXPECT(m == m_original);
+}
+
+// concat axis moved because rank(in_dims) < rank(out_dims)
+// parameter inputs are not the same rank
```
bpickrel, line 956:
```suggestion
// matched by find_concat_multibroadcasts but parameter inputs are not the same rank
```

test/simplify_reshapes_test.cpp
```diff
@@ -930,6 +931,48 @@ TEST_CASE(concat_multibroadcasts4)
...
+    auto mb_x = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mb_lens0}}), x);
+    auto mb_y = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mb_lens1}}), y);
+    auto concat_xy = m.add_instruction(migraphx::make_op("concat", {{"axis", 4}}), mb_x, mb_y);
+    m.add_return({concat_xy});
+    auto m_original = m;
+    run_pass(m);
+    EXPECT(m == m_original);
+}
+
+// concat axis moved because rank(in_dims) < rank(out_dims)
```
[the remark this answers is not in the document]
CharlieL7: I say we keep it. The most important part of this test is that the rank of the parameters is different. Can add another test if we want to hit the leading axis code and the other dims check.

test/simplify_reshapes_test.cpp
```diff
@@ -918,9 +918,10 @@ TEST_CASE(concat_multibroadcasts3)
     EXPECT(new_concat->get_operator().to_value()["axis"].to<int>() == 2);
 }
 
+// Broadcasted batch dim, axis is broadcasted dim
+// skips this case
```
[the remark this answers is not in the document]
CharlieL7: I think it's somewhat obvious that this is looking to test the `find_concat_multibroadcasts` matcher based on the name of the test. But the additional comment is fine.

test/simplify_reshapes_test.cpp
```diff
@@ -918,9 +918,10 @@ TEST_CASE(concat_multibroadcasts3)
     EXPECT(new_concat->get_operator().to_value()["axis"].to<int>() == 2);
 }
 
+// Broadcasted batch dim, axis is broadcasted dim
```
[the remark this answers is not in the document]
CharlieL7: > A similar problem would come up if the order of passes in simplify_reshapes changes and causes a match where there was none before, or vice versa--a different matcher modifies an instruction graph before the intended matcher can see it.

That's possible, but quite unlikely given how small these test cases are. The only way I can think of to actually prevent this is to have passes that only run one matcher.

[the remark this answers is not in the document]
CharlieL7: > We're going to have a problem with these tests in the future if any change to simplify_reshapes breaks one of them, in that it's not made clear which of the simplify_reshapes matchers each test is intended to match.

I don't follow your logic here. All we would have to do is turn on `MIGRAPHX_TRACE_MATCHES`.

test/simplify_reshapes_test.cpp
```diff
@@ -930,6 +931,48 @@ TEST_CASE(concat_multibroadcasts4)
...
+    std::vector<std::size_t> mb_lens0 = {1, 12, 60, 64, 64};
+    std::vector<std::size_t> mb_lens1 = {1, 12, 60, 64, 192};
+    auto mb_x = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mb_lens0}}), x);
+    auto mb_y = m.add_instruction(migraphx::make_op("multibroadcast", {{"out_lens", mb_lens1}}), y);
+    auto concat_xy = m.add_instruction(migraphx::make_op("concat", {{"axis", 4}}), mb_x, mb_y);
+    m.add_return({concat_xy});
+    auto m_original = m;
+    run_pass(m);
+    EXPECT(m == m_original);
+}
```
pfultz2, line 974: Can you add TODO comments on these test that we will simplify them in the future?

pfultz2:
Another test case that should be added is for the inputs `{64, 64}` and `{60, 1, 192}` with axis=2. We should not do the transformation for this case ever(so no TODOs in the comments). 

For bonus points, we should also reject `{1, 64, 64}` and `{60, 1, 192}` even when they are the same rank. For that you need to check there is a common broadcasted axes across all inputs. There is an example of how to that [here](https://github.com/ROCm/AMDMIGraphX/blob/dcedcee77a82b0f93777aecb2990cf38a0a9ffed/src/simplify_algebra.cpp#L527) in this old version of `find_inner_broadcast`(since then its been replaced), but you could use it here.

src/simplify_reshapes.cpp
```diff
@@ -253,32 +268,60 @@ struct find_concat_multibroadcasts
...
+        if(mb_inputs.front()->get_shape().lens().size() < concat_out_lens.size())
+            concat_op.axis -=
+                std::count(front_mb_strides.begin(), front_mb_strides.begin() + concat_op.axis, 0);
+
+        // Inputs to multibroadcasts should have the same dimensions except for the axis to
+        // concatenate over
+        const auto& front_in_lens = mb_inputs.front()->get_shape().lens();
+        for(std::size_t ax = 0; ax < front_in_lens.size(); ++ax)
+        {
+            if(ax != concat_op.axis)
```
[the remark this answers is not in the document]
CharlieL7: I thought that `normalize_compute_shape()` would replace the operator with another one with a normalized `axis` attribute?

test/simplify_reshapes_test.cpp
CharlieL7: Not sure I understand what you mean by simplify these in the future. I'm only intending to change what the comment says.

CharlieL7: Do you mean when we rewrite the matcher for doing a broadcast before if there's atleast one common broadcast axis?

bpickrel (approved):

PR #3098, opened 2024-05-16
Use the package name hip-dev/devel rather than using the deprecated package name hip-base.

raramakr:
While installing versioned packages, the dependency list also contains non versioned package hip-dev. This was due to the usage of deprecated package name hip-base in the dependency list of mipgraphx. Replaced the same with hip-dev/devel

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,8 +343,8 @@ if(MIGRAPHX_USE_ROCBLAS)
     list(APPEND PACKAGE_DEPENDS rocblas)
 endif()
 
-rocm_package_add_deb_dependencies(DEPENDS "hip-dev")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -343,11 +343,14 @@ if(MIGRAPHX_USE_ROCBLAS)
     list(APPEND PACKAGE_DEPENDS rocblas)
 endif()
 
+rocm_package_add_deb_dependencies(DEPENDS "hip-dev")
```
arvindcheru, line 346: Do we need do add this as shared dependency "rocm_package_add_dependencies(**SHARED_DEPENDS** "

nunnikri (approved):

ahsan-ca (approved):

PR #3100, opened 2024-05-17
Fix remove rocblas tests

tvukovic-amd:
- Added unsupported fp8 type for dot in case when MIGRAPHX_USE_ROCBLAS is OFF since mlir dot doesn't support fp8
- execute gemm tune tests when MIGRAPHX_USE_ROCBLAS is ON since they use rocBLAS

@apwojcik

### DIFF
--- a/src/targets/gpu/target.cpp
+++ b/src/targets/gpu/target.cpp
@@ -104,9 +104,9 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         unsupported_fp8_ops.insert("quant_dot");
     }
 #else
-        // mlir doesn't support fp8 dot
-        unsupported_fp8_ops.insert("dot");
-        unsupported_fp8_ops.insert("quant_dot");
+    // mlir doesn't support fp8 dot
+    unsupported_fp8_ops.insert("dot");
+    unsupported_fp8_ops.insert("quant_dot");
 #endif
     // MIOpen doesn't have support for fp8 pooling yet.
     unsupported_fp8_ops.insert("pooling");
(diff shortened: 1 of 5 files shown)


src/targets/gpu/target.cpp
```diff
@@ -103,6 +103,10 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         unsupported_fp8_ops.insert("dot");
         unsupported_fp8_ops.insert("quant_dot");
     }
+#else
+    // mlir doesn't support fp8 dot
+    unsupported_fp8_ops.insert("dot");
+    unsupported_fp8_ops.insert("quant_dot");
```
umangyadav, line 109: This shouldn't be necessary. 
https://github.com/ROCm/AMDMIGraphX/blob/7011bda8a316bdff7b43a873ff0d6088155bcf0a/src/targets/gpu/fuse_mlir.cpp#L234 

MIGraphX doesn't pick rocMLIR for the FP8 dot/quant_dots. 

Do you have a case where rocBLAS is disbaled in MIGraphX ?

apwojcik: Yes, that is required when rocBLAS is not used (see `MIGRAPHX_USE_ROCBLAS=OFF`).

tvukovic-amd: Yes, this is added for case where rocBLAS is disabled in MIGraphX. There is variable MIGRAPHX_USE_ROCBLAS in develop branch, so when it is set to OFF we need to add unsupported fp8 ops for dot and quant dot.

umangyadav (approved):

src/targets/gpu/target.cpp
pfultz2, line 109: This just duplicating code for both branches. It would be better to have `rocblas_fp8_available` function that just always returns false when rocblas is disabled.

PR #3104, opened 2024-05-18
Better simplification of shape transformation operators such as reshape/transpose/broadcast

pfultz2:
Currently, in migraphx, we only simplify shape transformation of the same operator such as repeated transpose or reshapes.

However, this will simplify across reshape/transpose/broadcast. It produces a much simpler set of transformations even across our current unit tests.

It also canonicalizes many shape transformations so they will be done with the same set of operators where possible or in the best way to best preserve broadcasting and layout. It will use squeeze and unsqueeze where possible to preserve layout and it will try to end with a broadcast. So starting with a dimension of `2, 32, 1` and doing a `multibroadcast[out_lens={2, 32, 256}]` -> `reshape[dims={2, 32, 16, 16}]` now becomes `unsqueeze[axes={3}]` -> `multibroadcast[out_lens={2, 32, 16, 16}]`.

These simplifications are all done with a data structure that I called `shape_transform_descriptor`(perhaps this isn't the best name for it, so open to ideas for a better name). This class will record the transformations being done, apply simplifications and then generate the operators needed to produce such transformation.


CharlieL7 (commented):
For the `reshape` operator: do we want it to do the operations that are not covered by the other operators? Since `reshape` & `reshape_lazy` together can technically do all of the other transformations

pfultz2:
> For the reshape operator: do we want it to do the operations that are not covered by the other operators?

What do you mean? `reshape` can only do reshape, it cant do transpose or broadcast.

src/include/migraphx/algorithm.hpp
```diff
@@ -76,6 +96,53 @@ void group_unique(Iterator start, Iterator last, Output out, Predicate pred)
...
+void group_find(Iterator start, Iterator last, Predicate pred, Output out)
+{
+    start = std::find_if(start, last, pred);
+    while(start != last)
+    {
+        auto it = std::find_if_not(start, last, pred);
+        out(start, it);
+        start = std::find_if(it, last, pred);
+    }
+}
```
umangyadav, line 109: To me it seems like this is better implemented with just single `for` or `while` loop. 
e.g. 
```
auto pred_it = last;
auto neg_pred_it = last;
curr_it = start;
while(curr_it != last) {
    if(pred(curr_it) {
        pred_it = curr_it;
    } else if(pred_it != last) { // pred_it must be set 
        neg_pred_it = curr_it;
    }
    if(pred_it != last && neg_pred_it != last) {
        out_fun(pred_it, neg_pred_it)
        pred_it = last;
        neg_pred_it = last;
    }
    curr_it++;
}
```

src/include/migraphx/algorithm.hpp
```diff
@@ -76,6 +96,53 @@ void group_unique(Iterator start, Iterator last, Output out, Predicate pred)
     }
 }
 
+template <class Iterator, class Predicate, class Output>
+void group_find(Iterator start, Iterator last, Predicate pred, Output out)
```
umangyadav, line 100: is `out` here a Callable object ?

src/include/migraphx/algorithm.hpp
```diff
@@ -76,6 +96,53 @@ void group_unique(Iterator start, Iterator last, Output out, Predicate pred)
...
+    while(start != last)
+    {
+        auto it = std::find_if_not(start, last, pred);
+        out(start, it);
+        start = std::find_if(it, last, pred);
+    }
+}
+
+template <class Iterator, class Predicate>
+Iterator adjacent_remove_if(Iterator first, Iterator last, Predicate p)
```
umangyadav, line 112: Can you add docstrings for all the new additions in `algorithm.hpp` ? 
e.g. this one has poitner arithmetics making it difficult to read.

src/include/migraphx/algorithm.hpp
```diff
@@ -54,6 +55,25 @@ T transform_accumulate(Iterator first, Iterator last, T init, BinaryOp binop, Un
         first, last, first, init, binop, [&](auto&& x, auto&&) { return unaryop(x); });
 }
 
+template <class Iterator, class OutputIterator, class BinaryOperation, class UnaryOp>
+OutputIterator transform_partial_sum(
+    Iterator first, Iterator last, OutputIterator d_first, BinaryOperation binop, UnaryOp unaryop)
```
umangyadav, line 60: Add docstring. Ideally for such functions, better if you can demonstrate example usecase in docstring.

src/include/migraphx/shape_transform_descriptor.hpp
```diff
@@ -0,0 +1,112 @@
...
+//
+// For example, if we start with an initial dimensions as `[x, y, z]` then
+// each dimension will have one subdimension that corresponds to each
+// original dimension: `[[x:0]], [[y:1]], [[z:2]]`.
+//
+// When a transpose is applied we would just permutate the dimensions.
+//
+// When a reshape that would merge dimensions together then the subdimensions
+// are copied to the same subdimension. So if we reshape the dimensions as `
+// [x*y, z]` then it would become `[[x:0], [y:1]], [[z:1]]`. If the reshape
```
umangyadav, line 62:
```suggestion
// [x*y, z]` then it would become `[[x:0], [y:0]], [[z:1]]`. If the reshape
```
??

src/simplify_reshapes.cpp
```diff
@@ -1,3 +1,4 @@
+
```
umangyadav, line 1:
```suggestion
```

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,830 @@
...
+static auto compute_end_dim(Iterator start, Iterator last, std::size_t dim, Projection proj)
+{
+    std::size_t x = 1;
+    auto it       = std::find_if(start, last, [&](auto d) {
+        x *= proj(d);
+        return x >= dim;
+    });
+    if(x != dim)
+        return start;
+    return it;
```
umangyadav, line 52:
```suggestion
    std::size_t x = 1;
    auto it       = std::find_if(start, last, [&](auto d) {
        x *= proj(d);
        return x == dim;
    });
    if(it != last) 
            return it;
    return start;
```

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,830 @@
...
+}
+
+void debug_print(const std::vector<dimension::sub>& subs, bool new_line = true)
+{
+    for(const auto& s : subs)
+    {
+        std::cout << s.len << ":" << to_string_range(s.axis, "x");
+        if(s.hidden_axis.has_value())
+            std::cout << "$" << s.hidden_axis.value();
+        std::cout << ",";
```
umangyadav, line 62: For the last subdimension you probably do not want to print `,`

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,830 @@
...
+        std::cout << std::endl;
+}
+void debug_print(const dimension& dim) { debug_print(dim.subdimensions); }
+void debug_print(const std::vector<dimension>& dims)
+{
+    for(const auto& d : dims)
+    {
+        std::cout << "[";
+        debug_print(d.subdimensions, false);
+        std::cout << "],";
```
umangyadav, line 74: For the last dimension you probably do not want to print `,`

src/include/migraphx/shape_transform_descriptor.hpp
```diff
@@ -0,0 +1,112 @@
...
+// splits the dimension then the subdimension is copied to each dimension and
+// the axis is updated to maintain the order. So a reshape of `[2, x/2, y,
+// z]` would become: `[[2:0,0]], [[x/2:0,1]], [[y:1]], [[z:2]]`.
```
umangyadav, line 65: In `[[2:0, 0]]` what does second element mean `,0]]` ? 
`2:0` means subdimension with len 2 originating from axis 0.  but then what is second element ?

src/include/migraphx/output_iterator.hpp
```diff
@@ -64,6 +64,13 @@ function_output_iterator<F> make_function_output_iterator(F f)
     return {std::move(f)};
 }
 
+template <class Container>
+auto join_back_inserter(Container& c)
+{
+    return make_function_output_iterator(
+        [&](const auto& r) { c.insert(c.end(), r.begin(), r.end()); });
+}
```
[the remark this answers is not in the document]
pfultz2: Because its joining the inputs which could be a container of containers of elements to just a container of elements:

```cpp
std::vector<std::vector<int>> input;
std::vector<int> output;
std::copy(input.begin(), input.end(), join_back_inserter(output));
```

src/include/migraphx/algorithm.hpp
pfultz2: Yes just like the other group algorithms.

src/include/migraphx/algorithm.hpp
```diff
@@ -76,6 +96,53 @@ void group_unique(Iterator start, Iterator last, Output out, Predicate pred)
...
+    *first = std::move(*i);
+    ++first;
+    return first;
+}
+
+template <class Iterator, class F>
+Iterator adjacent_for_each(Iterator first, Iterator last, F f)
+{
+    if(first == last)
+        return last;
```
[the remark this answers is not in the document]
pfultz2: `std::distance` isnt always a O(1) operation, so for forward iterators this would require traversing the container twice.

src/shape_transform_descriptor.cpp
pfultz2: No that wont work when the reshape is not divisible by the dimensions. So a reshape from `[2, 3, 5]` to `[10, 3]` wont have an end dim in this case. So we return false, since this would require fractional axes which the shape_transform_descriptor doesnt support. 

Let me add a unit test for this.

src/simplify_reshapes.cpp
```diff
@@ -86,25 +88,59 @@ bool is_no_transpose(const std::vector<int64_t>& dims)
...
+              x->get_operator().name() == "contiguous")
+        {
+            ops.push_back(x->get_operator());
+            x = x->inputs().front();
+        }
+        if(x->get_shape().scalar())
+        {
+            m.replace_instruction(
+                ins, make_op("multibroadcast", {{"out_lens", ins->get_shape().lens()}}), x);
+        }
```
CharlieL7, line 132: I think this bugs out if you have a broadcasted scalar. Not sure if it's possible to get a broadcasted scalar with no shape transform instructions before it.

test/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+TEST_CASE(optimize_reshape_2_squeeze)
+{
+    EXPECT(migraphx::optimize_shape_transforms({3, 1, 5, 1, 2, 1, 1},
+                                               {
+                                                   make_op("reshape", {{"dims", {3, 5, 2}}}),
+                                               }) ==
+           ops{
+               make_op("squeeze", {{"axes", {1, 3, 5, 6}}}),
+           });
```
CharlieL7, line 300: minor: Really poor formatting by clang-format here. Probably better to add a newline to L294

test/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,389 @@
...
+    EXPECT(migraphx::optimize_shape_transforms({1},
+                                               {
+                                                   make_op("multibroadcast", {{"out_lens", {2}}}),
+                                                   make_op("unsqueeze", {{"axes", {1}}}),
+                                               }) ==
+           ops{
+               make_op("multibroadcast", {{"out_lens", {2, 1}}}),
+           });
+}
+
```
CharlieL7, line 388: We should have some tests for `flatten` also.

test/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,389 @@
...
+TEST_CASE(optimize_squeeze_broadcast)
+{
+    EXPECT(migraphx::optimize_shape_transforms(
+               {256, 1, 1},
+               {
+                   make_op("squeeze"),
+                   make_op("broadcast", {{"axis", 0}, {"out_lens", {256, 64, 1, 1}}}),
+               }) == ops{
+                         make_op("unsqueeze", {{"axes", {3}}}),
+                         make_op("multibroadcast", {{"out_lens", {256, 64, 1, 1}}}),
```
CharlieL7, line 348: It looks like this code will prioritize `unsqueeze` over `squeeze`? Is there a benefit to this or is it arbitrary?

src/include/migraphx/algorithm.hpp
pfultz2: > e.g. this one has pointer arithmetics making it difficult to read code.

It was also difficult to write which is why I added unit tests for it.

src/include/migraphx/shape_transform_descriptor.hpp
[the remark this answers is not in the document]
pfultz2: > Shouldn't this be [[x:0], [y:1]], [[z:2]]?

Yes.

> The docstring would also be clearer if we use a notation like:

Its written to match the debug print out. 

> {{len=x, axis=[0]}}, {{len=y, axis=[1]}}, {{len=z, axis=[2]}}

I think thats a little too verbose to read. Maybe I can describe the notation some more.

src/simplify_reshapes.cpp
pfultz2: If there is no shape transform instructions then this is not matched. There has to be at least two shape transforms for this to match.

test/shape_transform_descriptor.cpp
pfultz2: I dont know if its prioritization, but rather the way we interpret transforms.

This starts with `[[256:0]], [[64:]], [[1:1]], [[1:2]]`, so it first tries to insert the broadcast. The `broadcast` op wont work for this case since the "succeeding sizes must match"(which it would be nice to remove this restriction in the future), so it inserts a multibroadcast instead. 

Then it flattens it to `[[256:0]], [[1:]], [[1:1]], [[1:2]]`, which will insert an unsqueeze to create the dimension which has an empty axis.

src/include/migraphx/algorithm.hpp
pfultz2: Using `find_if` is easier to read though and it matches the pattern of how the other group algorithms work.

src/include/migraphx/shape_transform_descriptor.hpp
umangyadav: > The first elements represents the original axis, 

Original axis of the original input before applying any transformation ? 
e.g. 
let's say I have input  of shape `[2, 5, 10]` --> which would be represented as  `[[2:0]], [[5:1]], [[10:2]]` in current scheme. 
Now, Applying first reshape it becomes `[2, 5, 5, 2]` --> which would become `[[2:0]], [[5:1]], [[5:2, 2]], [[2:2, 2]]` 
Now let's apply second reshape and make it `[2, 25, 2]` --> 

1. Original  axis here should it taken from original input `[2, 5, 10]` ? or 
In this case, output dimension with len 25 has two parent axes. 
2. Should it be considered from Input to the second reshape which is `[2, 5, 5, 2]`?

Or to frame question differently, 
what is the definition of original axis ?

umangyadav: > but each additional element represents where the axis is split. 

To me it seems like size of `subdimension.axis` can at max be 2 and min 1. 

for the case when size is two. Second element would denote axis in the output of the operation. 

For example
> So a reshape of `[2, x/2, y,
// z]` would become: `[[2:0,0]], [[x/2:0,1]], [[y:1]], [[z:2]]`. 

here `[[x/2:0, 1]]` mean 1 is axis in the output for that dimension.   We can instead make the axis always of length 2. where first element would be parent axis and second element would be axis in the output. 
`[[2:0,0]], [[x/2:0,1]], [[y:1, 2]], [[z:2, 3]]`

> So if we reshape the dimensions as `
// [x*y, z]` then it would become `[[x:0], [y:1]], [[z:2]]`.

E.g. in this case it would be like `[[x:0, 0], [y:1, 0]], [[z:2, 1]]`

src/include/migraphx/shape_transform_descriptor.hpp
```diff
@@ -0,0 +1,112 @@
...
+            // length greater then 1, so it cant be directly associated with
+            // the axis. However, it still needs to accounted for. After we
+            // generate the broadcast we will set the axis to the hidden
+            // axis, and then length to 1.
+            optional<std::size_t> hidden_axis = nullopt;
+        };
+        std::vector<sub> subdimensions;
+    };
+    std::vector<dimension> dimensions;
+    std::size_t rank = 0;
```
umangyadav, line 104: Is this rank, rank of the original input before any reshape ops are applied ? or is it the rank before EACH reshape operation is applied ?

src/include/migraphx/shape_transform_descriptor.hpp
```diff
@@ -0,0 +1,112 @@
...
+// [x*y, z]` then it would become `[[x:0], [y:1]], [[z:2]]`. If the reshape
+// splits the dimension then the subdimension is copied to each dimension and
+// the axis is updated to maintain the order. So a reshape of `[2, x/2, y,
+// z]` would become: `[[2:0,0]], [[x/2:0,1]], [[y:1]], [[z:2]]`.
+//
+// After recording the operators, `simplify` method is used to simplify the
+// data structure such as merging adjacent dimension, etc. The `generate`
+// method is called to generate the operators need to do this
+// transformation.
+struct shape_transform_descriptor
```
umangyadav, line 71: Many of my questions are stemming from confusion if this `shape_transform_descriptor` is meant to represent only ONE operation or series of operation ?

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,830 @@
...
+        }
+        // squeeze
+        else if(rdim > idim)
+        {
+            auto start = subs.begin() + i;
+            auto it = compute_end_dim(start, subs.end(), rdim, std::mem_fn(&dimension::sub::len));
+            if(it == start)
+                return false;
+            auto n = it - start;
+            i += n;
```
umangyadav, line 183:
```suggestion
            auto n = it - start +1;
            i += n;
```

umangyadav: I see you are doing `i++` and `r++` at the bottom to account for additional 1. I find it better to it here.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,830 @@
...
+    {
+        for(auto d : range(rdims.begin() + new_dims.size(), rdims.end()))
+        {
+            if(d != 1)
+                return false;
+            new_dims.push_back({{dimension::sub{1}}});
+        }
+    }
+
+    if(rdims.size() != new_dims.size())
```
umangyadav, line 218: assert would suffice here since for loop earlier would make sure both are of same size.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,830 @@
...
+                           MIGRAPHX_THROW("Wrong out_lens for broadcast");
+                       return {{dimension::sub{len, {}}}};
+                   });
+    std::transform(out_lens.begin() + offset + dimensions.size(),
+                   out_lens.end(),
+                   std::back_inserter(new_dims),
+                   [&](auto len) -> dimension {
+                       return {{dimension::sub{len, {}}}};
+                   });
+    dimensions = new_dims;
```
umangyadav, line 260: assert `out_lens.size() == new_dims.size()`

src/include/migraphx/shape_transform_descriptor.hpp
pfultz2: > Now, Applying first reshape it becomes [2, 5, 5, 2] --> which would become [[2:0]], [[5:1]], [[5:2, 2]], [[2:2, 2]]

No, it becomes `[[2:0]], [[5:1]], [[5:2, 0]], [[2:2, 1]]`.

> Now let's apply second reshape and make it [2, 25, 2] 

So this would become `[[2:0]], [[5:1], [5:2, 0]], [[2:2, 1]]`.

> Original axis here should it taken from original input [2, 5, 10] ?

Yes in fact all the axes are from the original input we just add extra numbers to track the splitting of the axis.

> To me it seems like size of subdimension.axis can at max be 2 and min 1.

No that isnt the case. It can also be empty for when its unsqueezed and broadcasted.

It is a max of 2 only after `simplify` because it renumbers all axes, but before that it can be more than 2.

For example, if we have `[[12:0]]` that we reshape to `[2, 6]` so it would become `[[2:0, 0], [6:0, 1]]`. And then if there is another reshape to `[2, 2, 3]` then this will become `[[2:0, 0], [2:0, 1, 0], [3:0, 1, 1]]`.  

> here [[x/2:0, 1]] mean 1 is axis in the output for that dimension.

No, its just used to track the order of the dimensions after the axis is split by a reshape. You can see the logic here:

https://github.com/ROCm/AMDMIGraphX/blob/9a43b2bed0117080cfcf6ca29603417d7a976da4/src/shape_transform_descriptor.cpp#L198

We `push_back` the numbers from the range (0, n], where n is the number of splits that reshape is doing.

src/include/migraphx/shape_transform_descriptor.hpp
pfultz2: Its the rank of the original input.

src/include/migraphx/shape_transform_descriptor.hpp
pfultz2: Its meant to represent a series of operations.

src/shape_transform_descriptor.cpp
pfultz2: But then I would need to increment the other variables in all branches:

```cpp
        if(idim == rdim)
        {
            ...
            i++;
            r++;
        }
        // squeeze
        else if(rdim > idim)
        {
            ...
            auto n = it - start + 1;
            i += n;
            r++;
            ...
        }
        // unsqueeze
        else // if(rdim < idim)
        {
            ...
            auto n = it - start + 1;
            r += n;
            i++;
            ...
        }
```

It seems simpler to increment them at the end together. This will ensure that I dont forget to increment one of them in one of the branches.

src/shape_transform_descriptor.cpp
pfultz2: I can add an assert, but I still would like this branch so if something goes wrong in a release version it will just bail.

pfultz2:
The unit test failure will most likely be fixed by #3188.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,835 @@
...
+        if(d1.axis.size() < 2)
+            return;
+        if(d2.axis.size() < 2)
+            return;
+        if(not std::equal(d1.axis.begin(), d1.axis.end() - 1, d2.axis.begin(), d2.axis.end() - 1))
+            return;
+        auto a1 = d1.axis.back();
+        auto a2 = d2.axis.back();
+        if(a2 <= a1)
+            return;
```
umangyadav, line 290: This strictly needs to be `a2 < a1` if `a2 == a1` then both subdimensions have exactly the same values for all  axes which seems unlikely/invalid and  i think it should throw.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,835 @@
...
+        last_axis = std::prev(axes_map.end())->second.back()->axis;
+
+        // Find missing axes. This will store a mapping between the missing
+        // axis and the next available axis.
+        for(auto axis : range(rank))
+        {
+            if(contains(axes_map, axis))
+                continue;
+            auto it            = axes_map.upper_bound(axis);
+            missing_axes[axis] = it == axes_map.end() ? rank : it->first;
```
umangyadav, line 455:
```suggestion
            missing_axes[axis] = it == axes_map.end() ? rank -1  : it->first;
```
Rank `r` would have axes from `[0, 1, .. r-1]`. Therefore i think it should be `rank -1`

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,835 @@
...
+        return false;
+    const auto& sub = d.subdimensions.front();
+    return sub.axis.empty();
+}
+static bool missing_leading_axis(const dimension& d)
+{
+    if(d.subdimensions.empty())
+        return true;
+    const auto& sub = d.subdimensions.front();
+    return sub.axis.empty();
```
umangyadav, line 345: Why is it only checking if the first subdimension is empty ? 
Broadcasted subdimension could be anywhere in the list.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,835 @@
...
+        // squeeze
+        else if(rdim > idim)
+        {
+            auto start = subs.begin() + i;
+            auto it = compute_end_dim(start, subs.end(), rdim, std::mem_fn(&dimension::sub::len));
+            if(it == start)
+                return false;
+            auto n = it - start;
+            i += n;
+            new_dims.push_back({{start, it + 1}});
```
CharlieL7, line 184: I don't think this is right. `it` should point to the end dimension already? Somehow it works right now. Using LLDB it seems that this code is optimized out. But when I put in a new test and debug statements I get out of range segmentation faults.

Here's the test I added: 
```
TEST_CASE(record_reshape_merge)
{
    auto desc = make_descriptor({3, 4, 5}, make_op("reshape", {{"dims", {3, 20}}}));
    EXPECT(get_final_lens(desc) == final_lens{3, 20});
    EXPECT(get_all_lens(desc) == all_lens{{3}, {4, 5}});
    EXPECT(get_all_axes(desc) ==
           all_axes{d_axes{{0}}, d_axes{{1}, {2}}});
}
```

I added these debug statements:
```
            auto n = it - start;
            std::cout << "n: " << n << "\n";
            std::cout << "i before: " << i << "\n";
            i += n;
            std::cout << "i after: " << i << "\n";
            std::cout << "it + 1 - start: " << it + 1 - start << "\n";
            std::cout << "(it + 1)->len" << (it + 1)->len << "\n";
            std::cout << "(it + 1)->axis.front()" << (it + 1)->axis.front() << "\n";
            new_dims.push_back({{start, it + 1}});
```

CharlieL7: Figured it out, using begin to end iterators

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,835 @@
...
+void remove_1_sub_dims(std::vector<dimension::sub>& subdimensions)
+{
+    subdimensions.erase(std::remove_if(subdimensions.begin(),
+                                       subdimensions.end(),
+                                       [&](const dimension::sub& d) { return d.len == 1; }),
+                        subdimensions.end());
+}
+
+void dimension::simplify()
+{
```
CharlieL7, line 275: Should have unit tests for this

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+        {
+            subs[0]->axis = {axis};
+        }
+        else
+        {
+            std::sort(subs.begin(), subs.end(), by(std::less<>{}, [](const dimension::sub* s) {
+                          return s->axis;
+                      }));
+            for(std::size_t i : range(subs.size()))
+                subs[i]->axis = {axis, i};
```
shivadbhavsar, line 392: how does this work for the multiple split case (ie. same dim got split multiple times)? Eg. you have 3 subdims like 
`[[len=x, axis={0,2,0}]], [[len=y, axis={0,2,1}]], [[len=z, axis={0,2,2}]]`
and then simplify is able to simplify 2 of them st. you have remaining:
`[[len=x*y, axis={0,2,1}]], [[len=z, axis={0,2,2}]]`
If I understand correctly, renumbering should make those axis {0,2,0} and {0,2,1} right?

shivadbhavsar: Further along the same train of thought, if simplify finds that all of them are adjacent and gives
`[[len=x*y*z, axis={0,2,2}]]`
Then is there a way somewhere to remove that last value from the axis vector entirely?

src/shape_transform_descriptor.cpp
pfultz2: It using this to group the broadcasted dimensions(using the `group_find` algorithm). So the one-past-the-end should have an axis on the first subdimension so it can figure out where to place the dimensions of 1 which could be broadcasted.

It is true that within the group it may include dimensions that start with a broadcasted subdimension but may include other dimensions so we probably need to filter it out when we construct the `std::deque`.

src/shape_transform_descriptor.cpp
pfultz2: >  and then simplify is able to simplify 2 of them st. you have remaining:

Simplify is able to simplify all three of them into the same dim, which will be `{0, 2, 2}`. 

>  If I understand correctly, renumbering should make those axis {0,2,0} and {0,2,1} right?

Its kind of missing what the other axes are, but lets assume the other axes couldn't be simplified and are: `{0, 0, 0}`, `{0, 0, 1}`, `{0, 1, 0}`, `{0, 1, 1}`. So renumbering will change the all axes to:

* `{0, 0, 0}` -> `{0, 0}`
* `{0, 0, 1}` -> `{0, 1}`
* `{0, 1, 0}` -> `{0, 2}`
* `{0, 1, 1}` -> `{0, 3}`
* `{0, 2, 2}` -> `{0, 4}`

pfultz2: >  Then is there a way somewhere to remove that last value from the axis vector entirely?

If all them got simplified then it will just set the axis to `{0}`. Thats what the condition above does(ie under `if(subs.size() == 1)`).

src/shape_transform_descriptor.cpp
umangyadav: I do not see why proposed change will not work. Logically it would also return false for the case you mentioned  `[2, 3, 5]` to `[10, 3]`

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+static bool is_broadcast_dim(const dimension& d)
+{
+    if(d.len() == 1)
+        return false;
+    assert(not d.subdimensions.empty());
+    if(d.subdimensions.size() != 1)
+        return false;
+    const auto& sub = d.subdimensions.front();
+    return sub.axis.empty();
+}
```
umangyadav, line 340:
```suggestion
}

```

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+
+void dimension::simplify()
+{
+    if(subdimensions.size() < 2)
+        return;
+    remove_1_sub_dims(subdimensions);
+    // Flatten adjacent dimensions
+    adjacent_for_each(subdimensions.begin(), subdimensions.end(), [&](sub& d1, sub& d2) {
+        if(d1.axis.size() < 2)
```
umangyadav, line 282: For this to work, i think subdimensions need to be in sorted order based on axes.  I do not see it being sorted before calling to `simplify()`

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+    {
+        // Group axes
+        auto axes_map = group_axes(dimensions);
+        if(axes_map.empty())
+            return;
+
+        renumber_axes(axes_map);
+
+        // Find last axis
+        last_axis = std::prev(axes_map.end())->second.back()->axis;
```
umangyadav, line 447:
```suggestion
        // Find last axis
        last_axis = std::prev(axes_map.end())->first;
```

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+            if(last != dimensions.end())
+            {
+                assert(not last->subdimensions.empty());
+                const auto& sub = last->subdimensions.front();
+                assert(not sub.axis.empty());
+                axis = sub.axis.front();
+            }
+            std::deque<std::size_t> dims(std::distance(start, last));
+            std::iota(dims.begin(), dims.end(), std::distance(dimensions.begin(), start));
+            broadcast_dims_map[axis] = dims;
```
umangyadav, line 475: Would this work in case of transposes ? When there are transposes, dimensions would not be in sorted order based on original axis.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+            }
+            else
+            {
+                auto next = std::find_if(std::next(it), sub->end(), [&](const dimension::sub& s) {
+                    if(s.len != 1)
+                        return true;
+                    if(s.axis.empty())
+                        return true;
+                    return s.axis.front() > missing_axis;
+                });
```
umangyadav, line 508: So the `it` is pointer to subdimension with `last_axis`. 
and
```
 // Find last axis
        last_axis = std::prev(axes_map.end())->second.back()->axis;
```

`last_axis` is axis after `renumbering` whereas here `std::find()` is finding it over dimensions without re-numbering. 
How does that work out ?

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+            // Search for the subdimension that has the next axis and try to
+            // insert the axis before it will be in order.
+            auto [sub, it, prev] = find_subdimension(*this, [&](const dimension::sub& s) {
+                if(s.axis.empty())
+                    return false;
+                if(s.axis.front() != next_axis)
+                    return false;
+                if(s.axis.size() == 1)
+                    return true;
+                assert(s.axis.size() == 2);
```
umangyadav, line 523: what's the reasoning behind this assert ? Why does it have to be of size 2 ?

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+                return s.axis.back() == 0;
+            });
+            bool in_order        = false;
+            if(prev.has_value() and not(*prev)->axis.empty())
```
umangyadav, line 527:
```suggestion
            if(prev.has_value() and not (*prev)->axis.empty())
```

src/shape_transform_descriptor.cpp
umangyadav: I see `renumbering` is renumbering all subdimensions because it is using pointer to those.

src/shape_transform_descriptor.cpp
pfultz2: They dont need to be sorted. This is checking adjacent axes in a subdimension and making the first one a len of 1(which will be removed). So `{0, 3}, {0, 1}, {0, 2}` becomes `{0, 3}, {0, 2}`(later these will be renumbered to `{0, 2}, {0, 1}`.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,836 @@
...
+            return;
+
+        renumber_axes(axes_map);
+
+        // Find last axis
+        last_axis = std::prev(axes_map.end())->second.back()->axis;
+
+        // Find missing axes. This will store a mapping between the missing
+        // axis and the next available axis.
+        for(auto axis : range(rank))
```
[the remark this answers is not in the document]
pfultz2: No, this used because we remove suddimensions of len 1 in  the `simplify` method, which can remove an entire axis, so we need to find the axes that are missing so we can place them somewhere.

src/shape_transform_descriptor.cpp
umangyadav: It is checking if axes in adjacent subdimension differ by 1 at the last position.   For that check to work in all cases, i think it needs to be sorted.

src/shape_transform_descriptor.cpp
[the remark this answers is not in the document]
umangyadav: True,  but let's say there was no subdimension with len 1 and `ouput_rank > input_rank` in that case `missing_axes` would be empty.

src/shape_transform_descriptor.cpp
pfultz2: > Why does it have to be of size 2 ?

We are only comparing the last element in the axis.

After renumbering the axis size should not be greater than 2 but the assert is there just in case this changes(which would make this logic incorrect).

src/shape_transform_descriptor.cpp
pfultz2: I see, if its last you return `start`. This is different from the original which could return `last` from `find_if`, but looking how its used this would be an error because we do `it + 1` to get the last iterator. I am not sure how we would hit such a case since we check the number of elements are the same.

src/shape_transform_descriptor.cpp
pfultz2: > For that check to work in all cases, i think it needs to be sorted.

In what case would it not work?

src/shape_transform_descriptor.cpp
pfultz2: I'll add an assert for it.

src/shape_transform_descriptor.cpp
pfultz2: This is meant to store the index of the dimension. Let me update the comment to make that clearer.

src/shape_transform_descriptor.cpp
pfultz2: Yea clang-format doesn't allow that.

src/shape_transform_descriptor.cpp
umangyadav: e.g. 
`[[0, 0, 1], [0, 0, 3], [0, 0, 2]]`

pfultz2: But how would that not work? Those axes can't be removed.

src/shape_transform_descriptor.cpp
```diff
@@ -0,0 +1,840 @@
...
+//
+// 1. Reshape/unsqueeze
+// 2. Transpose
+// 3. Broadcast
+// 4. Reshape/squeeze
+// 5. Broadcast
+//
+// This will generate operators backwrds starting at 5 and going up. Steps 1-3
+// are generated from the subdimesions and steps 4-5 are generated with the
+// dimensions.
```
umangyadav, line 706:
```suggestion
// This will generate operators backwards starting at 5 and going up. Steps 1-3
// are generated from the subdimensions and steps 4-5 are generated with the
// dimensions.
```

shivadbhavsar:
There are some models that are failing in torch benchmarks with this PR. I am saving off these mxrs in our nas and I'll list them here. 
`/mnt/nas_share/migraphx/models/torch_benchmarks/paul_shape_transform_pr`

- crossvit_9_240.mxr

pfultz2:
> There are some models that are failing in torch benchmarks with this PR. I am saving off these mxrs in our nas and I'll list them here. `/mnt/nas_share/migraphx/models/torch_benchmarks/paul_shape_transform_pr`
> 
>     * crossvit_9_240
> 
>     * fbnetc_100
> 
>     * mnasnet_100
> 
>     * mobilenetv2_100
> 
>     * mobilenetv3_large_100
> 
>     * mobilevit_s
> 
>     * spnasnet_100
> 
>     * tf_efficientnet_b0
> 
>     * tinynet_a
> 
> 
> I only saw 2 unique errors so most likely they are all failing due to the same bug(s)

These error happen in simplify_algebra. They look like bugs there and not related to this PR, but this PR probably rewrites it in a way to expose this bug. Probably should open separate tickets to fix those issue. These are the snippets I got that might help describe the error and which class this happens in:

```
find_concat_op:
@509 = unsqueeze[axes={4},steps={}](@505) -> float_type, {128, 1, 1, 1, 1}, {1, 1, 1, 1, 1}
@511 = unsqueeze[axes={4},steps={}](@506) -> float_type, {1, 3, 1, 1, 1}, {3, 1, 1, 1, 1}
@513 = unsqueeze[axes={4},steps={}](@507) -> float_type, {1, 1, 224, 1, 1}, {224, 224, 1, 1, 1}
@515 = unsqueeze[axes={4},steps={}](@508) -> float_type, {1, 1, 1, 224, 1}, {224, 224, 224, 1, 1}
@510 = multibroadcast[out_lens={128, 3, 224, 224, 1},out_dyn_dims={}](@509) -> float_type, {128, 3, 224, 224, 1}, {1, 0, 0, 0, 1}
@512 = multibroadcast[out_lens={128, 3, 224, 224, 1},out_dyn_dims={}](@511) -> float_type, {128, 3, 224, 224, 1}, {0, 1, 0, 0, 1}
@514 = multibroadcast[out_lens={128, 3, 224, 224, 1},out_dyn_dims={}](@513) -> float_type, {128, 3, 224, 224, 1}, {0, 0, 1, 0, 1}
@516 = multibroadcast[out_lens={128, 3, 224, 224, 1},out_dyn_dims={}](@515) -> float_type, {128, 3, 224, 224, 1}, {0, 0, 0, 1, 1}
@517 = concat[axis=4](@510,@512,@514,@516) -> float_type, {128, 3, 224, 224, 4}, {602112, 200704, 896, 4, 1}

find_splits:
@760 = convolution[padding={0, 0, 0, 0},stride={1, 1},dilation={1, 1},group=1,padding_mode=0](@759,@758) -> half_type, {128, 384, 28, 28}, {301056, 784, 28, 1}
@762 = convolution[padding={0, 0, 0, 0},stride={1, 1},dilation={1, 1},group=1,padding_mode=0](@757,@287) -> half_type, {128, 192, 28, 28}, {784, 100352, 28, 1}
@763 = slice[axes={1},starts={0},ends={192}](@760) -> half_type, {128, 192, 28, 28}, {301056, 784, 28, 1}
@764 = add(@762,@763) -> half_type, {128, 192, 28, 28}, {784, 100352, 28, 1}
@814 = slice[axes={1},starts={192},ends={384}](@760) -> half_type, {128, 192, 28, 28}, {301056, 784, 28, 1}
@815 = convolution[padding={0, 0, 0, 0},stride={1, 1},dilation={1, 1},group=1,padding_mode=0](@809,@282) -> half_type, {128, 192, 28, 28}, {150528, 784, 28, 1}
@816 = add(@814,@815) -> half_type, {128, 192, 28, 28}, {150528, 784, 28, 1}
```

causten (changes requested):
Needs a follow up from Shiv

src/shape_transform_descriptor.cpp
umangyadav: Resolved offline.

subdimensions doesn't need to be sorted. Need to maintain memory layout and therefore only remove ones that are adjacent in memory.

causten (approved):

umangyadav (approved):
Looks good to for most of the parts. good to me for the most of the parts. Approving to not block any improvements for 6.3.

CharlieL7 (approved):

causten:
@pfultz2  need to fix the CI failures

umangyadav:
@pfultz2  windows build is failing. Please take a look

pfultz2:
Windows CI is passing now.

PR #3106, opened 2024-05-21
Enable splitk with convolution

pfultz2:


src/targets/gpu/jit/mlir.cpp
```diff
@@ -77,7 +77,7 @@ struct mlir_compiler : compiler<mlir_compiler>
         auto* smod = ins->module_inputs().front();
         assert(smod->get_parameter_names().size() == ins->inputs().size() - 1);
         auto gemm_ins = std::find_if(smod->begin(), smod->end(), [&](const auto& i) {
-            return i.name() == "dot" or i.name() == "quant_dot";
+            return contains({"dot", "quant_dot", "convolution", "quant_convolution"}, i.name());
```
umangyadav, line 80: Is there a test for this one ? 

If not can you add one please  ? E.g. I'd added one for the GEMMs in this PR https://github.com/ROCm/AMDMIGraphX/pull/3082/files

src/targets/gpu/jit/mlir.cpp
```diff
@@ -77,7 +77,7 @@ struct mlir_compiler : compiler<mlir_compiler>
         auto* smod = ins->module_inputs().front();
         assert(smod->get_parameter_names().size() == ins->inputs().size() - 1);
         auto gemm_ins = std::find_if(smod->begin(), smod->end(), [&](const auto& i) {
```
umangyadav, line 79:
```suggestion
        auto gemm_based_ins = std::find_if(smod->begin(), smod->end(), [&](const auto& i) {
```

src/targets/gpu/jit/mlir.cpp
```diff
@@ -77,7 +77,7 @@ struct mlir_compiler : compiler<mlir_compiler>
...
-            return i.name() == "dot" or i.name() == "quant_dot";
+            return contains({"dot", "quant_dot", "convolution", "quant_convolution"}, i.name());
         });
         // check if (a) module is fused (b) contains a dot instruction and (c) perfConfig can not
```
manupak, line 82: This comment is outdated now.

src/targets/gpu/jit/mlir.cpp
```diff
@@ -77,7 +77,7 @@ struct mlir_compiler : compiler<mlir_compiler>
...
         // allow fused module
```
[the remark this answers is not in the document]
umangyadav: https://github.com/ROCm/AMDMIGraphX/pull/3022#discussion_r1597211056  

See this discussion

test/verify/test_conv_add_tune.cpp
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+#include <migraphx/op/common.hpp>
+
+template <migraphx::shape::type_t DType>
+struct test_conv_add_tune : verify_program<test_conv_add_tune<DType>>
```
umangyadav, line 32: @pfultz2  added test. 
Sizes seem large but i had to pick sizes for the convolution that would offload to MLIR by default and would pick split-k perf config as the fastest one. 

This test uses Split-k perf config for the convolution on MI100 . 

Without fix from this PR, this test fails.

umangyadav (approved):

test/verify/test_conv_add_tune.cpp
umangyadav: It is not guranteed to pick split-k perf config all the time though. So this test may be dud for those cases. On MI250, i do not see it picking split-k config as the fastest one.

causten (approved):

PR #3111, opened 2024-05-22
Fix hipExtModuleLaunchKernel function

slojosic-amd:
hipExtModuleLaunchKernel function name should not be mangled in case of cross-platform compilation. Mangled hipExtModuleLaunchKernel function name causes linking issues with migraphx_gpu.dll on Windows.
image


pfultz2 (approved):

umangyadav (approved):

causten:
FYI you don't need to merge develop multiple times in a row.

PR #3113, opened 2024-05-23
Fuse broadcasts with pointwise after splitting the reduction

pfultz2:


src/fuse_pointwise.cpp
```diff
@@ -200,8 +200,43 @@ struct pointwise_reshape : rewrite_reshapes_base
     static std::string name() { return "pointwise"; }
 };
 
+struct pointwise_broadcast_pointwise
```
umangyadav, line 203: add unit-tests

src/fuse_pointwise.cpp
```diff
@@ -200,8 +200,43 @@ struct pointwise_reshape : rewrite_reshapes_base
...
+        auto x_ins         = r.instructions["x"];
+
+        auto broadcast = broadcast_ins->get_operator();
+
+        auto x_inputs = x_ins->inputs();
+        std::transform(x_inputs.begin(), x_inputs.end(), x_inputs.begin(), [&](auto input) {
+            return m.insert_instruction(broadcast_ins, broadcast, input);
+        });
+
+        m.replace_instruction(
```
umangyadav, line 227: How does it help to broadcast all the inputs ? 

Is it guranteed that all the inputs are broadcastable ?

pfultz2: > How does it help to broadcast all the inputs ?

So the output of the operator will have the same dimensions as the original broadcasted instruction which feeds into another pointwise(which all have the same dimesions).

> Is it guranteed that all the inputs are broadcastable ?

Yes, the inputs and outputs are the same dimensions for the pointwise operator so if we can broadcast the output then we can broadcast all the inputs.

src/fuse_pointwise.cpp
pfultz2: Unit test added.

umangyadav (approved):

Issue #3033, opened 2024-05-02
benchmark: No configs to tune


shivadbhavsar:

bpickrel:
Note to self:  "No configs to tune" really means that an exception was thrown during benchmarking, in `insert_compiles()` in `compile_ops.cpp` such that the array named `compiles` didn't get populated.  Exception messages during the run included
```
@6Pass: simplify_algebra
x0
, {1}, 0({} -> @00{ -> ]{}, { -> 1, 1, 1, 1, 1{{0half_type}@param@8, x0 -> 0float_type, {()}Exception in fused_reduce: /workspace/pt2_benchmarks/AMDMIGraphX/src/targets/gpu/compile_gen.cpp:393: operator(): Unknown operator: hip::allocate -> half_type, {1}, {0}
@return@3 = mul(@2,@2) -> half_type, {1}, {0}
```
and
```
@18 = gpu::code_object[code_object=4920,symbol_name=div_kernel,global=1204224,local=1024,](@13,@17,output) -> half_type, {64, 1, 768, 7, 7}, {37632, 37632, 49, 7, 1}

Exception in fused_reduce: /workspace/pt2_benchmarks/AMDMIGraphX/src/targets/gpu/compile_gen.cpp:393: operator(): Unknown operator: hip::allocate
x2 = @param:x2 -> half_type, {64, 1, 96, 56, 56}, {301056, 301056, 3136, 56, 1}

```

bpickrel:
The "No configs to tune" error message was a red herring.  Further investigation traced the bug to a `fused_reduce` instruction that was created with an invalid arguments list during fusion of a module that should not have been fused, and later fails a validation assertion if you run it in debug mode.  Not calling `set_bypass()` in the first place was an oversight.

shivadbhavsar:
Seems like we are running into this error for another model in the latest run. I saved it in the nas at `/migraphx/models/torch_benchmarks/gmixer_24_224/gmixer_24_224.mxr`.
@bpickrel  can you do a quick check if this compiles successfully with your bypass change? If not, I'll create a new issue.

bpickrel:
[AMD Official Use Only - AMD Internal Distribution Only]

will do right now.

PR #3116, opened 2024-05-23
added set_bypass() call

bpickrel:
Bug fix:  added a  `set_bypass()` call that was missing from the `reduce_reshape` variant of the `rewrite_reshapes` pass in fusion reduction (the other variants already have it).  This sets a flag to prevent the created module from being fused again (i.e. subsequent compiler passes that see the flag skip over the module), which could lead to an exception because it would try to create a `fused_reduce` instruction with an invalid argument list.  Closes Issue #3033


pfultz2 (approved):
This looks good.

pfultz2:
For the other reviewers, adding the extra check in fused_reduce operator will get coverage for this issue with the current unit tests, so there is no need to write a new unit test.

Also its very likely such unit test would be removed in the future, which I dont have time to explain now.

bpickrel:
Don't be too quick!  The change broke an existing test, which I'm working on now.

umangyadav (approved):

lakhinderwalia (approved):

PR #3124, opened 2024-05-27
Add remove MIOpen build option

tvukovic-amd:
- Use MIGRAPHX_USE_MIOPEN ON/OFF to keep/exclude MIOpen in MIGraphX (by default is ON)
- Modified following tests (which additionally fail when MIGRAPHX_USE_MIOPEN is OFF):
1.  test_gpu_fuse_ops
2. test_api_gpu
3. test_verify_conv
4. test_verify_general
(add condition if MIOpen is included to execute them because without MIOpen there is no gpu::convolution, gpu::pooling and gpu::lrn supported)
- Note: test_gpu_adjust_allocation fails when MIOpen is excluded - reference and compiler outputs are different (they are the same when MIOpen is used)

@apwojcik


causten:
Looks like there is a cppcheck failure

apwojcik:
@causten The cppcheck is failing on the line that has been moved a step down due to inserting the `#ifdef` preprocessor macro. The original code was not edited. Fixing the cppcheck will result in additional changes not related to this PR. Do you think we should proceed with those changes, too?

apwojcik:
> Looks like there is a cppcheck failure

 @causten, the cppcheck failure is related to the line that was moved one line down by adding `#ifdef` before. The line itself has stayed the same. We have investigated that, and fixing the cppcheck will increase the number of changed files, and those changes are not related to the topic PR is addressing.

causten:
The Format check found a problem.  Please resolve

src/targets/gpu/lowering.cpp
```diff
@@ -87,24 +87,26 @@ struct miopen_apply
...
         add_extend_op("rnn_var_sl_shift_sequence");
         add_extend_op("topk");
 
+#if MIGRAPHX_USE_MIOPEN
         add_convolution_op("convolution");
         add_convolution_op("convolution_backwards");
         add_convolution_op("quant_convolution");
+        add_extend_op("lrn");
+        add_extend_op("pooling");
+        add_generic_op("contiguous");
```
pfultz2, line 108: Contiguous is not an miopen operator it shouldn't be listed here.

src/targets/gpu/jit/roialign.cpp
```diff
@@ -25,6 +25,9 @@
 #include <migraphx/gpu/compile_hip_code_object.hpp>
 #include <migraphx/gpu/context.hpp>
 #include <migraphx/gpu/compile_hip.hpp>
+#if MIGRAPHX_USE_MIOPEN == 0
+#include <migraphx/op/pooling.hpp>
```
pfultz2, line 29: Why is this conditionally added here? Its needed for the enum used below.

test/api/test_gpu.cpp
```diff
@@ -28,6 +28,8 @@
 #include <migraphx/manage_ptr.hpp>
 #include "test.hpp"
 
+// gpu::convolution not supported since MIOpen is OFF
+#if MIGRAPHX_USE_MIOPEN
```
pfultz2, line 32: Is it necessary to disable the tests? Wont it just fallback to cpu?

test/verify/main.cpp
```diff
@@ -63,40 +63,58 @@ void validate_gpu(const migraphx::program& p, const migraphx::parameter_map& m)
     EXPECT(is_shared(ctx, p.get_context()));
 }
 
+// clang-format off
```
apwojcik, line 66: We had to turn clang-format off here because of the differences in versions used between Windows and Linux, hip sdk, and dockers — it was impossible to satisfy all of them!

apwojcik: formated manually to look clean

test/api/test_gpu.cpp
tvukovic-amd: I think that this should be kept since there is loading conv_relu_maxpool_test.onnx which has pooling layer (gpu::pooling without miopen not merged yet to develop)

src/targets/gpu/include/migraphx/gpu/abs.hpp
```diff
@@ -56,7 +57,7 @@ struct miopen_abs
         return shapes.size() - 1;
     }
 };
-
+#endif
```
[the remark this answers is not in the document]
tvukovic-amd: I removed operator, but when MIGRAPHX_USE_MIOPEN is OFF, if you think that I should remove it in all cases, I can change it.

test/api/test_gpu.cpp
pfultz2: This makes it harder to figure out how to re-enable the pooling tests as they are all scattered across. It would be better to just have a cpu fallback and just keep all the tests enabled.

tvukovic-amd: Do you know how we can make GPU test fallback to CPU. Wouldn't it be misleading for results interpretation since those are GPU tests?

pfultz2: We fallback to cpu version for nonmaxsuppression. We do this in lowering so you can follow how it was done for that operator to do something similar for pooling:

https://github.com/ROCm/AMDMIGraphX/blob/c6fc4c18f7ac63f1af5916ad2e8e53c06a656c59/src/targets/gpu/lowering.cpp#L358

tvukovic-amd: Pushed changes to fallback pooling, lrn and backwards convolution to CPU and undo changes to skip tests when MIOpen is disabled

src/targets/gpu/lowering.cpp
```diff
@@ -374,6 +381,66 @@ struct miopen_apply
...
+                    return mod->insert_instruction(ins, make_op("hip::copy_from_gpu"), in);
+                });
+            cpu_inputs.front() =
+                mod->insert_instruction(ins, make_op("hip::sync_stream"), cpu_inputs);
+            auto cpu_out = mod->insert_instruction(ins, ins->get_operator(), cpu_inputs);
+            auto gpu_out =
+                mod->insert_instruction(ins, make_op("hip::copy_to_gpu"), cpu_out, output);
+            return mod->replace_instruction(ins, gpu_out);
+        });
+    }
```
pfultz2, line 402: This shouldn't be needed now that we have a GPU pooling.

apwojcik (approved):

PR #3129, opened 2024-05-29
Add ONNX parsing for SimplifiedLayerNormalization

turneram:


CharlieL7:
Where is this spec for this operator?

src/onnx/parse_simplified_layer_normalization.cpp
```diff
@@ -0,0 +1,98 @@
...
+        if(x_rank < 2 or x_rank > 3)
+        {
+            MIGRAPHX_THROW("PARSE_SIMPLIFIED_LAYER_NORMALIZATION: invalid input shape");
+        }
+
+        auto x_sq = info.add_common_op("mul", x, x);
+        auto rms  = info.add_instruction(make_op("reduce_mean", {{"axes", {axis}}}), x_sq);
+        auto mean = rms;
+        epsilon =
+            (x_dtype == migraphx::shape::half_type and std::abs(epsilon) < 1e-7) ? 1e-7 : epsilon;
```
CharlieL7, line 85: Why are we limiting the epsilon for half type? It looks like a user input

CharlieL7 (commented):
Looks fine, minor comments. I would like to see the equation that it's supposed to be doing.

turneram:
> Where is this spec for this operator?

There isn't actually one for SimplifiedLayerNormalization, but [this](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftskipsimplifiedlayernormalization) is the spec for SkipSimplifiedLayerNormalization, which is just add + SLN. That spec does include an optional bias input, but neither of the ORT implementations utilize it, so I omitted it from ours.

turneram:
> Looks fine, minor comments. I would like to see the equation that it's supposed to be doing.

The equation is the same as [RMS LayerNorm](https://arxiv.org/abs/1910.07467).

test/onnx/verify/simplified_layer_normalization.cpp
```diff
@@ -0,0 +1,84 @@
...
+                        half{0.3},
+                        half{-0.6},
+                        half{10.0},
+                        half{-1.0},
+                        half{0.0},
+                        half{1.0},
+                        half{1.2},
+                        half{3.2},
+                        half{-4.1},
+                        half{5.3}};
```
umangyadav, line 48: You shouldn't require casting all the elements to half. Same applies to all other places.

src/onnx/parse_simplified_layer_normalization.cpp
```diff
@@ -0,0 +1,98 @@
...
+        auto eps    = info.add_literal(migraphx::literal{migraphx::shape{x_dtype}, {epsilon}});
+        rms         = info.add_common_op("add", rms, eps);
+        auto rrms   = info.add_instruction(make_op("rsqrt"), rms);
+        auto result = info.add_common_op("mul", x, rrms);
+        result      = info.add_common_op("mul", result, scale);
+
+        return {result, mean, rrms};
```
umangyadav, line 92: Is this being matched with LayerNorm kernel on the GPU target ?

TedThemistokleous (approved):

umangyadav (approved):

PR #3140, opened 2024-05-31
Add ONNX parsing for SkipSimplifiedLayerNormalization

turneram:


CharlieL7 (commented):
looks fine, I have the same questions as for #3129. Answering one should answer the other.

src/onnx/parse_skip_simplified_layer_normalization.cpp
```diff
@@ -0,0 +1,124 @@
...
+        // input : T
+        // 3D input tensor with shape (batch_size, sequence_length, hidden_size) Or 2D input tensor
+        // with shape (token_count, hidden_size) skip : T 3D input tensor with shape (batch_size,
+        // sequence_length, hidden_size) Or 2D input tensor with shape (token_count, hidden_size)
+        // gamma : T
+        // 1D input tensor with shape (hidden_size)
+        // bias (optional) : T
+        // 1D bias tensor with shape (hidden_size) - not used by ORT
+
+        if(args.size() != 3)
```
TedThemistokleous, line 75: Is there not a 4th input to handle then? Based on the doc its specifying 3-4 inputs.

turneram: Onnxruntime didn't implement theirs to utilize the bias input, possibly because the only model that currently uses this op doesn't call it with a bias input. I could add it if we want to be inline with the specs, but since I was using theirs as a baseline for correctness I wrote ours the same way.

TedThemistokleous: Yeah I get it. Nothing wrong with what you did. I think its better to cover our bases on this though as we've seen fun outcomes with some of these optional inputs

My only concern really is because microsoft wrote: 
_This version of the operator has been available since version 1 of the 'com.microsoft' operator set._

https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.SkipSimplifiedLayerNormalization

TedThemistokleous (changes requested):
Add the optional input parameters and attributes and outputs. Likely the outputs can just add be composite of existing ops we have like ReduceSum for the mean and bias sum outputs

- [x]  bias input
- [x] mean (output)
- [x] inv_std_var (output)
- [x] input_skip_bias_sum (output)

test/onnx/verify/skip_simplified_layer_normalization.cpp
```diff
@@ -0,0 +1,245 @@
...
+                        half{0.3},
+                        half{-0.6},
+                        half{10.0},
+                        half{-1.0},
+                        half{0.0},
+                        half{1.0},
+                        half{1.2},
+                        half{3.2},
+                        half{-4.1},
+                        half{5.3}};
```
umangyadav, line 48: You shouldn't require casts

turneram: Removing the casts causes compilation errors. It looks like the way I have it is what we do in other tests using the half type.

umangyadav (approved):

test/onnx/verify/skip_simplified_layer_normalization.cpp
umangyadav: It does look like we use it like this, but just as an alternate solution you can try passing Float vector to `migraphx::argument` which has "half_type" as its shape.

src/onnx/parse_skip_simplified_layer_normalization.cpp
```diff
@@ -0,0 +1,137 @@
...
+        auto x_dtype       = x_shape.type();
+        int64_t x_rank     = x_shape.ndim();
+        int64_t skip_rank  = skip->get_shape().ndim();
+        int64_t gamma_rank = gamma->get_shape().ndim();
+        // axis = hidden_size dim
+        int64_t axis = x_rank - 1;
+
+        if(x_rank < 2 or x_rank > 3 or x_rank != skip_rank or gamma_rank != 1)
+        {
+            MIGRAPHX_THROW("PARSE_SKIPSIMPLIFIEDLAYERNORMALIZATION: invalid input shape");
```
TedThemistokleous, line 99: Add some test cases for the invalid input cases that trigger this and the above throws

TedThemistokleous (approved):
CI for clang, hiprtc,etc is all green, Not sure why its not reflected on this PR. 

Add the two minor tests to check the throw cases apart from that looks good. Will kick off a merge from develop to see if that solves CI blockage with windows

umangyadav:
@turneram  fix tidy

PR #3158, opened 2024-06-06
Update gemm parse to re-order transpose operation

ahsan-ca:


src/onnx/parse_gemm.cpp
```diff
@@ -71,6 +71,14 @@ struct parse_gemm : op_parser<parse_gemm>
 
         std::vector<int64_t> perm = {1, 0};
         auto dot_type             = a_arg->get_shape().type();
+
+        a_arg = (trans_a)
```
umangyadav, line 75: I do not see why this would necessary

umangyadav: This would make `mul` operation take transposed inputs in pointwise kernel. It may impact performance ( i am 100% sure).

ahsan-ca:
This change was proposed to use `add_apply_alpha_beta` for gemm_fp8_test. Discussed this change with Umang and decided to abandon this change and make changes to the test case instead and not use  `add_apply_alpha_beta` .

PR #3160, opened 2024-06-06
Increased stack size for migraphx-hiprtc-driver to 4MB on Windows

slojosic-amd:
Default 1MB stack size is not enough for migraphx-hiprtc-driver.exe on Windows.
@apwojcik


umangyadav (approved):

src/targets/gpu/hiprtc/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ add_executable(migraphx-hiprtc-driver
     main.cpp
 )
 rocm_clang_tidy_check(migraphx-hiprtc-driver)
+if(WIN32)
+    target_link_options(migraphx-hiprtc-driver PRIVATE -Xlinker /stack:4194304)
```
pfultz2, line 30: I think this option is only valid when using MSVC linker. We should probably check for that.

apwojcik: There is no need. The llvm-ld.exe is just a frontend, and on Windows, it invokes the MSVC linker (link.exe).

apwojcik: The only exception is when MINGW (or CYGWIN) is used then the GNU linker is used.

pfultz2: Sure, but we could add a check for GCC and skip adding this flag so it wont break any users that might want to build with gcc in the future.

apwojcik: @pfultz2 done

Issue #2640, opened 2024-01-12
Make naming consistent for test/verify cpp files


kahmed10:
Currently under test/verify/, we have files with inconsistent naming. Either remove "test" from the name (ex. `test_abs.cpp` to `abs.cpp`) or prepend "test" to all operators (ex. `gemm_add.cpp` to `test_gemm_add.cpp`).

PR #3168, opened 2024-06-10
Make verify tests naming consistent

eddieliao:
Used the following script to make the naming of verify tests consistent:
```python
import os, glob

files = [f for f in glob.glob("*.cpp") if not (f.startswith("test") 
                                               or f.startswith("auto") 
                                               or f.startswith("main")
                                               or f.startswith("run")
                                               or f.startswith("verify"))]

for f in files:
    print("Modifying " + f)
    find = f.split(".cpp")[0]
    replace = "test_" + find
    with open(f, 'r') as file:
        contents = file.read()
        contents = contents.replace(find, replace)
    with open(f, 'w') as file:
        file.write(contents)
    os.rename(f, "test_" + f)

print("---Done---")
```

Addresses #2640.


umangyadav (approved):
Thanks.

PR #3171, opened 2024-06-12
Parentheses readability

### LINKED ISSUE #2546: Update matchers formatting to be more readable
I think we should be formatting using tabs and whitespace more liberally for this usecase. The current syntax is nearly impossible to read.

eddieliao:
Add a cppcheck rule for four or more consecutive parentheses; closes #2546.

### DIFF
--- a/tools/cppcheck/migraphx.py
+++ b/tools/cppcheck/migraphx.py
@@ -1,7 +1,7 @@
 #####################################################################################
 # The MIT License (MIT)
 #
-# Copyright (c) 2015-2023 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
--- a/tools/cppcheck/migraphx.py
+++ b/tools/cppcheck/migraphx.py
@@ -426,3 +426,10 @@ def useStlAlgorithms(cfg, data):
                                  "Use std::equal_range instead.")
         elif token.str == 'memchr':
             cppcheck.reportError(token, "style", "Use std::find instead.")
+
+@cppcheck.checker
+def NestedParentheses(cfg, data):
+    for token in cfg.tokenlist:
+        if not simpleMatch(token, ") ) ) )"):
+            continue
+        cppcheck.reportError(token, "style", "Too many nested parentheses can affect readability; consider using variables instead.")
--- a/tools/cppcheck/migraphx.py
+++ b/tools/cppcheck/migraphx.py
@@ -433,9 +433,7 @@ def MatcherNestedParentheses(cfg, data):
     for token in cfg.tokenlist:
         if not simpleMatch(token, "matcher ( ) const {"):
             continue
-        for tok2 in cfg.tokenlist[cfg.tokenlist.index(token):]:
-            if simpleMatch(tok2, "}"):
-                break
+        for tok2 in token.tokAt(4).forward(token.linkAt(4)):
             if not simpleMatch(tok2, ") ) ) )"):
                 continue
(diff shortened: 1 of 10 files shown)


causten:
Can you add a before/after screenshot

CharlieL7:
We probably want to fix the current warnings in this PR also. I can make some commits if it's not clear how to fix them

pfultz2:
There are a lot of warning outside of matchers, which I think a limit of 4 is too aggressive. We could just target the check to the `matcher()` functions with something like this:

```python
@cppcheck.checker
def MatcherNestedParentheses(cfg, data):
    for token in cfg.tokenlist:
        if not simpleMatch(token, 'matcher ( ) const {'):
            continue
        for tok2 in token.tokAt(4, token.linkAt(4)):
            if not simpleMatch(tok2, ") ) ) )"):
                continue
            cppcheck.reportError(
                tok2, "style",
                "Too many nested parentheses can affect readability; consider using variables instead."
            )
            break
```

eddieliao:
> Can you add a before/after screenshot

What screenshots would you like here?

> We probably want to fix the current warnings in this PR also. I can make some commits if it's not clear how to fix them

That would be appreciated; just an example or two would help.



> There are a lot of warning outside of matchers, which I think a limit of 4 is too aggressive. We could just target the check to the `matcher()` functions with something like this:
> 
> ```python
> @cppcheck.checker
> def MatcherNestedParentheses(cfg, data):
>     for token in cfg.tokenlist:
>         if not simpleMatch(token, 'matcher ( ) const {'):
>             continue
>         for tok2 in token.tokAt(4, token.linkAt(4)):
>             if not simpleMatch(tok2, ") ) ) )"):
>                 continue
>             cppcheck.reportError(
>                 tok2, "style",
>                 "Too many nested parentheses can affect readability; consider using variables instead."
>             )
>             break
> ```

4 is pretty aggressive I agree, although I was just basing this off of the original comment of "more than three parentheses". I can go ahead and use your suggestion to only target `matcher()`; let me know if there's another number you have in mind for consecutive parentheses.

CharlieL7:
Fixed the fuse_concat matcher syntax. The other warnings will disappear with the cppcheck change.

tools/cppcheck/migraphx.py
```diff
@@ -429,11 +429,15 @@ def useStlAlgorithms(cfg, data):
...
+def MatcherNestedParentheses(cfg, data):
     for token in cfg.tokenlist:
-        if not simpleMatch(token, ") ) ) )"):
+        if not simpleMatch(token, 'matcher ( ) const {'):
             continue
-        cppcheck.reportError(
-            token, "style",
-            "Too many nested parentheses can affect readability; consider using variables instead."
-        )
+        for tok2 in token.tokAt(4, token.linkAt(4)):
```
pfultz2: You need to do `token.tokAt(4).forward(token.linkAt(4))`. This is most likely where the python error is coming from.

tools/cppcheck/migraphx.py
```diff
@@ -431,9 +431,11 @@ def useStlAlgorithms(cfg, data):
 @cppcheck.checker
 def MatcherNestedParentheses(cfg, data):
     for token in cfg.tokenlist:
-        if not simpleMatch(token, 'matcher ( ) const {'):
+        if not simpleMatch(token, "matcher ( ) const {"):
             continue
-        for tok2 in token.tokAt(4, token.linkAt(4)):
+        for tok2 in cfg.tokenlist[cfg.tokenlist.index(token):]:
```
pfultz2, line 436: It better to just use the `forward` method to iterate the tokens lazily:

https://github.com/danmar/cppcheck/blob/main/addons/cppcheckdata.py#L503

eddieliao: Changed, thanks

CharlieL7 (approved):

PR #3182, opened 2024-06-14
Add flag to dump mlir modules to mxrs

pfultz2:


src/targets/gpu/jit/mlir.cpp
```diff
@@ -198,8 +200,14 @@ struct mlir_compiler : compiler<mlir_compiler>
                                               const operation&,
                                               bool exhaustive) const
     {
+        static const auto mxr_loc = string_value_of(MIGRAPHX_MLIR_DUMP_TO_MXR{});
```
umangyadav, line 203: add to ENV var document. 

Also what's the usefulness of having MXRs for each MLIR module ?

pfultz2: Added docs.

umangyadav (approved):

shivadbhavsar (approved):

PR #3191, opened 2024-06-17
Add onnx_options external data path setter

stsokolo:
There was no support for setting external_data_path property in onnx_options structure. This is needed so that this path cen be set from onnxruntime.

### DIFF
--- a/src/api/api.cpp
+++ b/src/api/api.cpp
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
- * Copyright (c) 2015-2023 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -162,6 +162,11 @@ void set_default_dyn_dim_value(onnx_options& options, const shape::dynamic_dimen
 void set_default_loop_iterations(onnx_options& options, int64_t value)
 {
     options.max_loop_iterations = value;
+}
+
+void set_external_data_path(onnx_options& options, const char* external_data_path)
+{
+    options.external_data_path = std::string(external_data_path);
 }
 
 void set_limit_loop_iterations(onnx_options& options, int64_t value)
@@ -1916,6 +1921,18 @@ migraphx_onnx_options_set_limit_loop_iterations(migraphx_onnx_options_t onnx_opt
         if(onnx_options == nullptr)
             MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter onnx_options: Null pointer");
         migraphx::set_limit_loop_iterations((onnx_options->object), (value));
+    });
+    return api_error_result;
+}
+
+extern "C" migraphx_status
+migraphx_onnx_options_set_external_data_path(migraphx_onnx_options_t onnx_options,
+                                             const char* external_data_path)
+{
+    auto api_error_result = migraphx::try_([&] {
+        if(onnx_options == nullptr)
+            MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter onnx_options: Null pointer");
+        migraphx::set_external_data_path((onnx_options->object), (external_data_path));
     });
     return api_error_result;
 }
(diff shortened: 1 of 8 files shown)


src/api/api.cpp
```diff
@@ -169,6 +169,12 @@ void set_limit_loop_iterations(onnx_options& options, int64_t value)
     options.limit_max_iterations = value;
 }
 
+void set_external_data_path(onnx_options& options,
```
umangyadav, line 172: https://github.com/ROCm/AMDMIGraphX/blob/eed559feb338b8972335b2aee5d48602d89583b0/tools/api/api.cpp#L157

this API file is auto generated. 

You need to add this `set_external_data_path` function in `tools/api.cpp` instead of directly modifying this src file.

umangyadav (changes requested):
Thanks for the contribution. 

APIs are auto-generated. You need to make changes inside `tools/api.cpp` and `src/api/migraphx.py` and then 
run `python3 tools/generate.py` which would all necessary APIs inside `src/api/*.h *hpp *.cpp` files.

src/api/include/migraphx/migraphx.h
```diff
@@ -518,6 +518,9 @@ MIGRAPHX_C_EXPORT migraphx_status migraphx_onnx_options_set_default_loop_iterati
 MIGRAPHX_C_EXPORT migraphx_status migraphx_onnx_options_set_limit_loop_iterations(
     migraphx_onnx_options_t onnx_options, int64_t value);
 
+MIGRAPHX_C_EXPORT migraphx_status migraphx_onnx_options_set_external_data_path(
```
umangyadav, line 521: These APIs are auto generated. 
You would need to declare them here : 
https://github.com/ROCm/AMDMIGraphX/blob/eed559feb338b8972335b2aee5d48602d89583b0/src/api/migraphx.py#L324

instead of directly modifying these files.

src/api/include/migraphx/migraphx.hpp
```diff
@@ -1343,6 +1343,14 @@ struct onnx_options : MIGRAPHX_HANDLE_BASE(onnx_options)
     {
         call(&migraphx_onnx_options_set_limit_loop_iterations, this->get_handle_ptr(), value);
     }
+
+    /// Set absolute path for external data files
+    void set_external_data_path(const std::string& external_data_path)
```
umangyadav, line 1348: Need to add an option in python API as well for this. 
you can add bindings in pybind here 
https://github.com/ROCm/AMDMIGraphX/blob/eed559feb338b8972335b2aee5d48602d89583b0/src/py/migraphx_py.cpp#L480

src/api/api.cpp
stsokolo: Done.

src/api/include/migraphx/migraphx.hpp
stsokolo: Done.

src/api/include/migraphx/migraphx.h
stsokolo: Done.

src/api/include/migraphx/migraphx.hpp
umangyadav, line 1348: Can you add a simple test example with external data path ? 
for example something like this 
https://github.com/ROCm/AMDMIGraphX/blob/2c4dd4a9e4af590539ddee7e021b99443cff9cda/test/api/test_cpu.cpp#L104 ?? 

we have onnx simple onnx file with ext data here which you can make use of or else you can generate a new one on your own. 

https://github.com/ROCm/AMDMIGraphX/tree/develop/test/onnx/ext_path

stsokolo: Done.

umangyadav: Can you add `compile` and `eval` similar to this one ? 
https://github.com/ROCm/AMDMIGraphX/blob/2c4dd4a9e4af590539ddee7e021b99443cff9cda/test/api/test_cpu.cpp#L110

stsokolo: I added the changes.

umangyadav (approved):

TedThemistokleous (approved):

PR #3202, opened 2024-06-19
Layernorm Perf improvement: increase wg size

lakhinderwalia:
Increase performance of layernorm for unet model by ~ 33%: tensor {float, 2, 32, 81920} was evaluated as a test case.


pfultz2:
What happens to the performance for sizes like `{float, 2, 32, 512}` and `{float, 2, 32, 1024}`?

lakhinderwalia:
> What happens to the performance for sizes like `{float, 2, 32, 512}` and `{float, 2, 32, 1024}`?

I see a smaller perf improvement. Maybe 10% is a number I can quote for the smaller kernel here, but the variation is an important caveat here, and it is hard to be very precise. Thanks.

pfultz2:
> What happens to the performance for sizes like {float, 2, 32, 512} and {float, 2, 32, 1024}?

Actually these sizes are too small, as they still use block size of 256.

I ran this across different sizes to compare where it would pick a different block size:

```
develop:
gpu::code_object[code_object=10256,symbol_name=layernorm_kernel,global=2097152,local=256,] -> float_type, {256, 32, 2096}, {67072, 2096, 1}: 0.110351ms
gpu::code_object[code_object=11024,symbol_name=layernorm_kernel,global=2097152,local=256,] -> float_type, {256, 32, 8192}, {262144, 8192, 1}: 0.428912ms
gpu::code_object[code_object=25936,symbol_name=layernorm_kernel,global=2097152,local=256,] -> float_type, {256, 32, 81920}, {2621440, 81920, 1}: 4.54644ms
gpu::code_object[code_object=9744,symbol_name=layernorm_kernel,global=2097152,local=256,] -> float_type, {256, 32, 327680}, {10485760, 327680, 1}: 32.6882ms
PR:
gpu::code_object[code_object=10000,symbol_name=layernorm_kernel,global=4194304,local=512,] -> float_type, {256, 32, 2096}, {67072, 2096, 1}: 0.107564ms
gpu::code_object[code_object=10256,symbol_name=layernorm_kernel,global=4194304,local=512,] -> float_type, {256, 32, 8192}, {262144, 8192, 1}: 0.42505ms
gpu::code_object[code_object=16976,symbol_name=layernorm_kernel,global=4194304,local=512,] -> float_type, {256, 32, 81920}, {2621440, 81920, 1}: 4.69303ms
gpu::code_object[code_object=60432,symbol_name=layernorm_kernel,global=4194304,local=512,] -> float_type, {256, 32, 327680}, {10485760, 327680, 1}: 53.006ms
```

With smaller sizes like `2096` its 10% faster, where as `8192` there is almost no difference. However, once it gets larger it get slower. `81920` is only 3% slower, but `327680` is 60% slower. I dont think this is a good heuristic. 

Without sampling a bunch of size to get a better formula for the block size, I think it would be better to just make the block size a tuning parameter so we can time it for each size we see.

lakhinderwalia:
I am a strong proponent of the block size as a tuning parameter, based on my detailed testing results posted elsewhere. So agree on that future direction, which would apply not just to this operator, because there would be many sizes where other operators today with a fixed block size would show a degradation, is my guess.

But, bs=512 for layernorm didn't degrade in my tests while compared to bs=256. And that is why I recommend it in the PR. I couldn't run {256, 32, 327680}, which failed to run. The large size of  {256, 32, 81920} ran approx. 20% improvement seen with BS=512, compared to BS=256. 

(BTW, I use modified sizes in test_layernorm_test -- which actually is a layernorm+mul+add kernel'.)

CharlieL7 (approved):

PR #3203, opened 2024-06-20
Load and save problem_cache from a file

pfultz2:


src/targets/gpu/problem_cache.cpp
```diff
@@ -24,11 +24,34 @@
...
+#include <migraphx/json.hpp>
+#include <migraphx/env.hpp>
+#include <migraphx/serialize.hpp>
+#include <migraphx/file_buffer.hpp>
 
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_PROBLEM_CACHE)
```
umangyadav, line 36: Need to add this to ENV var docs

src/targets/gpu/problem_cache.cpp
```diff
@@ -24,11 +24,34 @@
...
+
+void problem_cache::load()
+{
+    auto pc_path = string_value_of(MIGRAPHX_PROBLEM_CACHE{});
+    if(pc_path.empty())
+        return;
+    if(not fs::exists(pc_path))
+        return;
```
umangyadav, line 44: It would be nice to print out warning/log message saying path is non-existent.

src/targets/gpu/problem_cache.cpp
```diff
@@ -24,11 +24,34 @@
...
+    from_value(from_json_string(read_string(pc_path)), cache);
+}
+void problem_cache::save() const
+{
+    auto pc_path = string_value_of(MIGRAPHX_PROBLEM_CACHE{});
+    if(pc_path.empty())
+        return;
+    write_string(pc_path, to_pretty_json_string(to_value(cache)));
```
umangyadav, line 52: This would overwrite exisiting cache on same path i suppose.

src/include/migraphx/serialize.hpp
```diff
@@ -191,9 +197,27 @@ template <class T>
 auto from_value_impl(rank<4>, const value& v, T& x)
     -> decltype(x.insert(*x.begin()), std::declval<typename T::mapped_type>(), void())
 {
-    x.clear();
-    for(auto&& e : v)
-        x.emplace(e.get_key(), from_value<typename T::mapped_type>(e));
+    if(v.is_object())
```
umangyadav, line 200: add tests for these

src/include/migraphx/serialize.hpp
```diff
@@ -191,9 +197,27 @@ template <class T>
...
+        for(auto&& e : v)
+            x.emplace(e.get_key(), from_value<typename T::mapped_type>(e));
+    }
+    else if(v.is_array())
+    {
+        x.clear();
+        for(auto&& e : v)
+        {
+            if(e.size() != 2)
+                MIGRAPHX_THROW("Expected a pair");
```
umangyadav, line 212: Why this restriction ?

pfultz2: Because its loading into a map.

src/targets/gpu/problem_cache.cpp
pfultz2: Yes it loads it in memory and then writes it out at the end.

src/include/migraphx/serialize.hpp
pfultz2: Tests added.

src/targets/gpu/problem_cache.cpp
pfultz2: added.

src/targets/gpu/problem_cache.cpp
umangyadav: This is unresolved

pfultz2: Added docs.

umangyadav (approved):

kahmed10 (approved):

PR #3206, opened 2024-06-20
Implement `reduce_any` and `reduce_all`

CharlieL7:
* `reduce_any` = reduce using logical_or
* `reduce_all` = reduce using logical_and
* Not in the ONNX spec, needed for the related PyTorch operators
* Output type of `reduce_any` and `reduce_all` is the original type (not bool).
* Update to float8 to convert to bool


pfultz2 (approved):

src/include/migraphx/float8.hpp
```diff
@@ -116,6 +116,8 @@ struct float8
         return migraphx::fp8::impl::cast_from_f8<2, 5, float, FNUZ /*negative_zero_nan*/>(data);
     }
 
+    inline explicit constexpr operator bool() const { return not is_zero(); }
```
umangyadav, line 119: Can you add some unit-tests for this one inside `test/fp8e*.cpp` ?

umangyadav (approved):

PR #3219, opened 2024-06-26
Fix bug in 1arg dynamic reshape

CharlieL7:
* Fixes a bug when the input shape and `dims` attribute have a different number of dimensions
* Fixes bug seen in bertsquad-12

### DIFF
--- a/src/include/migraphx/op/reshape.hpp
+++ b/src/include/migraphx/op/reshape.hpp
@@ -76,22 +76,24 @@ struct reshape
         const bool has_negative_dim_attr = neg_dim_num < dims.size();
         // construct output dynamic shape from dims attribute
         std::vector<shape::dynamic_dimension> output_dyn_dims(dims.size());
-        std::transform(dims.begin(),
-                       dims.end(),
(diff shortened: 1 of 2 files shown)


umangyadav (approved):

src/include/migraphx/op/reshape.hpp
```diff
@@ -76,22 +76,24 @@ struct reshape
         const bool has_negative_dim_attr = neg_dim_num < dims.size();
         // construct output dynamic shape from dims attribute
         std::vector<shape::dynamic_dimension> output_dyn_dims(dims.size());
-        std::transform(dims.begin(),
-                       dims.end(),
-                       input_dyn_dims.begin(),
```
umangyadav, line 81: this loop would run for `dims.size()` times.  I do not see difference in logic between current change and new logic.

CharlieL7: Simply that we don't access `input_dyn_dims` out of bounds

ahsan-ca (approved):

PR #3222, opened 2024-06-26
Add weight streaming

### LINKED ISSUE #3156: Add weight streaming at runtime
Figure out a way to have weight streaming at runtime i.e. be able to fit large models on gpu without needing to know literal size ahead of time

- [x] Define/determine an allocation of literals to be streamed
- [x] Move copy instructions to separate stream
- [x] Investigate why `@literal` instructions take up so much time
- [x] Decrease time spent on `@literal` instruction
- [ ] Move from naive allocation to "smart" allocation (figure out best way to mask time taken to copy)

eddieliao:
Add weight streaming to allow running of large models on GPUs with low memory.

Closes #3156.


eddieliao:
17 of 18 test failures right now for `test/schedule_test` is due to a check of `EXPECT(not t.has_stream(one));`. This is failing due to the fact that I have added both literals and their associated `copy_to_gpu` to a stream. Should literals thus not be included on the stream?

eddieliao:
Doing some very basic perf testing adding the literals onto the stream with copies seems to very marginally improve performance. Still not sure if the tests can be ignored since this is a new feature.

eddieliao:
Seems to be stuck on a Windows build for some reason? Looking into why but would appreciate help if anybody has any pointers

eddieliao:
Removed call to `make_shared_array` to avoid slowdown due to `std::copy`.

Running `./build/bin/driver perf models/resnet50-v2-7.onnx --enable-weight-streaming`:

Before:
```
Summary:
hip::copy_to_gpu: 55.82ms / 117 = 0.477094ms, 61%
@literal: 31.1419ms / 117 = 0.26617ms, 34%
gpu::miopen_fusion: 1.33101ms / 12 = 0.110917ms, 2%
gpu::code_object::mlir_convolution_add_relu: 0.784531ms / 17 = 0.0461489ms, 1%
gpu::convolution: 0.573863ms / 4 = 0.143466ms, 1%
load: 0.317283ms / 191 = 0.00166117ms, 1%
gpu::code_object::mlir_convolution_add: 0.278956ms / 8 = 0.0348696ms, 1%
gpu::code_object::mul_add_relu_kernel: 0.244727ms / 8 = 0.0305909ms, 1%
gpu::code_object::mlir_convolution: 0.236977ms / 5 = 0.0473954ms, 1%
gpu::code_object::mlir_convolution_mul_add_relu: 0.232219ms / 4 = 0.0580549ms, 1%
multibroadcast: 0.206982ms / 67 = 0.00308928ms, 1%
gpu::gemm: 0.167417ms / 1 = 0.167417ms, 1%
gpu::code_object::mlir_convolution_mul_add_add_relu: 0.124884ms / 3 = 0.0416281ms, 1%
gpu::pooling: 0.105871ms / 1 = 0.105871ms, 1%
gpu::code_object::add_relu_noop_concat_noop_kernel: 0.085841ms / 3 = 0.0286137ms, 1%
gpu::code_object::mul_kernel: 0.0637746ms / 2 = 0.0318873ms, 1%
gpu::code_object::concat_add_relu_kernel: 0.0290406ms / 1 = 0.0290406ms, 1%
gpu::code_object::mul_add_add_relu_mul_reduce_sum_kernel: 0.02897ms / 1 = 0.02897ms, 1%
gpu::code_object::add_relu_kernel: 0.0275658ms / 1 = 0.0275658ms, 1%
step: 0.0107792ms / 3 = 0.00359307ms, 1%
hip::hip_copy_literal: 0.009445ms / 2 = 0.0047225ms, 1%
@param: 0.0049534ms / 2 = 0.0024767ms, 1%
hip::hip_allocate_memory: 0.0043382ms / 1 = 0.0043382ms, 1%
check_context::migraphx::gpu::context: 0.0036888ms / 1 = 0.0036888ms, 1%
reshape_lazy: 0.003683ms / 1 = 0.003683ms, 1%

Batch size: 1
Rate: 19.0196 inferences/sec
Total time: 52.5772ms
Total instructions time: 91.8387ms
Overhead time: 0.135608ms, -39.2615ms
Overhead: 0%, -75%
```

After:
```
Summary:
hip::copy_to_gpu: 6.37803ms / 117 = 0.0545131ms, 60%
gpu::miopen_fusion: 1.12564ms / 12 = 0.0938037ms, 11%
gpu::code_object::mlir_convolution_add_relu: 0.747901ms / 17 = 0.0439942ms, 7%
load: 0.378693ms / 191 = 0.00198269ms, 4%
gpu::convolution: 0.349078ms / 4 = 0.0872696ms, 4%
gpu::code_object::mlir_convolution_add: 0.267458ms / 8 = 0.0334322ms, 3%
multibroadcast: 0.257707ms / 67 = 0.00384638ms, 3%
gpu::code_object::mlir_convolution: 0.225739ms / 5 = 0.0451478ms, 3%
gpu::code_object::mul_add_relu_kernel: 0.220406ms / 8 = 0.0275508ms, 3%
gpu::code_object::mlir_convolution_mul_add_relu: 0.212062ms / 4 = 0.0530154ms, 2%
gpu::code_object::mlir_convolution_mul_add_add_relu: 0.122514ms / 3 = 0.040838ms, 2%
gpu::code_object::add_relu_noop_concat_noop_kernel: 0.0843656ms / 3 = 0.0281219ms, 1%
gpu::gemm: 0.0727626ms / 1 = 0.0727626ms, 1%
gpu::pooling: 0.067244ms / 1 = 0.067244ms, 1%
gpu::code_object::mul_kernel: 0.051746ms / 2 = 0.025873ms, 1%
@literal: 0.0353714ms / 117 = 0.00030232ms, 1%
gpu::code_object::concat_add_relu_kernel: 0.0287566ms / 1 = 0.0287566ms, 1%
gpu::code_object::mul_add_add_relu_mul_reduce_sum_kernel: 0.0274078ms / 1 = 0.0274078ms, 1%
gpu::code_object::add_relu_kernel: 0.0251514ms / 1 = 0.0251514ms, 1%
step: 0.0067122ms / 3 = 0.0022374ms, 1%
hip::hip_copy_literal: 0.0028988ms / 2 = 0.0014494ms, 1%
reshape_lazy: 0.0017368ms / 1 = 0.0017368ms, 1%
@param: 0.0017358ms / 2 = 0.0008679ms, 1%
hip::hip_allocate_memory: 0.0012642ms / 1 = 0.0012642ms, 1%
check_context::migraphx::gpu::context: 0.0010434ms / 1 = 0.0010434ms, 1%

Batch size: 1
Rate: 162.816 inferences/sec
Total time: 6.1419ms
Total instructions time: 10.6934ms
Overhead time: 0.132798ms, -4.55153ms
Overhead: 2%, -74%
```

eddieliao:
Failing Windows build right now, seems to be a problem with `fuse_mlir` from develop

TedThemistokleous:
> Failing Windows build right now, seems to be a problem with `fuse_mlir` from develop

All good. That'll get resolved on its own once develop changes.

 Make sure to also add some more test coverage as well. Codecov/patch says you have about 62.50% coverage on your change set. We try to maintain about 92-93% coverage so add a few more unit tests that exercise the code paths you've added

eddieliao:
> Make sure to also add some more test coverage as well. Codecov/patch says you have about 62.50% coverage on your change set. We try to maintain about 92-93% coverage so add a few more unit tests that exercise the code paths you've added

The changes in `src/program.cpp` I think are mostly due to the addition of passing a bool into `generic_eval()`, which wasn't originally covered. I can add some tests to cover that although it would require the setting of an env var to change the trace level, is there a way to do this through the tests?

eddieliao:
Discussed with Umang and Alan, will ignore codecov failure for now as the code path not covered is purely for debugging. Can add later if needed if it's possible to set env vars inside of a test.

src/include/migraphx/literal.hpp
```diff
@@ -98,10 +98,18 @@ struct literal : raw_data<literal>
     std::vector<literal> get_sub_objects() const { return {}; }
 
     /// Convert the data to an argument
-    argument get_argument() const
+    argument get_argument(bool weight_streaming = false) const
```
pfultz2, line 101: There is no need to do this. You can create an operator that stores an `argument` and just returns that:

```cpp
struct literal_as_argument
{
    std::string name() const { return "gpu::literal_as_argument"; }
    argument data;

    {
        return pack(f(self.data, "data"));
    }

    shape compute_shape(const std::vector<shape>&) const
    {
        return data.get_shape();
    }
    argument compute_shape(const std::vector<argument>&) const
    {
        return data;
    }
};
MIGRAPHX_REGISTER_OP(literal_as_argument);
```

And then replace the literals with this operator in the write_literals pass.

eddieliao: Should this operator be created in a new header file or in the same file as `literal.hpp`?

pfultz2: I would create it directly in the write_literals.cpp file.

eddieliao: I tried putting it directly into the `write_literals.cpp` file but it's not as drop and replace as I expected. During execution it comes up with a `/code/AMDMIGraphX/src/include/migraphx/operation.hpp:359: compute_op: Not computable: gpu::literal_as_argument` error, any idea what this means? Not too familiar with how operators work.

pfultz2: Yea its missing the `compute` method. I mistyped the `compute` function. The second `compute_shape` is supposed to be `compute`.

eddieliao: Hmm that doesn't seem to work either (same error), I have the following as my operator:
```cpp
struct literal_as_argument
{
    std::string name() const { return "gpu::literal_as_argument"; }
    argument data;
    std::string data_name = "{ ... }";

    {
        return pack(f(self.data_name, "data"));
    }

    shape compute_shape(const std::vector<shape>&) const
    {
        return data.get_shape();
    }
    argument compute(const std::vector<argument>&) const
    {
        return data;
    }
};
MIGRAPHX_REGISTER_OP(literal_as_argument);
```

I changed the reflect function to avoid having the entire data field print out every time, but that shouldn't be affecting the operation.

eddieliao: Fixed and pushed, thanks

src/targets/gpu/write_literals.cpp
```diff
@@ -27,36 +27,142 @@
...
+struct literal_as_argument
+{
+    std::string name() const { return "gpu::literal_as_argument"; }
+    argument data;
+    std::string data_name = "{ ... }";
+
+    template <class Self, class F>
+    static auto reflect(Self& self, F f)
+    {
+        return pack(f(self.data_name, "data"));
```
pfultz2, line 48: You arent serializing the data.

eddieliao: Added back in the serialization of data, however it does print out the entire data argument which makes it very difficult to read the printout of the instructions. Tried to overwrite the printout with the following to no avail:
```cpp
friend std::ostream& operator<<(std::ostream& os, const operation& op) { return os << "testing..."; }
```

eddieliao: Fixed, thanks

src/include/migraphx/compile_options.hpp
```diff
@@ -40,6 +41,8 @@ struct compile_options
 
     bool fast_math       = true;
     bool exhaustive_tune = false;
+    bool weight_streaming = false;
```
pfultz2, line 44: I dont think we should have a flag to enable weight streaming, if you run out of memory then it is the only way to run the model, so it should be enabled by default.

src/include/migraphx/compile_options.hpp
```diff
@@ -40,6 +41,8 @@ struct compile_options
...
+    long streaming_budget = LONG_MAX;
```
pfultz2, line 45: I would prefer this option to be called `max_memory`, so the user can specify a max memory they want the model to use.

src/include/migraphx/schedule.hpp
```diff
@@ -41,6 +41,7 @@ struct MIGRAPHX_EXPORT schedule
 {
     schedule_model model{};
     bool enable = true;
+    bool weight_streaming = false;
```
pfultz2, line 44: The schedule shouldn't be aware directly of aware of weight_streaming. It doesnt make sense on non-async targets which wont have multiple streams. Instead, it we want to have some operators to be scheduled on specific stream then we should update the stream_model to support that.

src/program.cpp
```diff
@@ -445,6 +445,11 @@ std::vector<argument> generic_eval(const module* mod,
         {
             results.emplace(ins, trace(ins, [&] { return ins->get_literal().get_argument(); }));
         }
+        // else if(name == "gpu::literal_as_argument")
+        // {
+        //     results.emplace(
+        //         ins, trace(ins, [&] { return ins->get_literal(); }));
+        // }
```
pfultz2, line 452: The commented code should be removed.

src/targets/gpu/target.cpp
```diff
@@ -84,6 +84,9 @@ MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_CK)
 
 std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_options& options) const
 {
+    size_t free_mem = 0;
+    auto status     = hipMemGetInfo(&free_mem, nullptr);
+    std::cout << "Free memory: " << free_mem << " status: " << status << std::endl;
```
pfultz2, line 89: These debug prints shouldn't be in the `get_passes`.

src/include/migraphx/compile_options.hpp
eddieliao: How would we determine if the model runs out of memory? Unless the model is first compiled and run with no weight streaming to test if there is sufficient memory, but that would also require re-compiling anyways after a failure.

src/include/migraphx/compile_options.hpp
eddieliao: Changed, thanks

src/include/migraphx/schedule.hpp
eddieliao: Ended up removing any streaming implementation in the scheduler for the time being given that it doesn't currently improve performance.

src/program.cpp
eddieliao: Removed, thanks

src/targets/gpu/target.cpp
eddieliao: Removed, thanks

src/schedule.cpp
```diff
@@ -79,6 +79,8 @@ struct stream_info
                 // This will ensure a stream will be assigned to return
                 if(op.name() == "@return")
                     weight = 1;
+                if(stream_literals and op.name() == "@literal")
+                    weight = 2;
```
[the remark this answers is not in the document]
eddieliao: Removed, thanks

src/include/migraphx/compile_options.hpp
pfultz2: We can take the estimated scratch+literals+parameters to determine how much the memory the model will use. The estimated scratch might be an overestimate, but this is a boderline case. Its better to nder use the memory the over use it and fail. Furthermore, if the user really wants to use it on the GPU they can set the max memory larger.

src/include/migraphx/compile_options.hpp
```diff
@@ -40,6 +41,8 @@ struct compile_options
 
     bool fast_math       = true;
     bool exhaustive_tune = false;
+    bool weight_streaming = false;
+    long max_memory       = LONG_MAX;
```
pfultz2, line 45: Actually thinking about this more, I think we want it set to `0` for default, which would mean to use the memory available on the gpu, and then the user can set max to always disable weight streaming.

Also, this should probably be `std::size_t` type.

src/targets/gpu/write_literals.cpp
```diff
@@ -27,36 +27,155 @@
...
 void write_literals::apply(module& m) const
 {
     assert(ctx != nullptr);
     std::size_t n = 0;
-    for(auto ins : iterator_for(m))
+
+    if(weight_streaming)
     {
-        if(ins->name() == "@literal")
+        std::size_t bytes_on_gpu = 0;
```
pfultz2, line 76: This needs to be set to the size of the scratch memor+total bytes of all parameters.

src/targets/gpu/write_literals.cpp
```diff
@@ -27,36 +27,155 @@
...
+            {
+                ins_list.push_back(ins);
+                size_of_literals += ins->get_shape().bytes();
+            }
+        }
+
+        long budget = max_memory;
+        if(budget == LONG_MAX)
+        {
+            budget = static_cast<long>(size_of_literals / 4);
```
pfultz2, line 113: Why isnt this set to the `free_memory`?

src/targets/gpu/write_literals.cpp
```diff
@@ -27,36 +27,155 @@
...
+        }
+
+        size_t free_memory = 0;
```
pfultz2, line 116: This value is never used. It should be used to set the `budget` when its not been set.

src/include/migraphx/compile_options.hpp
eddieliao: What would be the point of setting the max if it always disables weight streaming? I understand the idea of 0 for default (aka 0 = make it run no matter what), but if setting the max disables weight streaming than it would either a. always fail with not enough memory, or b. always work since the max is > size of literals.

src/targets/gpu/write_literals.cpp
eddieliao: The `size_of_literals/4` is arbitrary; in this case it was chosen just to approximately pre-allocate a quarter of the literals. I can change this to be the user-defined `max_memory` (or `free_memory` if `max_memory = 0`); I'm assuming when you're referring to `free_memory` you mean `total free - (scratch+allocated literals+parameters)`?

src/targets/gpu/write_literals.cpp
eddieliao: Changed, thanks

src/targets/gpu/write_literals.cpp
eddieliao: Changed, thanks

src/targets/gpu/write_literals.cpp
eddieliao: Changed, thanks

kahmed10:
Also consider #4049 for design revisit

Issue #3224, opened 2024-06-26
Bug in find_concat_op


shivadbhavsar:
Repro:

```
# concat_err.py
import numpy as np
import migraphx

p = migraphx.program()
mm = p.get_main_module()

lit1 = mm.add_literal(np.random.randn(128, 1, 1, 1, 1).astype(np.float32))
lit2 = mm.add_literal(np.random.randn(1, 3, 1, 1, 1).astype(np.float32))

lit1_bc = mm.add_instruction(migraphx.op('multibroadcast', out_lens=[128, 3, 224, 224, 1]), [lit1])
lit2_bc = mm.add_instruction(migraphx.op('multibroadcast', out_lens=[128, 3, 224, 224, 1]), [lit2])

cat = mm.add_instruction(migraphx.op('concat', axis=4), [lit1_bc, lit2_bc])

mm.add_return([cat])
```

Compile:
`migraphx-driver compile concat_err.py `

PR #3242, opened 2024-07-02
fix bug in find_concat_op when input broadcasts are on different axes

shivadbhavsar:
Fixes issue #3224 
Which in turn fixes one of the issues exposed by #3104

### DIFF
--- a/src/simplify_algebra.cpp
+++ b/src/simplify_algebra.cpp
@@ -811,6 +811,18 @@ struct find_concat_op
                op.attributes().contains("pointwise");
     }
 
+    static bool is_valid_concat(std::vector<instruction_ref> ins, size_t axis)
+    {
+        auto concat_lens = ins.front()->get_shape().lens();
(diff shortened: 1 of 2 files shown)


src/simplify_algebra.cpp
```diff
@@ -811,6 +811,18 @@ struct find_concat_op
                op.attributes().contains("pointwise");
     }
 
+    static bool is_valid_concat(std::vector<instruction_ref> ins, size_t axis)
```
umangyadav, line 814: Looks like there is another bug on line 886

```
   auto pred = [](auto i, auto j) {
            return i->get_operator() == j->get_operator() and
                   i->inputs().size() == i->inputs().size() and
                   i->outputs().size() == i->outputs().size();
        };
```

shivadbhavsar: Tracking in #3245

TedThemistokleous (approved):

Issue #3218, opened 2024-06-26
Bug in find_splits pass


shivadbhavsar:
repo:

```
#splits_err.py
import numpy as np
import migraphx

p = migraphx.program()
mm = p.get_main_module()

in1 = mm.add_literal(np.random.randn(128, 24, 48, 48).astype(np.float32))
conv_550_w = mm.add_literal(np.random.randn(288, 24, 1, 1).astype(np.float32))
conv_550 = mm.add_instruction(migraphx.op('convolution', padding=[0, 0, 0, 0],stride=[1, 1],dilation=[1, 1],group=1,padding_mode=0), [in1, conv_550_w])

slc_552 = mm.add_instruction(migraphx.op('slice', axes=[1], starts=[0], ends=[144]), [conv_550])
slc_617 = mm.add_instruction(migraphx.op('slice', axes=[1], starts=[144], ends=[288]), [conv_550])

add_553_lit = mm.add_literal(np.random.randn(128, 144, 48, 48).astype(np.float32))
add_533 = mm.add_instruction(migraphx.op('add'), [add_553_lit, slc_552])

add_619_lit = mm.add_literal(np.random.randn(128, 144, 48, 48).astype(np.float32))
add_619 = mm.add_instruction(migraphx.op('add'), [slc_617, add_619_lit])

mm.add_return([add_533, add_619])
```

Compile using:
`migraphx-driver compile splits_err.py`

PR #3244, opened 2024-07-03
find_splits bugfix

shivadbhavsar:
Solves #3218 

Given a set of ops:
input: {2, 4, 4} -> slc1 {2,2,4}, slc2 {2,2,4} -> add(c1, slc1), add(c2, slc2) 

Current logic assumes that the slices are always at the same input location in the binary op. This PR handles the case when graph has something like the following instead:
input: {2, 4, 4} -> slc1 {2,2,4}, slc2 {2,2,4} -> add(slc1, c1), add(c2, slc2) 

In this case `add` is a commutative op so this simplification can still be done, so we force all the slices to be at input 0. In the case of non commutative ops, the matcher will return without performing a simplification.


src/simplify_algebra.cpp
```diff
@@ -1081,6 +1081,46 @@ struct find_splits
...
+        auto group_op = group.front()->get_operator();
+        assert(std::all_of(
+            group.begin(), group.end(), [&](auto i) { return i->get_operator() == group_op; }));
+
+        for(auto i : group)
+        {
+            if(std::find(splits.begin(), splits.end(), i->inputs().at(split_idx)) == splits.end())
+            {
+                auto args = i->inputs();
+                std::reverse(args.begin(), args.end());
```
umangyadav, line 1118: reverse trick only works if it binary op

umangyadav: you can put an assert that `assert(args.size() == 2)`

shivadbhavsar: Ya I can add an assert. The binary op assumption is made because its called from if branch that checks `start->inputs().size() == 2`

umangyadav (approved):

TedThemistokleous (approved):

PR #3253, opened 2024-07-09
Windows/fix memory issue

ivarusic-amd:
Disabling the check for 'get_available_gpu_memory' on Windows, which relies on hipMemGetInfo to determine free memory.

### DIFF
--- a/src/targets/gpu/hip.cpp
+++ b/src/targets/gpu/hip.cpp
@@ -108,10 +108,10 @@ static host_ptr_cache& get_host_ptr_cache()
 
 std::shared_ptr<void> allocate_gpu(std::size_t sz, bool host = false)
 {
-    #ifndef _WIN32
+#ifndef _WIN32
     if(sz > get_available_gpu_memory())
         MIGRAPHX_THROW("Memory not available to allocate buffer: " + std::to_string(sz));
-    #endif
+#endif
     void* alloc_ptr = nullptr;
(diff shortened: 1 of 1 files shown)


src/targets/gpu/hip.cpp
```diff
@@ -108,8 +108,10 @@ static host_ptr_cache& get_host_ptr_cache()
 
 std::shared_ptr<void> allocate_gpu(std::size_t sz, bool host = false)
 {
+    #ifndef _WIN32
     if(sz > get_available_gpu_memory())
         MIGRAPHX_THROW("Memory not available to allocate buffer: " + std::to_string(sz));
```
umangyadav, line 113: Is `hipMemGetInfo` doesn't work on windows ? 

Also simply disabling the check may not print useful message when it fails.  It would need to rely on hipHostMalloc or hipMalloc to produce error code

ivarusic-amd: We assume it's not working correctly . This will be a workaround for our testing. You can find more information at the following links. Comments on this: https://confluence.amd.com/display/VPGFXAT/Stable+Diffusion+malfunction+investigation?focusedCommentId=1372236680&refresh=1720075365089#comment-1372236680 , and here https://ontrack-internal.amd.com/browse/SWDEV-471422

umangyadav (approved):

src/targets/gpu/hip.cpp
ivarusic-amd: We assume it's not working correctly . This will be a workaround for our testing.

PR #3256, opened 2024-07-09
use unitAttr for split-k attribute

### LINKED ISSUE #3254: Make `enable_splitk_for_tuning` unit attr
It is unit attr, https://mlir.llvm.org/docs/Dialects/Builtin/#unitattr

If it is present, MLIR will enable split-k configs, no matter what its value is

MIGraphX currently sets its value to `false` which has no effect. Instead MIGraphX shouldn't include attribute at all to disable split-k since it is Unit Attr

umangyadav:
Fixes #3254


manupak (approved):
LGTM.

Just a note for future, if this needs to be disabled, then  "enable_splitk_for_tuning" should be removed as a key here.

TedThemistokleous (approved):

PR #3264, opened 2024-07-10
Comment out qlinear reuse matcher

CharlieL7:
Comment out the qlinear_reused matcher because of an accuracy error for quantized resnet50

### DIFF
--- a/src/simplify_qdq.cpp
+++ b/src/simplify_qdq.cpp
@@ -324,38 +324,39 @@ bool precedes(Iterator x, Iterator y, Iterator last)
     return any_of(iterator_for(r), [&](auto it) { return it == y; });
 }
 
-struct match_qlinear_reused
-{
-    auto matcher() const
-    {
-        return match::name("quantizelinear")(
-            match::used_once(), match::arg(0)(match::none_of(match::used_once()).bind("x")));
-    }
-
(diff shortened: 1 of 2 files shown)


umangyadav (approved):

shivadbhavsar (approved):

umangyadav:
Is this required if we were to pull in changes from MLIR ?

CharlieL7:
> Is this required if we were to pull in changes from MLIR ?

We won't hit the verify "accuracy" bug for the quantized resnet50 model with the change from MLIR. However, the matcher itself is questionable. I'm proposing to remove it permanently here in #3269.

pfultz2:
> We won't hit the verify "accuracy" bug for the quantized resnet50 model with the change from MLIR.

If we dont hit an accuracy error then we shouldn't remove it.

CharlieL7:
> > We won't hit the verify "accuracy" bug for the quantized resnet50 model with the change from MLIR.
> 
> If we dont hit an accuracy error then we shouldn't remove it.

Matcher is brittle, if we change the randomized values slightly it will even fail without MLIR.

CharlieL7:
Closing: Should be able to pass verify accuracy with this compiler pass enable with the MLIR fix https://github.com/ROCm/rocMLIR/pull/1567

PR #3268, opened 2024-07-11
Find dot slice

### LINKED ISSUE #3267: Optimize Dot + Slice
RNN has patterns where 
Output of GEMM is immediately Sliced. 

MIGraphX can slice inputs of the GEMMs instead and only compute portion of the GEMM that is useful instead.

umangyadav:
Fixes #3267 

Didn't add verify tests since they are already covered through GRU/RNN tests.


src/simplify_algebra.cpp
```diff
@@ -266,6 +266,84 @@ struct find_mul_dot
     }
 };
 
+struct find_dot_slice
```
CharlieL7, line 269: Minor: It might be more descriptive to name this matcher `find_dot_slice_used_once`.

src/simplify_algebra.cpp
```diff
@@ -266,6 +266,84 @@ struct find_mul_dot
...
+        auto starts    = slice_op["starts"].to_vector<int64_t>();
+        auto ends      = slice_op["ends"].to_vector<int64_t>();
+        assert(starts.size() == ends.size() and starts.size() == axes.size());
+        auto has_neg_vals = [](auto vec) {
+            return std::any_of(vec.begin(), vec.end(), [](auto i) { return i < 0; });
+        };
+        if(has_neg_vals(starts) or has_neg_vals(ends) or has_neg_vals(axes))
+        {
+            return;
+        }
```
CharlieL7, line 292: Maybe this should be a throw or a warning because having a negative value here would probably be a problem.

CharlieL7 (approved):

test/simplify_algebra_test.cpp
```diff
@@ -4081,6 +4081,163 @@ TEST_CASE(dot_mul_b_non_const)
...
+TEST_CASE(dot_slice_not_applicable_2)
+{
+    migraphx::shape as{migraphx::shape::float_type, {2, 256, 32}};
+    migraphx::shape bs{migraphx::shape::float_type, {2, 32, 128}};
+    migraphx::module m1;
+    {
+        auto a     = m1.add_parameter("a", as);
+        auto b     = m1.add_parameter("b", bs);
+        auto dot   = m1.add_instruction(migraphx::make_op("dot"), a, b);
+        auto slice = m1.add_instruction(
```
shivadbhavsar, line 4230: In theory, the result of this should be the same as `dot_slice_ab` case right? Would it ever get to this pass without normalized axes?

shivadbhavsar (approved):

src/simplify_algebra.cpp
umangyadav: "Dot" should have one use only. 
Not the slice. it is not clear from the name `find_dot_slice_used_once`. We don't have "used_once"  inside names of any matchers

src/simplify_algebra.cpp
umangyadav: We have `normlize_ops` pass which runs at the beginning which would convert all the neg axes to positive ones. 
Therefore i think retuning simply is okay.

CharlieL7: I think it should print a warning if possible, otherwise if we mess something up wrt. `normalize_ops` this could be passed over without us knowing.

umangyadav: made changes to use "normalized_operator" and also enabled test for neg vals.

PR #3278, opened 2024-07-15
update python module versioning for dev builds

shivadbhavsar:
Currently when we do a local build, the python version is just set to "dev". Ie:
```
>>> import migraphx
>>> migraphx.__version__
'dev'
```
This is not very useful. This PR uses the version header file to populate this field properly for dev builds. Refer to https://packaging.python.org/en/latest/discussions/versioning/ for official packaging library recommendation for versioning dev builds/releases.

### DIFF
--- a/src/py/migraphx_py.cpp
+++ b/src/py/migraphx_py.cpp
@@ -613,8 +613,13 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
 #ifdef VERSION_INFO
     m.attr("__version__") = VERSION_INFO;
 #else
-    m.attr("__version__") = std::to_string(MIGRAPHX_VERSION_MAJOR) + "." +
-                            std::to_string(MIGRAPHX_VERSION_MINOR) + "." +
-                            std::to_string(MIGRAPHX_VERSION_PATCH) + ".dev";
+    auto version_string = std::to_string(MIGRAPHX_VERSION_MAJOR) + "." +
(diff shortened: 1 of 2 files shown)


causten (approved):

umangyadav (approved):

src/py/migraphx_py.cpp
```diff
@@ -612,6 +613,8 @@ MIGRAPHX_PYBIND11_MODULE(migraphx, m)
 #ifdef VERSION_INFO
     m.attr("__version__") = VERSION_INFO;
 #else
-    m.attr("__version__") = "dev";
+    m.attr("__version__") = std::to_string(MIGRAPHX_VERSION_MAJOR) + "." +
```
umangyadav, line 616: Inside `migraphx-driver` it also prints SHA with `MIGRAPHX_VERSION_TWEAK`
It has been useful for debugging.  We can consider adding that but that would be deviation from https://packaging.python.org/en/latest/discussions/versioning/

umangyadav: You can also do something like “0.5.dev1+gd00980f”
https://packaging.python.org/en/latest/discussions/versioning/#local-version-identifiers

shivadbhavsar: oh ya, I'll update it to add tweak in that form

umangyadav (approved):

CharlieL7 (approved):

Issue #3212, opened 2024-06-21
Fuse reductions with MLIR with multi-outputs


pfultz2:
This issue has two parts.

The first part is to fuse reductions(including split reductions) with MLIR, including any pointwise.

The second part is to use multiple outputs when fusing, to fuse even when its used multiple times. So we may have something like `convoluttion -> add -> reduce` but the `add` maybe used more than once so we will need to output that computation along with the output of reduce.

We should also handle multiple reductions as well.

pfultz2:
The goal of this is to fuse layernorm with with two convs/gemms. In #3010, it will fuse whats after the reductions with the following convolution, but we still need to fuse the reduction step. In #3097, it will spit the reduction, but we need to then fuse it with mlir. 

Here is an example from unet where we want to do the fusion:

```python
p = migraphx.program()
mmain = p.get_main_module()
x_main_module_0 = mmain.add_literal(migraphx.create_argument(migraphx.shape(type="float_type", lens=[1]), [1e-4]))
x_sample = mmain.add_parameter("sample",migraphx.shape(type="float_type", lens=[2,4,64,64]))
x_main_module_4 = mmain.add_literal(migraphx.generate_argument(migraphx.shape(type="float_type", lens=[320,1,1]), 1))
x_main_module_5 = mmain.add_literal(migraphx.generate_argument(migraphx.shape(type="float_type", lens=[320,1,1]), 2))
x_main_module_6 = mmain.add_literal(migraphx.generate_argument(migraphx.shape(type="float_type", lens=[320,320,3,3]), 3))
x_main_module_7 = mmain.add_literal(migraphx.generate_argument(migraphx.shape(type="float_type", lens=[320]), 4))
x_main_module_8 = mmain.add_literal(migraphx.generate_argument(migraphx.shape(type="float_type", lens=[320,4,3,3]), 5))
x_main_module_9 = mmain.add_instruction(migraphx.op("convolution", padding=[1,1,1,1]), [x_sample, x_main_module_8])
x_main_module_10 = mmain.add_instruction(migraphx.op("broadcast", axis=1, out_lens=[2,320,64,64]), [x_main_module_7])
x_main_module_11 = mmain.add_instruction(migraphx.op("add"), [x_main_module_9, x_main_module_10])
x_main_module_12 = mmain.add_instruction(migraphx.op("contiguous"), [x_main_module_11])
[...]

PR #3280, opened 2024-07-16
Fuse mlir_op and pointwise across reshapes

### LINKED ISSUE #2813: Fuse GEMM across a reshape
From the 22 Feb 2024 performance model review of Distilgpt2:

There are several cases of dot+reshape+pointwise:

```
@47 = gpu::code_object[code_object=5688,symbol_name=mlir_reshape_dot,global=67584,local=256,](@44,@45,@46) -> half_type, {348, 3072}, {3072, 1}
@48 = hip::hip_copy_literal[id=main:@literal:66] -> half_type, {3072}, {1}
@49 = reshape_lazy[dims={1, 348, 3072}](@47) -> half_type, {1, 348, 3072}, {1069056, 3072, 1}
@50 = load[offset=1069056,end=3207168](@1) -> half_type, {1, 348, 3072}, {1069056, 3072, 1}
@51 = multibroadcast[out_lens={348, 3072},out_dyn_dims={}](@48) -> half_type, {348, 3072}, {0, 1}
@52 = reshape_lazy[dims={1, 348, 3072}](@51) -> half_type, {1, 348, 3072}, {0, 0, 1}
@53 = gpu::code_object[code_object=9624,symbol_name=add_mul_mul_mul_mul_add_neg_sub_exp_add_div_mul_kernel,global=534528,local=1024,](@49,@52,@50) -> half_type, {1, 348, 3072}, {1069056, 3072, 1}
```

And also here:

```
@32 = gpu::code_object[code_object=5520,symbol_name=mlir_transpose_reshape_dot,global=18432,local=256,](@29,@30,@31) -> half_type, {348, 768}, {768, 1}
@33 = hip::hip_copy_literal[id=main:@literal:70] -> half_type, {768}, {1}
@34 = load[offset=534528,end=1069056](@1) -> half_type, {1, 348, 768}, {267264, 768, 1}
@35 = multibroadcast[out_lens={348, 768},out_dyn_dims={}](@33) -> half_type, {348, 768}, {0, 1}
@36 = reshape_lazy[dims={1, 348, 768}](@32) -> half_type, {1, 348, 768}, {267264, 768, 1}
@37 = reshape_lazy[dims={1, 348, 768}](@35) -> half_type, {1, 348, 768}, {0, 0, 1}
@38 = gpu::code_object[code_object=9416,symbol_name=add_add_kernel,global=133632,local=1024,](@36,@37,@10,@34) -> half_type, {1, 348, 768}, {267264, 768, 1}
```

The dot is only used once so we could fuse this with MLIR.
MLIR should already handle a reshape as it already fuses them on the inputs to dot, but we could do some tweaking in migraphx to fuse it when its inbetween the dot and pointwise.

Deliverables:
* Need to write a matcher in MIGX that will handle this case of fusing dot -> reshape -> pointwise to pass to MLIR.

umangyadav:
Currently MLIR pointwise fusion happens only for the cases where pointwise op appears immediately after Dot/conv. 

This PR makes changes to allow certain reshapes operations in between Dot/Conv and pointwise op.
Part of #3212 
Fixes #2813


test/gpu/fuse_mlir.cpp
```diff
@@ -66,6 +66,48 @@ migraphx::instruction_ref add_mlir(migraphx::program& p,
...
+            {x, a, b},
+            {"x2", "y0", "y1"},
+            [=](auto* pm, const auto& inputs) {
+                auto dot = pm->add_instruction(migraphx::make_op("dot"), inputs[1], inputs[2]);
+                auto dot_trans = pm->add_instruction(
+                    migraphx::make_op("transpose", {{"permutation", {0, 2, 1}}}), dot);
+                auto dot_rsp = pm->add_instruction(migraphx::make_op("reshape", {{"dims", {3, 3}}}),
+                                                   dot_trans);
+                auto add     = pm->add_instruction(migraphx::make_op("add"), dot_rsp, inputs[0]);
+                return std::make_tuple(dot, add);
```
CharlieL7, line 104: Why does this return a tuple of `dot` and `add`?

umangyadav: That's how it's used inside utility `add_mlir`. 

First element of the tuple is the core MLIR op. second element of the tuple is the return instruction for the MLIR module.

src/module.cpp
```diff
@@ -901,7 +901,10 @@ generic_split(const module& m,
                    splits.end(),
                    std::back_inserter(outputs),
                    [&](instruction_ref ins) { return map_ins1.at(ins); });
-    m1.add_return(outputs);
+    if(not outputs.empty())
+    {
+        m1.add_return(outputs);
+    }
```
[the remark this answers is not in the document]
umangyadav: It could also occur when `split` instruction is not found in the module that is being split. 
For example : 
case : {gemm + reshape + pointwise} --> in this case original module is first split with gemm instruction as the split.
so it becomes {gemm} + {reshape + pointwise} and then 
{reshape + pointwises} is split with reshape_ins as the split instruction. 

But for the case where we only have `{gemm + pointwise}` in this case 

First split would do `{gemm} + {Pointwise}` later second split would try to split `{pointwise}` with `std::prev{pointwise}` as the split instruction that would be `gemm` and it won't be found in `{pointwise}`.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -165,21 +165,30 @@ MIGRAPHX_REGISTER_OP(mlir_op);
 
 namespace {
 
+const auto& reshaper_names()
```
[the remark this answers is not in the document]
umangyadav: Both have same effect.  I changed simplify_reshapes to use anonymous namespace.

src/targets/gpu/jit/mlir.cpp
```diff
@@ -171,9 +183,58 @@ struct mlir_compiler : compiler<mlir_compiler>
...
+                             &inputs_rep_map,
+                             [](module& insert_mod,
+                                instruction_ref insert_loc,
+                                const operation& op,
+                                const std::vector<instruction_ref>& inputs,
+                                const std::vector<module_ref>& mod_args) -> instruction_ref {
+                                 if(op.name() == "reshape")
+                                 {
+                                     // TODO: Add proper lowering for the reshape op
+                                     return insert_mod.insert_instruction(
```
pfultz2, line 210: Should we check if the input is standard and then insert a contiguous?

umangyadav: `reshape_lazy` can work some limited non-standard shapes too. 

Proper fix need to emulate what we do with `eliminate_contiguous`  that is try compute_shape() and if that throws an exception then only add `gpu::contiguous`. 

I added an `assert` for now

src/targets/gpu/jit/mlir.cpp
```diff
@@ -79,29 +79,33 @@ struct mlir_compiler : compiler<mlir_compiler>
...
+        // check if (a) module is fused (b) contains a "gemm/conv" instruction and (c)
+        // perfConfig can not allow fused module
+        if(gemm_like_ins != smod->end() and pointwise_ins != smod->end() and
            not is_module_fusible(*smod, ctx, solution))
         {
             auto input_args = ins->inputs();
+            // remove alloc buffer
             input_args.pop_back();
-            auto mod_splits      = smod->split(input_args, {gemm_like_ins});
+            auto split_ins = std::prev(pointwise_ins);
```
umangyadav, line 94: @pfultz2  fusing reshapes with Dot is working. I removed changes for lowering reshapes module in between.

src/module.cpp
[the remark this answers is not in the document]
umangyadav: Reverted these changes. Marking this comment as resolved.

CharlieL7 (approved):

shivadbhavsar (approved):

PR #3285, opened 2024-07-17
adjust stride ordering rules for standard shape: stride can be anythi…

bpickrel:
…ng in a dimension of size 1.  This is for issue https://github.com/ROCm/AMDMIGraphX/issues/3117


test/shape_test.cpp
```diff
@@ -526,7 +537,7 @@ TEST_CASE(test_shape_broadcasted)
 TEST_CASE(test_shape_broadcasted2)
 {
     migraphx::shape s{migraphx::shape::float_type, {1, 2}, {0, 1}};
-    EXPECT(not s.standard());
+    EXPECT(s.standard());
```
shivadbhavsar, line 540: I think we dont allow a shape to be both standard and broadcasted as far as I've seen. Maybe problematic? @umangyadav might know better

umangyadav: Agree with shiv here

bpickrel:
Confirmed it also works on MI300 for good measure, although this should have been GPU-independent.

test/shape_test.cpp
bpickrel: Then I could put a check for broadcasted in the same spot in code.  Is that what we want?

If a shape being both standard and broadcasted is a problem, I wonder why the change didn't cause something else to fail.

bpickrel: I just found an existing test case that is both standard and broadcasted:
```
TEST_CASE(test_shape_scalar2)
{
    migraphx::shape s{migraphx::shape::float_type, {1}, {0}};
    EXPECT(s.standard());
    EXPECT(s.packed());
    EXPECT(not s.transposed());
    EXPECT(s.broadcasted());
}
```

shivadbhavsar: Hmm ok, seems like its allowed for scalars. 
Off the top of my head Im not sure where/if this is problematic. Might want to confirm with Paul when hes back

bpickrel:
Added @pfultz2 Paul, the question is whether it ever makes sense to define a shape as both standard and broadcast, if the 0-length dimension is on an axis we're ignoring for the `standard` check.  The only test case we know of that has both standard and broadcast properties is a scalar.

pfultz2:
> the question is whether it ever makes sense to define a shape as both standard and broadcast

Yes it does make sense. "Standard" shape basically means that the offset and index are the same.

shivadbhavsar (approved):

umangyadav (approved):

pfultz2 (approved):

causten:
@bpickrel   CI Failure...

```
[2024-07-25T00:26:40.785Z] [   RUN    ] test_shape_alloc
[2024-07-25T00:26:40.785Z] test_verify: /home/jenkins/workspace/AMDMIGraphX_PR-3285/src/program.cpp:511: std::vector<argument> migraphx::generic_eval(const module *, std::vector<context> &, std::unordered_map<std::string, argument>, std::unordered_map<instruction_ref, argument>, F) [F = (lambda at /home/jenkins/workspace/AMDMIGraphX_PR-3285/src/program.cpp:530:57)]: Assertion `ins->get_shape().any_of_dynamic() or results.at(ins).get_shape() == ins->get_shape()' failed.

[2024-07-25T00:26:40.785Z] CMake Error at gdb/test_test_verify_conv/run.cmake:16 (message):
[2024-07-25T00:26:40.785Z]   Test failed
```

causten (changes requested):
fix CI failure and then it can be merged

bpickrel:
Not an obvious fix to test failure--back to draft status.

src/include/migraphx/shape.hpp
```diff
@@ -296,6 +296,11 @@ struct MIGRAPHX_EXPORT shape
     /// not transposed.
     bool standard() const;
 
+    /// Returns true if the shapes are compatible.  TODO: better description
+    // Paul, How would you describe the purpose of that equality check for the shapes?  I'm trying to come up with a 
+    // function description that explains why it's ok for the strides not to match sometimes.
+    bool compatible_lens(const shape& s2) const;
```
pfultz2, line 303: The function should probably be in program.cpp file since its only used for that. Either way it should be a free function so we can make clear which is the actual shape and which is the expected shape. And should be called `is_compatible_shape`.

src/shape.cpp
```diff
@@ -490,6 +490,26 @@ bool shape::scalar() const
 
 bool shape::standard() const { return impl->m_standard; }
 
+
+bool shape::compatible_lens(const shape& s2) const
+{
+    if(dynamic() or s2.dynamic()) return true;
```
pfultz2: This needs to account for sub_shapes.

src/shape.cpp
```diff
@@ -490,6 +490,26 @@ bool shape::scalar() const
...
+    
+    // Lens must be the same; strides must be same except that
+    // axes with len=1 don't matter
+    for(size_t ind = 0; ind < lens().size(); ind++)
+    {
+        size_t l_ind(lens()[ind]);
+        if(l_ind != s2.lens()[ind] or
+          (l_ind != 1 and strides()[ind] != s2.strides()[ind]))
+            return false;
+    }
```
pfultz2, line 507: No need to for this loop, if its both standard shape then just check that the lens are equal otherwise return false:

```cpp
static bool is_compatible_shape(const shape& actual, const shape& expected)
{
    // Check subshapes
    if(expected.type() == shape::tuple_type)
        return equal(actual.sub_shapes(), expected.sub_shapes(), &is_compatible_shape);
    // Only the expected can be dynamic
    if(expected.dynamic())
        return true;
    if(actual == expected)
        return true;
    if(actual.type() != expected.type())
        return false;
    if(actual.standard() and expected.standard())
        return actual.lens() == expected.lens();
    return false;
}
```

bpickrel: It's still not clear to me what you mean by "actual" and "expected"  here.  I need to know what order to place the arguments when calling this, since only `expected` can be dynamic.

bpickrel: Do we know that the recursion will always exit?  i.e. that `sub_shapes()` will always end with a shape that is not `tuple_type`?

pfultz2: > It's still not clear to me what you mean by "actual" and "expected" here.

The "actual" is the shape being used at runtime, and the expected is the computed shape in the IR. For the case of dynamic shape the comptued IR will have a dynamic shape, but at runtime it will be a static shape, so we should just check that the type matches(and probably the ndim as well).

src/program.cpp
```diff
@@ -507,8 +526,10 @@ std::vector<argument> generic_eval(const module* mod,
                 }));
         }
         assert(results.find(ins) != results.end());
+        // TODO: what order do the arguments to is_compatible_shape() come in?  One
+        // can be dynamic.
         assert(ins->get_shape().any_of_dynamic() or
-               results.at(ins).get_shape() == ins->get_shape());
```
pfultz2, line 511: Remove the `ins->get_shape().any_of_dynamic()` as the function checks dynamic shapes.

src/program.cpp
```diff
@@ -507,8 +526,10 @@ std::vector<argument> generic_eval(const module* mod,
...
+               is_compatible_shape(ins->get_shape(), results.at(ins).get_shape()));
```
pfultz2, line 532: Args should be flipped: `is_compatible_shape(results.at(ins).get_shape(), ins->get_shape()))`.

bpickrel:
Don't re-review until I get current build fails fixed.

src/program.cpp
```diff
@@ -426,6 +426,25 @@ void preview_argument(std::ostream& os, const argument& a)
         });
 }
 
+static bool is_compatible_shape(const shape& actual, const shape& expected)
```
umangyadav, line 429: since this function is  only used inside `assert` it will not have any uses when doing Release builds. 
`assert` are only enabled for the Debug builds. 

You can do a fake call to `is_compatible_shape()` to avoid build error 
`(void)(is_compatible_shape(shape{}, shape{})`

bpickrel: Is there any problem with wrapping it in a `#ifndef NDEBUG` instead?

umangyadav: you can do that too. That's valid solution as well

src/program.cpp
bpickrel: done

bpickrel:
Now seeing a test fail in this code:
```
    /*!
     * Check all shapes have the same layout.
     */
    const check_shapes& same_layout() const
    {
        if(not this->same([](const shape& s) { return find_permutation(s); }))
            MIGRAPHX_THROW(prefix() + "Layouts do not match");
        return *this;
    }
```

causten (approved):

src/include/migraphx/check_shapes.hpp
```diff
@@ -235,11 +235,14 @@ struct check_shapes
     }
 
     /*!
-     * Check all shapes have the same layout.
+     * Check all shapes have the same layout, with minor differences allowed.
      */
-    const check_shapes& same_layout() const
+    const check_shapes& compatible_layout() const
```
pfultz2, line 240: Let not make the changes here. This might cause problems with MIOpen. Instead we should just update auto_contiguous to insert contiguous even if its standard and not transposed instead:

```cpp
        if(not s.dynamic() and not s.standard() and not s.transposed() and s.elements() != 0)
        {
            auto c = m.insert_instruction(std::next(ins), make_op("contiguous"), ins);
            m.replace_instruction(ins, c);
        }
```

Which comes from [here](https://github.com/ROCm/AMDMIGraphX/blob/49fd94aff22084f8848886a7d058a43670e8afbc/src/auto_contiguous.cpp#L67).

src/include/migraphx/shape.hpp
```diff
@@ -431,6 +431,30 @@ struct MIGRAPHX_EXPORT shape
...
+    if(actual == expected)
+        return true;
+    if(actual.type() != expected.type())
+        return false;
+    // If both shapes are standard and lens match, they are considered compatible
+    // even if strides are different.
+    if(actual.standard() and expected.standard())
+        return actual.lens() == expected.lens();
+    return false;
+}
```
pfultz2, line 456: THis should just go into the program.cpp file.

src/include/migraphx/check_shapes.hpp
bpickrel: After some discussion, I reverted as @pfultz2 requested but with a correction to the Line 67 logic above.  Paul, please note that if this check is done incorrectly and a `contiguous`  gets added at the wrong time, it can result in an infinite loop.  You can see an example of this in commit `97a0b7e5b9a1bdcf223d680febd1dc99be9d6d1b` which also has a descriptive comment at this same point.

src/include/migraphx/shape.hpp
bpickrel: done, part of the above request.

src/auto_contiguous.cpp
```diff
@@ -64,7 +64,10 @@ void auto_contiguous::apply(module& m) const
         if(ins->outputs().empty() and ins != last)
             continue;
         shape s = ins->get_shape();
-        if(not s.dynamic() and not s.standard() and s.elements() != 0)
+        // If s is not standard layout or has out of sequence strides, insert "contiguous" op
+        // to make a standard shape
+        if(not s.dynamic() and (not s.transposed() or s.normalize_standard() != s) and
```
pfultz2, line 69: This wrong, it should be `not s.standard() or s.normalize_standard() != s`, that is replace `transposed` with `standard`.

bpickrel: Blame it on a garbled `git merge`.  Fixed.

pfultz2: Actually, thinking about it more, we could probably simplify this to just `s.as_standard() != s`, so the entire `if` statement becomes `if(not s.dynamic() and s.as_standard() != s and s.elements() != 0)`.

bpickrel: Ran `make check` successfully on my machine but somehow in the commit and CI process here on Github, `test_ref` failed for both variations, both times in non-obvious ways.  Back to the drawing board for now.

pfultz2: Remove `not s.transposed()` because we want to insert contiguous when its transposed. 

Also to fix the tests we need to skip contiguous for single scalars, so we can change the condition from `s.elements() != 0` to `s.elements() > 1`, that should fix the issue.

So the final condition should look like: `if(not s.dynamic() and s.as_standard() != s and s.elements() > 1)`.

bpickrel: Tried it, but I'm still seeing multiple test failures with the same message, for instance:
```
[   RUN    ] test_select_module_conv
test_verify: /workspace/AMDMIGraphX/src/shape.cpp:57: migraphx::shape_impl::shape_impl(shape::type_t, std::vector<std::size_t>): Assertion `t != shape::tuple_type' failed.
```
Not sure what we'd change to fix that.

bpickrel: Update: Also a few tests failed with this message:
```
 34/357 Test #351: test_api_gpu ..............................................................Subprocess aborted***Exception:   0.49 sec
[   RUN    ] load_and_run
MIGraphX Error: /workspace/AMDMIGraphX/src/onnx/onnx_parser.cpp:407: parse_graph: PARSE_GRAPH: invalid onnx file. Input "0" is unavailable due to unordered nodes!
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to call function
```
Can an `onnx_parser` error be caused by a bug in an `auto_contiguous` pass?

src/auto_contiguous.cpp
```diff
@@ -64,7 +64,10 @@ void auto_contiguous::apply(module& m) const
         if(ins->outputs().empty() and ins != last)
             continue;
         shape s = ins->get_shape();
-        if(not s.dynamic() and not s.standard() and s.elements() != 0)
+        // If s is not standard layout or has out of sequence strides, insert "contiguous" op
+        // to make a standard shape
+        if(not s.dynamic() and (not s.standard() or s.normalize_standard() != s) and
+            s.elements() != 0)
```
pfultz2, line 70: This needs to be `s.elements() > 1` to fix the tests.

bpickrel: I made that comment after having made and tested that change on my machine but I didn't push it.  I just pushed the change so you can see for yourself that it still fails.

src/auto_contiguous.cpp
```diff
@@ -64,7 +64,9 @@ void auto_contiguous::apply(module& m) const
         if(ins->outputs().empty() and ins != last)
             continue;
         shape s = ins->get_shape();
-        if(not s.dynamic() and not s.standard() and s.elements() != 0)
+        // If s is not standard layout or has out of sequence strides, insert "contiguous" op
+        // to make a standard shape
+        if(not s.dynamic() and (s.as_standard() != s) and s.elements() > 1)
```
pfultz2, line 69: Instead of `s.as_standard() != s`, we probably need to make it `not s.standard() or s.normalize_standard() != s` like it was before, that should avoid the `t != shape::tuple_type` assertion failures.

bpickrel: That cleared most of the errors, but it's still failing in onnx parsing.  Investigating...

src/include/migraphx/check_shapes.hpp
```diff
@@ -332,7 +332,7 @@ struct check_shapes
      */
     const check_shapes& not_broadcasted() const
     {
-        if(not this->all_of([](const shape& s) { return not s.broadcasted(); }))
+        if(not this->all_of([](const shape& s) { return s.standard() or not s.broadcasted(); }))
```
pfultz2, line 335: I think this change is fine.

bpickrel: That solved some errors, but in some of the perf tests I'm now seeing: `RuntimeError: /src/AMDMIGraphX/src/targets/gpu/gemm_impl.cpp:85: blas_shape: GPU_GEMM: matrix dimensions can't be broadcasted` with this change.  Not sure how to properly correct that check.

pfultz2: What test does that fail on?

pfultz2: I wonder why we are not inserting a contiguous here.

bpickrel: > What test does that fail on?

You can see them by going to the Conversation tab of this PR and looking for the big red X icons in the perf. test results.  All of the tests that threw errors had this same error.  They are:
```
torchvision-resnet50_1
cadene-resnext101_1
resnet50v1_fp16
resnet50v1_int8
```

bpickrel: > I wonder why we are not inserting a contiguous here.

In `check_shapes`?  That doesn't seem a very appropriate place to do it.

pfultz2: No, contiguous is inserted in auto_contiguous for when the shapes are not standard(or not the same as the normalize_standard shape). Then after lowering in eliminate_contiguous, we try to remove the contiguous, but only if it doesnt throw an error like this one, which means there should be a contiguous inserted to prevent this error, but somehow it escaped. 

We should try to get a backtrace to see where this happens, or run it with MIGRAPHX_TRACE_COMPILE=1 to see which pass it fails on.

bpickrel: I've been trying to replicate the failures on `aus-navi3x-04.amd.com` and can't--commands such as `MIGRAPHX_TRACE_COMPILE=1 bin/driver verify --onnx  ../../resnext101_64x4di1.onnx` for instance, handle the same model files just fine.  I can't troubleshoot the location of the problem until I can replicate it.  Any ideas on how to track down what's different?

pfultz2: But what about the failures for resnet50 models? Can you reproduce those?

Also, these are a compile failure so you just do `./bin/driver compile` instead of verify(that should be much quicker).

pfultz2: > No, I just tried those again, and they don't fail either.

Ok, I think you need to try on MI series. The navi always uses MLIR which mean rocblas is never used and this error comes from rocblas.

> if I'm reading the aliasing in the script files correctly--tests with resnet50v1_fp16 and resnet50v1_int8 both use the same model, resnet50/resnet50_v1.onnx but with command line arguments --fp16 and --int8 respectively.

Yes.

bpickrel:
Took some time but replicated the fail on rocm-rome-6.  Examining the output now.  Output file:
[rocblas_fail.txt](https://github.com/user-attachments/files/16937199/rocblas_fail.txt)

pfultz2:
> Took some time but replicated the fail on rocm-rome-6. Examining the output now. Output file:

So the failure is in fuse_ops, which means we have a bug somewhere where it is changing the shape.

bpickrel:
> > Took some time but replicated the fail on rocm-rome-6. Examining the output now. Output file:
> 
> So the failure is in fuse_ops, which means we have a bug somewhere where it is changing the shape.
> 
> A backtrace will help find out which matcher this is happening.

I think I found it; just pushed for testing

src/program.cpp
bpickrel: Marked this conversation unresolved again because I thought codecov would not flag debug-only code lines 446 and 451.  Why does it want this assertion code to be covered by a test?

bpickrel:
Late followup:  I repeated the perf test for resnet50v1_int8 on server `banff-200-c2-1` (MI-300) with both this branch (commit 57b37f6c4) and previous develop (commit 990071c73).  There is still a significant speed deficit with this branch.

```
stride_ordering_for_mlir Rate: 1313.84 inferences/sec
develop                  Rate: 1412.36 inferences/sec

```

PR #3286, opened 2024-07-18
remove splitk support

causten:
The splitk feature grew the time it took for --exhaustive-tuning to hours.  Backing out until a better solution is developed


umangyadav:
I would suggest gating split-k with an env flag. We need to continue testing Split-k in CI.

umangyadav (approved):

PR #3289, opened 2024-07-19
Maxpool. Kernel of size 1: stride being ignored

lakhinderwalia:
A Resnet50 model isn't being verified properly. The error stems from a JIT maxpool kernel.

### DIFF
--- a/src/include/migraphx/op/pooling.hpp
+++ b/src/include/migraphx/op/pooling.hpp
@@ -331,7 +331,7 @@ struct pooling
                 int end;
                 std::size_t dilated_kernel_dim = dilate_dim(kernel_dims[d_2], dilations[d_2]);
                 // NOLINT
-                if(count_include_pad and (mode != pooling_mode::max))
+                if(count_include_pad and mode == pooling_mode::average)
                 {
                     // Even when using padding, if in ceil_mode a window
                     // could extend beyond the end of both input and
--- a/src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
+++ b/src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
@@ -131,10 +131,8 @@ struct window
     constexpr void visit(Index i, F f) const
     {
         auto win_start = generate_array<diff_int>(rank{}, [&](auto j) {
-            diff_int w   = win[j];
             diff_int dim = i[j];
-            (void)w;
-            MIGRAPHX_ASSERT(w >= 1);
+            MIGRAPHX_ASSERT(win[j] >= 1);
             diff_int s = stride[j];
             diff_int p = padding[j];
             return (dim * s) - p;
--- a/src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
+++ b/src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
@@ -133,7 +133,7 @@ struct window
         auto win_start = generate_array<diff_int>(rank{}, [&](auto j) {
             diff_int w   = win[j];
             diff_int dim = i[j];
-	    (void) w;
+            (void)w;
             MIGRAPHX_ASSERT(w >= 1);
             diff_int s = stride[j];
             diff_int p = padding[j];
--- a/src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
+++ b/src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
@@ -133,9 +133,8 @@ struct window
         auto win_start = generate_array<diff_int>(rank{}, [&](auto j) {
             diff_int w   = win[j];
             diff_int dim = i[j];
+	    (void) w;
             MIGRAPHX_ASSERT(w >= 1);
-            if(w == 1)
-                return dim;
             diff_int s = stride[j];
             diff_int p = padding[j];
             return (dim * s) - p;
--- a/src/targets/gpu/lowering.cpp
+++ b/src/targets/gpu/lowering.cpp
@@ -303,7 +303,8 @@ struct miopen_apply
         apply_map.emplace("pooling", [=](instruction_ref ins) {
             auto&& op   = ins->get_operator();
             auto op_val = op.to_value();
-            if(op_val.at("count_include_pad").to<bool>())
+            if(op_val.at("count_include_pad").to<bool>() and
+               op_val["mode"].to<op::pooling_mode>() == op::pooling_mode::average)
             {
                 return insert_precompile_op(ins);
             }
--- a/test/verify/test_max_pooling_1x1_s2.cpp
+++ b/test/verify/test_max_pooling_1x1_s2.cpp
@@ -1,0 +1,48 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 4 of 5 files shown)


src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
```diff
@@ -133,9 +133,8 @@ struct window
         auto win_start = generate_array<diff_int>(rank{}, [&](auto j) {
             diff_int w   = win[j];
             diff_int dim = i[j];
+	    (void) w;
```
umangyadav, line 136: FYI, testing of JIT pooling in CI disabled. 

https://github.com/ROCm/AMDMIGraphX/issues/3234

lakhinderwalia: Thanks. This is already being verified by forcing it manually.

umangyadav: > This is already being verified by forcing it manually -- in a separate build. 
Where ?

src/include/migraphx/op/pooling.hpp
```diff
@@ -331,7 +331,7 @@ struct pooling
                 int end;
                 std::size_t dilated_kernel_dim = dilate_dim(kernel_dims[d_2], dilations[d_2]);
                 // NOLINT
-                if(count_include_pad and (mode != pooling_mode::max))
+                if(count_include_pad and mode == pooling_mode::average)
```
umangyadav, line 334: This would disable count_include_pad for the lp_norm

lakhinderwalia: https://onnx.ai/onnx/operators/onnx__LpNormalization.html

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
lakhinderwalia: Sorry, I meant in a private build. This is being tested by disabling this clause: https://github.com/ROCm/AMDMIGraphX/blob/b8c33b0754041cd61a0bd41396a1838b7c88a8c7/src/targets/gpu/lowering.cpp#L314, to turn on JIT kernels for pooling. So this will test all the pooling modes. `make check` and `test_verify` have been run successfully.

test/verify/test_max_pooling_1x1_s2.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+struct test_max_pooling_1x1_s2 : verify_program<test_max_pooling_1x1_s2>
+{
+    migraphx::program create_program() const
+    {
+        migraphx::program p;
+        auto* mm = p.get_main_module();
+
+        auto input =
+            mm->add_parameter("x", migraphx::shape{migraphx::shape::float_type, {1, 4, 112, 112}});
```
umangyadav, line 39: You can probably make dimensions smaller for the unit-test to not slow down verify test.

umangyadav (approved):

test/verify/test_max_pooling_1x1_s2.cpp
lakhinderwalia: Done.

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
```diff
@@ -133,9 +133,8 @@ struct window
         auto win_start = generate_array<diff_int>(rank{}, [&](auto j) {
             diff_int w   = win[j];
             diff_int dim = i[j];
+            (void)w;
             MIGRAPHX_ASSERT(w >= 1);
-            if(w == 1)
-                return dim;
```
pfultz2, line 138: Wouldn't it be better to change this to `return dim * s` instead of removing the condition?

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
```diff
@@ -133,9 +133,8 @@ struct window
         auto win_start = generate_array<diff_int>(rank{}, [&](auto j) {
             diff_int w   = win[j];
             diff_int dim = i[j];
+            (void)w;
```
umangyadav, line 136: instead of adding `(void)(w)` it might be better to do 
`MIGRAPHX_ASSERT(win[j] >= 1)`

lakhinderwalia: In some build, where assert is not active, the void statement is necessary-- for compilation. Adding another assert won't help in that direction.

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
lakhinderwalia: If the rest of the following lines do the same thing.. why add an exception?

causten:
Would like to move to 6.2.1 as well

src/targets/gpu/kernels/include/migraphx/kernels/pooling.hpp
pfultz2: Umang is saying to remove the variable instead of using a void statement.

src/include/migraphx/op/pooling.hpp
pfultz2, line 334: Why is this change needed? The `count_include_pad` flag should only be skipped for max since it doesn't use an average of the pixel values.

src/targets/gpu/lowering.cpp
```diff
@@ -303,7 +303,8 @@ struct miopen_apply
         apply_map.emplace("pooling", [=](instruction_ref ins) {
             auto&& op   = ins->get_operator();
             auto op_val = op.to_value();
-            if(op_val.at("count_include_pad").to<bool>())
+            if(op_val.at("count_include_pad").to<bool>() and
+               op_val["mode"].to<op::pooling_mode>() == op::pooling_mode::average)
```
pfultz2, line 307:
```suggestion
            if(op_val.at("count_include_pad").to<bool>() and
               op_val["mode"].to<op::pooling_mode>() != op::pooling_mode::max)
```

src/include/migraphx/op/pooling.hpp
lakhinderwalia: Per Onnx specfications, count_include_pad as an operator parameter is used only with pooling_mode::average. Not with other pooling modes, e.g. max or lpnorm.

pfultz2: This is not an onnx operator. We have this attribute on this operator and it should work for all pooling modes that use a count.

pfultz2: Ah wait, lpnorm doesnt use the count. Nevermind.

bpickrel (changes requested):
see earlier comment about the `count_include_pad` param

src/include/migraphx/op/pooling.hpp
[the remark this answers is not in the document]
pfultz2: lpnorm and max pooling doesn't use a count so this flag is not applicable. You would get the same result if it were true or false.

pfultz2 (approved):

src/include/migraphx/op/pooling.hpp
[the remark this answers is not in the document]
lakhinderwalia: (Thanks, @pfultz2 for responding to this. I had missed this comment from Brian.)

bpickrel (approved):
Assuming the tests pass, should be good.

PR #3290, opened 2024-07-19
ENV variable MIGRAPHX_VERIFY_DUMP_DIFF

lakhinderwalia:
Using this environment variable to dump large set of differing data if test_verify fails. Currently we only do it for upto 32 bytes max.

### DIFF
--- a/docs/dev/env_vars.rst
+++ b/docs/dev/env_vars.rst
@@ -314,3 +314,8 @@ Prints the reference and target programs even if the verify tests pass.
 
 Set to "1", "enable", "enabled", "yes", or "true" to use.
 Dumps verify tests to ``.mxr`` files.
+
+.. envvar:: MIGRAPHX_VERIFY_DUMP_DIFF
+
+Set to "1", "enable", "enabled", "yes", or "true" to use.
+Dumps the output of the test (and the reference) results when they differ.
--- a/src/verify_args.cpp
+++ b/src/verify_args.cpp
@@ -24,6 +24,8 @@
 
 #include <migraphx/verify_args.hpp>
 
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_VERIFY_DUMP_DIFF);
+
 namespace migraphx {
 inline namespace MIGRAPHX_INLINE_NS {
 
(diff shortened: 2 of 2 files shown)


src/verify_args.cpp
```diff
@@ -24,6 +24,8 @@
 
 #include <migraphx/verify_args.hpp>
 
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_VERIFY_DUMP_DIFF);
```
umangyadav, line 27: add documentation inside `env_var.rst`

lakhinderwalia: Thanks. Done.

docs/dev/env_vars.rst
```diff
@@ -314,3 +314,8 @@ Prints the reference and target programs even if the verify tests pass.
 
 Set to "1", "enable", "enabled", "yes", or "true" to use.
 Dumps verify tests to ``.mxr`` files.
+
+.. envvar:: MIGRAPHX_VERIFY_DUMP_DIFF
+
+Set to "1", "enable", "enabled", "yes", or "true" to use.
+Dumps the output of the test (and the reference) results when they differ.
```
randyh62, line 321: Maybe add that the default is 0, disabled, no, or false?

randyh62 (approved):

umangyadav (approved):

src/verify_args.cpp
```diff
@@ -42,9 +44,9 @@ bool verify_args(const std::string& name,
             // TODO: Check for nans
             std::cout << "FAILED: " << name << std::endl;
             std::cout << "RMS Error: " << rms_error << std::endl;
-            if(ref.size() < 32)
+            if(ref.size() < 32 or enabled(MIGRAPHX_VERIFY_DUMP_DIFF{}))
```
bpickrel, line 47: Suggestion with a little bit nicer formatting and making sure both outputs are displayed together:
```
            if((ref.size() < 32 and target.size() < 32) or enabled(MIGRAPHX_VERIFY_DUMP_DIFF{}))
            {
                std::cout << "ref: " << ref << std::endl << std::endl;
                std::cout << "target: " << target << std::endl;
            }
            else   
                std::cout << "(Use env MIGRAPHX_VERIFY_DUMP_DIFF to display non-matching results)" << std::endl;
```

PR #3291, opened 2024-07-21
update find_package to msgpack-cxx

acxz:
msgpack has changed its api in 6.0.0

related: https://github.com/ROCm/AMDMIGraphX/issues/1602 and https://github.com/ROCm/AMDMIGraphX/pull/1603


src/CMakeLists.txt
```diff
@@ -310,8 +310,8 @@ target_link_libraries(migraphx PRIVATE nlohmann_json::nlohmann_json)
 find_package(SQLite3 REQUIRED)
 target_link_libraries(migraphx PRIVATE SQLite::SQLite3)
 
-find_package(msgpackc-cxx QUIET)
-if(NOT msgpackc-cxx_FOUND)
+find_package(msgpack-cxx QUIET)
```
umangyadav, line 313: MIGraphX msgpack is pinned at 3.3.0
https://github.com/ROCm/AMDMIGraphX/blob/05c9d1ba5467e695d0b52839f5e95b8240fdb48a/requirements.txt#L28

6.0.0 API is not tested with MIGraphX in case there is any incompatibilities.

acxz: That is understood. By API I meant in context of the cmake find_package API.

acxz: Thanks for that link @umangyadav!

PR #3297, opened 2024-07-23
Add --use_rocm flag to Onnxrt build

TedThemistokleous:
Need to add the use_rocm flag in our build script as our Onnxruntime changes now require that each EP can be built separately or together.


umangyadav (approved):

causten:
Dont we need to just build/test our migx ep ?

TedThemistokleous:
> Dont we need to just build/test our migx ep ?

Some of the tests in DLM use ROCm EP + MIGraphX EP similar to how CUDA EP is used with Tensor RT EP. It also changes the functionality of what was previously expected. If we're okay with that then I can close this and not merge.

TedThemistokleous:
Closing this out after speaking with @causten . We want to speed up our CI and ROCm EP is tested by QA already for regressions off devop's build with both --use_rocm and --use_migraphx flags

PR #3299, opened 2024-07-23
Allow multiple outputs for the MLIR + Pointwise fusions 

umangyadav:
At present when doing `mlir + pointwise` fusion, It is not checking if "conv/gemm" has only use or not. 

Therefore for the contrived case, it can lead to pessimization where it is recomputing "dot/conv" again. 

This PR fixes that. 

Currently MLIR compilation is broken for the multiple outputs, Fix for that would come from rocMLIR. This PR wouldn't change after rocMLIR fix. https://github.com/ROCm/rocMLIR-internal/issues/1546


umangyadav:
> Will this handle cases like this?
> 
> ```
> ┌─┐   
> │a│   
> └┬┘   
> ┌▽──┐ 
> │b  │ 
> └┬─┬┘ 
>  │┌▽┐ 
>  ││c│ 
>  │└┬┘ 
>  │┌▽─┐
>  ││d │
>  │└┬─┘
> ┌▽─▽┐ 
> │e  │ 
> └───┘ 
> ```
> 
> Especially if `c` or `d` is not fusible with mlir.

I don't think that will be handled well even with existing logic on develop branch or with this PR. 

It is implictly assumed that only one of the input to the pointwise would be produced from GEMM/conv that is being fused.

pfultz2:
> I don't think that will be handled well even with existing logic on develop branch or with this PR.
> 
> It is implictly assumed that only ONE of the inputs to the pointwise op would be produced from GEMM/conv that is being fused. Rest must be produced by other ops.

What about if `b` is convolution and `c` and `e` are pointwise ops but `d` is not(like pooiling or something else)? Currently we dont check if `b` is used once so it would create two convolutions.

However, as I understand it, this PR would try to create multiple outputs instead. So would it fuse both `c` and `e` with the convolution? Or would it just fuse `c` and output the convolution to be used later?

umangyadav:
>  Or would it just fuse c and output the convolution to be used later? 

It will do this one.

umangyadav:
>  So would it fuse both c and e with the convolution? Or would it just fuse c and output the convolution to be used later? 

Following with same example where `b` is convolution `c` and `e` are poitnwise but `d` is a pooling. 

Whether `b + e` is matched first or `b + c` is matched first, it depends on how graph is traversed. 

if `b + e` is matched then it would produce in incorrect result or crash. That case is not handled .  Because as i mentioned earlier is that it is implictly assumed that only one of the inputs to the pointwise must be produced from Convolution. Rest must be produced from some other ops. 

If `b + c` is matched and fused then it would work as expected.

pfultz2:
> If b + c is matched and fused then it would work as expected. b would be part of multiple outputs and then later consumed by e . This PR would handle this case very well.

Are you able to add a test case for this scenario so we break this in the future?

umangyadav:
> Are you able to add a test case for this scenario so we dont break this in the future?

Both the unit-tests that i added are of this type.

umangyadav:
Verify test is failing here: fix for that is here #3316

pfultz2 (approved):

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -461,16 +461,51 @@ struct find_mlir_fused_ops
...
         param_map[x_ins] = prev_input; // this is to avoid adding parameter for gemm/conv reshaped
                                        // input to pointwise in new fused module
-        mm->add_return(mm->fuse(*pm, pw_ins->inputs(), &param_map));
+        bool gemm_has_multi_outs = gemm_based_op->outputs().size() > 1;
+        auto reshaped_gemm       = x_ins;
+        std::vector<instruction_ref> reshapes_vec;
+        while(reshaped_gemm != gemm_based_op)
+        {
+            reshapes_vec.push_back(reshaped_gemm);
+            gemm_has_multi_outs |= reshaped_gemm->outputs().size() > 1;
```
CharlieL7, line 470:
```suggestion
            gemm_has_multi_outs = gemm_has_multi_outs or reshaped_gemm->outputs().size() > 1;
```
Minor: weird to me to see bitwise or assignment for this

test/verify/test_transpose_reshape_add_sub_mul.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include "verify_program.hpp"
+#include <migraphx/program.hpp>
+#include <migraphx/generate.hpp>
+#include <migraphx/make_op.hpp>
+#include <migraphx/op/common.hpp>
+#include <migraphx/apply_alpha_beta.hpp>
+
+template <migraphx::shape::type_t DType>
+struct test_transpose_reshape_add_sub_mul
```
CharlieL7, line 33: This doesn't use MLIR. What's the reason for adding this test?

umangyadav: Sorry, I forgot to mention. This PR is dependent on https://github.com/ROCm/AMDMIGraphX/pull/3346. 

This test is coming from that PR

CharlieL7 (approved):

PR #3313, opened 2024-07-26
rnnt_example_start

richagadgil:


TedThemistokleous:
Odd CI look like it was interrupted instead of a fail

shivadbhavsar (approved):
looks good to me. The migraphx calls and tensor copies can all be async (queued on default pytorch hip stream) but since youre not really doing any performance measurements, i think this is fine as it is.

causten (approved):

examples/nlp/python_rnnt/README.md
```diff
@@ -0,0 +1,47 @@
...
+## Steps
+
+To run the RNN-T, follow these steps below.
+
+1) Install MIGraphX to your environment. Please follow the steps to build MIGraphX given at https://github.com/ROCmSoftwarePlatform/AMDMIGraphX
+
+2) Require the python venv to installed
+
+```bash
+python3 -m venv sd_venv
```
kahmed10, line 14: this probably shouldn't be called `sd_venv`

examples/nlp/python_rnnt/rnnt.py
```diff
@@ -0,0 +1,352 @@
...
+    print("Create pytorch model...")
+    pytorch_model = pytorch_rnnt_model()
+
+    print("Export pytorch model to ONNX...")
+    from rnnt_onnx import export_rnnt_onnx
+    export_rnnt_onnx(pytorch_model, seq_length=seq_length)
+
+    print("Read MIGX model from ONNX and run...")
+    migx_model = RNNT_MGX(seq_length=seq_length)
+    sd_migx = GreedyDecoder(migx_model)
```
kahmed10, line 349: same here, `sd_migx` might cause some confusion

examples/nlp/python_rnnt/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+5) Run the inference, it will compile and run the model on the first sentence in [hf-internal-testing/librispeech_asr_dummy](https://huggingface.co/datasets/hf-internal-testing/librispeech_asr_dummy)
+
+```bash
+python3 rnnt.py 
+```
+
+This will do four things:
+- Create the PyTorch model 
+- Export the PyTorch model to ONNX
```
kahmed10, line 44: would it be better to split this up so that you don't have to run the export when doing MIGraphX inference?

TedThemistokleous (approved):

PR #3316, opened 2024-07-29
Fix Ref implementation of LpPool with Asym padding

umangyadav:
Ref implementation uses "insert_pad" pass to handle padding. 
It didn't initialize padding value for the lppool properly.

### DIFF
--- a/src/insert_pad.cpp
+++ b/src/insert_pad.cpp
@@ -21,6 +21,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+#include "migraphx/op/common.hpp"
 #include <migraphx/insert_pad.hpp>
 #include <migraphx/program.hpp>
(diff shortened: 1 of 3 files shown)


pfultz2 (approved):

src/insert_pad.cpp
```diff
@@ -91,8 +91,12 @@ static void update_pooling(const instruction_ref& input, const instruction_ref&
     std::copy(pads_l.begin(), pads_l.end(), padding.begin() + 2);
     std::copy(pads_r.begin(), pads_r.end(), padding.begin() + kdims + 2 + 2);
 
-    // maxpool uses lowest value for padding
-    float pad_val = std::numeric_limits<float>::lowest();
```
CharlieL7, line 95: Wasn't this also a problem for average pooling with count_include_pad?

CharlieL7 (approved):

src/insert_pad.cpp
umangyadav: not sure about it

PR #3097, opened 2024-05-16
Support two outputs in split_reduce

pfultz2:

PR #3319, opened 2024-07-30
Fuse Split-Reduce with MLIR

umangyadav:
Part of #3212 

Depends on #3097 #3299 and https://github.com/ROCm/rocMLIR/pull/1590


src/module.cpp
```diff
@@ -1041,8 +1041,8 @@ module::fuse(const module& m,
     if(map_ins == nullptr)
         map_ins = &default_map_ins;
     insert_params(*this, inputs, *map_ins);
-    auto param_map = m.get_ins_param_map(inputs);
-    for(auto&& [input, param] : param_map)
+    auto param_map = m.get_ins_param_map(inputs, true);
+    for(auto&& [param, input] : param_map)
```
pfultz2, line 1045: Why is this flipped?

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -423,6 +470,97 @@ std::vector<instruction_ref> mlir_contiguous(module_pass_manager& mpm,
...
+
+    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    {
+        auto reduce_ins = r.result;
+        auto gemm_ins   = r.instructions["gemm"];
+        auto* rm        = reduce_ins->module_inputs().front();
+        auto names      = rm->get_parameter_names();
+        std::sort(names.begin(), names.end());
+        module_ref gemm_old_mm = gemm_ins->module_inputs().front();
+        module_ref mm          = mpm.create_module(gemm_old_mm->name() + "_split_fused_reduce");
```
pfultz2, line 492: You can do `mpm.create_module(gemm_old_mm->name() + "_split_fused_reduce", gemm_ins->module_inputs().front())` to copy the module with a new name.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -423,6 +470,97 @@ std::vector<instruction_ref> mlir_contiguous(module_pass_manager& mpm,
...
+                         {
+                             for(const auto& skip_param : inputs)
+                             {
+                                 if(not contains(param_map, skip_param))
+                                 {
+                                     param_map[skip_param] =
+                                         skip_param; // skip adding parameter for inputs of
+                                                     // pointwise inside split_fused_reduce
+                                 }
+                             }
```
pfultz2, line 517: Why is this needed? If its not in the `param_map` then that seems like a bug.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -423,6 +470,97 @@ std::vector<instruction_ref> mlir_contiguous(module_pass_manager& mpm,
...
+                             auto* sub_pm     = mod_args.front();
+                             auto param_map_2 = create_param_map_with_literals(
+                                 &main_mod, sub_pm, op.compute_shape(to_shapes(inputs), mod_args));
+                             for(const auto& i : param_map_2)
+                             {
+                                 param_map.insert(i);
+                             }
```
pfultz2, line 524: You do `param_map.insert(param_map_2.begin(), param_map_2.end())` instead.

src/targets/gpu/mlir.cpp
```diff
@@ -951,11 +954,64 @@ struct mlir_program
...
+                i, migraphx::make_op("reduce_sum", {{"axes", new_reduce_axes}}), rsp_ins);
+            auto rsp_back = m.insert_instruction(
+                i, migraphx::make_op("reshape", {{"dims", reduce_lens}}), collapsed_reduce);
+            m.replace_instruction(i, rsp_back);
+            ins_to_remove.push_back(i);
+        }
+    }
+    std::for_each(ins_to_remove.begin(), ins_to_remove.end(), [&](const auto& remove_ins) {
+        m.remove_instruction(remove_ins);
+    });
```
pfultz2, line 1005: Just run dead_code_elimination.

src/targets/gpu/mlir.cpp
```diff
@@ -988,6 +1044,7 @@ std::string dump_mlir(const module& m, const std::vector<shape>& inputs)
         mr = &mm;
         adjust_param_shapes(mm, inputs);
     }
+    rewrite_reduce(mm);
```
pfultz2, line 1047: This wont dump correctly if the inputs are empty. Probably should take the module by value in the function and remove the `mm` and `mr` variables.

src/targets/gpu/mlir.cpp
```diff
@@ -1081,12 +1139,21 @@ tuning_config get_tuning_config_mlir(const context& migraphx_ctx,
                                      bool exhaustive)
 {
     adjust_param_shapes(m, inputs);
-
+    rewrite_reduce(m);
```
pfultz2, line 1142: We should probably make one function(like `mlir_rewrites`) that does `adjust_param_shapes` and `rewrite_reduce` together. This we have once place to add any additional rewrites in the future.

src/module.cpp
umangyadav: There are cases where same input instruction is mapped to multiple parameters. 
e.g. `split_fused_reduce(x, y, x)`  

In those cases, having mapping from `input--> param` would de-duplicate it and only add single parameter. 

Later 
https://github.com/ROCm/AMDMIGraphX/blob/ae2b02650767d53b9526f660f1b2017ec4bf56e0/src/module.cpp#L238
Here it won't find parameter in the `map_ins` and would try to add it and fails

src/targets/gpu/fuse_mlir.cpp
umangyadav: This is fusing pointwise module arg inside the fused_mlir_module.  Note that pointwise module is submodule to `split_fused_reduce` module as well.

Inputs to pointwise would be available in `map_ins` internally to `fuse` method. https://github.com/ROCm/AMDMIGraphX/blob/ae2b02650767d53b9526f660f1b2017ec4bf56e0/src/module.cpp#L1047

But they won't be passed down to "inserter" is invoked. https://github.com/ROCm/AMDMIGraphX/blob/ae2b02650767d53b9526f660f1b2017ec4bf56e0/src/module.cpp#L260

src/targets/gpu/fuse_mlir.cpp
umangyadav: Done.

src/targets/gpu/fuse_mlir.cpp
umangyadav: Done

src/targets/gpu/mlir.cpp
umangyadav: > This wont dump correctly if the inputs are empty. 

Why is that ? 
`rewrite_reduce` is just reshaping reduction ops. It doesn't touch inputs.

src/targets/gpu/mlir.cpp
umangyadav: I can do that after this PR.

src/targets/gpu/mlir.cpp
umangyadav: Done.

src/targets/gpu/fuse_mlir.cpp
pfultz2: I see thats because you are reusing the outer param_map. I am not sure that is needed though. You could just use another param_map:

```cpp
auto param_map_2 = create_param_map_with_literals(&main_mod, sub_pm, op.compute_shape(to_shapes(inputs), mod_args));
return main_mod.fuse(*sub_pm, inputs, &param_map_2).front();
```

src/targets/gpu/mlir.cpp
pfultz2: > > This wont dump correctly if the inputs are empty.
> 
> Why is that ? `rewrite_reduce` is just reshaping reduction ops. It doesn't touch inputs.

It rewrites the `mm` variable which is only set to the `m` variable(ie the module passed to the function) when the inputs are not empty.

CharlieL7 (commented):
We should also have a compiler pass test for the new fusion, right?

src/targets/gpu/fuse_mlir.cpp
umangyadav: I also need to add `inputs` of the pointwise op into `param_map` to avoid adding params for that. Adding to same `param_map` solves both issues.

umangyadav: > Sure, you can use get_ins_param_map for that.

 No, I can't it will create map from `param -> input` it won't skip adding parameters for the inputs. 

https://github.com/ROCm/AMDMIGraphX/blob/ae2b02650767d53b9526f660f1b2017ec4bf56e0/src/module.cpp#L1004

I need map from `input --> input`

umangyadav: Made some changes.

src/targets/gpu/mlir.cpp
umangyadav: Done.

umangyadav:
> We should also have a compiler pass test for the new fusion, right?

Yeah. They are a bit tricky to write. Let me add a one/two. I have verify test otherwise.

test/gpu/mlir.cpp
```diff
@@ -275,6 +275,35 @@ module {
...
+    auto reduce_sum = m.add_instruction(migraphx::make_op("reduce_sum", {{"axes", {2, 3}}}), conv);
+    m.add_return({reduce_sum});
+    auto s = migraphx::gpu::dump_mlir(m);
+    // Skip test if MLIR is not enabled
+    if(s.empty())
+        return;
+    auto mlir_output_with_attrs =
+        migraphx::interpolate_string(mlir_output, {{"attrs", get_attrs()}});
+    CHECK(encode(s) == encode(mlir_output_with_attrs));
+    // EXPECT(verify_mlir(m));
```
umangyadav, line 304: verify is failing. Therefore disabling it for now. Could be an issue with rocMLIR.

umangyadav:
> We should also have a compiler pass test for the new fusion, right?

Added tests

test/gpu/fuse_mlir.cpp
```diff
@@ -452,6 +455,191 @@ TEST_CASE(int_quant_dot_tanh_fails)
...
+    {
+        auto* mm = p2.get_main_module();
+        auto x   = mm->add_parameter("x", s_x);
+        auto w   = mm->add_parameter("w", s_w);
+        auto b   = mm->add_literal(migraphx::generate_literal(s_b));
+        auto mb  = mm->add_instruction(
+            migraphx::make_op("broadcast", {{"axis", 1}, {"out_lens", {2, 32, 10, 64, 64}}}), b);
+        auto fused =
+            add_mlir(p2,
+                     "mlir_main:pointwise0_main:split_reduce0",
```
CharlieL7, line 509: minor: why do we lose "convolution" in the name the MLIR instruction?

CharlieL7 (approved):

test/gpu/fuse_mlir.cpp
umangyadav: Names are constructed from modules that are fused.  convolution or dot would appear as `mlir_op[op=""` attribute. 
https://github.com/ROCm/AMDMIGraphX/blob/05b2ff452ae8585ed1d191d907e7e95c448bc318/src/targets/gpu/fuse_mlir.cpp#L140

src/targets/gpu/mlir.cpp
```diff
@@ -1081,12 +1134,21 @@ tuning_config get_tuning_config_mlir(const context& migraphx_ctx,
...
+    std::string problem_config = tc.problem.to<std::string>();
+    for(const auto i : iterator_for(m))
+    {
+        if(starts_with(i->name(), "@"))
+        {
+            continue;
+        }
+        problem_config += " " + i->name();
+    }
+    tc.problem       = problem_config;
```
umangyadav, line 1151: I had these changes to experiment and work around problem_cache issue. Reverting these

src/targets/gpu/mlir.cpp
umangyadav: https://github.com/ROCm/AMDMIGraphX/issues/3384

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -423,6 +470,95 @@ std::vector<instruction_ref> mlir_contiguous(module_pass_manager& mpm,
...
+                             auto* sub_pm     = mod_args.front();
+                             auto param_map_2 = create_param_map_with_literals(
+                                 &main_mod, sub_pm, op.compute_shape(to_shapes(inputs), mod_args));
+                             // skip adding parameter for inputs of
+                             // pointwise inside split_fused_reduce
+                             for(const auto& skip_input : inputs)
+                             {
+                                 param_map_2[skip_input] = skip_input;
+                             }
+                             return main_mod.fuse(*sub_pm, inputs, &param_map_2).front();
```
pfultz2, line 523: Actually, `fuse` is a poor choice here. Thats why you need to skip them in the param map. Also, it doesnt insert instruction at `pos`. Instead we should add a `insert_inline` method that can insert the instructions correctly:

```cpp
std::vector<instruction_ref>
module::insert_inline(instruction_ref ins,
                      const module& m,
                      const std::vector<instruction_ref>& inputs,
                      std::unordered_map<instruction_ref, instruction_ref>* map_ins,
                      module::inserter insert)
{
    std::unordered_map<instruction_ref, instruction_ref> default_map_ins;
    if(map_ins == nullptr)
        map_ins = &default_map_ins;
    auto param_map = m.get_ins_param_map(inputs, true);
    map_ins.insert(param_map.begin(), param_map.end());
    return this->insert_instructions(ins, &m, map_ins, std::move(insert));
}
```

Then you can do `main_mod.insert_inline(pos, *sub_pm, inputs, &param_map_2).front()`, and you can skip the `skip_input` changes.

umangyadav: Done. Thanks.

src/targets/gpu/include/migraphx/gpu/mlir.hpp
```diff
@@ -50,6 +50,8 @@ struct MIGRAPHX_GPU_EXPORT mlir_code_object
     std::vector<value> prefill_values   = {};
 };
 
+MIGRAPHX_GPU_EXPORT bool is_reduce(const instruction& ins);
```
pfultz2, line 53: I dont see this used outside of mlir.cpp. I think it can be removed from the header.

umangyadav: It is beign used on both `fuse_mlir.cpp` and `mlir.cpp`

pfultz2 (approved):

PR #3322, opened 2024-07-30
int4: disable const_folding for unpack_int4

lakhinderwalia:
disable const_folding for unpack_int4


src/quantize_int4.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        auto s = ins->get_shape();
+
+        auto mod_inputs = ins->module_inputs();
+
+        // Convert each of the inputs that are fp32 or fp16 to int4
+        auto inputs = ins->inputs();
+        std::transform(inputs.begin(), inputs.end(), inputs.begin(), [&](auto inp) {
+            auto sh = inp->get_shape();
+            if(sh.broadcasted() || sh.scalar())
```
umangyadav, line 58: tidy will complain on `||`

src/driver/main.cpp
```diff
@@ -502,6 +503,7 @@ struct compiler
         ap(to_fp16, {"--fp16"}, ap.help("Quantize for fp16"), ap.set_value(true));
         ap(to_int8, {"--int8"}, ap.help("Quantize for int8"), ap.set_value(true));
         ap(to_fp8, {"--fp8"}, ap.help("Quantize for fp8e4m3fnuz type"), ap.set_value(true));
+        ap(to_int4, {"--int4"}, ap.help("Quantize for int4"), ap.set_value(true));
```
umangyadav, line 506: Better to name it as `--int4-weights`

src/quantize_int4.cpp
lakhinderwalia: Yes. of course! Thanks.

pfultz2: There is no need to check for scalar. If its scalar then it is broadcasted.

src/propagate_constant.cpp
```diff
@@ -38,17 +38,22 @@ MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_PROPAGATE_CONSTANT)
...
 {
-    if(ins->name() == "contiguous")
+    if(ins->name() == "contiguous" or ins->name() == "dequantizelinear")
         return skip_propagate(ins->inputs().front());
+
     auto&& s = ins->get_shape();
+
+    if(ins->name() == "unpack_int4")
+        return true;
+
```
umangyadav, line 48: Would this work if let's say there is a "reshape" or some other operation in between `unpack_int4` and `dequantizelinear` ? 
e.g. `unpack_int4 --> convert --> DQ`

src/driver/main.cpp
lakhinderwalia: Would you consider dot operations to contain 'weights'?

umangyadav: Yes

src/propagate_constant.cpp
pfultz2: Why would there be a convert?

umangyadav: Contrived case i came up with.

Need to handle general case if possible pre-emptively.

umangyadav: It can any other op. "convert' is just an example

pfultz2: We are the ones that will insert these operators either in quantize_int4 or in the onnx parser, so i dont understand why we would insert a `convert`, since it wouldn't work.

umangyadav: model could also come from quantizer tool e.g. vitisAI quantizer.

src/quantize_int4.cpp
```diff
@@ -0,0 +1,108 @@
...
+                return inp;
+            auto lens = sh.lens();
+            if(lens[lens.size() - 1] % 2)
+                return inp; // even sized dimensions to pack
+
+            if(not inp->can_eval())
+                return inp;
+
+            std::vector<float> val;
+            inp->eval().visit([&](auto in_data) { val.assign(in_data.begin(), in_data.end()); });
```
pfultz2, line 71: You can do `inp->eval().to_vector<float>()`.

lakhinderwalia:
@pfultz2, @CharlieL7, this PR is being moved out of the draft mode. Some of its necessary functionality is in other PRs. For example, [block_quantization support.](https://github.com/ROCm/AMDMIGraphX/pull/3412)  Also int4 (i.e. [signed support](https://github.com/ROCm/AMDMIGraphX/pull/3359)) for the basic pack/unpack. [Parse int4 graph](https://github.com/ROCm/AMDMIGraphX/pull/3408).
At this stage, we are able to compile the test graphs that are checked-in, and also other full quark-generated graph. We also need some of these other PRs to merge first. Some other tests will be added as the review progresses. Thanks.

src/propagate_constant.cpp
```diff
@@ -36,26 +36,37 @@ inline namespace MIGRAPHX_INLINE_NS {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_TRACE_PROPAGATE_CONSTANT)
 
+static int const_pr_ct = 0;
```
pfultz2, line 39: Dont use a global variable, this is not thread-safe, and this pass is called across multiple threads.

src/propagate_constant.cpp
```diff
@@ -98,22 +109,31 @@ void propagate_constant::apply(module& m) const
...
+                inss.push_back(ins);
+            })(const_instrs_vec[i]);
+            if(not skip_pc)
+            {
+                if(enabled(MIGRAPHX_TRACE_PROPAGATE_CONSTANT{}))
+                    m.debug_print(inss);
+
+                assert(literals[i].get_shape() == const_instrs_vec[i]->get_shape());
+                auto l = m.add_literal(literals[i].get_shape(), literals[i].data());
+                m.replace_instruction(const_instrs_vec[i], l);
```
pfultz2, line 135: Dont refactor propagate_constant, there is already perf regressions across almost all models because its skipping a lot of const folding. 

For now just add dequatizelinear and unpack_int4 to skip_propagate like you did in earlier:

```cpp
bool skip_propagate(instruction_ref ins)
{
    if(ins->name() == "contiguous" or ins->name() == "dequantizelinear")
        return skip_propagate(ins->inputs().front());
    if(ins->name() == "unpack_int4")
        return true;
    auto&& s = ins->get_shape();
    if(s.broadcasted() and not s.scalar())
        return true;
    if(s.scalar() and s.elements() != 1)
        return true;
    return false;
}
```

This will handle the cases with `--int4-weights`. It wont handle other cases from other models, but we can tackle that in a later PR(I dont want this to block the other changes that are being made in this PR) as we need to discuss a better design on handling those cases including the additional cases for block quantization.

src/simplify_qdq.cpp
```diff
@@ -395,10 +417,32 @@ void remove_qdq_pairs(module& m)
...
+        }
+    }
+}
+
 } // namespace
 
 void simplify_qdq::apply(module& m) const
 {
+    // first step: add pack/unpack pair between qdq for int4 weights
+    add_int4_pack_unpack_pair(m);
```
pfultz2, line 445: Add unit tests for this pass.

src/targets/gpu/mlir.cpp
```diff
@@ -644,6 +645,10 @@ struct mlir_program
                 std::copy(padding.begin(), padding.end(), std::back_inserter(v.at("padding")));
             }
         }
+
+        if(op.name() == "unpack_int4")
+            v["axis"] = ins->get_shape().ndim() - 1;
```
pfultz2, line 650: This shouldn't override the axis value, but we really dont use this.

Ideally, we should just unpack across the fastest axis which is not always the same as the last axis. Let me open an issue to address this in the future.

src/targets/gpu/target.cpp
```diff
@@ -132,14 +133,20 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         dead_code_elimination{},
         normalize_ops{},
         dead_code_elimination{},
+        // 1. For int4 ops: first remove the identity op next to packed int4 data.
+        eliminate_identity{},
+        // 2. Next: (pack/unpack)_int4 handling in simplyfy_qdq.
         simplify_qdq{},
+        // 3. Next: pruning of pack_int4 related const branches.
+        propagate_constant{{}},
```
pfultz2, line 141: This should not be moved up here. This should happen in optimize_module, like always. I dont think this is needed to support `--int4-wieghts`.

src/targets/gpu/target.cpp
```diff
@@ -132,14 +133,20 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
         dead_code_elimination{},
         normalize_ops{},
         dead_code_elimination{},
+        // 1. For int4 ops: first remove the identity op next to packed int4 data.
```
pfultz2, line 136: I dont really like having just in4 specific comments here as these passes cover more than just int4 cases, so I think it is confusing.

src/include/migraphx/quantization.hpp
```diff
@@ -49,6 +49,8 @@ MIGRAPHX_EXPORT void quantize_int8(program& prog,
 MIGRAPHX_EXPORT void
 quantize_fp8(program& prog, const target& t, const std::vector<parameter_map>& calibration);
 
+MIGRAPHX_EXPORT void quantize_int4(program& prog);
```
pfultz2, line 52: This should be called `quantize_int4_weights` since it will only quantize the weights.

pfultz2 (changes requested):

src/propagate_constant.cpp
lakhinderwalia: Thanks. Yes, this simple case was tackled early on. But it was too far removed from what the actual quark-generated model did with its int4 zero_points, which was expected to be supported in the first-deliverable itself.

src/targets/gpu/mlir.cpp
lakhinderwalia: Agreed, we don't really use the axis for unpack/pack.

src/targets/gpu/target.cpp
lakhinderwalia: (Yes, this mayn't be necessary for supporting just the `--int4-weights`.)
This was moved up due to e.g. this simple test case needs to be supported: `test/onnx/int4_const_identity_block_sz_2_qdq_test.onnx` (new file in this PR).

src/propagate_constant.cpp
lakhinderwalia: I will mark this resolved for now. Don't see any reshape after `unpack_int4`. Thanks.

src/targets/gpu/target.cpp
lakhinderwalia: Reverted.

src/simplify_qdq.cpp
lakhinderwalia: Added.

src/propagate_constant.cpp
lakhinderwalia: For now, the changes are reverted to handing just `--int4-weights`.

src/targets/gpu/target.cpp
```diff
@@ -161,14 +162,20 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
...
+        eliminate_identity{},
+        // 2. Next: (pack/unpack)_int4 handling in simplyfy_qdq.
         simplify_qdq{},
+        // 3. Next: pruning of pack_int4 related const branches.
+        propagate_constant{{}},
```
pfultz2, line 170: I think this should be removed, its not necessary for int4-weights, and I am concerned it might cause perf regressions.

test/gpu/mlir.cpp
```diff
@@ -454,4 +454,51 @@ module {
...
+        migraphx::interpolate_string(mlir_output, {{"attrs", get_attrs()}});
+
+    CHECK(encode(s) == encode(mlir_output_with_attrs));
+}
+
+TEST_CASE(int4_unpack_conv)
+{
+    std::string mlir_output = R"__migraphx__(
+ module  {  func.func  @mlir_unpack_int4_quant_convolution(%arg0:  !migraphx.shaped<2x8x2x1xi8,  16x2x1x1>,  %arg1:  !migraphx.shaped<1x8x4x4xi8,  128x16x4x1>)  ->  !migraphx.shaped<1x2x3x3xi32,  18x9x3x1>  attributes  {arch  =  "",  kernel  =  "mixr",  num_cu  =  0  :  i64}  {  %0  =  migraphx.unpack  %arg0  {axis  =  3  :  i64,  isUnsigned  =  false}  :  <2x8x2x1xi8,  16x2x1x1>  ->  <2x8x2x2xi8,  32x4x2x1>  %1  =  migraphx.quant_convolution  %arg1,  %0  {dilation  =  [1,  1],  group  =  1  :  i64,  padding  =  [0,  0,  0,  0],  padding_mode  =  0  :  i64,  stride  =  [1,  1]}  :  <1x8x4x4xi8,  128x16x4x1>,  <2x8x2x2xi8,  32x4x2x1>  ->  <1x2x3x3xi32,  18x9x3x1>  return  %1  :  !migraphx.shaped<1x2x3x3xi32,  18x9x3x1>  }  }
+)__migraphx__";
```
pfultz2, line 487: This isn't very readable. Format this string better like the other tests do.

src/quantize_int4.cpp
pfultz2: Are you able to update this as well? @lakhinderwalia

src/targets/gpu/target.cpp
lakhinderwalia: You are right. I have missed pushing this particular file! Let us wait what the perf results are anyways!

src/quantize_int4.cpp
lakhinderwalia: I haven't done this yet. That is why the comment is left open. Will add it to my next set of changes. Thanks.

test/gpu/mlir.cpp
lakhinderwalia: Yes. Will do. We did talk about it.

src/targets/gpu/target.cpp
lakhinderwalia: It has been removed. However, I didn't notice any perf regression below.

src/quantize_int4.cpp
lakhinderwalia: Done.

lakhinderwalia: This change gave a compile failure. I just reverted it for now.

pfultz2 (approved):
This is still missing unit test for simplify_qdq. It really should be addressed either before merging or after merging.

lakhinderwalia:
> This is still missing unit test for simplify_qdq. It really should be addressed either before merging or after merging.

Added.

test/int4_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    migraphx::run_passes(m1, {migraphx::simplify_qdq{}});
+
+    auto chk_1 = match::name("quantizelinear")(
+                     match::output(match::name("pack_int4")(match::output(match::name(
+                         "unpack_int4")(match::output(match::name("dequantizelinear")))))))
+                     .bind("q");
+    auto res_1 = find_match(m1, chk_1);
+    EXPECT(migraphx::contains(res_1.instructions, "q"));
+}
```
pfultz2, line 113: This should be moved with the other simplify_qdq tests, but lets do that later. I think this should just be merged in now.

pfultz2:
There is an mlir failure on jenkins:

```
[2024-09-19T21:19:12.937Z] [   RUN    ] int4_unpack_conv

[2024-09-19T21:19:12.937Z] terminate called after throwing an instance of 'migraphx::version_1::exception'

[2024-09-19T21:19:12.937Z]   what():  /home/jenkins/workspace/AMDMIGraphX_PR-3322/src/targets/gpu/mlir.cpp:787: run_backend_pipeline: MLIR backend compilation failed: Error: The size of rock.alloc should be greather than zero.

[2024-09-19T21:19:12.937Z] Note: see current operation: %43 = "rock.alloc"() : () -> memref<16xi4, #gpu.address_space<private>>
```

Do you see this same error locally?

lakhinderwalia:
> There is an mlir failure on jenkins:
> 
> ```
> [2024-09-19T21:19:12.937Z] [   RUN    ] int4_unpack_conv
> 
> [2024-09-19T21:19:12.937Z] terminate called after throwing an instance of 'migraphx::version_1::exception'
> 
> [2024-09-19T21:19:12.937Z]   what():  /home/jenkins/workspace/AMDMIGraphX_PR-3322/src/targets/gpu/mlir.cpp:787: run_backend_pipeline: MLIR backend compilation failed: Error: The size of rock.alloc should be greather than zero.
> 
> [2024-09-19T21:19:12.937Z] Note: see current operation: %43 = "rock.alloc"() : () -> memref<16xi4, #gpu.address_space<private>>
> ```
> 
> Do you see this same error locally?

No.

pfultz2:
If this needs a change from mlir to work, then just comment out the broken test for now so we can merge this in.

lakhinderwalia:
Thanks. Actually, MLIR has a fix just commited-in. But I can disable the test. However, we do need one more approval :-)

causten:
I've got https://github.com/ROCm/AMDMIGraphX/pull/3467 which is pulling from mlir tip.  I"m planning to merge it asap

lakhinderwalia:
> I've got #3467 which is pulling from mlir tip. I"m planning to merge it asap

@causten, please include 1b99123a4e86629b07b1c2815668f6222d093d70 or newer. Thanks.

shivadbhavsar (approved):

PR #3342, opened 2024-08-05
Fix transposed padding

pfultz2:


src/targets/gpu/jit/pad.cpp
```diff
@@ -79,7 +79,7 @@ struct pad_compiler : compiler<pad_compiler>
 
         auto vinputs = inputs;
         vinputs.push_back(inputs.front().with_lens(offset_lens));
-        auto rinputs = reduce_dims(vinputs);
+        auto rinputs = reduce_dims(normalize_permutation(vinputs));
```
umangyadav, line 82: can you add a test which exposes and fixes this issue  ?

pfultz2: This fix is for optimization purposes not for correctness.

umangyadav (approved):

TedThemistokleous (approved):

PR #3343, opened 2024-08-05
Move auto_contiguous before lowering

pfultz2:
This should run before lowering. Running it earlier can cause contiguous to be incorrectly fused with mlir, which can cause mixed layout to be used when its not necessary.


umangyadav (approved):

TedThemistokleous (approved):

TedThemistokleous:
Fix CI otherwise looks good

causten (changes requested):
We reran this twice and in both cases performance tanked on many models.  This can not be merged as is

kahmed10:
@causten @pfultz2 inserting workaround to only move the pass if NHWC is enabled.

PR #3345, opened 2024-08-06
Use new args each time in eliminate_contiguous

umangyadav:
Eliminate Contiguous tries to replace each argument in a loop. 

Therefore need to create copy of new args each time inside the loop as they may change while looping.

I am not able to create a test as all the "Ref" ops allow Non-standard shapes. 

But I have a case like following with the CPU backend which doesn't allow non-std shape with Binary Ops. 

```cpp
at = transpose(a)
ac = contiguous(at)
br = reshape_lazy(b)
bc = contiguous(br)
add = binary_add(ac, bc)
```

Current logic doesn't remove `contiguous(br)`. Reason is it first creates `new_args = {ac, bc}` then it modifies them to be `{at, bc}` when it tries to remove `contiguous(at)`. It fails but then it doesn't revert `new_args` back to `{ac, bc}` it keeps them at `{at, bc}` and because of that when it tries to remove `contiguous(br)` it fails. Therefore `contigous(br)` is not eliminated.


CharlieL7 (approved):

pfultz2:
So it was written this way so we check if removing all contiguouses will work, where each individually dotn work.It can be the case that an operator(like for miopen's convolution) will fail if one is contiguous and the other is nhwc, but if both are nhwc it will work fine. We should probably add a unit test for that.

> But I have a case like following with the CPU backend which doesn't allow non-std shape with Binary Ops.

You can add a simple operator that requires standard shape similar to the `pass_standard_op` that is used in the current unit tests.

umangyadav:
> You can add a simple operator that requires standard shape similar to the pass_standard_op that is used in the current unit tests. 

Added a test.

shivadbhavsar (approved):

PR #3346, opened 2024-08-06
Fix reshapes when lowering for CPU backend

umangyadav:
Pipeline for  CPU backend is 
`auto_contiguous --> simplify_reshapes --> DCE --> .... --> lowering --> eliminate_contiguous`

This PR fixes two issues.

1. Lowering for the `reshape` copy operator was missing on CPU Backend. `reshape` copy operator would modify the buffer in place and can work on non-std shaped input buffers.  Therefore when lowering, it should  create another buffer for the output otherwise it can run into accuracy issues. 
2. Simplify Reshapes eliminates contiguouses in some cases which causes failures when lowering later. For example following https://github.com/ROCm/AMDMIGraphX/blob/a87416ae4533d677ed96411789162b3b34cb945d/src/simplify_reshapes.cpp#L642

For the case like 
```cpp
a = param(lens={2, 10})
b  = param(lens ={5, 4})
at = transpose(a)--> {10, 2}
ac = contigous(at) 
ar = reshape(ac) --> {5, 4}
add = binary_add(ar, b)
```

This is transformed by `find_reshapes_cont` to following : 
```
a = param(lens={2, 10})
b  = param(lens ={5, 4})
at = transpose(a)--> {10, 2}
br = reshape(b) --> {10, 2}
add = binary_add(at, br) --> {10, 2}
add_r = reshape(add) --> {5, 4}
```

When lowering for CPU backend, this would fails as `dnnl::binary()` op expects Standard shaped inputs only.

### DIFF
--- a/src/targets/cpu/lowering.cpp
+++ b/src/targets/cpu/lowering.cpp
@@ -478,22 +478,7 @@ struct cpu_apply
 
     instruction_ref replace(instruction_ref ins, const operation& op) const
     {
-        auto inputs = ins->inputs();
-        std::vector<instruction_ref> cont_inputs;
-        for(const auto& i : inputs)
-        {
-            if(not i->get_shape().standard())
-            {
-                auto alloc = insert_allocation(ins, i->get_shape().as_standard());
-                cont_inputs.push_back(
-                    modl->insert_instruction(ins, make_op("dnnl::reorder"), {i, alloc}));
-            }
-            else
-            {
-                cont_inputs.push_back(i);
-            }
-        }
-        return replace(ins, op, cont_inputs);
+        return replace(ins, op, ins->inputs());
     }
 
     instruction_ref
--- a/src/targets/cpu/lowering.cpp
+++ b/src/targets/cpu/lowering.cpp
@@ -397,6 +397,10 @@ struct cpu_apply
             {
                 apply_pooling(it);
             }
+            else if(it->name() == "reshape")
+            {
+                apply_reshape(it);
+            }
             else if(apply_map.count(it->name()) > 0)
             {
                 apply_map.at(it->name())(it);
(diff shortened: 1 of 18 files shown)


src/targets/cpu/lowering.cpp
```diff
@@ -451,7 +478,22 @@ struct cpu_apply
...
+        auto inputs = ins->inputs();
+        std::vector<instruction_ref> cont_inputs;
+        for(const auto& i : inputs)
+        {
+            if(not i->get_shape().standard())
+            {
+                auto alloc = insert_allocation(ins, i->get_shape().as_standard());
+                cont_inputs.push_back(
+                    modl->insert_instruction(ins, make_op("dnnl::reorder"), {i, alloc}));
+            }
```
pfultz2, line 490: This should already be done by `auto_contiguous`. We should probably move it right before lowering like what I did in #3343 for the gpu backend.

umangyadav: Yes it should be done except `simplify_reshapes` eliminates it

umangyadav: It fails when at `make check` if i move auto_contiguous
```
[   RUN    ] test_scatternd_add

module: "main"
@0 = @literal{4, 3, 1, 7} -> int64_type, {4, 1}, {1, 4}
@1 = allocate[shape=int64_type, {4, 1}, {1, 1},buf_type=nullopt] -> int64_type, {4, 1}, {1, 1}
@2 = contiguous(@0) -> int64_type, {4, 1}, {1, 1}
update = @param:update -> float_type, {4}, {1}
data = @param:data -> float_type, {8}, {1}
@5 = scatternd_add(data,@2,update) -> float_type, {8}, {1}
@6 = @return(@5)


module: "main"
update = @param:update -> float_type, {4}, {1}
@1 = @literal{4, 3, 1, 7} -> int64_type, {1, 4}, {4, 1}
data = @param:data -> float_type, {8}, {1}
@3 = convert[target_type=2](@1) -> float_type, {1, 4}, {4, 1}
@4 = convert[target_type=9](@3) -> int64_type, {1, 4}, {4, 1}
@5 = transpose[permutation={1, 0}](@4) -> int64_type, {4, 1}, {1, 4}
@6 = scatternd_add(data,@5,update) -> float_type, {8}, {1}
@7 = @return(@6)


FAILED: test_scatternd_add
    what(): ../../../../src/targets/cpu/dnnl.cpp:72: to_dnnl_memory_data_type: Unsupported data type

module: "main"
update = @param:update -> float_type, {4}, {1}
@1 = @literal{4, 3, 1, 7} -> int64_type, {1, 4}, {4, 1}
data = @param:data -> float_type, {8}, {1}
@3 = convert[target_type=2](@1) -> float_type, {1, 4}, {4, 1}
@4 = convert[target_type=9](@3) -> int64_type, {1, 4}, {4, 1}
@5 = transpose[permutation={1, 0}](@4) -> int64_type, {4, 1}, {1, 4}
@6 = scatternd_add(data,@5,update) -> float_type, {8}, {1}
@7 = @return(@6)


CMake Error at gdb/test_test_verify_general/run.cmake:16 (message):
  Test failed

```

umangyadav: > We should probably move it right before lowering like what I did in https://github.com/ROCm/AMDMIGraphX/pull/3343 for the gpu backend. 

Done

shivadbhavsar (approved):

pfultz2 (approved):

PR #3350, opened 2024-08-07
Improve performance of pooling kernel

pfultz2:
This should match much closer of the performance for MIOpen for many cases:

```
gpu::code_object[code_object=16512,symbol_name=pooling_kernel,global=1605632,local=256,] -> half_type, {64, 64, 56, 56}, {200704, 3136, 56, 1}: 0.068245ms

gpu::pooling[mode=max,padding={0, 0},padding_mode=0,stride={2, 2},lengths={3, 3},dilations={1, 1},ceil_mode=0,count_include_pad=0,lp_order=2,dyn_global=0] -> half_type, {64, 64, 56, 56}, {200704, 3136, 56, 1}: 0.0665585ms
```


pfultz2:
There was a bug, and now it seems its actually slower.

TedThemistokleous:
Curious how much slower? 10% or like < 5%?

PR #3352, opened 2024-08-09
Add Glu op

aarushjain29:


src/include/migraphx/op/glu.hpp
```diff
@@ -0,0 +1,80 @@
...
+            //Second Half
+            starts[dim] = mid;
+            ends[dim] = lens[dim];
+            auto second_half = migraphx::op::slice{{dim}, starts, ends}(x);
+
+            auto sigmoid_half = migraphx::op::sigmoid{}(second_half);
+
+            return migraphx::op::mul{}(first_half, sigmoid_half);
+        };
+    }
```
pfultz2, line 73: This is just a composite of other operators. An migraphx operator should not be added for this. Instead the onnx parser should just insert the migraphx operators needed to do the glu operator. Similar to how we implement other operators like GELU or SELU.

PR #3359, opened 2024-08-12
signed-int4 support for Pack/Unpack

lakhinderwalia:


src/include/migraphx/op/pack_int4.hpp
```diff
@@ -62,35 +62,62 @@ struct pack_int4
...
-            auto input_val = input[in_data_multi_idx];
-            // mask first 4 bits, keep it little endian.
-            output[i] = uint8_t(0x0F) & input_val;
-            in_data_multi_idx[axis] += 1;
-            input_val = input[in_data_multi_idx];
-            output[i] = (input_val << 4u) | output[i]; // NOLINT(hicpp-signed-bitwise)
+
+        visit_all(result, input)([&](auto out, auto inp) {
+            par_for(output_shape.elements(), [&](auto i) {
+                using type = typename decltype(inp)::value_type;
```
pfultz2, line 88: You should check the size and throw:

```cpp
if constexpr(sizeof(type) != 1)
    MIGRAPHX_THROW("Unsupported type");
```

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -62,35 +62,62 @@ struct pack_int4
...
+                    auto in_data_multi_idx = data_idx;
+                    in_data_multi_idx[axis] *= 2;
+                    int8_t val1 = inp[in_data_multi_idx];
+                    val1        = std::min(std::max(val1, min_4bit), max_4bit); // clip
+
+                    in_data_multi_idx[axis] += 1;
+                    int8_t val2 = inp[in_data_multi_idx];
+                    val2        = std::min(std::max(val2, min_4bit), max_4bit); // clip
+
+                    out[i] = (val2 << 4) | (val1 & 0xf); // pack
```
pfultz2, line 103: We are going need to add `NOLINT(hicpp-signed-bitwise)` on these pack lines.

src/include/migraphx/op/unpack_int4.hpp
```diff
@@ -62,30 +62,53 @@ struct unpack_int4
...
-            auto input_val = input[data_idx];
-            // mask first 4 bits, packing is assumed to be little endian
-            output[out_data_multi_idx] = uint8_t(0x0F) & input_val;
-            out_data_multi_idx[axis] += 1;
-            output[out_data_multi_idx] = input_val >> 4; // NOLINT(hicpp-signed-bitwise)
+
+        visit_all(result, input)([&](auto out, auto inp) {
+            par_for(in_shape.elements(), [&](auto i) {
+                using type    = typename decltype(out)::value_type;
+                auto data_idx = in_shape.multi(i);
```
pfultz2, line 85: Similar type check here:

```cpp
if constexpr(sizeof(type) != 1 or not std::is_integral<type>{})
    MIGRAPHX_THROW("Unsupported type");
```

test/ref/pack_int4.cpp
```diff
@@ -91,10 +107,8 @@ TEST_CASE(pack_int4_axis_0)
     EXPECT(migraphx::verify::verify_rms_range(results_vector, gold));
 }
 
-TEST_CASE(pack_int4_nchw)
+TEST_CASE(pack_uint4_nchw)
 {
-    // test with literal values such as 0x18 in which first 4 bits will be dropped, ideally
-    // quantizer should produce values that fit into 4 bits.
```
pfultz2, line 97: I would change this comment to: `test with literal values such as 0x18 in which first 4 bits will be dropped. The pack should clip these values`.

src/include/migraphx/op/pack_int4.hpp
```diff
@@ -62,35 +62,62 @@ struct pack_int4
...
+                }
+                else
+                {
+                    const uint8_t max_4bit = 15; // clip max value
```
umangyadav, line 107: only max and min values are differing. you can combine rest of the logic. You can set min value for unsigned as `0`

test/ref/pack_int4.cpp
```diff
@@ -115,13 +129,13 @@ TEST_CASE(pack_int4_nchw)
...
                               0xFE,
-                              0x10,
-                              0x32,
-                              0x54,
-                              0x76,
-                              0x98,
-                              0xBA,
-                              0xDC,
-                              0xFE};
+                              0xFF,
```
umangyadav, line 132: add tests "ref" tests for "unpack" as well

src/include/migraphx/op/pack_int4.hpp
pfultz2: You can also declare them using the `type` so it still uses the same type:

```cpp
type max_4bit = 0;
```

krzysz00 (commented):
I think the math's fine here, yeah

On the MLIR side, I'd prefer `unpack` to carry an attribute indicating signedness behavior instead, but that can be special-cased at the edge

src/include/migraphx/op/pack_int4.hpp
lakhinderwalia: Isn't it enough to check it in `normalize_compute_shape`?

test/ref/pack_int4.cpp
lakhinderwalia: There already is a test for pack_unpack. Isn't that enough?

src/include/migraphx/op/pack_int4.hpp
lakhinderwalia: Unfortunately, we get into issues with `visit_all`, where other types are getting involved in the compilation logic, and errors like this are showing up:
> error: use of overloaded operator '<<' is ambiguous (with operand types 'type' (aka 'half_float::half') and 'int')
And this is because it isn't just this line that is going to change.

lakhinderwalia: This code consolidation is taken care of. This constexpr is also avoided, and the code can now basically handle any type in visit_all clause here.

src/include/migraphx/op/pack_int4.hpp
lakhinderwalia: Actually we don't need this, since with the modified logic we do a visit_all() to handle all types. Thanks.

src/include/migraphx/op/pack_int4.hpp
lakhinderwalia: Taken care of (slightly differently, however).

src/include/migraphx/op/unpack_int4.hpp
lakhinderwalia: (Done slightly differently, that we don't need to check for this).

test/ref/pack_int4.cpp
lakhinderwalia: Just to clarify further, there are ref tests that check for pack operations, and then followed by its unpack counterpart. And those initial and final values are compared appropriately. Thanks.

test/ref/pack_int4.cpp
lakhinderwalia: Comments have been updated, though not verbatim. Thanks

pfultz2 (approved):
LGTM, but can we get another reviewer from migraphx team?

PR #3361, opened 2024-08-12
Add trace matchers

pfultz2:


src/include/migraphx/matcher.hpp
```diff
@@ -582,6 +584,52 @@ inline auto outputs()
...
+            ctx.debug_print(ins);
+            optional<instruction_ref> result = m.match(ctx, ins);
+            if(result.has_value())
+                std::cout << "Found\n";
+            else
+                std::cout << "Not Found\n";
+            return result;
+        });
+    };
+}
```
umangyadav, line 601: I was wondering, it would be better if we can print what was matched and what all instructions were replaced due to application of that matcher. 

e.g. we do something similar in propagate_constant pass 
https://github.com/ROCm/AMDMIGraphX/blob/7cdcefd0b989159407a1e19a5bdb34b0d29faf6a/src/propagate_constant.cpp#L101 

If i have "before" and "after"  instructions for each matcher, it would make debugging a lot easier (could be slower to run and print though).

src/include/migraphx/matcher.hpp
```diff
@@ -582,6 +584,52 @@ inline auto outputs()
     };
 }
 
+inline auto trace(std::string s)
```
umangyadav, line 587: what's the use case for this one ?

src/include/migraphx/matcher.hpp
pfultz2: Yea that is a very different case. I am not sure if you can easily print a subgraph(the propagate constant its easy to do since you know its a small sub graph in general).

src/include/migraphx/matcher.hpp
pfultz2: 1) To see whats get passed to the matcher(ie to see how far it makes it in the matcher).
2) To see the context of what is matching. You can place it on the outer matcher while you have traces on the inner matcher: `trace("outer")(some_matcher(trace("inner")(another_matcher)))`.

CharlieL7 (approved):

turneram (approved):

PR #3362, opened 2024-08-12
Preload tiles into LDS to improve performance of pointwise transposes

### LINKED ISSUE #3172: Preload transposed inputs in LDS for pointwise kernels
To improve performance for transpose kernels we should load the transposed inputs into LDS directly, and then read from LDS instead. We have function like `preload_copy` which will do this preloading with a vectorize copy, but this will preload the entire tensor.

Instead we should only preload a slice. Ideally this slice will be across the transposed axis, so in the case of NHWC we would slice across the C axis(ie axis 1) which will give us slices of CHW. The slice may contain more than one channel for the case where HW is very small as we want to make sure each workgroup has engough work to do.

The `preload_copy` function takes a boolean which enables preloading for the tensor. It uses a lambda callback with the new tensor so the LDS can be statically allocated and its lifetime will live as long as the tensor view. We should be able to reuse this function to do the preloading, we just need to do the slicing:

```cpp
template <index_int Axis, index_int N, class T>
__device__ auto auto_preload_tranposed(index idx, T x)
{
    return [=](auto f) {
        auto y = make_slice<Axis, N>(x);
        preload_copy<should_preload_transposed_slice<T, Axis>()>(idx, y)(f);
    };
}
```

Functions like `make_slice` and `should_preload_transposed_slice` will need to be implemented. And this could be extended for variadic parameters by using recursion:

```cpp
template <index_int Axis, index_int N, class T, class... Ts>
__device__ auto auto_preload_tranposed(index idx, T x, Ts... xs)
{
    return [=](auto f) {
        auto_preload_tranposed(idx, x)([=](auto y) {
            auto_preload_tranposed(idx, xs...)([=](auto... ys) {
                f(y, ys...);
            }); 
        });
    };
}
```

Currently, `preload_copy` will only work with packed tensors because it uses the original tensor shape to read the LDS buffer. We can make packed be a restriction at first, but we could get unpacked support working by tweak the `preload_copy` in the future.

Since we are reading a slice per workgroup the global calculation will need to be updated.

Since we are preloading transposed inputs we should prefer that the inputs are transposed rather than the outputs. This might require some adjustment to the permutation that we pass to `normalize_permutation`. 

Currently we call into the `pointwise` function to do the pointwise, but we probably want to make a `transposed_pointwise` function to handle the case we want to do the preloading, so its handled automatically.

pfultz2:
Fixes #3172.


src/targets/gpu/kernels/include/migraphx/kernels/functional.hpp
```diff
@@ -288,6 +288,7 @@ constexpr auto join(G g, F f)
 template <class G, class F, class... Fs>
 constexpr auto join(G g, F f, Fs... fs)
 {
+    // return f1([=](auto x) { return f2([=](auto y) { return g(x, y); }); });
```
TedThemistokleous, line 291: Why is this here?

src/targets/gpu/compile_gen.cpp
```diff
@@ -169,6 +172,116 @@ bool preload::is_preloading() const
...
 
+static std::size_t integer_divide_ceil(std::size_t x, std::size_t y)
+{
+    return (x + y - std::size_t{1}) / y;
+}
+
+static std::size_t compute_tile_factor(std::size_t r, std::size_t max_size = 64)
+{
+    std::size_t n = 1;
+    auto factors  = make_array(2, 3, 5, 7, 11);
```
TedThemistokleous, line 183: Do we want any higher primes as part of the factors? 13, 17 and 19, 23 are also close (eg 23 is 11*2 + 1 ), but is there some sort of trade between higher primes are better tile factors?

TedThemistokleous: Actually I suppose smaller primes yield more tiles and thus easier to fit  without the risk of running out of memory? More a question on my end vs a design concern. I get what you're doing here, just curious as to why not have some additional searches

src/targets/gpu/kernels/include/migraphx/kernels/functional.hpp
pfultz2: To show how it works when there is only three arguments.

src/targets/gpu/kernels/include/migraphx/kernels/tuple.hpp
```diff
@@ -142,12 +142,12 @@ struct tuple : tuple_detail::tuple_base<Ts...>
...
-                    return (xs < ys) ? -1 : (ys < xs) ? 1 : 0;
-                }...);
-            });
-        });
+                   return y([&](const auto&... ys) {
+                       return fold([&](auto a, auto b) { return a == 0 ? b() : a; })(0, [&] {
+                           return (xs < ys) ? -1 : (ys < xs) ? 1 : 0;
+                       }...);
+                   });
+               }) < 0;
```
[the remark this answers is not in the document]
pfultz2: So the fold is returning -1 if its less than, 0 if its equal and 1 if its greater than(ie `(xs < ys) ? -1 : (ys < xs) ? 1 : 0`). As you can see, if the tuple is greater than, it will return true for less than which is the opposite of what you want.

src/targets/gpu/compile_gen.cpp
pfultz2: I have not seen models using that high of prime factors in the dimensions. The important factor is that each wavefront is processing one row in the tile.

TedThemistokleous: Okay. That makes sense. Fair. I suppose until we hit a model with the higher dim, then leave them out.

TedThemistokleous:
@kahmed10  there's some other failures in CI with this one.

TedThemistokleous (approved):

CharlieL7:
Seems to be failing `test_spacetodepth_example_cpu` from ONNX backend. The test says CPU but the compiled program looks to be using the GPU...

src/targets/gpu/compile_gen.cpp
```diff
@@ -169,6 +172,116 @@ bool preload::is_preloading() const
...
+        result.args.begin(), result.args.end(), [](auto m) { return m != mode::none; });
+    // TODO: Handle tiling more than one arguments
+    if(nargs != 1)
+        return {};
+
+    const auto& s = inputs.front();
+    auto dim1     = compute_tile_factor(s.lens()[result.axis]);
+    auto dim2     = compute_tile_factor(s.lens().back(), 4096 / dim1);
+    if(dim1 == 1 or dim2 == 1)
+        return {};
```
CharlieL7, line 236: If I'm understanding this correctly, this code won't tile a larger prime number dimension like 1009. It's trying to find a tile that fits the tensor without going out of bounds. Can we also support these prime number dimensions by having a bounds check in the kernel?

src/targets/gpu/compile_gen.cpp
```diff
@@ -169,6 +172,116 @@ bool preload::is_preloading() const
...
+
+    result.outer = s.lens();
+    result.outer[result.axis] /= dim1;
+    result.outer.back() /= dim2;
+
+    auto tile_size  = dim1 * dim2;
+    result.ntiles   = s.elements() / tile_size;
+    auto tile_bytes = (tile_size + dim1) * s.type_size();
+    if(tile_bytes > 65536)
+        return {};
```
CharlieL7, line 251: Can we choose a smaller tile size fit into LDS by dividing by a prime factor? There's probably a point where the tile size is too small to be beneficial though.

src/targets/gpu/compile_gen.cpp
```diff
@@ -169,6 +172,116 @@ bool preload::is_preloading() const
...
+
+    auto nargs = std::count_if(
+        result.args.begin(), result.args.end(), [](auto m) { return m != mode::none; });
+    // TODO: Handle tiling more than one arguments
+    if(nargs != 1)
+        return {};
+
+    const auto& s = inputs.front();
+    auto dim1     = compute_tile_factor(s.lens()[result.axis]);
+    auto dim2     = compute_tile_factor(s.lens().back(), 4096 / dim1);
```
CharlieL7, line 234: The maximum tile size looks like it should depend on the data type.

src/targets/gpu/compile_gen.cpp
```diff
@@ -169,6 +172,116 @@ bool preload::is_preloading() const
...
+    result.inner[result.axis] = dim1;
+    result.inner.back()       = dim2;
+
+    result.outer = s.lens();
+    result.outer[result.axis] /= dim1;
+    result.outer.back() /= dim2;
+
+    auto tile_size  = dim1 * dim2;
+    result.ntiles   = s.elements() / tile_size;
+    auto tile_bytes = (tile_size + dim1) * s.type_size();
```
CharlieL7, line 249:
```suggestion
    // equivalent to dim1 * (dim2 + 1) to avoid bank conflicts 
    auto tile_bytes = (tile_size + dim1) * s.type_size();
```

src/targets/gpu/compile_gen.cpp
TedThemistokleous: I dont think we want a larger prime. My understanding was the use of primes was related to having the smallest possible irreducible kernel size and going above 13 wasn't used/ we never ran into a case for the higher primes.

CharlieL7: Not about larger primes for the factorization, I mean if we have a larger prime number for the dimension to tile. Since we're trying to use the largest tile size that fits on LDS?

pfultz2: > Can we also support these prime number dimensions by having a bounds check in the kernel?

We would need to do padding, which will be a little more complicated, but its definitely something we should probably look into the future.

src/targets/gpu/compile_gen.cpp
pfultz2: The `compute_tile_factor` should pick a smaller size, but if there is a bug somewhere(as these are called into two different functions) its good to have this check so we can bail instead of having a compilation error.

src/targets/gpu/compile_gen.cpp
CharlieL7: Issue to track https://github.com/ROCm/AMDMIGraphX/issues/3427

CharlieL7 (approved):

TedThemistokleous:
@pfultz2  Looks like in CI its still failing but on pooling for the all targets build in jenkins.

```
] 354/382 Test #365: test_api_custom_op_gpu ....................................................***Exception: SegFault  1.84 sec
```

Is this related to your remove padding from pooling PR #3423 ? Looks like some failures themselves with pooling

PR #3364, opened 2024-08-12
Move fuse_concat after fuse_mlir

pfultz2:


umangyadav:
Any models or cases benefitting from this one ? Or are there any existing issues that can be closed with this PR ?

TedThemistokleous (approved):

TedThemistokleous:
> Any models or cases benefitting from this one ? Or are there any existing issues that can be closed with this PR ?

I'm assuming we're letting MLIR handle any sort of Concats too that it can bundle before we then optimize them with our pass?

pfultz2:
> Any models or cases benefitting from this one ?

Not right now, we need to work on improving eliminate_concat to get some more benefit.

> I'm assuming we're letting MLIR handle any sort of Concats too that it can bundle before we then optimize them with our pass?

This allows mlir to still do output fusions instead of fusing it with concat, which could be eliminated later.

PR #3368, opened 2024-08-13
Optimize transposed concat

pfultz2:


src/targets/gpu/jit/concat.cpp
```diff
@@ -65,21 +65,39 @@ struct concat_compiler : compiler<concat_compiler>
 {
     std::vector<std::string> names() const { return {"fused_concat", "concat"}; }
 
+    static std::vector<shape> normalize(std::vector<shape> inputs, std::size_t& axis)
```
umangyadav, line 68: chnages look good, but i don't follow what it is trying to do. 

Can you put some explanation ?

pfultz2: This will reduce the dimensions and normalize the permutation, but it inserts an extra shape so it can track where the axis from this.

umangyadav (approved):

PR #3371, opened 2024-08-14
Enable hipBLASLt 

ahsan-ca:
This PR adds functionality to enable hipBLASLt in MIGraphX.
hipBLASLt can be enabled via MIGRAPHX_ENABLE_HIPBLASLT_GEMM env variable.


src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -126,6 +127,41 @@ struct hip_device
...
+        auto get_hipblaslt_workspace()
+        {
+            setup();
+            if(hbltworkspace == nullptr)
+            {
+                hbltworkspace = create_hipblaslt_workspace_ptr();
+            }
+            assert(hbltworkspace.get() != nullptr);
+            return hbltworkspace.get();
+        }
```
pfultz2, line 162: The workspace should not be stored in the context. Instead during lowering we can add a allocate instruction and then pass it as an argument to the operator. This is the way we handle the workspace in miopen.

src/targets/gpu/include/migraphx/gpu/context.hpp
```diff
@@ -126,6 +127,41 @@ struct hip_device
...
+
+        auto get_hipblaslt_preference()
+        {
+            setup();
+            if(hbltpreference == nullptr)
+            {
+                hbltpreference = create_hipblaslt_preference_ptr();
+            }
+            assert(hbltpreference.get() != nullptr);
+            return hbltpreference.get();
```
pfultz2, line 150: The preference pointer should be stored in the operator and created either during compile_blas or finalize.

src/targets/gpu/hip_gemm_impl.cpp
```diff
@@ -0,0 +1,599 @@
...
+        CHECK_HIPBLAS_ERROR(hipblasLtMatmulDescSetAttribute(
+            hipblaslt_desc, HIPBLASLT_MATMUL_DESC_TRANSA, &op_B, sizeof(int32_t)));
+    }
+
+    ~hip_gemm_impl()
+    {
+        CHECK_HIPBLAS_ERROR(hipblasLtMatmulDescDestroy(hipblaslt_desc));
+        CHECK_HIPBLAS_ERROR(hipblasLtMatrixLayoutDestroy(matA));
+        CHECK_HIPBLAS_ERROR(hipblasLtMatrixLayoutDestroy(matB));
+        CHECK_HIPBLAS_ERROR(hipblasLtMatrixLayoutDestroy(matC));
```
pfultz2, line 263: Use managed pointers.

src/targets/gpu/include/migraphx/gpu/context.hpp
pfultz2: Actually it should just be created in the solution class. anyways.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
```diff
@@ -0,0 +1,67 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#ifndef MIGRAPHX_GUARD_MIGRAPHLIB_HIPBLASLT_HPP
+#define MIGRAPHX_GUARD_MIGRAPHLIB_HIPBLASLT_HPP
+#include <migraphx/manage_ptr.hpp>
+#include <migraphx/gpu/config.hpp>
+#if MIGRAPHX_USE_HIPBLASLT
+#include <hipblaslt/hipblaslt.h>
+#define HIPBLASLT_WORKSPACE_SIZE (2 * 128 * 1024 * 1024)
```
pfultz2, line 30: Create a constant variable, dont use a macro.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
```diff
@@ -0,0 +1,67 @@
...
+#ifndef CHECK_HIPBLAS_ERROR
+#define CHECK_HIPBLAS_ERROR(error)                    \
```
pfultz2, line 32: Create a hipblaslt_invoke function to call the API and check the error code.

pfultz2:
You need to enable hipblaslt on the tidy builds as well.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
```diff
@@ -0,0 +1,67 @@
...
+    if(error != HIPBLAS_STATUS_SUCCESS)               \
+    {                                                 \
+        fprintf(stderr,                               \
+                "hipBLAS error: '%s'(%d) at %s:%d\n", \
+                hipblasStatusToString(error),         \
+                error,                                \
+                __FILE__,                             \
+                __LINE__);                            \
+        exit(EXIT_FAILURE);                           \
```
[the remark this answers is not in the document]
ahsan-ca: Done.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
ahsan-ca: Done.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
ahsan-ca: Done.

src/targets/gpu/compile_blas.cpp
```diff
@@ -0,0 +1,80 @@
...
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != "gpu::blas_gemm")
+            continue;
+        auto op        = any_cast<blas_gemm>(ins->get_operator()).op;
+        auto inputs    = ins->inputs();
+        auto alloc =
+            m.insert_instruction(ins, make_op("allocate", {{"shape", to_value(ins->get_shape())}}));
+        inputs.insert(std::prev(inputs.end()), alloc);
+        m.replace_instruction(ins, op, inputs);
```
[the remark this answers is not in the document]
ahsan-ca: Removed `compile_blas`.

src/targets/gpu/include/migraphx/gpu/context.hpp
ahsan-ca: Done.

src/targets/gpu/include/migraphx/gpu/context.hpp
ahsan-ca: Working on this...

src/targets/gpu/hip_gemm_impl.cpp
ahsan-ca: I was running into issues (test case failures) with using managed pointers for this. Need to look deeper into this. Any recommendations(example) on how to use managed pointers for this are welcome :).

src/targets/gpu/include/migraphx/gpu/context.hpp
ahsan-ca: Made the changes to add `allocate_gpu` in lowering.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
```diff
@@ -25,22 +25,46 @@
...
-    {                                                 \
-        fprintf(stderr,                               \
-                "hipBLAS error: '%s'(%d) at %s:%d\n", \
-                hipblasStatusToString(error),         \
-                error,                                \
-                __FILE__,                             \
-                __LINE__);                            \
-        exit(EXIT_FAILURE);                           \
+#ifndef HIPBLASLT_INVOKE_FUNCTION
+#define HIPBLASLT_INVOKE_FUNCTION
```
umangyadav, line 33: remove these ifdefs

causten:
testing with this branch is working.  Just need some approvals.  @TedThemistokleous @CharlieL7

TedThemistokleous (approved):

TedThemistokleous:
@ahsan-ca make sure we don't also need hipblas, not just hipblaslt and -dev/devel equivalent like we had in the SLES docker case if we're still missing deps.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
ahsan-ca: Done. I have also addressed the other comments that we discussed, in person, on Monday. Can you please also take a look at the last few commits to see if the comments have been addressed correctly? Thanks a lot for the help!

src/targets/gpu/include/migraphx/gpu/context.hpp
ahsan-ca: So the change in lowering for `allocate_gpu` would not work. I have removed the change and I will discuss this with Paul when he is back to see what we should do here.

ahsan-ca:
> > testing with this branch is working. Just need some approvals. @TedThemistokleous @CharlieL7
> 
> @causten I discussed the changes (made for addressing comments) with Umang yesterday, I need to update some things in these changes and fix some CI failures too. I will update here once done. Thanks!

I have made all the required changes and updated the PR with latest develop. This is good to go after CI run is green. Thanks! @CharlieL7 its ready for review. Thanks!

CharlieL7 (approved):

src/targets/gpu/include/migraphx/gpu/context.hpp
pfultz2: What the issue that you ran into?

src/targets/gpu/lowering.cpp
```diff
@@ -251,7 +252,12 @@ struct miopen_apply
             assert(refs.size() == 2);
             auto output = insert_allocation(ins, ins->get_shape());
```
pfultz2, line 253: You need to do another `insert_allocation` for the workspace as well.

pfultz2 (changes requested):

src/targets/gpu/include/migraphx/gpu/context.hpp
pfultz2: The workspace should be added before the output parameter.

ahsan-ca:
> You need to enable hipblaslt on the tidy builds as well.

@pfultz2 I was looking at this and the cmake argument for hipblaslt is turned on by default `option(MIGRAPHX_USE_HIPBLASLT "Enable MIGraphX to use hipBLASLt" ON)`. 
`MIGRAPHX_ENABLE_HIPBLASLT_GEMM` is an runtime environment variable that controls turning hipblaslt on/off for code.
In this case, we probably don't need to add anything to the `ci.yaml` file for CI.
Is my understanding correct?

src/targets/gpu/lowering.cpp
ahsan-ca: Made changes for allocating workspace in lowering.

src/targets/gpu/include/migraphx/gpu/context.hpp
ahsan-ca: Done.

hip-clang.docker
```diff
@@ -31,6 +31,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     libnuma-dev \
     miopen-hip \
     rocblas \
+    hipblaslt \
```
pfultz2, line 34: This should be installed in the install_preqs.sh file, not in the dockerfile.

src/targets/gpu/CMakeLists.txt
```diff
@@ -46,6 +49,15 @@ else()
     message(STATUS "MIGraphX build without rocBLAS")
 endif()
 
+if(MIGRAPHX_USE_HIPBLASLT)
+    # hipblaslt
+    find_package(hipblaslt REQUIRED)
+    find_package(hipblas REQUIRED)
```
pfultz2, line 55: Why are we looking for hipblas? We dont use it.

ahsan-ca: Removed.

hip-clang.docker
ahsan-ca: Moved to install_prereqs file.

tools/docker/sles.docker
```diff
@@ -23,7 +23,12 @@ RUN zypper --gpg-auto-import-keys install -y \
     gdb \
     git \
     python3-pip \
-    rpm-build
+    rpm-build \
+    perl-File-BaseDir \
+    libgfortran5 \
+    hipblaslt-devel \
+    hipblas-devel
```
pfultz2, line 30: These should also be added to the install_preqs.

tools/install_prereqs.sh
```diff
@@ -53,6 +53,7 @@ else
     python3-pip \
     python3-venv \
     rocblas-dev \
+    hipblaslt \
```
pfultz2, line 56: dont we need to install the dev package?

tools/docker/sles.docker
ahsan-ca: Done.

tools/install_prereqs.sh
ahsan-ca: Updated.

src/targets/gpu/CMakeLists.txt
ahsan-ca: Seems like hipblaslt uses hipblas. I saw sles failure after removing hipblas:
```
/opt/rocm-6.1.3/include/hipblaslt/hipblaslt.h:49:10: fatal error: 'hipblas/hipblas.h' file not found
   49 | #include <hipblas/hipblas.h>
      |          ^~~~~~~~~~~~~~~~~~~
```

ahsan-ca: This seems to be causing this error too:
```
File "/src/AMDMIGraphX/tools/accuracy/accuracy_checker.py", line 26, in
import migraphx
ImportError: /src/AMDMIGraphX/build/lib/libmigraphx_gpu.so.2011000: undefined symbol: hipblasSetStream
```

Dockerfile
```diff
@@ -53,8 +53,10 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
     rocm-smi-lib \
     rocm-dev \
     roctracer-dev \
+    libgfortran5 \
     hipcub  \
     hipblas  \
+    hipblaslt \
```
ahsan-ca, line 59: Remove this and libgfortran5.

src/targets/gpu/CMakeLists.txt
ahsan-ca: Ahsan to put a comment to add that we are doing this to workaround the broken hipblaslt package.
To Do:
1. Verify with hipBLASLt team if this is not fixed and open an issue to fix the hipblaslt package:
   a. Dev package needs to be fixed to inclue hipblas header files.
   b. Automatically add hipblas in cmake.

ahsan-ca: Added comment to explain its a workaround.

Dockerfile
ahsan-ca: Done.

src/targets/gpu/CMakeLists.txt
ahsan-ca: Chris checked and the answer to the broken issue is:
```
In 6.2, hipBLAS is a dependency of hipBLASLt, so you should be able to use them together.
In 6.3, the dependencies got change around a bit and you will need to also install a package
called hipBLAS-common.
When you are adding your code now, please don’t assume that just because you have
hipBLASlt installed that means you will have hipBLAS, since this assumption will go away in 6.3.
If you need hipBLAS, install it explicitly and #include it explicitly.
```
This explains how the code in develop that we were looking at would work. They have a new package hipBLAS-common.
```
#ifndef LEGACY_HIPBLAS_DIRECT
#include <hipblas-common/hipblas-common.h>
#else
#include <hipblas/hipblas.h>
#endif
```
@pfultz2 In light of above, do we want to keep hipblas for now since CI is at 6.1. Once we update the CI to 6.3, we can remove hipblas requirement and add `hipBLAS-common` requirement. Does this seem a reasonable approach for now?

ahsan-ca: I missed the step of updating CI to 6.2 before we update it to 6.3 :). I imagine that 6.2 should work fine though (with current workaround), since the changes for hipblaslt side went in, in 6.3.

pfultz2: Yea we need a comment for the workaround since there package is broken in 6.1(and possibly 6.2 as well).

If hipBLAS-common package is needed in 6.3, then they need to make sure its included by default. We are not using hipblas at all.

src/targets/gpu/hipblaslt.cpp
```diff
@@ -0,0 +1,75 @@
...
+static const size_t workspace_size = hipblaslt_workspace_size;
+
+hipblaslt_handle_ptr create_hipblaslt_handle_ptr()
+{
+    hipblasLtHandle_t handle;
+    hipblasLtCreate(&handle);
+    return hipblaslt_handle_ptr{handle};
+}
+
+hipblaslt_handle_ptr create_hipblaslt_handle_ptr(hipStream_t s)
```
pfultz2, line 43: This function can be removed, we already pass the stream when we call hipblasLtMatmul.

ahsan-ca: Sounds good. I will do that.

test/gpu/gemm_tune.cpp
```diff
@@ -130,7 +131,7 @@ TEST_CASE(gemm_tune_strided)
             break;
         }
     }
-#ifdef MIGRAPHX_USE_ROCBLAS_TUNING_API
+#if defined(MIGRAPHX_USE_ROCBLAS_TUNING_API) or defined(MIGRAPHX_USE_HIPBLASLT)
```
bpickrel, line 134: After using `target_compile_definitions` to set `MIGRAPHX_USE_HIPBLASLT` in the makefile, it appears to me that it will always be defined and set to either 0 or 1 at this point, which would make this `#if` always true.  Can you confirm that the code really doesn't get compiled when we don't want it to?

src/targets/gpu/hipblaslt.cpp
ahsan-ca: I have removed the function.

test/gpu/gemm_tune.cpp
ahsan-ca: Thanks for catching this. I will update this.

ahsan-ca: I have updated it to check for the value instead of definition.

ahsan-ca:
The Windows build is passing now.
The PR is good for review, all comments, so far, have been addressed.

test/gpu/gemm_tune.cpp
bpickrel: I wish we had a convention for preprocessor definitions that would make this mistake less likely, but the fix looks good.

bpickrel (approved):

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
```diff
@@ -0,0 +1,109 @@
...
+    return p([=](auto... ws) {
+        auto status = f(ws..., xs...);
+        if(status != HIPBLAS_STATUS_SUCCESS)
+        {
+            MIGRAPHX_THROW("hipblaslt_invoke: hipBlasLt call failed with status " +
+                           std::to_string(status));
+        }
+        return status;
+    });
+}
```
pfultz2, line 86: These functions should be in the migraphx::gpu namespace.

src/targets/gpu/include/migraphx/gpu/hip_gemm_impl.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <migraphx/shape.hpp>
+#include <migraphx/argument.hpp>
+#include <migraphx/gpu/context.hpp>
+
+// Set this environment variable to "true" to perform GEMM tuning even when the
+// --exhaustive-tune option isn't set.  Can be used to skip slow convolution tuning.
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_HIP_GEMM_TUNING);
+
+using milliseconds = std::chrono::duration<double, std::milli>;
+using microseconds = std::chrono::duration<double, std::micro>;
```
pfultz2, line 36: These shoudl be in the migraphx::gpu namespace as well.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <migraphx/errors.hpp>
+#if MIGRAPHX_USE_HIPBLASLT
+#include <hipblaslt/hipblaslt.h>
+#include <hipblaslt/hipblaslt-ext.hpp>
+
+// TODO: Remove hipblasStatusToString() function when hipblaslt
+// provides an API for doing this in hipBLASLt.
+
+// Convert hipblas_status to string
+inline const char* hipblasStatusToString(hipblasStatus_t status)
```
pfultz2, line 38: Function names should use snake_case. I dont know why tidy isn't complaining.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
ahsan-ca: Done.

pfultz2 (approved):

PR #3373, opened 2024-08-14
Update onnx.proto for int4

lakhinderwalia:
Update onnx.proto from https://github.com/onnx/onnx/blob/main/onnx/onnx.proto


causten:
@lakhinderwalia  please fix the format failure

kahmed10 (approved):

kahmed10:
> @lakhinderwalia please fix the format failure

We should be ignoring this file since it's coming from onnx.

pfultz2 (approved):

pfultz2:
Exception didnt work. @causten we should just merge it in. Since the format is done on a diff this wont affect other PRs.

PR #3377, opened 2024-08-15
Remove the hard coded path in MIGRAPHX_CXX_COMPILER

raramakr:
amdclang++/amdclang can be invoked without providing absolute path, since update-alternatives command has been run for these binaries This will also help to remove the hard coded paths in cmake target files for these compilers

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -285,20 +285,6 @@ rocm_install_targets(
     ${CMAKE_CURRENT_BINARY_DIR}/include
 )
 
-set(COMPILER_NAME ${CMAKE_CXX_COMPILER})
-if(NOT WIN32)
-    if(MIGRAPHX_ENABLE_CPU)
-        # Set the compiler as c++ for CPU compilation in linux
-        set(COMPILER_NAME "c++")
-    else()
-        # For GPU compilation, use amdclang
-        get_filename_component(GET_COMPILER_NAME ${CMAKE_CXX_COMPILER} NAME)
-        if(${GET_COMPILER_NAME} MATCHES "^amdclang")
-            set(COMPILER_NAME ${GET_COMPILER_NAME})
-        endif()
-    endif()# end of CPU/GPU check
-endif()
-
 if(NOT WIN32)
     check_cxx_linker_flag(-lstdc++fs HAS_LIB_STD_FILESYSTEM)
     if(HAS_LIB_STD_FILESYSTEM)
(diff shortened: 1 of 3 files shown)


pfultz2:
This should still use the absolute path so we can be sure its using the same compiler the developer set migraphx to build with. I dont see any reason why this should be changed either as it doesnt affect any users.

pfultz2:
It would be better to just disable adding the defines when using hiprtc to avoid the paths in the cmake targets file.

raramakr:
> This should still use the absolute path so we can be sure its using the same compiler the developer set migraphx to build with. I dont see any reason why this should be changed either as it doesnt affect any users.

As part of our efforts to make ROCm relocatable , these hard coded paths need to be removed.

src/CMakeLists.txt
```diff
@@ -283,6 +283,20 @@ rocm_install_targets(
...
+        # Set the compiler as c++ for CPU compilation in linux
+        set(COMPILER_NAME "c++")
+    else()
+        # For GPU compilation, use amdclang
+        get_filename_component(GET_COMPILER_NAME ${CMAKE_CXX_COMPILER} NAME)
+        if(${GET_COMPILER_NAME} MATCHES "^amdclang")
+            set(COMPILER_NAME ${GET_COMPILER_NAME})
+        endif()
+    endif()# end of CPU/GPU check
+endif()
```
pfultz2, line 298: There shouldn't be any changes to cmake. The [compiler_src.hpp](https://github.com/ROCm/AMDMIGraphX/blob/develop/src/include/migraphx/compile_src.hpp#L56) should be updated to only use `MIGRAPHX_CXX_COMPILER` macro when its not windows:

```cpp
struct MIGRAPHX_EXPORT src_compiler
{
#ifdef _WIN_32
    fs::path compiler                         = MIGRAPHX_CXX_COMPILER;
#else
    fs::path compiler                         = "c++";
#endif
    std::vector<std::string> flags            = {};
    fs::path output                           = {};
    fs::path launcher                         = {};
    std::string out_ext                       = ".o";
    std::function<fs::path(fs::path)> process = nullptr;
    std::vector<char> compile(const std::vector<src_file>& srcs) const;
};
```

pfultz2 (changes requested):

src/CMakeLists.txt
raramakr: Thank you @pfultz2 
Updated the  PR as suggested

pfultz2 (approved):

PR #3378, opened 2024-08-15
Add int4 & uint4 types to MigraphX

lakhinderwalia:
Add int4 & uint4 types to MigraphX

### DIFF
--- a/src/api/api.cpp
+++ b/src/api/api.cpp
@@ -98,6 +98,10 @@ shape::type_t to_shape_type(migraphx_shape_datatype_t t)
     switch(t)
     {
     case migraphx_shape_tuple_type: return shape::tuple_type;
+
+        //    case migraphx_shape_uint4_type: return shape::uint4_type;
+        //    case migraphx_shape_int4_type: return shape::int4_type;
+
 #define MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT(x, y) \
     case migraphx_shape_##x: return shape::x;
         MIGRAPHX_SHAPE_VISIT_TYPES(MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT)
@@ -111,6 +115,10 @@ migraphx_shape_datatype_t to_shape_type(shape::type_t t)
     switch(t)
     {
     case shape::tuple_type: return migraphx_shape_tuple_type;
+
+      //    case shape::uint4_type: return migraphx_shape_uint4_type;
+      //    case shape::int4_type: return migraphx_shape_int4_type;
+
 #define MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT(x, y) \
     case shape::x: return migraphx_shape_##x;
         MIGRAPHX_SHAPE_VISIT_TYPES(MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT)
--- a/src/api/api.cpp
+++ b/src/api/api.cpp
@@ -115,6 +115,9 @@ migraphx_shape_datatype_t to_shape_type(shape::type_t t)
     case shape::x: return migraphx_shape_##x;
         MIGRAPHX_SHAPE_VISIT_TYPES(MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT)
 #undef MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT
+    case shape::uint4_type:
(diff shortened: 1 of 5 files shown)


src/include/migraphx/shape.hpp
```diff
@@ -68,7 +68,9 @@ struct MIGRAPHX_EXPORT shape
 #define MIGRAPHX_SHAPE_GENERATE_ENUM_TYPES(x, t) x,
     enum type_t
     {
-        MIGRAPHX_SHAPE_VISIT_TYPES(MIGRAPHX_SHAPE_GENERATE_ENUM_TYPES) tuple_type
+        MIGRAPHX_SHAPE_VISIT_TYPES(MIGRAPHX_SHAPE_GENERATE_ENUM_TYPES) tuple_type,
+        uint4_type,
+        int4_type
```
pfultz2, line 73: We should have a macro to list these types.

src/api/api.cpp
```diff
@@ -115,6 +115,9 @@ migraphx_shape_datatype_t to_shape_type(shape::type_t t)
     case shape::x: return migraphx_shape_##x;
         MIGRAPHX_SHAPE_VISIT_TYPES(MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT)
 #undef MIGRAPHX_DETAIL_SHAPE_CASE_CONVERT
+    case shape::uint4_type:
+    case shape::int4_type:
+        break;
```
pfultz2, line 120: This wrong, the enum values should be exposed in the API and not throw an error.

lakhinderwalia: OK.

src/include/migraphx/shape.hpp
```diff
@@ -381,6 +383,8 @@ struct MIGRAPHX_EXPORT shape
     {
         switch(t)
         {
+        case uint4_type:
+        case int4_type:
```
pfultz2, line 387: The tuple visitor should not be called for these types. It should throw an error.

src/shape.cpp
```diff
@@ -48,20 +48,20 @@ struct shape_impl
...
     {
-        assert(t != shape::tuple_type);
+        assert(t != shape::tuple_type && t != shape::uint4_type && t != shape::int4_type);
     }
 
     shape_impl(shape::type_t t, std::vector<std::size_t> l)
         : m_type(t), m_lens(std::move(l)), m_standard(true)
     {
-        assert(t != shape::tuple_type);
+        assert(t != shape::tuple_type && t != shape::uint4_type && t != shape::int4_type);
```
pfultz2, line 57: This assert shouldn't be changed here. It should be valid to construct a shape of int4_type otherwise whats the point of adding this operator.

src/shape.cpp
```diff
@@ -744,7 +752,7 @@ std::vector<shape> flatten(const std::vector<shape>& shapes)
             auto subs = flatten(s.sub_shapes());
             result.insert(result.end(), subs.begin(), subs.end());
         }
-        else
+        else if(s.type() != shape::uint4_type && s.type() != shape::int4_type)
```
pfultz2, line 755: This shouldn't be changed, we should still be able to flatten an tuples of int4_type.

pfultz2:
In addition, we need to add unit tests for this type.

`check_shapes` needs to be updated to throw an error on these types with a flag to allow it for operators like unpack_int4, similar to the `dynamic_allowed` flag.

We also need a way to automatically insert the unpack_int4 for arguments in the onnx parser so we avoid passing it to unssuported operators(which is everything except unpack_int4).

I wont be around the next couple of weeks to help design or review this, so this will have to wait until september to tackle this further.

src/shape.cpp
lakhinderwalia: Thanks. This one is definitely required.

pfultz2:
Also, we need update the bytes calculation to be the correct amount. We also need to throw when calling index calculations.

src/shape.cpp
pfultz2: What do you mean?

lakhinderwalia: I meant you are correct. This constructor code, as you suggest, is definitely required for correct functioning. 
(I had just quickly hacked those places to make sure the CI would pass the compile stage -- as some switch statements etc. were failing.) Thanks.

PR #3385, opened 2024-08-17
Onnx parsers for Quantization & Dequantization: case when scales and zero point tensors of same dimension as the input tensor

lakhinderwalia:
Onnx parsers for Quantization Dequantization: case when scales and zero point tensors of same dimension as the input tensor:
Currently these parsers try to either broadcast or multibroadcast. Also these two operators already handle scales in some inconsistent manner.

### DIFF
--- a/src/onnx/parse_dequantizelinear.cpp
+++ b/src/onnx/parse_dequantizelinear.cpp
@@ -45,37 +45,47 @@ struct parse_dequantizelinear : op_parser<parse_dequantizelinear>
             axis = info.attributes.at("axis").i();
 
         auto input_lens = args[0]->get_shape().lens();
+        auto input_sz   = args[0]->get_shape().elements();
         auto n_dim      = input_lens.size();
 
-        instruction_ref x_scale;
-        if(args[1]->get_shape().elements() != 1)
+        instruction_ref x_scale = args[1];
+        auto scale_sz           = x_scale->get_shape().elements();
+        if(scale_sz != input_sz)
         {
-            auto tuned_axis = tune_axis(n_dim, axis, opd.op_name);
-            x_scale         = info.add_instruction(
-                make_op("broadcast", {{"axis", tuned_axis}, {"out_lens", input_lens}}), args[1]);
-        }
-        else
-        {
-            x_scale = info.add_instruction(make_op("multibroadcast", {{"out_lens", input_lens}}),
-                                           args[1]);
+            if(scale_sz > 1)
+            {
+                auto tuned_axis = tune_axis(n_dim, axis, opd.op_name);
+                x_scale         = info.add_instruction(
+                    make_op("broadcast", {{"axis", tuned_axis}, {"out_lens", input_lens}}),
+                    x_scale);
+            }
+            else
+            {
+                x_scale = info.add_instruction(
+                    make_op("multibroadcast", {{"out_lens", input_lens}}), x_scale);
+            }
         }
 
         if(args.size() == 3)
         {
             auto x_zero_point = args[2];
-            if(x_zero_point->get_shape().elements() != 1)
+            auto zp_sz        = args[2]->get_shape().elements();
+            if(zp_sz != input_sz)
             {
-                auto tuned_axis = tune_axis(n_dim, axis, opd.op_name);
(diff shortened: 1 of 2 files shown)


pfultz2:
> case when scales and zero point tensors of same dimension as the input tensor:

I dont think the scales and zero points are ever the same shape as the input:

- Per-tensor: The scales and zero points are scalars, so we insert a simple broadcast
- Per-axis quantization: The scales and zero points are a 1-D tensor that will be broadcasts across the `axis`(which is defined as an attribute in the ONNX operator,  see [here](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#attributes)). Since its 1-D we need to know the axis(we cant use a multibroadcast here).
- Blocked quantization: The scales and zero points have the same rank but different dimensions than the input. The `axis` of the dimension that is different than the input should be set to the `block_size`(which both `axis` and `block_size` are attributes define in the ONNX operator, see [here](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#attributes)). 

So I dont see how they could ever be the same shape.

lakhinderwalia:
> * Blocked quantization: The scales and zero points have the same rank but different dimensions than the input. The `axis` of the dimension that is different than the input should be set to the `block_size`(which both `axis` and `block_size` are attributes define in the ONNX operator, see [here](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#attributes)).

@pfultz2, I am trying to work with an int4 model without implementing blocked quantization. And assume `block_size = 1`. And thus in simple case here, the scale tensor is the same shape as the input weights.

CharlieL7 (changes requested):
Code changes look fine, should have ONNX parse tests to verify what it does.

lakhinderwalia:
> Code changes look fine, should have ONNX parse tests to verify what it does.

There are tests being added to another PR, along with int4 type tensors in it. If you like I can add one separately, just let me know, thanks. 

(BTW, this case is just a degenerate case of where no broadcast or multicast is required for scales or zero_points.)

pfultz2:
> > * Blocked quantization: The scales and zero points have the same rank but different dimensions than the input. The `axis` of the dimension that is different than the input should be set to the `block_size`(which both `axis` and `block_size` are attributes define in the ONNX operator, see [here](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#attributes)).
> 
> @pfultz2, I am trying to work with an int4 model while blocked quantization is not implemented. And assume `block_size = 1`. And thus in simple case here, the scale tensor is the same shape as the input weights.

#3412 already handles block quantization for any block_size.

lakhinderwalia:
> #3412 already handles block quantization for any block_size.

'Already'? As in back to the future? :-)

CharlieL7:
> > Code changes look fine, should have ONNX parse tests to verify what it does.
> 
> There are tests being added to another PR, along with int4 type tensors in it. If you like I can add one separately, just let me know, thanks.
> 
> (BTW, this case is just a degenerate case of where no broadcast or multicast is required for scales or zero_points.)

Would be better to have a test for this degenerate case, but not essentially right now. Approving

CharlieL7 (approved):

pfultz2 (changes requested):
I dont think this should be merged in. #3412 already handles this case, and if we merge this PR in, its just going to cause merge conflicts for #3412. I dont think this make sense to have all the extra churn for a "degenerate" case which is not used for any models.

lakhinderwalia:
> Would be better to have a test for this degenerate case, but not essentially right now. Approving

Thanks for approving, Charlie. The test graph for the degenerate case has actually already been added in a different PR.

lakhinderwalia:
> I dont think this should be merged in. #3412 already handles this case, and if we merge this PR in, its just going to cause merge conflicts for #3412. I dont think this make sense to have all the extra churn for a "degenerate" case which is not used for any models.

Thanks, @pfultz2, yes at this point we don't want the churn. I would have appreciated your early approval. This work was done 3 weeks ago, and now wasted.

lakhinderwalia:
Closing this one. This work has been subsequently done in [3412](https://github.com/ROCm/AMDMIGraphX/pull/3412).

PR #3389, opened 2024-08-19
Pooling improvements

pfultz2:
- Rewrites the pooling to use the generic reduction which seems to help improve the performance of the kernel(it seems like it generates much better assembly). This does get pooling much closer to miopen. 
- Improves performance for nhwc
- Works for assymetrical padding
- Add env variable to disable miopen pooling


src/targets/gpu/jit/pooling.cpp
```diff
@@ -98,16 +159,25 @@ struct pooling_compiler : compiler<pooling_compiler>
...
+        if(count_include_pad and mode == "average")
+            mode = "average_include_pad";
 
         std::string op = mode + "_pool";
         if(mode == "lpnorm")
             op += "<" + v.at("lp_order").to<std::string>() + ">";
 
-        std::string count_include_pad = v.get("count_include_pad", false) ? "true" : "false";
-
+        // algorithm algo{ctx, inputs.front(), window};
```
kahmed10, line 170:
```suggestion
```

kahmed10 (approved):

TedThemistokleous (approved):

Issue #2812, opened 2024-02-22
Fuse `where` into MLIR attention


CharlieL7:
From the 22 Feb 2024 performance model review of Distilgpt2:

There is a where before the softmax which prevents us from using flash attention:

```
@34 = gpu::code_object[code_object=9224,symbol_name=where_kernel,global=363312,local=1024,](@33,@30,@32) -> half_type, {1, 12, 348, 348}, {1453248, 121104, 348, 1}
@35 = load[offset=0,end=8352](@1) -> half_type, {1, 12, 348, 1}, {4176, 348, 1, 1}
@36 = gpu::code_object[code_object=9296,symbol_name=reduce_max_kernel,global=534528,local=128,](@34,@35) -> half_type, {1, 12, 348, 1}, {4176, 348, 1, 1}
```

MLIR all compile
```
@27 = gpu::code_object[code_object=5704,symbol_name=mul_where_reduce_max_sub_exp_reduce_sum_div_kernel,global=801792,local=192,](@24,@26,@25) -> half_type, {1, 12, 348, 348}, {1453248, 121104, 348, 1}, target_id=0: 0.015923ms, 2%
```

MLIR’s Flash Attention does support `where` before the `softmax` so it might be possible to just fuse it directly in MLIR. We need some tweaking to how we currently fuse FA. Also create a ticket for MLIR about this once investigated some more.

Deliverables:
* Create tickets on the MLIR repo:
  * Confirm that `where` is supported by MLIR
  * Provide an example of this fusion usecase with flash attention
*  Have the `where` fused with the MLIR flash attention kernel

pfultz2:
We probably need to fuse flash attention directly instead of relying on prefuse_ops. In this case we will have gemm+fused_reduce+gemm to match for. Then we will need check for pointwise+softmax in fused_reduce submodule, although softmax wont exist directly.

umangyadav:
> Confirm that where is supported by MLIR

https://github.com/ROCm/rocMLIR/blob/c58ccc552d8968b6e9cb18901b3a33360de6252b/mlir/include/mlir/Dialect/MIGraphX/IR/MIGraphX.td#L99 

MLIR supports `where`

umangyadav:
> We probably need to fuse flash attention directly instead of relying on prefuse_ops.  

IMO, We should do that for all the MLIR fusions ideally.

umangyadav:
> gemm+fused_reduce+gemm 

[...]

PR #3403, opened 2024-08-26
Generalize attention fusion

shivadbhavsar:
Comeplete solution for #2812 

Changes (applicable when mlir attention is enabled):

- No longer use gemm_softmax_gemm matcher in prefuse_ops pass
- Add matcher for softmax base ops
- Look for `dot -> fused_reduce -> dot (-> pointwise)` pattern in fuse_mlir pass
  - for valid attention fusion, fused_reduce should end with softmax base ops
  - create a single fused module consisting of these ops

Verified that the attention fusion works as before on various transformer models in our nas (bert, gpt, etc.)


test/gpu/fuse_mlir.cpp
```diff
@@ -770,6 +770,291 @@ TEST_CASE(conv_add_split_reduce_multi_use_conv)
     EXPECT(p1.sort() == p2.sort());
 }
 
+TEST_CASE(gemm_softmax_gemm)
```
lakhinderwalia, line 773: Please add a couple of lines of documentation, for p1 and p2.

test/module_test.cpp
```diff
@@ -691,4 +691,60 @@ TEST_CASE(fuse_module)
     EXPECT(m1 == m2);
 }
 
+TEST_CASE(get_inputs)
```
lakhinderwalia, line 694: What do these test do, please? The name doesn't explain to me. Maybe add a couple of lines of documentation. Thanks.

src/include/migraphx/match/softmax.hpp
```diff
@@ -0,0 +1,72 @@
...
+namespace detail {
+template <class F>
+struct softmax_matcher
+{
+    F f;
+
+    auto exp_x_minus_max() const
+    {
+        auto x_minus_max =
+            f("sub")(arg(0)(any().bind("x")), arg(1)(skip_broadcasts(f("reduce_max"))));
```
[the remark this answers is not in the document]
shivadbhavsar: yes L781 in fuse_mlir.cpp

src/module.cpp
```diff
@@ -858,6 +858,21 @@ module::get_ins_param_map(const std::vector<instruction_ref>& inputs, bool rever
     return result;
 }
 
+std::vector<instruction_ref>
+module::get_inputs(const std::unordered_map<instruction_ref, instruction_ref>& map_ins) const
+{
+    std::vector<instruction_ref> inputs;
+    auto params = this->get_parameters();
+    sort_params(params);
```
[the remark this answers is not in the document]
shivadbhavsar: L839 above is doing this and so I figured its necessary, maybe @pfultz2 knows better

src/targets/gpu/prefuse_ops.cpp
```diff
@@ -192,8 +179,7 @@ struct find_gemm_softmax_gemm
     auto matcher() const
     {
         auto gemm1 = match::skip(match::name("contiguous"))(match::name("dot")(
-            match::any_of(is_ck_gemm(), is_mlir_gemm(), is_test_gemm(enable_attention))
-                .bind("gemm1")));
+            match::any_of(is_ck_gemm(), is_test_gemm(enable_attention)).bind("gemm1")));
```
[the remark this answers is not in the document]
shivadbhavsar: Im not sure what this means, can you elaborate? 
`is_test_gemm` is used only for the test suite so that these matchers get tested regardless of which environment variables are set

test/module_test.cpp
shivadbhavsar: Its testing the get_inputs function added in module.cpp. Will add docs

test/gpu/fuse_mlir.cpp
shivadbhavsar: Done

src/targets/gpu/prefuse_ops.cpp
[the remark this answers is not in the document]
lakhinderwalia: I just meant that the naming is odd (to me at least). I could understand if it were `is_test` or `is_gemm` or something like that. But this isn't a big deal, just call it `is_nitpicking()`  :-)

lakhinderwalia (approved):

CharlieL7 (approved):
LGTM

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -678,93 +680,140 @@ using find_mlir_standalone_dot_op         = find_mlir_standalone_op<&is_mlir_dot
...
+        // fused_reduce submodule should end with a softmax
+        auto result = match::match_instruction(*m_attn, pw_softmax, match::softmax());
+        if(result.result != pw_softmax)
         {
-            auto bias_input = orig_inputs[2];
-            instruction_ref bias_param =
-                mm->add_parameter("y_bias", bias_input->get_shape().as_standard());
-            bias = mm->add_instruction(make_op("add"), scaled_gemm0, bias_param);
-            inputs.push_back(bias_input);
+            mpm.remove_module(m_attn->name());
```
pfultz2, line 771: Why do you remove the module? That should be removed automatically by DCE afterwards.

src/module.cpp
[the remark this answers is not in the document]
pfultz2: We map the inputs to the alphabetic order of the parameters, so I think this is necessary.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -678,93 +680,140 @@ using find_mlir_standalone_dot_op         = find_mlir_standalone_op<&is_mlir_dot
...
-    void apply(module_pass_manager& mpm, const match::matcher_result& r) const
+    std::unordered_map<instruction_ref, instruction_ref>
+    invert_map_ins(const std::unordered_map<instruction_ref, instruction_ref>& map_ins) const
     {
-        static size_t counter  = 0;
-        module_ref mm          = mpm.create_module("mlir_" + std::to_string(counter++));
-        auto gemm_softmax_gemm = r.instructions["gemm_softmax_gemm"];
-        mm->set_bypass();
+        std::unordered_map<instruction_ref, instruction_ref> inverse_map;
+        for(auto const& [i1, i2] : map_ins)
```
pfultz2, line 700: Can you name the variables like `key, value`? I think it will make it a little clearer.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -678,93 +680,140 @@ using find_mlir_standalone_dot_op         = find_mlir_standalone_op<&is_mlir_dot
...
+                        if(op.name() == "pointwise")
+                        {
+                            auto* sub_pm     = mod_args.front();
+                            auto param_map_2 = create_param_map_with_literals(
+                                &main_mod, sub_pm, op.compute_shape(to_shapes(inputs), mod_args));
+                            return main_mod.insert_inline(pos, *sub_pm, inputs, &param_map_2)
+                                .front(); // cppcheck-suppress returnDanglingLifetime;
+                        }
+                        return main_mod.insert_instruction(pos, op, inputs, mod_args);
+                    })
```
pfultz2, line 764: This is the same thing from `find_mlir_split_reduce`, can we move this to a function? If the lambda doesnt capture anything(which it looks like it doesnt) then we can just make a static function that we pass for both cases.

src/targets/gpu/fuse_mlir.cpp
shivadbhavsar: I think I had to add this because if we hit this condition in `find_mlir_attention_fused_ops`, then we will also hit it again for `find_mlir_standalone_attention_op`. Which means it will try to create a module with the same name before DCE is run.

shivadbhavsar: Actually, I added dce between those passes, so this can be removed

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -678,93 +680,140 @@ using find_mlir_standalone_dot_op         = find_mlir_standalone_op<&is_mlir_dot
...
+        if(axes.size() != 1)
+            return;
+
+        module_ref m_attn =
+            mpm.create_module("mlir_attn_" + fused_reduce->module_inputs().front()->name());
+        m_attn->set_bypass();
+        std::unordered_map<instruction_ref, instruction_ref> map_main_to_mattn;
+
+        // Add first gemm and fuse any input shape ops
+        module_ref fuse_gemm1 = mpm.create_module("reshapes_gemm1");
```
pfultz2, line 730: Why are you creating a module here? Also, if its only temporary, you could just construct a module varaible directly: `module fuse_gemm1;` instead of adding it to the program.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -678,93 +680,140 @@ using find_mlir_standalone_dot_op         = find_mlir_standalone_op<&is_mlir_dot
...
-        auto scaled_gemm0 = mm->add_instruction(make_op("mul"), gemm0, scale_lit_mbcast);
-
-        std::optional<instruction_ref> bias{nullopt};
-        if(orig_inputs.size() == 4)
+        auto axes = fused_reduce->get_operator().to_value()["axes"];
+        if(axes.size() != 1)
+            return;
+
+        module_ref m_attn =
+            mpm.create_module("mlir_attn_" + fused_reduce->module_inputs().front()->name());
```
pfultz2, line 725: Instead of creating the module and then removing if it doesn't work, you could just construct a module variable directly: `module m_attn` and then later if it can be added as attention you can do `mpm.create_module("mlir_attn_" + fused_reduce->module_inputs().front()->name(), std::move(m_attn))`.

src/include/migraphx/pass_manager.hpp
```diff
@@ -35,15 +35,16 @@ inline namespace MIGRAPHX_INLINE_NS {
...
+    module_pass_manager()                                                                = default;
+    module_pass_manager(const module_pass_manager&)                                      = delete;
+    virtual module& get_module()                                                         = 0;
+    virtual module* create_module(const std::string& name)                               = 0;
     virtual module* create_module(const std::string& name, module m)                     = 0;
     virtual void rename_module(const std::string& old_name, const std::string& new_name) = 0;
-    virtual module* get_common_parent()                    = 0;
-    virtual module* get_root_module()                      = 0;
-    virtual void run_pass(const pass& p)                   = 0;
+    virtual void remove_module(const std::string& name)                                  = 0;
```
pfultz2, line 44: I think this method can be removed since its not used anywhere.

pfultz2 (approved):

shivadbhavsar:
@pfultz2 @causten 
I added a commit here to accept graphs with the extra reshape here, it should fix perf regressions related to that issue. This can probably be generalized to accept some number of shape ops, I think that can be a topic for a separate PR.

PR #3412, opened 2024-09-04
Support block granularity for QuantizeLinear and DequantizeLinear

music-dino:
Add support for block level granularity in [QuantizeLinear](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#summary) and [DequantizeLinear](https://onnx.ai/onnx/operators/onnx__DequantizeLinear.html#summary). 

`y_scale` and `y_zero` point are transformed to match the shape of `x` by applying a unsqueeze->broadcast->reshape chain of transformations. 
If the final block of  `x` is smaller than given `block_size` the transformed `y_scale` and `y_zero_point` are sliced to remove excess elements.

Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/192


src/onnx/quantize_dequantize_linear.cpp
```diff
@@ -0,0 +1,139 @@
...
+        std::transform(args.begin() + 1, args.end(), args.begin() + 1, [&](auto ins) {
+            return info.add_instruction(
+                make_op("broadcast", {{"axis", axis}, {"out_lens", x_lens}}), ins);
+        });
+    }
+    // Blocked granularity
+    else
+    {
+        axis = tune_axis(x_rank, axis, op_name);
+        if(block_size == 0)
```
lakhinderwalia, line 76: Our quark generated graph doesn't use an explicit `block_size`. So this assumption about its being `0` needs to be tweaked a little bit. I think this parameter is optional. So we should work well in case it isn't supplied -- and not assume it is `0` then, and its final value should be computed to be = `block_size_min`. OTOH, if `block_size` is supplied with a model, and it isn't a `0`, then we should compare it within the lower and upper bounds. Thanks.

lakhinderwalia: Therefore, please remove this exception clause.

src/onnx/quantize_dequantize_linear.cpp
```diff
@@ -0,0 +1,139 @@
...
+                               ") shapes may only differ along provided axis(" + to_string(axis) +
+                               ")");
+            }
+        }
+
+        // Given x shape (D0, ..., Di, ..., Dn), y_scale shape (S0, ... Si, ...Sn) and
+        // axis=i, the accepted range is [ceil(Di/Si), ceil(Di/(Si-1))-1]
+        float di           = x_lens[axis];
+        float si           = y_scale_lens[axis];
+        int block_size_min = std::ceil(di / si);
```
lakhinderwalia, line 103: Sample code that can be added below if exception above is removed -- for `block_size == 0`.

`if(block_size ==0) block_size = block_size_min;`

src/onnx/parse_quantizelinear.cpp
```diff
@@ -37,47 +38,74 @@ struct parse_quantizelinear : op_parser<parse_quantizelinear>
...
+                "QuantizeLinear: output_type and y_zero_point type must match. output_type: " +
+                to_string(*output_type) +
+                +", y_zero_point type: " + to_string(args[2]->get_shape().type()));
+        }
+
+        transform_quantize_dequantize_linear_inputs(info, opd.op_name, block_size, axis, args);
 
-            common_args.push_back(y_zero_point);
+        if(parser.opset_version < 19)
+        {
```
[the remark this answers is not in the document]
music-dino: I thought about adding these as well, but decided against mainly because I haven't noticed that it's common practice to have the type constraints checked in parser code, although I might be wrong here.

src/onnx/parse_quantizelinear.cpp
```diff
@@ -37,47 +38,74 @@ struct parse_quantizelinear : op_parser<parse_quantizelinear>
...
     {
+        if(args.size() < 2 or args.size() > 3)
+        {
+            MIGRAPHX_THROW("QuantizeLinear: must have either 2 or 3 inputs, " +
+                           std::to_string(args.size()) + " inputs provided");
+        }
+
+        // Starting with version 19 ONNX introduced the constraint that x and y_scale types must be
+        // the same
+        if(parser.opset_version >= 19 and
```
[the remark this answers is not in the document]
music-dino: We have to flag it because the onnx spec states that it's a constraint for versions 19 and up. 
The common type derivation and conversion could be done for all cases, without a version check condition, but It'd be doing the extra work of common type calculation and looping over the arguments for opset versions 19+ to no avail, since we already know that the types are the same for that case.

src/onnx/quantize_dequantize_linear.cpp
music-dino: The ONNX spec states it is an optional attribute, with a default value of 0: 
https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#attributes

lakhinderwalia: We don't have the quark-generated graph compiling with your current code.  I can change this code later. Thanks.

src/onnx/parse_dequantizelinear.cpp
```diff
@@ -38,49 +39,41 @@ struct parse_dequantizelinear : op_parser<parse_dequantizelinear>
...
-            }
-            else
+            if(args[0]->get_shape().type() != args[2]->get_shape().type())
+                MIGRAPHX_THROW("DequantizeLinear: x and y_zero_point must be of same type");
+
+            if(args[1]->get_shape().lens() != args[2]->get_shape().lens())
             {
-                x_zero_point = info.add_instruction(
-                    make_op("multibroadcast", {{"out_lens", input_lens}}), x_zero_point);
+                MIGRAPHX_THROW("DequantizeLinear: y_scale and y_zero_point shapes must be equal. "
```
lakhinderwalia, line 57: "DequantizeLinear: y_scale and y_zero_point  have mismatched shapes"

src/onnx/parse_quantizelinear.cpp
```diff
@@ -37,47 +38,74 @@ struct parse_quantizelinear : op_parser<parse_quantizelinear>
...
+                +", y_zero_point type: " + to_string(args[2]->get_shape().type()));
+        }
+
+        transform_quantize_dequantize_linear_inputs(info, opd.op_name, block_size, axis, args);
 
-            common_args.push_back(y_zero_point);
+        if(parser.opset_version < 19)
+        {
+            auto common_type = common_shape({args[0]->get_shape(), args[1]->get_shape()}).type();
+            std::transform(args.begin(), args.begin() + 2, args.begin(), [&](auto ins) {
```
lakhinderwalia, line 96: Just trying to understand here: Why is it args.begin() + 2. And not args.end(). Thanks.

src/onnx/parse_quantizelinear.cpp
```diff
@@ -37,47 +38,74 @@ struct parse_quantizelinear : op_parser<parse_quantizelinear>
...
-            auto tuned_axis = tune_axis(n_dim, axis, opd.op_name);
-            y_scale         = info.add_instruction(
-                make_op("broadcast", {{"axis", tuned_axis}, {"out_lens", input_lens}}), args[1]);
+            output_type = get_type(info.attributes.at("output_dtype").i());
         }
 
-        auto common_args = add_common_args(*info.mod, {args[0], y_scale});
-
-        if(args.size() == 3)
+        if(output_type.has_value() and args.size() == 3 and
```
lakhinderwalia, line 82: Style: Please do the exception processing in one clause, on line 59 above.

lakhinderwalia (approved):
Left you some very minor comments. They are optional. 
Approved.

src/onnx/quantize_dequantize_linear.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+namespace onnx {
+
+void transform_quantize_dequantize_linear_inputs(const onnx_parser::node_info& info,
+                                                 const std::string& op_name,
+                                                 int block_size,
+                                                 int axis,
+                                                 std::vector<instruction_ref>& args)
```
CharlieL7, line 39: style: It would be more clear that we're changing the arguments to the Q/DQ if this returned a new `std::vector<instruction_ref`. Since `instruction_ref` is just an iterator, the copy is minimal wrt. memory. With this change the op parsers could also make a copy of the arguments.

CharlieL7 (changes requested):
Please add some onnx parse tests that show the expected MIGraphX IR output from these changes.

src/onnx/parse_quantizelinear.cpp
music-dino: Prior to version 19, the first two inputs(x and y_scales) can have different float types, so a conversion to common type is needed to make the mgx operator work. The optional third input will have a type of int8 or uint8, and we want to leave it that way.

src/onnx/quantize_dequantize_linear.cpp
music-dino: Done

music-dino:
> Please add some onnx parse tests that show the expected MIGraphX IR output from these changes.

I've added a couple of parse tests.

CharlieL7 (approved):

PR #3428, opened 2024-09-06
Move auto_contiguous before lowering

pfultz2:
This should fix the perf isssues.


causten:
Perf report looks pretty bad.  Is there something else needed?

pfultz2:
Perf regression should be fixed by #3494.

TedThemistokleous:
> Perf regression should be fixed by #3494.

Do we close this out then or get 3494 in before this?

shivadbhavsar (approved):
LGTM, but wait to merge till other PR is merged first and verify perf drop fix

PR #3445, opened 2024-09-16
Parser changes to handle MatMulIntegerToFloat

TedThemistokleous:
Changes to MatMul parser to handle the Microsoft Contrib operator MatMulintegarToFloat

Since we have the scale and zero points in our operands we can just perform a multiplied after int8 biases are added and then insert a regular dot on the scaled input values which should give the same output as the input data types. 

Able to leverage the existing set of tests for matmul

~Needs https://github.com/ROCm/AMDMIGraphX/pull/3526 as there's a bug with dequantizelinear this has uncovered~


TedThemistokleous:
TODO:

- [ ] Add Parser tests

src/onnx/parse_matmul.cpp
```diff
@@ -106,6 +108,62 @@ struct parse_matmul : op_parser<parse_matmul>
...
+                                         const std::vector<instruction_ref>& args,
+                                         const int index)
+    {
+        instruction_ref scale_arg                            = args[index];
+        std::set<migraphx::shape::type_t> supported_dq_types = {migraphx::shape::float_type,
+                                                                migraphx::shape::half_type};
+
+        if(not(contains(supported_dq_types, scale_arg->get_shape().type())))
+        {
+            MIGRAPHX_THROW("PARSE_QUANT_DOT_SCALDED: Scales must be float or half_type");
```
lakhinderwalia, line 121: SCALDED?

src/onnx/parse_matmul.cpp
```diff
@@ -106,6 +108,62 @@ struct parse_matmul : op_parser<parse_matmul>
...
+
+    static instruction_ref set_scale_bias(const std::vector<instruction_ref>& args,
+                                          const int index,
+                                          const migraphx::shape& scale_arg_shape,
+                                          const instruction_ref& compare_arg,
+                                          bool& has_valid_scale_bias)
+    {
+        has_valid_scale_bias = false;
+
+        if(args.size() > index)
```
lakhinderwalia, line 140: I don't see an index defined in MatMulIntegerToFloat. Is this for some other operator. Thanks.

src/onnx/parse_matmul.cpp
lakhinderwalia: Also, would this message be proper for `MatMul` operator?

src/onnx/parse_matmul.cpp
```diff
@@ -106,6 +108,62 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            {
+                MIGRAPHX_THROW("PARSE_QUANT_DOT_SCALED: Bias must be the same type as scales");
+            }
+
+            if(scale_bias_arg->get_shape().lens().at(0) != compare_arg->get_shape().lens().at(1))
+            {
+                MIGRAPHX_THROW("PARSE_QUANT_DOT_SCALED: Bias have same dim as matrix B column");
+            }
+
+            has_valid_scale_bias = true;
```
lakhinderwalia, line 161: As against `invalid`? ;-)

src/onnx/parse_matmul.cpp
```diff
@@ -148,6 +206,108 @@ struct parse_matmul : op_parser<parse_matmul>
         }
     }
 
+    static void handle_scaled_transposes(const onnx_parser::node_info& info,
+                                         instruction_ref& scale_a0,
+                                         instruction_ref& zp_a0,
+                                         bool no_zp)
+    {
+        if(no_zp)
```
lakhinderwalia, line 214: It seems only `zp_a0` needs to be in the if clause..?

src/onnx/parse_matmul.cpp
```diff
@@ -148,6 +206,108 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            auto bc_zp_a0 = info.add_instruction(
+                make_op("multibroadcast", {{"out_lens", a0->get_shape().lens()}}), zp_a0);
+
+            dequantized_op =
+                info.add_instruction(make_op("dequantizelinear"), a0, bc_scale_a0, bc_zp_a0);
+        }
+        return dequantized_op;
+    }
+
+    static instruction_ref handle_scaled_output(const onnx_parser::node_info& info,
```
lakhinderwalia, line 255: Too many parameters. Ideally they should be handled by a `struct` parameter.

src/onnx/parse_matmul.cpp
```diff
@@ -173,12 +333,20 @@ struct parse_matmul : op_parser<parse_matmul>
         }
 
         auto is_quant_dot = opd.op_name == "quant_dot";
+        auto has_scales   = opd.op_name == "quant_dot_scaled";
```
lakhinderwalia, line 336: A little confusing naming convention. Between `quant_dots` and `Matmul`**. And then there is `has_scales`: which is presumably also a `quant_dot`.

src/onnx/parse_matmul.cpp
```diff
@@ -200,23 +368,50 @@ struct parse_matmul : op_parser<parse_matmul>
             auto s0_lens        = a0->get_shape().lens();
             auto s1_lens        = a1->get_shape().lens();
 
-            if(not is_quant_dot and args.size() > 2)
+            if(not is_quant_dot and args.size() > 2 and not has_scales)
```
lakhinderwalia, line 371: Would it be simpler to just check if it is just a `dot`, instead of looking for `quant_dot` and `quant_dot_scaled`, as this clause seems to be doing? Thanks.

src/onnx/parse_matmul.cpp
```diff
@@ -200,23 +368,50 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            }
 
             // Only INT8 or UINT8 type currently supported
             std::set<migraphx::shape::type_t> supported_types = {migraphx::shape::uint8_type,
                                                                  migraphx::shape::int8_type};
             const auto a0_type                                = a0->get_shape().type();
             const auto a1_type                                = a1->get_shape().type();
 
-            if(is_quant_dot and
+            if((is_quant_dot or has_scales) and
```
lakhinderwalia, line 414: If it `has_scales`, then it is perhaps not a `MATLMULINTEGER`: as shown in the exception message.

src/onnx/parse_matmul.cpp
TedThemistokleous: It won't reach there as its gated by whether the operator contains the scaled inputs. This variant of MatMul also includes the dequantize to convert the quantized input types to float

src/onnx/parse_matmul.cpp
TedThemistokleous: Its argument index. We're doing the check here so its done for every arg

src/onnx/parse_matmul.cpp
TedThemistokleous: If scale bias doesn't exist there isn't a bias at the end of the matmulintergertofloat added then.

src/onnx/parse_matmul.cpp
TedThemistokleous: No the second input as its bound by column of the input vector (which is 1-d always). 

I broke this out instead of adding it inline to encapsulate logic.

src/onnx/parse_matmul.cpp
TedThemistokleous: They're the same amount of a parameters gathered by the operator. These are all needed for dequantize steps and adding the proper unsqueeze->transpose paths. Order matters here with respect to matrix input A or B

src/onnx/parse_matmul.cpp
TedThemistokleous: What do you suggest I name it? quant_dot_dequant? This operator essentially takes in quantized input and dequantizes the output.

src/onnx/parse_matmul.cpp
TedThemistokleous: Sure that can easily be swapped

src/onnx/parse_matmul.cpp
TedThemistokleous: Sure, simple to just add op.name() here as part of the string. Both MatMulInteger and MatMulIntegerToFloat have the same error on this

src/onnx/parse_matmul.cpp
TedThemistokleous: renamed it

src/onnx/parse_matmul.cpp
TedThemistokleous: The parser changes here due to how we handle the input arg. Arguments are shifted for the Microsoft variant

src/onnx/parse_matmul.cpp
TedThemistokleous: Added naming here instead so errors shared by parsers show the correct parser name

src/onnx/parse_matmul.cpp
```diff
@@ -200,61 +364,102 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            }
+
+            instruction_ref ba0 = set_bias_arg(op_name, args, a0_zp_index, a0, has_ba0);
+            instruction_ref ba1 = set_bias_arg(op_name, args, a1_zp_index, a1, has_ba1);
+
+            // handle optional bias arg to the result
+            instruction_ref scaled_bias;
+            if(has_scales)
+            {
+                scaled_bias = set_scale_bias(args, 6, scale_a1->get_shape(), a1, has_scale_bias);
```
lakhinderwalia, line 401: Please give a name to `6`, just as for `a0_zp_index` above.

src/onnx/parse_matmul.cpp
```diff
@@ -200,61 +364,102 @@ struct parse_matmul : op_parser<parse_matmul>
...
+            }
 
             // Only INT8 or UINT8 type currently supported
             std::set<migraphx::shape::type_t> supported_types = {migraphx::shape::uint8_type,
                                                                  migraphx::shape::int8_type};
             const auto a0_type                                = a0->get_shape().type();
             const auto a1_type                                = a1->get_shape().type();
 
-            if(is_quant_dot and
+            if((not is_dot) and
```
lakhinderwalia, line 410: Style: `if(is_quant_dot || is_quant_dot_scaled)` is much clearer of the intent rather than a negative `if(not is_dot)`. Thanks.

src/onnx/parse_matmul.cpp
```diff
@@ -200,61 +364,102 @@ struct parse_matmul : op_parser<parse_matmul>
             auto s0_lens        = a0->get_shape().lens();
             auto s1_lens        = a1->get_shape().lens();
 
-            if(not is_quant_dot and args.size() > 2)
+            if(is_dot and args.size() > 2)
             {
-                MIGRAPHX_THROW("PARSE_MATMUL: Bias Args not supported for MatMul");
+                MIGRAPHX_THROW(op_name + ": Bias Args not supported for MatMul");
```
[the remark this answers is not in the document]
TedThemistokleous: MatMul on its own doesn't have bias used as an input

src/onnx/parse_matmul.cpp
```diff
@@ -200,61 +364,102 @@ struct parse_matmul : op_parser<parse_matmul>
...
             {
-                shift_input_and_bias(info, offset_op, has_ba0, a0, ba0);
+                if(((a0_type == migraphx::shape::uint8_type) or
+                    (a1_type == migraphx::shape::uint8_type)))
+                {
+                    offset_op = info.add_literal(
+                        migraphx::literal{migraphx::shape{migraphx::shape::half_type}, {-128}});
+                }
+
+                // always convert uint8 to int8 to avoid rollover
```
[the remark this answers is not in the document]
TedThemistokleous: I'm trying to contain stuff for quant_dot and quant_dot_scaled. Quant_dot doesn't need a dequantized output which resolves to float regardless of uint8 or int8 input. In the quant_dot case, there's shifts added to make both inputs match

src/onnx/parse_matmul.cpp
TedThemistokleous: Done

src/onnx/parse_matmul.cpp
TedThemistokleous: Done

src/onnx/parse_matmul.cpp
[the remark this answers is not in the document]
TedThemistokleous: I cleaned this section up so that its just one repeated call per input to handle uint8 for quant_dot case

src/onnx/parse_matmul.cpp
TedThemistokleous: Okay renamed it to is_quant_dot_scaled

src/onnx/parse_matmul.cpp
```diff
@@ -106,7 +108,64 @@ struct parse_matmul : op_parser<parse_matmul>
...
+
+        if(args.size() > index)
+        {
+            instruction_ref scale_bias_arg                       = args[index];
+            std::set<migraphx::shape::type_t> supported_dq_types = {migraphx::shape::float_type,
+                                                                    migraphx::shape::half_type};
+
+            if(not(contains(supported_dq_types, scale_bias_arg->get_shape().type())))
+            {
+                MIGRAPHX_THROW("PARSE_QUANT_DOT_SCALDED: Bias must be float or half_type");
```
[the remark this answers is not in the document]
TedThemistokleous: Fixed

src/onnx/parse_matmul.cpp
lakhinderwalia: I don't see it. Perhaps you have some pending changes somewhere.

src/onnx/parse_matmul.cpp
lakhinderwalia: (sorry, I am looking at stale change..)

src/onnx/parse_matmul.cpp
TedThemistokleous: its changed as part of the update

src/onnx/parse_matmul.cpp
TedThemistokleous: Done

src/onnx/parse_matmul.cpp
TedThemistokleous: Done

lakhinderwalia (approved):
Thank you for following up the comments. Approved.

CharlieL7 (changes requested):
From our conversation, need to test/handle higher dimensional matrix contractions (matrix mul). Also transpose with permutation = `{0, 1}` probably does nothing.

TedThemistokleous:
> From our conversation, need to test/handle higher dimensional matrix contractions (matrix mul). Also transpose with permutation = `{0, 1}` probably does nothing.

You're right but I think when I tried this wasn't doing what I thought it was in conjunction with some the squeeze at -1 . Fixed this an realized I don't need the transpose here. 

Also I think this should handle N-dim now since everything is broadcasted on a per column basis for scale/bias inputs which should be matched to the column of the input matrix.

src/onnx/parse_matmul.cpp
```diff
@@ -148,11 +224,124 @@ struct parse_matmul : op_parser<parse_matmul>
...
+        bool a0_has_no_zp = (a0 == zp_a0);
+        bool a1_has_no_zp = (a1 == zp_a1);
+
+        if(not a0_has_no_zp)
+        {
+            unsq_zp_a0 = info.add_instruction(make_op("unsqueeze", {{"axes", {0}}}), zp_a0);
+            if(zp_a0->get_shape().scalar())
+            {
+                unsq_zp_a0 =
+                    info.add_instruction(make_op("unsqueeze", {{"axes", {0}}}), unsq_zp_a0);
```
CharlieL7, line 294: Minor: Can these unsqueeze operators be merged to either be `unsqueeze {{axes, 0}}` and `unsqueeze {{axes, 0, 1}`?

CharlieL7 (approved):
Minor: would be good to have a higher ndim test

src/onnx/parse_matmul.cpp
TedThemistokleous: I dont think so here as one is assume the input is scalar and needs the additional added 1 dimension

PR #3448, opened 2024-09-16
Call register function so ROCMLIR_DEBUG_FLAGS can be used to pass options to libRockCompiler.

pcf000:
Call register function so ROCMLIR_DEBUG_FLAGS can be used to pass options to libRockCompiler.

Typical anticipated use is something like

    MIGRAPHX_GPU_COMPILE_PARALLEL=1 ROCMLIR_DEBUG_FLAGS='--mlir-disable-threading --mlir-print-ir-after-all' ./build/bin/test_verify ...

to print pass-by-pass output from MLIR from with the MIGraphX test program.  The parallel and threading settings are there to keep the debug output clean.

### DIFF
--- a/src/targets/gpu/mlir.cpp
+++ b/src/targets/gpu/mlir.cpp
@@ -729,6 +729,7 @@ struct mlir_program
     void run_high_level_pipeline()
     {
         mlir_pass_manager pm_front{mlirPassManagerCreate(ctx.get())};
+        mlirRegisterRocMLIROptions();
         mlirMIGraphXAddHighLevelPipeline(pm_front.get());
         logger.clear();
         if(mlirLogicalResultIsFailure(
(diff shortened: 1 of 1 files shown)


src/targets/gpu/mlir.cpp
```diff
@@ -729,6 +729,7 @@ struct mlir_program
     void run_high_level_pipeline()
     {
         mlir_pass_manager pm_front{mlirPassManagerCreate(ctx.get())};
+        mlirRegisterRocMLIROptions();
```
pcf000, line 732: The calls have to go here because --mlir-print-ir-after-all, for instance, is a pass-manager option and can't be interpreted with the call-once MLIR registration functions.  That also means we can't hide the call in rocMLIR code.

pfultz2: Also, why cant mlir just call this method in `mlirMIGraphXAddHighLevelPipeline` and `mlirMIGraphXAddBackendPipeline`?

pcf000: None of the calls from the registration function take the pass manager as input.  I think there's cl::opt magic that creates the options local to the pass manager and also enables parsing the command line to find them.

I'll try calling it in the other functions;  that would be a better choice.  I'll be back with the results.

pfultz2: But do you need a mutex for that? This sounds like its not thread-safe.

pcf000:
Closing here because we can do it entirely in rocMLIR.

PR #3451, opened 2024-09-17
Increase iterations in optimize_module

pfultz2:


ahsan-ca (approved):

causten:
Performance is all over the place.  I can't merge this without some more feedback

PR #3460, opened 2024-09-18
driver verify bisect

richagadgil:


src/driver/verify.cpp
```diff
@@ -237,6 +238,57 @@ void verify_reduced_program(const program& p,
...
+                            verify::tolerance tols)
+{
+    const auto* mm = p.get_main_module();
+    auto right         = std::distance(mm->begin(), mm->end());
+    std::size_t left = 0;
+    bool passed;
+
+    std::cout << "Bisect Verify steps: " << right << std::endl;
+    while (left <= right) {
+        std::size_t mid = left + (right - left) / 2;
```
pfultz2, line 281: You need to skip the instructions that are const foldable.

pfultz2: You can probably create a vector to do the mapping:

```cpp
std::vector<std::size_t> trim_map;
std::size_t i = 0;
for(const auto& ins:*mm)
{
    if(not ins.can_eval())
        trim_map.push_back(i);
    i++;
}
```

Then do the bisection over `trim_map` and then use it to get the actual trim value: `verify_bisect(p, trim_map[mid], t, options, vo, inputs, tols)`.

src/dom_info.cpp
```diff
@@ -55,13 +55,25 @@ struct module_visitor
...
+        if(options.ignore_constants)
+        {
+            copy_if(v.get_children(ins), std::back_inserter(children), [&](auto child) {
+                return not child->can_eval();
+            });
+        }
+        else
+        {
+            children = v.get_children(ins);
+        }
```
pfultz2, line 76: Let me revert these changes to the dominator since its not needed anymore.

src/driver/verify.cpp
```diff
@@ -237,6 +241,93 @@ void verify_reduced_program(const program& p,
...
+        {
+            left = mid + 1;
+        }
+        else
+        {
+            failed = trim;
+            right  = mid - 1;
+        }
+    }
+    if(failed != -1)
```
pfultz2, line 325: I would do `if (failed > 0)`.

shivadbhavsar (approved):

PR #3466, opened 2024-09-21
Defaulting HIP_PLATFORM to `amd` on Windows

apwojcik:
The HIP SDK on Windows does not detect the platform correctly. Before calling `find_package(hip ...)`, it needs to be defaulted to "amd".

### DIFF
--- a/src/targets/gpu/CMakeLists.txt
+++ b/src/targets/gpu/CMakeLists.txt
@@ -22,6 +22,11 @@
 # THE SOFTWARE.
 # ####################################################################################
 
+# HIP SDK on Windows does not detect platform correctly, defaulting to "amd"
+if(NOT HIP_PLATFORM AND CMAKE_SYSTEM_NAME STREQUAL "Windows")
+    set(HIP_PLATFORM "amd")
+endif()
+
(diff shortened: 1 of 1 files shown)


pfultz2:
This should be fixed in hip, this is not a migraphx bug. As a workaround you can just pass `-DHIP_PLATFORM=amd` to cmake command until hip fixes it.

causten (approved):

pfultz2:
@causten Why was this merged? My comments was not even addressed. We shouldn't put hip fixes in out repo, they should be fixed in hip. And there is a simple workaround that can already be used. I think I am going to revert this.

Dnonmi:
> @causten Why was this merged? My comments was not even addressed. We shouldn't put hip fixes in out repo, they should be fixed in hip. And there is a simple workaround that can already be used. I think I am going to revert this.

My question is why create a delay for fixes that were already deemed fine for other branches on the same software stack?
Namely MIOpen: https://github.com/ROCm/MIOpen/pull/3263

With this lack of consistency, It's not assuring the hip side of things will be fixed soon so other people can move on with their jobs.

Issue #3177, opened 2024-06-13
Lower unsupported pooling sizes for the CPU to Reference backend


umangyadav:
https://github.com/ROCm/AMDMIGraphX/pull/3081/files#r1638157103

aditya-167:
Hi @umangyadav , I would like to take this if no one is working on this.. (I am new to the graph compiler). Here is my rough idea-

### Suggested Changes:
Upon checking OneDNN has a kernel size limitation (for Direct Algorithm on SIMD): specifies a maximum dimension size of 14 for the kernel ("weights tensor"). This is noe implemented in MIGraphX, which may lead to failures during execution.

`Check Kernel Size and Padding:` Add a check for the kernel size and padding limits before applying the OneDNN pooling operator.

`Fallback to Reference Backend:` If the condition fails (e.g., kernel size exceeds 14 ), replace the pooling operation with a reference backend operation. For invalid configurations, the pooling operation is replaced with a reference backend pooling operator (ref::pooling) CPU based fallback. Valid configurations continue to use the OneDNN pooling operation (dnnl::pooling)

let me know you have suggestions!

umangyadav:
@aditya-167  Thanks for taking initiative. Your suggested changes look good. I suggest you raise a PR with suggested changes and tag @pfultz2  or @CharlieL7  as reviewers who can help you more with changes.

PR #3468, opened 2024-09-22
Fix for Lower unsupported pooling sizes for the CPU to Reference backend #3177

aditya-167:
Hi @pfultz2,@CharlieL7  

I am working on : https://github.com/ROCm/AMDMIGraphX/issues/3177 (this is my first PR) as suggested by @umangyadav that you would help and review my changes.

# Problem Description
OneDNN has a known limitation when handling specific pooling configurations related to padding, stride, and kernel 
size. 

# Solution Overview
OneDNN fails for certain combinations such as padding = {2}, stride = {1}, and lengths = {2}. This configuration leads to a failure in the pooling operation.
OneDNN has a kernel size limitation: The referenced OneDNN documentation specifies a maximum dimension size of 14 for the pooling kernel (referred to as the "weights tensor"). This limitation is not currently enforced by MIGraphX, which can lead to failures during execution.
The goal of this PR is to address these limitations by detecting such problematic pooling configurations and falling back to the reference backend (a CPU-based implementation) when OneDNN is unable to execute the pooling operation.

### Condition Checking:
The pooling operator is analyzed to check if the pooling configuration violates OneDNN's known limitations:
Check if the kernel size (lengths) exceeds the maximum size of 14.
Check for combinations of padding, stride, and kernel size that are known to fail with OneDNN.
If OneDNN cannot execute the pooling operation due to these limitations, the code falls back to the reference backend (a CPU-based pooling implementation).

### Fallback Mechanism:
For invalid configurations, the pooling operation is replaced with a reference backend pooling operator (ref::pooling).
Valid configurations continue to use the OneDNN pooling operation (dnnl::pooling).

## Code Changes:
https://github.com/aditya-167/AMDMIGraphX/blob/054055f93d6e65aa8b3d7c46a701156b97d15b8d/src/targets/cpu/lowering.cpp#L432C5-L453C10

Please suggest if any any improvment needed (this is my first PR) :-) , also let me know I how to test this as I was thinking of creating a test file for the same however I was wondering if there exists already one in the repo that I can add to ? but I couldnt find one for this operation.


test/CMakeLists.txt
```diff
@@ -162,3 +162,4 @@ if(MIGRAPHX_ENABLE_FPGA)
 endif()
 
 
+
```
CharlieL7, line 165: whitespace

test/onnx/.onnxrt-commit
```diff
@@ -1 +1 @@
-da3bd45cdd50782a22fa919a343101683468002d
+59b7b6bb7cbb7bcc86dab590f1b4d5ed50d53dec
```
CharlieL7, line 1: Why was this changed?

CharlieL7 (changes requested):
Looks good overall. Add a verify test in the `test/verify/` to make sure this works. Verify tests run the whole compilation process on all targets. The test should look something like this: [test/verify/test_pooling_autopad.cpp`](https://github.com/ROCm/AMDMIGraphX/blob/develop/test/verify/test_pooling_autopad.cpp)

aditya-167:
> Looks good overall. Add a verify test in the `test/verify/` to make sure this works. Verify tests run the whole compilation process on all targets. The test should look something like this: [test/verify/test_pooling_autopad.cpp`](https://github.com/ROCm/AMDMIGraphX/blob/develop/test/verify/test_pooling_autopad.cpp)

okay, I have added a test: test_pooling_fallback.cpp that checks for 3 cases, one for valid other 2 for kernel size >14 & invalid pooling that should trigger fallback

CharlieL7:
Failing to compile, needs fix

aditya-167:
> Failing to compile, needs fix. Check CI runs for the errors.

I have updated the tests, seems like I was handling it in not the the right way. compiles in my local repo..

CharlieL7:
I'm going to help this PR along and make some changes to pass the rest of the CI.

aditya-167:
> I'm going to help this PR along and make some changes to pass the rest of the CI.

thanks @CharlieL7 .. I reviewed the CI.. it fails at cppcheck and format.. not sure how to resolve this, it compiles and tests completely in my local..

src/targets/cpu/lowering.cpp
```diff
@@ -435,8 +435,23 @@ struct cpu_apply
...
            not v["ceil_mode"].to<bool>() and
-           v["mode"].to<op::pooling_mode>() != op::pooling_mode::lpnorm)
+           v["mode"].to<op::pooling_mode>() != op::pooling_mode::lpnorm){
+            // Check for OneDNN's kernel size limit and padding issues
+            auto lengths = v["lengths"].to_vector<int>();
+            auto padding = v["padding"].to_vector<int>();
+
+            // Ensure that the kernel size does not exceed OneDNN's max of 14
+            if(std::any_of(lengths.begin(), lengths.end(), [](int len) { return len > 14; }) ||
+            std::any_of(padding.begin(), padding.end(), [](int pad) { return pad > 14; }))
```
CharlieL7, line 445: This condition is not tested and I'm pretty sure is not the correct limitation for OneDNN. I'm also not seeing the 14 kernel length limitation in the OneDNN spec: https://oneapi-src.github.io/oneDNN/dev_guide_pooling.html#doxid-dev-guide-pooling-1dg-pool-impl-limits.

CharlieL7 (changes requested):
It's not clear to me what the actual limitation for this kernel is from OneDNN. Umang did write a TODO comment, but the link provided does not mention such limitations. Need to find a source that lays out the limitation or other proof that OneDNN cannot handle these cases.

aditya-167:
> It's not clear to me what the actual limitation for this kernel is from OneDNN. Umang did write a TODO comment, but the link provided does not mention such limitations. Need to find a source that lays out the limitation or other proof that OneDNN cannot handle these cases.



> It's not clear to me what the actual limitation for this kernel is from OneDNN. Umang did write a TODO comment, but the link provided does not mention such limitations. Need to find a source that lays out the limitation or other proof that OneDNN cannot handle these cases.

Earlier I was referring to https://oneapi-src.github.io/oneDNN/dev_guide_convolution.html#doxid-dev-guide-convolution if you take a look at oneDNN algorithm section, there you may see "Weights tensor width does not exceed 14." I think I mixed up few things here.. I need more digging into this and test few cases where it fails..

CharlieL7:
Right, I see the limit for the direct algorithm for convolution; but this is for pooling. Technically pooling can be implemented as a convolution, but the pooling page doesn't mention anything about having the same limitations.

aditya-167:
@pfultz2 as rightly pointed by @CharlieL7, the doc OneDNN spec: **https://oneapi-src.github.io/oneDNN/dev_guide_pooling.html#doxid-dev-guide-pooling-1dg-pool-impl-limits** doesnt explicitly mention about length size and stride for pooling, earlier I was referring to conv.. however in the comments @umangyadav mentioned about two cases on padding, when length = {3} fails. Now my question is do I have to test each configuration for pooling, and check when it first fails, and put on those condition to fallback on cpu? I am thinking of using binary search for a range of lengths, where first it starts to fail and put the condition for refernce fallback cpu.

CharlieL7:
> Now my question is do I have to test each configuration for pooling, and check when it first fails, and put on those condition to fallback on cpu? I am thinking of using binary search for a range of lengths, where first it starts to fail and put the condition for refernce fallback cpu.

Hello @aditya-167, your search idea sounds like a good plan. It might be faster to ask the OneDNN team for the limitations. Also, we're looking in to changing over to ZenDNN (https://github.com/amd/ZenDNN) over OneDNN in the future.

CharlieL7:
Converted to draft because this PR is stale.

adityas-amd:
> Converted to draft because this PR is stale.

@CharlieL7 is this still relevant? I originally pushed these changes before I joined AMD, so I wanted to confirm whether we should still move this forward?

CharlieL7:
> > Converted to draft because this PR is stale.
> 
> @CharlieL7 is this still relevant? I originally pushed these changes before I joined AMD, so I wanted to confirm whether we should still move this forward?

This is not a priority right now so you can close this if you think the changes here will not be used in the foreseeable future.

PR #3469, opened 2024-09-23
Make hipcc part of install prereqs and cleanup commented line

TedThemistokleous:


hip-clang.docker
```diff
@@ -45,9 +44,6 @@ RUN ldconfig
 # Workaround broken miopen cmake files
 RUN sed -i 's,;/usr/lib/x86_64-linux-gnu/librt.so,,g' /opt/rocm/lib/cmake/miopen/miopen-targets.cmake
 
-# Workaround for broken hip-clang
-# RUN sed -i -e '/if(WIN32)/, /endif()/d' /opt/rocm/lib/cmake/hip/hip-config.cmake && sed -i -n 5p /opt/rocm/lib/cmake/hip/hip-config.cmake
```
pfultz2, line 49: I think the workaround can be removed from sles as well.

tools/install_prereqs.sh
```diff
@@ -60,6 +60,7 @@ else
     libgfortran5 \
     hipblas-dev \
     hipblaslt-dev \
+    hipcc \
```
pfultz2, line 63: There should probably be a comment that this is a workaround for broken hip-dev package.

pfultz2 (approved):

ahsan-ca (approved):

hip-clang.docker
TedThemistokleous: removed

tools/install_prereqs.sh
TedThemistokleous: Added above

PR #3473, opened 2024-09-23
add migraphx load and save functions, allowing the hiding of an error message

apwojcik:
The MIGraphX EP uses the `migraphx_load` and `migraphx_save` functions for MXR caching. If a model is not yet compiled (file missing from cache), a message in the console states an error occurred, which misled users. The v2 version of the functions allows error message reporting to be turned off.

The PR introduces the `migraphx_load_v2` and `migraphx_save_v2` functions. The original functions were replaced with macros, emulating the old behavior with the help of v2 functions.

### DIFF
--- a/src/api/api.cpp
+++ b/src/api/api.cpp
@@ -1805,27 +1805,35 @@ migraphx_operation_name(char* out, size_t out_size, migraphx_operation_t operati
     return api_error_result;
 }
 
-extern "C" migraphx_status
-migraphx_load(migraphx_program_t* out, const char* name, migraphx_file_options_t options)
+extern "C" migraphx_status migraphx_load_v2(migraphx_program_t* out,
+                                            const char* name,
+                                            migraphx_file_options_t options,
+                                            bool output_error)
 {
-    auto api_error_result = migraphx::try_([&] {
-        if(options == nullptr)
-            MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter options: Null pointer");
-        *out = allocate<migraphx_program_t>(migraphx::load((name), (options->object)));
-    });
+    auto api_error_result = migraphx::try_(
+        [&] {
+            if(options == nullptr)
+                MIGRAPHX_THROW(migraphx_status_bad_param, "Bad parameter options: Null pointer");
+            *out = allocate<migraphx_program_t>(migraphx::load((name), (options->object)));
+        },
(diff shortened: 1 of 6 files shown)


TedThemistokleous:
Did you test this with a rebuild and the latest stuff off repo.radeon.com? 

Concerned since we're removing the old signature and Cpp name mangles based on arguments you'll have a runtime  issue between libraries if a user updates MIGraphX package with this change but doesn't rebuild Onnxruntime with the new macro. 

@apwojcik can you try the following with this change

1.  Install the latest Onnxruntime from repo.radeon on your system with MIGraphX latest
2 . Build this change as a package and install it over the existing MIGraphX in the system
3.  Try to run a model with the Onnxruntime MIGraphX EP 

If you hit an error you may need to put the old signature back in for now.

pfultz2:
This breaks binary compatibility. Why cant we just check in onnxruntime if the file exists first before we load it?

pfultz2:
> Concerned since we're removing the old signature and Cpp name mangles based on arguments you'll have a runtime issue between libraries if a user updates MIGraphX package with this change but doesn't rebuild Onnxruntime with the new macro.

There is no name mangling since this a C API, but you are correct that binary compatibility is broken. A simpler solution would be to just check if the file exists before loading it.

apwojcik:
> This breaks binary compatibility. Why cant we just check in onnxruntime if the file exists first before we load it?

That would be too easy :-) But you have a valid point. I am closing this PR. Thanks!

apwojcik:
> Did you test this with a rebuild and the latest stuff off repo.radeon.com?
> 
> Concerned since we're removing the old signature and Cpp name mangles based on arguments you'll have a runtime issue between libraries if a user updates MIGraphX package with this change but doesn't rebuild Onnxruntime with the new macro.
> 
> @apwojcik can you try the following with this change
> 
> 1. Install the latest Onnxruntime from repo.radeon on your system with MIGraphX latest
>    2 . Build this change as a package and install it over the existing MIGraphX in the system
> 2. Try to run a model with the Onnxruntime MIGraphX EP
> 
> If you hit an error you may need to put the old signature back in for now.

Ted, your concern is justified. I have not considered such a scenario. Anyway, I am closing the PR and using Paul's suggested more straightforward solution.

apwojcik:
Closing the PR.

PR #3474, opened 2024-09-25
Update RNN-T Imports

richagadgil:
Imports causing RNN-T Errors


causten (approved):

examples/nlp/python_rnnt/torch_requirements.txt
```diff
@@ -0,0 +1,2 @@
+--index-url https://download.pytorch.org/whl/rocm6.1/
```
shivadbhavsar, line 1: The readme will need instructions to also install these reqs i think

spolifroni-amd (approved):
No changes to any wording/language, so it's fine from a docs perspective.

shivadbhavsar (approved):

PR #3480, opened 2024-09-25
Add check for fusable ops to attention fusion

### LINKED ISSUE #3479: unregistered operation 'migraphx.max' found in dialect ('migraphx')
MLIR attention does not seem to support the pointwise max operation, need to add check for it in the fuse_mlir pass.

Adjacent problem to #3439. Occurs when compiling the same model using mlir attention.

shivadbhavsar:
Not all pointwise ops are mlir fusable, and so this check is required. Fixes #3479


test/gpu/fuse_mlir.cpp
```diff
@@ -1069,6 +1069,136 @@ TEST_CASE(gemm_pw_softmax_gemm_pw)
...
+            {gemm1, eight, ten},
+            {3},
+            [&](auto* rm,
+                const auto& inputs,
+                const auto& axes) -> std::vector<migraphx::instruction_ref> {
+                auto pw = add_pointwise(
+                    p1, rm, "main:pointwise0", inputs, [](auto* pm, const auto& pw_inputs) {
+                        auto mul = pm->add_instruction(
+                            migraphx::make_op("mul"), pw_inputs[0], pw_inputs[1]);
+                        return pm->add_instruction(migraphx::make_op("max"), mul, pw_inputs[2]);
```
pfultz2, line 1108: It would probably be good to add a pointwise op that mlir will never support(ie its only added for the tests), that way we dont need to update the test when `max` is supported.

CharlieL7 (approved):
LGTM

pfultz2 (approved):

Issue #143, opened 2018-12-20
conv_bias_relu fusion slows down inception_v3 batch size of 2


mei-ye:
inception v3 batch size of 2 has about 2X slow -down with fusion.  I discovered this while I was working on stream-scheduling task and had to temporarily disable fusion.

To get the .onnx file, use Mev's instructions:
https://github.com/mvermeulen/migraphx_onnx

In specific, I used the following script to create .onnx file.
import torch
import torchvision.models as models

batch_size = 2

inception = models.inception_v3(pretrained=True)
inception.eval()
torch.onnx.export(inception, torch.randn(batch_size,3,299,299), "inception.onnx")

mvermeulen:
Not sure on the causes, but using HCC_PROFILE=2 and then rpt on perf_onnx (minus the compile step and minus the individual instruction timing); suggests that there might also be a "mis-tuned" situation for batch size 2 relative to batch size 4.

I've attached files but item that stands out is the kernel "gcnAsmConv1x1U" which takes 12x longer in batch size 2 than in batch size 4 - and hence dominates 60% of the batch size 2 time and is only 10% of the batch size 4 time.

[inceptionv3i4.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/2852512/inceptionv3i4.txt)
[inceptionv3i2.txt](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/files/2852513/inceptionv3i2.txt)

Running a MIOPEN_FIND_ENFORCE=3 tuning on the batch size 2 to see if this speeds up this case.

mvermeulen:
After tuning, with MIOPEN_FIND_ENFORCE, the speed for inceptionv3 batch size goes from 55 images/sec to 177 images/sec.

This now fits between batch size 1 (164 images/sec) and batch size 4 (222 images/sec), but will also try tuning inceptionv3 to see if that also speeds up.

The tuning added the following entries on my Vega64 system
'''
48-35-35-5x5-64-35-35-2-2x2-1x1-1x1-0-NCHW-FP32-F=ConvOclDirectFwd:16,16,16,32,1,2,4,4,2;ConvOclDirectFwdFused:8,32,16,32,2,1,4,4,2
[...]

PR #3482, opened 2024-09-26
Roialign fix and half_pixel mode support

bpickrel:
Fix bugs in the implementation of ROIAlign operation which were found when attempting to run it with the `half_pixel` coordinate conversion mode, to include more thorough tests.  Some bugs are mode-specific and some are not.

The ROIAlign operation was first proposed in a paper at [https://arxiv.org/abs/1703.06870v3](https://arxiv.org/abs/1703.06870v3) which introduced the `Mask R-CNN` model.  It was a variant of the ROIPool operation which was found to give significantly better accuracy.  In the implementations in Torch, Onnxruntime, and Migraphx, ROIPool and ROIAlign are implemented in the same op. with different choices for the mode attribute, with `output_half_pixel` for ROIPool and `half_pixel` for ROIAlign; thus, there is no ROIAlign op without fixing the `half_pixel` mode.

Note, by the way, that these same coordinate conversion modes are also attributes of the `Resize` op.

MIGraphX uses the Onnxruntime implementation of ROIAlign as its functional specification and should give identical results.

This change is prerequisite for [torch-migraphx PR #143](https://github.com/ROCm/torch_migraphx/issues/143) but does not close it.


bpickrel:
Don't review this yet!  It's very incomplete and I just created the PR to make it easy to visualize the changes so far.

test/onnx/parse/roialign_default_test.cpp
```diff
@@ -36,16 +36,31 @@ TEST_CASE(roialign_default_test)
     auto rois = mm->add_parameter("rois", srois);
     auto bi   = mm->add_parameter("batch_ind", sbi);
 
-    // Due to the onnx model using opset 12, the coordinate_transformation_mode should be set to
-    // output_half_pixel
+    // Depending on whether the model was built for Onnx opset 16 or earlier, the default
+    // coordinate_transformation_mode is different.  These model files had explicit opset given
```
spolifroni-amd, line 40:
```suggestion
    // coordinate_transformation_mode will be different.  These model files had explicit opset given
```

test/onnx/verify/roialign_verify_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+    std::vector<int64_t> bi_data = {1, 0};
+
+    pp["rois"]      = migraphx::argument(srois, rois_data.data());
+    pp["batch_ind"] = migraphx::argument(sbi, bi_data.data());
+
+    auto result = p.eval(pp).back();
+    std::vector<float> result_vector;
+    result.visit([&](auto output) { result_vector.assign(output.begin(), output.end()); });
+
+    // gold results were generated with onnxruntime
```
bpickrel, line 54:
```suggestion
    // gold values were generated with onnxruntime
```

test/onnx/parse/roialign_default_test.cpp
```diff
@@ -36,16 +36,31 @@ TEST_CASE(roialign_default_test)
...
         x,
         rois,
         bi);
     mm->add_return({r});
-
     auto prog = read_onnx("roialign_default_test.onnx");
-
     EXPECT(p == prog);
+
+    // Opset 12 program
```
shivadbhavsar, line 51: I suggest making this a separate test so if there is a failure you know which one of the two opsets had the problem

tools/build_and_test_onnxrt.sh
```diff
@@ -36,8 +36,8 @@ export CXXFLAGS="-D__HIP_PLATFORM_AMD__=1 -w"
...
-echo 'InferenceSessionTests.CheckRunProfilerWithSessionOptions2' >> ../../../tools/ci_build/github/pai/migraphx-excluded-tests.txt
-echo 'InferenceSessionTests.Test3LayerNestedSubgraph' >> ../../../tools/ci_build/github/pai/migraphx-excluded-tests.txt
-echo 'InferenceSessionTests.Test2LayerNestedSubgraph' >> ../../../tools/ci_build/github/pai/migraphx-excluded-tests.txt
-../../../tools/ci_build/github/pai/pai_test_launcher.sh || (gdb ./onnxruntime_test_all core -batch -ex bt && exit 1)
+# echo 'InferenceSessionTests.CheckRunProfilerWithSessionOptions' >> ../../../tools/ci_build/github/pai/migraphx-excluded-tests.txt
+# echo 'InferenceSessionTests.CheckRunProfilerWithSessionOptions2' >> ../../../tools/ci_build/github/pai/migraphx-excluded-tests.txt
+# echo 'InferenceSessionTests.Test3LayerNestedSubgraph' >> ../../../tools/ci_build/github/pai/migraphx-excluded-tests.txt
+# echo 'InferenceSessionTests.Test2LayerNestedSubgraph' >> ../../../tools/ci_build/github/pai/migraphx-excluded-tests.txt
+# ../../../tools/ci_build/github/pai/pai_test_launcher.sh || (gdb ./onnxruntime_test_all core -batch -ex bt && exit 1)
```
shivadbhavsar, line 43: Is this intentional or a leftover debugging artifact?

test/verify/test_roialign_nondefault.cpp
```diff
@@ -40,7 +40,6 @@ struct test_roialign_nondefault : verify_program<test_roialign_nondefault>
 
         migraphx::shape ind_s{migraphx::shape::int64_type, {5}};
         std::vector<int64_t> ind_vec = {0, 2, 3, 4, 1};
-
```
shivadbhavsar, line 43: revert

shivadbhavsar (commented):
Looks fine, just a few small things. I haven't been able to fully wrap my head around all the math in the ref and gpu impl, the index changes look reasonable. Do we have a way to directly test against ORT (without maunally extracting gold outputs)? If so, I think it would be worthwhile to add a few more tests comparing with ORT

tools/build_and_test_onnxrt.sh
bpickrel: reverted

test/verify/test_roialign.cpp
```diff
@@ -44,20 +77,23 @@ struct test_roialign : verify_program<test_roialign<DType>>
...
+            x,
+            roi,
+            ind);
         mm->add_return({r});
 
         return p;
     }
 };
 
+template struct test_roialign_half_pixel<migraphx::shape::float_type>;
```
bpickrel, line 96: Is there any point in adding other data types for the half_pixel (i.e. the "real" ROIAlign) test?

test/onnx/verify/roialign_half_pixel_verify_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    pp["y"] = migraphx::argument(s, data.data());
+
+    migraphx::shape srois{migraphx::shape::float_type, {2, 4}};
+    std::vector<float> rois_data = {1.1, 0.73, 1.7, 1.13, 1.1, 0.73, 2.6, 1.13};
+    migraphx::shape sbi{migraphx::shape::int64_type, {2}}; // batch_index
+    std::vector<int64_t> bi_data = {0, 1};
+
+    pp["rois"]      = migraphx::argument(srois, rois_data.data());
+    pp["batch_ind"] = migraphx::argument(sbi, bi_data.data());
+    pp["y"]         = migraphx::argument(s, data.data());
```
bpickrel, line 48: It's not clear to me whether Y, the output, is even needed for this test.  What role does it play?

test/onnx/parse/roialign_test.cpp
```diff
@@ -41,7 +41,7 @@ TEST_CASE(roialign_test)
                           {{"coordinate_transformation_mode", "output_half_pixel"},
                            {"spatial_scale", 2.0f},
                            {"output_height", 5},
-                           {"output_width", 5},
+                           {"output_width", 3},
```
bpickrel, line 44: This particular change was a big deal, btw, since the old code appeared to work fine until I gave it an output_height and output_width that were not the same.

test/onnx/verify/roialign_half_pixel_verify_test.cpp
shivadbhavsar: Likely because `roialign_half_pixel_test.onnx` expects it as input. I am not sure why its configured like that. Maybe it needs an allocated buffer to write the output to?
I am not very familiar with the verify onnx tests, I would check if this is common in other tests also

test/onnx/verify/roialign_half_pixel_verify_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::vector<float> data(2 * 2 * 4 * 3);
+    std::iota(data.begin(), data.end(), 0.f);
+    migraphx::parameter_map pp;
+    pp["x"] = migraphx::argument(s, data.data());
+    pp["y"] = migraphx::argument(s, data.data());
+
+    migraphx::shape srois{migraphx::shape::float_type, {2, 4}};
+    std::vector<float> rois_data = {1.1, 0.73, 1.7, 1.13, 1.1, 0.73, 2.6, 1.13};
+    migraphx::shape sbi{migraphx::shape::int64_type, {2}}; // batch_index
+    std::vector<int64_t> bi_data = {0, 1};
```
shivadbhavsar, line 44: for good measure, can you add a tests with the following cases:
- Repeated batch indices
- Missing batch indices (ie. not all batch items are computed on)
- Number of ROIs != batch_size 

You can probably just create one test case to get all these. Make the input batch_size 3 and the batch_indices something like {1,2,2,1} (and hence the rois shape will be {4,4})
Would be good to have a gpu verify test for this same case too just to be sure gpu impl matches

docs/dev/onnx_operators.rst
```diff
@@ -697,7 +697,7 @@ Operator Support Matrix
 |                          |           |                 | functions are                |
 |                          |           |                 | not enabled                  |
 +--------------------------+-----------+-----------------+------------------------------+
-| RoiAlign                 | ✅        | FP8, FP16,      |                              |
+| RoiAlign                 | ✅        | FP8, FP16,      |                               |
```
spolifroni-amd, line 700: The extra whitespace at the end of this row is causing the table to be improperly formatted and not appear on the doc page.

```suggestion
| RoiAlign                 | ✅        | FP8, FP16,      |                              |
```

spolifroni-amd (changes requested):
A whitespace in the table needs to be removed.

test/onnx/parse/roialign_default_test.cpp
bpickrel: Done

test/onnx/verify/roialign_half_pixel_verify_test.cpp
bpickrel: Added such a case in ref. tests, along with other updates.  Some of the new cases fail and I'm now debugging those.

test/ref/roialign.cpp
```diff
@@ -84,114 +84,164 @@ TEST_CASE(roialign_out_of_bound_test)
     }
 }
 
+auto create_program(const std::string& trans_mode = "half_pixel",
```
bpickrel, line 87: Notes on ref tests:  added cases with all 4 combinations of trans_mode and pooling_mode and split them apart into separate named cases.  The modified `create_program()` has a reshaped input with multiple channels, multiple layers, an ROI list that doesn't match 1-1 with the layers, non-unity scale and sampling ratio, one negative data value (any float value is legal) and an ROI that goes out of bounds (also legal).  Also, the output height and width are no longer equal, which masked errors in the original implementation.

Need to debug the max pooling cases!

bpickrel:
The licensing check fail now occurring is for a file not related to this PR: 
```
Error: The licenses for the following 1 file(s) either... do not match the year of commit, have a different copyright format or have not been synced from the latest roialign_fix branch:
['src/targets/gpu/kernels/include/migraphx/kernels/float8.hpp']
```

bpickrel:
> Looks fine, just a few small things. I haven't been able to fully wrap my head around all the math in the ref and gpu impl, the index changes look reasonable. Do we have a way to directly test against ORT (without maunally extracting gold outputs)? If so, I think it would be worthwhile to add a few more tests comparing with ORT

I think it would be possible to add a test following the model of the existing tests in test/py/.  With luck it wouldn't be very much extra work, half a day or so.  @pfultz2 what do you think?  The rationale for adding an op test here is that the ROIAlign op is defined in terms of the Onnxruntime implementation so it makes sense to have a specialized test with ORT as the reference.

Note my recent comment that I learned the ORT implementation of the max pooling option is buggy and can't be used for a test reference until the fix is released.  I don't know whether max pooling is widely used with this op or not.

bpickrel:
> Looks fine, just a few small things. I haven't been able to fully wrap my head around all the math in the ref and gpu impl, the index changes look reasonable.

Do you want me to go over it with you?  I can explain the intent of nearly everything but the indexing is still very difficult to unravel.

test/py/test_roialign.py
```diff
@@ -0,0 +1,80 @@
...
+    #     TODO:  clean up this output:
+
+    # root@rocm-rome-6:/workspace/AMDMIGraphX/test/py# python3 ./test_roialign.py 
+    # 2024-10-28 22:13:54.675372471 [W:onnxruntime:, graph.cc:109 MergeShapeInfo] Error merging shape info for output. 'y' source:{2,2,3,2} target:{3,2,2,2}. Falling back to lenient merge.
+    #  XXXXX 0x562d956ec950    (0x562d956ec8f0 + 1 * 2 + channel 0) * 4 * 3
+    #  XXXXX 0x562d956ec980    (0x562d956ec8f0 + 1 * 2 + channel 1) * 4 * 3
+    #  XXXXX 0x562d956ec8f0    (0x562d956ec8f0 + 0 * 2 + channel 0) * 4 * 3
+    #  XXXXX 0x562d956ec920    (0x562d956ec8f0 + 0 * 2 + channel 1) * 4 * 3        
+    res = sess.run(['y'], {'x': data, 'rois': roi_data, 'batch_ind': index_data})
+    assert np.allclose(mgx_result, res, rtol=1e-05, atol=1e-08, equal_nan=False)
```
bpickrel, line 73: tolerances are the Numpy defaults

pfultz2:
You should capture the onnxruntime results and just create a ref test.

test/py/test_roialign.py
```diff
@@ -0,0 +1,75 @@
...
+    })
+    assert np.allclose(mgx_result,
+                       res[-1],
+                       rtol=1e-05,
+                       atol=1e-08,
+                       equal_nan=False)
+
+
+if __name__ == "__main__":
+    test_roialign()
```
pfultz2, line 75: This file is not even used in our test suite. It should just be removed and a ref test should be used.

Issue #3487, opened 2024-09-27
[BUG] Compile Error when importing torch after migraphx


richagadgil:
Description:

model.compile() results in string of hiprtc errors when torch is imported after migraphx

Steps to Reproduce:

1. pip install -r torch_requirements.txt
2. pip install -r requirements.txt 
3. python3 compile_model.py 

[compile_model_failure.zip](https://github.com/user-attachments/files/17170287/compile_model_failure.zip)

Full Error Log:

[compile_error.txt](https://github.com/user-attachments/files/17170350/compile_error.txt)

PR #3493, opened 2024-10-01
Call compile with correct executable 

richagadgil:
Fixing Issue #3487 where compiling with incorrect hiprtc function results in error


shivadbhavsar (approved):

src/targets/gpu/compile_hip.cpp
```diff
@@ -228,9 +228,15 @@ std::vector<std::vector<char>> compile_hip_src_with_hiprtc(std::vector<hiprtc_sr
 bool hip_has_flags(const std::vector<std::string>& flags)
 {
     hiprtc_program prog{" "};
```
pfultz2, line 230: This is variable is not used. Also the other `hip_has_flags` function should be removed.

PR #3496, opened 2024-10-02
Add support for MatMulNBits

music-dino:
Resolves https://github.com/migraphx-benchmark/AMDMIGraphX/issues/193


src/onnx/parse_matmulnbits.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    instruction_ref parse(const op_desc& /*opd*/,
+                          const onnx_parser& parser,
+                          onnx_parser::node_info info,
+                          const std::vector<instruction_ref>& args) const
+    {
+        const int N          = parse_attribute(parser, info, "N");
+        const int K          = parse_attribute(parser, info, "K");
+        const int bits       = parse_attribute(parser, info, "bits");
+        const int block_size = parse_attribute(parser, info, "block_size");
```
TedThemistokleous, line 51: use size_t

TedThemistokleous:
Odd CI is all green but here its showing it didn't run. Rekicked this as I don't think this update is showing up on github.

src/onnx/parse_matmulnbits.cpp
```diff
@@ -0,0 +1,192 @@
...
+        if(bits != 4)
+            MIGRAPHX_THROW("MatMulNBits: bits only supported for value of 4, actual value " +
+                           std::to_string(bits));
+
+        if(block_size < 16 or (block_size & (block_size - 1)) != 0)
+            MIGRAPHX_THROW("MatMulNBits: block_size must be a power of 2 and >=16, actual value " +
+                           std::to_string(block_size));
+
+        const size_t n_blocks_per_col = (k + block_size - 1) / block_size;
+        const size_t blob_size        = std::ceil(block_size * bits / 8.0f);
```
TedThemistokleous, line 61: does this need to be a float for 8.0f here and the ceil? block_size and bits are size_t.  Are you relying on type promotion here? Wouldn't the result still be floored after the divide?

TedThemistokleous (commented):
Overall looks alright. Will see what CI says in the morning. Just another question I have about how you handle blob_sizes.

src/onnx/parse_matmulnbits.cpp
music-dino: Yes, I'm relying on type promotion so the division result won't be floored. Since the divisor is a float(8.0f) the size_t dividend(block_size * bits result) will get promoted to float and float division is performed. The round computes the least integer value not less than its input meaning the mantissa of the result will be 0,  so conversion back to size_t will just cut off zeroes.
However for this particular case it doesn't really matter because block_size is always >=16 and a power of 2 so correct results would be produced even with integer division, but I left it this way to conform with the documentation. The promotion does matter for the expected_zp_lens calculation a couple of lines down.

music-dino:
> Overall looks alright. Will see what CI says in the morning. Just another question I have about how you handle blob_sizes.

There's a clang_debug failure, but I can't access the logs.

TedThemistokleous:
> > Overall looks alright. Will see what CI says in the morning. Just another question I have about how you handle blob_sizes.
> 
> There's a clang_debug failure, but I can't access the logs.

Looks clean from my end. Dont worry about it.

src/onnx/parse_matmulnbits.cpp
TedThemistokleous: Got it just wanted to be clear that was your intention and yes I did see that for the cases below.

TedThemistokleous (approved):
Approved. looks fine and nothing else jumps out at me.

PR #3500, opened 2024-10-03
Adding MIGRAPHX_TIME_MATCHERS

aarushjain29:


pfultz2 (changes requested):
1) The time measurements should be done in the `find_matches_for` function.
2) There should also be an environment variable to enable it like `MIGRAPHX_TIME_MATCHERS`. 
3) Trace filter should probably be used to reduce down the amount of printouts.

src/simplify_algebra.cpp
bpickrel, line 1: Update copyright date (think we have a script for that; can't remember how to run it)

bpickrel:
> 1. The time measurements should be done in the `find_matches_for` function.

@pfultz2 does every single matcher go through `find_matches_for`?  If we're going to add this feature to the `develop` branch then I suggest you add it for every matcher, not just the selected ones that Aarushi added the code to.

> 2. There should also be an environment variable to enable it like `MIGRAPHX_TIME_MATCHERS`.

This is a good way to globally enable/disable match timing.  But maybe we also want to enable/disable timing programmatically for individual matchers?  Aarushi, in your experience would this be helpful?

> 3. Trace filter should probably be used to reduce down the amount of printouts.

I suggest wrapping the timer function in a macro so you can easily insert/remove it from the list in individual passes' `find_matches()` methods, eg. 
```
void simplify_algebra::apply(module& m) const
{
    // Run simplifications multiple times
    m.repeat_while_changes(8, [&] {
        match::find_matches(m,
                            LOG_MATCHER_TIMER(find_inner_broadcast{}),    <===new macro
                            find_dot_broadcast{},
 etc.
```
which would be less repetitive work than wrapping each matcher in curly brackets as Aarushi has done.

What do you think?

pfultz2:
> does every single matcher go through find_matches_for?

Yes.

> This is a good way to globally enable/disable match timing. But maybe we also want to enable/disable timing programmatically for individual matchers?

The `MIGRAPHX_TRACE_MATCHES_FOR` lets you filter the matchers if we use it for timing as well.

> What do you think?

I would like to avoid needing to change the source code to get timings. We dont do it for the passes. This will let us get timings on deployed systems so we dont need to rebuild migraphx.

src/include/migraphx/matcher.hpp
```diff
@@ -415,29 +418,35 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
                 return;
             if(trace > 1 and trace_for)
                 std::cout << "Match: " << matcher_name << std::endl;
-            auto r = match_instruction(get_module(mod), ins, m.matcher());
-            if(r.result == get_module(mod).end())
+            auto elapsed_time = time<std::chrono::nanoseconds>([&] {
+                auto r = match_instruction(get_module(mod), ins, m.matcher());
```
pfultz2, line 422: We should only be timing the match and apply methods. We should also print out the time for each.

src/include/migraphx/matcher.hpp
```diff
@@ -418,34 +418,41 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
                 return;
             if(trace > 1 and trace_for)
                 std::cout << "Match: " << matcher_name << std::endl;
-            auto elapsed_time = time<std::chrono::nanoseconds>([&] {
-                auto r = match_instruction(get_module(mod), ins, m.matcher());
-                if(r.result == get_module(mod).end())
+
+            timer match_timer;
+            auto r = match_instruction(get_module(mod), ins, m.matcher());
+            auto match_time = match_timer.record<std::chrono::nanoseconds>();
```
pfultz2, line 424: Use the `time` function.

src/include/migraphx/matcher.hpp
```diff
@@ -418,34 +418,41 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
...
+            {
+                std::cout << "Matched by " << matcher_name << std::endl;
+                get_module(mod).debug_print(ins);
+            }
+            // If its already invalid dont validate it again
+            bool invalidated = validate and get_module(mod).validate() != get_module(mod).end();
+                
+            timer apply_timer;
+            m.apply(mod, r);
+            auto apply_time = apply_timer.record<std::chrono::nanoseconds>();
```
pfultz2, line 438: Use the `time` function.

src/include/migraphx/matcher.hpp
pfultz2: Also dont use `std::chrono::nanoseconds`, as it uses integral type. Instead use microseconds with floating-point.

src/include/migraphx/matcher.hpp
```diff
@@ -418,34 +418,41 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
...
                 }
-                match = true;
-            });
-            if(time_matchers and trace_for)
+            }
+            match = true;
+            if(time_matchers)
             {
-                std::cout << "Matcher " << matcher_name << " took " << elapsed_time << "ns." << std::endl;
+                std::cout << "Matching for " << matcher_name << " took " << match_time << "ns." << std::endl;
```
pfultz2, line 454: Print this after matching not at the end.

src/include/migraphx/matcher.hpp
```diff
@@ -438,6 +449,11 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
                 }
             }
             match = true;
+            if(time_matchers)
```
pfultz2, line 452: You should also check `trace_for`.

src/include/migraphx/matcher.hpp
aarushjain29: I have used the timer struct from /src/include/migraphx/time.hpp
Should I use separate `time` function for this?

src/include/migraphx/matcher.hpp
```diff
@@ -415,7 +418,16 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
                 return;
             if(trace > 1 and trace_for)
                 std::cout << "Match: " << matcher_name << std::endl;
+
+            timer match_timer;            
             auto r = match_instruction(get_module(mod), ins, m.matcher());
+            auto match_time = match_timer.record<std::chrono::duration<double, std::micro>>();
```
aarushjain29, line 424: We can't directly use the existing time function from time.hpp for timing the match_instruction because it only records the time and doesn't return any value. Since match_instruction needs to assign its result to r, we would require a separate time template function that can both capture the result and measure the elapsed time as overloading is not possible as it is leading to ambiguity in function selection. This additional template would return a std::pair, containing both the result and the time, Any thoughts on this @pfultz2 ?

src/include/migraphx/matcher.hpp
```diff
@@ -415,7 +420,15 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
                 return;
             if(trace > 1 and trace_for)
                 std::cout << "Match: " << matcher_name << std::endl;
+
             auto r = match_instruction(get_module(mod), ins, m.matcher());
+            const auto match_time = match_timer.record<std::chrono::duration<double, std::micro>>();
+            
+            if(time_matchers or trace_for)
+            {
+                std::cout << "Matching for " << matcher_name << " took " << match_time << "ns." << std::endl;
```
bpickrel, line 429:
```suggestion
                std::cout << "Matching for " << matcher_name << " took " << match_time << "us." << std::endl;
```

src/include/migraphx/matcher.hpp
```diff
@@ -425,7 +438,16 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
...
             bool invalidated = validate and get_module(mod).validate() != get_module(mod).end();
-            m.apply(mod, r);
+                
+            auto apply_time = time<std::chrono::duration<double, std::micro>>([&] {
+                m.apply(mod, r);
+            });
+            
+            if(time_matchers or trace_for)
+            {
+                std::cout << "Apply for " << matcher_name << " took " << apply_time << "ns." << std::endl;
```
bpickrel, line 448:
```suggestion
                std::cout << "Apply for " << matcher_name << " took " << apply_time << "us." << std::endl;
```

bpickrel (approved):

src/include/migraphx/matcher.hpp
```diff
@@ -421,7 +425,15 @@ void find_matches_for(source_location location, Mod& mod, instruction_ref ins, M
                 return;
             if(trace > 1 and trace_for)
                 std::cout << "Match: " << matcher_name << std::endl;
+
             auto r = match_instruction(get_module(mod), ins, m.matcher());
+            const auto match_time = match_timer.record<std::chrono::duration<double, std::micro>>();
```
lakhinderwalia, line 430: Perhaps the time measurement should be done only when the flags are set.

lakhinderwalia (approved):
It seems there is very little difference in the debug outputs, between the new environment variable, `MIGRAPHX_TIME_MATCHERS` being set, and the result of `trace_for` option match. 

Does one even need this new variable in that case: unless your print clauses are gated by `if(time_matchers and trace_for)` -- The `or` is replaced by `and`.

I am approving this PR, never-the-less, for it gives out the necessary time-related info that one is interested in.

CharlieL7 (changes requested):
Just add documentation update.

CharlieL7 (approved):

spolifroni-amd (approved):

PR #3507, opened 2024-10-04
MLIR fusion for int4

lakhinderwalia:


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -878,6 +878,50 @@ struct find_pointwise_mlir
...
+        std::unordered_map<instruction_ref, instruction_ref> map_ins;
+        int ct = 0;
+
+        for(auto inp : ins->inputs())
+        {
+            instruction_ref t_ins;
+            if(inp->get_operator().name() == "unpack_int4")
+            {
+                auto chld = inp->inputs()[0];
+                t_ins     = nm->add_parameter(chld->name() + std::to_string(++ct),
```
pfultz2, line 906: Dont use `to_string` to make the parameters names as that wont work if there is more than 10 parameters. Instead use `param_name(ct++)`.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -878,6 +878,50 @@ struct find_pointwise_mlir
...
+        auto ins      = r.result;
+        auto* mm      = ins->module_inputs().front();
+        module_ref nm = mpm.create_module("int4:" + mm->name());
+        nm->set_bypass();
+
+        std::vector<instruction_ref> x_in;
+        std::unordered_map<instruction_ref, instruction_ref> map_ins;
+        int ct = 0;
+
+        for(auto inp : ins->inputs())
```
pfultz2, line 900: There is no tax on letters in variables, this could be named `input`, to make it clearer.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -878,6 +878,50 @@ struct find_pointwise_mlir
...
+        {
+            instruction_ref t_ins;
+            if(inp->get_operator().name() == "unpack_int4")
+            {
+                auto chld = inp->inputs()[0];
```
pfultz2, line 905: This is missing an `i`, but `child` is strange name for what seems to be the parent. Either way, I think `unpack_input` would be a clearer name.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -878,6 +878,50 @@ struct find_pointwise_mlir
...
+                t_ins     = nm->add_parameter(chld->name() + std::to_string(++ct),
+                                          chld->get_shape().as_standard());
+                t_ins     = nm->add_instruction(inp->get_operator(), t_ins);
```
pfultz2, line 908: Just do `map_ins[input] = nm->add_instruction(input->get_operator(), t_ins)` directly no need for the extra variable.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -878,6 +878,50 @@ struct find_pointwise_mlir
...
+                x_in.push_back(chld);
+            }
+            else
+            {
+                t_ins = nm->add_parameter(inp->name() + std::to_string(++ct),
```
pfultz2, line 913: Same here, you can just do map_ins[input] = nm->add_parameter(...)`.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -930,6 +974,10 @@ void fuse_mlir::apply(module_pass_manager& mpm) const
     {
         match::find_matches(mpm, find_pointwise_mlir{});
     }
+
+    match::find_matches(mpm, find_unpack_int4_mlir_op{});
+    mpm.run_pass(dead_code_elimination{});
```
pfultz2, line 979: Dont run dead_code_elimination here. We already run it after the pass.

src/targets/gpu/fuse_mlir.cpp
lakhinderwalia: Yes, it is a strange name, exactly the opposite. :-)

src/targets/gpu/fuse_mlir.cpp
lakhinderwalia: BTW, it is the same variable :-) And the map_ins is updated in both if/else clause .

pfultz2:
The test failures come from using a global counter to name the modules. As new tests are added that shifts the numbering. #3511 should fix that issue.

test/gpu/fuse_mlir.cpp
```diff
@@ -413,6 +413,101 @@ TEST_CASE(add_dot)
...
+            {"wt_int4", "x"},
+            [=](auto* pm, const auto& inputs) {
+                auto unpk = pm->add_instruction(migraphx::make_op("unpack_int4"), inputs[0]);
+                auto q    = pm->add_instruction(migraphx::make_op("quant_dot"), inputs[1], unpk);
+                return std::make_tuple(q->get_operator(), q);
+            });
+        m->add_return({fused});
+    }
+
+    // todo: fix literal names before comparing p1 and p2.
```
pfultz2, line 448: Due to the global counter the names will change depending on the order of tests. #3511 will fix that issue, so you can just set the name to what it is when you run the test as standalone(ie `./bin/test_gpu_fuse_mlir unpack_int4_dot`) or running with `-c`.

test/gpu/fuse_mlir.cpp
```diff
@@ -413,6 +413,101 @@ TEST_CASE(add_dot)
...
+    EXPECT(true);
```
pfultz2, line 449: These tests need to be enabled to make sure they work.

pfultz2 (changes requested):

test/onnx/gen_onnx.py
```diff
@@ -5992,14 +5992,22 @@ def int4_const_identity_qdq_test():
...
         perm=[1, 0],
     )
 
+    x2_t = helper.make_tensor_value_info('x2', TensorProto.FLOAT16, [2, 4])
+
+    dot_node = helper.make_node(
+        'MatMul',
+        inputs=['t_y', 'x2'],
+        outputs=['y'],
+    )
```
pfultz2, line 6005: The onnx files are not even being tested. Either way this should go into another PR.

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -887,6 +887,48 @@ struct find_pointwise_mlir
...
+            {
+                auto unpack_input     = input->inputs()[0];
+                instruction_ref t_ins = nm->add_parameter(unpack_input->name() + param_name(++ct),
+                                                          unpack_input->get_shape().as_standard());
+                map_ins[input]        = nm->add_instruction(input->get_operator(), t_ins);
+                x_in.push_back(unpack_input);
+            }
+            else
+            {
+                map_ins[input] = nm->add_parameter(input->name() + param_name(++ct),
```
pfultz2, line 921: Remove `input->name()`.

test/onnx/gen_onnx.py
lakhinderwalia: These files are already being used, at least one of them. And we can add additional tests later -- or now. Thanks.

test/gpu/fuse_mlir.cpp
lakhinderwalia: These tests are enabled now. Thanks.

pfultz2 (approved):

turneram (approved):

PR #3513, opened 2024-10-07
Enable mlir input fusion for unary operators and dequantizelinear

pfultz2:
- Enable mlir input fusion for unary operators and dequantizelinear as there doesnt seem to be a perf regression for these cases. This is also needed for int4.
- Fuse input reshapes as well like we do for the input to gemms. 
- Fix a bug where `insert_common_op` would convert everything to the same type which doesnt work for operators like `dequantizelinear`. Instead an option is added to disable this behavior.


test/gpu/fuse_mlir.cpp
```diff
@@ -413,6 +413,122 @@ TEST_CASE(add_dot)
...
+                         auto relu = pm->add_instruction(migraphx::make_op("relu"), inputs[0]);
+                         auto dot  = pm->add_instruction(migraphx::make_op("dot"), relu, inputs[1]);
+                         return std::make_tuple(dot->get_operator(), dot);
+                     });
+        mm->add_return({fused});
+    }
+    EXPECT(p1.sort() == p2.sort());
+}
+
+TEST_CASE(dequantizelinear_dot)
```
lakhinderwalia, line 449: Can't we please have a simpler test for `dequantizelinear_dot`?

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -843,6 +863,21 @@ struct find_pointwise_mlir
             mlir_input_pointwise(match::used_once()).bind("pointwise")));
     }
 
+    static bool is_simple_op(const_module_ref pm, std::initializer_list<std::string> op_names)
+    {
+        auto last = std::prev(pm->end());
+        assert(last->name() == "@return");
```
lakhinderwalia, line 869: Nit: Instead of asserting here, perhaps one could return a `false` here.

lakhinderwalia (approved):

PR #3521, opened 2024-10-10
Fixed logic error of instruction::replace()

tcgu-amd:
Related issue: https://github.com/ROCm/AMDMIGraphX/issues/3110

instruction::replace() performs replace operation on the current instruction then recursively calls itself on any down stream instructions that depends on the output. This is equivalent to performing a pre-order tree walk which is problematic. A simple example would be

op1(op2(op4(x)), op3(op4(x))).

In this example, op1 is a binary operation and op2, op3, op4 are unary operations. The parse tree is simply

  op4
 /   \
op2  op3
 \   /
  op1

An pre-order tree walk on op4.replace will be called in the order of op4 -> op2 -> op1 -> op3

However, at op1, we will encounter an error of shape mismatch because op2 has already been replaced but op3 still hasn't

The correct implementation would be to use BFS, which would yield op4, op2, op3, op1.


src/instruction.cpp
```diff
@@ -59,14 +59,27 @@ instruction::instruction(literal l)
...
 {
-    if(r != result)
-    {
+    if (r != result){
         result = r;
-        for(auto&& ins : output)
+        std::deque<instruction_ref> q;
+        for (instruction_ref ins: output){
+            q.push_back(ins);
+        }
```
pfultz2, line 67: Initialize the `deque` with the `output`:  `std::deque<instruction_ref> q(output.begin(), output.end())`.

src/instruction.cpp
```diff
@@ -59,14 +59,27 @@ instruction::instruction(literal l)
...
-            ins->recompute_shape();
-        }
+            shape new_r = compute_shape(ins->op, ins->arguments, ins->module_args);
+            if (new_r != ins->result)
+            {
+                ins->result = new_r;
+                for(instruction_ref child_ins: ins->output)
+                {
+                    q.push_back(child_ins);
+                }
```
pfultz2, line 80: Use `std::copy` here: `std::copy(ins->output.begin(), ins->output.end(), std::back_inserter(q))`.

src/include/migraphx/instruction.hpp
```diff
@@ -33,6 +33,7 @@
 #include <migraphx/config.hpp>
 #include <string>
 #include <utility>
+#include <deque>
```
pfultz2, line 36: Put the include in the .cpp file.

pfultz2:
Unit tests need to be added for this.

src/instruction.cpp
tcgu-amd: Fixed

src/instruction.cpp
tcgu-amd: Fixed

src/include/migraphx/instruction.hpp
tcgu-amd: Fixed

tcgu-amd:
> Unit tests need to be added for this.

@pfultz2  Should they be included in the scope of this PR? Thanks!

pfultz2:
> Should they be included in the scope of this PR?

Yes, unit tests need to be added to verify the PR works as expected(see [Contributing](https://github.com/ROCm/AMDMIGraphX/wiki/Developer's-Guide#contributing) section). A test case can be added under test/instruction.cpp file.

pfultz2:
Thanks for the test! Just need to fix the licensing and cppcheck issue.

PR #3522, opened 2024-10-10
Integrating generic_float struct for adding datatypes

richagadgil:


pfultz2:
A couple of things:

- Move generic_float to migraphx/generic_float.hpp
- Add a specialization for `std::numeric_limits`
- Add a specialization for `migraphx::is_floating_point`(and remove the half one)
- Add a specialization for `std::common_type`(and remove the half one)
- Add a test/generic_float.cpp with the fp32 tests I wrote and add some tests for operator overloads

The specializations should use the template type like:

```cpp
template<unsigned int E, unsigned int M, unsigned int F>
class numeric_limits<migraphx::generic_float<E, M, F>>
{
...
};
```

pfultz2:
Also, we can use the fp8 template type as well to reduce the number common_type overloads:

```cpp
template<unsigned int E, unsigned int M, unsigned int F, migraphx::fp8::f8_type T, bool FNUZ>
struct common_type<migraphx::generic_float<E, M, F>, migraphx::fp8::float8<T, FNUZ>> : std::common_type<float, float>
{};

template<unsigned int E, unsigned int M, unsigned int F, migraphx::fp8::f8_type T, bool FNUZ>
struct common_type<migraphx::fp8::float8<T, FNUZ>, migraphx::generic_float<E, M, F>> : std::common_type<float, float>
{};
```

src/include/migraphx/half.hpp
```diff
@@ -53,61 +54,56 @@ struct deduce<half_float::detail::expr>
 template <class T>
 using deduce = typename detail::deduce<T>::type;
 
+template <class T>
+struct is_floating_point : std::false_type {};
+
+template <unsigned int E, unsigned int M, unsigned int F>
+struct is_floating_point<generic_float<E, M, F>> : std::true_type {};
```
pfultz2, line 61: Should go into generic_float.hpp.

src/include/migraphx/half.hpp
```diff
@@ -53,61 +54,56 @@ struct deduce<half_float::detail::expr>
...
+
 } // namespace MIGRAPHX_INLINE_NS
 } // namespace migraphx
 
 namespace std {
 
-template <class T>
-struct common_type<migraphx::half, T> : std::common_type<float, T> // NOLINT
+template<unsigned int E, unsigned int M, unsigned int F>
+class numeric_limits<migraphx::generic_float<E, M, F>>
```
pfultz2, line 69: Should go into generic_float.hpp.

src/include/migraphx/half.hpp
```diff
@@ -53,61 +54,56 @@ struct deduce<half_float::detail::expr>
...
+template<unsigned int E, unsigned int M, unsigned int F, class T>
+struct common_type<migraphx::generic_float<E, M, F>, T> : std::common_type<float, T> // NOLINT
 {
-    using type = float;
 };
 
-template <>
-struct common_type<migraphx::half, migraphx::fp8::fp8e5m2>
+template<unsigned int E, unsigned int M, unsigned int F, class T>
+struct common_type<T, migraphx::generic_float<E, M, F>> : std::common_type<float, T> // NOLINT
```
pfultz2, line 91: `common_type` specializations should go into generic_float.hpp.

src/include/migraphx/half.hpp
```diff
@@ -53,61 +54,56 @@ struct deduce<half_float::detail::expr>
...
 namespace std {
 
-template <class T>
-struct common_type<migraphx::half, T> : std::common_type<float, T> // NOLINT
+template<unsigned int E, unsigned int M, unsigned int F>
+class numeric_limits<migraphx::generic_float<E, M, F>>
 {
-};
+    public:
+    static constexpr bool has_infinity = false;
```
pfultz2, line 72: This should be set to true.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,268 @@
...
+ * THE SOFTWARE.
+ */
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+template<unsigned int N>
+constexpr unsigned int all_ones() noexcept
+{
+    return (1 << N) - 1;
```
pfultz2, line 31: I think you will need to write it as `(1u << N) - 1u` to get rid of the tidy warnings.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,350 @@
...
+        return generic_float(float(x) op float(y)); \
+    }
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(*)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(-)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(+)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(/)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(<)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(<=)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(>)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(>=)
```
pfultz2, line 271: Just realized the comparison operators need to return bool instead of `generic_float`. We should make a macro for those operators.

test/float32.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <limits>
+
+using fp32 = migraphx::generic_float<23, 8>;
+
+#define CHECK_FLOAT(x, y) \
+    CHECK(migraphx::float_equal(x, y)); \
+    CHECK(migraphx::float_equal(x, y.to_float())); \
+    CHECK(migraphx::float_equal(fp32{x}, y)); \
+    CHECK(migraphx::float_equal(fp32{x}.to_float(), y.to_float()))
```
pfultz2, line 38: This needs to compare the bits not the float values. 

```cpp
template<class T, class U>
bool bit_equal(const T& x, const U& y)
{
    static_assert(sizeof(T) == sizeof(U));
    using type = std::array<char, sizeof(T);
    return bit_cast<type>(x) == bit_cast<type>(y);
}
```

test/float16.cpp
```diff
@@ -0,0 +1,84 @@
...
+    CHECK(migraphx::float_equal(fp16{x}, y)); \
+    CHECK(migraphx::float_equal(fp16{x}.to_float(), y.to_float()))
+
+
+TEST_CASE(fp16_values)
+{   
+    
+    CHECK_FLOAT(1.0f, fp16{1.0f});
+    CHECK_FLOAT(-1.0f, fp16{-1.0f});
+    CHECK_FLOAT(std::numeric_limits<float>::min(), fp16::min());
```
pfultz2, line 45: This should not be equal.

test/float16.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * THE SOFTWARE.
+ */
+#include <cmath>
+#include <migraphx/float_equal.hpp>
+#include <migraphx/ranges.hpp>
+#include "test.hpp"
+
+#include <limits>
+
+using fp16 = migraphx::half;
```
pfultz2, line 31: Dont create another typedef, just use the one in the half.hpp header. It also would be good to name the test file as half.cpp.

test/float32.cpp
```diff
@@ -0,0 +1,85 @@
...
+    CHECK_FLOAT(-1.0f, fp32{-1.0f});
+    CHECK_FLOAT(std::numeric_limits<float>::min(), fp32::min());
+    CHECK_FLOAT(std::numeric_limits<float>::lowest(), fp32::lowest());
+    CHECK_FLOAT(std::numeric_limits<float>::max(), fp32::max());
+    CHECK_FLOAT(std::numeric_limits<float>::epsilon(), fp32::epsilon());
+    CHECK_FLOAT(std::numeric_limits<float>::denorm_min(), fp32::denorm_min());
+
+    // CHECK_FLOAT(std::numeric_limits<float>::infinity(), fp32::infinity());
+    // CHECK_FLOAT(std::numeric_limits<float>::quiet_NaN(), fp32::qnan());
+    // CHECK_FLOAT(std::numeric_limits<float>::signaling_NaN(), fp32::snan());
```
pfultz2, line 53: These checks should not be disabled.

pfultz2:
For the fp16 tests, we want test similiar to the fp8, but instead of having an array lookup table, we would sample some values into a map and test that:

```cpp
TEST_CASE(check_half_values)
{
    for(auto [x, f] : half_lut)
    {
        auto h = migraphx::bit_cast<migraphx::half>(x);
        if(std::isnan(f))
        {
            CHECK(std::isnan(h));
        }
        else if(std::isinf(f))
        {
            CHECK(std::isinf(h));
            CHECK((h < 0) == (f < 0));
            CHECK(bit_equal(x, migraphx::half(f)));
        }
        else
        {
            CHECK(migraphx::float_equal(float(h), f));
            CHECK(bit_equal(x, migraphx::half(f)));
        }
    }
}
```

I have a map of a thousand or so values we can use for this test. Also we will want to test the numeric limits by checking the bits match what we would expect:

```cpp
TEST_CASE(check_numeric_limits)
{
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::min(), uint16_t{0x0400}));
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::lowest(), uint16_t{0xfbff}));
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::max(), uint16_t{0x7bff}));
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::epsilon(), uint16_t{0x1400}));
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::denorm_min(), uint16_t{0x0001}));
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::infinity(), uint16_t{0x7c00}));
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::quiet_NaN(), uint16_t{0x7fff}));
    CHECK(bit_equal(std::numeric_limits<migraphx::half>::signaling_NaN(), uint16_t{0x7dff}));
}
```

In addition, it would be good to have some tests for overflow and underflow like for `std::numeric_limits<half>::max() + std::numeric_limits<float>::epsilon()`.

test/half.cpp
```diff
@@ -0,0 +1,1118 @@
...
+    }
+}
+
+TEST_CASE(check_flows)
+{
+    // check positive underflow
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::min() * std::numeric_limits<migraphx::half>::min(), migraphx::half(0)));
+
+    // check overflow
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::max() + std::numeric_limits<migraphx::half>::max(), std::numeric_limits<migraphx::half>::infinity()));
```
pfultz2, line 1111: You should also try max + min.

test/half.cpp
```diff
@@ -0,0 +1,1118 @@
...
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::max() / std::numeric_limits<migraphx::half>::epsilon(), std::numeric_limits<migraphx::half>::infinity()));
+
+    // check negative underflow 
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::lowest() + std::numeric_limits<migraphx::half>::lowest(), -std::numeric_limits<migraphx::half>::infinity()));
```
pfultz2, line 1115: Also try lowest - min.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,392 @@
...
+        return result;
+    }
+
+    constexpr float to_float() const noexcept
+    {
+        float32_parts f{};
+
+        f.sign = sign;
+
+        if(exponent == 0)
```
pfultz2, line 110: Good to have a comment that this is for subnormal FPs.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,392 @@
...
+            else
+            {
+                int shift  = 0;
+                f.mantissa = mantissa;
+
+                while((f.mantissa & (1 << MantissaSize)) == 0)
+                {
+                    f.mantissa <<= 1;
+                    shift++;
+                }
```
pfultz2, line 128: This looks like its trying to count the bits. I think it would be a good idea to add a `popcount` function to do this(C++20 provides one but we can't use it yet) using Brian Kernighan's algorithm:

```cpp
constexpr unsigned int popcount(unsigned int v)
{
    unsigned int c = 0;
    for (; v; c++)
        v &= v - 1;
    return c;
}
```

Then you could just do:

```cpp
auto shift = popcount(mantissa);
f.mantissa = mantissa << shift;
```

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,392 @@
...
+
+                f.mantissa &= all_ones<MantissaSize>();
```
pfultz2, line 130: Why is this needed? It look like to me it doesnt do anything. The mantissa is not going to ever be bigger than `all_ones<MantissaSize>()`.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,392 @@
...
+};
+
+constexpr float32_parts get_parts(float f)
+{
+    return migraphx::bit_cast<float32_parts>(f);
+}
+
+
+template<unsigned int MantissaSize, unsigned int ExponentSize, unsigned int Flags = 0>
+struct __attribute__((packed)) generic_float
```
pfultz2, line 75: Why was `__attribute__((packed))` added? Was there some warning?

pfultz2:
You will also need to add pragmas to bit_cast.hpp header to disable the warning for that function.

test/half.cpp
richagadgil: max + min = max. I wonder if there is rounding happening.

test/half.cpp
richagadgil: lowest - min = min (same issue as above). I wonder if there is rounding happening.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,392 @@
...
+                exponent = all_ones<ExponentSize>();
+                mantissa = 0;
+            }
+            else if(e <= 0)
+            {
+                exponent = 0;
+
+                auto shift = diff - int(f.exponent);
+                mantissa =
+                    (f.mantissa | (1 << static_cast<int>(float32_parts::mantissa_width()))) >>
```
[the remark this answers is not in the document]
richagadgil: This is the correct way to convert a subnormal fp. It is a 2-step process:

- Add the implicit leading 1 `(f.mantissa | (1 << static_cast<int>(float32_parts::mantissa_width())))` required for subnormal representations
- Shifting the mantissa left by shift to account for the smaller FP16 exponent range (`shift`), truncating the mantissa (`float32_parts::mantissa_width() - MantissaSize`). Additional shift by `1` bit to account for the added leading 1 in the subnormal case. Total shift: ` shift + float32_parts::mantissa_width() - MantissaSize + 1`

Subnormal numbers have no implicit 1 and represent values of the form 0.m rather than 1.m.

src/include/migraphx/generic_float.hpp
richagadgil: Working on integrating this

src/include/migraphx/generic_float.hpp
richagadgil:
```
error: static assertion failed due to requirement 'sizeof(std::array<char, 2>) == sizeof(migraphx::generic_float<10, 5, 0>)'
   44 |     static_assert(sizeof(To) == sizeof(From));
```

sizeof(migraphx::generic_float<10, 5, 0>) becomes 4 instead of 2 bytes without packing.

src/include/migraphx/generic_float.hpp
pfultz2: Actually, looking at this again, I dont know if this was correct. It looks more like you are trying to count the bits after `MantissaSize` until it reaches a bit that is 0. It does seems a little strange because I would assume you would want to check the highest bit that is set(which it doesnt like thats what this loops does). Perhaps we need the [`countl_zero`](https://en.cppreference.com/w/cpp/numeric/countl_zero) function instead(but we will need to write our own implementation).

richagadgil: I think with std::countl_zero would assume we're counting the amount of 0's from the end of f.mantissa size (which is 23 believe), not MantissaSize (which is 10)

richagadgil: So we would need to pass in MantissaSize to this function. Is that ok?

pfultz2: > I think with std::countl_zero would assume we're counting the amount of 0's from the end of f.mantissa size (which is 23 believe), not MantissaSize (which is 10)

First you can pass `mantissa` instead of `f.mantissa`(because it hasn't been assigned yet). Also, this will count the zeros for an `unsigned int` which is 32 bits. So you would need to subtract off the difference: `(32 - MantissaSize)`.

> So we would need to pass in MantissaSize to this function. Is that ok?

No the function should match the signature used for C++20 so we can replace it with that function in the future.

src/include/migraphx/bit_cast.hpp
```diff
@@ -30,8 +30,12 @@
 #include <migraphx/requires.hpp>
 #include <migraphx/config.hpp>
 
+#if defined(__GNUC__) and !defined(__clang__)
+#define MIGRAPHX_CONST_FOLD(x) (x)
+#else
 // NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
 #define MIGRAPHX_CONST_FOLD(x) (__builtin_constant_p(x) ? (x) : (x))
+#endif
```
pfultz2, line 38: This is not correct. This needs to be defined for gcc. Only the warning should be disabled for the line below where its used.

src/include/migraphx/generic_float.hpp
```diff
@@ -38,6 +37,23 @@ constexpr unsigned int all_ones() noexcept
...
+
+    int count          = 0;
+    constexpr int bits = sizeof(T) * 8;
+
+    while(count < bits && (value & (static_cast<T>(1) << (bits - 1 - count))) == 0)
+    {
+        count++;
+    }
+
+    return count;
```
pfultz2, line 54: I think doing this would be much simpler:

```cpp
unsigned int r = 0;
for (;value != 0; value >>= 1)
    r++;
return 8*sizeof(value) - r;
```

src/include/migraphx/generic_float.hpp
```diff
@@ -121,13 +135,13 @@ struct __attribute__((packed)) generic_float
                 int shift  = 0;
                 f.mantissa = mantissa;
 
-                while((f.mantissa & (1 << MantissaSize)) == 0)
+                if(MantissaSize < float32_parts::mantissa_width())
                 {
-                    f.mantissa <<= 1;
-                    shift++;
+                    shift = MantissaSize - (32 - countl_zero(mantissa));
```
pfultz2, line 143: I see, you are wanting to compute the log2 of the mantissa. You should do `sizeof(mantissa)` instead of `32`.

pfultz2: Then do `sizeof(unsigned int)`.

richagadgil: sizeof(unsigned int) * 8?

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,410 @@
...
+
+    static constexpr int exponent_bias() { return all_ones<7>(); }
+
+    constexpr float to_float() const noexcept { return migraphx::bit_cast<float>(*this); }
+};
+
+constexpr float32_parts get_parts(float f) { return migraphx::bit_cast<float32_parts>(f); }
+
+template <unsigned int MantissaSize, unsigned int ExponentSize, unsigned int Flags = 0>
+struct __attribute__((packed, may_alias)) generic_float
```
CharlieL7, line 68: checking my understanding: `__attribute__((packed, may_alias))` signals to the compiler to make sure the struct's members are contiguous in memory and that we may reinterpret the bits?

src/include/migraphx/half.hpp
```diff
@@ -28,11 +28,12 @@
 #include <half/half.hpp>
 #include <migraphx/config.hpp>
 #include <migraphx/float8.hpp>
+#include <migraphx/generic_float.hpp>
```
CharlieL7, line 31: What does the `ifdef` on L45 do? If that is removed, could we remove the `<half/half.hpp>` include here?

CharlieL7 (approved):
LGTM

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,410 @@
...
+
+// template<unsigned int E, unsigned int M, unsigned int F, migraphx::fp8::f8_type T, bool FNUZ>
+// struct common_type<migraphx::generic_float<E, M, F>, migraphx::fp8::float8<T, FNUZ>> :
+// std::common_type<float, float>
+// {};
+
+// template<unsigned int E, unsigned int M, unsigned int F, migraphx::fp8::f8_type T, bool FNUZ>
+// struct common_type<migraphx::fp8::float8<T, FNUZ>, migraphx::generic_float<E, M, F>> :
+// std::common_type<float, float>
+// {};
```
CharlieL7, line 396: What's the current outlook for the fp8 types? Will we keep the fp8 data structure on ref MIGX or will we specialize generic_float for the fp8 types?

src/include/migraphx/half.hpp
pfultz2: Yea I think we should remove `<half/half.hpp>` includes everywhere from migraphx, and we can remove the ifdef from below.

src/include/migraphx/generic_float.hpp
```diff
@@ -213,15 +214,15 @@ struct __attribute__((packed, may_alias)) generic_float
     {
         generic_float x{};
         x.exponent = all_ones<ExponentSize>();
-        x.mantissa = 1 << (MantissaSize - 2);
+        x.mantissa = 1u << static_cast<unsigned int>(MantissaSize - 2);
```
pfultz2, line 217: Write this as `1u << (MantissaSize - 2u)` instead of using a cast.

src/include/migraphx/generic_float.hpp
```diff
@@ -213,15 +214,15 @@ struct __attribute__((packed, may_alias)) generic_float
...
         return x;
     }
 
     static constexpr generic_float qnan()
     {
         generic_float x{};
         x.exponent = all_ones<ExponentSize>();
-        x.mantissa = 1 << (MantissaSize - 1);
+        x.mantissa = 1u << static_cast<unsigned int>(MantissaSize - 1);
```
pfultz2, line 225: Same here.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,411 @@
...
+        return generic_float(float(x) op float(y));                                            \
+    }
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(*)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(-)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(+)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(/)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(<)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(<=)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(>)
+    MIGRAPHX_GENERIC_FLOAT_BINARY_OP(>=)
```
pfultz2, line 297: Dont use `MIGRAPHX_GENERIC_FLOAT_BINARY_OP` for comparison operators. They should return `bool` not `generic_float`. I would create another macro:

```cpp
#define MIGRAPHX_GENERIC_FLOAT_COMPARE_OP(op)                                         \
    friend constexpr bool operator op(const generic_float& x, const generic_float& y) \
    {                                                                                 \
        return float(x) op float(y);                                                  \
    }
```

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,411 @@
...
+
+    explicit constexpr generic_float(float f = 0.0) noexcept { from_float(get_parts(f)); }
+
+    constexpr generic_float& operator=(float f) noexcept
+    {
+        from_float(get_parts(f));
+        return *this;
+    }
+
+    constexpr generic_float operator-() const noexcept
```
pfultz2, line 84: For completeness, we should add unary `+` as well(which actually does nothing).

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,411 @@
...
+            return false;
+        return std::tie(x.mantissa, x.exponent, x.sign) == std::tie(y.mantissa, y.exponent, y.sign);
+    }
+
+    friend constexpr bool operator!=(const generic_float& x, const generic_float& y)
+    {
+        return not(x == y);
+    }
+
+    constexpr generic_float& operator++() noexcept
```
pfultz2, line 311: For completeness, we should add post-increment as well as pre and post decrement.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,424 @@
...
+    {
+        *this += generic_float(1.0f);
+        return *this;
+    }
+
+    constexpr generic_float operator++(int) noexcept
```
pfultz2, line 323: For the tidy warning, I think it means to return `const generic_float`.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,424 @@
...
+class numeric_limits<migraphx::generic_float<E, M, F>>
+{
+    public:
+    static constexpr bool has_infinity = true;
+    static constexpr migraphx::generic_float<E, M, F> epsilon()
+    {
+        return migraphx::generic_float<E, M, F>::epsilon();
+    }
+
+    static constexpr migraphx::generic_float<E, M, F> quiet_NaN()
```
pfultz2, line 346: I think you can set [`readability-identifier-naming.ConstexprMethodIgnoredRegexp`](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html#cmdoption-arg-ConstexprMethodIgnoredRegexp) to `"quiet_NaN|signaling_NaN"` in the .clang-tidy yaml file to ignore these warnings.

src/include/migraphx/generic_float.hpp
```diff
@@ -320,7 +320,7 @@ struct __attribute__((packed, may_alias)) generic_float
         return *this;
     }
 
-    constexpr generic_float operator++(int) noexcept
+    const generic_float operator++(int) noexcept
```
pfultz2, line 323: It looks like that produces another tidy warning. We should probably do what we did before and suppress the warning in cmake.

src/include/migraphx/generic_float.hpp
```diff
@@ -331,6 +331,7 @@ struct __attribute__((packed, may_alias)) generic_float
 } // namespace MIGRAPHX_INLINE_NS
 } // namespace migraphx
 
+// NOLINT(cert-dcl58-cpp)
```
pfultz2, line 334: The comment needs to go on the line where the class is specialized.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,425 @@
...
+constexpr unsigned int all_ones() noexcept
+{
+    return (1u << N) - 1u;
+}
+
+template <typename T>
+constexpr int countl_zero(T value)
+{
+    unsigned int r = 0;
+    for(; value != 0; value >>= 1)
```
pfultz2, line 45: Probably need to use `0u` and `1u` to fix the tidy warning.

pfultz2:
Why was this closed?

test/float32.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+template <class T, class U>
+bool bit_equal(const T& x, const U& y)
+{
+    static_assert(sizeof(T) == sizeof(U));
+    using type = std::array<char, sizeof(T)>;
+    return migraphx::bit_cast<type>(x) == migraphx::bit_cast<type>(y);
+}
+
+#define MIGRAPHX_CHECK_FLOAT(x, y)     \
```
pfultz2, line 43: Need to add a `// NOLINTNEXTLINE` above to avoid the error.

test/half.cpp
```diff
@@ -0,0 +1,1239 @@
...
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::lowest(), uint16_t{0xfbff}));
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::max(), uint16_t{0x7bff}));
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::epsilon(), uint16_t{0x1400}));
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::denorm_min(), uint16_t{0x0001}));
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::infinity(), uint16_t{0x7c00}));
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::quiet_NaN(), uint16_t{0x7e00}));
+    CHECK(bit_equal(std::numeric_limits<migraphx::half>::signaling_NaN(), uint16_t{0x7d00}));
+}
+
+static const std::map<uint16_t, float> half_lut = {
```
pfultz2, line 54: We probably need to wrap this in a function to fix the tidy warning:

```cpp
const std::map<uint16_t, float>& half_lut()
{
    static const std::map<uint16_t, float> result = { ... };
    return result;
}
```

pfultz2:
Overall this looks, we just need to fix the tidy warnings.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    constexpr generic_float operator+() const noexcept { return *this; }
+
+    constexpr float to_float() const noexcept
+    {
+        float32_parts f{};
+        f.sign = sign;
+
+        if(exponent == 0 and ExponentSize != 8) // subnormal fps
```
pfultz2, line 98: Why are we checking the `ExponentSize` is `8`? The subnormal should be enabled for all sizes.

pfultz2 (approved):
Looks good, just need to fix the CI checks.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,435 @@
...
+    static constexpr unsigned int max_exponent() { return all_ones<8>(); }
+
+    static constexpr int exponent_bias() { return all_ones<7>(); }
+
+    constexpr float to_float() const noexcept { return migraphx::bit_cast<float>(*this); }
+};
+
+constexpr float32_parts get_parts(float f) { return migraphx::bit_cast<float32_parts>(f); }
+
+#pragma pack(push, 1)
```
pfultz2, line 67: This needs to be surrounded by `#ifdef _MSC_VER`.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,435 @@
...
+template <unsigned int MantissaSize, unsigned int ExponentSize, unsigned int Flags = 0>
+struct alignas(1) __attribute__((may_alias)) generic_float
```
pfultz2, line 69: You need the packed attribute for gcc/clang, but this probably breaks windows though. Instead of using macros, doing `[[gnu::packed, gnu::may_alias]]` may work instead.

pfultz2, line 69: Also alignment is wrong. It should be `alignas((MantissaSize+ExponentSize+1)/8)`, I dont know if that compiles.

src/include/migraphx/generic_float.hpp
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    const generic_float operator++(int) noexcept // NOLINT(readability-const-return-type)
+    {
+        generic_float temp = *this;
+        *this += generic_float(1.0f);
+        return temp;
+    }
+};
+#pragma pack(pop)
```
pfultz2, line 337: Also needs a `#ifdef _MSC_VER`.

PR #3526, opened 2024-10-11
Fix dequantizelinaer simple case with no zero point present

TedThemistokleous:
Dequantielinear was failing when only two inputs were given (input, scale) as the zero point type check was applied always instead of when there's 3 inputs. Missed due to missing test cases in verify which would have exercised compute_shape() for the op

Reused existing generated onnx files in parser.

Arose when adding changes for #3445  which uses dequantizelinear before the final dot is produced.

### DIFF
--- a/src/include/migraphx/op/dequantizelinear.hpp
+++ b/src/include/migraphx/op/dequantizelinear.hpp
@@ -49,10 +49,13 @@ struct dequantizelinear
     std::string name() const { return "dequantizelinear"; }
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.same_dims().has(2, 3);
-        if(inputs.size() == 3 and inputs[0].type() != inputs[2].type())
+        if(inputs.size() == 3)
         {
-            MIGRAPHX_THROW("DEQUANTIZELINEAR: Zero point and input should be the same type.");
+            check_shapes{inputs, *this}.same_dims().has(2, 3);
+            if(inputs[0].type() != inputs[2].type())
+            {
+                MIGRAPHX_THROW("DEQUANTIZELINEAR: Zero point and input should be the same type.");
+            }
         }
         return {inputs[1].type(), inputs[0].lens(), inputs[0].strides()};
     }
--- a/test/onnx/verify/dequantizelinear_simple_test.cpp
+++ b/test/onnx/verify/dequantizelinear_simple_test.cpp
@@ -1,0 +1,82 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2024 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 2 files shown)


shivadbhavsar (approved):

src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -49,10 +49,13 @@ struct dequantizelinear
     std::string name() const { return "dequantizelinear"; }
     shape compute_shape(std::vector<shape> inputs) const
     {
-        check_shapes{inputs, *this}.same_dims().has(2, 3);
-        if(inputs.size() == 3 and inputs[0].type() != inputs[2].type())
+        if(inputs.size() == 3)
```
lakhinderwalia, line 52: Thanks for fixing this, but perhaps this Operator is still missing an important check when there are 2 or 3  inputs. And `x_scale` is a 1-D Tensor. We should check this 1-D Tensor against the axis dimension of `X`. I don't see that check in parse_dequantizelinear.cpp either. Is it done somewhere else? Thanks.

lakhinderwalia (approved):
Approving this. If you want to fix something additional (left a comment) if that is an issue, please feel free to fix it as well.

src/include/migraphx/op/dequantizelinear.hpp
TedThemistokleous: I believe it should be. If not I'll make another issue and PR for that case

src/include/migraphx/op/dequantizelinear.hpp
```diff
@@ -49,10 +49,13 @@ struct dequantizelinear
...
         {
-            MIGRAPHX_THROW("DEQUANTIZELINEAR: Zero point and input should be the same type.");
+            check_shapes{inputs, *this}.same_dims().has(2, 3);
```
pfultz2, line 54: Why is this moved down here? We should still check that the inputs are the same dims even when 2 inputs are given. It makes sense up at the top.

src/include/migraphx/op/dequantizelinear.hpp
pfultz2: > And x_scale is a 1-D Tensor. We should check this 1-D Tensor against the axis dimension of X.

No, this is not the onnx `DequantizeLinear` operator. In migraphx, all inputs should have the same dimensions(which is what `check_shapes` checks for). If in onnx, the `x_scale` is a 1-D tensor then a broadcast is inserted to make it have the same dimensions here:

https://github.com/ROCm/AMDMIGraphX/blob/fab88b0eca40b85d8bd53a40ff75acc587ed019d/src/onnx/quantize_dequantize_linear.cpp#L70

pfultz2 (changes requested):
We should still check that all inputs have the same dimensions even when 2 parameters are used.

src/include/migraphx/op/dequantizelinear.hpp
lakhinderwalia: Thanks, Paul. For the onnx case, I do see the 1D tensor shape is being checked here: [quantize_dequantize_linear.cpp](https://github.com/ROCm/AMDMIGraphX/blob/fab88b0eca40b85d8bd53a40ff75acc587ed019d/src/onnx/quantize_dequantize_linear.cpp#L60)

src/include/migraphx/op/dequantizelinear.hpp
TedThemistokleous: That wont be the case though in this operator would it? I guess we broadcast out scale if its a scalar or its 1-d for the case of scale being a tensor

TedThemistokleous: That's where the issue was coming from when invoking dequantize_linear, zero points weren't present and resulted in an error.

pfultz2: What was the error specifically? This allows either 2 or 3 inputs so it shouldn't throw an error for 2 inputs.

TedThemistokleous: image

would invoke

```[   RUN    ] matmulintegertofloat_test
  what():  /code/AMDMIGraphX/src/include/migraphx/check_shapes.hpp:220: same_dims: dequantizelinear: Dimensions do not match```

TedThemistokleous: Unless you want us to always broadcast the scale dimensions (even if its 1-d) but then that means we're inserting a broadcast to remove later. For the scale here, we're never supposed to have the same dimensions between input and scale vectors. Its why I gated this check by input size.

TedThemistokleous:
Not a bug. In MIGraphX we handle this as pointwise so we need to have everything broadcasted to the input to avoid index operations

PR #3527, opened 2024-10-14
Handle more reshapes with the shape transform descriptor

pfultz2:
This handles cases where there is a merge and split used together, such as `{3, 10, 16} -> {3, 40, 2, 2}`. This currently is not supported. We need to keep track of a remainder in the loop to properly do this. Since it is fairly complicated to do, this just reuses common_dims, which already handles the remainder. So it reshapes to the common_dims and then reshapes to the final dimensions.

### DIFF
--- a/src/include/migraphx/shape_transform_descriptor.hpp
+++ b/src/include/migraphx/shape_transform_descriptor.hpp
@@ -76,6 +76,7 @@ struct MIGRAPHX_EXPORT shape_transform_descriptor
 
     bool apply(const std::vector<operation>& ops);
     bool apply_reshape(const std::vector<std::size_t>& rdims);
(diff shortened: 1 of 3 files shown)


shivadbhavsar (approved):
LGTM. 
For already supported cases the following is true right?
For just merge (ie. {3, 10, 16} -> {30, 16}), the common dims stay the same as input dims so the cdims transform will do nothing.
And for just split (ie. {3, 10, 16} -> {3, 10, 4, 4}), the common dims will be the same as rdims and so the cdims transform will apply the transform and then the rdims apply call will do nothing.

pfultz2:
> For already supported cases the following is true right?

Yes thats correct. As an optimization I could probably exit `apply_reshape_impl` early when rdims and the current dimensions are the same.

TedThemistokleous (approved):

PR #3531, opened 2024-10-15
Remove zero point parameter for dequantizelinear when its zero

pfultz2:
This removes the zero point from dequantizelinear when its zero. We dont do this for quantizelinear since its necessary for deducing the output type.


test/simplify_qdq_test.cpp
```diff
@@ -162,7 +165,7 @@ TEST_CASE(qdq_different_scales)
         auto t2     = m1.add_parameter("t2", sh2);
         auto scale1 = m1.add_literal(0.5f);
         auto scale2 = m1.add_literal(0.4f);
-        auto zero   = m1.add_literal(std::int8_t{0});
+        auto zero   = m1.add_literal(std::int8_t{1});
```
lakhinderwalia, line 168: This is no longer a variable called `zero`.  It is initialized to a `1`.  Maybe it should be called `zp`? Also, are you changing its value so it doesn't get eliminated?

src/simplify_qdq.cpp
```diff
@@ -416,6 +416,28 @@ void remove_qdq_pairs(module& m)
     }
 }
 
+void remove_zero_point(module& m)
+{
+    for(auto ins : iterator_for(m))
+    {
+        if(ins->name() != "dequantizelinear")
```
lakhinderwalia, line 423: For case of `quantizelinear`, could this not replace a bunch of `0`s with a scalar broadcast..? Thanks.

src/simplify_qdq.cpp
```diff
@@ -416,6 +416,28 @@ void remove_qdq_pairs(module& m)
...
+            continue;
+        if(ins->inputs().size() != 3)
+            continue;
+        auto zp = ins->inputs().at(2);
+        if(not zp->can_eval())
+            continue;
+        auto a       = zp->eval();
+        bool is_zero = false;
+        a.visit([&](auto t) {
+            is_zero = std::all_of(t.begin(), t.end(), [](auto x) { return float_equal(x, 0); });
```
TedThemistokleous, line 433: Isn't zero point based on the input type here for dequantize? If so why are we using float_equal then? Is this more to cover the case of say fp8 as well as int8, int4, etc?

TedThemistokleous (changes requested):
One question, otherwise I get what you're doing here

src/simplify_qdq.cpp
pfultz2: The visitor will visit all data types, so we need the `float_equal` if the zero point is a floating point(which is the case for fp8).

TedThemistokleous (approved):

TedThemistokleous:
Fix CI but I think this is fine with what you're doing here

PR #3533, opened 2024-10-16
Dont convert squeeze/unsqueeze to reshape when fusing mlir

pfultz2:
Unsqueeze/squeeze/lazy_reshape perserve the layout whereas reshape doesnt. The mlir_op uses this to compute the output shape which could be different after we fuse with mlir. Instead, just convert them to reshape when we are converting to mlir.


shivadbhavsar (approved):
lgtm

PR #3536, opened 2024-10-16
Fix compilation when using STL headers for GCC 14

sohaibnd:
For gcc 13 and earlier, including `<mutex>` would include `<exception>` internally so there would be no need to do so explicitly to use std::exception_ptr . For gcc 14, "Some C++ Standard Library headers have been changed to no longer include other headers that were being used internally by the library" ([source](https://gcc.gnu.org/gcc-14/porting_to.html)). This likely includes `<mutex>` so `<exception>` will have to be included explicitly.

### DIFF
--- a/src/include/migraphx/lexing.hpp
+++ b/src/include/migraphx/lexing.hpp
@@ -27,6 +27,7 @@
 #include <functional>
 #include <string>
 #include <vector>
+#include <algorithm>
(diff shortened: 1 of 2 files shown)


pfultz2 (approved):

sohaibnd:
`<algorithm>` was also needed in src/include/migraphx/lexing.hpp

PR #3539, opened 2024-10-18
Fix GroupQueryAttention Max Sequence Length Propagation

turneram:
Previous version resulted in crashes when using max seqlen < 4096. 
Modified one test to use max seqlen of 2048 to verify.

### DIFF
--- a/src/targets/gpu/include/migraphx/gpu/group_query_attention.hpp
+++ b/src/targets/gpu/include/migraphx/gpu/group_query_attention.hpp
@@ -90,7 +90,7 @@ static inline gqa_parameters init_params(const std::vector<shape>& inputs, const
     auto local_window_size  = v.at("local_window_size").to<std::uint32_t>();
     auto rotary_interleaved = v.at("rotary_interleaved").to<bool>();
     auto scale              = v.at("scale").to<float>();
-    auto present_kv_seqlen  = v.at("present_kv_seqlen").to<std::size_t>();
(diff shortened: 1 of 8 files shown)


TedThemistokleous (approved):

src/targets/gpu/include/migraphx/gpu/group_query_attention.hpp
```diff
@@ -90,7 +90,7 @@ static inline gqa_parameters init_params(const std::vector<shape>& inputs, const
     auto local_window_size  = v.at("local_window_size").to<std::uint32_t>();
     auto rotary_interleaved = v.at("rotary_interleaved").to<bool>();
     auto scale              = v.at("scale").to<float>();
-    auto present_kv_seqlen  = v.at("present_kv_seqlen").to<std::size_t>();
+    auto present_kv_seqlen  = inputs[1].lens().size() == 4 ? inputs[1].lens()[2] : 0;
```
pfultz2, line 93: Why does it get set to 0 for non-4d tensors?

turneram: `qga_rotary_embedding` doesn't take either of the kv-cache tensors as inputs and doesn't use the `present_kv_seqlen` parameter. Since its `inputs[1]` is 2d and the value of `present_kv_seqlen` doesn't matter, I just set it to zero and avoid trying to read `lens()[2]`.

pfultz2 (approved):

PR #3541, opened 2024-10-18
Enable non-packed inputs for mlir

pfultz2:


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -152,14 +152,40 @@ struct mlir_op
...
+                                     assert(y != 0);
+                                     if((x % y) != 0)
+                                         return 0;
+                                     return x / y;
+                                 });
+        return std::equal(stride_ratios.begin() + 1,
+                          stride_ratios.end(),
+                          ns.lens().begin() + 1,
+                          [](auto ratio, auto len) { return ratio >= len; });
+    }
```
pfultz2, line 178: @krzysz00 This is checking if the shapes can be used by mlir(ie can be constructed from slice/reshape/transpose/broadcast). I know there are similar checks in mlir. Are you able to check that this is checking the shapes correctly?

krzysz00: https://github.com/ROCm/rocMLIR/blob/develop/mlir/lib/Dialect/MIGraphX/IR/MIGraphX.cpp#L182 is our logic.

This code _feels_ similar, though I'd need to run through some examples with broadcast dimensions

krzysz00 (commented):
Assuming this doesn't read to non-packed outputs, seems fine to me

shivadbhavsar (approved):

src/targets/gpu/fuse_mlir.cpp
```diff
@@ -152,14 +152,40 @@ struct mlir_op
...
+    static bool is_mlir_compatible(const shape& s)
+    {
+        if(s.standard() or s.packed() or s.scalar() or s.ndim() == 1)
+            return true;
+        auto ns = reorder_shape(s, find_permutation(s));
+        std::vector<std::size_t> stride_ratios;
+        auto last = std::find(ns.strides().begin(), ns.strides().end(), 0);
+        if(*std::prev(last) != 1)
+            return false;
+        std::adjacent_difference(ns.strides().begin(),
```
TedThemistokleous, line 165: Do we want to possible add a parallel execution policy here or are we not concerned about this difference being to large? Looks like adjacent difference has the ability to be parallelized.

TedThemistokleous (approved):
Looks good unless @krzysz00  has any concerns

CI is all green not sure why github isn't reporting that correctly in jenkins here.

PR #3550, opened 2024-10-23
Printing Final MLIR Module

richagadgil:
Takes flag MIGRAPHX_MLIR_DUMP_TO_FILE to save final MLIR module as file.

### DIFF
--- a/src/targets/gpu/jit/mlir.cpp
+++ b/src/targets/gpu/jit/mlir.cpp
@@ -230,7 +230,6 @@ struct mlir_compiler : compiler<mlir_compiler>
 
     static void trace(std::ostream& os, instruction_ref ins)
     {
-        std::cout << "tracing" << std::endl;
         auto shapes = to_shapes(ins->inputs());
         auto* smod  = ins->module_inputs().front();
         os << dump_mlir(*smod, shapes);
--- a/src/targets/gpu/jit/mlir.cpp
+++ b/src/targets/gpu/jit/mlir.cpp
@@ -39,6 +39,7 @@ inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_DUMP_TO_MXR);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_DUMP_TO_FILE);
 
 static module create_pointwise_module(module_ref in_mod)
 {
@@ -212,6 +213,7 @@ struct mlir_compiler : compiler<mlir_compiler>
                                               bool exhaustive) const
     {
         static const auto mxr_loc = string_value_of(MIGRAPHX_MLIR_DUMP_TO_MXR{});
+        static const auto mlir_loc = string_value_of(MIGRAPHX_MLIR_DUMP_TO_FILE{});
 
         auto shapes = to_shapes(ins->inputs());
         auto* smod  = ins->module_inputs().front();
@@ -219,11 +221,16 @@ struct mlir_compiler : compiler<mlir_compiler>
         {
             dump_mlir_to_mxr(*smod, ins->inputs(), mxr_loc);
         }
+        if(not mlir_loc.empty())
+        {
+            dump_mlir(*smod, shapes, mlir_loc);
+        }
         return get_tuning_config_mlir(ctx, *smod, shapes, exhaustive);
     }
 
     static void trace(std::ostream& os, instruction_ref ins)
     {
+        std::cout << "tracing" << std::endl;
         auto shapes = to_shapes(ins->inputs());
         auto* smod  = ins->module_inputs().front();
         os << dump_mlir(*smod, shapes);
--- a/src/targets/gpu/jit/mlir.cpp
+++ b/src/targets/gpu/jit/mlir.cpp
@@ -39,7 +39,7 @@ inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_DUMP_TO_MXR);
-MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_DUMP_TO_FILE);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_DUMP);
 
 static module create_pointwise_module(module_ref in_mod)
 {
@@ -213,7 +213,7 @@ struct mlir_compiler : compiler<mlir_compiler>
                                               bool exhaustive) const
     {
         static const auto mxr_loc = string_value_of(MIGRAPHX_MLIR_DUMP_TO_MXR{});
-        static const auto mlir_loc = string_value_of(MIGRAPHX_MLIR_DUMP_TO_FILE{});
+        static const auto mlir_loc = string_value_of(MIGRAPHX_MLIR_DUMP{});
 
         auto shapes = to_shapes(ins->inputs());
         auto* smod  = ins->module_inputs().front();
@@ -223,7 +223,7 @@ struct mlir_compiler : compiler<mlir_compiler>
         }
         if(not mlir_loc.empty())
         {
-            dump_mlir(*smod, shapes, mlir_loc);
+            dump_mlir_to_file(*smod, shapes, mlir_loc);
         }
         return get_tuning_config_mlir(ctx, *smod, shapes, exhaustive);
     }
--- a/src/targets/gpu/mlir.cpp
+++ b/src/targets/gpu/mlir.cpp
@@ -1153,7 +1153,6 @@ mlir_code_object compile_mlir(const context& migraphx_ctx,
     if(trace)
     {
         const std::lock_guard<std::mutex> lock(mutex);
-        std::cout << "printing mlir" << std::endl;
         std::cout << mlir_print(&mlirOperationPrint, mod_op) << std::endl;
     }
     auto co            = mp.compile(solution);
--- a/src/targets/gpu/mlir.cpp
+++ b/src/targets/gpu/mlir.cpp
@@ -1102,9 +1102,8 @@ static std::string compute_dump_name(const module& m, const std::string& ext)
     std::vector<instruction_ref> sizes;
     for(auto ins : iterator_for(m))
     {
-        if(not contains({"convolution", "dot"}, ins->name()))
-            continue;
-        sizes.insert(sizes.end(), ins->inputs().begin(), ins->inputs().end());
+        if(contains({"convolution", "dot"}, ins->name()))
+            sizes.insert(sizes.end(), ins->inputs().begin(), ins->inputs().end());
     }
     auto name =
         mlir_program::get_symbol_name(m) + "_" + shape::to_sizes_string(to_shapes(sizes)) + ext;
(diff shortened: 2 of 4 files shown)


src/targets/gpu/jit/mlir.cpp
```diff
@@ -39,6 +39,7 @@ inline namespace MIGRAPHX_INLINE_NS {
 namespace gpu {
 
 MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_DUMP_TO_MXR);
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_MLIR_DUMP_TO_FILE);
```
pfultz2, line 42: I would just call this `MIGRAPHX_MLIR_DUMP`.

src/targets/gpu/mlir.cpp
```diff
@@ -1096,6 +1096,37 @@ std::string dump_mlir(module m, const std::vector<shape>& inputs)
...
+
+    std::string mlir_str = mlir_print(&mlirOperationPrint, mod_op);
+    
+    std::ofstream outfile(f, std::ios_base::app);
+    if (outfile.is_open()) {
+        outfile << mlir_str;
+        outfile.close();
+    } else {
+        throw std::runtime_error("Unable to open file");
+    }
```
pfultz2, line 1125: Use `write_string(f, mlir_str)` to write it to a file.

src/targets/gpu/mlir.cpp
```diff
@@ -1096,6 +1096,37 @@ std::string dump_mlir(module m, const std::vector<shape>& inputs)
     return mlir_print(&mlirOperationPrint, mod_op);
 }
 
+std::string dump_mlir(module m, const std::vector<shape>& inputs, const fs::path& location)
+{
+    auto name =
+        mlir_program::get_symbol_name(m) + ".mlir";
+    replace_string_inplace(name, ", ", "_");
+    replace_string_inplace(name, ":", "s");
```
pfultz2, line 1104: This is missing the sizes. We should probably put this in a seperate function to compute the file name and reuse it for this function and `dump_mlir_to_mxr`, something like:

```cpp
static std::string compute_dump_name(const module& m, const std::string ext)
{
    std::vector<instruction_ref> sizes;
    {
        if(not contains({"convolution", "dot"}, ins->name()))
            continue;
        sizes.insert(sizes.end(), ins->inputs().begin(), ins->inputs().end());
    }
    auto name =
        mlir_program::get_symbol_name(m) + "_" + shape::to_sizes_string(to_shapes(sizes)) + ext;
    replace_string_inplace(name, ", ", "_");
    replace_string_inplace(name, ":", "s");
    return name;
}
```

This will need to be called after `rewrite_reduce`.

src/targets/gpu/mlir.cpp
```diff
@@ -1096,6 +1096,37 @@ std::string dump_mlir(module m, const std::vector<shape>& inputs)
     return mlir_print(&mlirOperationPrint, mod_op);
 }
 
+std::string dump_mlir(module m, const std::vector<shape>& inputs, const fs::path& location)
+{
```
pfultz2, line 1100: Should probably add a mutex to avoid writing to a file from multiple threads.

src/targets/gpu/mlir.cpp
```diff
@@ -1096,6 +1097,46 @@ std::string dump_mlir(module m, const std::vector<shape>& inputs)
...
+    {
+        adjust_param_shapes(m, inputs);
+    }
+    rewrite_reduce(m);
+
+    auto name = compute_dump_name(m, ".mlir");
+    auto f    = location / name;
+
+    mlir_program mp;
+    mp.parse(*mr);
```
pfultz2, line 1132: Do `mp.parse(&m)` and remove the `mr` variable.

richagadgil: You mean `mp.parse(m)`?

src/targets/gpu/mlir.cpp
```diff
@@ -1217,10 +1258,7 @@ void dump_mlir_to_mxr(module m,
             continue;
         sizes.insert(sizes.end(), ins->inputs().begin(), ins->inputs().end());
     }
-    auto name =
-        mlir_program::get_symbol_name(m) + "_" + shape::to_sizes_string(to_shapes(sizes)) + ".mxr";
-    replace_string_inplace(name, ", ", "_");
-    replace_string_inplace(name, ":", "s");
+    auto name = compute_dump_name(m, ".mxr");
```
lakhinderwalia, line 1261: Please also print the file name, of the dumped mxr file, if successful. Thanks.

lakhinderwalia: To stdout.

lakhinderwalia (approved):
Approved.

Ideally, the file name (and the directory) should be printed after being successful. And you should be catching exceptions incase somebody gives an improper value to this environment variable. Please see if you could try with some such weird name.

lakhinderwalia:
Another simple enhancement, as I am using this PR to dump some MLIR related stuff (in a multitude of files) is also to have an absolute incrementing number printed in the files. So if there are 10 of them, it would be easier to relate to the progression of the compile -- to differentiate very similar kernels of the graph. Thanks.

lakhinderwalia (changes requested):
Just found out what is likely a bug. For an identical kernel name, if there are many.. those n-1 kernels are getting written over (the same file name) by the last kernel dump. So you do need an incrementing counter number to avoid an overwrite.

causten:
> Just found out what is likely a bug. For an identical kernel name, if there are many.. those n-1 kernels are getting written over (the same file name) by the last kernel dump. So you do need an incrementing counter number to avoid an overwrite.

The objective of the PR is to only document the final solution to be used.  Overwriting the same file would mean there is either not enough to distinguish in the name or over the course of the entire model there are exact duplicates.

pfultz2 (approved):

richagadgil:
@lakhinderwalia so for the first case there was only 1 MLIR and for the `fp16` case there were 3 MLIR files?

lakhinderwalia:
> @lakhinderwalia so for the first case there was only 1 MLIR file and for the `fp16` case there were 3 MLIR files?

Yes.

richagadgil:
Can you share the file names if you don't mind for each?

lakhinderwalia:
Here are the file names, @richagadgil:

```
mlir_quant_convolution_dequantizelinear_add_mul_add_relu_quantizelinear_.mlir
```
vs:
```
mlir_convolution_mul_add_add_relu_1x128x28x28_512x128x1x1.mlir
mlir_convolution_mul_add_add_relu_1x256x14x14_1024x256x1x1.mlir
mlir_convolution_mul_add_add_relu_1x64x56x56_256x64x1x1.mlir
```

pfultz2:
Yea we need to add `quant_convolution` and `quant_dot` to the list of operators.

src/targets/gpu/mlir.cpp
```diff
@@ -1096,6 +1097,46 @@ std::string dump_mlir(module m, const std::vector<shape>& inputs)
     return mlir_print(&mlirOperationPrint, mod_op);
 }
 
+static std::string compute_dump_name(const module& m, const std::string& ext)
+{
+    std::vector<instruction_ref> sizes;
+    for(auto ins : iterator_for(m))
+    {
+        if(not contains({"convolution", "dot"}, ins->name()))
```
pfultz2, line 1105: `quant_convolution` and `quant_dot` need to be added to the list.

lakhinderwalia:
Please take a large model, and then you might find a lot of value in seeing that the dumped files (names) that have an incremental number are more helpful: That would hopefully help one in locating that actual layer, say via Netron. Rather than searching and checking the input sizes. But that is arguable perhaps, but as long as the issue if fixed for the quant operators, I am fine.

src/targets/gpu/mlir.cpp
lakhinderwalia: Why not just search for the term `mlir`, if that is available at this stage?

src/targets/gpu/mlir.cpp
```diff
@@ -1096,6 +1097,46 @@ std::string dump_mlir(module m, const std::vector<shape>& inputs)
...
+    std::cout << "Dumping MLIR file to: " << name << std::endl;
+    auto f    = location / name;
+
+    mlir_program mp;
+    mp.parse(m);
+    auto mod_op = mlirModuleGetOperation(mp.mmodule.get());
+
+    std::string mlir_str = mlir_print(&mlirOperationPrint, mod_op);
+
+    write_string(f, mlir_str);
```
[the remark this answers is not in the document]
richagadgil: It looks like `write_string` in src/file_buffer.cpp doesn't account for exceptions. But that functionality should be added there rather than here, I believe. I can open a seperate PR. @pfultz2

src/targets/gpu/mlir.cpp
```diff
@@ -1258,7 +1258,7 @@ void dump_mlir_to_mxr(module m,
     }
     auto name = compute_dump_name(m, ".mxr");
     std::cout << "Dumping MXR file to: " << name << std::endl;
-    auto f    = location / name;
+    auto f = location / name;
```
lakhinderwalia, line 1261: Will this work on Windows? a forward slash `/`?

src/targets/gpu/mlir.cpp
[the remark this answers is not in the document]
lakhinderwalia: Yes, please, that would be a good idea. But regardless, you might still have to catch the exception here, in that case. Thanks.

May I suggest that you could please print the message to `cout` after calling `write_string()`, and print `f`, rather than `name`. This way if only it is successful do we get the printed message -- with the full path in it.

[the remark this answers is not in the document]
lakhinderwalia: And would the '/' in the file path work for Windows? (Repeat question, I added that somewhere else too).

src/targets/gpu/mlir.cpp
lakhinderwalia: (I think that would be either/or situation). If `mlir` can be searched for, that should be enough.. @pfultz2 ?

src/targets/gpu/mlir.cpp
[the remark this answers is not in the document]
richagadgil: I think it would be better idea to have exception handling in write_string() so it applies to all instances of it.

src/targets/gpu/mlir.cpp
pfultz2: There is no mlir op.

> Should I add quant_convolution and quant_dot to the .mxr (not mlir) file dumping function?

This is just for adding the sizes, which should be the same for both mxr and mlir file.

src/targets/gpu/mlir.cpp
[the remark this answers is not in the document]
pfultz2: Dont add exception handling.

src/targets/gpu/mlir.cpp
pfultz2: Yes it will.

pfultz2:
> Please take a large model, and then you might find a lot of value in seeing that the dumped files (names) that have an incremental number are more helpful:

We already use the sizes in the current develop. This PR is only extending the current dump to dump a mlir instead of mxr.  I think its beyond the scope of this PR to change the current naming scheme.

Also, using this with MXR files, I dont seen any benefit in adding incremental values. And its not really feasible either since these are compiled in parallel(before scheduling) so adding increments would not be in the order of the instructions.

> That would hopefully help one in locating that actual layer, say via Netron. Rather than searching and checking the input sizes.

The actual layer is in the MLIR/MXR file with the sizes(and layouts) so I dont see any need to lookup or search anything in Netron.

lakhinderwalia (approved):

PR #3553, opened 2024-10-24
Fixed instruction::replace() logic. 

tcgu-amd:
The previous fix with BFS doesn't fully work in more complex cases (e.g. it will fail in the newly added test case `check_replace_dag`). This fix implements topological sorting to replace instructions in topological order which should work for all cases.

More details:

In a dummy scenario of `add2(reduce(x), add1(abs(reduce(x)), sin(reduce(x))))`, we will have a dependency tree looking like
```tree
reduce _
        \_abs__
         \_sin__\_add1_
          \_____________\_add2
```
If we call reduce.replace(), BFS will visit the instructions in the following order:
```
reduce -> abs -> sin -> add2 -> add1
```
This will causes an error of shape mismatch at `add2` because it is called before its input `add1`.

Topological sorting the instruction tree will yield:

```
reduce -> sin -> abs -> add1 -> add2
```

Which is the correct order to process the instructions.

This should be able to extend to more complex cases.

### DIFF
--- a/src/include/migraphx/output_iterator.hpp
+++ b/src/include/migraphx/output_iterator.hpp
@@ -72,6 +72,12 @@ auto join_back_inserter(Container& c)
         [&](const auto& r) { c.insert(c.end(), r.begin(), r.end()); });
 }
 
+template <class Container>
+auto push_inserter(Container& c)
+{
+    return make_function_output_iterator([&](const auto& x) { c.push(x); });
+}
 } // namespace MIGRAPHX_INLINE_NS
 } // namespace migraphx
+
 #endif // MIGRAPHX_GUARD_MIGRAPHX_OUTPUT_ITERATOR_HPP
--- a/src/instruction.cpp
+++ b/src/instruction.cpp
@@ -63,17 +63,38 @@ void instruction::replace(const shape& r)
     if(r != result)
     {
         result = r;
-        std::deque<instruction_ref> q(output.begin(), output.end());
-        while(not q.empty())
+        std::deque<instruction_ref> s(output.begin(), output.end());
+        std::deque<instruction_ref> sorted_ins;
+        std::set<instruction*> marked;
+        std::set<instruction*> visited;
+        while(not s.empty())
         {
-            instruction_ref ins = q.front();
-            q.pop_front();
+            instruction_ref ins = s.back();
+            instruction* addr = as_address(ins);
+            if(marked.count(addr))
+            {
+               s.pop_back();
+               continue;
+            }
+            if(visited.count(addr) or ins->output.empty())
+            {
+               marked.insert(addr);
+               sorted_ins.push_front(ins);
+               s.pop_back();
+               continue;
+            }
+            visited.insert(addr);
+            std::copy(ins->output.begin(), ins->output.end(), std::back_inserter(s));
+        }
+        while(not sorted_ins.empty())
+        {
+            instruction_ref ins = sorted_ins.front();
+            sorted_ins.pop_front();
             assert(ins->name() == "@return" or ins->name().front() != '@');
             shape new_r = compute_shape(ins->op, ins->arguments, ins->module_args);
             if(new_r != ins->result)
             {
                 ins->result = new_r;
-                std::copy(ins->output.begin(), ins->output.end(), std::back_inserter(q));
             }
         }
     }
--- a/src/instruction.cpp
+++ b/src/instruction.cpp
@@ -26,6 +26,7 @@
 #include <migraphx/erase.hpp>
 #include <migraphx/module.hpp>
 #include <migraphx/ranges.hpp>
+#include <migraphx/output_iterator.hpp>
 #include <queue>
 
 namespace migraphx {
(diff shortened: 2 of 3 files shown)


pfultz2:
This seems like it will be really slow since it needs topologically sort until end of the model instead of just until the shapes no longer change.

tcgu-amd:
> This seems like it will be really slow since it needs topologically sort until end of the model instead of just until the shapes no longer change.

Yes unfortunately I think this is definitely going to be slower than the previous implementations. I am not quite sure if there's potentially a better approach since we don't know the dependencies of instructions beforehand until after the sort. 

One way I can think of is to take an optimistic approach and perform BFS assuming everything is going to be fine, and on shape mismatch just push the instruction to the back of the queue. Only return the error if all instructions in the queue are shape mismatches. This is a little bit unconventional so I will need to test it to make sure it is going to generate correct results.

pfultz2:
There might be a way to traverse up the inputs to check for dependencies. I would need to think about it more.

pfultz2:
I would think instead you would check if the inputs reaches the instruction and then add that to a `revisit` queue:

```cpp
void instruction::replace(const shape& r)
{
    if(r != result)
    {
        result = r;
        std::deque<instruction_ref> q(output.begin(), output.end());
        std::deque<instruction_ref> revisit;
        std::unordered_set<instruction_ref> visited;
        while(not q.empty())
        {
            instruction_ref ins = q.front();
            q.pop_front();
            if(not visited.insert(ins).second)
                continue;
            assert(ins->name() == "@return" or ins->name().front() != '@');
            shape new_r = compute_shape(ins->op, ins->arguments, ins->module_args);
            if(new_r != ins->result)
            {
                ins->result = new_r;
                for(auto out:ins->outputs())
                {
                    if(any_of(out->inputs(), [&](instruction_ref x) { return x != ins and reaches(ins, x); }))
                    {
                        revisit.push_back(out);
                    }
                    else
                    {
                        q.push_back(ins);
                    }
                }
            }
            if(q.empty())
            {
                q.insert(q.end(), revisit.begin(), revisit.end());
                revisit.clear();
            }
        }
    }
}
```

This would fix the simple case you presented but I am not sure it would handle more complicated cases.

pfultz2:
Actually, I think it might be much simpler if we just use the order in the instruction list as that should already be in order. So we could just use a `priority_queue` instead:

```cpp
struct replace_shape_order
{
    instruction_ref start;

    std::size_t location(instruction_ref x) const
    {
        return std::distance(start, x);
    }

    bool operator()(instruction_ref x, instruction_ref y) const
    {
        return location(x) > location(y);
    }
};

void instruction::replace(const shape& r)
{
    if(r != result)
    {
        result = r;
        auto start = std::find_if(output.front()->inputs().begin(), output.front()->inputs().end(), [&](instruction_ref x) {
            return this == as_address(x);
        });
        assert(as_address(*start) == this);
        std::priority_queue<instruction_ref, std::vector<instruction_ref>, replace_shape_order> q(output, replace_shape_order{*start});
        while(not q.empty())
        {
            instruction_ref ins = q.top();
            q.pop();
            assert(ins->name() == "@return" or ins->name().front() != '@');
            shape new_r = compute_shape(ins->op, ins->arguments, ins->module_args);
            if(new_r != ins->result)
            {
                ins->result = new_r;
                std::copy(ins->output.begin(), ins->output.end(), push_inserter(q));
            }
        }
    }
}
```

tcgu-amd:
> Actually, I think it might be much simpler if we just use the order in the instruction list as that should already be in order. So we could just use a `priority_queue` instead:
> 
> ```c++
> struct replace_shape_order
> {
>     instruction_ref start;
> 
>     std::size_t location(instruction_ref x) const
>     {
>         return std::distance(start, x);
>     }
> 
>     bool operator()(instruction_ref x, instruction_ref y) const
>     {
>         return location(x) > location(y);
>     }
> };
> 
> void instruction::replace(const shape& r)
> {
>     if(r != result)
>     {
>         result = r;
>         auto start = std::find_if(output.front()->inputs().begin(), output.front()->inputs().end(), [&](instruction_ref x) {
>             return this == as_address(x);
>         });
>         assert(as_address(*start) == this);
>         std::priority_queue<instruction_ref, std::vector<instruction_ref>, replace_shape_order> q(output, replace_shape_order{*start});
>         while(not q.empty())
>         {
>             instruction_ref ins = q.top();
>             q.pop();
>             assert(ins->name() == "@return" or ins->name().front() != '@');
>             shape new_r = compute_shape(ins->op, ins->arguments, ins->module_args);
>             if(new_r != ins->result)
>             {
>                 ins->result = new_r;
>                 std::copy(ins->output.begin(), ins->output.end(), push_inserter(q));
>             }
>         }
>     }
> }
> ```

This makes sense! Much more elegant too! I will test it out.

tcgu-amd:
@pfultz2 I pushed a commit with the new solution you proposed. Seems like it is working. Worth noting that there is no std::inserter for priority_queue so we can't use std::copy to insert the instruction outputs.

pfultz2:
> Worth noting that there is no std::inserter for priority_queue so we can't use std::copy to insert the instruction outputs.

You could add one to the migraphx/output_iterator.hpp header:

```cpp
template <class Container>
auto push_inserter(Container& c)
{
    return make_function_output_iterator([&](const auto& x) { c.push(x); });
}
```

pfultz2 (approved):

PR #3563, opened 2024-10-28
Resize Onnx parsing

bpickrel:
The old logic stepped through the args list in order, then assumed the first float-type input was the "Scales".  This led to an error in an obscure case where the "ROIs" input, which is not supported by MigraphX, was *not* left empty and was thereby misinterpreted. 

Changed logic to simply refer to the inputs by array reference; means we're allowed to assume that positional inputs all exist even if they're given as blank `''` quotes.

Updated a few parsing tests to include correct blank positional inputs.

The reason for the older args parsing was apparently an effort to cover the fact that Resize requires one but not both of inputs `scales` and `sizes` which are of different types, and someone thought that positional arguments could not be guaranteed.


kahmed10:
Looks to be fixed from develop already. This PR adds more tests that are passing.

TedThemistokleous (approved):
Fix licensing otherwise LGTM

PR #3566, opened 2024-10-28
matmulnbits zero_point fix

lakhinderwalia:
Currently the matmulnbits parsing introduces a fixed type `uint8_type` for zero_point, and misses `int8_type`.

### DIFF
--- a/src/onnx/parse_matmulnbits.cpp
+++ b/src/onnx/parse_matmulnbits.cpp
@@ -119,7 +119,7 @@ struct parse_matmulnbits : op_parser<parse_matmulnbits>
         }
         else
         {
-            zp = info.add_literal(literal{shape{shape::uint8_type, {1}}, {8}});
+            zp = info.add_literal(literal{shape{args[1]->get_shape().type(), {1}}, {8}});
             zp = info.add_instruction(
                 make_op("multibroadcast", {{"out_lens", b->get_shape().lens()}}), zp);
         }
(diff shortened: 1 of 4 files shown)


kahmed10 (approved):

kahmed10:
https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.MatMulNBits according to the spec zero point can only be uint8/int32/float16/float?

TedThemistokleous:
> > https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.MatMulNBits according to the spec zero point can only be uint8/int32/float16/float?
> 
> And if the zero point isn't specified, it can be inferred to be `uint8` or `int32`, would be my guess. I should fix my test case, and do it for `int32` instead of `int8`.
> 
> @kahmed10, btw, I don't see any type checking in this parser code -- unless it is done somewhere else.

I don't think we can assume it is handled. If things are type constrained you'll need to add that for those inputs

causten (changes requested):
Is this still needed?  Too many failures to approve

lakhinderwalia:
> Is this still needed? Too many failures to approve

These failures have likely nothing to do with this change, @causten, because I added a fix for an `int8` type, which isn't exercised anywhere in the codebase. But this PR is still important to debug our current fusion issues, as I am debugging one right now. It is kept as a back-burner item. Thanks.

lakhinderwalia:
Closing this. This test-model was used to verify a simple int8 test-case; The type checking can be added later, because `int8` case should have been flagged by the parser.

PR #3568, opened 2024-10-29
Rewrite 1x1 convolutions to gemm

pfultz2:
This allows us to use rocblas for some gemms.


src/rewrite_dot.cpp
```diff
@@ -0,0 +1,120 @@
...
+        return false;
+    auto v = ins->get_operator().to_value();
+    if(not all_of(v.at("stride"), [](const value& x) { return x.to<std::size_t>() == 1; }))
+        return false;
+    if(not all_of(v.at("padding"), [](const value& x) { return x.to<std::size_t>() == 0; }))
+        return false;
+    if(not all_of(v.at("dilation"), [](const value& x) { return x.to<std::size_t>() == 1; }))
+        return false;
+    auto w = ins->inputs().at(1)->get_shape();
+    return std::all_of(w.lens().begin() + 2, w.lens().end(), [](std::size_t i) { return i == 1; });
```
shivadbhavsar, line 50: Might be worthwhile to have a 1x1x1 test case too since its allowed here to make sure it generalizes to the 3D case

pfultz2:
Need to investigate the nasnet failure.

CharlieL7 (approved):
LGTM. Is this the only case where doing a convolution as a GEMM is beneficial? Technically any convolution can be done as a GEMM.

pfultz2:
> LGTM. Is this the only case where doing a convolution as a GEMM is beneficial? Technically any convolution can be done as a GEMM.

For other sizes, we need to do windowing(and possibly some padding) which will require an extra copy when we reshape to a gemm, with tensors that are larger then the original input.

causten:
If you wrap the 1x1 with an env variable it could get merged

src/rewrite_dot.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <migraphx/rewrite_dot.hpp>
+#include <migraphx/module.hpp>
+#include <migraphx/matcher.hpp>
+#include <migraphx/make_op.hpp>
+#include <migraphx/permutation.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_DISABLE_REWRITE_DOT);
```
kahmed10, line 34: Can we flip the logic on this for now? Until we know it's always faster or at least as fast as 1x1 convolutions.

kahmed10: also needs documentation update

causten (changes requested):
Need to resolve @kahmed10 comments.  The 1x1 conf to gemms can not be turned by default.  I want to merge the feature that would allow a user to turn on a flag that would force all 1x1 to gemms for performance testing, but as it is written this cant be merged

causten:
Still need to add an entry in to the docs/dev/env_vars.rst  for MIGRAPHX_ENABLE_REWRITE_DOT

docs/dev/env_vars.rst
```diff
@@ -116,6 +116,11 @@ Disables the ``schedule`` pass.
 Set to "1", "enable", "enabled", "yes", or "true" to use.
 Disables the ``fuse_reduce`` pass.
 
+.. envvar:: MIGRAPHX_ENABLE_REWRITE_DOT
+
+Set to "1", "enable", "enabled", "yes", or "true" to use.
```
spolifroni-amd, line 121: When you say "to use", it's not specific enough. It needs to be "to use rewrite dot" or "to enable dot rewriting". People might be putting this through a translator, so a detail like that is important.

spolifroni-amd (changes requested):

docs/dev/env_vars.rst
spolifroni-amd: It shouldn't be "to use" for any of them then. It's ambiguous. The documentation should clarify things for the user since the main reason a user will consult documentation is if they need disambiguation and clarity.

This needs to be clarified. If this is a bigger project than this PR, then a documentation issue needs to be opened. But this needs to be fixed.

causten: Let's fix the "to use" in a separate PR

causten (approved):
Doc added, consistent with the existing format

docs/dev/env_vars.rst
causten: I opened https://github.com/ROCm/AMDMIGraphX/issues/3915 and would love help to make the working better

spolifroni-amd (approved):

PR #3569, opened 2024-10-29
Fuse reshapes on pointwise inputs for mlir output fusion

pfultz2:


src/targets/gpu/fuse_mlir.cpp
```diff
@@ -619,68 +630,62 @@ struct find_mlir_fused_ops
...
-        for(const auto& op : reverse(op_stream))
-        {
-            prev_input = mm->add_instruction(op, {prev_input});
-        }
-        assert(prev_input->get_shape().lens() == x_ins->get_shape().lens());
-        param_map[x_ins] = prev_input; // this is to avoid adding parameter for gemm/conv reshaped
-                                       // input to pointwise in new fused module
+        fuse_input_ops(mm, gemm_based_op->inputs(), &map_ins);
+
         bool gemm_has_multi_outs = gemm_based_op->outputs().size() > 1;
```
shivadbhavsar, line 639: Might be tangential to this PR, but what happens if one of the reshape ops has multi outputs? ie. something like:
```
dot -> reshape -> pointwise -> .... -> dot
          |_____________________________|
```

shivadbhavsar (approved):
Approving, comment is probably a topic for another PR if its even an issue

TedThemistokleous (approved):

src/targets/gpu/fuse_mlir.cpp
pfultz2: The loop below checks that all the reshapes between dot and pointwise have only one output. For the aux inputs to pointwise, it doesnt really matter:

```
┌───┐┌─┐      
│dot││x│      
└┬──┘└┬┘      
 │┌───▽──────┐
 ││reshape   │
 │└┬───────┬─┘
┌▽─▽──────┐│  
│pointwise││  
└┬────────┘│  
┌▽─────────▽┐ 
│convolution│ 
└───────────┘ 
```

After mlir fusion it just becomes: 

```
┌──────────────────────────┐┌─┐ 
│mlir_dot_reshape_pointwise││x│ 
└┬─────────────────────────┘└┬┘ 
 │┌──────────────────────────▽─┐
 ││reshape                     │
 │└┬───────────────────────────┘
┌▽─▽────────┐                   
│convolution│                   
└───────────┘ 
```

DCE wont remove the `reshape` since its used, and the duplicated `reshape` doesnt really matter since it is potentially an aliasing-like operator.

pfultz2:
Tests updated.

PR #3575, opened 2024-10-30
skip passed work

causten:
Still testing, but the idea is to skip a passed test if the current commit hash matches is marked as passed already

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,14 +41,18 @@ def getJobStatus(String variant) {
                            https://api.github.com/repos/ROCm/AMDMIGraphX/commits/${commit_hash}/status """,
             returnStdout: true
         ).trim()
+        echo "Print entire response"
         echo response
 
+        echo "Slurp"
         def jstr = new JsonSlurper().parseText( response )
         jstr.each { println it }
         
         def statuses = jstr.statuses
+        echo "Made it to statuses"
         echo statuses 
 
+        echo "Now planning to find the correct context"
         def contextStatus = statuses.find { it.context == "Jenkins - ${variant}" }
         echo contextStatus
         
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ def getJobStatus(String variant) {
 
     echo "getJobStatus"
  
-    withCredentials([string(credentialsId: "${env.migraphx_ci_creds}", variable: 'GITHUB_TOKEN')]) {
+    withCredentials([usernamePassword(credentialsId: "${env.migraphx_ci_creds}", usernameVariable: 'GITHUB_USERNAME', passwordVariable: 'GITHUB_TOKEN')]) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -45,11 +58,12 @@ def rocmtestnode(Map conf) {
             cmake -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_DEV=On -DCMAKE_EXECUTE_PROCESS_COMMAND_ECHO=STDOUT -DMIGRAPHX_DISABLE_VIRTUAL_ENV=ON ${flags} ..
             git diff
             git diff-index --quiet HEAD || (echo "Git repo is not clean after running cmake." && exit 1)
-            make -j\$(nproc) generate VERBOSE=1
+            #make -j\$(nproc) generate VERBOSE=1
             git diff
             git diff-index --quiet HEAD || (echo "Generated files are different. Please run make generate and commit the changes." && exit 1)
-            make -j\$(nproc) all package check VERBOSE=1
+            #make -j\$(nproc) all package check VERBOSE=1
```
pfultz2, line 64: We should skip the entire stage including docker/cmake builds and even possibly checkout. This should probably be done in the `node(name)` section below, or before if possible since we are just checking a commit hash.

Jenkinsfile
```diff
@@ -45,11 +58,12 @@ def rocmtestnode(Map conf) {
...
             md5sum ./*.deb
+            echo ${GIT_COMMIT_SHA}  > "${name}_PR"
```
pfultz2, line 66: I dont think storing it locally in the workspace is useful since it wont exist on every node. Instead you should check the github commit status, something like this:

```groovy
    withCredentials([string(credentialsId: "${env.migraphx_ci_creds}", variable: 'GITHUB_TOKEN')]) {
        def commit_hash = sh(script: 'git rev-parse --short HEAD', returnStdout: true).trim()
        def response = sh(
            script: """curl -H "Accept: application/vnd.github+json" \
                           -H "Authorization: token ${GITHUB_TOKEN}" \
                           https://api.github.com/repos/ROCm/AMDMIGraphX/commits/${commit_hash}/status""",
            returnStdout: true
        ).trim()
        def statuses = response.statuses
        def contextStatus = statuses.find { it.context == "Jenkins - ${variant}"
        env.COMMIT_PASSED = contextStatus != null && contextStatus.state == 'success'
    }
```

Jenkinsfile
pfultz2: Even better would be to do it in(or around) the `Check image` stage. Set an env variable to a list of the contexes that are `success`, and then we can check that variable before we even allocate a node. It would also reduce the number github API calls we need to make(it would be once per build instead of once per build per stage).

Jenkinsfile
causten: interesting, I was trying to use archieveAritfact since it would be shared across nodes, but I hadn't figured out a way to pull the file down.  I like your idea and will try that

PR #3577, opened 2024-10-30
Enable GEMM/dot for FP8 using hipblasLT

CharlieL7:
Looks to work correctly on Navi4X


src/targets/gpu/target.cpp
```diff
@@ -129,9 +129,11 @@ std::vector<pass> target::get_passes(migraphx::context& gctx, const compile_opti
     unsupported_fp8e4m3fnuz_ops.insert("argmin");
 
     std::set<std::string> unsupported_fp8ocp_ops = {};
-    // TODO update with hipBLASLt support
+    // TODO: remove this when the flag is removed
+#if !MIGRAPHX_ENABLE_HIPBLASLT_GEMM
```
ahsan-ca, line 133: Probably need to do `MIGRAPHX_DECLARE_ENV_VAR(MIGRAPHX_ENABLE_HIPBLASLT_GEMM)` at the top of the file...

ahsan-ca (changes requested):

src/targets/gpu/target.cpp
CharlieL7: I'm not sure what that does exactly. Is the env variable not declared elsewhere?

ahsan-ca: This seems to be the way MIGraphX handles env vars. Doing this provides functions like enabled/disabled etc to work with the env vars, so we can set any of the following values for the env variable:
```
```

Also, I missed to mention that we should do `if(not enabled(MIGRAPHX_ENABLE_HIPBLASLT_GEMM{}))` instead of `!MIGRAPHX_ENABLE_HIPBLASLT_GEMM`, for guarding using the env variable.

lakhinderwalia (approved):
Approved. Is there any test case that verifies this set of Ops. Thanks.

CharlieL7:
> Approved. Is there any test case that verifies this set of Ops. Thanks.

Yes, there are the `test_verify gemm` tests that will use hipblaslt now if the flag is enabled.

src/targets/gpu/target.cpp
lakhinderwalia: Does this environment variable not clash with the one already declared in lowering.cpp? Thanks.

ahsan-ca: I don't think they should clash.

ahsan-ca (approved):

CharlieL7:
Created an issue tracking the preprocessor conditional I added: https://github.com/ROCm/AMDMIGraphX/issues/3592

lakhinderwalia (approved):

ahsan-ca (approved):

CharlieL7:
I tried using CMake's `check_type_size` or `check_symbol_exists` to detect if the types are in the hipblaslt version. The `HIP_R_F8...`are not really types so it doesn't work. I didn't get `check_symbol_exists` to work. The last thing I found was using `check_source_compiles` but that's way too troublesome to use.

pfultz2:
The types are defined as enum in hip not in hipblaslt:

https://github.com/ROCm/HIP/blob/3d60bd3a6415c280bd1fe63767ae8e10eea4d2d1/include/hip/library_types.h#L61

So we shouldn't check the hipblaslt version. It looks like hipblaslt already checks the hip version and then defines `ROCM_USE_FLOAT8`:

https://github.com/ROCm/hipBLASLt/blob/b2adca84509dd31e31b8f42044389128d199b62e/library/include/hipblaslt.h#L67

Which looks like it is to enable the non-fnuz types:

https://github.com/ROCm/hipBLASLt/blob/b2adca84509dd31e31b8f42044389128d199b62e/library/src/amd_detail/rocblaslt/src/utility.cpp#L79

So we can probably use `#ifdef ROCM_USE_FLOAT8` for this.

CharlieL7:
Works well with the variable

PR #3578, opened 2024-10-30
Implement BF16 using generic_float class

richagadgil:
Uses generic float class (#3522) to create bf16 class.

BF16 has 1 sign bit, 8 bits for the exponent, and 7 bits for the mantissa: `bf16 = migraphx::generic_float<7, 8>;`

Summary of changes:
1. generic_float.cpp : Change the subnormal (when exponent==0) conversion to differentiate between FP16 and BF16 types
2. migraphx.h, shape.hpp, hip_gemm_impl.cpp, gemm_impl.cpp: Add BF16 shape type
3. type_traits.hpp: Add traits for BF16 type
4. tests/: Add tests for BF16 type


src/include/migraphx/bf16.hpp
```diff
@@ -0,0 +1,105 @@
...
+};
+
+template <>
+struct common_type<migraphx::bf16, migraphx::bf16>
+{
+    using type = migraphx::bf16;
+};
+
+template <>
+struct common_type<migraphx::bf16, migraphx::half>
```
pfultz2, line 98: So #3606 enables specializing using the template so there is no need for these specializations and can be removed.

pfultz2:
Probably should add some ref tests that use bf16 type.

src/include/migraphx/bf16.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <migraphx/config.hpp>
+#include <migraphx/float8.hpp>
+
+namespace migraphx {
+inline namespace MIGRAPHX_INLINE_NS {
+
+using bf16 = migraphx::generic_float<7, 8>;
+
+template <class T>
+using deduce = typename detail::deduce<T>::type;
```
pfultz2, line 38: Remove this, its already defined in half.hpp, and its not necessary for bf16.

src/include/migraphx/bf16.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#ifndef MIGRAPHX_GUARD_RTGLIB_BF16_HPP
+#define MIGRAPHX_GUARD_RTGLIB_BF16_HPP
+
+#include <migraphx/half.hpp>
```
pfultz2, line 28: Dont include half, just include generic_float.hpp.

src/include/migraphx/bf16.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <migraphx/config.hpp>
+#include <migraphx/float8.hpp>
```
pfultz2, line 30: Dont include float8.

src/include/migraphx/shape.hpp
```diff
@@ -52,6 +53,7 @@ struct MIGRAPHX_EXPORT shape
 #define MIGRAPHX_SHAPE_VISIT_TYPES(m) \
     m(bool_type, bool) \
     m(half_type, half) \
+    m(bf16_type, bf16) \
```
CharlieL7, line 56: It looks like this changes the order of the visit types. @TedThemistokleous do you know if this breaks ABI?

test/op_shape_test.cpp
```diff
@@ -2166,7 +2166,7 @@ TEST_CASE(multinomial)
     migraphx::shape s1{migraphx::shape::float_type, {1, 2}};
     migraphx::shape s2{migraphx::shape::float_type, {3, 4}};
     migraphx::shape s3{migraphx::shape::float_type, {1, 4}};
-    int dtype = 2;
+    int dtype = 3;
```
CharlieL7, line 2169: These should be using the `migraphx::shape`. Should work with `migraphx::shape::float_type`.

src/include/migraphx/shape.hpp
CharlieL7: Can probably move the new type to the bottom of the list to fix if it does break abi

TedThemistokleous: I think the change should happen in migraphx.h

src/py/migraphx_py.cpp
```diff
@@ -180,6 +180,17 @@ struct npy_format_descriptor<migraphx::fp8::fp8e5m2>
     static constexpr auto name() { return _("fp8e5m2"); }
 };
 
+template <>
+struct npy_format_descriptor<migraphx::bf16>
+{
+    static std::string format()
+    {
+        // TODO: no standard format in numpy for bf16
```
TedThemistokleous, line 188: Should this be an issue or is this tracked somewhere already?

src/api/include/migraphx/migraphx.h
```diff
@@ -35,6 +35,7 @@
 #define MIGRAPHX_SHAPE_VISIT_TYPES(m) \
     m(bool_type, bool) \
     m(half_type, half) \
+    m(bf16_type, bf16) \
```
pfultz2, line 38: This breaks ABI, the bf16 needs to be at the end.

test/op_shape_test.cpp
pfultz2: We should not be changing the value of `float_type`.

test/op_shape_test.cpp
```diff
@@ -2175,7 +2175,7 @@ TEST_CASE(multinomial_0size_input)
 {
     migraphx::shape s1{migraphx::shape::float_type, {1, 2}};
     migraphx::shape s2{migraphx::shape::float_type, {}};
-    int dtype = 2;
+    int dtype = 3;
```
pfultz2, line 2178: Revert this, the value should not be changed.

src/include/migraphx/shape.hpp
richagadgil: Does it break ABI or not?

CharlieL7 (approved):

TedThemistokleous (approved):

pfultz2 (approved):

pfultz2:
For the CI failures, just add an if statement to skip bf16 for now in the `compile_math` case in jit.cpp. That should get it passing and we can work on enabling it on a later PR.

PR #3587, opened 2024-11-04
Dont use mixed layouts with convolution

pfultz2:
This refactors the `layout_nhwc` to either make the weights the same layout as the inputs or sets the input and weights to nhwc. Since this only applies to convolution, I renamed the pass to `layout_convolution` instead.


TedThemistokleous (approved):

aarushjain29 (approved):
./build/bin/driver perf /models/slim/vgg16_i1.pb --int8

Rate: 1695.36 inferences/sec
Total time: 0.589847ms
Total instructions time: 0.911128ms
Overhead time: 0.0127777ms, -0.321281ms
Overhead: 2%, -54%

PR #3591, opened 2024-11-05
Generic "truncate_float" class for bf16 and fp16 quantization

richagadgil:


src/include/migraphx/bf16.hpp
```diff
@@ -0,0 +1,105 @@
...
+    using type = migraphx::bf16;
+};
+
+template <>
+struct common_type<migraphx::bf16, migraphx::generic_float<10, 5>>
+{
+    using type = float;
+};
+
+} // namespace std
```
pfultz2, line 103: The common_type specializations should not be needed and can be removed.

PR #3593, opened 2024-11-06
Update finalize function for hipBLASLt to find tuned solution only when solution index is 0.

ahsan-ca:
Currently, finalize function runs tuning for all solution indices.
This PR modifies the function to run tuning only when
solution index is '0'. For all other solution indices, it runs validate.

This PR also fixes a bug in functionality of hipBLASLt validate function.

### DIFF
--- a/src/targets/gpu/hip_gemm_impl.cpp
+++ b/src/targets/gpu/hip_gemm_impl.cpp
@@ -462,10 +462,9 @@ struct hip_gemm_impl
     int32_t
     validate(context& ctx, const std::vector<argument>& input_args, int32_t solution_idx) // const
     {
-        hipblasStatus_t check_valid(HIPBLAS_STATUS_SUCCESS);
         auto common_args = create_hipblaslt_args_common(ctx, input_args, solution_idx);
-        check_valid      = hipblaslt_invoke(&hipblasLtMatmul, common_args);
-        if(check_valid == HIPBLAS_STATUS_SUCCESS)
+        auto check_valid = hipblaslt_invoke(&hipblasLtMatmul, common_args);
+        if(check_valid != HIPBLAS_STATUS_SUCCESS)
         {
             std::cerr << "WARNING:  tuned solution is invalid; reverting to default" << std::endl;
             return 0;
@@ -633,9 +632,19 @@ int32_t hip_gemm_finalize(context& ctx,
                           float beta,
                           int32_t solution_idx)
 {
-    auto gemm_item   = hip_gemm_impl(output_shape, input_shapes, alpha, beta);
-    int32_t solution = gemm_item.tune(ctx, input_shapes);
-    hip_gemm_save_solution(ctx, output_shape, input_shapes, solution_idx);
+    int solution;
+    auto gemm_item = hip_gemm_impl(output_shape, input_shapes, alpha, beta);
+    if(solution_idx == 0)
+    {
+        solution = gemm_item.tune(ctx, input_shapes);
+        hip_gemm_save_solution(ctx, output_shape, input_shapes, solution);
+    }
+    // If a tuned solution index is already given, don't tune again but validate
+    // in case the data was tuned with a different hipBLAS version.
+    else
+    {
+        solution = gemm_item.validate(ctx, input_shapes, solution_idx);
+    }
     return solution;
 }
 
--- a/src/targets/gpu/hip_gemm_impl.cpp
+++ b/src/targets/gpu/hip_gemm_impl.cpp
@@ -632,20 +632,19 @@ int32_t hip_gemm_finalize(context& ctx,
                           float beta,
                           int32_t solution_idx)
 {
-    int solution;
     auto gemm_item = hip_gemm_impl(output_shape, input_shapes, alpha, beta);
     if(solution_idx == 0)
     {
-        solution = gemm_item.tune(ctx, input_shapes);
-        hip_gemm_save_solution(ctx, output_shape, input_shapes, solution);
+        solution_idx = gemm_item.tune(ctx, input_shapes);
+        hip_gemm_save_solution(ctx, output_shape, input_shapes, solution_idx);
     }
     // If a tuned solution index is already given, don't tune again but validate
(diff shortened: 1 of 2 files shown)


ahsan-ca:
Windows build failures seem unrelated.

src/targets/gpu/hip_gemm_impl.cpp
```diff
@@ -633,9 +632,19 @@ int32_t hip_gemm_finalize(context& ctx,
...
+    if(solution_idx == 0)
+    {
+        solution = gemm_item.tune(ctx, input_shapes);
+        hip_gemm_save_solution(ctx, output_shape, input_shapes, solution);
+    }
+    // If a tuned solution index is already given, don't tune again but validate
+    // in case the data was tuned with a different hipBLAS version.
+    else
+    {
+        solution = gemm_item.validate(ctx, input_shapes, solution_idx);
```
bpickrel, line 646: This defaults the solution if it doesn't validate.  If the given index was tuned but doesn't validate, why wouldn't we want to tune again?

src/targets/gpu/hip_gemm_impl.cpp
```diff
@@ -633,9 +632,19 @@ int32_t hip_gemm_finalize(context& ctx,
                           float beta,
                           int32_t solution_idx)
 {
-    auto gemm_item   = hip_gemm_impl(output_shape, input_shapes, alpha, beta);
-    int32_t solution = gemm_item.tune(ctx, input_shapes);
-    hip_gemm_save_solution(ctx, output_shape, input_shapes, solution_idx);
+    int solution;
```
pfultz2, line 635: Remove this variable and use `solution_idx` variable instead.

src/targets/gpu/hip_gemm_impl.cpp
pfultz2: > If the given index was tuned but doesn't validate, why wouldn't we want to tune again?

This is mainly for loading a compiled model from a file as the solution may not be valid with a different version of hipblaslt. Since we are loading the model it should happen somewhat quickly, so we should not be doing benchmarking.

ahsan-ca: > This defaults the solution if it doesn't validate. If the given index was tuned but doesn't validate, why wouldn't we want to tune again?

@bpickrel Just to clarify, did you mean something like this?
```
    // If a tuned solution index is already given, don't tune again but validate
    // in case the data was tuned with a different hipBLASLt version.
    if(solution_idx != 0)
    {
        solution_idx = gemm_item.validate(ctx, input_shapes, solution_idx);
    }
    if(solution_idx == 0)
    {
        solution_idx = gemm_item.tune(ctx, input_shapes);
        hip_gemm_save_solution(ctx, output_shape, input_shapes, solution_idx);
    }
```

src/targets/gpu/hip_gemm_impl.cpp
ahsan-ca: Done.

src/targets/gpu/hip_gemm_impl.cpp
pfultz2: Yes

pfultz2: We shouldnt retune if the solution is invalid.

src/targets/gpu/hip_gemm_impl.cpp
```diff
@@ -462,10 +462,9 @@ struct hip_gemm_impl
...
     validate(context& ctx, const std::vector<argument>& input_args, int32_t solution_idx) // const
     {
-        hipblasStatus_t check_valid(HIPBLAS_STATUS_SUCCESS);
         auto common_args = create_hipblaslt_args_common(ctx, input_args, solution_idx);
-        check_valid      = hipblaslt_invoke(&hipblasLtMatmul, common_args);
-        if(check_valid == HIPBLAS_STATUS_SUCCESS)
+        auto check_valid = hipblaslt_invoke(&hipblasLtMatmul, common_args);
+        if(check_valid != HIPBLAS_STATUS_SUCCESS)
         {
             std::cerr << "WARNING:  tuned solution is invalid; reverting to default" << std::endl;
```
bpickrel, line 469: How can you ever get here?  hipblaslt_invoke() throws an exception if any result besides HIPBLAS_STATUS_SUCCESS occurs.  Is there code I don't see that catches an exception and takes us back here?

ahsan-ca: I think you are right. Thanks for catching this!

ahsan-ca: I have made the change to `hipblaslt_invoke()` to take an optional parameter that controls if we silently return the status or throw an exception with an error message.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
```diff
@@ -76,14 +76,17 @@ inline auto hipblaslt_invoke(F f, Ts... xs)
 }
 
 template <class F, class Pack, class... Ts>
```
bpickrel, line 78:
```suggestion
// Invoke a Hipblaslt call.  If used to validate a call, set fatal_error = false to prevent
// throwing an exception on failure.
template <class F, class Pack, class... Ts>
```

bpickrel (approved):
One suggested comment add, but looks OK to me now.

src/targets/gpu/include/migraphx/gpu/hipblaslt.hpp
ahsan-ca: Thanks for the suggestion, I have added the comment.

ahsan-ca:
> Code looks good. I'm assuming this is faster to compile because we're not running tune every time?

Yes, this should be faster because previously we were running tuning regardless of the solution index provided. Now we only do tuning when the solution index provided is zero.

> Also, does a solution index of zero always mean it needs tuning?

We do tuning if the solution index provided is zero. However, we do not try to tune if the non-zero solution index is invalidated by the `validate()` function to `solution index 0`.

